PR #722, opened 2021-07-15
Support a light terminal theme

siggisim:
Adds support for a light terminal theme that the user can switch to in `Personal Settings`.
Screen Shot 2021-07-15 at 2 36 05 PM

Does so by replacing the simple `denseMode` boolean with a simple `UserPreferences` object that manages dense mode and this new log viewer theme.

Also cleans up a few things I bumped into along the way:
- Updates invocation query style to match
- Renames ActionCardComponent => InvocationActionCardComponent (we have two components named ActionCardComponent right now).
- Cleans up some unused imports

---

**Version bump**: Minor

### DIFF
--- a/app/preferences/preferences.ts
+++ b/app/preferences/preferences.ts
@@ -8,16 +8,16 @@ const terminalThemeLightValue = "LIGHT";
 declare var window: any;
 
 export default class UserPreferences {
-
   handlePreferencesChanged: () => void;
 
   constructor(handlePreferencesChanged: () => void) {
     this.handlePreferencesChanged = handlePreferencesChanged;
   }
 
-  denseModeEnabled = viewModeKey in window.localStorage
-    ? window.localStorage.getItem(viewModeKey) == denseModeValue
-    : window.buildbuddyConfig && window.buildbuddyConfig.default_to_dense_mode;
+  denseModeEnabled =
+    viewModeKey in window.localStorage
+      ? window.localStorage.getItem(viewModeKey) == denseModeValue
+      : window.buildbuddyConfig && window.buildbuddyConfig.default_to_dense_mode;
   lightTerminalEnabled = window.localStorage.getItem(terminalThemeKey) == terminalThemeLightValue || false;
 
   toggleDenseMode() {
--- a/app/preferences/preferences.ts
+++ b/app/preferences/preferences.ts
@@ -1,0 +1,34 @@
+const viewModeKey = "VIEW_MODE";
+const denseModeValue = "DENSE";
+const comfyModeValue = "COMFY";
+
+const terminalThemeKey = "TERMINAL_THEME";
+const terminalThemeLightValue = "LIGHT";
+
+declare var window: any;
+
+export default class UserPreferences {
+
+  handlePreferencesChanged: () => void;
+
+  constructor(handlePreferencesChanged: () => void) {
+    this.handlePreferencesChanged = handlePreferencesChanged;
+  }
+
+  denseModeEnabled = viewModeKey in window.localStorage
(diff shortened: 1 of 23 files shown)


app/preferences/preferences.ts
```diff
@@ -0,0 +1,34 @@
...
+    window.localStorage.setItem(viewModeKey, this.denseModeEnabled ? denseModeValue : comfyModeValue);
+    this.handlePreferencesChanged();
+  }
+
+  toggleLightTerminal() {
+    this.lightTerminalEnabled = !this.lightTerminalEnabled;
+    window.localStorage.setItem(terminalThemeKey, this.lightTerminalEnabled ? terminalThemeLightValue : "");
+    this.handlePreferencesChanged();
+  }
+}
```
bduffany, line 34: nit: needs formatting with prettier? (not sure if it fixes trailing newline)

enterprise/app/root/root.tsx
```diff
@@ -83,10 +81,8 @@ export default class EnterpriseRootComponent extends React.Component {
     capabilities.didNavigateToPath();
   }
 
-  handleToggleDenseClicked() {
-    let newDenseMode = !this.state.denseMode;
-    this.setState({ ...this.state, denseMode: newDenseMode });
-    window.localStorage.setItem(denseModeKey, newDenseMode ? denseModeValue : "");
+  handlePreferencesChanged() {
+    this.setState({ ...this.state, preferences: this.state.preferences });
```
bduffany, line 85: I think you need to clone the object, otherwise child components will see that the prop didn't change, and won't re-render

bduffany (changes requested):
Nice!

enterprise/app/root/root.tsx
bduffany: Might want to make UserPreferences a global like we do with rpcService etc. then expose an observable preferences object which is just a plain old JS object -- that way you can listen to changes using the `subscribe()` API and then clone it easily with `this.setState({ preferences: {...preferences.values}})`

app/preferences/preferences.ts
siggisim: Looks like my prettier was broken, fixed

enterprise/app/root/root.tsx
siggisim: According to [the docs](https://reactjs.org/docs/react-component.html#setstate), setState will always re-render unless you implement shouldComponentUpdate which we don't anywhere.

> setState() will always lead to a re-render unless shouldComponentUpdate() returns false. If mutable objects are being used and conditional rendering logic cannot be implemented in shouldComponentUpdate(), calling setState() only when the new state differs from the previous state will avoid unnecessary re-renders.

bduffany: Ah, I'm thinking of `React.PureComponent` which we aren't using anywhere yet. Pure components don't re-render if their props don't change -- https://codesandbox.io/s/interesting-nobel-nmkgu?file=/src/App.js

Pure components are an easy way to improve performance if re-render time becomes a bottleneck, and it would be great if we could avoid introducing patterns that break from the "reactive" model and make it harder to use pure components down the road (like classes whose reference doesn't change, but their internal state does). As of now I think everything could be made pure, except components which use InvocationModel.

siggisim: I think components that have 'User' as a prop (and their children which is pretty much every component) would also be an issue if we wanted to use PureComponents.

There are also quite a few components that have mutable objects as part of their state. 

It's a nice goal, but we haven't focused on immutability in state / props. If render performance becomes an issue, we're always going to have to go component by component and enable that optimizations for components that support it - and make things immutable before converting components that don't.

bduffany (approved):

PR #724, opened 2021-07-16
Integrate exhaustive analyzer.

vadimberezniker:
---

**Version bump**: None


enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -152,6 +152,7 @@ func (s *InvocationSearchService) QueryInvocations(ctx context.Context, req *inp
 		q.SetOrderBy("created_at_usec", sort.Ascending)
 	case inpb.InvocationSort_UPDATED_AT_USEC_SORT_FIELD:
 		q.SetOrderBy("updated_at_usec", sort.Ascending)
+	default:
 	}
```
tylerwilliams, line 156: how bout:

sort := defaultSortParams()
if req.Sort != nil {
	sort = req.Sort
}

switch sort.SortField {
	case inpb.InvocationSort_CREATED_AT_USEC_SORT_FIELD:
		q.SetOrderBy("created_at_usec", sort.Ascending)
	case inpb.InvocationSort_UPDATED_AT_USEC_SORT_FIELD:
		q.SetOrderBy("updated_at_usec", sort.Ascending)
	case inpb.InvocationSort_UNKNOWN_SORT_FIELD:
        // do nothing -- sort is not required
	}

tylerwilliams (approved):
LGTM (minus the empty default case)

enterprise/server/invocation_search_service/invocation_search_service.go
vadimberezniker: I added a case for InvocationSort_UNKNOWN_SORT_FIELD and marked it as an unexpected event.
We don't expect to receive a a search request with a sort message but with sort field not set so we should never reach this case.

PR #725, opened 2021-07-16
Shutdown the executor gracefully on scale-down.

tylerwilliams:
Ensure in-flight work is cancelled before the shutdown and no more messages are sent.

### DIFF
--- a/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
+++ b/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
@@ -181,8 +181,8 @@ func NewPriorityTaskScheduler(env environment.Env, exec *executor.Executor, opti
 	// shutdown timeout, at which point we hard-cancel it.
 	env.GetHealthChecker().RegisterShutdownFunction(func(ctx context.Context) error {
 		qes.mu.Lock()
-		defer qes.mu.Unlock()
 		qes.shuttingDown = true
+		qes.mu.Unlock()
 		log.Debug("PriorityTaskScheduler received shutdown signal")
 
 		deadline, ok := ctx.Deadline()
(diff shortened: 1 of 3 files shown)


enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -157,41 +158,45 @@ func NewPriorityTaskScheduler(env environment.Env, exec *executor.Executor, opti
...
-			qes.mu.Lock()
-			activeTasks := len(qes.activeTaskCancelFuncs)
-			qes.mu.Unlock()
-			if activeTasks == 0 {
-				break
-			}
-			time.Sleep(100 * time.Millisecond)
+		deadline, ok := ctx.Deadline()
+		if !ok {
+			rootCancel()
```
vadimberezniker, line 189: Would this case ever be hit?

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -157,41 +158,45 @@ func NewPriorityTaskScheduler(env environment.Env, exec *executor.Executor, opti
...
-	// but not started yet, ensuring that someone else will have a chance to get it.
+
+	// This func ensures that we don't attempt to claim work that was
+	// enqueued but not started yet, allowing another executor a chance
+	// to complete it. This is client-side "graceful" stop -- we stop
+	// processing queued work as soon as we receive a shutdown signal, but
+	// permit in-progress work to continue, up until just before the
+	// shutdown timeout, at which point we hard-cancel it.
 	env.GetHealthChecker().RegisterShutdownFunction(func(ctx context.Context) error {
 		qes.mu.Lock()
```
vadimberezniker, line 182: It looks like the lock is being hold while waiting for the shutdown deadline.

vadimberezniker (changes requested):

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
tylerwilliams: my bad; dashed this off before lunch without testing first oO

vadimberezniker (approved):
LGTM

vadimberezniker (approved):

PR #726, opened 2021-07-16
Fix file arrays in buildfix scripts

bduffany:
Tested on Linux and Mac.

We should really just switch to Python... but then we have a dep on Python. Probably OK, but let's think about that later.

---

**Version bump**: TODO


vadimberezniker (approved):
LGTM

PR #727, opened 2021-07-16
Take resource buffer into account when determining if a task will fit on an executor

siggisim:
---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #728, opened 2021-07-16
Skip non-default partition directories when loading default partition.

vadimberezniker:
Also fixed the max size config option since it was renamed from max_size
to max_size_bytes.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #729, opened 2021-07-17
Fix timezone formatting on disk cache statusz page.

vadimberezniker:
MST is the expected timezone placeholder.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #730, opened 2021-07-19
Fix filesystem permissions for local docker-based executor

bduffany:
Add a config option to inherit the uid and gid of the current user, and enable it for local development.

Before this change, when running the executor locally, all files are written with uid=0 and gid=0, causing various permissions errors.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #732, opened 2021-07-19
Only skip partition dirs when initializing default partition.

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #735, opened 2021-07-19
Use the public GetExecutionNodes in tests instead of test-only GetAllExecutionNodes

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #736, opened 2021-07-19
Remove the test timeout grace period

siggisim:
This should no longer be needed.

---

**Version bump**: Minor


tylerwilliams (approved):
LGTM

PR #737, opened 2021-07-19
Support jailer and networking

tylerwilliams:
This change runs firecracker inside of the jailer and also enables networking for VMs.

The [jailer](https://github.com/firecracker-microvm/firecracker/blob/main/docs/jailer.md) is another binary shipped with firecracker that creates a chroot + network namespace and runs the firecracker process inside. It's good for security, but also important for snapshotting + networking as it allows snapshots to have the same firecracker/vsock unix socket paths and IPs, and co-exist on the same host because they are in different chroot / namespaces.

Post this change, networking in containers works out of the box as soon as the default route is added (example:
ip route add default via 192.168.241.1). I'll send another CL after this one that makes that default route addition happen as part of init.

Want to run this CL locally?
 - sudo ./tools/enable_local_firecracker.sh on a linux machine (x86-64)
 - blaze build enterprise/server/remote_execution/containers/firecracker:firecracker_test
 - bazel-bin/enterprise/server/remote_execution/containers/firecracker/firecracker_test_/firecracker_test

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -36,7 +36,7 @@ import (
 
 const (
 	// How long to wait for the VMM to listen on the firecracker socket.
-	firecrackerSocketWaitTimeout = 30 * time.Second
+	firecrackerSocketWaitTimeout = 3 * time.Second
 
 	// How long to wait when dialing the vmexec server inside the VM.
 	vSocketDialTimeout = 3 * time.Second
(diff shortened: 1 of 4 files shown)


siggisim (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -562,6 +820,10 @@ func (c *firecrackerContainer) Unpause(ctx context.Context) error {
 	return nil
 }
 
+func (c *firecrackerContainer) Wait(ctx context.Context) error {
```
bduffany, line 823: nit: Document this since it's not part of the Container interface?

tylerwilliams (commented):
Done.

PR #738, opened 2021-07-19
Truncate invocation patterns based on char count

bduffany:
* Always show the first pattern
* Beyond that, truncate based on char count instead of target count

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/664


server/util/invocationutil/invocationutil.go
```diff
@@ -0,0 +1,34 @@
+package invocationutil
```
tylerwilliams, line 1: Awesome that you factored this out!

Could we call it something different than util/invocationutil tho? It stutters (util), and also i don't think it's broadly useful.

How about we put a package called "format" or "invocation_format" inside of server/build_event_protocol?

server/util/invocationutil/invocationutil.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	patternCharLimit = 50
+	listSeparator    = ", "
```
tylerwilliams, line 10: Please comment these so that someone else could figure out how to set them.

tylerwilliams (changes requested):

server/util/invocationutil/invocationutil.go
bduffany: sg, I wasn't happy with the name invocationutil either :P moved to `server/build_event_protocol/invocation_format`

server/util/invocationutil/invocationutil.go
bduffany: Done

tylerwilliams (approved):
LGTM

PR #739, opened 2021-07-19
More optimized redaction of build events

bduffany:
---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/731, https://github.com/buildbuddy-io/buildbuddy-internal/issues/661


enterprise/server/auth/auth.go
```diff
@@ -534,6 +534,10 @@ func (a *OpenIDAuthenticator) ParseAPIKeyFromString(input string) string {
 }
 
 func (a *OpenIDAuthenticator) AuthContextFromAPIKey(ctx context.Context, apiKey string) context.Context {
+	if _, ok := ctx.Value(APIKeyHeader).(string); ok {
+		log.Warningf("Overwriting existing value of %q in context.", APIKeyHeader)
+	}
+	ctx = context.WithValue(ctx, APIKeyHeader, apiKey)
```
tylerwilliams, line 540: Will this trigger in normal operation?

server/util/redact/redact.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/golang/protobuf/proto"
+
+	bespb "github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
+)
+
+const (
+	RedactionFlagAPIKey = 1 << iota
+)
+
+func RedactAPIKey(ctx context.Context, env environment.Env, event *bespb.BuildEvent) error {
```
tylerwilliams, line 18: I am more guilty than most of naming stuff like this -- but the "right" way to name this in go is probably to call this method APIKey() and then your code would read:

if err := redact.ApiKey(asdasd); err != nil {
return err
}

server/util/redact/redact.go
```diff
@@ -0,0 +1,77 @@
...
+
+	contextKey, ok := ctx.Value("x-buildbuddy-api-key").(string)
+	if ok {
+		txt = strings.ReplaceAll(txt, contextKey, "<REDACTED>")
+	}
+
+	return proto.UnmarshalText(txt, event)
+}
+
+func getConfiguredAPIKey(env environment.Env) string {
```
tylerwilliams, line 72: can you add a comment on this / possibly also name it differently so it's clear what this API key is? (it's not the one sent in the request)

tylerwilliams (changes requested):

enterprise/server/auth/auth.go
bduffany: Nope -- switched to `alert.UnexpectedEvent`

server/util/redact/redact.go
bduffany: sg, done. I think it looks kind of weird within the package itself, but it's probably better to make it more readable at the callsite.

server/util/redact/redact.go
bduffany: Done

server/util/redact/redact.go
```diff
@@ -0,0 +1,79 @@
...
+	"strings"
+
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/golang/protobuf/proto"
+
+	bespb "github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
+)
+
+const (
+	RedactionFlagAPIKey = 1 << iota
```
tylerwilliams, line 15: Since this is written into the database -- can you explicitly define it?

tylerwilliams (approved):
LGTM

server/util/redact/redact.go
bduffany: Done

PR #740, opened 2021-07-19
Fix panic in case BES stream contains no events

bduffany:
Ran into this while trying to replay an invocation and I had a typo in the invocation ID.

---

**Version bump**: TODO


tylerwilliams (approved):
LGTM

PR #741, opened 2021-07-19
Set Ulimits when creating docker container

siggisim:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -25,12 +25,14 @@ import (
 	dockertypes "github.com/docker/docker/api/types"
 	dockercontainer "github.com/docker/docker/api/types/container"
 	dockerclient "github.com/docker/docker/client"
+	units "github.com/docker/go-units"
 	gstatus "google.golang.org/grpc/status"
 )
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):
LGTM

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -25,12 +25,14 @@ import (
 	dockertypes "github.com/docker/docker/api/types"
 	dockercontainer "github.com/docker/docker/api/types/container"
 	dockerclient "github.com/docker/docker/client"
+	units "github.com/docker/go-units"
```
bduffany, line 28: nit: prefix import alias with `docker` (for consistency with above imports)

bduffany (approved):

enterprise/server/remote_execution/containers/docker/docker.go
siggisim: this library is called `go-units` it just happens to be in the docker organization: https://github.com/docker/go-units

bduffany: woops -- units sg

PR #742, opened 2021-07-20
Fix cache mode displayed in UI when remote_executor is set

bduffany:
`--remote_cache` defaults to the value of `--remote_executor` if it is not set. See [here](https://github.com/bazelbuild/bazel/blob/718279b6c157769f48fbaa13e0aacb94682916ef/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java#L256-L260)

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/666


vadimberezniker (approved):
Thanks!

PR #744, opened 2021-07-20
Add support for storing executor information in Redis.

vadimberezniker:
We create a hash for each executor pool. The key is the executor ID and
the value is a new RegisteredExecutionNode proto.

Additionally we create a per-group set that tracks all the available
pools.

---

**Version bump**: Minor


tylerwilliams (approved):
LGTM

PR #745, opened 2021-07-20
Add prettier to checkstyle workflow

bduffany:
Now that prettier is included in `buildfix.sh`, we can include it in checkstyle with less hassle when there are errors. Formats the following file types: js, jsx, ts, tsx, html, css, yaml, json, md

---

**Version bump**: TODO

### DIFF
--- a/.github/workflows/checkstyle.yaml
+++ b/.github/workflows/checkstyle.yaml
@@ -50,9 +50,8 @@ jobs:
         run: |
           git merge-base HEAD origin/master |
             xargs git diff --name-only --diff-filter=AMRCT |
-            grep -P '\.(js|jsx|ts|tsx|html|css|yaml|yml|json|md)$' |
-            ( xargs --no-run-if-empty --delimiter='\n' npx prettier@2.2.1 --check || true ) |
-            ( grep -P '^\[warn\]' || true ) > prettier-errors.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/checkstyle.yaml
```diff
@@ -44,6 +44,18 @@ jobs:
           echo "clang format errors:"
           cat clang-format-errors.txt
 
+      - name: prettier
+        # NOTE: Only run prettier on files that differ from master, since
+        # prettier can be slow.
+        run: |
+          git merge-base HEAD origin/master |
+            xargs git diff --name-only --diff-filter=AMRCT |
+            grep -P '\.(js|jsx|ts|tsx|html|css|yaml|json|md|xml)$' |
```
siggisim, line 53: I'm sure there are a bunch of yaml, css, html, and md files that don't conform to prettier's style standard.

Should we fix those / see what they look like before enabling?

bduffany: Sure -- https://github.com/buildbuddy-io/buildbuddy/pull/754

siggisim: Seems reasonable if @tylerwilliams  is okay with it.

tylerwilliams: 👍

tylerwilliams (approved):
LGTM

PR #747, opened 2021-07-20
Fix race condition with executor shutdown

bduffany:
* Make all workspace operations mutually exclusive. In particular, make sure `Remove` waits for uploads to complete.
* After `Remove` is called, return an error from all other workspace operations.
* Parallelize all runner removals in `runner.Pool.Shutdown`, since removal is now blocked on upload completion. Otherwise, if a single workspace takes a long time to upload its outputs (while the pool is shutting down), other runners would not be removed.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -724,8 +724,8 @@ func (p *Pool) Shutdown(ctx context.Context) error {
 			removeResults <- r.RemoveWithTimeout(ctx)
 		}()
 	}
-	errs := []error{}
-	for _, _ = range runners {
+	errs := make([]error, 0)
+	for range runners {
 		if err := <-removeResults; err != nil {
 			errs = append(errs, err)
 		}
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -33,6 +33,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/golang/protobuf/proto"
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
 
 	aclpb "github.com/buildbuddy-io/buildbuddy/proto/acl"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -713,9 +713,20 @@ func (p *Pool) Shutdown(ctx context.Context) error {
...
+		// Remove runners in parallel, since each deletion is blocked on uploads
+		// to finish (if applicable). A single runner that takes a long time to
+		// upload its outputs should not block other runners from working on
+		// workspace removal in the meantime.
+		r := r
+		go func() {
+			removeResults <- r.RemoveWithTimeout(ctx)
+		}()
+	}
+	errs := []error{}
```
tylerwilliams, line 727: errs := make([]error, 0)

tylerwilliams (changes requested):
LGTM but have @vadimberezniker give it a once over before submitting too please.

enterprise/server/remote_execution/workspace/workspace.go
```diff
@@ -22,6 +23,14 @@ import (
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 )
 
+var (
+	// WorkspaceMarkedForRemovalError is returned from workspace operations
+	// whenever Remove was previously called on the workspace.
+	WorkspaceMarkedForRemovalError = status.UnavailableError("workspace is marked for removal")
```
siggisim, line 29: We should make sure these are retried

vadimberezniker (approved):
LGTM

enterprise/server/remote_execution/runner/runner.go
bduffany: Done

enterprise/server/remote_execution/workspace/workspace.go
bduffany: Yep, unavailable errors are retried.

bduffany:
Tested this by running some local builds (~10 trials before and after) with a few artificial targets that generate a large amount of small output files, and having the executor send SIGTERM to itself right at the start of `workspace.UploadOutputs`.

With this test setup, before this change I was consistently seeing errors saying that certain outputs were expected, but not present. After the change, I was able to reliably build with no errors.

tylerwilliams (approved):
LGTM

PR #750, opened 2021-07-20
Setup default route with init script

tylerwilliams:
This sets the default route so that networking works out of the box on newly spawned VMs.

routes will look like this on a new VM with networking enabled:
~ # ip route
default via 192.168.241.1 dev eth0
192.168.241.0/24 dev eth0 scope link  src 192.168.241.2
~ #


enterprise/server/cmd/goinit/main.go
```diff
@@ -98,8 +122,13 @@ func main() {
 		fmt.Printf("Error configuring logging: %s", err)
 		os.Exit(1) // in case log.Fatalf does not work.
 	}
+	log.Printf("os.Args before: %q", os.Args)
```
vadimberezniker, line 125: did you mean to leave these log statements in?

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -414,15 +416,23 @@ func (c *firecrackerContainer) getJailerCommand(ctx context.Context) *exec.Cmd {
...
 		bootArgs = "--debug_mode " + bootArgs
-	} else {
-		bootArgs += " quiet"
 	}
 	if c.enableNetworking {
-		bootArgs += " " + machineIPBootArgs
+		bootArgs = "--set_default_route " + bootArgs
 	}
-
+	log.Errorf("boot args: %q", bootArgs)
```
vadimberezniker, line 435: remove?

vadimberezniker (approved):
LGTM

enterprise/server/cmd/goinit/main.go
tylerwilliams: Nope, done.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
tylerwilliams: Done.

PR #752, opened 2021-07-20
Add `remote-dev` .bazelrc config

siggisim:
---

**Version bump**: None


tylerwilliams (approved):
Heh, feels like I've been doing this locally for years now!? Love it.

PR #754, opened 2021-07-20
Format JS, TS, and markup files with prettier

bduffany:
Formats all files with these extensions: yaml, yml, json, md (as well as README without md extension), html, css, ts, tsx, js, jsx.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #756, opened 2021-07-21
Allow firecracker based isolation via platform prop + executor config flag

tylerwilliams:
RFR

Bare mac executors do not need to change. Executors that want to support multiple workload isolation options may now do so by setting the appropriate flags in their configs. Clients may select a preferred workload isolation option which will be used if enabled, an error is returned otherwise.

This will allow us to enable firecracker alongside docker and use both in parallel while we test.


bduffany (commented):
Looks good so far :+1:

bduffany (approved):

PR #759, opened 2021-07-21
Page artifacts by output instead of target

bduffany:
Some targets have a large number of outputs, which causes the UI spend a long time rendering. This change shaves about 5s off of the page load time for our infamous 13 second invocation, and also reduces the time it takes to switch between the "app" and "artifacts" tabs from O(seconds) to O(~100ms).

Since we now only partially load targets, this PR adds little indicators in the list to make it clear that we're not showing all artifacts from each target.

---

image
---

Also fixed artifact filtering, which was broken in a previous PR but was unnoticed.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/661

### DIFF
--- a/app/invocation/invocation_artifacts_card.tsx
+++ b/app/invocation/invocation_artifacts_card.tsx
@@ -91,7 +91,7 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
                   ))}
                   {target.hiddenOutputCount > 0 && (
                     <div className="artifact-hidden-count">
-                      + {target.hiddenOutputCount} more {target.hiddenOutputCount === 1 ? "artifact" : "artifacts"} for
+                      {target.hiddenOutputCount} more {target.hiddenOutputCount === 1 ? "artifact" : "artifacts"} for
                       this target
                     </div>
                   )}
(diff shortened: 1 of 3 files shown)


app/invocation/invocation_artifacts_card.tsx
```diff
@@ -35,50 +34,78 @@ export default class ArtifactsCardComponent extends React.Component {
...
+                  )}
                 </div>
-              ))}
-            {targets.flatMap((completed) => completed.completed.importantOutput).length == 0 && (
-              <span>{this.props.filter ? "No matching artifacts" : "No artifacts"}</span>
+              </div>
+            ))}
+            {hiddenTargetCount > 0 && (
+              <div className="artifact-hidden-count">
+                + {hiddenTargetCount} more {hiddenTargetCount === 1 ? "target" : "targets"} with artifacts
```
siggisim, line 103: Should clicking on this (or the one below) load more artifacts?

siggisim (approved):
Nice!

app/invocation/invocation_artifacts_card.tsx
bduffany: Since it's hard to say what the most intuitive thing to do is when clicking those, I just made them less "clickable" looking per our offline discussion (see the updated screenshot).

siggisim (approved):

PR #762, opened 2021-07-22
Add `Login with SSO` button

siggisim:
---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -413,7 +413,12 @@ func NewOpenIDAuthenticator(ctx context.Context, env environment.Env) (*OpenIDAu
 		return nil, status.FailedPreconditionErrorf("No auth providers specified in config!")
 	}
 
-	return newOpenIDAuthenticator(ctx, env, authConfigs)
+	a, err := newOpenIDAuthenticator(ctx, env, authConfigs)
+	if err != nil {
+		alert.UnexpectedEvent("authentication_configuration_failed", "Failed to configure authentication: %s", err)
+	}
+
+	return a, err
 }
 
 func sameHostname(urlStringA, urlStringB string) bool {
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -452,7 +452,7 @@ func (a *OpenIDAuthenticator) getAuthConfig(issuer string) authenticator {
 
 func (a *OpenIDAuthenticator) getAuthConfigForSlug(slug string) authenticator {
 	for _, a := range a.authenticators {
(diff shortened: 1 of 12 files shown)


enterprise/server/auth/auth.go
```diff
@@ -437,6 +445,15 @@ func (a *OpenIDAuthenticator) getAuthConfig(issuer string) authenticator {
 	return nil
 }
 
+func (a *OpenIDAuthenticator) getAuthConfigForSlug(slug string) authenticator {
+	for _, a := range a.authenticators {
+		if strings.ToLower(a.getSlug()) == strings.ToLower(slug) {
```
tylerwilliams, line 450: could you use strings.EqualFold here?

enterprise/server/cmd/server/main.go
```diff
@@ -104,7 +105,7 @@ func convertToProdOrDie(ctx context.Context, env *real_environment.RealEnv) {
 	if err == nil {
 		env.SetAuthenticator(authenticator)
 	} else {
-		log.Infof("No authentication will be configured: %s", err)
+		alert.UnexpectedEvent("authentication_configuration_failed", "Failed to configure authentication: %s", err)
```
tylerwilliams, line 108: we alert on this metric (it will send a page). Is this really an unexpected event?

server/static/static.go
```diff
@@ -83,8 +83,13 @@ func handleRootPaths(env environment.Env, rootPaths []string, template *template
...
 func serveIndexTemplate(env environment.Env, tpl *template.Template, version string, w http.ResponseWriter) {
 	issuers := make([]string, 0)
+	ssoEnabled := false
 	for _, provider := range env.GetConfigurator().GetAuthOauthProviders() {
-		issuers = append(issuers, provider.IssuerURL)
+		if provider.Slug == "" {
+			issuers = append(issuers, provider.IssuerURL)
+		} else {
+			ssoEnabled = true
+		}
```
tylerwilliams, line 92: nit: add a comment like:

// Assemble a slice of the supported issuers. Omit "private" issuers, which have a slug,
// and set ssoEnabled = true if any private issuers are present in the config.

tylerwilliams (approved):
LGTM (few nits)

enterprise/server/cmd/server/main.go
siggisim: Moved the alert to where we actually try to configure at least one issuer, but it fails.

server/static/static.go
siggisim: Done

enterprise/server/auth/auth.go
siggisim: I forgot about that nifty (and poorly named) function. Done!

PR #764, opened 2021-07-22
Record cache stats on disconnect (buildbuddy/buildbuddy-internal#742)

tempoz:
We had been only writing cache stats in FinalizeInvocation; this adds that behavior to
MarkInvocationDisconnected as well.

---

**Version bump**: Patch


tempoz:
Tested locally and got cache data on a disconnected invocation.

tylerwilliams (approved):
Woot, nice and easy. Thanks!

PR #765, opened 2021-07-22
Add Redactor skeleton for migrating redaction logic from event_parser

bduffany:
This PR is a NOP for now but lays the groundwork for us to be able to factor the redaction logic out of `StreamingEventParser.ParseEvent` (and into `StreamingRedactor.RedactMetadata`) without breaking anything.

---

**Version bump**: Patch

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -72,13 +72,12 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
 		chunkFileSizeBytes = defaultChunkFileSizeBytes
 	}
 	buildEventAccumulator := accumulator.NewBEValues(iid)
-	redactor := redact.NewStreamingRedactor(b.env)
 	return &EventChannel{
 		env:                     b.env,
 		ctx:                     ctx,
 		pw:                      protofile.NewBufferedProtoWriter(b.env.GetBlobstore(), iid, chunkFileSizeBytes),
 		beValues:                buildEventAccumulator,
-		redactor:                redactor,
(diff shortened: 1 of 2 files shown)


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -72,11 +72,13 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
 		chunkFileSizeBytes = defaultChunkFileSizeBytes
 	}
 	buildEventAccumulator := accumulator.NewBEValues(iid)
```
vadimberezniker, line 74: You can inline this.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -72,11 +72,13 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
...
+	redactor := redact.NewStreamingRedactor(b.env)
```
vadimberezniker, line 75: Will the redactor contain any state? If not, should we instantiate a single instance in NewBuildEventHandler?

bduffany: Yes, on each build event stream, the BuildStartedEvent contains a piece of BuildMetadata called ALLOW_ENV and we need to store that value in the redactor so it knows which pieces of the env need to be redacted (the env comes in a subsequent event)

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Done

vadimberezniker (approved):
LGTM

tylerwilliams (approved):
LGTM

PR #767, opened 2021-07-24
Add initial frontend for global invocation filter

bduffany:
Only has a DatePicker for now (will add role filters later). Does not actually do any filtering yet; this is just the frontend piece for updating the URLSearchParams.

* Date defaults to "last 7 days" -- this should probably depend on the page. For invocations we could default to "all time" and just rely on paging to limit the invocations displayed. For the test grid and trends we could use 7 days, which is our current default.
* Update URL search params when picking a new date (using `replaceState` to avoid polluting history)
* Changes the background when filters are applied, so users aren't confused as to why they aren't seeing their most recent invocations whenever they reload. (The only other visual indicators are the URL and the displayed date range, which are both very subtle).
* Provide a way to clear filters (X button).
* Some basic date range formatting -- it's not perfect. Will find a better way to do this later.
* Only added to history page; will add to other pages in subsequent PRs.

Hidden behind the server flag `app.enable_global_filter`.

https://user-images.githubusercontent.com/2414826/126851302-75fdc032-7438-43b3-b6f8-323a663dad99.mp4

---

**Version bump**: Patch


BUILD
```diff
@@ -62,6 +62,8 @@ exports_files([
     "package.json",
     "yarn.lock",
     "VERSION",
+    "node_modules/react-date-range/dist/styles.css",
```
siggisim, line 65: I think we should just "vendor" these so they're easier to edit anyway.

You can create separate css files for them in the filter directory.

enterprise/app/filter/filter.tsx
```diff
@@ -0,0 +1,91 @@
...
+            />
+          </Popup>
+        </div>
+      </div>
+    );
+  }
+}
+
+// Exclude "Days from today" label since it's not useful.
+const INPUT_RANGES = [defaultInputRanges[0]];
```
siggisim, line 91: Is it easy to do custom ones like: Last 7 days, Last 30 days, ... ?

enterprise/app/history/history.tsx
```diff
@@ -376,6 +378,7 @@ export default class HistoryComponent extends React.Component {
           )}
         </div>
         {this.props.hash === "#users" && <OrgJoinRequestsComponent user={this.props.user} />}
+        {capabilities.globalFilter && <FilterComponent search={this.props.search} />}
```
siggisim, line 381: What do you think about putting this here vs making it feel more "global" - i.e. in the top right where the AW date picker was? I can see pros and cons to each approach.

siggisim (changes requested):
Look good, just a couple of comments

enterprise/app/filter/filter.tsx
bduffany: Yeah, the API is fairly simple -- you specify them via a "staticRanges" prop, and each range defines a label and a func that returns `{ startDate, endDate }`. (These INPUT_RANGES here correspond to rows with input boxes where you can enter a custom range, like "___ days up to today." I'm planning on cleaning this up as well as setting up the custom staticRanges in a subsequent PR.)

siggisim (approved):
Before submitting, can you do a build with and without this change to see how much it changes the js bundle size?

bduffany:
> Before submitting, can you do a build with and without this change to see how much it changes the js bundle size?

JS size diff is 329KB uncompressed, 48KB compressed (increase of 3% uncompressed, 4% compressed)

PR #768, opened 2021-07-26
Sort skyframe threads numerically

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/454

### DIFF
--- a/app/flame_chart/profile_model.ts
+++ b/app/flame_chart/profile_model.ts
@@ -31,14 +31,9 @@ export type ThreadTimeline = {
   maxDepth: number;
 };
 
-type NumberedThreadName = {
-  prefix: string;
-  number: number;
-};
-
 // Matches strings like "skyframe evaluator 1", "grpc-command-0", etc., splitting the
 // non-numeric prefix and numeric suffix into separate match groups.
-const NUMBERED_THREAD_NAME_PATTERN = /^([^\d]+)(\d+)$/;
+const NUMBERED_THREAD_NAME_PATTERN = /^(?<prefix>[^\d]+)(?<number>\d+)$/;
 
 function eventComparator(a: TraceEvent, b: TraceEvent) {
   // Group by thread ID.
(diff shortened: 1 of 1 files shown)


app/flame_chart/profile_model.ts
```diff
@@ -27,10 +27,13 @@ export type ThreadEvent = TraceEvent & {
 export type ThreadTimeline = {
   tid: number;
   threadName: string;
+  skyframeNumber?: number;
   events: ThreadEvent[];
   maxDepth: number;
 };
 
+const SKYFRAME_EVALUATOR_PATTERN = /skyframe evaluator (\d+)/;
```
siggisim, line 35: this seems a bit brittle, i.e. in the case that bazel changes the thread names used in the profile

you could just simplify the comparator to something like

```parseInt(a.threadName.match( /\d+/g )) - parseInt(b.threadName.match( /\d+/g ))```

**Review by siggisim — APPROVED** [review]

File: app/flame_chart/profile_model.ts (hunk 1 above)
Comment (bduffany) [reply app/flame_chart/profile_model.ts]: There's also things like "grpc-command-0" so I think matching just the number is also brittle -- changed the impl to also keep track of the prefix that comes before the number, and for any threads matching the pattern we sort by the number (within the prefix group).

File: app/flame_chart/profile_model.ts [hunk 2]
```diff
@@ -31,6 +31,15 @@ export type ThreadTimeline = {
   maxDepth: number;
 };
 
+type NumberedThreadName = {
```
siggisim, line 34: you can probably avoid creating this little struct and the parseNumberedThreadName function by using named capture groups:
https://javascript.info/regexp-groups#named-groups

siggisim (approved):

app/flame_chart/profile_model.ts
bduffany: Ah nice, yeah that makes the change a bit smaller and also makes the regex less opaque :) done

app/flame_chart/profile_model.ts
```diff
@@ -47,8 +51,16 @@ function eventComparator(a: TraceEvent, b: TraceEvent) {
...
 }
 
-function threadNameComparator(a: ThreadTimeline, b: ThreadTimeline) {
-  // Sort by thread name.
+function timelineComparator(a: ThreadTimeline, b: ThreadTimeline) {
+  // Within numbered thread names (e.g. "skyframe evaluator 0", "grpc-command-0"), sort
+  // numerically.
+  const matchA = a.threadName.match(NUMBERED_THREAD_NAME_PATTERN)?.groups;
+  const matchB = b.threadName.match(NUMBERED_THREAD_NAME_PATTERN)?.groups;
+  if (matchA && matchB && matchA["prefix"] === matchB["prefix"]) {
```
siggisim, line 59: think it's a little more readable as 
```
if (matchA && matchB && matchA.prefix === matchB.prefix) {
  return Number(matchA.number) - Number(matchB.number);
}
```

bduffany: woops, missed this. Will clean up the next time I pass through here

PR #770, opened 2021-07-26
Display timeouts and other types of test failures with red backgrounds

siggisim:
---

**Version bump**: Minor

### DIFF
--- a/app/target/target.tsx
+++ b/app/target/target.tsx
@@ -84,6 +84,8 @@ export default class TargetComponent extends React.Component {
         return "test-failed";
       case build_event_stream.TestStatus.INCOMPLETE:
         return "test-error";
+      default:
+        return "test-failed";
     }
   }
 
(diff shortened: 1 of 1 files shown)


app/target/target.tsx
```diff
@@ -84,6 +84,8 @@ export default class TargetComponent extends React.Component {
         return "test-failed";
       case build_event_stream.TestStatus.INCOMPLETE:
         return "test-error";
+      default:
+        return "test-failed";
```
bduffany, line 88: Would "test-error" make more sense here? Seems like "TIMEOUT" is more like "INCOMPLETE" than "FAILED" -- "failed" seems like more of a "normal" status whereas INCOMPLETE, TIMEOUT, etc. are like "something weird happened and it neither passed nor failed"

bduffany (approved):

app/target/target.tsx
siggisim: Other places in the UI we lump Timeouts in with Failures. We can change that, but we should do it across the board.

I've updated the switch statement to be more explicit about the grouping and made the default case `test-error`.

PR #771, opened 2021-07-26
Introduce a build rule that calculates the SHA of it inputs

siggisim:
This is used when generating a static js bundle release for serving via CDN.

---

**Version bump**: None

### DIFF
--- a/rules/sha/index.bzl
+++ b/rules/sha/index.bzl
@@ -1,4 +1,4 @@
-# Calculates the sha1 of the contents of the files in `srcs` and outputs it to a file called `name`.txt.
+# Calculates the sha1 of the contents of the files in `srcs` and outputs it to a file called `name`.sum.
 def sha(name, srcs, **kwargs):
     native.genrule(
         name = name,
--- a/rules/sha/index.bzl
+++ b/rules/sha/index.bzl
@@ -1,10 +1,10 @@
-# Calculates a sha of the contents of the files in `srcs` and outputs it to a file called `name`.txt.
+# Calculates the sha1 of the contents of the files in `srcs` and outputs it to a file called `name`.txt.
 def sha(name, srcs, **kwargs):
     native.genrule(
         name = name,
         srcs = srcs,
-        outs = [name + ".txt"],
-        cmd = "echo \"$$(echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 }')\" > $@",
+        outs = [name + ".sum"],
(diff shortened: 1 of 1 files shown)


rules/sha/index.bzl
```diff
@@ -0,0 +1,10 @@
+# Calculates a sha of the contents of the files in `srcs` and outputs it to a file called `name`.txt.
+def sha(name, srcs, **kwargs):
+    native.genrule(
+        name = name,
+        srcs = srcs,
+        outs = [name + ".txt"],
+        cmd = "echo \"$$(echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 }')\" > $@",
```
bduffany, line 7: nit: I think you could remove the subshell and run

```
echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@
```

bduffany: Also, uber nit: run `xargs` with `--delimiter '\n'`

bduffany (approved):

rules/sha/index.bzl
```diff
@@ -0,0 +1,10 @@
+# Calculates a sha of the contents of the files in `srcs` and outputs it to a file called `name`.txt.
+def sha(name, srcs, **kwargs):
```
bduffany, line 2: nit: consider calling this sha1 (or add to the documentation above?)

rules/sha/index.bzl
```diff
@@ -0,0 +1,10 @@
...
+    native.genrule(
+        name = name,
+        srcs = srcs,
+        outs = [name + ".txt"],
```
bduffany, line 6: nit: consider making the file extension `.sum`

bduffany (commented):
two more nits (sorry)

rules/sha/index.bzl
siggisim: > nit: I think you could remove the subshell
Done


> Also, uber nit: `xargs --delimiter '\n' find` in case file names contain spaces
SRCS is space delimited. Looks like this is a more general bazel problem: https://github.com/bazelbuild/bazel/issues/4327

rules/sha/index.bzl
siggisim: Added to the documentation - would add an arg if folks wanted to change which sha algo to use.

rules/sha/index.bzl
siggisim: Done

PR #772, opened 2021-07-26
Use hash of js bundle in the `js_entry_point_path`

siggisim:
This replaces instance of `{APP_BUNDLE_HASH}` with that app bundle's hash in the `js_entry_point_path`.

We then place this as a query parameter when loading the default js bundle.

Like:
```
/app/app_bundle/app.js?hash=abc123
```

During local development, this means you should no longer have to force refresh when we make front-end changes and allows us to set aggressive cache headers.

In our production configs, this allows us to use the same variable to point do a CDN release like:
```
--js_entry_point_path=https://storage.googleapis.com/buildbuddy-static/app/{APP_BUNDLE_HASH}/app_bundle/app.js
```
or
```
--js_entry_point_path=https://static.buildbuddy.io/app/{APP_BUNDLE_HASH}/app_bundle/app.js
```

---

**Version bump**: Minor

### DIFF
--- a/enterprise/app/BUILD.bazel
+++ b/enterprise/app/BUILD.bazel
@@ -73,3 +73,10 @@ genrule(
         cat out > $@;
     """,
 )
+
+load("//rules/sha:index.bzl", "sha")
+
+sha(
+    name = "sha",
+    srcs = ["//enterprise/app:app_bundle"],
+)
--- a/rules/sha/index.bzl
+++ b/rules/sha/index.bzl
@@ -4,7 +4,7 @@ def sha(name, srcs, **kwargs):
         name = name,
         srcs = srcs,
         outs = [name + ".sum"],
-        cmd = "echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 } > $@",
+        cmd = "echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@",
         local = 1,
         **kwargs
     )
--- a/rules/sha/index.bzl
+++ b/rules/sha/index.bzl
@@ -1,5 +1,4 @@
-"""Calculates the sha1 of the contents of the files in `srcs` and outputs it to a file called `name`.sum"""
-
+# Calculates the sha1 of the contents of the files in `srcs` and outputs it to a file called `name`.sum.
 def sha(name, srcs, **kwargs):
     native.genrule(
         name = name,
--- a/rules/sha/index.bzl
+++ b/rules/sha/index.bzl
@@ -4,7 +4,7 @@ def sha(name, srcs, **kwargs):
         name = name,
         srcs = srcs,
         outs = [name + ".sum"],
-        cmd = "echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@",
+        cmd = "xargs find $(SRCS) -type f | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@",
         local = 1,
         **kwargs
     )
--- a/rules/sha/index.bzl
+++ b/rules/sha/index.bzl
@@ -4,7 +4,7 @@ def sha(name, srcs, **kwargs):
         name = name,
         srcs = srcs,
         outs = [name + ".sum"],
-        cmd = "xargs find $(SRCS) -type f | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@",
+        cmd = "find $(SRCS) -type f | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@",
         local = 1,
         **kwargs
     )
--- a/rules/sha/index.bzl
+++ b/rules/sha/index.bzl
@@ -1,0 +1,9 @@
+def sha(name, srcs, **kwargs):
+    native.genrule(
+        name = name,
+        srcs = srcs,
+        outs = [name + ".txt"],
+        cmd = "echo \"$$(echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 }')\" > $@",
+        local = 1,
+        **kwargs
+    )
--- a/server/static/static.go
+++ b/server/static/static.go
@@ -21,7 +21,7 @@ const (
 )
 
 var (
-	jsEntryPointPath = flag.String("js_entry_point_path", "/app/app_bundle/app.js", "Absolute URL path of the app JS entry point")
+	jsEntryPointPath = flag.String("js_entry_point_path", "/app/app_bundle/app.js?hash={APP_BUNDLE_HASH}", "Absolute URL path of the app JS entry point")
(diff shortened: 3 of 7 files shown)


server/static/static.go
```diff
@@ -120,10 +131,33 @@ func serveIndexTemplate(env environment.Env, tpl *template.Template, version str
...
+
+func AppBundleHash(bundleFS fs.FS) string {
+	var hashBytes []byte
+	if data, err := fs.ReadFile(bundleFS, "sha.sum"); err == nil {
+		hashBytes = data
+	}
+	if hashBytes != nil {
+		return strings.TrimSpace(string(hashBytes))
+	}
+	return ""
```
bduffany, line 150: nit: log the error if it is non-nil (and reading the data outside the `if` scope simplifies things a bit):

```go
data, err := fs.ReadFile(bundleFS, "sha.sum")
if err != nil {
  log.Errorf("Failed to read app checksum: %s", err)
  return ""
}
return strings.TrimSpace(string(hashBytes))
```

rules/sha/index.bzl
```diff
@@ -4,7 +4,7 @@ def sha(name, srcs, **kwargs):
         name = name,
         srcs = srcs,
         outs = [name + ".sum"],
-        cmd = "echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 } > $@",
+        cmd = "xargs find $(SRCS) -type f | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@",
```
bduffany, line 7: Instead of `xargs find` I think just `find` works

server/static/static.go
```diff
@@ -52,7 +54,16 @@ func NewStaticFileServer(env environment.Env, fs fs.FS, rootPaths []string) (*St
...
 		}
 
-		handler = handleRootPaths(env, rootPaths, template, version.AppVersion(), handler)
+		jsPath := *jsEntryPointPath
+		if appBundleHash != "" {
+			jsPath = strings.ReplaceAll(jsPath, "{APP_BUNDLE_HASH}", appBundleHash)
+		}
+
+		if strings.HasPrefix(jsPath, "http://") || strings.HasPrefix(jsPath, "https://") {
+			env.GetHealthChecker().AddHealthCheck("static_bundle", &healthChecker{jsPath: jsPath})
```
bduffany, line 63: uber nit: rename the health check to `"app_static_file_server"` (since the server handles stuff other than the static bundle)

enterprise/app/BUILD.bazel
```diff
@@ -73,3 +73,10 @@ genrule(
         cat out > $@;
     """,
 )
+
+load("//rules/sha:index.bzl", "sha")
+
+sha(
+    name = "sha",
```
bduffany, line 80: uber nit: rename to `app_bundle_sha` since we probably also want `style_sha` in here as well at some point

bduffany (approved):
Nice! I think we'll need to do this for `styles.css` as well before the need for force-refreshing goes away for good.

Ideally we'd also do this for all static assets (especially SVGs), because reloading those SVGs is the cause of some jankiness / reflow when loading the app and loading components for the first time (like the sidebar menu). We'll probably need something more general for that though (maybe some registry of path -> hash that gets generated as JS, and some `staticUrl` JS helper function that looks up the hash from the registry and automatically appends it).

server/static/static.go
siggisim: Made this a real error.

rules/sha/index.bzl
siggisim: Whoops, leftover from my pipe-ing days. Fixed.

server/static/static.go
siggisim: Done

enterprise/app/BUILD.bazel
siggisim: My plan is to ship all of the static files as a single bundle (when we add more than the js files).

rules/sha/index.bzl
```diff
@@ -4,7 +4,7 @@ def sha(name, srcs, **kwargs):
         name = name,
         srcs = srcs,
         outs = [name + ".sum"],
-        cmd = "echo \"$(SRCS)\" | xargs find | sort | xargs shasum | shasum | awk '{ print $$1 } > $@",
+        cmd = "find $(SRCS) -type f | sort | xargs shasum | shasum | awk '{ print $$1 }' > $@",
```
tylerwilliams, line 7: bruh i'm sure you can add at least another command in here :P

server/static/static.go
```diff
@@ -120,10 +131,30 @@ func serveIndexTemplate(env environment.Env, tpl *template.Template, version str
...
+		JsEntryPointPath: jsPath,
 		GaEnabled:        !*disableGA,
 	})
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
 }
+
+func AppBundleHash(bundleFS fs.FS) (string, error) {
+	hashBytes, err := fs.ReadFile(bundleFS, "sha.sum")
```
tylerwilliams, line 143: could we call this something app-bundle-specific so it's not too confusing if we add some other kinda hash into the bundle later?

bundle.sha256 or app.sha256 or something?

server/static/static.go
```diff
@@ -43,7 +45,7 @@ type StaticFileServer struct {
 
 // NewStaticFileServer returns a new static file server that will serve the
```
tylerwilliams, line 46: we should probably just rename this package from static -> app or pull the app stuff out into its own package

server/static/static.go
```diff
@@ -52,7 +54,16 @@ func NewStaticFileServer(env environment.Env, fs fs.FS, rootPaths []string) (*St
...
 
-		handler = handleRootPaths(env, rootPaths, template, version.AppVersion(), handler)
+		jsPath := *jsEntryPointPath
+		if appBundleHash != "" {
+			jsPath = strings.ReplaceAll(jsPath, "{APP_BUNDLE_HASH}", appBundleHash)
+		}
+
+		if strings.HasPrefix(jsPath, "http://") || strings.HasPrefix(jsPath, "https://") {
+			env.GetHealthChecker().AddHealthCheck("app_static_file_server", &healthChecker{jsPath: jsPath})
+		}
```
tylerwilliams, line 64: thanks for adding this

tylerwilliams (changes requested):
LGTM (few little things)

rules/sha/index.bzl
siggisim: lol i used to have a couple more

server/static/static.go
siggisim: i'd like this to eventually be the sha of the full static bundle that we upload to gcs on each release, not just the js app bundle

server/static/static.go
siggisim: i'd like to do this for all static resources (svgs, css files, images, etc), just starting with the app js bundle because it's the most pressing

server/static/static.go
siggisim: np

tylerwilliams (approved):

PR #774, opened 2021-07-27
Decode URL search params before filtering

siggisim:
Slashes don't currently filter correctly

---

**Version bump**: Minor

### DIFF
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -115,10 +115,6 @@ export default class InvocationComponent extends React.Component<Props, State> {
     }, 3000);
   }
 
-  decodedSearchParam(key: string) {
-    return this.props.search.get(key) ? decodeURIComponent(this.props.search.get(key)) : "";
-  }
-
   render() {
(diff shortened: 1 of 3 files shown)


app/invocation/invocation.tsx
```diff
@@ -115,6 +115,10 @@ export default class InvocationComponent extends React.Component<Props, State> {
     }, 3000);
   }
 
+  decodedSearchParam(key: string) {
```
bduffany, line 118: `urlSearchParams.get()` should already return the decoded value -- I think the problem is that we are double-encoding. In `InvocationFilterComponent.handleFilterChange` we're encoding `event.target.value` and then passing that to `router.updateParams`, which then calls `getModifiedUrl` which encodes the value again.

These days we should be able to avoid ever calling `encode`/`decodeURIComponent` and instead use `URLSearchParams` for everything (it automatically encodes/decodes for us).

siggisim: Good catch, fixed!

bduffany (approved):

PR #775, opened 2021-07-27
Add timestamp filtering fields to InvocationQuery

bduffany:
---

**Version bump**: Patch


proto/invocation.proto
```diff
@@ -167,6 +167,13 @@ message InvocationQuery {
 
   // The ROLE metadata set on the build.
   string role = 6;
+
+  // The timestamp on or after which the build must have been updated
+  // (inclusive).
+  google.protobuf.Timestamp updated_timestamp_range_start = 7;
+
+  // The timestamp up to which the build must have been updated (exclusive).
+  google.protobuf.Timestamp updated_timestamp_range_end = 8;
```
bduffany, line 176: These names might seem a bit long, but IMO are better than other alternatives I considered:

* `min_updated_timestamp` / `max_updated_timestamp`: problematic because `max` suggests inclusive, but we probably want the right limit to be exclusive because that's usually the convention and probably the least surprising
* `start_timestamp` / `end_timestamp`: slightly ambiguous because they could be referring to build start / build end -- documentation would clarify this, but it's better if the fields are self-documenting

proto/invocation.proto
```diff
@@ -170,6 +170,12 @@ message InvocationQuery {
 
   // The ROLE metadata set on the build.
   string role = 6;
+
+  // The timestamp on or after which the build was last updated (inclusive).
+  google.protobuf.Timestamp updated_timestamp_range_start = 7;
+
+  // The timestamp up to which the build was last updated (exclusive).
+  google.protobuf.Timestamp updated_timestamp_range_end = 8;
```
tylerwilliams, line 178: (idk what this looks like on the client side but another way to do this would be to have a timestamp + a duration)

What about calling these fields:
updated_after and updated_before?

tylerwilliams (changes requested):

proto/invocation.proto
bduffany: Those names SG (Done) -- it's more ergonomic for the client to use start/end dates so will keep as timestamps

tylerwilliams (approved):
LGTM

PR #776, opened 2021-07-27
Wire up datepicker to invocation history query

bduffany:
Datepicker is wired up to the `updated_at_usec` field (not the `created_at_usec` field) since we are using that field for the history query already (in the `ORDER BY`) and we don't have indexes on `created_at_usec`. The trends page uses that timestamp as well.

Tried running the resulting history query on groups with the largest number of invocations; query performance was the same (since both queries with/without the range filter, are ordering by `updated_at_usec` and hitting the row limit of 126)

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

tylerwilliams:
(timeutil comes through nicely)

siggisim (approved):

PR #777, opened 2021-07-27
Clean up deps.bzl

tempoz:
This PR does several things to clean up our external dependency specifications:

- Makes go.mod reference tyler's firecracker fork instead of only doing that in deps.bzl (which gopls can't read).

- Runs `go mod tidy` and `gazelle -- update-repos`.

- Fixes internal imports in `exhaustive` to prevent warnings being printed during build.

- Pins `exhaustive` version

- Disables generation of the test protos in `protoreflect`, as they aren't necessary and wwere creating build noise.

---

**Version bump**: Patch


deps.bzl
```diff
@@ -173,8 +173,8 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
     go_repository(
         name = "com_github_azure_go_autorest_autorest",
         importpath = "github.com/Azure/go-autorest/autorest",
-        sum = "h1:gI8ytXbxMfI+IVbI9mP2JGCTXIuhHLgRlvQ9X4PsnHE=",
-        version = "v0.11.12",
+        sum = "h1:gI8ytXbxMfI+IVbI9mP2JGCTXIuhHLgRlvQ9X4PsnHE=",  # keep
```
vadimberezniker, line 176: My suggestion is to not fight gazelle here and let it downgrade the dependencies. This will be easier to maintain going forward. This matches the behavior of plain Go. 

For transitive dependencies, Go picks the highest compatible version amongst all the modules in the graph that depend on it. You can see the dependency graph by running `go mod graph`.

vadimberezniker (commented):
Thanks for the cleanup!

deps.bzl
tempoz: Fixed!

deps.bzl
```diff
@@ -2356,8 +2375,8 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         importpath = "github.com/nishanths/exhaustive",
         patch_args = ["-p1"],
         patches = ["@%s//buildpatches:com_github_nishanths_exhaustive" % workspace_name],
-        sum = "h1:+ANTMqRNrqwInnP9aszg/0jDo+zbXa4x66U19Bx/oTk=",
-        version = "v0.2.3",
+        sum = "h1:eMV1t2NQRc3r1k3guWiv/zEeqZZP6kPvpUfy6byfL1g=",
+        version = "v0.0.0-20200811152831-6cf413ae40e0",
```
vadimberezniker, line 2379: This dep is a bit of a special case because it's not referenced by any Go file in our project (it's an analyzer plugin). 
Let's pin this one to the latest version.

vadimberezniker (approved):
Let's pin `exhaustive` to the latest version. Otherwise LGTM!

deps.bzl
tempoz: Done ^-^

PR #778, opened 2021-07-27
Add tool for sanity checking JS bundle size diff

bduffany:
Example output:

```
BEFORE
size (KB)       4597
gzipped (KB)    814

AFTER
size (KB)       4746
gzipped (KB)    848

DIFF
size (KB)       148
gzipped (KB)    33
```

---

**Version bump**: TODO

### DIFF
--- a/tools/js_size_diff.sh
+++ b/tools/js_size_diff.sh
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 get_bundle_info() {
-  bazel build //enterprise/app:app_bundle >&2
+  bazel build -c opt //enterprise/app:app_bundle >&2
   local bundle_path="bazel-bin/enterprise/app/app_bundle/app.js"
   local size_bytes
   local compressed_size_bytes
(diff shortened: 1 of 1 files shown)


tools/js_size_diff.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+set -euo pipefail
+
+get_bundle_info() {
+  bazel build //enterprise/app:app_bundle >&2
+  local bundle_path="bazel-bin/enterprise/app/app_bundle/app.js"
+  local size_bytes
+  local compressed_size_bytes
+  size_bytes=$(cat "$bundle_path" | wc -c)
+  compressed_size_bytes=$(cat "$bundle_path" | gzip | wc -c)
```
siggisim, line 10: The build we serve in production is built with `bazel build -c opt` which minifies the js bundle also

bduffany: Ah right -- fixed

siggisim (approved):

PR #780, opened 2021-07-27
Move most basic redaction logic to StreamingRedactor

bduffany:
This PR moves most "trivial" pieces of redaction logic from the event parser to the redactor. Since we redact before calling the event parser (as of #765), this is effectively a NOP.

Will do the build metadata and env var redaction in the next PR since it's not 100% straightforward (compared to the changes in this PR).

---

**Version bump**: Patch

### DIFF
--- a/server/util/redact/redact_test.go
+++ b/server/util/redact/redact_test.go
@@ -3,40 +3,54 @@ package redact_test
 import (
 	"testing"
 
-	"github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
 	"github.com/buildbuddy-io/buildbuddy/proto/command_line"
 	"github.com/buildbuddy-io/buildbuddy/server/testutil/testenv"
 	"github.com/buildbuddy-io/buildbuddy/server/util/redact"
 	"github.com/stretchr/testify/assert"
+
+	bespb "github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
 )
 
-func singleFile() *build_event_stream.File {
-	return &build_event_stream.File{
-		Name: "afile",
-		File: &build_event_stream.File_Uri{
-			Uri: "213wZJyTUyhXkj381312@uri",
-		},
+func fileWithURI(uri string) *bespb.File {
+	return &bespb.File{
+		Name: "foo.txt",
+		File: &bespb.File_Uri{Uri: uri},
 	}
 }
 
-func singleFiles() []*build_event_stream.File {
-	return []*build_event_stream.File{
-		singleFile(),
-	}
-}
-
-func TestRedact(t *testing.T) {
+func TestRedactMetadata_BuildStarted_StripsSecrets(t *testing.T) {
 	redactor := redact.NewStreamingRedactor(testenv.GetTestEnv(t))
-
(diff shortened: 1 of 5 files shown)


bduffany:
Woops, need to fix failing tests. Marking WIP for now

bduffany:
Fixed tests; should be good to go.

bduffany:
This diff might be useful as an extra sanity check -- it shows only what was excluded from the ParseEvent function body when copying over to RedactMetadata: https://gist.github.com/bduffany/a920cabe7527da7edf464d1514590f07

server/util/redact/redact_test.go
```diff
@@ -0,0 +1,165 @@
...
+func singleFile() *build_event_stream.File {
+	return &build_event_stream.File{
+		Name: "afile",
+		File: &build_event_stream.File_Uri{
+			Uri: "213wZJyTUyhXkj381312@uri",
+		},
+	}
+}
+
+func singleFiles() []*build_event_stream.File {
```
vadimberezniker, line 22: What do you think about calling this singleFileSlice? singleFiles is a bit awkward.

server/util/redact/redact_test.go
```diff
@@ -0,0 +1,165 @@
+package redact_test
```
vadimberezniker, line 1: Thanks for adding tests!

server/util/redact/redact_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	redactor.RedactMetadata(&build_event_stream.BuildEvent{
+		Payload: &build_event_stream.BuildEvent_OptionsParsed{OptionsParsed: optionsParsed},
+	})
+
+	assert.Equal(t, []string{"foo"}, optionsParsed.CmdLine)
+	assert.Equal(t, []string{"explicit"}, optionsParsed.ExplicitCmdLine)
+
+	actionExecuted := &build_event_stream.ActionExecuted{
+		Stdout:             singleFile(),
```
vadimberezniker, line 86: What do you think about passing the URI as an argument to singleFile() so that person reading the test can see both the input and the expected output in the test?

server/util/redact/redact_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+
+func singleFiles() []*build_event_stream.File {
+	return []*build_event_stream.File{
+		singleFile(),
+	}
+}
+
+func TestRedact(t *testing.T) {
```
vadimberezniker, line 28: nit: this test is on the long side, consider splitting it up into multiple test functions

vadimberezniker (approved):
LGTM, only some minor test comments

server/util/redact/redact_test.go
bduffany: SG, done (and renamed `singleFile` to `fileWithURI`)

server/util/redact/redact_test.go
bduffany: Done

server/util/redact/redact_test.go
bduffany: Removed and now constructing the slice directly

PR #781, opened 2021-07-27
Fix gazelle's proto packaging for `bojand/ghz` dep

tempoz:
Fixes the proto packaging in `bojand/ghz` dep so that gazelle stops complaining.

---

**Version bump**: Patch


vadimberezniker (approved):
LGTM

PR #783, opened 2021-07-27
Add bazel rule for uploading files to GCS

siggisim:
This will be used to release the static app bundle.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

tylerwilliams:
Would be nice if there were a pipe-able command that was commonly available that would let you check for confirmation (that you could use in the delete step)

PR #785, opened 2021-07-28
Move WorkspaceStatus and BuildMetadata redaction to redactor

bduffany:
* Strip repo URL credentials from WorkspaceStatus and BuildMetadata in redactor instead of event parser
* Strip repo URL credentials more uniformly
* Beef up `build_event_handler_test` to ensure the redactor is wired up to the build event handler properly whether or not `--enable_optimized_redaction` is set. Includes tests for StructuredCommandLine redaction, which will be useful for the next PR (which moves env var redaction to the redactor as well).

---

**Version bump**: Patch


vadimberezniker (approved):
LGTM

PR #786, opened 2021-07-28
Remove globs from BUILD files for gazelle

tempoz:
Gazelle can't understand globs, so every time we run `fix` it produces
noise about these BUILD targets. Their input files are neither overly
numerous nor frequently changing, so it makes sense to just list them.

---

**Version bump**: Patch


bduffany (approved):
(Disabled auto-merge since Tyler is doing a release)

PR #788, opened 2021-07-28
Turn on chunked event logs in dev and local

tempoz:
This enables chunked event logs in local and dev configs.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

(please keep an eye on this in dev after you submit just to make sure things are looking good!)

PR #792, opened 2021-07-29
Add a release bazel rule

siggisim:
This will enable us to keep the master `dev.apply` rule that will deploy both the GCS bucket and then apply our k8s configs.

I'm working on adding a `.diff` action to the gcs deploy (in a separate PR) so we can keep the diff functionality.

There is prior art: https://github.com/atlassian/bazel-tools/tree/master/multirun

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #680, opened 2021-07-07
Temporarily disable tests that invoke bazel.

bduffany:
They are flaky because `bazel shutdown` sometimes takes forever.

---

**Version bump**: Patch

PR #793, opened 2021-07-29
Re-enable tests that invoke bazel

bduffany:
Reverts https://github.com/buildbuddy-io/buildbuddy/pull/680

---

**Version bump**: Patch


siggisim (approved):

bduffany:
@siggisim I realized after your approval that tests were failing, since the app binary spawned by the tests couldn't find the new sha.sum, and that's because we don't build the app into `//:bundle` when running tests under fastbuild mode.

Added a flag to `libmain.go` that allows disabling the static file server just for tests. LMK if you think that's a reasonable solution. A simpler solution is to run tests with `-c opt` but I think that means the app would then be a dependency of our unit tests which is probably not what we want.

siggisim:
> @siggisim I realized after your approval that these tests were failing, since the app binary spawned by the tests couldn't find the new `sha.sum`, and that's because we don't build the app into `//:bundle` when running tests under fastbuild mode.
> 
> Added a flag to `libmain.go` that allows disabling the static file server just for tests. LMK if you think that's a reasonable solution. A simpler solution is to run tests with `-c opt` but I think that means the app would then be a dependency of our unit tests which is probably not what we want.

Hm, I don't love this solution. Let me see why the `sha.sum` file is not being found.

These aren't really unit tests if they're spinning up Bazel right?

siggisim:
Looks like we just need to give the server a path to the static and app directories - then we won't use the app bundle. This is what we do both during local development and during production.

You can add these in server/testutil/app/app.go:
```
"--static_directory=static",
"--app_directory=/enterprise/app",
```

Though we'll probably need a way to pass those paths for enterprise tests and these for non-enterprise:
```
"--static_directory=static",
"--app_directory=/app",
```

Maybe we can pass the app directory into `app.Run()`.

bduffany:
Done -- I thought about that approach too, but somehow convinced myself that it wouldn't work -- guess I should have just tried it lol :shrug:

siggisim (approved):

PR #797, opened 2021-07-29
Fix time range query for invocation history

bduffany:
The code was assuming that `time.Unix(0, 0).IsZero()` returns true, but it actually returns false. As a result, the history query would always filter to the time range `epoch_local_time <= t < epoch_local_time`, returning 0 results.

---

**Version bump**: Patch


vadimberezniker (approved):
LGTM

PR #799, opened 2021-07-29
Lower read buffer size to avoid bazel error

siggisim:
This should fix Bazel errors that look like:
```
(Exit 34): io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: gRPC message exceeds maximum size 4194304: 4194305
```
---

**Version bump**: Minor


tylerwilliams (approved):
LGTM

PR #800, opened 2021-07-30
Add cache-snapshot support

tylerwilliams:
Snapshots are read and written from the cache as directories containing all of the necessary files (kernel, initrd, containerfs, workspacefs, etc). This does not yet attach a different workspace to a snapshot-resumed VM.


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -83,7 +88,7 @@ const (
 
 	// Workspace slack space is how much extra space will be allocated in the
 	// workspace disk image beyond the size of the existing files.
-	workspaceSlackBytes = 1e9
+	workspaceSlackBytes = 100 * 1e6 // 100MB
```
siggisim, line 91: is this effectively a limit on the size of the outputs or am i misunderstanding what this is?

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -341,36 +283,168 @@ func NewContainer(ctx context.Context, opts ContainerOpts) (*firecrackerContaine
...
+		c.vmIdx = opts.ForceVMIdx
+	}
 	return c, nil
 }
 
+func (c *FirecrackerContainer) SaveSnapshot(ctx context.Context, instanceName string) (string, error) {
+	if err := c.machine.PauseVM(ctx); err != nil {
+		return "", err
+	}
+	memSnapshotPath := filepath.Join(c.getChroot(), fullDiskSnapshotName)
```
siggisim, line 322: i think mem and disk naming is swapped here

siggisim (approved):
looks good, just a couple of ubernits

enterprise/server/remote_execution/containers/firecracker/firecracker.go
tylerwilliams: Done.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
tylerwilliams: Until fuse lands, yup :(

PR #801, opened 2021-07-30
Dedupe writes differently depending on isolation type.

tylerwilliams:
AC items may be re-written and we want to allow this -- so check that the isolation type is == CAS before deduping cache writes.

Also, clean up the tests slightly.


siggisim (approved):

vadimberezniker (approved):
LGTM

PR #802, opened 2021-07-30
Add config option for specifying location of executor data on host.

vadimberezniker:
Local SSDs will be mounted using a `hostPath` volume which will appear
in a different location.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #804, opened 2021-07-30
Fix redis connection refused errors in tests

bduffany:
* Add a startup healthcheck before allowing the test to use the server instance, to ensure that the server is ready to accept connections.
* Use a Unix socket instead of a random free TCP port to avoid race conditions where the port gets taken after the port is determined but before the server actually listens on it.

Tested locally with `--runs_per_test=2000` (12K total test funcs)

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/767

### DIFF
--- a/enterprise/server/testutil/testredis/testredis.go
+++ b/enterprise/server/testutil/testredis/testredis.go
@@ -62,8 +62,8 @@ func Start(t testing.TB) string {
 	args = append(args, "--maxmemory-policy", "noeviction")
 	cmd := exec.CommandContext(ctx, redisBinPath, args...)
 	log.Printf("Starting redis server: %s", cmd)
-	cmd.Stdout = &logWriter{t}
-	cmd.Stderr = &logWriter{t}
+	cmd.Stdout = &logWriter{}
+	cmd.Stderr = &logWriter{}
 	err = cmd.Start()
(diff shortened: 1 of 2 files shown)


enterprise/server/testutil/testredis/testredis.go
```diff
@@ -68,12 +74,37 @@ func Start(t *testing.T) string {
...
+	return target
+}
+
+func waitUntilHealthy(t testing.TB, target string) {
+	start := time.Now()
+	ctx := context.Background()
+	r := redis.NewClient(redisutil.TargetToOptions(target))
+	hc := redisutil.HealthChecker{Rdb: r}
+	for {
+		err := hc.Check(ctx)
```
vadimberezniker, line 88: nit: i think it's simpler to call `r.Ping(ctx).Err` without needing to instantiate a healthchecker.

vadimberezniker (approved):
LGTM thanks!

enterprise/server/testutil/testredis/testredis.go
bduffany: Done

vadimberezniker (approved):

PR #805, opened 2021-07-30
Show Input File Tree

pariparajuli:
Expandable & scrollable input file tree in the actions page with downloadable files


pariparajuli:
Screen Shot 2021-07-30 at 12 28 11 PM

pariparajuli:
Screen Shot 2021-07-30 at 12 38 02 PM

pariparajuli:
Screen Shot 2021-07-30 at 1 03 01 PM

app/invocation/invocation_action_card.tsx
```diff
@@ -196,6 +225,47 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+            error: "Error loading action profile. Make sure your cache is correctly configured.",
+          });
+        });
+      return;
+    } else {
+      try {
+        rpcService.downloadBytestreamFile(node.name, dirFile, this.props.model.getId());
+      } catch {
+        console.error("Error downloading bytestream timing profile");
+      }
```
bduffany, line 265: I think the `catch` isn't needed here -- if there's an error then the user will see it in the download status, or the `/file/download` will show a 404 if it's not found or something

app/invocation/invocation_action_card.tsx
```diff
@@ -196,6 +225,47 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+        })
+        .catch(() => {
+          console.error("Error loading bytestream file profile!");
+          this.setState({
+            ...this.state,
+            error: "Error loading action profile. Make sure your cache is correctly configured.",
+          });
+        });
+      return;
+    } else {
```
bduffany, line 260: nit: the "else" block can be removed since the above block ends with `return` -- makes the code a little bit less indented and easier to read

app/invocation/invocation_action_card.tsx
```diff
@@ -236,6 +306,24 @@ export default class InvocationActionCardComponent extends React.Component<Props
                       <div>Default</div>
                     )}
                   </div>
+                  <div className="action-section">
+                    <div
+                      title="List of required supported NodeProperty [build.bazel.remote.execution.v2.NodeProperty] keys."
+                      className="action-property-title">
+                      Input File Tree
```
bduffany, line 313: nit: "Input files" or "Input tree" (sentence case) might be a bit more concise

app/invocation/invocation_input_file_node.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
```
bduffany, line 1: nit: consider renaming this file `invocation_action_input_node.tsx` (remove "file" since it can also represent a dir, and add "action" to be consistent with `invocation_action_card`)

bduffany (changes requested):

app/invocation/invocation_action_card.tsx
```diff
@@ -46,6 +53,27 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+      .fetchBytestreamFile(inputRootFile, this.props.model.getId(), "arraybuffer")
+      .then((root_buff: any) => {
+        let tempRoot = build.bazel.remote.execution.v2.Directory.decode(new Uint8Array(root_buff));
+        this.setState({
+          ...this.state,
+          inputRoot: tempRoot,
+        });
+      })
+      .catch(() => {
+        console.error("Error loading bytestream inputRoot profile!");
```
siggisim, line 69: nit: remove " profile"

app/invocation/invocation_action_card.tsx
```diff
@@ -196,6 +225,47 @@ export default class InvocationActionCardComponent extends React.Component<Props
     );
   }
 
+  handleFileClicked(node: any) {
+    let digest_string = node.digest.hash + "/" + node.digest.sizeBytes;
+    let dirFile = "bytestream://" + this.getCacheAddress() + "/blobs/" + digest_string;
+    if (!node.hasOwnProperty("isExecutable")) {
```
siggisim, line 231: is there a cleaner way we can tell if this node is not a directory?

app/invocation/invocation_action_card.tsx
siggisim, line 260: this would probably be less complex if you did an early exit instead of an if/else 

```
if (node.hasOwnProperty("isExecutable")) {
      try {
        rpcService.downloadBytestreamFile(node.name, dirFile, this.props.model.getId());
      } catch {
        console.error("Error downloading bytestream timing profile");
      }
}

The rest of the code above ...
```

app/invocation/invocation_action_card.tsx
```diff
@@ -236,6 +306,24 @@ export default class InvocationActionCardComponent extends React.Component<Props
                       <div>Default</div>
                     )}
                   </div>
+                  <div className="action-section">
+                    <div
+                      title="List of required supported NodeProperty [build.bazel.remote.execution.v2.NodeProperty] keys."
```
siggisim, line 311: can we just remove this title

siggisim (commented):
wow this is really cool!

pariparajuli:
Screen Shot 2021-07-30 at 5 08 52 PM

siggisim:
I'd say either a lighter gray or just a simple `border: 1px solid #eee;`

app/invocation/invocation_action_card.tsx
siggisim: you could keep them in separate lists/maps instead of shoving them all in a "children" list. that might make it difficult to say sort them in alphabetical order later

another option would be to wrap them in another object that keeps track of their type when you fetch them

pariparajuli:
Screen Shot 2021-08-02 at 11 21 39 AM

app/invocation/invocation_action_card.tsx
```diff
@@ -77,20 +90,15 @@ export default class InvocationActionCardComponent extends React.Component<Props
       action.commandDigest.sizeBytes;
     rpcService
       .fetchBytestreamFile(commandFile, this.props.model.getId(), "arraybuffer")
-      .then((action_buff: any) => {
-        let temp_array = new Uint8Array(action_buff);
+      .then((actionBuff: any) => {
+        let tempArray = new Uint8Array(actionBuff);
         this.setState({
           ...this.state,
```
bduffany, line 96: nit: no need to spread `...state`, and also would be cleaner without the `tempArray` variable

app/invocation/invocation_action_card.tsx
```diff
@@ -77,20 +90,15 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
-          command: build.bazel.remote.execution.v2.Command.decode(temp_array),
+          command: build.bazel.remote.execution.v2.Command.decode(tempArray),
         });
+        this.fetchInputRoot(action.inputRootDigest);
```
bduffany, line 99: Since this only depends on the action and not the command in the response, can this be moved to a separate `fetchInputRoot` function and done in parallel with `fetchCommand`?

app/invocation/invocation_action_card.tsx
```diff
@@ -196,6 +204,36 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+      rpcService.downloadBytestreamFile(node.name, dirUrl, this.props.model.getId());
+      return;
+    }
+    rpcService
+      .fetchBytestreamFile(dirUrl, this.props.model.getId(), "arraybuffer")
+      .then((dirBuff: any) => {
+        let tempDir = build.bazel.remote.execution.v2.Directory.decode(new Uint8Array(dirBuff));
+        this.state.treeShaToExpanded.set(digestString, true);
+        let contents: Node[] = [];
+        this.state.treeShaToChildrenMap.set(digestString, contents.concat(tempDir.directories).concat(tempDir.files));
```
bduffany, line 230: nit: this can be condensed to

```tsx
this.state.treeShaToChildrenMap.set(
    digestString,
    [...dir.directories, ...dir.files]
);
```

app/invocation/invocation_action_card.tsx
```diff
@@ -196,6 +204,36 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+      return;
+    }
+    if (this.isFileNode(node)) {
+      rpcService.downloadBytestreamFile(node.name, dirUrl, this.props.model.getId());
+      return;
+    }
+    rpcService
+      .fetchBytestreamFile(dirUrl, this.props.model.getId(), "arraybuffer")
+      .then((dirBuff: any) => {
+        let tempDir = build.bazel.remote.execution.v2.Directory.decode(new Uint8Array(dirBuff));
```
bduffany, line 227: nit: avoid using "temp" in variable names -- local variables are always temporary so the "temp" prefix isn't adding much value

app/invocation/invocation_action_input_node.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+
+interface Props {
+  node: any;
```
bduffany, line 4: nit: import your `Node` type from `invocation_action_card` and use it here -- or better yet, define the `Node` type here and have `invocation_action_card` import it from here instead (it probably makes more sense to define `Node` in this file because this component is the thing that renders the `Node`)

bduffany (changes requested):

app/invocation/invocation_action_card.tsx
pariparajuli: i ended up wrapping them in another object

app/invocation/invocation_action_card.tsx
```diff
@@ -28,43 +38,46 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+      .fetchBytestreamFile(inputRootFile, this.props.model.getId(), "arraybuffer")
+      .then((rootBuff: any) => {
+        let tempRoot = build.bazel.remote.execution.v2.Directory.decode(new Uint8Array(rootBuff));
         this.setState({
           ...this.state,
-          error: "Error loading action profile. Make sure your cache is correctly configured.",
+          inputRoot: tempRoot,
         });
-      });
+      })
```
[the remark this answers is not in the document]
pariparajuli: I ended up having to add inputDirs when i added the wrapper object because of some type mismatching stuff

app/invocation/invocation_action_card.tsx
pariparajuli: can i do it after i fetch the command in the fetchAction function?

bduffany:
> > Since this only depends on the action and not the command in the response, can this be moved to a separate fetchInputRoot function and done in parallel with fetchCommand?
>
> can i do it after i fetch the command in the fetchAction function?

Yep that should work

app/invocation/invocation_action_card.tsx
```diff
@@ -28,43 +37,48 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
-        let tempAction = build.bazel.remote.execution.v2.Action.decode(new Uint8Array(action_buff));
+      .then((buffer: any) => {
         this.setState({
-          ...this.state,
-          contents: action_buff,
-          action: tempAction,
+          action: build.bazel.remote.execution.v2.Action.decode(new Uint8Array(buffer)),
         });
-        this.fetchCommand(tempAction);
+        this.fetchCommand(this.state.action);
```
bduffany, line 44: React doesn't always guarantee that `setState` will immediately update the state -- https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

> Think of setState() as a request rather than an immediate command to update the component. For better perceived performance, React may delay it, and then update several components in a single pass. React does not guarantee that the state changes are applied immediately.

So you could put `action` in a local variable then use that local instead of `this.state.action`

app/invocation/invocation_action_card.tsx
bduffany: (unresolving)

bduffany (changes requested):

app/invocation/invocation_action_card.tsx
```diff
@@ -196,6 +203,46 @@ export default class InvocationActionCardComponent extends React.Component<Props
     );
   }
 
+  handleFileClicked(node: InputNode) {
+    console.log("reached");
```
bduffany, line 207: remove `console.log`

bduffany: (`console.log` is still there -- forgot to `git push`?)

bduffany (approved):

PR #806, opened 2021-07-30
Timelines for action page

pariparajuli:
Separated timelines for execution metadata

### DIFF
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -14,7 +14,10 @@
 
 .action-timeline .timeline-event-label {
   width: 100%;
-  padding: 2px 4px;
+  padding: 2px 6px;
+  text-align: right;
+  font-size: 12px;
+  color: rgb(38, 50, 56, 0.7);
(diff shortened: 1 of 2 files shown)


pariparajuli:
Screen Shot 2021-07-30 at 3 21 18 PM

bduffany:
Much improved!

Mind double-checking that the timing numbers on the right are readable (high enough contrast) if they are displayed on top of the colors? You can easily check by setting the flex weight of the gray bar to 0.

pariparajuli:
the way i have the divs set up, the text can't sit on top of colors, if there is a bar that's too long, the text ends up overflowing to the side, i was thinking of putting the text within the bars on the right side above the bar itself

bduffany:
> i was thinking of putting the text within the bars on the right side above the bar itself

That sounds like a good option

pariparajuli:
Screen Shot 2021-08-02 at 2 53 17 PM

bduffany (approved):
So much nicer!

PR #807, opened 2021-07-30
Add branch info to invocations and propagate to client (buildbuddy-io/buildbuddy-internal#616)

tempoz:
Adds the branch information to the invocation and propagates it into the database and through to the client. Also adds branches to the sidebar, in the same style as the commits and repos tabs.

---

**Version bump**: Minor


siggisim (commented):
Wow, this looks really great - before submitting we should chat with @tylerwilliams about how we should go about adding this new column & index.

siggisim (approved):
lgtm-ing, but let's chat with tyler about adding the column before submitting

server/build_event_protocol/accumulator/accumulator.go
```diff
@@ -151,6 +157,12 @@ func (v *BEValues) populateWorkspaceInfoFromStructuredCommandLine(commandLine *c
 			switch environmentVariable {
 			case "CIRCLE_REPOSITORY_URL", "GITHUB_REPOSITORY", "BUILDKITE_REPO", "TRAVIS_REPO_SLUG", "GIT_URL", "CI_REPOSITORY_URL", "REPO_URL":
 				v.setStringValue(repoURLFieldName, value)
+			case "CIRCLE_BRANCH", "GITHUB_HEAD_REF", "BUILDKITE_BRANCH", "TRAVIS_BRANCH", "GIT_BRANCH", "CI_COMMIT_BRANCH":
+				v.setStringValue(branchNameFieldName, value)
+			case "GITHUB_REF":
+				if strings.HasPrefix(value, "refs/heads/") {
+					v.setStringValue(branchNameFieldName, strings.TrimPrefix(value, "refs/heads/"))
+				}
```
tylerwilliams, line 165: "TrimPrefix returns s without the provided leading prefix string. If s doesn't start with prefix, s is returned unchanged."

so we could just do v.setStringValue(branchNameFieldName, strings.TrimPrefix(value, "refs/heads/"))

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -523,6 +523,7 @@ func tableInvocationFromProto(p *inpb.Invocation, blobID string) *tables.Invocat
 	if *enableOptimizedRedaction {
 		i.RedactionFlags = redact.RedactionFlagStandardRedactions
 	}
+	i.BranchName = p.BranchName
```
tylerwilliams, line 526: nit: for clarity i would move this up next to RepoUrl and CommitSHA

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -570,5 +571,6 @@ func TableInvocationToProto(i *tables.Invocation) *inpb.Invocation {
 	if i.LastChunkId != "" {
 		out.HasChunkedEventLogs = true
 	}
+	out.BranchName = i.BranchName
```
tylerwilliams, line 574: (same as above)

server/tables/tables.go
```diff
@@ -97,6 +97,7 @@ type Invocation struct {
 	RepoURL      string `gorm:"index:repo_url_index"`
 	CommitSHA    string `gorm:"index:commit_sha_index"`
 	LastChunkId  string
+	BranchName   string `gorm:"index:branch_name_index"`
```
tylerwilliams, line 100: I'm not sure we need this -- what query will hit this index?

server/tables/tables.go
```diff
@@ -545,6 +546,7 @@ func PostAutoMigrate(db *gorm.DB) error {
 		"invocations_stats_user_index":        "(`group_id`, `user`, `action_count`, `duration_usec`, `updated_at_usec`, `success`, `invocation_status`)",
 		"invocations_stats_host_index":        "(`group_id`, `host`, `action_count`, `duration_usec`, `updated_at_usec`, `success`, `invocation_status`)",
 		"invocations_stats_repo_index":        "(`group_id`, `repo_url`, `action_count`, `duration_usec`, `updated_at_usec`, `success`, `invocation_status`)",
+		"invocations_stats_branch_index":      "(`group_id`, `branch_name`, `action_count`, `duration_usec`, `updated_at_usec`, `success`, `invocation_status`)",
```
tylerwilliams, line 549: This looks good but please test and make sure all the stats and trends pages are still hitting these indexes

tylerwilliams (changes requested):

proto/invocation.proto
```diff
@@ -176,6 +179,8 @@ message InvocationQuery {
 
   // The timestamp up to which the build was last updated (exclusive).
   google.protobuf.Timestamp updated_before = 8;
+  // The git branch used for the build.
```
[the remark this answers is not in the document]
tylerwilliams: (Please fix these both)

tylerwilliams:
This looks really good overall :) Excited to finally have branch support!

proto/target.proto
```diff
@@ -57,6 +57,9 @@ message TargetHistory {
   // If multiple targets were run at the same commit, the latest run will be
   // returned.
   repeated TargetStatus target_status = 3;
+
+  // The git branch the build was for.
+  string branch_name = 4;
```
siggisim, line 62: i think we can remove this, are we even setting it?

server/tables/tables.go
tempoz: Tested it; trends certainly is. What stats pages are you talking about?

server/tables/tables.go
tylerwilliams: Those links just take me back to this CL -- but a good way to check would be to run "explain ...." on those queries in mysql and see if these are being used?

server/tables/tables.go
tylerwilliams: I meant like when you hit https://app.buildbuddy.io/#repos or https://app.buildbuddy.io/#hosts or whatever -- are those still hitting an index.

tylerwilliams (approved):
LGTM (please validate the index stuff just to make sure we're hitting them & not spending ram on something unused)

tylerwilliams:
@vadimberezniker FYI for next release (index changes)

server/tables/tables.go
tempoz: yup, they get used when you visit the page for an individual repo, commit, or branch.

server/tables/tables.go
tempoz: Checked #repos; looks like it is.

PR #808, opened 2021-07-30
Improve loading sequence for chunked event logs

bduffany:
* Fetch the first chunk in parallel with the invocation, instead of making 2 requests in serial (GetInvocationRequest followed by GetEventLogChunkRequest)
* Update server so that if chunking is not enabled for an invocation, we return EOF instead of an error. This allows us to make the requests in parallel, because we can now always fetch event log chunks even if we don't know whether chunking is enabled (the enabled bit is present in the GetInvocationResponse).
* Show a loading state while the chunk is loading, if it has not loaded by the time the GetInvocationResponse comes back. For non-chunked invocations, we never show this loading state, since we have a separate full-page loading state until the InvocationModel is created and the InvocationModel contains the logs. For chunked invocations, a loading spinner is displayed in the logs card until we see any logs in the GetEventLogChunkResponse:

image

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/765

### DIFF
--- a/app/invocation/invocation_build_logs_card.tsx
+++ b/app/invocation/invocation_build_logs_card.tsx
@@ -11,7 +11,10 @@ interface Props {
 export default class BuildLogsCardComponent extends React.Component<Props> {
   render() {
     return (
-      <div className={`card ${this.props.dark ? "dark" : "light-terminal"} ${this.props.expanded ? "expanded" : ""}`}>
+      <div
+        className={`card build-logs-card ${this.props.dark ? "dark" : "light-terminal"} ${
+          this.props.expanded ? "expanded" : ""
+        }`}>
         <img className="icon" src={this.props.dark ? "/image/log-circle-light.svg" : "/image/log-circle.svg"} />
         <div className="content">
           <div className="title">Build logs </div>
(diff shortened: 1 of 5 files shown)


app/invocation/invocation_build_logs_card.tsx
```diff
@@ -1,89 +1,30 @@
...
         <img className="icon" src={this.props.dark ? "/image/log-circle-light.svg" : "/image/log-circle.svg"} />
         <div className="content">
           <div className="title">Build logs </div>
           <div className="details">
-            <TerminalComponent value={this.getConsoleBuffer()} lightTheme={!this.props.dark} />
+            {this.props.loading ? (
+              // "terminal" class makes sure the loading container's size
+              // is the same as the terminal that will take its place.
+              <div className="terminal">
+                <div className="loading" />
```
siggisim, line 23: does this handle dark and light terminal well?

siggisim (approved):

app/invocation/invocation_build_logs_card.tsx
bduffany: good catch - fixed

image

siggisim (approved):

tempoz (approved):
LGTM

PR #811, opened 2021-07-30
Rule tweaks

siggisim:
---

**Version bump**: None

### DIFF
--- a/rules/gcs/index.bzl
+++ b/rules/gcs/index.bzl
@@ -43,6 +43,17 @@ def gcs(name, srcs, bucket, gsutil = "gsutil", prefix = "", sha_prefix = "", zip
         **kwargs
     )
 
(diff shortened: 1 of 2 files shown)


siggisim:
Had an implementation for diffing, but it's gross

```
cmd = "echo \"find -f $(SRCS) | sed 's|$$(echo $(SRCS) | awk '{ print $$1 }' | xargs dirname)/||g' | grep . | sort > local && %s ls -r gs://%s/%s | sed 's|gs://%s/%s||g' | sed 's|/:||g' | sed 's|:||g' | grep . | sort > remote && diff local remote\" > $@" % (gsutil, bucket, prefix, bucket, prefix),
```

tylerwilliams (approved):
LGTM

PR #812, opened 2021-07-31
Re-enable ComplexActionIO test

bduffany:
I'm no longer able to repro the flakiness w/ 100 test runs, probably due to misc improvements we've been making to the executor + testing infra. However, it was very slow (5s) since it was copying around 100MB+ worth of files several times. So this PR updates the test to do a bit less IO (now it's about 1s which is more reasonable).

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/462


tylerwilliams (approved):
LGTM

PR #816, opened 2021-08-02
Explains emtpy action page

pariparajuli:
Notifies user that empty action page with no error means it was removed from CAS


app/invocation/invocation_action_card.tsx
```diff
@@ -237,11 +237,16 @@ export default class InvocationActionCardComponent extends React.Component<Props
                     )}
                   </div>
                 </div>
+              ) : (
+                <div>
+                  No action profile was found. If there are no error messages, the action profile was likely removed
```
siggisim, line 242: It's a bit more complicated than this, I'd just keep it simple for now and say something like `No action details were found.`

bduffany (approved):

siggisim (approved):

PR #817, opened 2021-08-02
Fix darwin executor build

siggisim:
Looks like I messed up this function signature when splitting darwin/linux firecracker packages.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

argh, my bad

PR #818, opened 2021-08-02
Downloadable output files in action page

pariparajuli:
Output files in the action page can now be downloaded

### DIFF
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -48,7 +48,7 @@
   color: #222;
 }
 
-.action-section.div {
+.action-section div {
   flex-grow: 1;
 }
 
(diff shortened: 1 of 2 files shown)


app/invocation/invocation.css
```diff
@@ -40,15 +40,15 @@
...
 }
 
 .action-section {
   display: flex;
   margin-bottom: 16px;
+  color: #222;
 }
 
-.action-section div {
+.action-section.div {
```
bduffany, line 51: Looks like this was unintentional?

bduffany (changes requested):
Nice!

bduffany (approved):

PR #823, opened 2021-08-03
Add role filtering support for invocation stats query

bduffany:
The proto already allows filtering by role, but the actual filtering wasn't implemented. This adds the SQL wiring and index necessary to make the query complete in a reasonable amount of time.

See performance details here https://github.com/buildbuddy-io/buildbuddy-internal/issues/617

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/617


tylerwilliams (approved):
LGTM

PR #824, opened 2021-08-03
Add role selector to global filter + tweaks

bduffany:
* Add filter dropdown with role + status
* Wire up selected role to history + stats queries (https://github.com/buildbuddy-io/buildbuddy/pull/823 contains the backend impl + optimizations)
* Move datepicker to the upper right of the page and remove blue background when filters are active, in favor of a more subtle filled-button indicator
* Improve styling of our "select" widget so it's more consistent with the button widget
* Set `overflow-y: scroll;` on the main page content to force the scrollbar to always be displayed, even if the page is not scrollable (this excludes Mac, where scrollbars are rendered as an overlay on top of the page and don't affect layout, IIUC). This reduces jank when navigating between pages, setting new filters, etc. (the scrollbar pops in and out of view somewhat erratically as content is loaded and unloaded, causing the UI to reflow and creating a more jarring experience when navigating)

---

**Version bump**: Patch


siggisim:
The select dropdown is a bit visually distracting and takes up a lot of space. I'm also thinking we'll add some more filters here eventually (branch, repo, user etc).

What do you think about putting just a little filter icon to the left of the date picker which, when clicked, opens a little overlay similar to the date picker? There we could do a multi-select role (allowing people to select CI runs and normal invocations, but not workflows for example). It'll also be less visually distracting and give us space to add the other filters.

Open to other ideas, just wanted to throw this proposal out there.

bduffany:
@siggisim A dropdown sounds good, seems like we'd be able to scale better that way. Will play around with some concepts tomorrow

bduffany:
Updated so that filtering options are in a dropdown:

https://user-images.githubusercontent.com/2414826/128405625-edc06abf-434e-49f8-a7d9-5d5fb032dbfb.mp4

Status options aren't implemented yet; will add it in another PR.

enterprise/app/filter/filter.tsx
```diff
@@ -49,43 +68,88 @@ export default class FilterComponent extends React.Component<FilterProps, State>
     const startDate = (startDateParam ? moment(startDateParam) : moment().subtract(7, "days")).toDate();
     const endDate = (endDateParam ? moment(endDateParam) : moment()).toDate();
 
-    const isFiltering = Boolean(this.props.search.toString());
+    const canClearFilters = Boolean(this.props.search.toString());
```
siggisim, line 71: Will this trigger from the artifact or target filters?

enterprise/app/filter/filter.tsx
```diff
@@ -49,43 +68,88 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
         )}
-        <div className="date-picker-container">
-          <OutlinedButton className="date-picker-button" onClick={this.onOpenDatePicker.bind(this)}>
-            <img src="/image/calendar.svg" alt="" />
+        <div className="popup-wrapper">
+          <OutlinedButton className="filter-menu-button icon-text-button" onClick={this.onOpenFilterMenu.bind(this)}>
+            <img className="subtle-icon" src="/image/filter.svg" alt="" />
+            {roleValue === "CI" && <span className="role-badge CI">CI</span>}
+            {roleValue === "CI_RUNNER" && <span className="role-badge CI_RUNNER">Workflow</span>}
+            {!roleValue && <span className="filter-menu-button-label">Filter...</span>}
```
siggisim, line 87: Can we just make it the icon if no filter is selected? i.e. remove `Filter...`

enterprise/app/filter/filter.tsx
```diff
@@ -49,43 +68,88 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
-            onClick={this.onClickClearFilters.bind(this)}>
-            <img src="/image/x.svg" alt="" />
-          </OutlinedButton>
-        ) : (
-          <div className="filter-icon-container">
-            <img src="/image/filter.svg" alt="" />
-          </div>
+      <div className={`global-filter ${canClearFilters ? "is-filtering" : ""}`}>
+        {canClearFilters && (
+          <FilledButton className="square" title="Clear filters" onClick={this.onClickClearFilters.bind(this)}>
```
siggisim, line 78: Should we just show this if a filter is applied - not a date range?

siggisim (approved):
This looks good, just a couple of nits.

enterprise/app/filter/filter.tsx
bduffany: OK, sounds good -- maybe we can have a dedicated "Default" time range in the datepicker instead, and maybe make the text a different color when a non-default range is selected, or something.

bduffany: Done

enterprise/app/filter/filter.tsx
bduffany: (Updated this logic so it only depends on the params controlled via the dropdown)

enterprise/app/filter/filter.tsx
bduffany: Done!

siggisim (approved):

PR #826, opened 2021-08-03
Clear test.xml file if one is not found

siggisim:
This fixes the issue where switching between test shards where some have and some don't have test.xml files leads to stale data.

---

**Version bump**: Minor


tylerwilliams (approved):
LGTM

bduffany:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/764

PR #831, opened 2021-08-04
Add status filter field to invocation query protos

bduffany:
Since the client's notion of "status" is really derived from 2 fields (`invocation_status` and `success`) it makes sense to define a new enum that represents this derived status, so that we can specify a list of statuses to fetch (which will be driven by the new global filtering UI).

In upcoming PRs, will update the client to always send all possible status filter values (by default -- this will later be controlled by the filtering UI), and then update the server to handle this new field appropriately.

Confirmed that status filtering does not make queries significantly slower for group IDs with the largest number of invocations (tried history, stats, and trends queries) -- even when combined with `role` filtering.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/617, https://github.com/buildbuddy-io/buildbuddy-internal/issues/647

### DIFF
--- a/proto/invocation.proto
+++ b/proto/invocation.proto
@@ -177,7 +177,8 @@ message InvocationQuery {
   // The timestamp up to which the build was last updated (exclusive).
   google.protobuf.Timestamp updated_before = 8;
 
-  // Allowed invocation statuses. Must be explicitly specified, or no invocations will be returned.
+  // Allowed invocation statuses. Must be explicitly specified, or no
+  // invocations will be returned.
   repeated StatusFilterValue statuses = 9;
 }
 
@@ -308,7 +309,8 @@ message InvocationStatQuery {
   // The timestamp up to which the build was last updated (exclusive).
   google.protobuf.Timestamp updated_before = 8;
 
-  // Allowed invocation statuses. Must be explicitly specified, or no invocations will be returned.
+  // Allowed invocation statuses. Must be explicitly specified, or no
+  // invocations will be returned.
   repeated StatusFilterValue statuses = 9;
 }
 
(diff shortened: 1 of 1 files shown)


proto/invocation.proto
```diff
@@ -176,6 +176,25 @@ message InvocationQuery {
 
   // The timestamp up to which the build was last updated (exclusive).
   google.protobuf.Timestamp updated_before = 8;
+
+  // Allowed invocation statuses.
+  repeated StatusFilterValue statuses = 9;
+}
+
+enum StatusFilterValue {
```
bduffany, line 184: Also considered `CompositeStatus` and `ClientStatus` but not sure if those are very intuitive -- open to other suggestions

proto/invocation.proto
```diff
@@ -288,6 +308,10 @@ message InvocationStatQuery {
 
   // The timestamp up to which the build was last updated (exclusive).
   google.protobuf.Timestamp updated_before = 8;
+
+  // Allowed invocation statuses. Must be explicitly specified, or no
+  // invocations will be returned.
+  repeated StatusFilterValue statuses = 9;
```
tylerwilliams, line 314: status

proto/invocation.proto
tylerwilliams: OverallStatus? InvocationStatus is taken I guess? Maybe @siggisim has ideas

tylerwilliams (changes requested):

proto/invocation.proto
bduffany: `OverallStatus` SG, if that works for Siggi :)

proto/invocation.proto
bduffany: Fixed

siggisim (approved):

tylerwilliams (approved):
LGTM

PR #839, opened 2021-08-05
Filtering for RBE actions

pariparajuli:
Filter/search with actionDigest for RBE actions

### DIFF
--- a/app/invocation/invocation_execution_card.tsx
+++ b/app/invocation/invocation_execution_card.tsx
@@ -261,7 +261,7 @@ export default class ExecutionCardComponent extends React.Component {
         incompleteCount++;
       }
     }
-    console.log(this.props.filter);
+
     const filteredActions = this.state.executions.filter(
       (action) =>
         !this.props.filter ||
@@ -269,7 +269,6 @@ export default class ExecutionCardComponent extends React.Component {
           .toLowerCase()
           .includes(this.props.filter.toLowerCase())
     );
-    console.log(filteredActions);
 
     return (
       <div className={`card expanded`}>
--- a/app/invocation/invocation_execution_card.tsx
+++ b/app/invocation/invocation_execution_card.tsx
@@ -11,6 +11,7 @@ import rpcService from "../service/rpc_service";
 interface Props {
   model: InvocationModel;
   inProgress: boolean;
+  filter: string;
 }
 
 interface State {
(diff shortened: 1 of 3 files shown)


pariparajuli:
Screen Shot 2021-08-05 at 4 38 33 PM
Screen Shot 2021-08-05 at 4 39 52 PM
Screen Shot 2021-08-05 at 4 42 02 PM

app/invocation/invocation_execution_card.tsx
```diff
@@ -260,6 +261,15 @@ export default class ExecutionCardComponent extends React.Component {
         incompleteCount++;
       }
     }
+    console.log(this.props.filter);
```
bduffany, line 264: nit: remove `console.log`

app/invocation/invocation_execution_card.tsx
```diff
@@ -260,6 +261,15 @@ export default class ExecutionCardComponent extends React.Component {
...
+    const filteredActions = this.state.executions.filter(
+      (action) =>
+        !this.props.filter ||
+        `${action.actionDigest.hash}/${action.actionDigest.sizeBytes}`
```
bduffany, line 268: Seems useful to be able to filter by the command, e.g. `go`, `bash`, `tsc`, ... if it doesn't slow down the search by too much, mind adding that here too?

pariparajuli: I added that and made the placeholder for the filter input a bit different so it's clear what the user can filter by

app/invocation/invocation_execution_card.tsx
```diff
@@ -316,42 +326,48 @@ export default class ExecutionCardComponent extends React.Component {
...
                     </div>
-                  </div>
-                </div>
-              );
-            })}
+                  );
+                })}
+              </div>
+            ) : (
+              <div>No matching actions.</div>
```
bduffany, line 369: Sorry I didn't notice this on the last round -- I think maybe we should show "No actions found" if the user is not filtering?

bduffany (approved):

app/invocation/invocation_execution_card.tsx
bduffany: Resolved offline -- when no actions are found there's a separate message that gets displayed.

PR #840, opened 2021-08-06
Don't panic if js healthcheck returns error

siggisim:
Also caches a successful result so we make sure the js bundle exists on CDN, but we don't fetch a ~1MB file every 3 seconds from every App. 

---

**Version bump**: Minor


tylerwilliams (approved):
LGTM

PR #841, opened 2021-08-06
Implement server-side invocation history status filtering and trends filtering

bduffany:
* Implement status filtering for `SearchInvocationRequest`, `GetInvocationStatRequest`, `GetTrendRequest`
  * Change the documentated semantics for the `status` filter -- rather than requiring all desired statuses to be specified, it is more natural (and also backwards-compatible) to select all statuses by default, and for each filter to map one-to-one with an OR clause in SQL. It is also more natural for the user to build up a list of statuses they want to see (starting from an empty list, meaning "all statuses"), rather than subtract the ones they don't want to see.
* Update `GetTrendRequest` to allow filtering by the new timestamp fields instead of the `lookback_window`. Will remove the `lookback_window` support in another PR (once clients are using the new timestamp fields).

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/invocation_stat_service/invocation_stat_service.go
+++ b/enterprise/server/invocation_stat_service/invocation_stat_service.go
@@ -58,14 +58,6 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
 		return nil, status.ResourceExhaustedErrorf("Too many rows.")
 	}
 
-	lookbackWindowDays := 7 * 24 * time.Hour
-	if w := req.GetLookbackWindowDays(); w != 0 {
-		if w < 1 || w > 365 {
-			return nil, status.InvalidArgumentErrorf("lookback_window_days must be between 0 and 366")
-		}
-		lookbackWindowDays = time.Duration(w*24) * time.Hour
-	}
-
 	q := query_builder.NewQuery(fmt.Sprintf("SELECT %s as name,", i.h.DateFromUsecTimestamp("updated_at_usec")) + `
 	    SUM(CASE WHEN duration_usec > 0 THEN duration_usec END) as total_build_time_usec,
 	    COUNT(1) as total_num_builds,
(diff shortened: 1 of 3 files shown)


enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -112,7 +104,27 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
...
+		// or default to 7 days.
+		lookbackWindowDays := 7 * 24 * time.Hour
+		if w := req.GetLookbackWindowDays(); w != 0 {
+			if w < 1 || w > 365 {
+				return nil, status.InvalidArgumentErrorf("lookback_window_days must be between 0 and 366")
+			}
+			lookbackWindowDays = time.Duration(w*24) * time.Hour
+		}
+		q.AddWhereClause("updated_at_usec >= ?", timeutil.ToUsec(time.Now().Add(-lookbackWindowDays)))
+	}
```
tylerwilliams, line 120: Are there other places in the code we do this? if not, you can add a comment that this can be deleted in the next release (same for the proto field)

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -112,7 +104,27 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
...
+
+	if end := req.GetQuery().GetUpdatedBefore(); end.IsValid() {
+		q.AddWhereClause("updated_at_usec < ?", timeutil.ToUsec(end.AsTime()))
+	}
+
+	addStatusPredicate(q, req.GetQuery().GetStatus())
```
tylerwilliams, line 126: I think this would be clearer if "addStatusPredicate" returned a *query_builder.OrClauses and then you added them to the query here. That way someone reading this code can kind of see the whole query / how it's built in one place.

If we did this, we'd probably also want to call addStatusPredicate buildStatusPredicate instead.

tylerwilliams (changes requested):
looks good, minus a couples of nits

enterprise/server/invocation_stat_service/invocation_stat_service.go
bduffany: Nope, we don't have other `updated_at` filters that would need to be migrated to the "new way." Added a comment here and to the proto as well.

enterprise/server/invocation_stat_service/invocation_stat_service.go
bduffany: sg, went with `toStatusClauses` instead of `buildStatusPredicate` to avoid confusion since we're calling `.Build` on the thing being returned

tylerwilliams (approved):
LGTM

PR #842, opened 2021-08-06
Set cache headers for static resources with a `hash` query parameter

siggisim:
---

**Version bump**: Minor


tylerwilliams (approved):
LGTM

PR #845, opened 2021-08-06
Removed filter for empty actions

pariparajuli:
Moved filterComponenet inside executionComponent to remove it when there are no actions


pariparajuli:
Screen Shot 2021-08-06 at 5 07 41 PM
Screen Shot 2021-08-06 at 5 08 37 PM

bduffany (approved):

PR #851, opened 2021-08-09
Clickable bars on trends page

pariparajuli:
Drill into chart details for builds, build duration, users, commits, hosts, and repos charts


enterprise/app/trends/trends_chart.tsx
```diff
@@ -51,9 +65,12 @@ const TrendsChartTooltip = ({
 export default class TrendsChartComponent extends React.Component {
   props: Props;
 
+  handleClick(date: any) {
```
bduffany, line 68: nit: change `any` to `string`

enterprise/app/trends/trends_chart.tsx
```diff
@@ -77,7 +94,20 @@ export default class TrendsChartComponent extends React.Component {
...
+              {this.props.data.map((date, index) => (
+                <Cell
+                  cursor={this.props.clickableBars ? "pointer" : "default"}
+                  key={`cell-${index}`}
+                  onClick={this.props.clickableBars ? this.handleClick.bind(this, date) : null}
+                />
+              ))}
+            </Bar>
             {this.props.extractSecondaryValue && this.props.secondaryLine && (
               <Line
```
bduffany, line 112: Is there a way to set `pointer-events: none` on this line? Sometimes it gets in the way of the bars and it causes flickering:

https://user-images.githubusercontent.com/2414826/128788440-0e8d4c25-393e-44eb-91bc-bdf4c052291f.mp4

enterprise/app/trends/trends_chart.tsx
```diff
@@ -51,9 +65,12 @@ const TrendsChartTooltip = ({
 export default class TrendsChartComponent extends React.Component {
   props: Props;
 
+  handleClick(date: any) {
+    router.navigateTo("/?start=" + date + "&end=" + date + this.props.hash);
```
bduffany, line 69: nit: if you replace `this.props.hash` with `(this.props.hash || "")` then you can avoid setting the empty `hash` prop on some of the charts

bduffany (dismissed):
Looks awesome :) just some nits

enterprise/app/trends/trends.tsx
```diff
@@ -260,6 +264,8 @@ export default class TrendsComponent extends React.Component<Props> {
                 formatHoverLabel={this.formatLongDate}
                 formatHoverValue={(value) => (value || 0) + " users"}
                 name="users with builds"
+                clickableBars={true}
```
siggisim, line 267: Rather than passing in a boolean and a hash and having the chart component know how to navigate - I think it would be cleaner to pass in an `onDateClicked` function. 

If this function prop is set, the bars become clickable - and then the chart doesn't have to know what a router is or what a hash is.

siggisim (changes requested):

enterprise/app/trends/trends.tsx
pariparajuli: Is there a way to do this without passing in the hash? I'm not completely familiar with the intricacies but if i pass an onValueClicked function into a chart, is there a way to pass a hash parameter in with the function and then put in the date parameter later?

pariparajuli: Got it!

enterprise/app/trends/trends.tsx
```diff
@@ -186,6 +191,7 @@ export default class TrendsComponent extends React.Component<Props> {
                 secondaryName="average build time seconds"
                 secondaryLine={true}
                 separateAxis={true}
+                barClicked={this.onBarClicked.bind(this, "")}
```
bduffany, line 194: nit: usually callback function parameters are prefixed with `on` -- can you rename to `onBarClicked`?

siggisim (commented):
Should this live behind @bduffany's global filter capabilities flag, or should we wait to submit it until those changes are in?

bduffany:
> Should this live behind @bduffany's global filter capabilities flag, or should we wait to submit it until those changes are in?

Yeah, this should go behind that flag because otherwise there is no obvious way to change the date filter once it is applied (especially since URL params are now sticky). And I changed the impl so that we only respect the filtering params when the capabilities flag is set.

pariparajuli:
I'll throw this behind the global filter flag

siggisim (approved):
This looks great!

PR #852, opened 2021-08-09
Add global filter to trends page

bduffany:
* Behind `capabilities.globalFilter`, remove the `7d`, `30d` etc. tabs as well as the "Show CI builds only" checkbox, and show the global filter instead. Don't show the "workflows" option since it's confusing at the moment -- the "All" option doesn't request workflows (to avoid double-counting build times), and it would be awkward to write "All except workflows"
* Improve the "static ranges" available in the date picker -- for now, setting them to match the options for trends (7d, 30d, 90d, 180d, 365d)
* Use the correct date picker defaults to be consistent with what we have today ("All time" on history page, 30d on trends page -- for the TAP page it'll be 7d)
* Not messing with timezones for now -- will switch to a consistent timezone in a follow-up PR. This PR should not affect any of our existing queries, since it's all behind a flag.

https://user-images.githubusercontent.com/2414826/128776200-58ac305c-26c1-47d2-86cc-f3908a7e9fce.mp4

---

**Version bump**: Patch


app/format/format.tsx
```diff
@@ -113,10 +116,26 @@ export function formatTimestamp(timestamp: { seconds?: number | Long; nanos?: nu
...
+export function formatDateRange(startDate: Date, endDate: Date, { now = new Date() } = {}) {
   let startFormat, endFormat;
-  const now = new Date();
+
+  // Special cases for date range picker default options
+  if (isSameDay(now, endDate)) {
+    if (isSameDay(startDate, LOCAL_EPOCH)) {
+      return "All time";
+    }
+    for (const n of LAST_N_DAYS_OPTIONS) {
```
siggisim, line 129: Could this be simplified as:
```
if (isSameDay(now, endDate)) {
    if (isSameDay(startDate, LOCAL_EPOCH)) {
      return "All time";
    }
  return `Last ${daysBetween(startDate, endDate)} days`;
}
```

Seems weird for `format.tsx` to have to know about `LAST_N_DAYS_OPTIONS`, and it's probably more readable to say `Last 37 days` than showing the dates.

enterprise/app/filter/filter.tsx
```diff
@@ -99,15 +130,50 @@ export default class FilterComponent extends React.Component<FilterProps, State>
     const startDateParam = this.props.search.get(START_DATE_PARAM_NAME);
     const endDateParam = this.props.search.get(END_DATE_PARAM_NAME);
 
-    const startDate = (startDateParam ? moment(startDateParam) : moment().subtract(7, "days")).toDate();
-    const endDate = (endDateParam ? moment(endDateParam) : moment()).toDate();
+    const defaultStartDate = getDefaultStartDateForPage(this.props.path);
```
siggisim, line 133: Let's just keep it simple and do Last 30 days for everything instead of having a different default start date per page.

enterprise/app/filter/filter.tsx
```diff
@@ -99,15 +130,50 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
     const roleValue = this.props.search.get(ROLE_PARAM_NAME) || "";
     const statusValue = this.props.search.get(STATUS_PARAM_NAME) || "";
 
     const isFiltering = Boolean(roleValue || statusValue);
     const selectedStatuses = new Set(parseStatusParam(statusValue));
 
+    const now = new Date();
+    const presetDateRanges: PresetRange[] = [
+      {
+        label: `Default (${formatDateRange(defaultStartDate, now, { now })})`,
```
siggisim, line 146: Do people need to know or care about the "default" date range?

Can we just default them to last 30 days and then they can pick whatever date range they want?

enterprise/app/filter/filter.tsx
```diff
@@ -147,10 +213,13 @@ export default class FilterComponent extends React.Component<FilterProps, State>
                     <Radio value="CI" checked={roleValue === "CI"} />
                     <span className="role-badge CI">CI</span>
                   </label>
-                  <label onClick={this.onRoleChange.bind(this, "CI_RUNNER")}>
-                    <Radio value="CI_RUNNER" checked={roleValue === "CI_RUNNER"} />
-                    <span className="role-badge CI_RUNNER">Workflow</span>
-                  </label>
+                  {/* TODO(bduffany): Figure out how to deal with workflows for the trends page */}
```
siggisim, line 216: Can we just leave this in instead of adding special case logic?

enterprise/app/trends/trends.tsx
```diff
@@ -66,14 +70,32 @@ export default class TrendsComponent extends React.Component<Props> {
...
-      request.query.role = ["CI"];
+    if (capabilities.globalFilter) {
+      const filterParams = getProtoFilterParams(this.props.path, this.props.search);
+      if (filterParams.role) {
+        request.query.role = [filterParams.role];
+      } else {
+        // Note: Technically we're filtering out workflows and unknown roles,
+        // even though the user has selected "All roles". But we do this to
+        // avoid double-counting build times for workflows and their nested CI runs.
+        request.query.role = ["", "CI"];
```
siggisim, line 86: I think we should just let people pick if they want to see all or if they want to filter out workflows (using the new picker). I think we should avoid this kind of implicit behavior.

enterprise/app/filter/filter.tsx
```diff
@@ -99,15 +130,50 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
+      },
+      ...LAST_N_DAYS_OPTIONS.map((n) => {
+        const start = moment(now)
+          .add(-n + 1, "days")
+          .toDate();
+        return {
+          label: formatDateRange(start, now, { now }),
+          isSelected: (range: Range) => {
+            return (
+              this.props.search.get(START_DATE_PARAM_NAME) &&
```
siggisim, line 164: What happens if a users selects Last 90 days and then refreshes the next day? Will they see yesterday's data and their selection is lost?

siggisim (changes requested):
Looks good, just trying to simplify some of the complexity here.

enterprise/app/filter/filter.tsx
bduffany: Done -- the intent here was just to be able to "clear" the filter, but now that we can express relative date ranges in the URL, I think the ability to clear isn't as important.

enterprise/app/filter/filter.tsx
bduffany: Done -- after https://github.com/buildbuddy-io/buildbuddy/pull/861 is submitted the concern here will go away.

enterprise/app/filter/filter.tsx
```diff
@@ -177,6 +233,17 @@ export default class FilterComponent extends React.Component<FilterProps, State>
             <DateRangePicker
               ranges={[{ startDate, endDate, key: "selection" }]}
               onChange={this.onDateChange.bind(this)}
+              // When showing "All time" we don't want to set the currently
+              // visible month to the Unix epoch... so always show the end
+              // date when initially rendering the component
+              shownDate={endDate}
+              staticRanges={
+                // We've added a custom "isDefault" prop to the date range
```
siggisim, line 241: Comment is out of date now I think

siggisim (approved):

enterprise/app/filter/filter.tsx
bduffany: Fixed

PR #859, opened 2021-08-10
Pull container image outside of execution context timeout

tylerwilliams:
See title.

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -152,7 +152,7 @@ type CommandRunner struct {
 	diskUsageBytes   int64
 }
 
-func (r *CommandRunner) PrepareForTask(task *repb.ExecutionTask) error {
+func (r *CommandRunner) PrepareForTask(ctx context.Context, task *repb.ExecutionTask) error {
 	r.Workspace.SetTask(task)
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -165,6 +165,12 @@ func (r *CommandRunner) PrepareForTask(task *repb.ExecutionTask) error {
 	if err := r.Workspace.CreateOutputDirs(); err != nil {
 		return status.UnavailableErrorf("Error creating output directory: %s", err.Error())
 	}
+
+	// Pull the container image before Run() is called, so that we don't
+	// use up the whole exec ctx timeout with a slow container pull.
+	if err := r.Container.PullImageIfNecessary(ctx); err != nil {
+		return status.UnavailableErrorf("Error pulling container: %s", err)
+	}
```
bduffany, line 173: Maybe add a comment to the `Run` func that the `PullImageIfNecessary` should be a NOP since we pull in `PrepareTask`? (I understand not wanting to remove the pull from that func, but might be confusing to future readers why we pull twice)

bduffany (approved):
Nice! I think this will also make our `ExecutedActionMetadata` metrics a bit more accurate since execution time was previously including the image pull.

PR #861, opened 2021-08-10
Make role filters work the same way as status filters

bduffany:
Instead of radio buttons, now using checkboxes.

Required an API change to allow selecting any combination of roles. The API change is backwards-compatible.

image

---

**Version bump**: Patch


siggisim (approved):

tylerwilliams (approved):
proto LGTM

PR #863, opened 2021-08-11
Add a very basic runner implementation.

tylerwilliams:
See title.


enterprise/server/runner/runner.go
```diff
@@ -0,0 +1,247 @@
...
+			"--pushed_branch=master",
+			"--target_branch=master",
+			"--invocation_id=" + invocationID,
+		},
+		Platform: &repb.Platform{
+			Properties: []*repb.Platform_Property{
+				{Name: "container-image", Value: "docker://gcr.io/flame-public/buildbuddy-ci-runner:v2.2.7"},
+				{Name: "recycle-runner", Value: "true"},
+				{Name: "preserve-workspace", Value: "true"},
+				{Name: "workload-isolation-type", Value: "firecracker"},
```
bduffany, line 132: :fire: :fire: :fire:

enterprise/server/runner/runner.go
```diff
@@ -0,0 +1,247 @@
...
+			"--pushed_repo_url=" + req.GetGitRepo().GetRepoUrl(),
+			"--target_repo_url=" + req.GetGitRepo().GetRepoUrl(),
+			"--bazel_sub_command=" + req.GetBazelCommand(),
+			"--pushed_branch=master",
+			"--target_branch=master",
+			"--invocation_id=" + invocationID,
+		},
+		Platform: &repb.Platform{
+			Properties: []*repb.Platform_Property{
+				{Name: "container-image", Value: "docker://gcr.io/flame-public/buildbuddy-ci-runner:v2.2.7"},
```
bduffany, line 129: nit: mind putting a reminder at the top of the same const `workflow/service.go` to remember to update this string if that one changes? (A shared library might also make sense but I get that we might not want to bother with that for now)

enterprise/server/runner/runner.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return nil, err
+	}
+	invocationID := guid.String()
+	actionDigest, err := r.createAction(ctx, req, invocationID)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Printf("Uploaded action to cache. Digest: %s/%d", actionDigest.GetHash(), actionDigest.GetSizeBytes())
```
bduffany, line 170: nit: Infof or Debugf?

bduffany (approved):

enterprise/server/runner/runner.go
```diff
@@ -0,0 +1,247 @@
...
+			{Name: "REPO_TOKEN", Value: req.GetGitRepo().GetAccessToken()},
+		},
+		Arguments: []string{
+			"./" + runnerName,
+			"--bes_backend=" + conf.GetAppEventsAPIURL(),
+			"--bes_results_url=" + conf.GetAppBuildBuddyURL() + "/invocation/",
+			"--commit_sha=" + req.GetRepoState().GetCommitSha(),
+			"--pushed_repo_url=" + req.GetGitRepo().GetRepoUrl(),
+			"--target_repo_url=" + req.GetGitRepo().GetRepoUrl(),
+			"--bazel_sub_command=" + req.GetBazelCommand(),
```
bduffany, line 122: I think your original idea of calling this `bazel_args` actually makes more sense now (I originally thought it would be a string like `build`, `test` etc. which I think of as the sub-command, but I see now it also includes the target etc.)

tylerwilliams:
Closing because this was merged as https://github.com/buildbuddy-io/buildbuddy/pull/1177

PR #864, opened 2021-08-11
Add some unit tests for query_builder.OrClauses

bduffany:
There's a few places in the code where we rely on `query_builder.OrClauses` returning an empty string when there are no clauses present, but it feels a bit brittle since that function could just as easily return `" "` which still seems like a valid result (especially since the current impl pads the returned string with whitespace in the non-empty case). Adding a unit test to enforce this expected behavior and avoid accidentally breaking stuff in the future.

---

**Version bump**: Patch


vadimberezniker (approved):
LGTM

PR #865, opened 2021-08-11
Filtering actions by status

pariparajuli:
Filtering actions by status: all, starting, cache check, queued, executing, succeeded, and errored

### DIFF
--- a/app/invocation/invocation_execution_card.tsx
+++ b/app/invocation/invocation_execution_card.tsx
@@ -283,10 +283,8 @@ export default class ExecutionCardComponent extends React.Component {
       )
       .filter(
         (action) =>
-          this.state.statusFilter.includes("all") ||
-          getExecutionStatus(action).name.toLowerCase().includes(this.state.statusFilter.toLowerCase()) ||
-          (getExecutionStatus(action).name.toLowerCase().includes("failed") &&
-            this.state.statusFilter.includes("error"))
+          this.state.statusFilter === "all" ||
+          getExecutionStatus(action).name.toLowerCase().startsWith(this.state.statusFilter)
       );
 
     return (
@@ -313,6 +311,7 @@ export default class ExecutionCardComponent extends React.Component {
                   <Option value="queued">Queued</Option>
                   <Option value="executing">Executing</Option>
                   <Option value="succeeded">Succeeded</Option>
+                  <Option value="failed">Failed</Option>
                   <Option value="error">Errored</Option>
                 </Select>
                 <span className="invocation-sort-title">Sort by</span>
--- a/app/invocation/invocation_execution_card.tsx
+++ b/app/invocation/invocation_execution_card.tsx
@@ -21,6 +21,7 @@ interface State {
   executions: execution_stats.Execution[];
   sort: string;
   direction: "asc" | "desc";
+  statusFilter: string;
 }
 
 const ExecutionStage = build.bazel.remote.execution.v2.ExecutionStage;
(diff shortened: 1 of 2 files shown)


pariparajuli:
Screen Shot 2021-08-11 at 1 41 34 PM
Screen Shot 2021-08-11 at 1 41 54 PM
Screen Shot 2021-08-11 at 1 42 28 PM

app/invocation/invocation_execution_card.tsx
```diff
@@ -264,14 +272,22 @@ export default class ExecutionCardComponent extends React.Component {
...
+        (action) =>
+          !this.props.filter ||
+          `${action.actionDigest.hash}/${action.actionDigest.sizeBytes}`
+            .toLowerCase()
+            .includes(this.props.filter.toLowerCase()) ||
+          action.commandSnippet.toLowerCase().includes(this.props.filter.toLowerCase())
+      )
+      .filter(
+        (action) =>
+          this.state.statusFilter.includes("all") ||
```
bduffany, line 286: You can use `===` here instead of `includes` for `this.state.statusFilter`, here and below

app/invocation/invocation_execution_card.tsx
```diff
@@ -264,14 +272,22 @@ export default class ExecutionCardComponent extends React.Component {
...
+          getExecutionStatus(action).name.toLowerCase().includes(this.state.statusFilter.toLowerCase()) ||
```
bduffany, line 287: nit: `this.state.statusFilter` should already be lowercase, no need to lowercase it

app/invocation/invocation_execution_card.tsx
```diff
@@ -289,6 +305,16 @@ export default class ExecutionCardComponent extends React.Component {
...
               <div className="invocation-sort-controls">
+                <span className="invocation-filter-title">Show</span>
+                <Select onChange={this.handleStatusFilterChange.bind(this)} value={this.state.statusFilter}>
+                  <Option value="all">All</Option>
+                  <Option value="starting">Starting</Option>
+                  <Option value="cache check">Cache Check</Option>
+                  <Option value="queued">Queued</Option>
+                  <Option value="executing">Executing</Option>
+                  <Option value="succeeded">Succeeded</Option>
+                  <Option value="error">Errored</Option>
```
bduffany, line 316: I think we want another option value `Failed` just below the `Succeeded` value, which corresponds to actions with `Failed (exit code ${blah})`

app/invocation/invocation_execution_card.tsx
```diff
@@ -264,14 +272,22 @@ export default class ExecutionCardComponent extends React.Component {
...
+          `${action.actionDigest.hash}/${action.actionDigest.sizeBytes}`
+            .toLowerCase()
+            .includes(this.props.filter.toLowerCase()) ||
+          action.commandSnippet.toLowerCase().includes(this.props.filter.toLowerCase())
+      )
+      .filter(
+        (action) =>
+          this.state.statusFilter.includes("all") ||
+          getExecutionStatus(action).name.toLowerCase().includes(this.state.statusFilter.toLowerCase()) ||
+          (getExecutionStatus(action).name.toLowerCase().includes("failed") &&
```
bduffany, line 288: I think the logic should be: If the lowercase `executionStatus.name` starts with "failed" then we should show it only if `filter === "failed'` -- and if it starts with "Error" then we show it only if `filter === "error"` (we want `startsWith` instead of `includes` bc errored actions might include FAILED, e.g. `Error (FAILED_PRECONDITION)`)

bduffany (changes requested):

app/invocation/invocation_execution_card.tsx
pariparajuli: Ah I see, I thought we wanted to consolidate the error & failed ones into one option

siggisim (approved):
Looks great!

PR #870, opened 2021-08-11
Use Pacific Time for stats and trends queries

bduffany:
Queries that aggregate by calendar day (`YYYY-MM-DD`) suffer from timezone issues because they try to convert Unix timestamps to calendar days. Currently, we implicitly use UTC as the timezone when doing this conversion. This is a problem because the new global filter does date filtering according to the client's local timezone, which may not be UTC.

We decided that we don't yet want to implement full timezone support (which would require passing the timezone string from the client to the server). But for the global filter, we also don't want the client to filter according to their local time, because we don't support that for stats/trends queries and it is confusing to not use the same timezone across the UI.

So, we decided to standardize on using PT (`America/Los_Angeles`) everywhere, and this PR is the first step towards that.

* Our deployed MySQL instances have timezone data available for `America/Los_Angeles` (I tested the new query against one of our replicas), so we are covered on that front.
* In JS we can reliably convert a `Date` to a YYYY-MM-DD date in Pacific Time using `date.toLocaleDateString("fr-CA", {timeZone: "America/Los_Angeles"})`, so we are covered on the client as well.

In subsequent PRs we'll update the UI to use Pacific Time as well, including the timestamps requested by the client. The client will make all requests using the same time zone that we expect on the server. So for example, when filtering to 2021-01-01, they'll see results only from 2021-01-01 PT.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/invocation_stat_service/invocation_stat_service.go
+++ b/enterprise/server/invocation_stat_service/invocation_stat_service.go
@@ -42,7 +42,7 @@ func (i *InvocationStatService) getAggColumn(aggType inpb.AggType) string {
 	case inpb.AggType_COMMIT_SHA_AGGREGATION_TYPE:
 		return "commit_sha"
 	case inpb.AggType_DATE_AGGREGATION_TYPE:
-		return i.h.DateFromUsecTimestamp("updated_at_usec")
+		return i.h.PTDateFromUsecTimestamp("updated_at_usec")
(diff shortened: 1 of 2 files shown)


vadimberezniker (approved):
LGTM

bduffany:
After some discussion, we decided to use a timezone offset instead. Closing

PR #873, opened 2021-08-12
Use local timezone offsets for date aggregations

bduffany:
After a bit of discussion, we decided to not use Pacific Time after all and go with a timezone offset, which is simpler to implement on the client and also provides better UX than standardizing on Pacific Time everywhere.

Because the client calculates dates using local historical timezone data (and not using this offset-based method), stuff can still get a little weird around daylight savings time. e.g. if PDT is currently being observed, then we will incorrectly calculate the date for builds near midnight on days where PST was observed. The effect of this is that when clicking certain bars on the trends page, some builds that contribute to the data for those bars will not be visible in the resulting history view.

In the future, we can correct this by issuing a quick SQL query while initializing the DB to see which time zone data is available (e.g. `SELECT * FROM mysql.time_zone_name`), and if found in that list then accept IANA strings from the client, otherwise fall back to the client's timezone offset.

---

**Version bump**: Patch


vadimberezniker (approved):
LGTM

tylerwilliams (approved):
LGTM

siggisim (approved):

PR #874, opened 2021-08-12
Fix ReadFile calls when using bundle filesystem

siggisim:
We call `fs.ReadFile(bundleFS, "sha.sum")` to get the sha of the app js bundle.

`ReadFile`'s documentation reads:
> If fs implements ReadFileFS, ReadFile calls fs.ReadFile. Otherwise ReadFile calls fs.Open and uses Read and Close on the returned file.

Because we were embedding `embed.FS` into `aliasFS`, we were implementing `ReadFileFS` which would not take our aliasing into account.

This only turned out to be a problem for builds built with `-c opt` that didn't specify an `app_directory`. Unfortunately this included the open source docker image builds between versions `2.3.3` and `2.4.9`.

Enterprise docker images and builds run with `bazel run` where not affected.

---

**Version bump**: Patch

### DIFF
--- a/bundle.go
+++ b/bundle.go
@@ -17,7 +17,7 @@ import (
 var all embed.FS
 
 type aliasFS struct {
-	embed.FS
+	FS            embed.FS
 	remappedPaths map[string]string
 }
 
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
LGTM

PR #875, opened 2021-08-12
Debug log stdout/stderr of execution failures.

tylerwilliams:
This makes it easier to see why executions are failing, which shortens the time it takes us to debug weird remote exec problems. I think the extra debug-level noise is worth it.


vadimberezniker (commented):
Is the exit code non zero in this case?

PR #877, opened 2021-08-12
Don't trigger handoff callback until the local write is done.

vadimberezniker:
---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #879, opened 2021-08-13
Display email address in sidebar and single-user group name if no name is set

siggisim:
---

**Version bump**: Minor

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -2,6 +2,7 @@ package tables
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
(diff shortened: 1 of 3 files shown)


server/tables/tables.go
```diff
@@ -225,7 +226,7 @@ func (u *User) ToProto() *uspb.DisplayUser {
 			Id: u.UserID,
 		},
 		Name: &uspb.Name{
-			Full:  u.FirstName + " " + u.LastName,
+			Full:  strings.TrimSpace(u.FirstName + " " + u.LastName),
```
bduffany, line 229: nit: should this be `TrimSpace(firstName) + " " + TrimSpace(lastName)` ?

bduffany (approved):

server/tables/tables.go
siggisim: The only thing I'm trying to achieve here is remove the space in case FirstName or LastName (or neither) are set.

tylerwilliams (approved):
LGTM

PR #748, opened 2021-07-20
Track unclaimed tasks in Redis.

vadimberezniker:
Sharing the state across schedulers will give us better scheduling
during upscaling.

We use a sorted set per executor pool to track unclaimed tasks.

---

**Version bump**: Patch

PR #883, opened 2021-08-13
Track unclaimed tasks in Redis.

vadimberezniker:
Sharing the state across schedulers will give us better scheduling
during upscaling.

We use a sorted set per executor pool to track unclaimed tasks.

This is a roll-forward of
https://github.com/buildbuddy-io/buildbuddy/pull/748.
This time around we don't rely on the availability of the ZRANDMEMBER
command which requires Redis 6.2
Instead we fetch the entire set of task IDs and do our own sampling.
The task ID set has a relatively small upper bound (10k) and in practice should be much smaller than that.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -1,7 +1,6 @@
 package scheduler_server
 
 import (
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):
LGTM

PR #884, opened 2021-08-13
Display errors passed through the `?error=` url query param using the error service

siggisim:
This uses the same BuildBuddyError parsing logic that will properly display golang-style errors.

Here's what this looks like:
Screen Shot 2021-08-13 at 11 56 02 AM

---

**Version bump**: Minor

### DIFF
--- a/app/errors/BUILD
+++ b/app/errors/BUILD
@@ -10,6 +10,7 @@ ts_library(
     ]),
     deps = [
         "//app/alert",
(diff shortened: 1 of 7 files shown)


siggisim:
cc @tylerwilliams

tylerwilliams:
This is great! Really stoked to finally bubble up those auth errors instead of just console logging them :)

PR #885, opened 2021-08-13
Update server side error redirects to pass error via new `?error=` url param

siggisim:
This depends on https://github.com/buildbuddy-io/buildbuddy/pull/884 in order to display these errors to users

---

**Version bump**: Minor

### DIFF
--- a/server/backends/github/github.go
+++ b/server/backends/github/github.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


server/backends/github/github.go
```diff
@@ -256,3 +250,8 @@ func getCookie(r *http.Request, name string) string {
 	}
 	return ""
 }
+
+func redirectWithError(w http.ResponseWriter, r *http.Request, err error) {
+	log.Warningf(err.Error())
```
tylerwilliams, line 255: uberubernit: here and other places, you don't need the "*f" variants of status errors or log functions if you're not doing formatting. so this could be log.Warning(err.Error())

this is such a nit i feel bad after even writing this comment.

tylerwilliams (approved):
LGTM

server/backends/github/github.go
siggisim: lol fixed, my logic is always just been to use the `f` version everywhere because it's an easy change when you inevitably want to print out some extra random crap

PR #886, opened 2021-08-13
Add an explicit error for when we try to create a user without an email address

siggisim:
This has been the source of much strife a couple of times now. Better error messaging will help us to detect this early.

---

**Version bump**: Minor


tylerwilliams (approved):
LGTM

PR #887, opened 2021-08-13
Display (real) errors coming from CreateUser or GetUser using the ErrorService

siggisim:
Right now we only console log errors from CreateUser / GetUser (we don't display them to the user).

This is because some of the errors are benign - specifically "User not found" from GetUser and "No user token available to fill user" from CreateUser.

This is mostly fine, but when we get real errors from CreateUser (like the auth token not containing an email address) - we just console log these errors which effectively swallows them.

This change clearly marks all of the server side errors that the client auth flow depends on and switches them to UNAUTHENTICATED errors. From the [status code docs](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L86):
>   // The caller does not have permission to execute the specified
  // operation. `PERMISSION_DENIED` must not be used for rejections
  // caused by exhausting some resource (use `RESOURCE_EXHAUSTED`
  // instead for those errors). `PERMISSION_DENIED` must not be
  // used if the caller can not be identified (use `UNAUTHENTICATED`
  // instead for those errors). This error code does not imply the
  // request is valid or the requested entity exists or satisfies
  // other pre-conditions.

It then updates the client to look for those `UNAUTHENTICATED` statuses instead of doing more brittle error message string matching.

We still accept the string matching approach for the client side for one release to handle the case where a new client hits an old server during the rollout. After the next release, we can remove any string matching.

---

**Version bump**: Minor

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -5,6 +5,8 @@ import { user_id } from "../../proto/user_id_ts_proto";
 import { user } from "../../proto/user_ts_proto";
 import capabilities from "../capabilities/capabilities";
 import rpcService from "../service/rpc_service";
+import errorService from "../errors/error_service";
+import { BuildBuddyError } from "../../app/util/errors";
 
 const SELECTED_GROUP_ID_LOCAL_STORAGE_KEY = "selected_group_id";
 
@@ -38,9 +40,7 @@ export class AuthService {
         this.emitUser(this.userFromResponse(response));
       })
       .catch((error: any) => {
-        console.log(error);
-        // TODO(siggisim): make this more robust.
-        if (error.includes("not found")) {
+        if (BuildBuddyError.parse(error).code == "Unauthenticated" || error.includes("not found")) {
           this.createUser();
         } else {
           this.onUserRpcError(error);
@@ -67,14 +67,18 @@ export class AuthService {
         this.refreshUser();
       })
       .catch((error: any) => {
-        this.onUserRpcError(error);
+        if (BuildBuddyError.parse(error).code == "Unauthenticated" || error.includes("No user token")) {
+          console.log("User was not created because no auth cookie was set, this is normal.");
+          this.emitUser(null);
+        } else {
+          this.onUserRpcError(error);
+        }
       });
   }
 
   onUserRpcError(error: any) {
-    console.log(error);
+    errorService.handleError(error);
     this.emitUser(null);
-    // TODO(siggisim): figure out what we should do in this case.
   }
 
   userFromResponse(response: user.GetUserResponse) {
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -40,6 +40,7 @@ export class AuthService {
         this.emitUser(this.userFromResponse(response));
       })
       .catch((error: any) => {
+        // TODO(siggisim): Remove "not found" string matching after the next release.
         if (BuildBuddyError.parse(error).code == "Unauthenticated" || error.includes("not found")) {
           this.createUser();
         } else {
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -68,7 +68,8 @@ export class AuthService {
         this.refreshUser();
       })
       .catch((error: any) => {
-        if (BuildBuddyError.parse(error).code == "Unauthenticated") {
+        // TODO(siggisim): Remove "No user token" string matching after the next release.
+        if (BuildBuddyError.parse(error).code == "Unauthenticated" || error.includes("No user token")) {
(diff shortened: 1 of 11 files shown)


server/util/capabilities/capabilities.go
```diff
@@ -55,7 +53,7 @@ func IsGranted(ctx context.Context, env environment.Env, cap akpb.ApiKey_Capabil
 	}
 	user, err := a.AuthenticatedUser(ctx)
 	if err != nil {
-		if isAnonymousUser := gstatus.Code(err) == codes.PermissionDenied; isAnonymousUser {
+		if isAnonymousUser := status.IsPermissionDeniedError(err) || status.IsUnauthenticatedError(err); isAnonymousUser {
```
tylerwilliams, line 56: @bduffany probably knows better here -- but do we need both or would it be better if isAnonymous were only true for unauthenticated errors?

app/auth/auth_service.ts
```diff
@@ -67,14 +68,19 @@ export class AuthService {
         this.refreshUser();
       })
       .catch((error: any) => {
-        this.onUserRpcError(error);
+        // TODO(siggisim): Remove "No user token" string matching after the next release.
+        if (BuildBuddyError.parse(error).code == "Unauthenticated" || error.includes("No user token")) {
```
tylerwilliams, line 72: Since I'm a typescript baby -- what's the reason for using BuildBuddyError instead of a status object here? (i went down this road because I was going to ask why you compared the code with a string instead of an enum / const)

tylerwilliams (approved):
LGTM (but i have questions ... lol)

server/util/capabilities/capabilities.go
siggisim: I think we should probably go through at some point and audit all uses for PermissionDenied vs Unauthenticated and make sure they conform to the status proto comment's definitions. This PR was just focused on the GetUser/CreateUser flows.

I think once we're confident that we're returning the right statuses in the right places, we can probably switch this (and the two other places where we do these kinds of checks) to just check for Unauthenticated. Until we do that, I think changing this could have unintended consequences.

bduffany: I agree with not doing it in this PR -- and I also agree that we should make the anonymous user case a bit more clean and less error-prone eventually. Maybe we could treat anonymous users as just a special kind of user with the `ANON` (or empty) group ID and then the anonymous concept is no longer needed (except in some rarer cases where we want to prevent anonymous users from doing certain things... in those cases we could check that the group ID is `ANON`). I think @vadimberezniker also was looking at some related logic semi-recently and he might have thoughts as well.

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -543,10 +543,10 @@ func (p *Pool) Get(ctx context.Context, task *repb.ExecutionTask) (*CommandRunne
 		return nil, err
 	}
 
-	// PermissionDenied and Unimplemented both imply that this is an
+	// PermissionDenied, Unauthenticated, Unimplemented both imply that this is an
```
bduffany, line 546: nit: s/both/all

app/auth/auth_service.ts
siggisim: The server currently (and I'd love to fix this at some point) returns http code `500` for all errors with just a text response that looks like this `rpc error: code = PermissionDenied desc = User not found`.

There's no status proto for us to parse, or http code for us to check. Changing this will require figuring out why the server isn't returning the right http codes (or returning status protos everywhere) and updating all of the places we parse errors to support both the current implementation and the new one.

Our error parser parses the current error format into an error code and an error description. 

To answer your actual question though, the error code looks like a string, but it's actually effectively an enum (the Typescript compiler will complain if it's an invalid value) thanks to this [union type](https://github.com/buildbuddy-io/buildbuddy/blob/master/app/util/errors.ts#L1) (@bduffany likes these, but I prefer enums).

bduffany (commented):
nit: consider updating `testauth` and `nullauth` if applicable?

app/auth/auth_service.ts
bduffany: I recently learned that enums in TypeScript can be used the same way as those string unions, and I agree they're easier to understand -- going to switch to enums from now on :)

enterprise/server/remote_execution/runner/runner.go
siggisim: Done

siggisim:
> nit: consider updating `testauth` and `nullauth` if applicable?

Done

server/util/capabilities/capabilities.go
siggisim: Yeah, we can have this discussion separately from this PR - I'd like to keep the scope small.

PR #889, opened 2021-08-16
Handle bytestream /Read requests where read_offset == digest size

siggisim:
The fix is the 3 line change in `byte_stream_server.go`, but this PR also adds tests for this scenario in:
- distributed_test.go
- cacheproxy_test.go
- disk_cache_test.go
- byte_stream_server_test.go

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/util/cacheproxy/cacheproxy_test.go
+++ b/enterprise/server/util/cacheproxy/cacheproxy_test.go
@@ -131,6 +131,56 @@ func TestReader(t *testing.T) {
 	}
 }
 
+func TestReaderMaxOffset(t *testing.T) {
+	ctx := context.Background()
+	flags.Set(t, "auth.enable_anonymous_usage", "true")
+	te := getTestEnv(t, emptyUserMap)
+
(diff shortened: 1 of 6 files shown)


enterprise/server/util/cacheproxy/cacheproxy_test.go
```diff
@@ -131,6 +131,56 @@ func TestReader(t *testing.T) {
...
+	io.CopyN(buf, randomSrc, 100)
+	readSeeker := bytes.NewReader(buf.Bytes())
+
+	// Compute a digest for the random bytes.
+	d, err := digest.Compute(readSeeker)
+	if err != nil {
+		t.Fatal(err)
+	}
+	readSeeker.Seek(0, 0)
+
```
tylerwilliams, line 164: nit: set instanceName := "foo" and use that variable below?

tylerwilliams (approved):
LGTM

enterprise/server/util/cacheproxy/cacheproxy_test.go
siggisim: Done

PR #890, opened 2021-08-17
Clean up target tracking

siggisim:
The main goal of this PR is to move status tracking from the TestSummary event to the Finished event to support invocations that use the `--test_tag_filters=` flag.

This PR also:
- Pulls permissions logic out into its own function so we have a better idea of which errors should be warnings and which should be debug
- Add more logging and include invocation ID for better debugging
- Cleans up switch statement by breaking the cases out into their own functions

---

**Version bump**: Minor

### DIFF
--- a/server/build_event_protocol/target_tracker/target_tracker.go
+++ b/server/build_event_protocol/target_tracker/target_tracker.go
@@ -163,13 +163,13 @@ func (t *TargetTracker) testTargetsInAtLeastState(state targetState) bool {
 	return true
 }
 
-func (t *TargetTracker) permissionsFromContext(ctx context.Context) *perms.UserGroupPerm {
+func (t *TargetTracker) permissionsFromContext(ctx context.Context) (*perms.UserGroupPerm, error) {
 	if auth := t.env.GetAuthenticator(); auth != nil {
 		if u, err := auth.AuthenticatedUser(ctx); err == nil && u.GetGroupID() != "" {
-			return perms.GroupAuthPermissions(u.GetGroupID())
+			return perms.GroupAuthPermissions(u.GetGroupID()), nil
 		}
 	}
-	return nil
+	return nil, status.UnauthenticatedError("Context did not contain auth information")
 }
 
 func (t *TargetTracker) writeTestTargets(ctx context.Context, permissions *perms.UserGroupPerm) error {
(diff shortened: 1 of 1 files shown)


server/build_event_protocol/target_tracker/target_tracker.go
```diff
@@ -209,29 +209,20 @@ func (t *TargetTracker) writeTestTargets(ctx context.Context) error {
 	}
 	if len(updatedTargets) > 0 {
 		if err := updateTargets(ctx, t.env, updatedTargets); err != nil {
-			log.Warningf("Error updating targets: %s", err.Error())
+			log.Warningf("Error updating %s targets: %s", t.buildEventAccumulator.InvocationID(), err.Error())
```
tylerwilliams, line 212: nit: you can use %q for the invocationID and it'll get quoted nicely

tylerwilliams: (here and below a couple places too)

server/build_event_protocol/target_tracker/target_tracker.go
```diff
@@ -154,7 +154,7 @@ func isTest(t *target) bool {
 	return strings.HasSuffix(strings.ToLower(t.ruleType), "test")
 }
 
-func (t *TargetTracker) testTargetsInAtleastState(state targetState) bool {
+func (t *TargetTracker) testTargetsInAtLeastState(state targetState) bool {
```
tylerwilliams, line 157: lol i had to read this like 10 times to see the delta. At least I finally got it.

tylerwilliams (approved):
LGTM

server/build_event_protocol/target_tracker/target_tracker.go
```diff
@@ -163,16 +163,16 @@ func (t *TargetTracker) testTargetsInAtleastState(state targetState) bool {
...
 	if auth := t.env.GetAuthenticator(); auth != nil {
 		if u, err := auth.AuthenticatedUser(ctx); err == nil && u.GetGroupID() != "" {
-			permissions = perms.GroupAuthPermissions(u.GetGroupID())
+			return perms.GroupAuthPermissions(u.GetGroupID())
 		}
 	}
-	if permissions == nil {
-		return status.FailedPreconditionError("Permissions were nil -- not writing target data.")
-	}
+	return nil
```
vadimberezniker, line 172: I'm not a big fan of returning nil pointers from helper functions to indicate error as it could easily lead to a future nil dereference. 
I would change the function to return (*perms.UserGroupPerm, error)

vadimberezniker (approved):
LGTM

server/build_event_protocol/target_tracker/target_tracker.go
siggisim: itsAsubtleOne

server/build_event_protocol/target_tracker/target_tracker.go
siggisim: Done

server/build_event_protocol/target_tracker/target_tracker.go
siggisim: Done

PR #892, opened 2021-08-17
Execution tab when RBE enabled

pariparajuli:
Tab for executions available, accessible page during dense mode

### DIFF
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -291,7 +291,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
             />
           )}
 
-          {isBazelInvocation && (activeTab == "all" || activeTab == "execution") && (
+          {isBazelInvocation && activeTab == "execution" && (
             <ExecutionCardComponent
               model={this.state.model}
               inProgress={this.state.inProgress}
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -223,10 +223,11 @@ export default class InvocationComponent extends React.Component<Props, State> {
             hash={this.props.hash}
             denseMode={this.props.preferences.denseModeEnabled}
             role={this.state.model.getRole()}
+            rbeEnabled={this.state.model.getIsRBEEnabled() ? true : false}
           />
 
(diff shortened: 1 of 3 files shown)


pariparajuli:
Screen Shot 2021-08-17 at 12 06 37 PM
Screen Shot 2021-08-17 at 12 14 33 PM

app/invocation/invocation.tsx
```diff
@@ -286,9 +291,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
             />
           )}
 
-          {activeTab == "timing" && <TimingCardComponent model={this.state.model} />}
-
-          {activeTab == "execution" && (
+          {isBazelInvocation && (activeTab == "all" || activeTab == "execution") && (
```
siggisim, line 294: I don't think we want the executions tab to show up on the all page

siggisim (changes requested):
Looks good, just a minor comment

app/invocation/invocation_execution_card.tsx
```diff
@@ -275,12 +274,8 @@ export default class ExecutionCardComponent extends React.Component {
 
     return (
       <div>
-        <InvocationFilterComponent
-          hash={"#execution"}
-          search={this.props.search}
-          placeholder={"Filter by digest or command..."}
-        />
         <div className={`card expanded`}>
+          <img className="icon" src="/image/cloud-regular.svg" />
```
siggisim, line 278: I'm not seeing this icon in the screenshots - where does this show up?

siggisim (changes requested):

app/invocation/invocation_execution_card.tsx
pariparajuli: Realized that the screenshots are from before i added the icon but this is what it looks like after 
Screen Shot 2021-08-19 at 4 20 36 PM

siggisim: Could we maybe just leave it out? It pushes all of the content to the right.

siggisim (approved):
Looks great!

PR #893, opened 2021-08-17
Don't redact environment variables we will later use

siggisim:
Also: 
- adds missing environment variables `COMMIT_SHA` and `REPO_URL`
- updates docs

---

**Version bump**: Minor


vadimberezniker (approved):
LGTM

PR #895, opened 2021-08-17
Don't show error alerts on invocations page for errors we already have pages for

siggisim:
Currently for invocations that we have error pages for, we also pop an errorService alert modal.

Example for not found: https://app.buildbuddy.io/invocation/dfldksfjlsfa

This PR fixes that.

---

**Version bump**: Minor


vadimberezniker (approved):
LGTM

PR #896, opened 2021-08-18
Don't use a full CPU for small tests

siggisim:
Use the default CPU estimate instead

---

**Version bump**: Minor


vadimberezniker (approved):
LGTM

PR #904, opened 2021-08-23
Fix logs not being displayed while invocation is in progress

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/813, Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/694

### DIFF
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -139,7 +139,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
     if (!this.state.model.hasChunkedEventLogs()) {
       return false;
     }
-    return this.logsModel.isFetching();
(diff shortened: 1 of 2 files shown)


siggisim:
What do you think about putting this here? https://github.com/buildbuddy-io/buildbuddy/blob/master/app/invocation/invocation.tsx#L138

It seems like from the component's perspective if "loading" is set to true, it should display the loading indicator.

bduffany:
> What do you think about putting this here? https://github.com/buildbuddy-io/buildbuddy/blob/master/app/invocation/invocation.tsx#L138
> 
> It seems like from the component's perspective if "loading" is set to true, it should display the loading indicator.

sg, done

siggisim (approved):

PR #905, opened 2021-08-23
Delete DB executor support.

vadimberezniker:
All environments are using Redis as of the last release.

---

**Version bump**: None


bduffany (approved):
Nice!

PR #907, opened 2021-08-24
Clean up optimized redaction flag usage

bduffany:
The flag is now `true` in prod.

---

**Version bump**: Patch


vadimberezniker (approved):
LGTM

PR #909, opened 2021-08-24
Simplify and fix bugs with history component

bduffany:
* Fix bug that we would sometimes call certain RPCs unnecessarily. For example, we would always fetch invocations when looking at aggregate pages, and wait for the invocations to come back before displaying the aggregate stats, even though the invocations aren't needed in order to show aggregate stats. Now we always fetch _either_ invocations and summary stats, _or_ aggregated stats.
* Fix bug that we'd fetch summary stats when scoped to a user, repo, etc. but not actually display the stats.
* Fix bug that navigating to the users view would show "No users found" while the users are being fetched (can repro this easily in prod by going to the Users tab in the probers org)
* Fix bug that stale RPC data could be displayed (now we cancel all RPCs properly).

* Commented, renamed, and updated types of some state fields to improve readability. (I kept getting confused between `summaryStat` and `invocationStat` -- I find the new names and types harder to confuse)

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/history/history.tsx
+++ b/enterprise/app/history/history.tsx
@@ -84,7 +84,7 @@ export default class HistoryComponent extends React.Component<Props, State> {
       }),
       pageToken: nextPage ? this.state.pageToken : "",
       // TODO(siggisim): This gives us 2 nice rows of 63 blocks each. Handle this better.
-      count: 10,
+      count: 126,
     });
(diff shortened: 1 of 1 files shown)


siggisim (approved):
LGTM - talked offline about moving from rxjs to cancellable promises in a separate PR.

PR #934, opened 2021-08-27
CASFS integration (except Firecracker) (behind flag & platform property).

vadimberezniker:
This PR only implements the basic `Run` path, the other path will be
handle via a separate PR.

This PR changes the container `Run` and `Exec` methods to take a
`repb.ExecutionTask` instead of a `repb.Command` as we need access to
both the remote instance name and task ID for the CASFS.

---

**Version bump**: None


enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -83,6 +91,34 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
...
+			return result
+		}
+		defer func() {
+			err := cfs.Unmount()
+			if err != nil {
+				log.Warningf("CASFS unmount failed: %s", err)
+			}
+		}()
+		containerRootDir = casfsDir
+	}
```
bduffany, line 120: Am I understanding this right -- the only change to the container impl is that we're just swapping out the implementation of the `workingDir` from a regular directory to a FUSE-based one?

If so it seems like this logic should done at the Runner level (rather than the container level) and the container implementations don't need to care about how the `workingDir` is implemented -- otherwise you have to do this same logic for every container impl.

bduffany (changes requested):

enterprise/server/remote_execution/containers/docker/docker.go
vadimberezniker: For Firecracker, the FUSE filesystem needs to be mounted within the Guest VM so a generalized approach won't work for all container impls.

bduffany: I think maybe it makes sense then to do something like this in `runner.go`:

```
if fuseEnabled && !firecracker {
  workspaceDir = createFUSEWorkspace(ctx, env, workspaceDir, ...)
}
```

Because other container impls (bare, containerd, podman etc.) would all use the same logic.

Trying to avoid adding business logic to the container implementations as much as possible, since it's simpler to understand and gives us more flexibility if the container interface maps more or less directly to the equivalent CLI command.

vadimberezniker: Done.

enterprise/server/remote_execution/container/container.go
```diff
@@ -50,7 +50,7 @@ type CommandContainer interface {
 	//
 	// It is approximately the same as calling PullImageIfNecessary, Create,
 	// Exec, then Remove.
-	Run(ctx context.Context, command *repb.Command, workingDir string, creds PullCredentials) *interfaces.CommandResult
+	Run(ctx context.Context, task *repb.ExecutionTask, workingDir string, creds PullCredentials, fsLayout *FileSystemLayout) *interfaces.CommandResult
```
bduffany, line 53: I guess `task` and `fsLayout` are still being passed in because firecracker will need them -- but in the interest of simplifying the `CommandContainer` interface, I think it would be a bit cleaner to also do the firecracker-specific FS stuff inside `PrepareForTask` in runner.go as well. Kinda like

```
if ctr, ok := r.CommandContainer.(*FireCrackerContainer); ok {
  ctr.SetupCASFSInGuestVM(task, fsLayout)
} else {
  workspaceDir = makeCASFS()
}
```

vadimberezniker: Done.

bduffany (approved):

PR #944, opened 2021-08-30
Redact container registry platform props

bduffany:
We don't support these props just yet, but preemptively redacting them before we do.

---

**Version bump**: Patch

### DIFF
--- a/server/util/redact/BUILD
+++ b/server/util/redact/BUILD
@@ -9,11 +9,8 @@ go_library(
         "//proto:build_event_stream_go_proto",
         "//proto:command_line_go_proto",
         "//server/environment",
-        "//server/util/alert",
         "//server/util/git",
-        "//server/util/log",
         "@com_github_golang_protobuf//proto:go_default_library",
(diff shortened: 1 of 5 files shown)


server/util/redact/BUILD
```diff
@@ -9,8 +9,11 @@ go_library(
         "//proto:build_event_stream_go_proto",
         "//proto:command_line_go_proto",
         "//server/environment",
+        "//server/util/alert",
         "//server/util/git",
+        "//server/util/log",
         "@com_github_golang_protobuf//proto:go_default_library",
+        "@com_github_mattn_go_shellwords//:go-shellwords",
```
siggisim, line 16: we probably don't need any of these new imports anymore

siggisim (approved):
Looks good, just one comment

server/util/redact/BUILD
bduffany: Fixed -- checkstyle noticed this too :)

vadimberezniker (approved):

PR #947, opened 2021-08-31
Implement private registry auth via platform props

bduffany:
Allows using `container-registry-username` and `container-registry-password` as a source of image pull credentials, and wires up the `ImageCacheAuthenticator` to `PullImageIfNecessary` so that we can avoid re-checking these credentials on every action.

Pull credentials now come from the following sources (in order):
1. Platform props
2. Executor config (`executor.container_registries` section)
3. Credential helpers (`~/.docker/config.json`)

Depends on https://github.com/buildbuddy-io/buildbuddy/pull/944 so that we redact properly.

Tested:
* Was able to run a build with `--remote_default_exec_properties=container-registry-username=AWS --remote_default_exec_properties=container-registry-password=$(aws ecr get-login-password --region us-east-2)` and pull from a private ECR registry
* If `container-registry-password` is removed, the error `Unavailable: Error pulling container: Error response from daemon: Head <URL>: no basic auth credentials` is returned back from AWS, as expected. After adding the creds back in, no longer get the error. This shows that we create new tokens for different credential values.
* If I delete the container image, it is indeed re-pulled, showing that we check `IsImageCached` properly and re-pull if needed.

---

**Version bump**: Patch


siggisim (approved):
Looks good to me, but @vadimberezniker should probably take a peak at the go details.

vadimberezniker (approved):
Looks great.

PR #948, opened 2021-08-31
Add unexpected events chart

bduffany:
Good news: no unexpected events in dev :)

Also fix grafana startup URL so it takes us to the editable view by default.

Query is

```promql
sum by (name) (rate(buildbuddy_unexpected_event[${window}]))
```

---

**Version bump**: Patch

### DIFF
--- a/tools/metrics/grafana/dashboards/buildbuddy.json
+++ b/tools/metrics/grafana/dashboards/buildbuddy.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": 2,
-  "iteration": 1630346094925,
+  "iteration": 1630423834333,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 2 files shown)


vadimberezniker:
> Good news: no unexpected events in dev :)

There's an alert setup for unexpected events so not surprised there aren't any :)

tools/metrics/grafana/dashboards/buildbuddy.json
```diff
@@ -10367,13 +10367,111 @@
...
+            "alignLevel": null
+          }
         }
       ],
       "title": "Internal",
       "type": "row"
     }
   ],
-  "refresh": "1m",
+  "refresh": "5s",
```
vadimberezniker, line 10474: Did you mean to change this?

vadimberezniker (approved):

tools/metrics/grafana/dashboards/buildbuddy.json
bduffany: Nope, good catch -- reverted

PR #950, opened 2021-08-31
Remove default_override flags

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/server/util/redact/redact.go
+++ b/server/util/redact/redact.go
@@ -2,6 +2,7 @@ package redact
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


server/util/redact/redact.go
```diff
@@ -130,10 +131,12 @@ func redactStructuredCommandLine(commandLine *clpb.CommandLine, allowedEnvVars [
...
 			option.CombinedForm = stripURLSecrets(option.CombinedForm)
 
-			// Redact remote header values
-			if option.OptionName == "remote_header" || option.OptionName == "remote_cache_header" {
-				option.OptionValue = envVarRedactedPlaceholder
-				option.CombinedForm = envVarPrefix + option.OptionName + envVarSeparator + envVarRedactedPlaceholder
+			// Redact remote header values. Also redact default_overrides for now
+			// since we don't have a use for them (yet) and they may contain sensitive
+			// info.
+			if option.OptionName == "remote_header" || option.OptionName == "remote_cache_header" || option.OptionName == "default_override" {
```
siggisim, line 137: I think we should drop these altogether - we see a ton of these. Basically one per line in your .bazelrc

bduffany: Done

siggisim (approved):

PR #952, opened 2021-08-31
FUSE lib hack to avoid accessing /dev/null before it's available.

vadimberezniker:
We delay the splice package init() call until the FUSE server is
instantiated.

---

**Version bump**: None


bduffany (approved):

bduffany:
Maybe add a TODO to the `goinit` binary to run this as a binary to avoid needing to patch?

vadimberezniker:
> Maybe add a TODO to the `goinit` binary to run this as its own binary to avoid needing to patch?

Done.

PR #954, opened 2021-08-31
Add docs for private registry authentication platform props

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/docs/rbe-platforms.md
+++ b/docs/rbe-platforms.md
@@ -48,20 +48,23 @@ flags to your build:
 
 ```
 --remote_default_exec_properties=container-registry-username=USERNAME
---remote_default_exec_properties=container-registry-password=PASSWORD
+--remote_default_exec_properties=container-registry-password=ACCESS_TOKEN
 ```
 
-For the password, we recommend generating a short lived token using
-the appropriate utility for your cloud provider.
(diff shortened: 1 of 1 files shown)


docs/rbe-platforms.md
```diff
@@ -41,11 +41,41 @@ You can then pass this configuration to BuildBuddy RBE with the following flag:
...
+
+For Amazon ECR (Elastic Container Registry), the username must be `AWS`
+and a token can be generated with `aws ecr get-login-password --region REGION`
+(replace `REGION` with the region matching the ECR image URL):
+
+```
+--remote_default_exec_properties=container-registry-username=AWS
+--remote_default_exec_properties=container-registry-password="$(aws ecr get-login-password --region REGION)"
+```
+
```
siggisim, line 73: This is great, could you add an example using a custom platform & GCR: https://docs.buildbuddy.io/docs/rbe-platforms#using-a-custom-docker-image

I think this would likely have to use a long lived token (unless you get fancy and generate the platform on the fly - which we can do later in our toolchain).

bduffany: Done

siggisim (approved):
Awesome!

PR #955, opened 2021-09-01
Fix image caching related bug in Firecracker impl.

vadimberezniker:
The image path is only set in `PullImage` which is not called if
`IsImageCached` returns true.

Since this function can have side-effects perhaps it could use a better
name like `CheckCachedImage` or `FindCachedImage`?

---

**Version bump**: None


bduffany (approved):
Thanks for catching this! And sorry if it caused any inconvenience.

I think maybe we could make `containerFSPath` a function and then maybe cache its return value if we are concerned about performance.

vadimberezniker:
> Thanks for catching this! And sorry if it caused any inconvenience.
> 
> I think maybe we could make `containerFSPath` a function and then maybe cache its return value if we are concerned about performance.

I don't think we need to worry about performance here. The cache check is likely to be dwarfed by the other work that we do for each execution.

PR #965, opened 2021-09-01
Guest side changes for Firecracker CASFS support.

vadimberezniker:
Setup CASFS if the execution request contains filesystem layout
information.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -459,11 +459,14 @@ type FileMap map[digest.Key][]*FilePointer
 type BatchFileFetcher struct {
 	ctx          context.Context
 	instanceName string
-	fileCache    interfaces.FileCache
+	fileCache    interfaces.FileCache // OPTIONAL
 	bsClient     bspb.ByteStreamClient
-	casClient    repb.ContentAddressableStorageClient
+	casClient    repb.ContentAddressableStorageClient // OPTIONAL
 }
 
+// NewBatchFileFetcher creates a CAS fetcher that can automatically batch small requests and stream large files.
+// `fileCache` is optional. If present, it's used to cache a copy of the data for use by future reads.
+// `casClient` is optional. If not specified, all requests will use the ByteStream API.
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -475,6 +478,10 @@ func NewBatchFileFetcher(ctx context.Context, instanceName string, fileCache int
 }
 
 func (ff *BatchFileFetcher) batchDownloadFiles(ctx context.Context, req *repb.BatchReadBlobsRequest, filesToFetch FileMap, opts *DownloadTreeOpts) error {
+	if ff.casClient == nil {
+		return status.FailedPreconditionErrorf("cannot batch download files when casClient is noet set")
```
bduffany, line 482: uber nit: typo

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -558,7 +565,7 @@ func (ff *BatchFileFetcher) FetchFiles(filesToFetch FileMap, opts *DownloadTreeO
 		// fit in the batch call, so we'll have to bytestream
 		// it.
 		size := d.GetSizeBytes()
-		if size > gRPCMaxSize {
+		if size > gRPCMaxSize || ff.casClient == nil {
```
bduffany, line 568: Is there a reason to disable batching for this use case? If not, maybe just pass a CAS client to the BatchFileFetcher to avoid adding more conditional logic

bduffany (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
vadimberezniker: Done.

enterprise/server/remote_execution/dirtools/dirtools.go
vadimberezniker: In v0 for firecracker, I only implemented a ByteStream proxy because 1) reads from the guest generally can't be batched and 2) for uploads, it's not clear how much perf benefit there is to using the CAS API between the guest and the host. 

Once v0 is working I will test the upload perf to determine if we should also add a CAS API proxy.

PR #966, opened 2021-09-01
Update `vmstart` tool to optionally mount the FS for a given action

vadimberezniker:
using CASFS.

---

**Version bump**: None


tools/vmstart/vmstart.go
```diff
@@ -30,6 +37,8 @@ var (
 	remoteInstanceName = flag.String("remote_instance_name", "", "The remote_instance_name for caching snapshots")
```
bduffany, line 37: nit: update the docs to mention that it's for caching snapshots and also for fetching the action, if applicable

bduffany (approved):

tools/vmstart/vmstart.go
vadimberezniker: Done.

PR #967, opened 2021-09-02
Support short-lived container registry credentials via remote_header

bduffany:
Confirmed that passing a short-lived token via platform props causes the cache to be invalidated on every build which is obviously undesirable. Confirmed that with the remote header approach, the cache is not invalidated on every build.

---

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/491


siggisim (commented):
This looks pretty clean to me, and seems like a good way to support short-lived tokens.

I don't think we need to support multiple credentials to start.

tylerwilliams (approved):
LGTM

PR #968, opened 2021-09-02
Implement usage page RPC and UI view

bduffany:
* Allow people to view up to most recent 6 months of usage data
* Allow configuring a `start_date` for usage data -- we'll set this to `2021-10-01T00:00:00Z` so that folks can view their usage data starting from October 1st but no earlier (since we won't have any data collected before then)
* For simplicity, return the data in reverse-chronological order and guarantee that at least one month is returned, so the UI can just render the months in a dropdown picker directly, and set the current selection to the first (latest) month

image

---

**Version bump**: Patch


enterprise/server/usage_service/usage_service.go
```diff
@@ -17,5 +26,87 @@ func New(env environment.Env) *usageService {
...
+	// Get the timestamp corresponding to the start of the current month in UTC,
+	// with a month offset so that we return the desired number of months of
+	// usage.
+	now := time.Now().UTC()
+	maxNumHistoricalMonths := maxNumMonthsOfUsageToReturn - 1
+	createdAfter := addCalendarMonths(startOfMonth(now), -maxNumHistoricalMonths)
+	createdBefore := addCalendarMonths(startOfMonth(now), 1)
+
+	rows, err := db.Raw(`
+	SELECT `+db.UTCMonthFromUsecTimestamp("created_at_usec")+` AS period,
```
vadimberezniker, line 45: nit: indent

vadimberezniker (approved):
:notes: so fresh, so clean :notes:

bduffany:
Closing for now -- we want to have a separate Usage table to keep track of usage and it's probably not worth trying to migrate from the invocation-based page to the usage-table based page.

Will reopen once we implement the Usage table, and use the Usage table data for the SQL query.

bduffany:
@vadimberezniker -- modified the SQL queries in `usage_service.go` to pull from the new `Usage` table instead of `Invocation` -- PTAL

enterprise/app/usage/usage.tsx
```diff
@@ -19,21 +25,64 @@ export default class UsageComponent extends React.Component<UsageProps, State> {
     this.setState({ loading: true });
     rpcService.service
       .getUsage(new usage.GetUsageRequest())
-      .then((response) => this.setState({ response }))
+      .then((response) => {
+        this.setState({ response, selectedPeriod: response.usage[0].period });
```
siggisim, line 29: any chance we could get a could not find `.period` of undefined

siggisim (approved):
front-end changes look good - just one nit

enterprise/app/usage/usage.tsx
bduffany: That can happen if the configured usage start date is in the future. Updated the server code so that it just returns an empty month of usage for the current month in that case -- and also logs an error on server startup about the potential misconfiguration.

siggisim: might also be worth throwing a `?.period` in there to be safe in case the server ever changes

bduffany: Throwing an error in that case so that we only set the response if it's valid (and don't have to worry about edge cases anywhere in the rendering logic).

vadimberezniker (approved):

PR #969, opened 2021-09-02
Add `clean-workspace-inputs` platform argument that allows users to specify how workspace inputs are cleaned up when preserving workspaces

siggisim:
Add `clean-workspace-inputs` platform argument that allows users to specify how workspace inputs are cleaned up when preserving workspaces.

- If unset or empty, the current behavior will be preserved where we only overwrite inputs that have changed between actions.
- If set to `*`, all old inputs that are not used in the current action are cleaned up.
- If set to another value, we split on commas and clean up any inputs glob matching those strings
  - For example  `clean-workspace-inputs: **.framework/**,**.h ` will clean up any old inputs that end in `.h` or contain `.framework/`.

This fixes: https://github.com/buildbuddy-io/buildbuddy-internal/issues/852

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/workspace/workspace.go
+++ b/enterprise/server/remote_execution/workspace/workspace.go
@@ -3,6 +3,7 @@ package workspace
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -17,6 +18,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
+	"github.com/gobwas/glob"
 	"github.com/google/uuid"
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/workspace/workspace.go
```diff
@@ -139,6 +144,38 @@ func (ws *Workspace) DownloadInputs(ctx context.Context, tree *repb.Tree) (*dirt
...
+			inputFilesToCleanUp[path] = digest
+			continue
+		}
+		patterns := strings.Split(ws.opts.CleanInputs, ",")
+		for _, pattern := range patterns {
+			if strings.Contains(path, pattern) {
+				inputFilesToCleanUp[path] = digest
+				continue
+			}
+		}
```
bduffany, line 163: I think we should use `filepath.Match`on the basename of the path (`filepath.Base`), instead of supporting a mix of `strings.Contains` and an explicit "match-all" pattern here:

* It supports single-star globs already, so there's no need to treat `*` specially (which might mislead people into thinking that other globs will work as well).
* It's less error-prone since it matches the whole string. For example `contains(".h")` will also match `foo.hs` while `match("*.h")` will only match `.h` files.
* It's backwards-compatible with full glob expressions (i.e. globs including `**`, like `include/**/*.h`), if we want to support that later on

Glob patterns would need an explicit `*`, like `*.h,*.framework`, and the impl would have to work on the basename and not the full path. https://play.golang.org/p/FiX_WYiS-c1

bduffany (changes requested):

enterprise/server/remote_execution/workspace/workspace_test.go
```diff
@@ -139,3 +139,93 @@ func TestWorkspaceCleanup_PreserveWorkspace_PreservesAllFilesExceptOutputs(t *te
...
+	for _, file := range filePaths {
+		ws.Inputs[file] = &repb.Digest{}
+	}
+
+	keep := map[string]*repb.Digest{
+		"KEEPME":         &repb.Digest{},
+		"foo/KEEPME":     &repb.Digest{},
+		"foo/bar/KEEPME": &repb.Digest{}}
+
+	ws.CleanInputsIfNecessary(keep)
```
bduffany, line 165: nit: assert that there's no error:

```
err := ws.CleanInputsIfNecessary(keey)

require.NoError(t, err)
```

(same in the test below)

enterprise/server/remote_execution/workspace/workspace.go
bduffany: Resolved offline - `filepath.Match` + `filepath.Base` won't work if the patterns are trying to match a directory rather than a basename. We can later use a separate platform prop like `clean-workspace-input-patterns` if we want full glob support

bduffany (approved):

enterprise/server/remote_execution/workspace/workspace_test.go
siggisim: Done

enterprise/server/remote_execution/workspace/workspace.go
siggisim: Ended up switching to a glob that supports ** since that's what my original intent was - and we're already pulling in a library that supports this.

bduffany (approved):

PR #974, opened 2021-09-05
Implement a real Xcode locator

siggisim:
This removes the need to have hardcoded XCode paths in our code (and for folks running executors to place XCode at those paths), and also allows us to support XCode versions more precise than two components (i.e. 12.5.1 or 12.5.1.12E507).

Xcode location logic is losely based on Bazel's XCode locator: https://github.com/bazelbuild/bazel/blob/master/tools/osx/xcode_locator.m

This removes the need for enableXcodeOverride.

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -115,9 +115,7 @@ func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChec
 		log.Infof("No authentication will be configured: %s", err)
 	}
 
-	xcodeLocator := xcode.NewXcodeLocator()
-	xcodeLocator.Locate()
-	realEnv.SetXCodeLocator(xcodeLocator)
+	realEnv.SetXCodeLocator(xcode.NewXcodeLocator())
 
 	if gcsCacheConfig := configurator.GetCacheGCSConfig(); gcsCacheConfig != nil {
 		opts := make([]option.ClientOption, 0)
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -33,6 +33,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/monitoring"
 	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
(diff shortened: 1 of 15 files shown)


enterprise/server/cmd/executor/executor.go
```diff
@@ -114,6 +115,10 @@ func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChec
 		log.Infof("No authentication will be configured: %s", err)
 	}
 
+	xcodeLocator := xcode.NewXcodeLocator()
+	xcodeLocator.Locate()
```
tylerwilliams, line 119: I think you could have "NewXcodeLocator()" call locate() internally before returning, then remove the "Locate()" method from the interface, and from being a public method

siggisim: Done. Also added support for SDKs, dropped enableXcodeOverride, and simplified the CFString stuff (I think all of that craziness was only needed when allocating a new string - which we don't do).

server/xcode/xcode_darwin.go
```diff
@@ -0,0 +1,171 @@
...
+
+		versions := expandXCodeVersions(xcodePlist.CFBundleShortVersionString, xcodePlist.ProductBuildVersion)
+		mostPreciseVersion := versions[len(versions)-1]
+		for _, version := range versions {
+			existingXcode, ok := versionMap[version]
+			if ok && mostPreciseVersion < existingXcode.version {
+				continue
+			}
+			log.Infof("Found XCode version: %s=>%s", version, developerDirPath)
+			log.Debugf("With SDKs %+v", sdks)
```
tylerwilliams, line 130: can you consolidate this into the line above (or rm it)?

tylerwilliams (approved):
LGTM (with one logging nit)

server/xcode/xcode_darwin.go
siggisim: Done

enterprise/server/cmd/executor/executor.go
```diff
@@ -114,6 +115,8 @@ func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChec
 		log.Infof("No authentication will be configured: %s", err)
 	}
 
+	realEnv.SetXCodeLocator(xcode.NewXcodeLocator())
```
brentleyjones, line 118: nit: It's `Xcode`, not `XCode`

PR #977, opened 2021-09-07
Fix cache perf test.

vadimberezniker:
Distributed disk cache requires isolation params to be set.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #978, opened 2021-09-08
Start a separate guest binary to handle CASFS setup.

vadimberezniker:
Still unclear why but having the fuse server in the same process as init
appears to cause some kind of deadlock when executing a command from the
vmexec service.

---

**Version bump**: None


enterprise/server/util/vsock/vsock.go
```diff
@@ -29,7 +29,9 @@ const (
 	minCID = 3
 	maxCID = math.MaxUint32
 
-	DefaultPort = 25415
+	VMExecPort              = 25415
+	VMCASFSPort             = 25416
+	HostByteStreamProxyPort = 25410
```
tylerwilliams, line 34: can you comment what each of these are?

enterprise/server/vmcasfs/vmcasfs.go
```diff
@@ -0,0 +1,109 @@
...
+	"google.golang.org/grpc"
+
+	vmfspb "github.com/buildbuddy-io/buildbuddy/proto/vmcasfs"
+	libVsock "github.com/mdlayher/vsock"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
+)
+
+var (
+	port = flag.Uint("port", vsock.VMCASFSPort, "The vsock port number to listen on")
+)
```
tylerwilliams, line 25: does this need to be a flag?

enterprise/server/vmcasfs/vmcasfs.go
```diff
@@ -0,0 +1,109 @@
...
+	// Context & cancel func for CAS RPS.
+	remoteRPCCtx       context.Context
+	cancelRemoteRPCCtx context.CancelFunc
+}
+
+func NewServer() (*casfsServer, error) {
+	if err := os.Mkdir("/casfs", 0755); err != nil {
+		return nil, err
+	}
+	cfs := casfs.New("/workspace", "/casfs/", &casfs.Options{})
```
tylerwilliams, line 41: do these paths have to match what is in goinit/main.go and firecracker/firecracker.go? Any ideas on how we could consolidate them? (maybe pass flags to init and to this func and define the consts in firecracker.go?)

enterprise/server/vmcasfs/vmcasfs.go
```diff
@@ -0,0 +1,109 @@
...
+	return &vmfspb.PrepareResponse{}, nil
+}
+
+func (s *casfsServer) Finish(ctx context.Context, request *vmfspb.FinishRequest) (*vmfspb.FinishResponse, error) {
+	s.mu.Lock()
+	if s.cancelRemoteRPCCtx != nil {
+		s.cancelRemoteRPCCtx()
+	}
+	s.mu.Unlock()
+	// TODO(vadim): implement
```
tylerwilliams, line 87: What do we need to do here?

proto/vmcasfs.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
+
+package vmcasfs;
+
+import "proto/remote_execution.proto";
+
```
tylerwilliams, line 6: nit: newlines after field definition, comment what messages are for if it's not obvious

// This message does foo.
message Foo {
  // comment 1
  string field = 1;

  // comment 2
  string field = 2;
}

proto/vmcasfs.proto
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// FileSystem is a service for controlling the FUSE CASFS filesystem inside a
+// VM.
+service FileSystem {
+  // Prepare the filesystem for use.
+  rpc Prepare(PrepareRequest) returns (PrepareResponse);
+  // Finish any outstanding work (such as uploading data to the CAS).
+  rpc Finish(FinishRequest) returns (FinishResponse);
+}
```
tylerwilliams, line 45: add a newline at the end of this file

enterprise/server/vmcasfs/vmcasfs.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (s *casfsServer) Prepare(ctx context.Context, req *vmfspb.PrepareRequest) (*vmfspb.PrepareResponse, error) {
+	s.mu.Lock()
+	if s.cancelRemoteRPCCtx != nil {
+		s.cancelRemoteRPCCtx()
+	}
+	rpcCtx, cancel := context.WithCancel(context.Background())
+	s.remoteRPCCtx = rpcCtx
+	s.cancelRemoteRPCCtx = cancel
```
tylerwilliams, line 69: I don't really understand the context stuff going on here -- could you add a comment explaining why a background context is used/saved in Prepare?

tylerwilliams, line 69: Nit: might simplify these names to call them:

s.ctx
s.cancelFunc

or s.remoteCtx
and s.remoteCancelFunc

tylerwilliams (changes requested):
Mostly LG, few nits and questions

proto/vmcasfs.proto
vadimberezniker: Done.

vadimberezniker (commented):
For some reason GitHub won't let me reply directly to your comments, so I'll reply to two of them here:

> do these paths have to match what is in goinit/main.go and firecracker/firecracker.go? Any ideas on how we could consolidate > them? (maybe pass flags to init and to this func and define the consts in firecracker.go?)

That's a good approach, but I just tried it and it doesn't work properly. It looks like any init args with a "=" are swallowed by the kernel. "--" works to separate the kernel & init args but firecracker SDK butchers the args order so it's not possible to do this. 
I'll leave the "/casfs" constant in the files for now. The "/workspace" references will go away with my latest changes.

> What do we need to do here?

With my newest changes, we won't need any more changes in Finish()

tylerwilliams (approved):
LGTM

PR #979, opened 2021-09-08
Bump scheduler LeaseTask stream EOF to a Warning.

vadimberezniker:
We normally do not expect to reach the stream EOF if the leasing process
is working properly.

---

**Version bump**: None


siggisim (approved):

tylerwilliams (approved):
LGTM

( a lot of these task lease messages are really noisy! are there others you want to clean up too? )

vadimberezniker:
> LGTM
> 
> ( a lot of these task lease messages are really noisy! are there others you want to clean up too? )

We could get rid of the LeaseTask `ping` logs. I think they are less useful now that things are pretty solid. 

BTW, we disabled debug logs in prod while you were away as the log volume (along with cost) went up quite a bit with more customer traffic.

PR #990, opened 2021-09-10
Add initial usage tracker with Redis recording only

bduffany:
* Add `UsageTracker` interface and initial impl that stores metrics for the current collection period in Redis
* Wire up `UsageTracker` to BES handler and `HitTracker`

This PR is effectively a NOP because we don't provision a UsageTracker unless `usage_tracking_enabled` is `true`.

This only adds Redis tracking -- DB storage / synchronization will come in a future PR.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/796


server/tables/tables.go
```diff
@@ -427,6 +427,12 @@ func (wf *Workflow) TableName() string {
 	return "Workflows"
 }
 
+type UsageCounts struct {
```
bduffany, line 430: FYI this lives in `tables.go` because it will be embedded into the `Usage` table.

enterprise/server/cmd/server/main.go
```diff
@@ -181,6 +183,21 @@ func main() {
 		realEnv.SetCacheRedisClient(redisClient)
 	}
 
+	if redisTarget := configurator.GetUsageRedisTarget(); redisTarget != "" {
+		redisClient := redisutil.NewClientWithOpts(&redis.Options{
+			Addr: redisTarget,
+			// Use db1 for usage data since we use SCAN to iterate over all usage
+			// keys and scanning over unrelated keys would hurt performance.
+			DB: 1,
```
tylerwilliams, line 191: I don't love introducing yet another redis client (YARC?), and also the notion of scanning the entire key space.

How about using the normal redis client, and keeping track of your "usage counters" by adding them to a set every time you increment them with SADD.

So, at the end of the minute (HHMM) your keyspace would look like this:

HHMM: set(GR123-HHMM, GR345-HHMM, GR567-HHMM)
GR123-HHMM: {cas_hits: 1, ac_hits: 2}
GR345-HHMM: {cas_hits: 3, ac_hits: 4}
GR567-HHMM: {cas_hits: 5, ac_hits: 6}
... other keys you don't care about...


Your set flow would be:

Pipeline:
  SADD(HHMM, GR123-HHMM)
  INCR(GR123-HHMM-cas_hits, 1)
  EXPIRE(GR123-HHMM-cas_hits, 120)

And your read flow would be:

Get global lock()
for period in HHMM, HHMM-1, HHMM-2 {
  for counter in period:
    flushToMySQL(counter)
  DEL period

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,151 @@
...
+
+func countsRedisKey(groupID string, t time.Time) string {
+	return fmt.Sprintf("%s%s/%d", redisCountsKeyPrefix, groupID, collectionPeriodStartUsec(t))
+}
+
+func invocationsRedisKey(groupID string, t time.Time) string {
+	return fmt.Sprintf("%s%s/%d", redisInvocationsKeyPrefix, groupID, collectionPeriodStartUsec(t))
+}
+
+func authenticatedGroupID(ctx context.Context, env environment.Env) (string, error) {
```
tylerwilliams, line 116: i would bet money we have a function somewhere that does this already. It's impossible we haven't already written every version of this function :P  can you check and or put this in a common package?

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,151 @@
...
+	"fmt"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/buildbuddy-io/buildbuddy/server/util/timeutil"
+	"github.com/go-redis/redis"
+	"github.com/jonboulle/clockwork"
```
tylerwilliams, line 15: We should be really careful about introducing new 3p packages -- we already depend on too many. Each one increases our security and update burden, so we should really only use stuff we *need*.

Can you use a timer here instead?

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,151 @@
...
+		return err
+	}
+	if len(m) == 0 {
+		return nil
+	}
+
+	uk := countsRedisKey(groupID, ut.clock.Now())
+	pipe := ut.rdb.TxPipeline()
+	for k, c := range m {
+		pipe.HIncrBy(ctx, uk, k, c)
```
tylerwilliams, line 71: I think we should expire these keys after 2-3x the flush period. It limits what we have to potentially scan (which is nice), and keeps us from potentially polluting redis with values we never read.

If we haven't picked them up in a couple of flush periods, we're not going to. (Also, in that case, would we even be logging new data? Seems like something bigger is wrong)

server/tables/tables.go
tylerwilliams, line 430: Should this table be registered and have a name func?

tylerwilliams (changes requested):
looks like this CL is still a WIP but I left a couple high level comments about the approach (mostly the SCAN stuff). lmk if you want to chat about it on slack!

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -341,6 +341,12 @@ func (e *EventChannel) handleEvent(event *pepb.PublishBuildToolEventStreamReques
 					}
 					return status.UnknownError(fmt.Sprintf("%v", authError))
 				}
+				// Record invocation usage as soon as we know the authenticated user.
```
tylerwilliams, line 344: is this happening once per event? Is that what you want?

server/remote_cache/hit_tracker/hit_tracker.go
```diff
@@ -255,6 +261,36 @@ func (h *HitTracker) TrackUpload(d *repb.Digest) *transferTimer {
 	}
 }
 
+func (h *HitTracker) recordUsage(actionCache bool, d *repb.Digest, actionCounter counterType) error {
+	if h.usage == nil {
+		return nil
+	}
+
+	// If we have an invocation ID, record that the invocation occurred.
+	if h.iid != "" {
```
tylerwilliams, line 270: kinda confused why ObserveInvocation is called here as well as inside the invocation handling code, do we need both? (ideally, rm this one)

server/tables/tables.go
tylerwilliams: ok, sgtm

tylerwilliams (changes requested):

server/remote_cache/hit_tracker/hit_tracker.go
bduffany: Removed (decided that we won't track invocation usage for cache-only invocations)

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,138 @@
...
+	// Start of minute
+	start := time.Date(
+		utc.Year(), utc.Month(), utc.Day(),
+		utc.Hour(), utc.Minute(), 0, 0,
+		utc.Location())
+	return timeutil.ToUsec(start)
+}
+
+func groupsRedisKey(t time.Time) string {
+	return fmt.Sprintf("%s%d", redisGroupsKeyPrefix, collectionPeriodStartUsec(t))
```
vadimberezniker, line 116: Can we use a more human friendly format for the timestamps in the keys? 
If we ever need to debug an issue using the data in Redis, it'll be a little annoying to have to manually convert/interpret every timestamp.

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func countsRedisKey(groupID string, t time.Time) string {
+	return fmt.Sprintf("%s%s/%d", redisCountsKeyPrefix, groupID, collectionPeriodStartUsec(t))
+}
+
+func countsToMap(tu *tables.UsageCounts) (map[string]int64, error) {
+	b, err := json.Marshal(tu)
```
vadimberezniker, line 124: What do you think about explicitly mapping the fields instead of relying on this json back & forth conversion?
Renaming a field in the struct will cause problems for usage tracking & there will be no direct code references to individual fields to suggest it's a bad idea.

server/util/perms/perms.go
```diff
@@ -255,3 +255,31 @@ func AuthenticateSelectedGroupID(ctx context.Context, env environment.Env, proto
...
+
+// AuthenticatedGroupID returns the authenticated group ID from the given
+// context. This is preferred for API requests, since the group ID can be
+// determined directly from the API key. UI requests should instead use
+// `AuthenticateSelectedGroupID`, since the API key is not available, and the
+// user's selected group ID needs to be taken into account.
+//
+// If there is no authenticated user and anonymous usage is enabled, this will
+// return the empty string for the user ID, and a nil error.
+func AuthenticatedGroupID(ctx context.Context, env environment.Env) (string, error) {
```
vadimberezniker, line 267: Should we keep this consistent with `AuthenticatedUser` and have it return an error for anonymous users? 
I understand that it makes the API slightly easier to use, but I'm not a fan of the fact that it behaves differently from the other functions in this package.

enterprise/server/usage/usage.go
bduffany: Sounds good, made the names `underscore_case` to simplify the DB field mapping in the next PR

server/util/perms/perms.go
bduffany: Yeah, it's probably better to have them behave consistently. Done

server/tables/tables.go
```diff
@@ -427,6 +427,13 @@ func (wf *Workflow) TableName() string {
...
+	Invocations            int64
+	CasCacheHits           int64
```
vadimberezniker, line 432: nit: CASCacheHits

vadimberezniker (approved):

server/tables/tables.go
bduffany: Done

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,138 @@
...
+	groupID, err := perms.AuthenticatedGroupID(ctx, ut.env)
+	if err != nil {
+		if perms.IsAnonymousUserError(err) && ut.env.GetConfigurator().GetAnonymousUsageEnabled() {
+			// Don't track anonymous usage for now.
+			return nil
+		}
+		return err
+	}
+
+	m, err := countsToMap(counts)
```
tylerwilliams, line 78: nit: call this counts?

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,138 @@
...
+
+	t := ut.clock.Now()
+
+	pipe := ut.rdb.TxPipeline()
+	// Add the group ID to the set of groups with usage
+	groupsKey := groupsRedisKey(t)
+	pipe.SAdd(ctx, groupsKey, groupID)
+	pipe.Expire(ctx, groupsKey, redisKeyTTL)
+	// Increment the hash values
+	countsKey := countsRedisKey(groupID, t)
```
tylerwilliams, line 94: nit: groupIDCollectionPeriod or something?

enterprise/server/usage/usage.go
```diff
@@ -0,0 +1,138 @@
...
+	for k, c := range m {
```
tylerwilliams, line 95: nit: this could be made more readable by naming your loop vars and map. For example:

for countField, count := range counts {
  pipe.HIncrBy(ctx, groupIDCollectionPeriod, countField, count)
}

server/util/perms/perms.go
```diff
@@ -255,3 +255,30 @@ func AuthenticateSelectedGroupID(ctx context.Context, env environment.Env, proto
...
+
+// IsAnonymousUserError can be used to check whether an error returned by
+// functions which return the authenticated user (such as AuthenticatedUser or
+// AuthenticateSelectedGroupID) is due to an anonymous user accessing the
+// service. This is useful for allowing anonymous users to proceed, in cases
+// where anonymous usage is explicitly enabled in the app config, and we support
+// anonymous usage for the part of the service where this is used.
+func IsAnonymousUserError(err error) bool {
+	return status.IsUnauthenticatedError(err) || status.IsPermissionDeniedError(err) || status.IsUnimplementedError(err)
+}
```
tylerwilliams, line 284: This feels really broad (and potentially dangerous).-- what do the current callsites that permit anonymous usage look like now?

tylerwilliams (changes requested):

server/util/perms/perms.go
bduffany: We do this in a few other places --

Screenshot from 2021-09-16 13-24-52

enterprise/server/usage/usage.go
bduffany: Went with groupIDCollectionPeriodKey and updated groupsKey above to follow a similar pattern

enterprise/server/usage/usage.go

enterprise/server/usage/usage.go

server/util/perms/perms.go
bduffany: Resolved offline to update these callsites in a follow-up PR.

tylerwilliams (approved):
LGTM

PR #991, opened 2021-09-10
Add redsync dep

bduffany:
This will be used so that we can acquire distributed locks using Redis. See https://github.com/go-redsync/redsync

This is needed for the usage tracker to make sure that only one app can flush to the DB at a time. 

---

**Version bump**: Patch


tylerwilliams (approved):
It seems like there is a way to use goredis/v8 (which we use) here: https://github.com/go-redsync/redsync/blob/master/examples/goredis/v8/main.go

I'm happy to approve this but some of the delta is a little concerning, like the addition of github.com/go-redis/redis/v7 and github.com/gomodule/redigo.

bduffany:
@tylerwilliams I added some `exclude` directives to `go.mod` to prevent Gazelle from including these old versions of Redis as well as the redigo driver in our `deps.bzl` when running `tools/update_deps.sh`.

bduffany:
@tylerwilliams I started running into an issue that my IDE would auto-complete `redis.<whatever>` with the V6 version of Redis which is really annoying. Will try to see if there's a way to avoid depending on the unnecessary packages.

bduffany:
I tried using an `exclude` directive to exclude the deps we didn't want (redigo + redis v6 and v7) but I started seeing errors like `go: github.com/go-redsync/redsync/v4@v4.4.1: github.com/go-redsync/redsync/v4(v4.4.1) depends on excluded github.com/go-redis/redis(v6.15.9+incompatible) with no newer version available`

Wound up creating a fork + script which deletes any files that have a dep on these unused packages -- github.com/bduffany/redsync

If we want to bump our version of `redsync` then we just need to checkout that repo and run `./publish_minimal_version.sh` (we can move that repo to the BB org if needed).

PR #994, opened 2021-09-12
m1 release

siggisim:
This PR enables a darwin-arm64 executor release for m1 macs.

This only enabled the executor build for now, since much more work is needed to release m1 mac builds (we'll probably need to upgrade to rules_nodejs 4.0 which is an involved upgrade).

While doing this release, I discovered that the executor currently bundles the js app. This is because of the dependency chain executor => scheduler_client => version => bundle => app_bundle => js.

This PR breaks the chain by pulling the executor version number from a stamp rather than the version file in the bundle for release builds.

There are enough tweaks to the workflow needed that it's cleaner as a separate workflow. Once we're able to build apps, it will be pretty straightforward to combine these workflows and use the release matrix.

The build does not yet take place on remote executors because we need to deploy m1 remote executors with these binaries first.

---

**Version bump**: None

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -76,6 +76,13 @@ build:release --stamp
 build:release --define release=true
 build:release --remote_download_toplevel
 
+# Configuration used for BuildBuddy release m1 workflow
+build:release-m1 --remote_instance_name=buildbuddy-io/buildbuddy/release
+build:release-m1 -c opt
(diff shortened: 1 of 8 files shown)


tylerwilliams (approved):
LGTM (nice find! I wonder if this will change executor binary size much?)

PR #996, opened 2021-09-13
Periodically flush usage from Redis to DB

bduffany:
* Every 1 minute, apps wake up and try to grab a Redis lock (implemented with `redsync`). If the attempt fails, exit.
* Read the keys `usage/groups/$t` for t = ~5 usage periods ago, to ~2 collection periods ago
* For each group, read the keys `usage/counts/$g/$t` and `usage/invocations/$g/$t`
* For each invocation, append to the `InvocationUsage` table if the invocation isn't already present
* For the usage counts, append to the current table row, but only if the collection period data has not yet been written

TODO in a separate PR: Optimize SQL performance. We probably want to batch the queries together.

---

**Version bump**: Patch


enterprise/server/usage/usage.go
```diff
@@ -37,6 +56,19 @@ const (
 	redisCountsKeyPrefix = redisUsageKeyPrefix + "counts/"
 
 	redisTimeKeyFormat = time.RFC3339
+
+	redisUsageLockKey    = "lock.usage"
+	redisUsageLockExpiry = 45 * time.Second
```
tylerwilliams, line 61: Can you please comment these (and give an example of time.RFC3339? It is not an RFC i know by heart, unlike RFC 1149)

enterprise/server/usage/usage.go
```diff
@@ -83,41 +128,235 @@ func (ut *tracker) Increment(ctx context.Context, uc *tables.UsageCounts) error
...
+// DB.
+//
+// Public for testing only; the server should call StartDBFlush to periodically
+// flush usage.
+func (ut *tracker) FlushToDB(ctx context.Context) error {
+	// redsync.ErrFailed can be safely ignored; it just means that another app
+	// is already trying to flush.
+	if err := ut.flushToDB(ctx); err != nil && err != redsync.ErrFailed {
+		return err
+	}
```
tylerwilliams, line 184: nit: since you're doing something special here, it's probably clearer to write it out:
err := ut.flushToDB(ctx)
if err != nil && err != redsync.ErrFailed {
  return err
}
return nil

enterprise/server/usage/usage.go
```diff
@@ -83,41 +128,235 @@ func (ut *tracker) Increment(ctx context.Context, uc *tables.UsageCounts) error
...
+func (ut *tracker) flushCounts(ctx context.Context, groupID string, c collectionPeriod, counts *tables.UsageCounts) error {
+	pk := &tables.Usage{
+		GroupID:         groupID,
+		PeriodStartUsec: timeutil.ToUsec(c.UsagePeriod().Start()),
+	}
+	dbh := ut.env.GetDBHandle()
+	return dbh.Transaction(ctx, func(tx *db.DB) error {
+		// Create a row for the corresponding usage period if one doesn't already
+		// exist
+		err := tx.Clauses(clause.OnConflict{DoNothing: true}).Create(pk).Error
```
tylerwilliams, line 255: Could you do INSERT ON DUPLICATE KEY IGNORE or something instead of relying on special gorm behavior here? I think it would be clearer to the reader and also not depend on gorm intricacies.

enterprise/server/usage/usage.go
```diff
@@ -83,41 +128,235 @@ func (ut *tracker) Increment(ctx context.Context, uc *tables.UsageCounts) error
...
+func (c collectionPeriod) End() time.Time {
+	return c.Start().Add(collectionPeriodDuration)
+}
+
+// Next returns the next collection period after this one.
+func (c collectionPeriod) Next() collectionPeriod {
+	return collectionPeriod(c.End())
+}
+
+// UsagePeriod returns the usage period that this collection period is contained
```
tylerwilliams, line 330: I don't totally follow the difference between a collectionPeriod and a usageperiod -- can you explain it?

enterprise/server/usage/usage.go
```diff
@@ -136,3 +375,40 @@ func countsToMap(tu *tables.UsageCounts) (map[string]int64, error) {
 	}
 	return counts, nil
 }
+
+// stringMapToCounts converts a Redis hashmap containing usage counts to
+// tables.UsageCounts.
+func stringMapToCounts(h map[string]string) (*tables.UsageCounts, error) {
+	hInt64 := map[string]int64{}
```
tylerwilliams, line 382: it is common in go to use make() to initialize maps -- especially in cases where you know the number of entries you plan to insert. This allows you program to make one allocation of the right size, rather than having to dynamically resize the map as you insert more and more items.

hInt64  := make(map[string]int64{}, len(h))

enterprise/server/usage/usage.go
```diff
@@ -136,3 +375,40 @@ func countsToMap(tu *tables.UsageCounts) (map[string]int64, error) {
...
+	}
+	return &tables.UsageCounts{
+		Invocations:            hInt64["invocations"],
+		CASCacheHits:           hInt64["cas_cache_hits"],
+		ActionCacheHits:        hInt64["action_cache_hits"],
+		TotalDownloadSizeBytes: hInt64["total_download_size_bytes"],
+	}, nil
+}
+
+// usageUpdates returns a GORM updates map that finalizes the usage row up to
```
tylerwilliams, line 398: lol, can we un-gormify this and do this with SQL?

tylerwilliams (changes requested):

enterprise/server/usage/usage.go
bduffany: Done (added the explanation to the comment here)

enterprise/server/usage/usage.go
bduffany: Knew about `make`, just forgot to use it here. Done

enterprise/server/usage/usage.go

tylerwilliams (changes requested):

enterprise/server/usage/usage.go
```diff
@@ -83,41 +146,272 @@ func (ut *tracker) Increment(ctx context.Context, uc *tables.UsageCounts) error
...
+		defer cancel()
+		if ok, err := ut.flushMutex.UnlockContext(ctx); !ok || err != nil {
+			log.Warningf("Failed to unlock Redis lock: ok=%v, err=%s", ok, err)
+		}
+	}()
+
+	// Loop through collection periods starting from the oldest collection period
+	// that may exist in Redis (based on key expiration time) and looping up until
+	// we hit a collection period which is not yet "settled".
+
```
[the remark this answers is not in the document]

enterprise/server/usage/usage.go
```diff
@@ -83,41 +146,272 @@ func (ut *tracker) Increment(ctx context.Context, uc *tables.UsageCounts) error
...
+	ctx, cancel := context.WithTimeout(ctx, flushTimeout)
+	defer cancel()
+
+	// Grab lock. This will immediately return `redsync.ErrFailed` if another
+	// client already holds the lock.
+	if err := ut.flushMutex.LockContext(ctx); err != nil {
+		return err
+	}
+	defer func() {
+		ctx, cancel := background.ExtendContextForFinalization(ctx, redisUsageLockFinalizationTimeout)
```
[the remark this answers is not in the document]

tylerwilliams (approved):
LGTM

woooo, thanks for bearing with me on this one!

Issue #997, opened 2021-09-13
Timing tab encounters a stack overflow


DavidANeil:
`RangeError: Maximum call stack size exceeded` occurs from apparently `flame_chart.tsx:66`

I imagine this happens due to the number of actions, but it would be nice for this view to at least fail smoothly, or possibly be able to segment the view to work for any number of actions.

bduffany:
Thanks for reporting this! Sent https://github.com/buildbuddy-io/buildbuddy/pull/998 to fix.

siggisim:
@DavidANeil This fix should go live in Wednesday's release.

DavidANeil:
Thank you for the outstanding resolution time.

PR #998, opened 2021-09-13
Fix 'Maximum call stack size exceeded' in flame chart

bduffany:
Argument lists in JS have a size limit. Using an array and reducing with `Math.max` fixes the problem.

```tsx
let arr = new Array(1000000).fill(0);
Math.max(...arr) // error: 'Uncaught RangeError: Maximum call stack size exceeded'
arr.reduce((a, b) => Math.max(a, b)) // 0
```

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy/issues/997


siggisim (approved):
LGTM!

PR #1001, opened 2021-09-14
If a group has sso enabled, use that as the default login mechanism on their join page

siggisim:
When a user visits a "join organization" page and they're not logged in and the organization has SSO enabled, clicking the big login / signup button behaves the same as clicking the "Login with SSO" button.

Also prioritizes groups with a url slug over groups without one when picking a "default group" (so users in their user group and an SSO organization without an associated domain default to their SSO organization rather than their personal group).

---

**Version bump**: Patch

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -91,6 +91,7 @@ export class AuthService {
     if (user.groups.length > 0) {
       user.selectedGroup =
         (selectedGroupId && user.groups.find((group) => group.id === selectedGroupId)) ||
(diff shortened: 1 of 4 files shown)


bduffany (approved):
This LG, but wdyt about hiding the non-SSO button instead?

siggisim:
> This LG, but wdyt about hiding the non-SSO button instead?

We can mess around with the design in a follow up, but the SSO button is pretty hidden right now and I mostly want the ability to switch on whether we're dealing with an SSO login vs a non-SSO login.

PR #1002, opened 2021-09-14
Fix anti-aliasing artifacts in logo

bduffany:
Collapsed the "BBBB" into a single contiguous bold "B" since the layered Bs were causing anti-aliasing artifacts, making the top of the logo look curved, and creating a few empty pixels inside the logo.

* Deleted all "B" `<path>` elements except for the rightmost B
* Deleted the parts of the path cutting out the "insides" of the B
* Updated the upper-left / bottom-left corner coordinates to match the original leftmost B.

Before | After (screenshots are from the sidebar in dense mode):

Screenshot from 2021-09-14 10-40-56

---

**Version bump**: Patch


siggisim (approved):
Thanks for doing this!

PR #1008, opened 2021-09-14
Split Authenticator interface

siggisim:
Splits the Authenticator interfaces into 4 interfaces:
- Authenticator
- HTTPAuthenticator
- GRPCAuthenticator
- APIKeyAuthenticator

Implements MultiHTTPAuthenticator and MultiAuthenticator to override those methods with the SAMLAuthenticator

Removes the concept of a fallback authenticator from SAMLAuthenticator and the extraneous methods that it didn't implement.

This PR does not change any functionality (except for displaying a "Logged Out!" message when logging out).

---

**Version bump**: Patch


siggisim:
Need to fix test, but high level comments welcome

PR #1009, opened 2021-09-14
Support conflict keys in case of hash collision.

tylerwilliams:
See title.

### DIFF
--- a/server/util/lru/lru.go
+++ b/server/util/lru/lru.go
@@ -214,13 +214,13 @@ func (c *LRU) removeOldest() {
 	}
 }
 
-func (c *LRU) lookupItem(pk, ck uint64) (*list.Element, bool) {
-	entries, ok := c.items[pk]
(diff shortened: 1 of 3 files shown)


server/util/lru/lru.go
```diff
@@ -30,16 +32,17 @@ type Config struct {
 type LRU struct {
 	sizeFn      SizeFn
 	evictList   *list.List
-	items       map[uint64]*list.Element
+	items       map[uint64][]*list.Element
```
bduffany, line 35: nit: might be able to get away with making the key a `struct{uint64, uint64}` -- then you don't need any list iteration logic

bduffany (approved):

server/util/lru/lru.go
bduffany: (only issue I can think of is that it increases the memory used by the keys, but you'd also be saving on the extra memory required for the slice values)

PR #1014, opened 2021-09-15
FUSE file system changes to forward most IO requests to the VFS server.

vadimberezniker:
Some functionality remains in the client FUSE FS, most notably the
directory layout management. I would like to move this to the VFS server
as well, but will likely do this after doing some work on FUSE/VFS perf.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1015, opened 2021-09-15
Firecracker changes to setup VFS server.

vadimberezniker:
---

**Version bump**: None


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -894,18 +789,22 @@ func (c *FirecrackerContainer) setupNetworking(ctx context.Context) error {
...
 
-func (c *FirecrackerContainer) setupByteStreamProxy() error {
-	if c.bsProxy != nil {
+func (c *FirecrackerContainer) setupVFSServer() error {
+	if c.vfsServer != nil {
 		return nil
 	}
 
-	vsockServerPath := filepath.Join(c.getChroot(), firecrackerVSockPath) + "_" + strconv.Itoa(vsock.HostByteStreamProxyPort)
+	vsockServerPath := filepath.Join(c.getChroot(), firecrackerVSockPath) + "_" + strconv.Itoa(vsock.HostVFSServerPort)
```
tylerwilliams, line 797: nit: this feels like a function that might be better in the vsock library

like vsockListenPath(vsockFile string, port int32) string {
  return vsockFile + "_" + " strconv.Itoa(port)
}

tylerwilliams (approved):
LGTM

enterprise/server/remote_execution/containers/firecracker/firecracker.go
vadimberezniker: Done.

PR #1017, opened 2021-09-16
Add role field to group proto

bduffany:
Adding `role` directly to `Group`, rather than creating a new `UserGroup` proto and then migrating the `GetUserResponse` from `repeated UserGroup` to `repeated Group`. Figured this would have little benefit since we only ever return `Group` protos for a particular user, so this proto already implicitly represents a user-group membership.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/868

### DIFF
--- a/proto/grp.proto
+++ b/proto/grp.proto
@@ -48,7 +48,11 @@ message Group {
 
   enum Role {
     UNKNOWN_ROLE = 0;
+    // Developers cannot perform certain privileged actions such as creating API
+    // keys and viewing usage data, but can perform most other common ations
+    // such as viewing invocation history.
     DEVELOPER_ROLE = 1;
+    // Admins have unrestricted access to any data owned by this group.
     ADMIN_ROLE = 2;
   }
 }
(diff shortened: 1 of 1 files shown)


siggisim (approved):
Looks good to me - but let's have @tylerwilliams take a look too since it's a proto change.

proto/grp.proto
```diff
@@ -39,6 +41,16 @@ message Group {
...
   bool use_group_owned_executors = 7;
+
+  // Role represent's the authenticated user's role within this group, which
+  // determines which actions the user is authorized to perform.
+  Role role = 9;
+
+  enum Role {
+    UNKNOWN_ROLE = 0;
+    DEVELOPER_ROLE = 1;
+    ADMIN_ROLE = 2;
```
tylerwilliams, line 52: comment at a high level what permissions these roles will entail or what the difference is?

tylerwilliams (approved):
LGTM

proto/grp.proto
bduffany: Done

PR #1019, opened 2021-09-16
Add role field to UserGroups table

bduffany:
* Add `role` field to `UserGroups` and a DB migration to initialize all existing `UserGroup` roles to admin after the new `role` column is added.
* When creating a user, make them `admin`s whenever they auto-join an empty group (including their self-owned group, as well as the configured "default" group if it is empty). Otherwise, give them the default [1] group role.
* Also assign this default [1] role when requesting to join a group, and when approving group membership.

Tested queries on both sqlite and MySQL.

[1] For now, the "default" role when joining a group is `admin`. After we add the user management UI, we can switch this to `developer`, after which point all new users that join groups will have "developer" role.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/868


enterprise/server/backends/userdb/userdb.go
```diff
@@ -579,7 +583,29 @@ func (d *UserDB) createUser(ctx context.Context, tx *db.DB, u *tables.User) erro
...
+		err = tx.Exec(`
+			UPDATE UserGroups
+			SET role = ?
+			WHERE group_group_id = ?
+			AND NOT EXISTS (
+				SELECT * FROM (
+					SELECT * FROM UserGroups
+					WHERE group_group_id = ?
+					AND user_user_id != ?
+				) AS PreExistingUserGroups
```
tylerwilliams, line 605: it feels like there is an extra "layer" of "select *" here -- is that needed?

server/util/perms/perms.go
```diff
@@ -28,6 +28,24 @@ const (
...
+	// AdminRole means a user has unrestricted access within a group.
+	AdminRole = 1
+
+	// DefaultRole is the role assigned to users when joining a group they did
+	// not create.
+	// TODO(bduffany): Change this to DeveloperRole once we have a way to manage
+	// roles via the UI (otherwise, there would be no easy way to promote new
+	// users to admins in the meantime).
+	DefaultRole = AdminRole
 )
```
tylerwilliams, line 49: nit: can you move these to their own const () block, just for clarity?

You might also consider, for example:

type role uint32

const (
  DeveloperRole role = 1 << 0
  AdminRole role = 1 << 1
  ManagerRole role = 1 << 2
  BadRole role = 1 << 100 // <- this line won't compile, because it overflows 'role'
)

tylerwilliams (changes requested):

enterprise/server/backends/userdb/userdb.go
bduffany: This is a workaround for this issue - https://stackoverflow.com/questions/45494/mysql-error-1093-cant-specify-target-table-for-update-in-from-clause

It's probably simpler to write this as 2 separate queries though, which I *think* should be OK since this is a transaction. (Done)

server/util/perms/perms.go
bduffany: Done

enterprise/server/backends/userdb/userdb.go
```diff
@@ -579,7 +589,32 @@ func (d *UserDB) createUser(ctx context.Context, tx *db.DB, u *tables.User) erro
...
+			VALUES(?, ?, ?, ?)
+			`, u.UserID, groupID, int32(grpb.GroupMembershipStatus_MEMBER), perms.DefaultRole,
+		).Error
+		if err != nil {
+			return err
+		}
+		// Promote from default role to admin if the user is the only one in the
+		// group after joining.
+		numPreExistingUsers := int64(0)
+		err = tx.Model(&tables.UserGroup{}).Where(
```
tylerwilliams, line 603: can you just make this a regular-old-query instead of a gorm WHERE query? Eventually we will remove / change gorm and these will be more difficult to migrate

tylerwilliams (changes requested):
LGTM (with a few nits)

enterprise/server/backends/userdb/userdb.go
bduffany: Done

tylerwilliams (approved):
LGTM

PR #1020, opened 2021-09-16
Remove remoteInstanceName from memory_cache CAS key (other cache types still tbd)

tylerwilliams:
See title.

### DIFF
--- a/server/backends/memory_cache/memory_cache.go
+++ b/server/backends/memory_cache/memory_cache.go
@@ -17,9 +17,10 @@ import (
 )
 
 type MemoryCache struct {
-	l      interfaces.LRU
-	lock   *sync.RWMutex
-	prefix string
+	l                  interfaces.LRU
+	lock               *sync.RWMutex
+	cacheType          interfaces.CacheType
+	remoteInstanceName string
 }
 
 func sizeFn(value interface{}) int64 {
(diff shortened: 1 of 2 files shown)


siggisim (approved):

bduffany (approved):
LGTM, just a few questions

Does this mean that the `--remote_instance_name=$(date +%s)` trick to effectively start fresh for an invocation will no longer work? If so, is there an alternative that we'll be able to use instead?

How will this interplay with Vadim's cache-prefixing stuff? Will we still keep things separate per-prefix?

tylerwilliams:
> LGTM, just a few questions / sanity checks
> 
> Does this mean that the `--remote_instance_name=$(date +%s)` trick to effectively start fresh for an invocation will no longer work? If so, is there an alternative that we'll be able to use instead?
> 
> How will this interplay with Vadim's cache-prefixing stuff? (Assuming we'll still keep things separate per-prefix?)

I realized I forgot to respond to this before submitting, sorry! The easiest way to grok this new behavior is to look at some examples, here are the ones from the test:

// caches with the same isolation are shared.
// action caches with the same isolation are shared.
// CAS caches with different remote instance names are shared.
// Action caches with different remote instance names are not shared.
// CAS and Action caches are not shared

At the end of the day, because we are verifying the hash of the bytes written to the CAS cache, there is no value in segmenting these items by remote_instance_name, since they are guaranteed by the hash to be exactly the same.

We do still segment them by groupID, to prevent a user in one group from "fishing" for something written to the cache by another group by trying different content values and seeing if the hash exists.

Hope that clarifies, lmk if not!

tylerwilliams:
Also, so far this is only for the memory_cache, other implementations (like disk) have not yet changed.

PR #1021, opened 2021-09-17
Add initial user management list view

bduffany:
Behind the config option `app.user_management_enabled`, show the "Members" tab in the settings page. It shows a list of group members. It also includes "Edit role" and "Remove" buttons, which don't do anything yet. The plan for "Edit role" is to show a modal dialog that shows the role selector as well as an explanation of each role. The plan for "Remove" is to show a modal with a confirm button. In both cases (edit role and remove) we'll show an extra warning if the changes will affect the current user (to protect against accidentally removing yourself or demoting yourself).

https://user-images.githubusercontent.com/2414826/133817736-ee7cbdeb-e63f-416a-877d-5d8b15deb42d.mp4

In a future PR, we should also consolidate the membership requests into this page.

---

**Version bump**: Patch

### DIFF
--- a/app/capabilities/capabilities.tsx
+++ b/app/capabilities/capabilities.tsx
@@ -23,6 +23,7 @@ export class Capabilities {
   sso: boolean;
   globalFilter: boolean;
   usage: boolean;
+  userManagement: boolean;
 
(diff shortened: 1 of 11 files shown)


siggisim (approved):
Woot!

PR #1023, opened 2021-09-17
Role field plumbing for user management API

bduffany:
Depends on #1019

Wire up `UpdateGroupUsersRequest.update.role` (proto) to `UserGroups.role` (DB)

Tested E2E via the UI (implemented in #1022)

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

(I think you have but please make sure these queries work in both sqlite and mysql)

PR #1026, opened 2021-09-20
Server-side implementation for serving live event logs

tempoz:
These are the server-side changes to implement caching and serving the "live" part of event
logs. This is the part that has not yet been written to disk, which also includes the tail that is
subject to change due to ANSI cursor codes.

---

**Version bump**: Minor


server/remote_cache/live_event_log/live_event_log.go
```diff
@@ -0,0 +1,284 @@
+package live_event_log
```
bduffany, line 1: Are these the right contents for this file? It looks like the implementation here is unchanged from the old `hit_tracker`

tempoz: oops, this file isn't supposed to exist :|

server/interfaces/interfaces.go
```diff
@@ -449,6 +451,16 @@ type MetricsCollector interface {
...
+// A ProtoStore allows for storing ephemeral protos globally.
+//
+// No guarantees are made about durability of ProtoStores -- they may be
+// evicted from the backing store that maintains them (usually memcache or
+// redis), so they should *not* be used in critical path code.
+type ProtoStore interface {
+	SetMessageByKey(ctx context.Context, key string, msg proto.Message) error
+	GetMessageByKey(ctx context.Context, key string, msg proto.Message) error
+}
+
```
bduffany, line 463: I think this interface would have higher utility if it were called `KeyValueStore` with `GetString` / `SetString` methods. That way the implementations don't need to care about what's being stored. Then you could have separate utility functions like `{read,write}Proto{To,From}KeyValueStore` which handles the proto serialization

server/backends/chunkstore/chunkstore.go
```diff
@@ -462,65 +470,103 @@ func (l *writeLoop) write() error {
...
+		return nil
 	}
 }
 
 func (l *writeLoop) run() {
 	l.flushTime = time.Now().Add(year)
 	l.open = true
 	for l.open {
-		l.readFromWriteChannel()
+		req := l.readFromWriteChannel()
```
bduffany, line 503: nit: I think it would help readability & maintainability to have a few comments sprinkled in throughout this loop -- the logic here is pretty dense

server/backends/chunkstore/chunkstore.go
```diff
@@ -462,65 +470,103 @@ func (l *writeLoop) write() error {
...
 	l.timeout = false
-	var p []byte
 	select {
-	case p, l.open = <-l.writeChannel:
-		if l.open && p == nil {
-			l.flushTime = time.Unix(0, 0)
-		} else {
-			l.chunk = append(l.chunk, p...)
-			l.lastWriteSize = len(p)
+	case req, open := <-l.writeChannel:
```
bduffany, line 476: nit: consider having `writeChannel` be a stream of `*WriteRequest` rather than `WriteRequest` since you return it by reference here anyway

bduffany (changes requested):

server/interfaces/interfaces.go
bduffany: Actually, you might be able to get away with using the existing memory cache / redis cache implementations instead of introducing a new interface -- you can serialize the proto to bytes then use `digest.ComputeForMessage` as the key

server/backends/chunkstore/chunkstore.go
tempoz: Hahaha, the readFromWriteChannel function used to just be in-line here, too 😶

I'll add more comments.

server/interfaces/interfaces.go
tempoz: Regarding the first point, I sort of feel like a proto already doesn't care about what's being stored. Anything, more or less, can be serialized to a proto, and protos provide much better structure than strings when it comes to serialization/deserialization. Do you disagree?

As far as the second point goes, you hit the nail on the head. I absolutely couldn't use CAS because I can't guarantee that I know the digest at retrieval time, and I actually tripped on that for a while at first before I realized I could just write an interface directly for redis/memory and implement a new cache type.

bduffany: My comment is mostly motivated by the fact that protos aren't really "native" to any of the underlying implementations (in-memory maps in golang, or Redis), while simpler datatypes like `string` and `[]byte` are. So effectively all the implementations have to include a translation layer from proto to the underlying storage mechanism, and if you want to use this to store something that isn't a proto, you have to create a "wrapper" proto to get it to work with this interface.

But more generally, the less the interface assumes about the type of data being passed in, the more powerful of an abstraction it is, since it's compatible with more things and is easier to implement. `io.Reader` and `io.Writer` for example both work with `[]byte` and they are super ubiquitous because `[]byte` is sort of the "least common denominator" among all types of data and Go makes it very convenient to work with `[]byte`. (Speaking of... the KeyValueStore interface I mentioned should probably also deal with `[]byte` for the values rather than `string`)

tempoz: I am much more onboard with using `byte[]`. It was encoding things as strings that was making me uncomfy :P

tempoz: Done!

server/interfaces/interfaces.go
```diff
@@ -449,6 +451,18 @@ type MetricsCollector interface {
...
+// A KeyValStore allows for storing ephemeral values globally.
+//
+// No guarantees are made about durability of KEyValStores -- they may be
+// evicted from the backing store that maintains them (usually memcache or
+// redis), so they should *not* be used in critical path code.
+type KeyValStore interface {
+	SetByKey(ctx context.Context, key string, val []byte) error
+	GetByKey(ctx context.Context, key string) ([]byte, error)
+	SetProtoByKey(ctx context.Context, key string, msg proto.Message) error
+	GetProtoByKey(ctx context.Context, key string, msg proto.Message) error
```
bduffany, line 463: nit: these could be util methods so that implementations of `KeyValStore` don't need to implement proto serialization / deserialization -- similar to how `cachetools.go` has utility methods for reading/writing protos from CAS:

```go
// file: server/util/keyval.go
package keyval

import (
)

func GetProto(store interfaces.KeyValStore, key string, msg proto.Message) error {
  ...
}

func SetProto(store interfaces.KeyValStore, key string, msg proto.Message) error {
  ...
}
```

server/interfaces/interfaces.go
```diff
@@ -449,6 +451,18 @@ type MetricsCollector interface {
...
 }
 
+// A KeyValStore allows for storing ephemeral values globally.
+//
+// No guarantees are made about durability of KEyValStores -- they may be
+// evicted from the backing store that maintains them (usually memcache or
+// redis), so they should *not* be used in critical path code.
+type KeyValStore interface {
+	SetByKey(ctx context.Context, key string, val []byte) error
+	GetByKey(ctx context.Context, key string) ([]byte, error)
```
bduffany, line 461: uber nit: these could just be `Get/Set` (since most functionality afforded by a key-value store requires a single key as input) -- also harmonious with `redis.Get` / `redis.Set`

tempoz: They can't be, since redis_cache is one of the things that needs to implement this interface, and Get and Set are already defined for the digest-style Cache, which it implements. First thing I tried :/

server/interfaces/interfaces.go
tempoz: This is neat, thanks.

tempoz: Done!

server/interfaces/interfaces.go
bduffany: I think `redis_cache` is maybe not the best place to implement this KV interface -- I get that it's convenient because `redis_cache` already owns a Redis client and has some Redis-related logic already in there, but our notion of "cache" today usually refers to the CAS and the Action Cache (AC), so I think it may be confusing to have this generalized KV interface implemented by something that is really intended to just implement the CAS and AC. For example, `disk_cache` is another cache implementation, but I'd guess we won't want a KV implementation that is disk-based.

I think it's fine to create a separate one-off implementation, maybe `enterprise/server/util/redis_key_value_store.go`

bduffany (approved):
LGTM

It's mildly smelly that `redis_cache.go` has some stuff in there that's not related to AC or CAS (like `IncrementCount` and now this new `KeyValueStore` implementation) -- might be worth cleaning this up in a separate PR.

tylerwilliams (approved):
LGTM

PR #1027, opened 2021-09-20
Remove some high-cardinality metric labels

bduffany:
Results of running this query from https://www.robustperception.io/which-are-my-biggest-metrics

```
topk(10, count by (__name__)({__name__=~".+"}))
```

Screenshot from 2021-09-20 15-11-00

The SQL and HTTP metrics are some of the highest cardinality metrics since the HTTP route and SQL query template are highly dynamic. Removing these labels for now, and updated the Grafana charts to no longer reference these labels.

---

**Version bump**: Patch

### DIFF
--- a/server/http/filters/BUILD
+++ b/server/http/filters/BUILD
@@ -10,7 +10,6 @@ go_library(
         "//server/http/protolet",
         "//server/metrics",
         "//server/util/log",
-        "//server/util/status",
         "//server/util/uuid",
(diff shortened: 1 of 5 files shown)


tylerwilliams (commented):
In terms of storage space, how does this compare to the savings from removing the (unused) k8s metrics in https://github.com/buildbuddy-io/buildbuddy-internal/pull/871?

PR #1035, opened 2021-09-22
Fix bugs with usage flush logic and improve concurrency tests

bduffany:
* Fix the `final_before_usec` update condition; we want to update the usage row only if it is finalized up to the _start_ of the current collection period, not the end
* Improve the concurrent access tests to disable Redlock, so we can test that concurrent flushing is safe even if Redis fails. Also assert that the result is correct rather than just asserting that there was no error (this is what caught the bug above).

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/usage/usage.go
+++ b/enterprise/server/usage/usage.go
@@ -197,7 +197,7 @@ func (ut *tracker) FlushToDB(ctx context.Context) error {
 	}
 	defer func() {
 		if err := ut.flushLock.Unlock(ctx); err != nil {
-			log.Warningf("Failed to unlock Redis lock: %s", err)
+			log.Warningf("Failed to unlock distributed lock: %s", err)
 		}
 	}()
 	return ut.flushToDB(ctx)
--- a/enterprise/server/usage/usage.go
+++ b/enterprise/server/usage/usage.go
@@ -256,9 +256,9 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, c collection
 		PeriodStartUsec: timeutil.ToUsec(c.UsagePeriod().Start()),
 	}
 	dbh := ut.env.GetDBHandle()
-	// Create a row for the corresponding usage period if one doesn't already
-	// exist.
 	return dbh.Transaction(ctx, func(tx *db.DB) error {
+		// Create a row for the corresponding usage period if one doesn't already
+		// exist.
 		res := tx.Exec(`
 			INSERT `+dbh.InsertIgnoreModifier()+` INTO Usages (
 				group_id,
--- a/enterprise/server/usage/usage.go
+++ b/enterprise/server/usage/usage.go
@@ -16,7 +16,6 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/timeutil"
 	"github.com/go-redis/redis/v8"
-
 )
 
 const (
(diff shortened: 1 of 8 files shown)


enterprise/server/usage/usage.go
```diff
@@ -79,16 +78,24 @@ var (
...
 
 type TrackerOpts struct {
+	// TEST ONLY options
+
+	// Clock overrides the clock used by the usage tracker.
 	Clock timeutil.Clock
+
+	// DisableRedlock disables the Redis locking mechanism used to prevent
+	// flush operations from accessing the DB concurrently.
+	DisableRedlock bool
```
tylerwilliams, line 88: Test only options that disable a big chunk of the impl are kinda gross -- could we pass in a distributed lock interface in the constructor and then implement that with a regular old lock or no-ops for testing?

type DistributedLock interface {
  Lock(ctx context.Context, key string, expiry time.Duration)
  Unlock(ctx context.Context, key string)
}

enterprise/server/usage/usage.go
```diff
@@ -254,12 +264,9 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, c collection
 		PeriodStartUsec: timeutil.ToUsec(c.UsagePeriod().Start()),
 	}
 	dbh := ut.env.GetDBHandle()
-	return dbh.Transaction(ctx, func(tx *db.DB) error {
```
tylerwilliams, line 257: https://www.percona.com/blog/2014/10/28/how-to-deal-with-mysql-deadlocks/

tylerwilliams (changes requested):
I'm hesitant to just remove all transactions here when we don't understand the reason we're getting a deadlock, even though I think what you're suggesting about the queries being safe is probably true right now?

enterprise/server/usage/usage.go

enterprise/server/usage/usage.go
bduffany: (Working on this)

bduffany: Decided that we'll keep this in a transaction and that we can tolerate these deadlocks once in a blue moon; the worst that happens is that we just log a warning -- MySQL will ensure that at least one of the contending transactions will go through.

enterprise/server/usage/usage.go
```diff
@@ -79,16 +78,23 @@ var (
...
 
 type TrackerOpts struct {
+	// TEST ONLY options
+
+	// Clock overrides the clock used by the usage tracker.
 	Clock timeutil.Clock
+
+	// FlushLock overrides the distributed lock to serialize access to the DB
+	// across flush jobs.
+	FlushLock interfaces.DistributedLock
```
tylerwilliams, line 88: nit: can you remove this from the opts and make it mandatory (like, accept it as the second param in the NewTracker func?)

enterprise/server/usage/usage.go
```diff
@@ -100,20 +106,16 @@ func NewTracker(env environment.Env, opts *TrackerOpts) (*tracker, error) {
...
-
-	pool := redsync_goredis.NewPool(rdb)
-	rs := redsync.New(pool)
-	flushMutex := rs.NewMutex(
-		redisUsageLockKey,
-		redsync.WithTries(1), redsync.WithExpiry(redisUsageLockExpiry),
-	)
-
+	flushLock := opts.FlushLock
+	if flushLock == nil {
```
tylerwilliams, line 110: nit: this conditional logic is still error-prone. If you accept the lock interface as a param and your impl just always uses it, then we'll be good

enterprise/server/usage/usage.go
tylerwilliams: SG

tylerwilliams (approved):
LGTM

enterprise/server/usage/usage.go

PR #1036, opened 2021-09-22
Reduce allocations in consistent_hash.GetAllReplicas()

tylerwilliams:
The current behavior is to allocate a new []string{} containing the 16 replicas on each call to GetAllReplicas(). Because a new string slice is allocated each time, this results in a significant number of allocations. See the use in prod for example:

image

This CL does the following:
 - stores the hash ring as a map of hash => valueIndex, rather than storing the value directly
 - maintains a cache of replica-slices
 - computes a replica-slice-key and serves a cached replica-slice if one is available, caching new ones for future use

This drastically reduces the number of allocations needed to serve replica sets. Before usage looked like this:

Screen Shot 2021-09-22 at 11 40 15 AM

After it looks like this:

Screen Shot 2021-09-22 at 11 40 08 AM


siggisim (approved):

server/util/consistent_hash/consistent_hash.go
```diff
@@ -38,21 +42,26 @@ func (c *ConsistentHash) GetItems() []string {
...
-func (c *ConsistentHash) Set(items ...string) {
+func (c *ConsistentHash) Set(items ...string) error {
+	if len(items) > 256 {
+		return status.InvalidArgumentError("Too many items in consistent hash, max allowed: 256")
+	}
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	c.keys = make([]int, 0)
-	c.ring = make(map[int]string, 0)
+	c.ring = make(map[int]uint8, 0)
```
vadimberezniker, line 52: We should reset `replicaSets` because the existing cache data won't be useful if the items change.

vadimberezniker (approved):
very nice

server/util/consistent_hash/consistent_hash.go
tylerwilliams: Done.

PR #1039, opened 2021-09-23
Refactor a few anonymous user checks to use new util func

bduffany:
We eventually will want a cleaner and less error-prone way to deal with anonymous users (maybe return a valid `UserInfo` object that just has an empty userID / groupID?) but this PR at least consolidates some of the existing ugliness into a centralized location.

---

**Version bump**: Patch

### DIFF
--- a/server/util/capabilities/capabilities.go
+++ b/server/util/capabilities/capabilities.go
@@ -52,7 +52,7 @@ func IsGranted(ctx context.Context, env environment.Env, cap akpb.ApiKey_Capabil
 		}
 		return int32(cap)&AnonymousUserCapabilitiesMask > 0, nil
 	}
-	user, err := perms.AuthenticatedUser(ctx, env)
+	user, err := a.AuthenticatedUser(ctx)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


server/util/capabilities/capabilities.go
```diff
@@ -43,18 +43,10 @@ func ToInt(caps []akpb.ApiKey_Capability) int32 {
 }
 
 func IsGranted(ctx context.Context, env environment.Env, cap akpb.ApiKey_Capability) (bool, error) {
```
tylerwilliams, line 45: Seems like the behavior here is changing slightly in the case where auth is not set up -- does this cause any issue?

tylerwilliams (approved):
LGTM

server/util/capabilities/capabilities.go
vadimberezniker: It looks like this function is called in 3 places and all 3 have an additional check (e.g. AttachUserPrefix) that already perform the anon check so this seems like a safe/good change. Can you confirm this was the motivation for it?

vadimberezniker (approved):

PR #1040, opened 2021-09-23
Frontend implementation for serving live event logs

tempoz:
- frontend changes to support "live" event logs

Truncates the logs to not include any live section, if it exists, whenever the log updates.
Sits on top of the server-side changes branch, only changes are in `app/invocation/invocation_logs_model.tsx`

---

**Version bump**: Minor

### DIFF
--- a/app/invocation/invocation_logs_model.tsx
+++ b/app/invocation/invocation_logs_model.tsx
@@ -18,6 +18,9 @@ export default class InvocationLogsModel {
   private logs = "";
   private responseSubscription: Subscription;
   private pollTailTimeout: number | null = null;
+
+  // Length of the log prefix which has already been persisted. The remainder of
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_logs_model.tsx
```diff
@@ -18,6 +18,7 @@ export default class InvocationLogsModel {
   private logs = "";
   private responseSubscription: Subscription;
   private pollTailTimeout: number | null = null;
+	private stableLogLength = 0;
```
bduffany, line 21: nit: Add a comment like `// Length of the log prefix which has already been persisted. The remainder of the log is considered "live" and may be updated on subsequent fetches`

bduffany (approved):

PR #1042, opened 2021-09-23
FUSE optimizations

vadimberezniker:
- Cache node attributes. `getattr` OPs are frequent so it helps to
avoid extra RPCs.
 - When opening small files for reading, return the file contents inline
to avoid the overhead of additional read RPCs.
 - Bring back `CopyFileRange` which got dropped in the vfs server
transition.
 - Don't lock on read requests as FUSE can fire off multiple read requests in parallel. 
 - Bump write buffer size to maximum supported by kernel

Other:
 - Instrument FUSE ops so we can know how much time is spent in the user
space code broken down by each OP type.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1043, opened 2021-09-23
Use different redis client for usage tracker

bduffany:
We don't set the redis_cache target in dev which makes the server fail to start when usage_tracking_enabled is configured as `true`.

Separately we'll think about whether it makes sense to define a new Redis target just for the usage tracker, and I can make that change afterwards. Just want to unbork dev.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM to unbork dev

PR #1045, opened 2021-09-24
Use jsonpb for capabilities config

bduffany:
This will allow us to use structured protos in the FrontendConfig (planning on using this for the per-role permissions config, which is a relatively small proto that doesn't seem worth adding an extra RPC for).

It's also nice to have the config available as a proto because we no longer need to extract individual fields from the proto and set them on the capabilities class.

---

**Version bump**: Patch

### DIFF
--- a/app/root/root.tsx
+++ b/app/root/root.tsx
@@ -23,6 +23,8 @@ interface State {
   preferences: UserPreferences;
 }
 
+capabilities.register("BuildBuddy Community Edition", false, [Path.invocationPath]);
+
 export default class RootComponent extends React.Component {
   state: State = {
     user: null,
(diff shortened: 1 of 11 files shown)


app/root/root.tsx
```diff
@@ -23,6 +23,8 @@ interface State {
   preferences: UserPreferences;
 }
 
+capabilities.register("BuildBuddy Community Edition", false, [Path.invocationPath]);
```
bduffany, line 26: fyi, moved the capabilities registration here so that `UserPreferences` can read them in the constructor

siggisim (approved):
this is great

siggisim:
lgtm

PR #1048, opened 2021-09-24
Don't mark disconnect for successful builds

tempoz:
Accumulates the status of the `BuildFinished` event so that we can avoid marking disconnect
for builds which succeed and then experience cancellation or other stream failures.

---

**Version bump**: Patch

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -159,21 +159,17 @@ func (e *EventChannel) writeCompletedBlob(ctx context.Context, blobID string, in
 }
 
 func (e *EventChannel) MarkInvocationDisconnected(ctx context.Context, iid string) error {
-	if !e.beValues.FinishedSuccess() {
-		e.statusReporter.ReportDisconnect(ctx)
+	if !e.beValues.BuildFinished() {
+		return e.FinalizeInvocation(iid)
 	}
+	e.statusReporter.ReportDisconnect(ctx)
 
 	if err := e.pw.Flush(ctx); err != nil {
 		return err
 	}
-
-	status := inpb.Invocation_DISCONNECTED_INVOCATION_STATUS
-	if e.beValues.FinishedSuccess() {
-		status = inpb.Invocation_COMPLETE_INVOCATION_STATUS
-	}
 	invocation := &inpb.Invocation{
 		InvocationId:     iid,
-		InvocationStatus: status,
+		InvocationStatus: inpb.Invocation_COMPLETE_INVOCATION_STATUS,
 	}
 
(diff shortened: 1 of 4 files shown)


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -159,14 +159,21 @@ func (e *EventChannel) writeCompletedBlob(ctx context.Context, blobID string, in
 }
 
 func (e *EventChannel) MarkInvocationDisconnected(ctx context.Context, iid string) error {
-	e.statusReporter.ReportDisconnect(ctx)
+	if !e.beValues.FinishedSuccess() {
```
siggisim, line 162: I'm a little nervous about this method diverging from FinalizeInvocation. What do you think about a simple early exit at the top:

```
if (e.besValues.Finished()) {
  return e.FinalizeInvocation(iid);
}
```

and deciding to do this whether or not the finish event was received rather than the exit code being successful. Looks like we already look at the exit code in fillInvocationFromEvents:
https://github.com/buildbuddy-io/buildbuddy/blob/f457925906010e188c35e961befb9ecf96134e4d/server/build_event_protocol/event_parser/event_parser.go#L145

Would probably be better to mark an invocation as "failed" if it received the finished event without success rather than disconnected.

Sorry I didn't investigate this well enough when commenting on the issue. Thanks for getting this done so quickly!

siggisim (changes requested):
Looks good, just one thought!

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: It looks like a good idea to me. And it's an easy change, no big deal :P

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -159,14 +159,17 @@ func (e *EventChannel) writeCompletedBlob(ctx context.Context, blobID string, in
...
+		return e.FinalizeInvocation(iid)
+	}
 	e.statusReporter.ReportDisconnect(ctx)
 
 	if err := e.pw.Flush(ctx); err != nil {
 		return err
 	}
 	invocation := &inpb.Invocation{
 		InvocationId:     iid,
-		InvocationStatus: inpb.Invocation_DISCONNECTED_INVOCATION_STATUS,
```
siggisim, line 169: I think this should flip back to DISCONNECTED

siggisim (approved):
Woot! This looks great, just one nit.

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: Oh whoa oops, thanks for the catch.

siggisim (approved):

PR #1049, opened 2021-09-24
Move selected group logic to the server

bduffany:
This will allow us to expand `GetUserResponse` with the list of resources (API endpoints) that are accessible to the user within their currently selected group.

Replicating the exact logic that we use on the client to determine the "fallback" selected group when a preference is not set or the user no longer has access to the selected group.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1050, opened 2021-09-24
Fix usage tracking for action_cache_hits

bduffany:
The usage tracking code was relying on the `actionCache` parameter to `makeCloseFunc`, but this was effectively always `false` and unused except by the usage tracker. `h.actionCache` is really what keeps track of whether the hit tracker is tracking hits for AC vs. CAS.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1052, opened 2021-09-24
Add default redis target

tempoz:
- Add default redis target and use it
- Use default redis target in local enterprise config

Defines a new redis target `app.default_redis_target` to use for usage tracking, metrics collection,
and event log caching. Falls back to the remote execution redis target if unspecified for migration reasons.
Acts as a fallback for the remote execution target if it is unspecified for intentional design reasons.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/server/main.go
+++ b/enterprise/server/cmd/server/main.go
@@ -193,11 +193,7 @@ func main() {
 	if redisTarget := configurator.GetDefaultRedisTarget(); redisTarget != "" {
 		redisClient := redisutil.NewClient(redisTarget, healthChecker, "default_redis")
 		realEnv.SetDefaultRedisClient(redisClient)
-		maxValueSizeBytes := int64(0)
-		if redisConfig := configurator.GetDefaultRedisConfig(); redisConfig != nil {
-			maxValueSizeBytes = redisConfig.MaxValueSizeBytes
-		}
-		r := redis_cache.NewCache(realEnv.GetDefaultRedisClient(), maxValueSizeBytes)
+		r := redis_cache.NewCache(realEnv.GetDefaultRedisClient(), 0)
 		realEnv.SetMetricsCollector(r)
 		realEnv.SetKeyValStore(r)
 	}
--- a/enterprise/server/cmd/server/main.go
+++ b/enterprise/server/cmd/server/main.go
@@ -193,6 +193,13 @@ func main() {
 	if redisTarget := configurator.GetDefaultRedisTarget(); redisTarget != "" {
 		redisClient := redisutil.NewClient(redisTarget, healthChecker, "default_redis")
 		realEnv.SetDefaultRedisClient(redisClient)
+		maxValueSizeBytes := int64(0)
+		if redisConfig := configurator.GetDefaultRedisConfig(); redisConfig != nil {
+			maxValueSizeBytes = redisConfig.MaxValueSizeBytes
+		}
+		r := redis_cache.NewCache(realEnv.GetDefaultRedisClient(), maxValueSizeBytes)
+		realEnv.SetMetricsCollector(r)
+		realEnv.SetKeyValStore(r)
 	}
 
 	if redisTarget := configurator.GetRemoteExecutionRedisTarget(); redisTarget != "" {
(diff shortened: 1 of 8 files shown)


enterprise/server/cmd/server/main.go
```diff
@@ -262,7 +267,7 @@ func main() {
 		log.Infof("Enabling memcache layer with targets: %s", mcTargets)
 		mc := memcache.NewCache(mcTargets...)
 		realEnv.SetCache(composable_cache.NewComposableCache(mc, realEnv.GetCache(), composable_cache.ModeReadThrough|composable_cache.ModeWriteThrough))
-	} else if redisTarget := configurator.GetCacheRedisTarget(); redisTarget != "" {
+	} else if redisTarget := configurator.GetDefaultRedisTarget(); redisTarget != "" {
```
bduffany, line 270: I think we want to keep the dedicated cache target here, but pull out the `SetMetricsCollector` and `SetKeyValStore` from this block and use the default target for those instead. That way, adding `app.default_redis_target` will not cause the Redis cache to be enabled (today, we intentionally don't specify `redis.cache_target` because we don't want to use the Redis cache).

enterprise/server/cmd/server/main.go
```diff
@@ -193,6 +193,13 @@ func main() {
 	if redisTarget := configurator.GetDefaultRedisTarget(); redisTarget != "" {
 		redisClient := redisutil.NewClient(redisTarget, healthChecker, "default_redis")
 		realEnv.SetDefaultRedisClient(redisClient)
+		maxValueSizeBytes := int64(0)
+		if redisConfig := configurator.GetDefaultRedisConfig(); redisConfig != nil {
+			maxValueSizeBytes = redisConfig.MaxValueSizeBytes
+		}
```
bduffany, line 199: The `maxValueSizeBytes` is only needed by the cache and is unused by the implementations of `MetricsCollector` and `KeyValStore`. I think we can hard-code this as zero for now, to avoid needing to introduce a `RedisConfig` section for the default Redis target, and also add a TODO to move the MetricsCollector and KeyValStore implementations out of `redis_cache.go` and into their own packages.

server/config/config.go
```diff
@@ -719,7 +729,13 @@ func (c *Configurator) GetRemoteExecutionRedisTarget() string {
 	if rec := c.GetRemoteExecutionConfig(); rec != nil && rec.RedisTarget != "" {
 		return rec.RedisTarget
 	}
-	// Fall back to the cache redis target if redis target is not specified in remote execution config.
+
+	// If no remote execution target is defined, use the default.
+	if c.gc.App.DefaultRedisTarget != "" {
+		return c.gc.App.DefaultRedisTarget
+	}
```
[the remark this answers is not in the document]
bduffany: Good point, if they are adding a default cache target on their own, then they are likely doing so based on some documentation they're reading, or as part of some change we're recommending, so it seems fine to address it via documentation.

bduffany (changes requested):

enterprise/server/cmd/server/main.go
tempoz: Sounds good.

server/config/config.go
```diff
@@ -174,9 +175,9 @@ type DistributedCacheConfig struct {
...
 }
 
-type RedisCacheConfig struct {
-	RedisTarget       string `yaml:"redis_target" usage:"A redis target for improved Caching/RBE performance. Target can be provided as either a redis connection URI or a host:port pair. URI schemas supported: redis[s]://[[USER][:PASSWORD]@][HOST][:PORT][/DATABASE] or unix://[[USER][:PASSWORD]@]SOCKET_PATH[?db=DATABASE] ** Enterprise only **"`
-	MaxValueSizeBytes int64  `yaml:"max_value_size_bytes" usage:"The maximum value size to cache in redis (in bytes)."`
+type RedisConfig struct {
+	RedisTarget       string `yaml:"redis_target" usage:"A redis target to use for usage tracking, metrics collection, caching, or remote execution, depending on configuration options. Target can be provided as either a redis connection URI or a host:port pair. URI schemas supported: redis[s]://[[USER][:PASSWORD]@][HOST][:PORT][/DATABASE] or unix://[[USER][:PASSWORD]@]SOCKET_PATH[?db=DATABASE] ** Enterprise only **"`
+	MaxValueSizeBytes int64  `yaml:"max_value_size_bytes" usage:"The maximum value size to store in redis (in bytes)."`
```
bduffany, line 180: nit: this diff chunk can be reverted

bduffany (approved):
LGTM modulo one nit -- thanks for doing this!

server/config/config.go
tempoz: Done.

PR #1053, opened 2021-09-24
Add group roles to UserInfo interface

bduffany:
In some places where we check that the user has access to a particular group, we'll now want to check whether they're in the group _and_ they fulfill a certain role within the group. This makes the change to also select the `role` field from `UserGroup` in all the places where we already join with the `UserGroups` table for group memberships.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -481,16 +481,16 @@ func lookupUserFromSubID(env environment.Env, ctx context.Context, subID string)
 		}
 		groupRows, err := tx.Raw(`
 			SELECT
-			  g.user_id,
-				g.group_id,
-				g.url_identifier,
-				g.name,
-				g.owned_domain,
-				g.github_token,
-				g.sharing_enabled,
-				g.use_group_owned_executors,
-				g.saml_idp_metadata_url,
-				ug.role
+			g.user_id,
+			g.group_id,
+			g.url_identifier,
+			g.name,
+			g.owned_domain,
+			g.github_token,
+			g.sharing_enabled,
+			g.use_group_owned_executors,
+			g.saml_idp_metadata_url,
+			ug.role
 			FROM Groups AS g, UserGroups AS ug
 			WHERE g.group_id = ug.group_group_id
 			AND ug.membership_status = ?
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -19,6 +19,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/db"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lru"
+	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/timeutil"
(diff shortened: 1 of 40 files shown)


enterprise/server/auth/auth.go
```diff
@@ -472,19 +481,46 @@ func lookupUserFromSubID(env environment.Env, ctx context.Context, subID string)
...
+			  g.user_id,
+				g.group_id,
+				g.url_identifier,
+				g.name,
+				g.owned_domain,
+				g.github_token,
+				g.sharing_enabled,
+				g.use_group_owned_executors,
+				g.saml_idp_metadata_url,
+				ug.role
```
bduffany, line 495: FYI this is selecting all the fields from Group except

* Model fields (created_at_usec, updated_at_usec) -- we don't reference these anywhere
* WriteToken and APIKey -- these aren't used anywhere either; they can be removed in a separate PR

enterprise/server/auth/auth.go
```diff
@@ -472,19 +479,46 @@ func lookupUserFromSubID(env environment.Env, ctx context.Context, subID string)
 		if err := userRow.Take(user).Error; err != nil {
 			return err
 		}
-		groupRows, err := tx.Raw(`SELECT g.* FROM `+"`Groups`"+` as g JOIN UserGroups as ug
-                                          ON g.group_id = ug.group_group_id
-                                          WHERE ug.user_user_id = ?`, user.UserID).Rows()
+		groupRows, err := tx.Raw(`
+			SELECT
+			  g.user_id,
```
vadimberezniker, line 484: nit: fix indent

enterprise/server/auth/auth.go
```diff
@@ -117,6 +120,10 @@ func (c *Claims) GetAllowedGroups() []string {
 	return c.AllowedGroups
 }
 
+func (c *Claims) GetGroupRoles() []uint32 {
```
vadimberezniker, line 123: Can we combine the group ID & role into a single struct instead of returning them separately? I don't think the user of this API should have to lookup this data in two slices. 

Same for the Claims struct, is there a reason not to store the per-group data (ID + role) together? We could introduce a new field and remove the old allowedgroups field later once it's no longer used.

enterprise/server/auth/auth.go
bduffany: This indent is intentional since the field list is syntactically a "child" of the SELECT

enterprise/server/auth/auth.go
bduffany: Done

enterprise/server/auth/auth.go
```diff
@@ -98,11 +100,12 @@ func jwtKeyFunc(token *jwt.Token) (interface{}, error) {
...
 type Claims struct {
 	jwt.StandardClaims
-	UserID                 string                   `json:"user_id"`
-	GroupID                string                   `json:"group_id"`
-	AllowedGroups          []string                 `json:"allowed_groups"`
-	Capabilities           []akpb.ApiKey_Capability `json:"capabilities"`
-	UseGroupOwnedExecutors bool                     `json:"use_group_owned_executors,omitempty"`
+	UserID                 string                        `json:"user_id"`
+	GroupID                string                        `json:"group_id"`
+	AllowedGroups          []string                      `json:"allowed_groups"`
```
vadimberezniker, line 105: Add a TODO to remove this field.

enterprise/server/auth/auth.go
vadimberezniker: The indentation you had is good. I was pointing out that for some reason the indentation for the first field was different from the remaining fields.

vadimberezniker (approved):

enterprise/server/auth/auth.go
bduffany: Gotcha, I think it was indented with spaces instead of tabs so I didn't notice it locally. Done

enterprise/server/auth/auth.go
bduffany: Done

PR #1054, opened 2021-09-24
Add an Azure blobstore implementation.

tylerwilliams:
Implements an Azure blobstore for storing invocations.

Tested bucket creation and reuse, normal invocation reading / writing. All work.


siggisim (approved):
Woot!

PR #1056, opened 2021-09-27
Implement server-side admin-only auth checks

bduffany:
Builds on https://github.com/buildbuddy-io/buildbuddy/pull/1053 (please review that one first)

---

**Version bump**: Patch

### DIFF
--- a/server/buildbuddy_server/buildbuddy_server.go
+++ b/server/buildbuddy_server/buildbuddy_server.go
@@ -118,10 +118,11 @@ func (s *BuildBuddyServer) DeleteInvocation(ctx context.Context, req *inpb.Delet
 	return &inpb.DeleteInvocationResponse{}, nil
 }
 
-func makeGroups(grps []*tables.Group) []*grpb.Group {
+func makeGroups(groupRoles []*tables.GroupRole) []*grpb.Group {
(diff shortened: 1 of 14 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -41,6 +42,24 @@ const (
 	actioncacheProtocolPrefix = "actioncache://"
 )
 
+var (
+	// AdminAPIMethods holds the list of service methods which are accessible
+	// only by group admins.
+	AdminAPIMethods = []string{
```
bduffany, line 48: FYI this is modeled as an explicit list of methods (rather than just checking `IsAdmin` in every method) so that we can return this to the client, so the client knows which RPCs it is authorized to perform without having to care about user roles.

vadimberezniker (commented):
Would it be better to implement this as an interceptor that is applicable to all BuildBuddyService RPCs with a mapping that identifies which checks apply to which APIs? This would allow us to have a single place to document which APIs need checks and which ones don't (and it would make it hard to miss adding checks for a new API method). 

As it stands currently, figuring out which permissions are needed by which API methods requires looking at each individual method implementation which is a bit cumbersome.

PR #1059, opened 2021-09-27
Clean up `Groups.api_key` column

bduffany:
This has been replaced with the API keys table.

GORM doesn't delete old columns for us, so I added a migration to explicitly remove the column. I am OK with removing this though if we don't think it's worth the added migration code and/or the risk of losing data permanently due to incorrect migration logic. (I figure since this column hasn't been in use for such a long time, the risk is minimal)

---

**Version bump**: Patch

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -161,6 +161,10 @@ type Group struct {
 	// email belonging to that domain may be added to this group.
 	OwnedDomain string `gorm:"index:owned_domain_index"`
 
+	// The group access token. This token allows writing data for this
+	// group.
+	WriteToken string `gorm:"index:write_token_index"`
+
 	// The group's Github API token.
 	GithubToken string
 	Model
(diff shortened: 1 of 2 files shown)


vadimberezniker (commented):
`GetAPIKeyGroupFromBasicAuth` seems to still be using the write_token field, which seems like a bug. 
Do you know where we use basic auth? Is that something that is actively used and needs to be supported? @siggisim @tylerwilliams

bduffany:
> `GetAPIKeyGroupFromBasicAuth` seems to still be using the write_token field, which seems like a bug. Do you know where we use basic auth? Is that something that is actively used and needs to be supported? @siggisim @tylerwilliams

Good catch, I didn't see that reference. Reverted the change for now and maybe later we can think about whether it's safe to delete.

server/tables/tables.go
```diff
@@ -531,7 +527,7 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
 	}
 
 	// Migrate Groups.APIKey to APIKey rows.
```
vadimberezniker, line 529: offtopic: is it time to get rid of this?

vadimberezniker (approved):

server/tables/tables.go
bduffany: I think maybe we can consider removing DB migrations as part of each major BB release, and having a separate tool to help customers upgrade their DB from older to newer DB versions between upgrades. (e.g. to upgrade from V1 to V2 to V3, the tool would check out the last BB release for each version then run the DB migration logic from each release).

PR #1060, opened 2021-09-28
Add option to force darwin executions to run on user-owned executors

siggisim:
---

**Version bump**: Patch


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -493,25 +497,25 @@ func NewSchedulerServerWithOptions(env environment.Env, options *Options) (*Sche
...
-func (s *SchedulerServer) GetGroupIDAndDefaultPoolForUser(ctx context.Context) (string, string, error) {
+func (s *SchedulerServer) GetGroupIDAndDefaultPoolForUser(ctx context.Context, os string) (string, string, error) {
 	defaultPool := s.env.GetConfigurator().GetRemoteExecutionConfig().DefaultPoolName
 	if !s.enableUserOwnedExecutors {
 		return "", defaultPool, nil
 	}
-
 	user, err := perms.AuthenticatedUser(ctx, s.env)
 	if err != nil {
 		if s.env.GetConfigurator().GetAnonymousUsageEnabled() {
```
vadimberezniker, line 521: This check would send anonymous executions to the shared pools since it happens before the check you added.

vadimberezniker (changes requested):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
siggisim: That would have been alright since that pool wouldn't have any executors.

I've made this more explicit by returning an error in this case.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -493,27 +497,30 @@ func NewSchedulerServerWithOptions(env environment.Env, options *Options) (*Sche
...
-func (s *SchedulerServer) GetGroupIDAndDefaultPoolForUser(ctx context.Context) (string, string, error) {
+func (s *SchedulerServer) GetGroupIDAndDefaultPoolForUser(ctx context.Context, os string) (string, string, error) {
 	defaultPool := s.env.GetConfigurator().GetRemoteExecutionConfig().DefaultPoolName
 	if !s.enableUserOwnedExecutors {
 		return "", defaultPool, nil
 	}
-
 	user, err := perms.AuthenticatedUser(ctx, s.env)
 	if err != nil {
+		if s.forceUserOwnedDarwinExecutors && os == "darwin" {
```
vadimberezniker, line 521: this check should go inside the following if statement

tylerwilliams (approved):
LGTM

(nit: make "darwin" a const?)

vadimberezniker (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
siggisim: Done

siggisim:
> (nit: make "darwin" a const?)

Done

PR #1061, opened 2021-09-28
Support usage tracking across regions

bduffany:
Adding `region` to the composite key for the usage table, so that flushes from Redis -> DB in each region only finalize the current collection period + group's data for the current region, rather than all regions.

As a bonus, this will make it easy to allow users to segment usage by region in the UI (if that's something they're interested in).

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/usage/usage.go
+++ b/enterprise/server/usage/usage.go
@@ -83,19 +83,31 @@ func NewFlushLock(env environment.Env) interfaces.DistributedLock {
 	return redisutil.NewWeakLock(env.GetDefaultRedisClient(), redisUsageLockKey, redisUsageLockExpiry)
 }
 
+type TrackerOpts struct {
+	// Region identifies the datacenter in which the Redis deployment is running.
+	// Usage is tracked separately per Redis deployment.
+	Region string
+}
+
 type tracker struct {
-	env   environment.Env
-	rdb   *redis.Client
-	clock timeutil.Clock
+	env    environment.Env
+	rdb    *redis.Client
+	clock  timeutil.Clock
+	region string
 
 	flushLock interfaces.DistributedLock
 	stopFlush chan struct{}
 }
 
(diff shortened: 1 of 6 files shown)


enterprise/server/usage/usage.go
```diff
@@ -83,19 +83,31 @@ func NewFlushLock(env environment.Env) interfaces.DistributedLock {
...
 
 	flushLock interfaces.DistributedLock
 	stopFlush chan struct{}
 }
 
-func NewTracker(env environment.Env, clock timeutil.Clock, flushLock interfaces.DistributedLock) *tracker {
+func NewTracker(env environment.Env, clock timeutil.Clock, flushLock interfaces.DistributedLock, opts *TrackerOpts) *tracker {
+	region := opts.Region
+	if region == "" {
+		log.Warningf("Region is not configured; usage may not be tracked properly in multi-region deployments.")
```
tylerwilliams, line 105: do you want to write data in this scenario?

enterprise/server/usage/usage_test.go
```diff
@@ -339,9 +344,52 @@ func TestUsageTracker_Flush_ConcurrentAccessAcrossApps(t *testing.T) {
...
 			GroupID:         "GR1",
+			Region:          "us-west1",
 			PeriodStartUsec: timeutil.ToUsec(usage1Start),
 			FinalBeforeUsec: timeutil.ToUsec(usage1Collection3Start),
 			UsageCounts:     tables.UsageCounts{CASCacheHits: 1001},
 		},
 	}, usages)
 }
+
+func TestUsageTracker_Flush_CrossRegion(t *testing.T) {
```
tylerwilliams, line 355: nice

tylerwilliams (approved):
LGTM

enterprise/server/usage/usage.go
bduffany: Probably not -- turned this into a fatal error in `main.go` instead.

PR #1067, opened 2021-09-29
Add config option to block executor registration on warmup.

vadimberezniker:
Also removed container creation from warmup since we now know it's not
necessary.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -302,11 +302,14 @@ func main() {
 	}()
 	go func() {
 		if executorConfig.StartupWarmupMaxWaitSecs != 0 {
+			warmupMaxWait := time.Duration(executorConfig.StartupWarmupMaxWaitSecs) * time.Second
 			select {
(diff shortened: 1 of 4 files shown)


tylerwilliams (approved):
LGTM

enterprise/server/cmd/executor/executor.go
```diff
@@ -293,7 +294,21 @@ func main() {
...
+	warmupDone := make(chan struct{})
+	go func() {
+		executionServer.Warmup()
+		close(warmupDone)
+	}()
+	go func() {
+		if executorConfig.StartupWarmupMaxWaitSecs != 0 {
+			select {
+			case <-warmupDone:
+			case <-time.After(time.Duration(executorConfig.StartupWarmupMaxWaitSecs) * time.Second):
```
tylerwilliams, line 307: what happens in this case?

vadimberezniker: We allow the startup to proceed without waiting any longer.

PR #1068, opened 2021-09-29
Set a keep-alive enforcement policy

siggisim:
Default is 5 min: https://pkg.go.dev/google.golang.org/grpc/keepalive#EnforcementPolicy

We set it to 10s to avoid sending: 
```
error code: ENHANCE_YOUR_CALM Received Goaway too_many_pings
```

---

**Version bump**: Patch

### DIFF
--- a/server/util/grpc_server/grpc_server.go
+++ b/server/util/grpc_server/grpc_server.go
@@ -83,8 +83,9 @@ func CommonGRPCServerOptions(env environment.Env) []grpc.ServerOption {
 		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
 		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
 		grpc.MaxRecvMsgSize(env.GetConfigurator().GetGRPCMaxRecvMsgSizeBytes()),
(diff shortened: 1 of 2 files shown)


server/util/grpc_server/grpc_server.go
```diff
@@ -82,5 +83,9 @@ func CommonGRPCServerOptions(env environment.Env) []grpc.ServerOption {
 		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
 		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
 		grpc.MaxRecvMsgSize(env.GetConfigurator().GetGRPCMaxRecvMsgSizeBytes()),
+		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
```
tylerwilliams, line 86: can you add a comment here about why this is set (that includes the error / ENHANCE_YOUR_CALM strings?)

tylerwilliams (approved):
LGTM

server/util/grpc_server/grpc_server.go
siggisim: Done

PR #1070, opened 2021-09-30
Add a primitive buffer for Redis counter values

bduffany:
This will be used to help reduce load on Redis; counter values will be collected via this buffer and periodically flushed to Redis at regular intervals.

This util on its own does not actually flush the values to Redis; the caller will be responsible for that.

The plan to integrate this with the app looks like this:
* Create one of these buffers and set them on the env
* In `main`, call `Flush` every 1s or so (will experiment in dev using some artificial invocations that generate a lot of INCRBY commands to see what a good value is for this)
* Update `RedisMetricsCollector` and `UsageTracker` to go through this buffer instead of writing directly to Redis
* In the build event handler, wait a few seconds at the end of the invocation (in the background) before writing invocation stats to the table, to give enough time for all buffers to be flushed. This means we'll need another write to the invocation table
  * This is not perfect, but probably better than what we have now (only writing 1/16 of the cache stats)
  * In the future we might want to consider some fancier synchronization protocol to avoid this delay and extra DB write

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/util/redisutil/redisutil.go
+++ b/enterprise/server/util/redisutil/redisutil.go
@@ -98,12 +98,14 @@ func (r *redlock) Unlock(ctx context.Context) error {
 	return err
 }
 
-// commandBuffer buffers and aggregates Redis commands in-memory. The buffer
+// CommandBuffer buffers and aggregates Redis commands in-memory. The buffer
 // This is useful for reducing the load placed on Redis, with a tradeoff that
 // data is not updated immediately and needs to be explicitly flushed.
 //
+// Callers should be careful
+//
 // It is safe for concurrent access.
-type commandBuffer struct {
+type CommandBuffer struct {
 	rdb *redis.Client
 
 	mu sync.Mutex // protects all fields below
@@ -118,13 +120,13 @@ type commandBuffer struct {
 }
 
 // NewCommandBuffer creates a buffer.
-func NewCommandBuffer(rdb *redis.Client) *commandBuffer {
-	c := &commandBuffer{rdb: rdb}
+func NewCommandBuffer(rdb *redis.Client) *CommandBuffer {
+	c := &CommandBuffer{rdb: rdb}
 	c.init()
 	return c
 }
 
-func (c *commandBuffer) init() {
+func (c *CommandBuffer) init() {
 	c.incr = map[string]int64{}
 	c.hincr = map[string]map[string]int64{}
 	c.sadd = map[string]map[interface{}]struct{}{}
@@ -133,7 +135,7 @@ func (c *commandBuffer) init() {
 
 // SAdd adds an SADD operation to the buffer. All buffered members of the set
 // will be added in a single SADD command.
-func (c *commandBuffer) SAdd(key string, members ...interface{}) {
+func (c *CommandBuffer) SAdd(key string, members ...interface{}) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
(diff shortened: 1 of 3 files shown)


enterprise/server/util/redisutil/redisutil.go
```diff
@@ -96,3 +97,141 @@ func (r *redlock) Unlock(ctx context.Context) error {
...
+
+	s, ok := c.sadd[key]
+	if !ok {
+		s = map[interface{}]struct{}{}
+		c.sadd[key] = s
+	}
+	for _, m := range members {
+		s[m] = struct{}{}
+	}
+}
```
tylerwilliams, line 181: set, ok := c.sadd[key]
	if !ok {
		set = make(map[interface{}]struct{}{})
		c.sadd[key] = set
	}
	for _, m := range members {
		set[m] = struct{}{}
	}
}

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -96,3 +97,141 @@ func (r *redlock) Unlock(ctx context.Context) error {
...
+
+// Expire adds an EXPIRE operation to the buffer, overwriting any previous
+// expiry currently buffered for the given key. Note that the expiry duration
+// will apply only from the time that the buffer is flushed. This implies that
+// the expiration does not affect any buffered keys.
```
tylerwilliams, line 186: "This implies that the expiration does not affect any buffered keys."

Is this true? Since you apply EXPIRE commands last, could a caller of this class do a hincrby, then expire, on the same key and have that work?

i'm just suggesting you remove this part of the comment, unless it's true and i'm misunderstanding!

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -96,3 +97,141 @@ func (r *redlock) Unlock(ctx context.Context) error {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.expire[key] = duration
+}
+
+// Flush flushes the buffered commands to Redis. The commands needed to flush
+// the buffers are issued in serial. If a command fails, the error is returned
+// and the rest of the commands in the series are not executed, and their data
+// is dropped from the buffer.
```
tylerwilliams, line 197: at first this struck me as wrong, but I like how simple it is, and in practice these things are not failing ever, right?

So props for simplicity.

If it's an issue in the future, i'd suggest logging errors and returning nil, or returning the most recent error or something simple like that if you hit one.

tylerwilliams (approved):
LGTM w/ a couple of nits

enterprise/server/util/redisutil/redisutil.go
bduffany: Reworded a bit -- trying to convey that the buffer does not actively apply the expirations, and because expirations overwrite the previous ones, some sequences of commands won't behave the same as if they were issued directly to Redis.

enterprise/server/util/redisutil/redisutil.go
bduffany: Yeah I figure since we can tolerate some occasional losses in cache stats or usage numbers, we can just log any errors that occur for now, and if we see issues happening frequently enough, then it's probably better to address the root cause of those errors rather than trying to safeguard against them here (which may just add more complexity and have problems of its own)

enterprise/server/util/redisutil/redisutil.go
bduffany: Done.

PR #1071, opened 2021-09-30
Clean up redis_cache.go

bduffany:
Moving the `MetricsCollector` and `KeyValueStore` implementations out of `redis_cache` and into their own packages, since these are unrelated to CAS and AC.

With the upcoming changes to buffer Redis counters in-memory, this should let us avoid adding more unrelated stuff to `redis_cache`, since the buffering will only be needed for the `MetricsCollector` implementation.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/redis_cache/redis_cache.go
+++ b/enterprise/server/backends/redis_cache/redis_cache.go
@@ -21,7 +21,7 @@ import (
 
 const (
 	defaultCutoffSizeBytes = 10000000
-	ttl                    = 259200 * time.Second // 3 days
+	ttl                    = 3 * 24 * time.Hour
 )
 
 var (
--- a/enterprise/server/backends/redis_cache/redis_cache.go
+++ b/enterprise/server/backends/redis_cache/redis_cache.go
@@ -329,23 +329,3 @@ func (c *Cache) Start() error {
 func (c *Cache) Stop() error {
 	return nil
 }
-
-func (c *Cache) SetByKey(ctx context.Context, key string, val []byte) error {
-	if val == nil {
-		c.rdb.Del(ctx, key)
-		return nil
-	}
-	return c.rdbSet(ctx, key, val)
-}
-
-func (c *Cache) GetByKey(ctx context.Context, key string) ([]byte, error) {
-	return c.rdbGet(ctx, key)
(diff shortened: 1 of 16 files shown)


enterprise/server/backends/redis_cache/redis_cache.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	defaultCutoffSizeBytes = 10000000
-	ttl                    = 259200 * time.Second // 3 days
+	ttl                    = 3 * 24 * time.Hour
```
tylerwilliams, line 24: lol, thank you :)

server/backends/redis_key_val_store/redis_key_val_store.go
```diff
@@ -0,0 +1,40 @@
+package redis_key_val_store
```
tylerwilliams, line 1: how bout redis_kvstore? (same for memory_kvstore)

server/backends/redis_key_val_store/redis_key_val_store.go
```diff
@@ -0,0 +1,40 @@
...
+	rdb *redis.Client
+}
+
+func New(rdb *redis.Client) *store {
+	return &store{rdb}
+}
+
+func (s *store) Set(ctx context.Context, key string, val []byte) error {
+	if val == nil {
+		s.rdb.Del(ctx, key)
```
tylerwilliams, line 25: does this return an error that you can return?

tylerwilliams (changes requested):

enterprise/server/backends/redis_cache/redis_cache.go
bduffany: :D

server/backends/redis_key_val_store/redis_key_val_store.go
bduffany: Done

server/backends/redis_key_val_store/redis_key_val_store.go
bduffany: Yep, done

tylerwilliams (approved):
LGTM

PR #1073, opened 2021-10-03
Allow node/attr information to be cached by the kernel.

vadimberezniker:
This improves performance of actions that interact with many
files/directories by significantly reducing the # of kernel -> FUSE
interactions.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1074, opened 2021-10-04
Redirect to main login page (with SSO) instead of using interstitial invocation login page

siggisim:
The interstitial page doesn't support SSO, and keeping the two login pages in sync is probably not worth the hassle.

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_not_found.tsx
+++ b/app/invocation/invocation_not_found.tsx
@@ -21,19 +21,7 @@ export default class InvocationNotFoundComponent extends React.Component {
     const canLogin = capabilities.auth && !this.props.user;
 
     if (invocationExists && canLogin) {
-      return (
-        <div className="login-interstitial">
-          <div className="container">
-            <div className="login-box">
-              <div className="login-buttons">
-                <h2>Sign in to continue</h2>
-                <button onClick={this.handleLoginClicked.bind(this)}>Sign up for BuildBuddy</button>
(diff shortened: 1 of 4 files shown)


app/invocation/invocation_not_found.tsx
```diff
@@ -21,19 +21,7 @@ export default class InvocationNotFoundComponent extends React.Component {
...
-              <div className="login-buttons">
-                <h2>Sign in to continue</h2>
-                <button onClick={this.handleLoginClicked.bind(this)}>Sign up for BuildBuddy</button>
-                <button onClick={this.handleLoginClicked.bind(this)}>Log in to BuildBuddy</button>
-              </div>
-            </div>
-          </div>
-        </div>
-      );
+      window.location.href = "/?" + new URLSearchParams({redirect_url: window.location.href});
```
bduffany, line 24: If the user presses the back button after getting redirected does it redirect them again? (Wondering if `replaceState` is needed)

bduffany (changes requested):

app/invocation/invocation_not_found.tsx
siggisim: Hitting the back button doesn't seem to redirect you again, it seems to take you to wherever you were prior to visiting the invocation page.

bduffany (approved):

app/invocation/invocation_not_found.tsx
bduffany: kk, thanks for checking :)

PR #1075, opened 2021-10-04
Remove readback from `MetricsCollector.IncrementCount` API

bduffany:
Once we add Redis buffering, `MetricsCollector.IncrementCount` will no longer be able to return the count after incrementing, because the Redis command is only issued when flushing, which happens periodically in the background.

We could get around this by issuing a `GET` immediately after the `INCRBY` but this sort of defeats the purpose of buffering.

Instead, we can just remove the readback capability from the `Increment` function and update consumers to no longer depend on this readback capability.

The only place where we depend on this readback is in `hit_tracker.go` where we do throughput calculations. Luckily we can replace this with an equivalent calculation that doesn't rely on the readback. See this demo which shows that throughput calculations are the same: https://gist.github.com/bduffany/c5908f4664f45caf3e3aa91876dd1303

---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/hit_tracker/hit_tracker.go
+++ b/server/remote_cache/hit_tracker/hit_tracker.go
@@ -45,8 +45,6 @@ const (
 
 	CachedActionExecUsec
 
-	UploadThroughputBytesPerSecond
-	DownloadThroughputBytesPerSecond
 	// New counter types go here!
 )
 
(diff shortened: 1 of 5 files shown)


server/remote_cache/hit_tracker/hit_tracker.go
```diff
@@ -132,7 +126,7 @@ func (h *HitTracker) TrackMiss(d *repb.Digest) error {
 		metrics.CacheTypeLabel:      h.cacheTypeLabel(),
 		metrics.CacheEventTypeLabel: missLabel,
 	}).Inc()
-	_, err := h.c.IncrementCount(h.ctx, h.counterName(Miss), 1)
+	err := h.c.IncrementCount(h.ctx, h.counterName(Miss), 1)
 	return err
```
tylerwilliams, line 130: you can just return h.c.IncrementCount(h.ctx, h.counterName(Miss), 1) directly, here and below

tylerwilliams (approved):
LGTM

server/remote_cache/hit_tracker/hit_tracker.go
bduffany: Done

PR #1077, opened 2021-10-04
Fix divide-by-0 issue in hit tracker

bduffany:
Happens when there are no CAS hits.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1078, opened 2021-10-04
Write cache stats to the DB in the background, after waiting for Redis buffers to be flushed

bduffany:
We're planning to update the implementation of `RedisMetricsCollector` to use the newly introduced `RedisCommandBuffer`, which allows buffering and aggregating Redis commands in memory to reduce the load placed on Redis.

Since this buffer is only flushed periodically, we need to wait until the Redis buffers are flushed from all apps to Redis, before copying the stats from Redis to the DB.

For simplicity's sake, we just wait for a short delay for the buffers to be flushed. This approach has a downside that if apps or Redis are under a very high load, we might miss out on some data. Separately, we can alert on the time between flush intervals so that we can know whether this is actually a problem in practice.

---

**Version bump**: Patch

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -44,6 +44,14 @@ import (
 
 const (
 	defaultChunkFileSizeBytes = 1000 * 100 // 100KB
+
+	// The time allowed for the cache hit trackers across all app instances to
+	// flush their stats from their local buffer to the backing storage (Redis).
+	// We wait this long before we write the invocation's cache stats to the DB.
+	cacheStatsFinalizationDelay = 500 * time.Millisecond
+
+	// How long to wait before giving up on writing cache stats to the DB.
+	writeCacheStatsTimeout = 30 * time.Second
 )
 
 var (
@@ -184,11 +192,12 @@ func (e *EventChannel) MarkInvocationDisconnected(ctx context.Context, iid strin
 	}
 
 	ti := tableInvocationFromProto(invocation, iid)
-	if cacheStats := hit_tracker.CollectCacheStats(e.ctx, e.env, iid); cacheStats != nil {
-		fillInvocationFromCacheStats(cacheStats, ti)
-	}
 	recordInvocationMetrics(ti)
-	return e.env.GetInvocationDB().InsertOrUpdateInvocation(ctx, ti)
+	if err := e.env.GetInvocationDB().InsertOrUpdateInvocation(ctx, ti); err != nil {
+		return err
+	}
+	go e.writeCacheStatsToDBWhenReady(context.Background(), iid)
+	return nil
 }
 
 func fillInvocationFromCacheStats(cacheStats *capb.CacheStats, ti *tables.Invocation) {
@@ -257,13 +266,14 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 	}
 
 	ti := tableInvocationFromProto(invocation, iid)
-	if cacheStats := hit_tracker.CollectCacheStats(e.ctx, e.env, iid); cacheStats != nil {
-		fillInvocationFromCacheStats(cacheStats, ti)
-	}
 	recordInvocationMetrics(ti)
 	if err := e.env.GetInvocationDB().InsertOrUpdateInvocation(e.ctx, ti); err != nil {
 		return err
 	}
+
+	// Wait for stats collected by the hit tracker to be flushed to the
+	// backing storage medium, then copy the stats to the DB.
+	go e.writeCacheStatsToDBWhenReady(context.Background(), iid)
 
 	// Notify our webhooks, if we have any.
 	for _, hook := range e.env.GetWebhooks() {
@@ -284,6 +294,22 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 		}()
 	}
 	return nil
+}
+
+// writeCacheStatsToDBWhenReady waits long enough for all apps to flush their
+// cache stats to the DB, then updates the invocation cache stats in the DB.
+func (e *EventChannel) writeCacheStatsToDBWhenReady(ctx context.Context, invocationID string) {
+	ctx, cancel := context.WithTimeout(ctx, writeCacheStatsTimeout)
+	defer cancel()
+
+	time.Sleep(cacheStatsFinalizationDelay)
+	ti := &tables.Invocation{InvocationID: invocationID}
+	if cacheStats := hit_tracker.CollectCacheStats(context.Background(), e.env, invocationID); cacheStats != nil {
+		fillInvocationFromCacheStats(cacheStats, ti)
+	}
+	if err := e.env.GetInvocationDB().InsertOrUpdateInvocation(ctx, ti); err != nil {
+		log.Errorf("Failed to write cache stats for invocation: %s", err)
+	}
 }
 
 func (e *EventChannel) HandleEvent(event *pepb.PublishBuildToolEventStreamRequest) error {


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -286,6 +294,23 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 	return nil
 }
 
+// writeCacheStatsToDBWhenReady waits long enough for all apps to flush their
+// cache stats from their in-memory buffers, then updates the invocation cache
+// stats in the DB.
+func (e *EventChannel) writeCacheStatsToDBWhenReady(ctx context.Context, invocationID string) {
+	ctx, cancel := context.WithTimeout(ctx, writeCacheStatsTimeout)
+	defer cancel()
```
tylerwilliams, line 302: I think it would be clearer if you set the timeout on the context *before* passing it to writeCacheStatsToDBWhenReady -- think of it as sort of a lease with a timeout and then you give that lease to the goroutine, rather than having the goroutine create one itself

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -184,11 +192,12 @@ func (e *EventChannel) MarkInvocationDisconnected(ctx context.Context, iid strin
...
-	if cacheStats := hit_tracker.CollectCacheStats(e.ctx, e.env, iid); cacheStats != nil {
-		fillInvocationFromCacheStats(cacheStats, ti)
-	}
 	recordInvocationMetrics(ti)
-	return e.env.GetInvocationDB().InsertOrUpdateInvocation(ctx, ti)
+	if err := e.env.GetInvocationDB().InsertOrUpdateInvocation(ctx, ti); err != nil {
+		return err
+	}
+	go e.writeCacheStatsToDBWhenReady(context.Background(), iid)
+	return nil
```
tylerwilliams, line 200: Maybe a better approach here would be to throw these invocation ids down a channel, and have something on the other end handling them once they are ready. This would allow you to also register a graceful shutdown function that closed and drained this queue, so you know these get processed when a server restarts, rather than possibly dropped.

tylerwilliams (changes requested):

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Hoisted this piece out of the function, but still creating the context inside the goroutine func so that we can still use the `defer cancel()` idiom. Not ideal, but I think that should at least make it more obvious that the goroutine has a definite timeout?

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Not sure I understand how these would get dropped, or why the risk of being dropped is lower with the queue + worker approach? Once the goroutine is started it can't be canceled just because the server is shutting down right?

server/build_event_protocol/build_event_handler/build_event_handler.go
tylerwilliams: You're right, since these outlive the lifetime of the request they are difficult to cancel. My suggestion missed that!

server/build_event_protocol/build_event_handler/build_event_handler.go
tylerwilliams: You said it! It can and will be cancelled as soon aS the main function ends, which is exactly why you want to put these in a channel. Then you can have a single goroutine that works on these, and you can call RegisterShutdownFunction and drain that queue, which will keep main from exiting until you've finished.


One way to think about what you have now is a go routine leak. At high qps you're going to be spinning up hundreds of these and taking up n*qps resources. If you put them in a queue instead you'll have lighter resource usage and better behavior during shutdown / rollout.

bduffany: Done.

server/libmain/libmain.go
```diff
@@ -182,7 +182,15 @@ func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChec
...
-	realEnv.SetBuildEventHandler(build_event_handler.NewBuildEventHandler(realEnv))
+	beHandler := build_event_handler.NewBuildEventHandler(realEnv)
+	realEnv.SetBuildEventHandler(beHandler)
+
+	sr := beHandler.GetStatsRecorder()
+	sr.Start()
+	realEnv.GetHealthChecker().RegisterShutdownFunction(func(ctx context.Context) error {
+		sr.Stop()
+		return nil
+	})
```
tylerwilliams, line 193: you have a reference to realEnv inside of build_event_handler -- would it be cleaner to do this inside of NewBuildEventHandler()?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -81,6 +102,102 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
 		hasReceivedStartedEvent: false,
 		eventsBeforeStarted:     make([]*inpb.InvocationEvent, 0),
 		logWriter:               logWriter,
+		onClose:                 onClose,
+	}
+}
+
+func (b *BuildEventHandler) GetStatsRecorder() *statsRecorder {
+	return b.statsRecorder
+}
```
tylerwilliams, line 111: nit: rm

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -81,6 +102,102 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
...
+	defer r.mu.Unlock()
+
+	r.stopped = true
+	close(r.reqs)
+
+	// Wait for all workers to be done recording stats.
+	for i := 0; i < numStatsRecorderWorkers; i++ {
+		<-r.workerDone
 	}
 }
```
tylerwilliams, line 202: I think you can simplify a fair bit of this code by using an errorgroup instead.

func (r *statsRecorder) Start() {
  r.eg := errgroup.Group{}
	for i := 0; i < numStatsRecorderWorkers; i++ {
		eg.Go(func() error {
            for req := range r.reqs {
			// Apply the finalization delay relative to when the invocation was marked
			// finalized, rather than relative to now. Otherwise each worker would be
			// unnecessarily throttled.
			time.Sleep(time.Until(req.requestedAt.Add(cacheStatsFinalizationDelay)))
			ti := &tables.Invocation{InvocationID: req.invocationID}
			if stats := hit_tracker.CollectCacheStats(ctx, r.env, req.invocationID); stats != nil {
				fillInvocationFromCacheStats(stats, ti)
			}
			if err := r.env.GetInvocationDB().InsertOrUpdateInvocation(ctx, ti); err != nil {
				log.Errorf("Failed to write cache stats for invocation: %s", err)
			}
		}
        })
	}
}

func (r *statsRecorder) Stop() {
	// Wait for all EventHandler channels to be closed to ensure there will be
	// no more calls to MarkFinalized.
	r.openChannels.Wait()

	r.mu.Lock()
	defer r.mu.Unlock()

	r.stopped = true
	close(r.reqs)
    r.eg.Wait()
}

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -81,6 +102,102 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
...
+			"stats_recorder_finalize_after_shutdown",
+			"Invocation %q was marked finalized after the stats recorder was shut down.",
+			invocationID)
+	}
+
+	req := &recordStatsRequest{
+		invocationID: invocationID,
+		requestedAt:  time.Now(),
+	}
+	r.reqs <- req
```
tylerwilliams, line 156: This will block if r.reqs is full. You should explicitly handle that case, maybe by logging an warning and dropping the invocationID?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -81,6 +102,102 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
...
+
+func (b *BuildEventHandler) GetStatsRecorder() *statsRecorder {
+	return b.statsRecorder
+}
+
+// recordStatsRequest is a request to record the stats for an invocation, which
+// will be fulfilled by one of the workers spawned by StatsRecorder.
+type recordStatsRequest struct {
+	invocationID string
+	requestedAt  time.Time
```
tylerwilliams, line 117: nit: can you add a comment clarifying what requestedAt means?

tylerwilliams (changes requested):

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Since this should only ever happen when the DB is under such a high load that we can't flush stats fast enough, the blocking behavior here seems roughly equivalent to the old behavior -- `FinalizeInvocation` would effectively block until the DB load lightened up and the invocation row was able to be written. And since the behavior is more or less the same either way, I'm leaning towards blocking rather than dropping the stats, since dropping the stats seems worse for the user. wdyt?

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Nice - done!

server/libmain/libmain.go
bduffany: Wasn't sure if this would cause problems for testing, but looks like it works just fine -- done

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Done

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Done (and renamed to createdAt)

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Resolved offline -- decided that since this blocks all builds from finishing and there might be other unexpected reasons why the buffer fills up, we'd rather just log an error.

PR #1080, opened 2021-10-05
Have the vfs file system retrieve layout from the vfs server.

vadimberezniker:
This change removes the need to propagate the same layout information
directly to both the vfs server and the fuse filesystem.

Additionally, the VFS server now takes existing files into account when
creating the layout to make it possible to reuse workspaces.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1083, opened 2021-10-06
Surface auth error in GetUser when the user is not found

bduffany:
This was useful for helping debug the "User already exists" error ("User already exists" was returned by `CreateUser`, which is only called when `GetUser` returns `Unauthenticated` -- in this case, it was returning `Unauthenticated` unexpectedly, so it was helpful to be able to see the root cause by looking at the response in the Network tab)

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1087, opened 2021-10-06
Use sql null types for group fields that can be null

bduffany:
When scanning into individual struct fields (e.g. with `db.Raw(someJoinQuery).Rows().Scan(&t1.Field1, &t2.Field2)`, GORM will give an error saying that it failed to convert NULL to string (or bool, etc.). We have NULLs in the DB in some places where we've added new fields but didn't mark them "not null" or give them default values when adding them initially.

---

**Version bump**: Patch

### DIFF
--- a/server/buildbuddy_server/buildbuddy_server.go
+++ b/server/buildbuddy_server/buildbuddy_server.go
@@ -2,6 +2,7 @@ package buildbuddy_server
 
 import (
 	"context"
+	"database/sql"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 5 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -256,7 +257,7 @@ func (s *BuildBuddyServer) CreateGroup(ctx context.Context, req *grpb.CreateGrou
 		Name:                   groupName,
 		OwnedDomain:            groupOwnedDomain,
 		SharingEnabled:         req.GetSharingEnabled(),
-		UseGroupOwnedExecutors: req.GetUseGroupOwnedExecutors(),
+		UseGroupOwnedExecutors: sql.NullBool{Valid: true, Bool: req.GetUseGroupOwnedExecutors()},
```
bduffany, line 260: To me it seems slightly error prone that we have to provide `Valid: true` when using these null types, although it is more convenient to get the values out of them (no need to check `!= nil` which is nice).

What do people think about using raw pointers instead? I think those are slightly error prone as well (it's possible to accidentally de-reference `nil` pointers) but I wonder if checking `nil` before de-referencing is more instinctual than remembering to pass `Valid: true`.

PR #1092, opened 2021-10-07
Roll-forward #1053: Add group roles to UserInfo interface

bduffany:
This is a clean roll-forward except for commit 63ba3fab477597a143ce74ada9c76ec952646fa7 and commit b64fc44c5d5ad36f8fba6707eb71b724e6070f3c

This is safe to roll forward now that we've fixed the error caused by attempting to convert NULL to string when scanning (in #1087)

---

**Version bump**: Patch


bduffany:
Accidentally merged this into the wrong branch -- will merge into master manually offline.

PR #1119, opened 2021-10-14
Escape `Groups` table name for newer MySQL versions

siggisim:
---

**Version bump**: Patch


siggisim:
cc @vadimberezniker and @tempoz - so we're all on the look-out for this in code reviews.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/914

siggisim:
Warning @tylerwilliams - auto-merge is enabled so don't approve if you have comments.

PR #1130, opened 2021-10-18
Bump golang version to 1.17.2

tempoz:
---

**Version bump**: Patch


tempoz:
Yeah, `go mod tidy` is very insistent about them. I should probably read and see if I can determine why.

tempoz:
It looks like indirect dependencies that aren't manually added go in a separate require block from everything else: https://github.com/golang/go/issues/27887

PR #1131, opened 2021-10-18
Use approximate LRU in disk cache.

tylerwilliams:
Plumb the approximate LRU into the disk cache. This CL modifies the disk cache so that only the hashes of filepaths are tracked. To perform evictions, the approximate LRU samples files randomly from directories, considers them in context with other samples, and evicts the oldest (deletes it from disk).

Performance wise, I am able to get approximately 2800 write QPS out of a full cache, which is on-par with the performance before this change.

### DIFF
--- a/server/backends/disk_cache/disk_cache.go
+++ b/server/backends/disk_cache/disk_cache.go
@@ -322,44 +322,88 @@ type partition struct {
 	randomSampleCh   chan *fileRecord
 }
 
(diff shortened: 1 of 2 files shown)


tylerwilliams:
This CL reduces memory use and GC but may not be necessary given that we recently moved apps to bigger machines. It is fine to review but I do not plan to submit it at this time.

PR #1141, opened 2021-10-20
Add regression test for adding a user to an empty group

bduffany:
---

**Version bump**: Patch


bduffany:
@tylerwilliams I don't think so -- I will see about adding a `buildbuddy_server_test.go`

PR #1147, opened 2021-10-21
Use a consistent view of platform props everywhere [POC]

bduffany:
Shows the overall effort needed to respect `--remote_header=x-buildbuddy-platform.KEY=VALUE` everywhere. In particular, this should fix the problem that overriding `pool` via this method does not currently work.

If it seems worth it, this could be split into smaller PRs with their own tests.

---

**Version bump**: TODO


tylerwilliams:
Just skimmed this -- why is a whole new RPC required?

PR #1157, opened 2021-10-22
Properly account for the executable bit in the filecache and workspace re-use

siggisim:
Currently the filecache assumes that files with the same digest can be used interchangeably.

This isn't exactly true, because the same file can be used in different actions with different values for the executable bit. This seems to be particularly problematic in `AndroidResourceLink` actions which expects `androidsdk/aapt2_binary` to have the executable bit set, which doesn't always seem to be the case.

This change updates the filecache to work with `repb.FileNode` rather than `repb.Digest`.

`repb.FileNode` is effectively just a struct that holds the digest + the IsExecutable bit.

With this change, executable files in the filecache are now stored with a `.executable` suffix - while non-executable files remain as-is. This will ensure backwards-compatibility and that only executable files will need to be re-downloaded on executors that already have a populated filecache.

This works around the limitation that hard-linked files cannot have different permissions.

This also fixes the case where during workspace re-use we "skip" existing files with the wrong permissions when their digests match.

---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/cachetools/cachetools.go
+++ b/server/remote_cache/cachetools/cachetools.go
@@ -552,6 +552,6 @@ func UploadProtoToAC(ctx context.Context, cache interfaces.Cache, instanceName s
 	return uploadProtoToCache(ctx, ac, instanceName, in)
 }
 
-func isExecutable(info *os.FileInfo) bool {
+func isExecutable(info os.FileInfo) bool {
 	return info.Mode()&0100 != 0
 }


bduffany:
Nice, seems like this might fix that go-protoc-bin issue that we were seeing with workflows :thinking: 

Building off the idea of adding the file suffix for backwards compatibility -- wdyt about encoding the whole file mode (in addition to the executable bit) as a string, and using that as the suffix? That way it'd fix any potential issues related to any of the permissions bits (https://github.com/buildbuddy-io/buildbuddy-internal/issues/472), in addition to the executable bit

PR #1179, opened 2021-10-28
Use parsed platform props for task sizing, and add tests

bduffany:
Before making any changes to task sizing / container sizing to support Mac workflows, clean up task sizing logic and also respect remote header overrides by going through `platform.ParseProperties`.

---

**Version bump**: Patch


enterprise/server/remote_execution/platform/platform.go
```diff
@@ -59,6 +60,10 @@ const (
 	// Using the property defined here: https://github.com/bazelbuild/bazel-toolchains/blob/v5.1.0/rules/exec_properties/exec_properties.bzl#L164
 	dockerRunAsRootPropertyName = "dockerRunAsRoot"
 
+	// A BuildBuddy Compute Unit is defined as 1 cpu and 2.5GB of memory.
+	EstimatedComputeUnitsPropertyName = "EstimatedComputeUnits"
+	EstimatedFreeDiskPropertyName     = "EstimatedFreeDiskBytes"
```
tylerwilliams, line 65: I don't follow what this is for based on the name. (Why would a client need to tell us the estimated free disk of a task?)

Can you add a comment?

bduffany: @siggisim do you know the context for the name of this prop?

siggisim: I think this was added by @vadimberezniker in this PR https://github.com/buildbuddy-io/buildbuddy/commit/0536ec1548e377a815bd7a5f62e4480428a90176#diff-9b6312e0c353cc319b101e1cb71608ed26be804af9e31d0b1c6d5ebe5552992eR16

vadimberezniker: This is used to specify how much "scratch space" beyond the input files a task requires to be able to function.
I added it this to support Managed Bazel + Firecracker because for Firecracker we need to decide ahead of time how big the workspace filesystem is going to be and Managed Bazel requires a relatively large amount of free space compared to typical actions.

tylerwilliams: SG, @bduffany can you add a comment to the property const?

bduffany: @tylerwilliams Done

tylerwilliams (approved):
LGTM

PR #1181, opened 2021-10-28
Remove outdated comment specifying blobstore error types

tempoz:
- comments
- add error check
- Remove outdated comment specifying blobstore error types

The comment specifying that blobstores return os-type errors is both untrue and no longer a design goal, so remove it.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1183, opened 2021-10-28
Make the role filter work for gRPC requests

bduffany:
---

**Version bump**: Patch


server/rpc/filters/filters.go
```diff
@@ -120,6 +122,20 @@ func authUnaryServerInterceptor(env environment.Env) grpc.UnaryServerInterceptor
 	return contextReplacingUnaryServerInterceptor(ctxFn)
 }
 
+func roleAuthUnaryServerInterceptor(env environment.Env) grpc.UnaryServerInterceptor {
```
tylerwilliams, line 125: no streaming filter is necessary? If we added a streaming RPC would it be unprotected?

server/role_filter/role_filter.go
```diff
@@ -38,6 +36,8 @@ var (
 		// don't require a group role.
 		"GetBazelConfig",
 		// API calls are role independent
+		// TODO(bduffany): prefix all of these with the service name,
```
tylerwilliams, line 39: good idea

enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
```diff
@@ -666,7 +668,10 @@ func (r *Env) newTestAuthenticator() *testauth.TestAuthenticator {
 	users[ExecutorAPIKey] = &testauth.TestUser{
 		GroupID:       ExecutorGroup,
 		AllowedGroups: []string{ExecutorGroup},
-		Capabilities:  []akpb.ApiKey_Capability{akpb.ApiKey_REGISTER_EXECUTOR_CAPABILITY},
+		// TODO(bduffany): Don't give this API key admin rights, since API keys
+		// don't actually have admin rights in real apps.
+		GroupMemberships: []*interfaces.GroupMembership{{GroupID: ExecutorGroup, Role: role.Admin}},
```
tylerwilliams, line 673: I'm having trouble squaring this comment (don't give admin rights) with the line below it: Role: admin. Could you elaborate on this a little bit?

tylerwilliams (changes requested):

vadimberezniker (approved):
Don't have anything to add beyond what's already covered by Tyler's comments.

server/rpc/filters/filters.go
bduffany: Good point, added a streaming filter too.

enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
bduffany: Reworded the comment to clarify. This is just a TODO to change `role.Admin` to `role.Default`, since this `TestUser` is actually representing an API key, and currently API keys cannot have `role.Admin`.

Addressing the TODO in this PR would take a bit of refactoring since the only way to get admin role currently is to authenticate over HTTP via protolet, and the RBE test env is not (yet?) set up for this.

tylerwilliams (approved):
LGTM

PR #1185, opened 2021-10-29
Add grpc keepalive to common options and restore failfast behaviour.

tylerwilliams:
WIP

### DIFF
--- a/enterprise/server/util/cacheproxy/cacheproxy.go
+++ b/enterprise/server/util/cacheproxy/cacheproxy.go
@@ -129,6 +129,7 @@ func (c *CacheProxy) getClient(ctx context.Context, peer string) (dcpb.Distribut
 		// failfast setting. This can happen when a replica goes away due to a rollout (or any other reason).
 		isReady := client.conn.GetState() == connectivity.Ready
 		if client.wasEverReady && !isReady {
+			client.conn.Connect()
 			return nil, status.UnavailableErrorf("connection to peer %q is not ready", peer)
 		}
 		client.wasEverReady = client.wasEverReady || isReady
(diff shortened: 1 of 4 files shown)


enterprise/server/util/cacheproxy/cacheproxy.go
```diff
@@ -123,6 +125,13 @@ func (c *CacheProxy) getClient(ctx context.Context, peer string) (dcpb.Distribut
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if client, ok := c.clients[peer]; ok {
```
vadimberezniker, line 127: I think you need to add logic to force a reconnect if the connection is idle.
While the keepalive changes will help normal connections from idling out, a broken connection will still put the connection nito IDLE state and it will stay there because we won't send any further RPCs.

vadimberezniker (approved):

enterprise/server/util/cacheproxy/cacheproxy.go
tylerwilliams: Yup, done.

tylerwilliams:
@siggisim for review
@tempoz for WithKeepalive FYI

server/util/grpc_client/grpc_client.go
```diff
@@ -66,5 +68,18 @@ func CommonGRPCClientOptions() []grpc.DialOption {
 		filters.GetUnaryClientInterceptor(),
 		filters.GetStreamClientInterceptor(),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(math.MaxInt32)),
+		grpc.WithKeepaliveParams(keepalive.ClientParameters{
+			// After a duration of this time if the client doesn't see any activity it
+			// pings the server to see if the transport is still alive.
+			Time: 10 * time.Second,
```
siggisim, line 74: Might want to set this to something a little higher than 10s, like 20s - because our grpc server enforcement policy is 10s.

https://github.com/buildbuddy-io/buildbuddy/blob/393e82020e9406dada2edb4708225b010c3f42ac/server/util/grpc_server/grpc_server.go#L88

siggisim (approved):

server/util/grpc_client/grpc_client.go
tylerwilliams: Done

PR #1186, opened 2021-10-29
Wire up code UI to new runner API

siggisim:
It's rough v0 (no commit sha or patch), and the response is just dumped into an alert - but thought it'd be nice to have these wired up.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/hostedrunner/hostedrunner.go
+++ b/enterprise/server/hostedrunner/hostedrunner.go
@@ -47,7 +47,7 @@ func New(env environment.Env) (*runnerService, error) {
 	}, nil
 }
 
-func (r *runnerService) lookupAPIKey(ctx context.Context) (string, error) {
+func (r *runnerService) lookupAPIKey(ctx context.Context, selectedGroup string) (string, error) {
 	auth := r.env.GetAuthenticator()
 	if auth == nil {
 		return "", status.FailedPreconditionError("Auth was not configured but is required")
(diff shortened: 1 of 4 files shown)


enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -56,12 +56,19 @@ func (r *runnerService) lookupAPIKey(ctx context.Context) (string, error) {
...
-	if u.GetGroupID() == "" {
+	group := u.GetGroupID()
+	if selectedGroup != "" {
+		for _, membership := range u.GetGroupMemberships() {
+			if membership.GroupID == selectedGroup {
+				group = selectedGroup
+			}
+		}
+	}
+	if group == "" {
```
tylerwilliams, line 67: @bduffany do we have any auth library code that will do this for us?

tylerwilliams (approved):
LGTM

enterprise/server/hostedrunner/hostedrunner.go
siggisim: Gunna submit this so I can play around with it, but will migrate if we have a better solution for getting a group id from either an HTTP or GRPC handler.

bduffany: We don't have any library code to get the preferred group ID that works for both HTTP and GRPC requests, but I could have benefited from a function like that in https://github.com/buildbuddy-io/buildbuddy/pull/1183 so I may just add a library func as part of that one.

PR #1187, opened 2021-11-01
Add EffectiveGroup function and use it in hostedrunner

bduffany:
---

**Version bump**: Patch


server/util/authutil/authutil.go
```diff
@@ -26,3 +30,30 @@ func AuthorizeGroupRole(u interfaces.UserInfo, groupID string, allowedRoles role
 	}
 	return nil
 }
+
+// EffectiveGroup returns the group membership effective for the request
+// context, given an authenticated user.
+func EffectiveGroup(ctx context.Context, u interfaces.UserInfo) (*interfaces.GroupMembership, error) {
```
tylerwilliams, line 36: Could we call this SelectedGroup? Or were you avoiding that because for group-based-auth nothing is selected?

tylerwilliams (approved):
LGTM

server/util/authutil/authutil.go
bduffany: Yeah, I figured for group-based auth it may be confusing to call it SelectedGroup.

PR #1188, opened 2021-11-02
Improve task and container sizing for workflows

bduffany:
Task sizing changes:
- Instead of `DefaultMemEstimate` (400 MB), increase workflow task memory usage estimate to 8GB whenever we see the `workflow-id` platform prop

Runner pool changes:
- When computing runner memory usage, fall back to the task size estimate if `container.Stats()` returns `MemoryUsageBytes == 0` (i.e. for bare runners)
- When pooling containers, evict based on combined estimated memory usage of the currently pooled runners, and not just the max runner count (which is calculated as (`0.8 * env['SYS_MEMORY_BYTES'] / 2.6GB`).
  - Alternatively, we could have just configured `max_runner_count: ${max_memory / 8GB}` in the `config.yaml`, or make the base runner memory estimate configurable and set it to 8GB only on workflow executors -- but I figured it's easier from an operations perspective to avoid introducing extra config options. Open to either approach if we're OK with configuring these explicitly.

---

**Version bump**: Patch


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -168,8 +173,8 @@ func (r *CommandRunner) pullCredentials() container.PullCredentials {
 	return container.GetPullCredentials(r.env, r.PlatformProperties)
 }
 
-func (r *CommandRunner) PrepareForTask(ctx context.Context, task *repb.ExecutionTask) error {
-	r.Workspace.SetTask(task)
+func (r *CommandRunner) PrepareForTask(ctx context.Context) error {
+	r.Workspace.SetTask(r.task)
```
tylerwilliams, line 177: since the runner can be reused, it can handle multiple tasks. Are you sure you want to be resetting the task instance variable directly all the time instead of passing the task in when you call PrepareForTask? that seems like a more natural flow...?

bduffany: So, the current flow is that when you check out a runner from the pool, you have to pass in a Task, and we pick a runner that is compatible with the Task (same container-image, same remote instance name, etc.). The runner is then "bound" to that task for as long as it is checked out from the pool -- this is already implicit in how we store the task's PlatformProperties in the runner state, and how we make use of those PlatformProperties in the various lifecycle funcs.

I think it is somewhat error prone to allow passing a Task into this function, because it'd be possible to make the mistake of checking a runner out of the pool that is able to run one task, but then prepare it for a different task, and then actually run yet another task. It seems cleaner to be able to check out a runner bound to a task, and then the caller just has the responsibility of walking that one task through its lifecycle (Prepare, then Run).

tylerwilliams: Got it, sg. Thanks for explaining it!

tylerwilliams (approved):
LGTM

PR #1189, opened 2021-11-02
Always use extended context in finalize

tempoz:
Fixes the stat recorder erroneously using the old context in FinalizeInvocation, as well
as adding an option to the chunkstore to set the context for writing and then making use
of it in FinalizeInvocation.

Refactors chunkstore somewhat for this purpose, as well as for readability, and fixes a bug
that would have resulted in buffer flushes happening more often than requested.

---

**Version bump**: Patch


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -471,7 +472,7 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 		return err
 	}
 
-	e.statsRecorder.Enqueue(e.ctx, invocation)
+	e.statsRecorder.Enqueue(ctx, invocation)
```
bduffany, line 475: I think this is a good change, so this is more just intended as a note: `Enqueue` only uses the context for the purpose of inspecting its values (`x-buildbuddy-jwt`), so I think the previous code would have worked even if the ctx deadline was exceeded

server/backends/chunkstore/chunkstore.go
```diff
@@ -436,116 +446,130 @@ func (w *ChunkstoreWriter) Close() error {
 	return err
 }
 
+func (w *ChunkstoreWriter) SetContext(ctx context.Context) {
+	w.ctx = ctx
```
bduffany, line 450: If I understand this correctly, this modifies the context without acquiring a lock, while the `writeLoop` may be accessing it concurrently. If so, there might be a more thread-safe way to do this.

Is extending the context the only thing you need this for? If so, you might be able to get away with something like this, and simplify some of the PR (I don't know if there are issues with this approach though so I would def get Tyler's opinion as well):

```
ctx = /* ... */
var extendOnce sync.Once
select {
  case <-ctx.Done():
    // Allow a one-time deadline extension for finalization.
   extendOnce.Do(func() { ctx = background.ExtendContextForFinalization(ctx) })
  if isDone(ctx) {
    // timed out
  }
}
```

server/backends/chunkstore/chunkstore.go
```diff
@@ -367,16 +369,19 @@ type WriteResult struct {
 	Err            error
 	Size           int
 	LastChunkIndex uint16
+	Timeout        bool
```
bduffany, line 372: Is this `Timeout` flag computable as `status.IsDeadlineExceeded(result.Err)` ?

bduffany (commented):
Some initial feedback on the `ctx` part -- I might need a few more passes for the bugfix parts. If it's feasible to split just the ctx changes into a separate PR, that would be super helpful in making it more obvious what the bugfixes are that are mentioned in the PR description

server/backends/chunkstore/chunkstore.go
tempoz: Nope, it tells you if the write is a flush due to timeout while waiting for enough bytes to hit the threshold to write to disk.

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: Good to know! Thank you for the info, as always ^-^

server/backends/chunkstore/chunkstore.go
tempoz: That isn't what it did; `ChunkstoreWriter` was entirely synchronous and only communicated with the writeLoop (the struct used in the goroutine) via the `WriteRequest` and `WriteResult` channels. Regardless, Tyler suggested that we just specify the context for every write, so the model has now shifted to that.

server/backends/chunkstore/chunkstore.go
```diff
@@ -388,164 +394,185 @@ type ChunkstoreWriter struct {
...
+				chunkIndex++
+			}
 		}
-		if len(l.chunk) == 0 {
+		if len(chunk) == 0 {
 			// There is no data to write; we don't need to wake to flush data.
-			l.flushTime = time.Now().Add(year)
-		} else if time.Until(l.flushTime) > (l.writeTimeoutDuration) {
+			flushTime = time.Now().Add(year)
+		} else if time.Now().Add(l.writeTimeoutDuration).Before(flushTime) || bytesFlushed > 0 {
```
tempoz, line 560: This is the bug fix; basically we were only resetting the flushTime when we flushed the whole buffer to disk, not every time we wrote. This reflows the time part of the conditional to be identical, but hopefully more readable, and adds the condition that we should reset the flush time when we write to disk.

tempoz: @bduffany

server/backends/chunkstore/chunkstore.go
```diff
@@ -388,164 +394,185 @@ type ChunkstoreWriter struct {
...
+			}
 		}
 		// We do not attempt to write if we have already encountered a write error
 		// in this iteration of the loop, and instead we simply move on.
 		// If the file is empty and we are closing it, we must write an empty chunk.
 		// Otherwise, if we have data cached to write and we are either flushing or
 		// closing the file, we need to write a chunk.
-		if err == nil && ((!l.open && l.chunkIndex == 0) || (len(l.chunk) > 0 && (time.Now().After(l.flushTime) || !l.open))) {
-			l.write()
+		if err == nil && ((!open && chunkIndex == 0) || (len(chunk) > 0 && (time.Now().After(flushTime) || !open))) {
```
bduffany, line 548: I don't understand the `chunkIndex == 0` part of this conditional, mind adding that to the comment?

Also, should `len(chunk) > 0` be required for all parts of the conditional? i.e. `err == nil && len(chunk) > 0 && ( ... )`

tempoz: I tried rewriting it, maybe it's easier to understand now?

bduffany (approved):
LGTM, nice :)

PR #1190, opened 2021-11-03
Fetch buildbuddy.yaml in the workflow service instead of the CI runner

bduffany:
Behind a flag, fetch buildbuddy.yaml as a standalone file before spawning the CI runner, and start one CI runner task per matched action, rather than spawning a single CI runner task that runs all matched actions. This is a pre-requisite for being able to specify the OS and Pool properties of the CI runner.

---

**Version bump**: Patch


enterprise/server/workflow/config/config.go
```diff
@@ -45,3 +53,53 @@ func NewConfig(r io.Reader) (*BuildBuddyConfig, error) {
 	}
 	return cfg, nil
 }
+
+// GetDefault returns the default workflow config, which tests all targets
+// when pushing to the given target branch, sending build events to BuildBuddy.
+func GetDefault(targetBranch, besBackend, besResultsURL, apiKey string) *BuildBuddyConfig {
+	return &BuildBuddyConfig{
```
tylerwilliams, line 60: why do we call this BuildBuddyConfig instead of WorkflowConfig?

enterprise/server/workflow/service/service.go
```diff
@@ -45,6 +48,8 @@ const (
 )
 
 var (
+	fetchWorkflowConfigBeforeExecution = flag.Bool("fetch_workflow_config_before_execution", false, "Whether to fetch the workflow config before executing the workflow and spawn the CI runner based on the workflow config.")
```
tylerwilliams, line 51: just checking -- do we need to flag this out and do you plan to remove the flag after this is released?

tylerwilliams (approved):
LGTM

enterprise/server/workflow/config/config.go
bduffany: No strong reason, other than that the file name is `buildbuddy.yaml` and presumably this is where all buildbuddy-related configuration would go. I can send a follow-up PR to rename it, if you think it's less confusing that way

enterprise/server/workflow/service/service.go
bduffany: Will remove this flag since we have enough time before the next release to test it, and it's just as easy to revert this behavior by pasting in a bit of code, as it is to switch the flag on / off if something goes wrong.

PR #1192, opened 2021-11-03
Enable RBE for code editor

siggisim:
Also opens invocation in a new tab rather than just alerting the response blob.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1193, opened 2021-11-03
Provision the OS-specific CI runner binary via the executor's data deps

bduffany:
Instead of uploading the CI runner to CAS and then having the executor download it, we now build the CI runner binary along with the executor build and inject the OS-specific CI runner directly from the bundle (or runfiles for development). This allows the CI runner binary to have the same architecture as the executor, which is required for Mac workflows.

Note: This does not apply to `remotebazel` yet, to avoid having to get this working with VFS. `remotebazel` currently only works on Linux anyway, so we can stick with our current strategy of uploading the Linux CI runner binary to the CAS and downloading it as an input (lazily).

---

**Version bump**: Patch


enterprise/server/remote_execution/workspace/workspace.go
```diff
@@ -146,6 +149,43 @@ func (ws *Workspace) DownloadInputs(ctx context.Context, tree *repb.Tree) (*dirt
...
+func (ws *Workspace) AddCIRunner() error {
+	destPath := path.Join(ws.Path(), "buildbuddy_ci_runner")
+	exists, err := disk.FileExists(destPath)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return nil
+	}
+	srcPath, err := bazelgo.Runfile("enterprise/server/cmd/ci_runner/ci_runner_/ci_runner")
```
tylerwilliams, line 163: see my slack comment for more details, but this won't work on our deployed macs, the runner needs to be a part of the bundle

bduffany: Done -- updated `bundle.go` to fall back to runfiles, so that we can exclude the CI runner from the bundle during development.

The CI runner adds ~20MB to the bundle, bringing the total bundle size from 81MB => 101MB (with `-c opt`)

enterprise/bundle.go
```diff
@@ -26,10 +31,34 @@ func (a *aliasFS) Open(name string) (fs.File, error) {
...
+	// If the file does not exist, attempt to read it from runfiles.
+	// Otherwise, return the original result.
+	if err != nil && os.IsNotExist(err) {
+		// Fall back to bazel runfiles. Note the limitation here that we can only get
+		// runfiles under the enterprise dir.
+		runfile, err := bazelgo.Runfile(filepath.Join("enterprise", name))
+		if err == nil {
+			return os.Open(runfile)
+		}
+	}
```
tylerwilliams, line 44: would it make sense to have a file resolver type util that looks at runfiles or the bundle (from env)? Then we could fix up all callsites that use bundle/runfiles directly to just call that util and get this behavior.

to me it seems confusing to put this here since if you're importing bundle, you expect to be reading from the bundle, not runfiles.

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -266,13 +266,19 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, task *repb.E
 	}
 
 	rxInfo := &dirtools.TransferInfo{}
-	// Don't download inputs if the FUSE-based filesystem is enabled.
+	// Don't download inputs or add the CI runner if the FUSE-based filesystem is
+	// enabled.
```
tylerwilliams, line 270: I think this breaks fuse support, since in that mode the runner won't exist at all, right? or is fuse able to map it in as is?

It's not a huge deal but we should file a bug / add a TODO to fix.

tylerwilliams (changes requested):
few comments (and i pinged you a question in slack)

enterprise/bundle.go
bduffany: Done

tylerwilliams (approved):
LGTM

tylerwilliams:
Thanks for splitting out the file resolver stuff, it really focused this CL!

PR #1194, opened 2021-11-03
Use bazel TEST_TMPDIR for tests

tempoz:
Tests were using `/tmp` for test file I/O, but bazel tests dhould use `TEST_TMPDIR`.

Redis sockets, however, must be in `/tmp`.

---

**Version bump**: None


server/testutil/testfs/testfs.go
```diff
@@ -14,19 +14,44 @@ import (
 
 // MakeTempDir creates and returns an empty directory that exists for the scope
 // of a test.
-func MakeTempDir(t testing.TB) string {
-	tmpDir, err := os.MkdirTemp("", "buildbuddy-test-*")
+func MakeTempDir(t testing.TB, dir string) string {
```
bduffany, line 17: It looks like there's three uses of the `dir` argument:

- Most callers will pass `""` to create a brand new temp dir
- One caller passes `/tmp` for creating a socket
- Other callers pass another string to create a child dir within a temp dir returned by `MakeTempDir`

For the sake of making tests easier to write, wdyt about dropping this extra arg? And then for the other use cases:

- For the socket use case, add another func like `TempSocket` which just calls `os.MkdirTemp` and then returns that path joined with `test.sock` (but doesn't actually create the file)
- For the child dir use case, I think we should rely on the cleanup func registered by the `MakeTempDir` to clean up the child dir, and then have a separate function like `testfs.MkdirAll(t testing.TB, rootDir, childPath string)` -- since the convention of the other functions in this package is to take a `rootDir` arg which should be the root test directory returned by `testfs.MakeTempDir`

tempoz: I like these suggestions a lot.

bduffany (approved):
Awesome!

server/testutil/webtester/webtester.go
```diff
@@ -77,21 +77,16 @@ func (wt *WebTester) screenshot(tag string) error {
 	// TEST_UNDECLARED_OUTPUTS_DIR is usually defined by Bazel. If this test is
 	// run outside of Bazel for whatever reason, this will just be an empty
 	// string, which is interpreted by CreateTemp as "use the OS-default temp dir"
-	testScreenshotDir := os.Getenv("TEST_UNDECLARED_OUTPUTS_DIR")
-	screenshotFile, err := os.CreateTemp(testScreenshotDir, fmt.Sprintf("%s.%s.screenshot.png", wt.t.Name(), tag))
-	if err != nil {
-		return err
-	}
-	defer screenshotFile.Close()
+	screenshotFile := testfs.MakeTempFile(wt.t, os.Getenv("TEST_UNDECLARED_OUTPUTS_DIR"), fmt.Sprintf("%s.%s.screenshot-*.png", wt.t.Name(), tag))
```
bduffany, line 80: I think we still want to use `os.CreateTemp` here -- IIUC `MakeTempFile` will delete the screenshot automatically on test cleanup

bduffany (changes requested):

server/testutil/webtester/webtester.go
bduffany: (sorry, caught this after approving)

tempoz: lol np, i'll fix it.

bduffany (approved):

PR #1195, opened 2021-11-04
Trim whitespace before normalizing/parsing repo urls

siggisim:
Should fix some issues with repo parsing we're seeing in prod

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1196, opened 2021-11-04
Improve error handling for InsertOrUpdateInvocation

bduffany:
Verified that Bazel does a retry in case we return an error here, so this should not negatively affect any existing cases where we return an error for whatever reason (e.g. due to context deadline exceeded).

---

**Version bump**: Patch

### DIFF
--- a/server/backends/invocationdb/invocationdb.go
+++ b/server/backends/invocationdb/invocationdb.go
@@ -8,7 +8,6 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
 	"github.com/buildbuddy-io/buildbuddy/server/util/db"
-	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
 	"github.com/buildbuddy-io/buildbuddy/server/util/query_builder"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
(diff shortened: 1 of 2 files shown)


server/backends/invocationdb/invocationdb.go
```diff
@@ -58,17 +57,12 @@ func (d *InvocationDB) InsertOrUpdateInvocation(ctx context.Context, ti *tables.
 	return d.h.Transaction(ctx, func(tx *db.DB) error {
 		var existing tables.Invocation
 		if err := tx.Where("invocation_id = ?", ti.InvocationID).First(&existing).Error; err != nil {
-			if db.IsRecordNotFound(err) {
-				return d.createInvocation(tx, ctx, ti)
-			}
-		} else {
-			err := tx.Model(&existing).Where("invocation_id = ?", ti.InvocationID).Updates(ti).Error
-			if err != nil {
-				log.Warningf("Error updating invocation %s: %s", ti.InvocationID, err.Error())
```
tylerwilliams, line 67: are we still logging the invocationID and err in another place?

tylerwilliams (approved):
LGTM

server/backends/invocationdb/invocationdb.go
bduffany: Yep, it manifests in the logs like this:

```
2021/11/04 13:45:46.290 WRN Disconnecting invocation "5eaa482a-5c4a-449e-84cc-d70fe02e1462": context deadline exceeded
...
2021/11/04 13:45:46.290 WRN Error finalizing invocation "5eaa482a-5c4a-449e-84cc-d70fe02e1462" during disconnect: context deadline exceeded
```

PR #1197, opened 2021-11-04
Fix bug that we finalize invocations twice in some cases

bduffany:
This fixes at least one possible cause of the AlreadyExists error (https://github.com/buildbuddy-io/buildbuddy-internal/issues/951), although there may be others (still looking into it).

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1198, opened 2021-11-04
Fix another source of AlreadyExists errors

bduffany:
It's possible that once we call `beValues.AddEvent(BuildFinished)`, we encounter an error, which would cause the invocation to terminate early and be finalized. However, because we saw the `BuildFinished` event, we'd mark the invocation with COMPLETED status, which would cause AlreadyExists to be returned on the subsequent retry.

Instead of switching on whether we saw the BuildFinished event, we now pass the status directly, basing the status on whether we're invoking FinalizeInvocation as the result of `disconnectWithErr` or as the result of successfully receiving all build events from Bazel.

Future work: Should we only mark the invocation COMPLETED after sending back all of the ACKs to Bazel? (Will Bazel attempt to retry if we don't ACK everything?)

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/951

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -435,19 +435,18 @@ func (e *EventChannel) Close() {
 	e.onClose()
 }
 
-func (e *EventChannel) FinalizeInvocation(iid string) error {
+func (e *EventChannel) FinalizeInvocation(iid string, status inpb.Invocation_InvocationStatus) error {
 	ctx, cancel := background.ExtendContextForFinalization(e.ctx, 10*time.Second)
 	defer cancel()
 
 	invocation := &inpb.Invocation{
 		InvocationId:     iid,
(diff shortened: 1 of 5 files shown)


tylerwilliams (approved):
LGTM

tempoz:
This leaves me concerned about this: https://github.com/buildbuddy-io/buildbuddy-internal/issues/875 , which was the motivation for the switch to `BuildFinished` in the first place. This bug description led me to believe that a build finished event was *the* defining factor as to whether or not an invocation is complete. Thoughts?

siggisim:
> This leaves me concerned about this: [buildbuddy-io/buildbuddy-internal#875](https://github.com/buildbuddy-io/buildbuddy-internal/issues/875) , which was the motivation for the switch to `BuildFinished` in the first place. This bug description led me to believe that a build finished event was _the_ defining factor as to whether or not an invocation is complete. Thoughts?

+1 this seems to revert the behavior that we were trying to achieve (not marking builds with a finished event as disconnected) - this probably means we have to be more lenient about the retry policy though.

PR #1200, opened 2021-11-04
Fix flaky invocation page test

bduffany:
Adding `--show_progress=0` prevents the logs from being as variable in length for invocations that happen to take longer to run (so we can do assertions more reliably on what is shown in the logs viewport).

---

**Version bump**: Patch


siggisim (approved):

tylerwilliams (approved):
LGTM

PR #1202, opened 2021-11-05
Add a fileresolver util to locate bundled files with runfiles fallback

bduffany:
Also use the fileresolver in a few places:
* firecracker.go (simplifies the logic for resolving paths to vmlinux, etc.)
* hostedrunner.go (fixes a bug that the server fails to start because we don't have runfiles available when running the enterprise app as a standalone binary)

This will also be useful for https://github.com/buildbuddy-io/buildbuddy/pull/1193

---

**Version bump**: Patch

### DIFF
--- a/server/environment/environment.go
+++ b/server/environment/environment.go
@@ -76,4 +76,5 @@ type Env interface {
 	GetUsageService() interfaces.UsageService
 	GetUsageTracker() interfaces.UsageTracker
 	GetXCodeLocator() interfaces.XcodeLocator
+	GetFileResolver() fs.FS
 }
--- a/server/environment/environment.go
+++ b/server/environment/environment.go
@@ -36,7 +36,11 @@ type Env interface {
 	// Optional dependencies below here. For example: enterprise-only things,
 	// or services that may not always be configured, like webhooks.
 	GetDBHandle() *db.DBHandle
+	// GetStaticFilesystem returns the FS that is used to serve from the /static
+	// directory.
 	GetStaticFilesystem() fs.FS
+	// GetAppFilesystem returns the FS used to serve JS and CSS resources needed
+	// by the app, including the app bundle and any lazily loaded JS chunks.
 	GetAppFilesystem() fs.FS
 	GetBlobstore() interfaces.Blobstore
 	GetInvocationDB() interfaces.InvocationDB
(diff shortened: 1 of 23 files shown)


server/environment/environment.go
```diff
@@ -76,4 +76,5 @@ type Env interface {
 	GetUsageService() interfaces.UsageService
 	GetUsageTracker() interfaces.UsageTracker
 	GetXCodeLocator() interfaces.XcodeLocator
+	GetFileResolver() fs.FS
```
tylerwilliams, line 79: Since we already have:
	GetStaticFilesystem() fs.FS
	GetAppFilesystem() fs.FS
Can we replace those two with this one? Or do we need this new one at all? 3 is a crowd :)

tylerwilliams (commented):
This stuff is sort of finicky and will require a lot of testing:
 (app|executor) x (opensource|enterprise) x (mac|linux)

But the CL mostly LGTM if the tests all look good.

One question I had was: why do you look in the bundle first and the runfiles second, instead of the other way round?

server/environment/environment.go
siggisim: Those two are *super* finicky and I've spent hours trying and failing to clean them up (something always breaks).

siggisim: For posterity - or to save whoever looks into the next some time: the two existing filesystems are actually performing 3 different functions:
- rendering and serving the "index" file (i.e. the html file that loads the app) at various paths (any path that the app is served from, i.e. `/`, `/trends/`, `/tests/`, `/settings/`, ...)
- serving static files (mostly icons and images) from the /static directory at the root
- serving generated "/app" files i.e. app.js, all of the app bundle chunks, style.css, etc

tylerwilliams: This is awesome, thanks for testing this so thoroughly!

tylerwilliams (approved):
LGTM

siggisim:
Looks like this brought back the executor depending on the bundle - which makes the executor depend on and build all of the typescript dependencies and javascript files again:

https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/cmd/executor/BUILD#L11

siggisim:
I'd removed this here: https://github.com/buildbuddy-io/buildbuddy/commit/54c93b4ee4705b3a4a1679d927bbe9759fdc12fd#diff-8571df13787671050df272c1e0cb65096505ca904670cdcd563756e93cf3d45bL12

The executor should probably have a different bundle than the app.

bduffany:
FWIW, the executor was already depending on the bundle indirectly via firecracker, but only on Linux:

https://github.com/buildbuddy-io/buildbuddy/pull/1202/files#diff-498169dee0b28bae776f175e7dc7a789bae8940508200991274f1521e3a57cc5L25

PR #1204, opened 2021-11-05
Add ci_runner option to report repo setup logs live.

vadimberezniker:
Logs are currently not reported until an "action" is run.

This can become the default once
https://github.com/buildbuddy-io/buildbuddy/pull/1190 is rolled out.

---

**Version bump**: None


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -122,6 +123,127 @@ type workspace struct {
 	logBuf bytes.Buffer
 }
 
+type buildEventReporter struct {
```
tylerwilliams, line 126: Could we move this to a separate package or does it need to be in here?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -433,12 +532,12 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace, startTime time.T
 	}
 
 	// Flush whenever the log buffer fills past a certain threshold.
-	ar.log.writeListener = func() {
-		if size := ar.log.Len(); size >= progressFlushThresholdBytes {
-			ar.flushProgress() // ignore error; it will surface in `bep.Wait()`
+	ar.reporter.log.writeListener = func() {
+		if size := ar.reporter.log.Len(); size >= progressFlushThresholdBytes {
+			ar.reporter.FlushProgress() // ignore error; it will surface in `bep.Wait()`
 		}
```
tylerwilliams, line 538: could this logic move inside the reporter class?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -452,16 +551,16 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace, startTime time.T
...
 		// time. The UI is expecting this invocation ID so that it can render a
 		// BuildBuddy invocation URL for each bazel_command that is executed.
 		args = append(args, fmt.Sprintf("--invocation_id=%s", iid))
 
-		runErr := runCommand(ctx, *bazelCommand, args, nil /*=env*/, ar.log)
+		runErr := runCommand(ctx, *bazelCommand, args, nil /*=env*/, ar.reporter.log)
 		exitCode := getExitCode(runErr)
 		if exitCode != noExitCode {
-			ar.log.Printf("%s(command exited with code %d)%s\n", ansiGray, exitCode, ansiReset)
+			ar.reporter.log.Printf("%s(command exited with code %d)%s\n", ansiGray, exitCode, ansiReset)
```
tylerwilliams, line 563: not really necessary to fix rn, and i know it wasn't your doing in this CL, but it's a little bit of a code smell to be reaching so deep into another objects internal state (ar.reporter.log.Fn()) -- my main worry with code like this is accidentally hitting a NPE, but in general i think it's a sign that some logic should be moved from the callsite where this is happening into the object itself

tylerwilliams (approved):
LGTM but please get brandon's sign off on this one too.

enterprise/server/cmd/ci_runner/main.go
bduffany: +1, might make sense to change `PublishStarted` to `Start`, move this logic inside `Start` (after publishing the Started event), and require a symmetric call to `Stop` which stops the flush?

The current logic makes it so that we publish Started, then WorkspaceStatus, and only then do we start flushing progress, but I think it's fine to start flushing progress as soon as we publish the Started event (Bazel does this anyway -- sometimes it even sends progress _before_ publishing the Started event but I think we are able to handle that properly as well, up to a certain number of events)

bduffany (approved):
LGTM modulo the nit about potentially moving the progress flush loop into the buildEventReporter

enterprise/server/cmd/ci_runner/main.go
vadimberezniker: Nothing is forcing us to leave this here, but also not a lot of motivation for moving it out. For now it's somewhat convenient to have it next to the runner code.

enterprise/server/cmd/ci_runner/main.go
vadimberezniker: I changed buildEventReporter to implement the logging methods so callers can use the reporter as a logger directly.

enterprise/server/cmd/ci_runner/main.go
vadimberezniker: done

PR #1205, opened 2021-11-05
Add patch support to the Runners implementation.

vadimberezniker:
---

**Version bump**: None


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -15,14 +15,21 @@ import (
...
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
+	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lockingbuffer"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/google/shlex"
 	"github.com/google/uuid"
 	"github.com/logrusorgru/aurora"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
```
tylerwilliams, line 31: nit: move down with other named imports

server/util/flagutil/flagutil.go
```diff
@@ -0,0 +1,16 @@
+package flagutil
```
tylerwilliams, line 1: <3 <3 <3

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -667,6 +684,27 @@ func (ws *workspace) setup(ctx context.Context) error {
 	return ws.sync(ctx)
 }
 
+func (ws *workspace) applyPatch(ctx context.Context, digestString string, bsClient bspb.ByteStreamClient) error {
```
tylerwilliams, line 687: uber uber nit but I think ordering like (ctx context.Context,  bsClient bspb.ByteStreamClient, digestString string) makes more sense (since digestString is the arg that changes most often)

tylerwilliams (approved):
LGTM

enterprise/server/cmd/ci_runner/main.go
vadimberezniker: done

enterprise/server/cmd/ci_runner/main.go
vadimberezniker: makes sense, done

PR #1207, opened 2021-11-08
Only increment invocation usage after inserting a new invocation row

bduffany:
This should provide a stronger guarantee that the usage count does not exceed the number of rows inserted into the DB (modulo the possibility of deleting invocations via the UI).

---

**Version bump**: Patch

### DIFF
--- a/server/backends/invocationdb/invocationdb.go
+++ b/server/backends/invocationdb/invocationdb.go
@@ -66,9 +66,9 @@ func (d *InvocationDB) InsertOrUpdateInvocation(ctx context.Context, ti *tables.
 				return err
 			}
 			if err := d.createInvocation(tx, ctx, ti); err != nil {
-				created = true
 				return err
 			}
+			created = true
 			return nil
 		}
 		return tx.Model(&existing).Where("invocation_id = ?", ti.InvocationID).Updates(ti).Error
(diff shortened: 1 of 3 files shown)


server/backends/invocationdb/invocationdb.go
```diff
@@ -53,17 +53,27 @@ func (d *InvocationDB) createInvocation(tx *db.DB, ctx context.Context, ti *tabl
...
+	created := false
+	err := d.h.Transaction(ctx, func(tx *db.DB) error {
 		var existing tables.Invocation
 		if err := tx.Where("invocation_id = ?", ti.InvocationID).First(&existing).Error; err != nil {
 			if !db.IsRecordNotFound(err) {
 				return err
 			}
-			return d.createInvocation(tx, ctx, ti)
+			if err := d.createInvocation(tx, ctx, ti); err != nil {
+				created = true
```
tempoz, line 69: The returned object from GORM's `Create` method contains [`RowsAffected`](https://gorm.io/docs/create.html#Create-Record), which would possibly be a better indicator for if we created a row than the value of the Error object is. Maybe change `createInvocation` to also return `RowsAffected`?

bduffany: If the invocation already exists when calling createInvocation, then gorm should return an AlreadyExists error, so I think checking the error is OK

tempoz: sounds good; just wanted to bring up the option.

tempoz (approved):
LGTM

PR #1211, opened 2021-11-08
Include patches in remote builds & tests

siggisim:
The patches generated by `diff-match-patch` are not compatible with git apply:
https://github.com/google/diff-match-patch/wiki/Unidiff

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1214, opened 2021-11-09
Pass effective group id along with jwt

siggisim:
Needed for ddisk to write user-initiated writes (managed bazel patches) with the correct prefix.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -20,6 +20,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lru"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
+	"github.com/buildbuddy-io/buildbuddy/server/util/request_context"
 	"github.com/buildbuddy-io/buildbuddy/server/util/role"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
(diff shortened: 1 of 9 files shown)


enterprise/server/auth/auth.go
```diff
@@ -647,7 +649,15 @@ func ClaimsFromSubID(env environment.Env, ctx context.Context, subID string) (*C
...
-	return userClaims(u), nil
+	eg := ""
+	if c := requestcontext.ProtoRequestContextFromContext(ctx); c != nil && c.GetGroupId() != "" {
+		for _, g := range u.Groups {
+			if g.Group.GroupID == c.GetGroupId() {
+				eg = c.GetGroupId()
+			}
+		}
+	}
+	return userClaims(u, eg), nil
```
bduffany, line 660: Should we update `authutil.EffectiveGroupID` to only consult u.GroupID instead of looking at the request context? (Just for the sake of making `u.GroupID` the source of truth for the preferred, authenticated group ID). Seems like it should be safe since we don't return user claims anywhere except for this function

siggisim: Removed `authutil.EffectiveGroup` altogether

tylerwilliams (approved):
LGTM

enterprise/server/auth/auth.go
bduffany: sg

bduffany (approved):

PR #1215, opened 2021-11-09
Add integration tests for BEP error handling & retry

bduffany:
Adds an integration test that uses a setup like the following:

```
bazel <=> test proxy <=> build event server
```

The proxy can be configured to return a single error back to Bazel when it first sees a stream event of a particular kind, causing Bazel to retry.

---

**Version bump**: Patch


server/test/integration/build_event_protocol/build_event_protocol_test.go
```diff
@@ -28,3 +43,306 @@ func TestBuildWithBESFlags_Success(t *testing.T) {
...
+		// Clear the error injector after it injects an error, to ensure we inject
+		// the error only once per stream proxy instance.
+		p.errorInjector = nil
+		return err
+	}
+	return nil
+}
+
+func (p *BEPProxyServer) PublishLifecycleEvent(ctx context.Context, req *pepb.PublishLifecycleEventRequest) (*empty.Empty, error) {
+	p.t.Log("Proxy: /PublishLifecycleEvent")
```
vadimberezniker, line 201: I would not recommend using the test logger outside of unit tests. 
The test log lines are printed separately from the rest of the logs and do not have timestamps which can make order of events hard to discern from test logs.

vadimberezniker (approved):
This is really awesome! 

Always happy to see more test coverage of Bazel <-> backed interactions.

server/test/integration/build_event_protocol/build_event_protocol_test.go
bduffany: Ah right, I was running with `-test.v` which shows the test logs in real time, but without `-v` I guess the logs would have to be printed at the end of the test. Fixed

PR #1216, opened 2021-11-09
Add prefixes to match git apply's filename expectations

siggisim:
Git seems to strip the first directory from filenames when git applying (likely because `git diff` uses `a/filename` and `b/filename` when rendering diffs.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1217, opened 2021-11-10
Mark runners that hit persistent worker related errors as doNotReuse and add persistent worker tests

siggisim:
This stops us from adding potentially borked persistent workers back to the runner pool and cleans up the persistent process.

---

**Version bump**: Patch


enterprise/server/remote_execution/runner/runner_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"path"
 	"testing"
 	"time"
+    "encoding/base64"
```
tylerwilliams, line 10: nit: ./buildfix.sh this?

tylerwilliams (approved):
LGTM

PR #1218, opened 2021-11-10
Use "raw unpack" when generating Firecracker container fs.

vadimberezniker:
The regular unpack command generates some extra data that is only
necessary for creating new layers which we don't need.

Verified that the Docker executor image we are building already pull the
latest version of umoci.

This is about 30% faster for the buildbuddy image.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1225, opened 2021-11-12
Strip ANSI sequences when downloading logs

bduffany:
Add a lightweight regex to strip ANSI styles, since they render as raw ANSI codes unless viewing the logs with `cat` or `less -r` (most people will probably use their editor or the default text viewer on their system)

---

**Version bump**: Patch

### DIFF
--- a/app/terminal/terminal.tsx
+++ b/app/terminal/terminal.tsx
@@ -10,17 +10,14 @@ interface TerminalState {
   wrap: boolean;
 }
 
-const wrapLocalStorageKey = "terminal-wrap";
-const wrapLocalStorageValue = "wrap";
+const WRAP_LOCAL_STORAGE_KEY = "terminal-wrap";
+const WRAP_LOCAL_STORAGE_VALUE = "wrap";
+const ANSI_STYLES_REGEX = /\x1b\[[\d;]+?m/g;
 
 export default class TerminalComponent extends React.Component<TerminalProps, TerminalState> {
-  state = { wrap: false };
+  state = { wrap: localStorage.getItem(WRAP_LOCAL_STORAGE_KEY) === WRAP_LOCAL_STORAGE_VALUE };
 
   terminalRef = React.createRef<HTMLDivElement>();
-
-  componentDidMount() {
-    this.setState({ wrap: localStorage.getItem(wrapLocalStorageKey) === wrapLocalStorageValue });
-  }
 
   render() {
     return (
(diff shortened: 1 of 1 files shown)


app/terminal/terminal.tsx
```diff
@@ -10,18 +10,15 @@ interface TerminalState {
...
+const ANSI_STYLES_REGEX = /\x1b\[[\d;]+?m/g;
 
 export default class TerminalComponent extends React.Component<TerminalProps, TerminalState> {
-  state = { wrap: false };
+  state = { wrap: localStorage.getItem(WRAP_LOCAL_STORAGE_KEY) === WRAP_LOCAL_STORAGE_VALUE };
 
   terminalRef = React.createRef<HTMLDivElement>();
 
-  componentDidMount() {
-    this.setState({ wrap: localStorage.getItem(wrapLocalStorageKey) === wrapLocalStorageValue });
```
siggisim, line 22: I initially had this in the constructor, but I moved this here because `wrapText` calls getBoundingClientRect to get the width of the terminal - and `this.terminalRef.current` wasn't available at construction time.

app/terminal/terminal.tsx
```diff
@@ -66,14 +63,15 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
...
     this.setState({ wrap: shouldWrap });
   }
 
   handleDownloadClicked() {
     var element = document.createElement("a");
-    element.setAttribute("href", "data:text/plain;charset=utf-8," + encodeURIComponent(this.props.value));
-    element.setAttribute("download", "build_logs.txt");
+    const unstyledLogs = this.props.value.replace(ANSI_STYLES_REGEX, "");
+    element.setAttribute("href", "data:text/plain;charset=utf-8," + encodeURIComponent(unstyledLogs));
+    element.setAttribute("download", "build.log");
```
siggisim, line 74: i'm kinda partial to .txt but ¯\\\_(ツ)\_/¯

app/terminal/terminal.tsx
bduffany: Gotcha, reverted

app/terminal/terminal.tsx
bduffany: Do you mean you prefer the `.txt` extension itself, or the plain-text rendering style of `.txt` vs `.log`? Here's a comparison of how VS Code handles the two different file types -- I think it does a decent job of improving the signal-to-noise ratio (some other editors might implement this as well):

image

tylerwilliams: I base this mostly on my own experience, but it seems like .txt is more widely supported / understood. Pretty mild preference here though, I feel like .log is easily figure-out-able, but it's annoying when it doesn't work automatically.

Agree it looks nicer when it does work.

Are there any numbers about this? How confident are you that everyone can open .log files?

bduffany: I will revert this for now though until I'm able to test on more machines. The more important change is stripping out the ANSI styles so that the text doesn't look messed up when opened with the user's preferred txt viewer.

app/terminal/terminal.tsx
siggisim: If you wanted to avoid the double render on the non-wrap case, you could check that wrap is enabled before calling setState. Not sure how bad the double render is in practice.

siggisim (approved):

bduffany:
> If you wanted to avoid the double render on the non-wrap case, you could check that wrap is enabled before calling setState. Not sure how bad the double render is in practice.

I haven't noticed much of an issue for the non-wrap case, though I haven't tested yet with very long logs. The wrap case does seem slightly janky on initial page load though, although it may just be the worse-than-usual hardware / internet connection I'm using atm

PR #1228, opened 2021-11-12
Fix issue that workflow_test fails locally if less than 8GB RAM is available

bduffany:
My laptop has 16GB RAM, but workflow_test fails locally because `resources.go` computes the available RAM as less than that.

Fixed by allowing `SYS_MEMORY_BYTES` and `SYS_MILLICPU` to be specified as flags (just like `MY_POOL`), and set those in the test to 10GB. The test workflows are running on tiny workspaces so should not need 8GB anyway. If they do, tests will fail with OOM errors, which seems OK.

---

**Version bump**: Patch


enterprise/server/test/integration/workflow/workflow_test.go
```diff
@@ -61,6 +61,8 @@ func setup(t *testing.T, gp interfaces.GitProvider) (*rbetest.Env, interfaces.Wo
 			env.SetInvocationSearchService(iss)
 		},
 	})
+	// Configure executors so that we can run at least one workflow task.
+	flags.Set(t, "executor.memory_bytes", fmt.Sprintf("%d", 10_000_000_000))
```
tylerwilliams, line 65: please link to / document where this 10G number comes from?

// Workflow tasks require 8Gi to schedule, but all of our workflow test cases
// typically use far less RAM. To enable this test to run on machines smaller
// than 8Gi, force-set the executor memory_bytes to 10Gi so the workflow
// tasks will schedule.

tylerwilliams (approved):
LGTM

tylerwilliams:
No preference, but please at least consider the other approach, where you modify the workflow task size instead?

enterprise/server/test/integration/workflow/workflow_test.go
bduffany: Thanks! Done

bduffany (commented):
>  consider the other approach, where you modify the workflow task size instead

After a little bit of thought, I think both approaches have roughly the same downsides -- in one case we're overriding the task size estimate which isn't something we'd do in production, and in another case we're overestimating our available RAM, which also isn't something we'd do in production. Will stick with this approach for now

PR #1229, opened 2021-11-12
Make portleaser threadsafe

tylerwilliams:
I rely on this heavily in gossip / raft tests where each node needs to open multiple ports and the test opens multiple nodes in parallel.


bduffany (approved):
Noice, definitely an improvement

PR #1230, opened 2021-11-12
Set BES args on code-initiated builds so you can debug test results

siggisim:
---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

(would be nice to set this from app > cache_api_url or something)

PR #1232, opened 2021-11-13
Save code editor state in local storage

siggisim:
We serialize the code editor's state object - this allows you to return to your changes later or refresh the page.

Now that we can serialize this state, it's easy to store this state on server / in cache instead of in local storage.

---

**Version bump**: Patch


MarcusSorealheis:
This is a good improvement because local storage is somewhat sketchy and vulnerable to exfiltration

tylerwilliams (approved):
LGTM

PR #1234, opened 2021-11-15
Add better cache miss debugging

tylerwilliams:
This CL adds support for an ActionCache miss scorecard, basically a list of targets and their associated action ids that were cache misses during a build. The data is collected using deferred counters in redis|memory, and finalized to blobstore at the same time the cache stats are finalized.

This adds a card to the cache tab that shows the items that were missing from the action cache, like this:

image

Future improvements (maybe someone more UI handy than me can help with):
 - pagination, in case the scorecard is large (it can be)
 - linking to a card to display the action, rather than downloading it, so users can debug a cache miss entirely in the UI

### DIFF
--- a/app/invocation/scorecard_card.tsx
+++ b/app/invocation/scorecard_card.tsx
@@ -1,0 +1,53 @@
+import React from "react";
+import InvocationModel from "./invocation_model";
+import rpcService from "../service/rpc_service";
+
+interface Props {
+  model: InvocationModel;
+}
+
+export default class ScorecardCardComponent extends React.Component {
+  props: Props;
+
+  getCacheAddress() {
+    let address = this.props.model.optionsMap.get("remote_cache").replace("grpc://", "");
+    address = address.replace("grpcs://", "");
+    if (this.props.model.optionsMap.get("remote_cache")) {
+      address = this.props.model.optionsMap.get("remote_cache").replace("grpc://", "");
+      address = address.replace("grpcs://", "");
+    }
+    if (this.props.model.optionsMap.get("remote_instance_name")) {
+      address = address + "/" + this.props.model.optionsMap.get("remote_instance_name");
+    }
+    return address;
+  }
+
+  getActionDownloadURL(actionId: string) {
+    let actionResultURI = "actioncache://" + this.getCacheAddress() + "/blobs/ac/" + actionId + "/141";
+    return rpcService.getBytestreamFileUrl(actionId, actionResultURI, this.props.model.getId());
+  }
+
+  render() {
+    return (
+      <div className="card scorecard">
+        <img className="icon" src="/image/x-circle-regular.svg" />
+        <div className="content">
+          <div className="title">Cache Misses</div>
+          <div className="details">
+            {this.props.model.scoreCard.misses.map((result) => (
+              <div>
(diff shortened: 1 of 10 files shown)


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -765,30 +800,60 @@ func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*in
 		}
 	}
 
+	invocationMu := sync.Mutex{}
```
bduffany, line 803: uber nit: to avoid a mutex, the first goroutine could read the ScoreCard, then after `eg.Wait()` tack the scorecard field onto the invocation

app/invocation/scorecard_card.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return address;
+  }
+
+  getActionDownloadURL(actionId: string) {
+    let actionResultURI = "actioncache://" + this.getCacheAddress() + "/blobs/ac/" + actionId + "/141";
+    return rpcService.getBytestreamFileUrl(actionId, actionResultURI, this.props.model.getId());
+  }
+
+  render() {
+    return (
```
bduffany, line 28: might need a guard clause like the following, to avoid an NPE if the invocation doesn't have a scorecard:

```
if (!this.props.model.scoreCard) return null;
```

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -197,6 +198,34 @@ func (r *statsRecorder) Enqueue(ctx context.Context, invocation *inpb.Invocation
 	}
 }
 
+func scoreCardBlobName(invocationID string) string {
+	blobFileName := invocationID + "-scorecard.chunk"
```
bduffany, line 202: uber nit: since this is a non-chunked blob, consider removing the `.chunk` suffix?

bduffany (changes requested):

app/invocation/scorecard_card.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+import InvocationModel from "./invocation_model";
+import rpcService from "../service/rpc_service";
+
+interface Props {
+  model: InvocationModel;
+}
+
+export default class ScorecardCardComponent extends React.Component {
```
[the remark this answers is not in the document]
tylerwilliams: yeah, planning to eventually add other stuff to scorecard

app/invocation/scorecard_card.tsx
tylerwilliams: Done.

server/build_event_protocol/build_event_handler/build_event_handler.go
tylerwilliams: Done.

server/build_event_protocol/build_event_handler/build_event_handler.go
tylerwilliams: ack; I think that's a little more finesse than I want to apply here :P I'd prefer to keep it really straightforward (esp since I don't believe contention will be a problem with only the 2 goroutines).

bduffany (approved):
LGTM, nice :)

PR #1235, opened 2021-11-15
Don't use a git credential helper for the workflow repo

bduffany:
This fixes an issue with Mac workflows that the osxkeychain credential helper shows an interactive dialog which blocks the `git fetch` until the dialog is clicked through.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1236, opened 2021-11-15
Allow configuring the OS for workflows in buildbuddy.yaml

bduffany:
Also, don't set `container-image` unless the OS is `linux`.

This requires us to read the workflow config when calling `ExecuteWorkflow`, so a little bit of refactoring was needed. `executeWorkflow` now requires the `workflowAction` config from `buildbuddy.yaml`. Since we have that object, we no longer need to pass `--action_name=...` as an extra CI runner arg, and can always set it based on the passed in `workflowAction`.

Note that this does not make any changes to `pool` - that'll be a separate PR.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1237, opened 2021-11-15
Run bazel shutdown when evicting bare workflow runners from the runner pool

bduffany:
When evicting a CI runner from the runner pool, we delete the runner's workspace directory, but the Bazel server is still left running, taking up memory resources.

This PR introduces a `shutdown` phase of the runner lifecycle that allows manual cleanup of processes (for bare runners only), and adds shutdown logic for CI runners to call the `buildbuddy_ci_runner` with a `--shutdown_and_exit` flag.

Tested locally with a Mac executor (I quit the Mac executor with Ctrl+C which triggers all runners to be removed, and saw the expected output being printed when running with `--debug_stream_command_outputs`).

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1239, opened 2021-11-16
Add paging for cache scorecard and link to action results

bduffany:
* Implement client-side paging (the same way we do for artifacts)
* Link to action results page
* Add action mnemonic to links
* Make the styling consistent with the test artifacts list

image

TODO in future PRs:
- Don't hard-code the action digest size as 141: either don't require a size to be able to view the action page, or find a way to use the actual digest size
- If the action is not found, show a help message with the list of potential reasons: `--noremote_upload_local_results` was set; a read-only API key was used; action expired or was manually deleted from CAS; ...?

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -120,7 +120,7 @@ export default class InvocationActionCardComponent extends React.Component<Props
   }
 
   getCacheAddress() {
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
LGTM

siggisim (approved):

PR #1241, opened 2021-11-17
Allow specifying CPU architecture in workflow config

bduffany:
This lets us support M1 Macs (by setting `os: darwin, arch: arm64`)

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1242, opened 2021-11-17
Embed bazelisk into the CI runner to simplify Mac workflow executor provisioning

bduffany:
Increases the size of the standalone CI runner binary by about 5MB (bringing the total size to 25MB).

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"sync"
@@ -20,6 +21,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/workflow/config"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
+	"github.com/buildbuddy-io/buildbuddy/server/util/bazelisk"
 	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
 	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lockingbuffer"
@@ -93,7 +95,7 @@ var (
 	patchDigests                flagutil.StringSliceFlag
 	reportLiveRepoSetupProgress = flag.Bool("report_live_repo_setup_progress", false, "If set, repo setup output will be streamed live to the invocation instead of being postponed until the action is run.")
 
-	bazelCommand      = flag.String("bazel_command", bazeliskBinaryName, "Bazel command to use.")
+	bazelCommand      = flag.String("bazel_command", "", "Bazel command to use.")
 	bazelStartupFlags = flag.String("bazel_startup_flags", "", "Startup flags to pass to bazel. The value can include spaces and will be properly tokenized.")
 	debug             = flag.Bool("debug", false, "Print additional debug information in the action logs.")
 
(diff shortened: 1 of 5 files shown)


server/util/bazelisk/BUILD
```diff
@@ -0,0 +1,24 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+genrule(
+    name = "bazelisk-1.10.1_crossplatform",
+    srcs = select({
+        "@bazel_tools//src/conditions:darwin_x86_64": ["@io_bazel_bazelisk-1.10.1-darwin-amd64//file:downloaded"],
+        "@bazel_tools//src/conditions:darwin_arm64": ["@io_bazel_bazelisk-1.10.1-darwin-arm64//file:downloaded"],
+        "//conditions:default": ["@io_bazel_bazelisk-1.10.1-linux-amd64//file:downloaded"],
```
brentleyjones, line 8: Does this end up downloading only the active condition, or all of them (https://github.com/bazelbuild/bazel/issues/12861)? With toolchains it would for sure only download the active condition. https://github.com/keith/buildifier-prebuilt is an example for buildifier.

bduffany: I tried running a clean build on a Linux machine, and it looks like we fetch only the linux-amd64 build: https://app.buildbuddy.io/invocation/191db983-3c49-4880-8e43-6296edc0241e#fetches

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -697,6 +703,39 @@ func ensurePath() error {
...
+		return "", err
+	}
+	// If bazelisk is already in the workspace root, use that.
+	if _, err := os.Stat(bazeliskBinaryName); err != nil {
+		if !os.IsNotExist(err) {
+			return "", err
+		}
+	} else {
+		return filepath.Join(wd, bazeliskBinaryName), nil
+	}
```
tylerwilliams, line 718: I would ignore this error and fall through to your case below. It also would make this code a little easier to follow:

	if _, err := os.Stat(bazeliskBinaryName); err == nil {
        return filepath.Join(wd, bazeliskBinaryName), nil
	}

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -697,6 +703,39 @@ func ensurePath() error {
...
+	// If bazelisk is embedded, copy it to the workspace root, and use that.
+	if f, err := bazelisk.Open(); err == nil {
```
tylerwilliams, line 720: the fact that this is just bazelisk.Open made me laugh. idk what i expected

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -697,6 +703,39 @@ func ensurePath() error {
...
+		defer dst.Close()
+		if _, err := io.Copy(dst, f); err != nil {
+			return "", err
+		}
+		return filepath.Join(wd, bazeliskBinaryName), nil
+	}
+	// If bazelisk is in PATH, use that.
+	if _, err := exec.LookPath(bazeliskBinaryName); err != nil {
+		return "", err
+	}
```
tylerwilliams, line 735: Should we move this up?

The behavior of using the bundled bazelisk, unless there was an error, in which case we fall back to the system one, is unusual. If we used the system one and fell back to built in, that would make sense, or alternatively, if we always used the bundled one or returned an error, that would also make sense.

tylerwilliams (changes requested):

enterprise/server/cmd/ci_runner/main.go
tylerwilliams: image

enterprise/server/cmd/ci_runner/main.go
bduffany: sg to always use the bundled one -- done. I made bazelisk a hard dependency instead of excluding it in fastbuild mode, which seems OK since it's pretty small. If we find that this slows down builds during development then we can maybe use a fileresolver to have it work with runfiles as well.

tylerwilliams (approved):
LGTM

PR #1244, opened 2021-11-17
Add a mock oauth option

tempoz:
Adds a config flag to enable an oauth authenticator that uses the buildbuddy
server to authenticate. The server always authenticates the client as the same
user. The flag additionally enables endpoints on the http mux to support the
oauth requests. The key for signing is specified in the code directly, and thus
is not secure, nor is it intended to be. This is necessary to ensure tokens still
work if the client reconnects to a different server.

Authenticators are now initialized on-demand so that the mock oauth authenticator
can use the openid-configuration endpoint on the server during initialization.

---

**Version bump**: Minor


enterprise/server/cmd/server/main.go
```diff
@@ -336,5 +346,15 @@ func main() {
...
 
+	if realEnv.GetConfigurator().GetMockOauthIssuer() != nil {
+		oauth, err := mockoauth.NewMockOauth(realEnv)
+		if err != nil {
+			log.Fatalf("Error initializing mock oauth: %s", err)
+		}
+		realEnv.GetAdditionalMuxEntries()[oauth.AuthorizationEndpoint().Path] = httpfilters.SetSecurityHeaders(http.HandlerFunc(oauth.Authorize))
+		realEnv.GetAdditionalMuxEntries()[oauth.TokenEndpoint().Path] = httpfilters.SetSecurityHeaders(http.HandlerFunc(oauth.AccessToken))
+		realEnv.GetAdditionalMuxEntries()[oauth.JwksEndpoint().Path] = httpfilters.SetSecurityHeaders(http.HandlerFunc(oauth.Jwks))
+		realEnv.GetAdditionalMuxEntries()["/.well-known/openid-configuration"] = httpfilters.SetSecurityHeaders(http.HandlerFunc(oauth.WellKnownOpenIDConfiguration))
```
bduffany, line 357: nit: might be slightly clearer as

```
mux := realEnv.GetAdditionalMuxEntries()
mux[...] = ...
mux[...] = ...
...
```

bduffany (changes requested):

enterprise/server/mockoauth/mockoauth.go
```diff
@@ -0,0 +1,253 @@
+package mockoauth
```
tylerwilliams, line 1: I think we should call this something besides mockauth -- since mocks are usually associated with test-only code and this is clearly not that, it's confusing for other folks working on the code, but also for customers in the config.

Instead of mockauth, what do you think about singleuserauth or selfauth?

enterprise/server/mockoauth/mockoauth.go
```diff
@@ -0,0 +1,253 @@
...
+	n.SetString(
+		"23854054546089486752143428681593425828175873941670684122390466251269713159"+
+			"245737985667367468143300414391912523520647401288672254625181245443791922"+
+			"131328655375075086232375550780596731379873494676833765863924320065467041"+
+			"534880370068444461206120370979494670081178693685915972335443995410513383"+
+			"318660804108686793740105181807521526108041735581000243371902164278468459"+
+			"474539325728897031680901033148357912833321597059542124576712419669819030"+
+			"691795237787947940066997209320864327657065503764510504770118773928022493"+
+			"125524472354132671270720851491074770807035708210053594371353758234068435"+
+			"932384388342343398119337427425316157863",
```
tylerwilliams, line 66: to do a raw string in go, you can use backticks (same below)

enterprise/server/mockoauth/mockoauth.go
```diff
@@ -0,0 +1,253 @@
...
+	signed, err := jwt.Sign(token, jwa.RS256, o.rsaPrivateKey)
+	if err != nil {
+		log.Errorf("Error signing token: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	writeJSONResponse(
+		w,
+		r,
+		&tokenJSON{
```
tylerwilliams, line 235: I think a more idiomatic formatting of these would be: writeJSONResponse(w, r, &tokenJSON{
....})

server/environment/environment.go
```diff
@@ -86,4 +87,5 @@ type Env interface {
 	//
 	// See server/util/fileresolver/fileresolver.go
 	GetFileResolver() fs.FS
+	GetAdditionalMuxEntries() map[string]http.Handler
```
tylerwilliams, line 90: I think instead of putting a map in the env like this you could store a mux in the env and then just add it to the root serve mux in libmain. This is nice because maps are sort of a useful but gross type to pass around in APIs. A mux would be a little bit cleaner.

https://stackoverflow.com/questions/23693520/go-how-to-combine-two-or-more-http-servemux

enterprise/server/auth/auth.go
```diff
@@ -226,9 +226,11 @@ type authenticator interface {
 type oidcAuthenticator struct {
 	oauth2Config *oauth2.Config
 	oidcConfig   *oidc.Config
-	provider     *oidc.Provider
+	Provider     *oidc.Provider
```
tylerwilliams, line 229: Does Provider need to be a public member? (I'm probably missing where this is used...)

enterprise/server/auth/auth.go
```diff
@@ -243,25 +245,40 @@ func extractToken(issuer, slug string, idToken *oidc.IDToken) (*userToken, error
...
 }
 
 func (a *oidcAuthenticator) verifyTokenAndExtractUser(ctx context.Context, jwt string, checkExpiry bool) (*userToken, error) {
+	if err := a.initialize(); err != nil {
+		return nil, err
+	}
 	conf := a.oidcConfig
 	conf.SkipExpiryCheck = !checkExpiry
-	validToken, err := a.provider.Verifier(conf).Verify(ctx, jwt)
+	validToken, err := a.Provider.Verifier(conf).Verify(ctx, jwt)
```
tylerwilliams, line 281: I'm maybe missing other occurrences -- but if this is the only place a method is called on provider -- maybe we could do the providerOnce.Do(init()) here and then we would not need initialize() elsewhere?

tylerwilliams (changes requested):
The meat of this (the single user authenticator) looks really good! I have a few comments about how it integrates with everything else, but pretty minor. Looking forward to getting this in.

enterprise/server/auth/auth.go
```diff
@@ -359,17 +372,32 @@ func createAuthenticatorsFromConfig(ctx context.Context, authConfigs []config.Oa
...
+		var once sync.Once
+		authenticator := &oidcAuthenticator{
 			slug:         authConfig.Slug,
 			issuer:       authConfig.IssuerURL,
 			oauth2Config: oauth2Config,
 			oidcConfig:   oidcConfig,
-			provider:     provider,
-		})
+		}
+		// initialize on-demand, since our mock oauth provider won't be reachable until the server starts
```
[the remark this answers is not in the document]
bduffany: How about having an internal bool to prevent it from running more than once?

```
AddHealthCheck(..., func(ctx ...) error {
  if initialized {
    return nil
  }
  if err := auth.Initialize(); err != nil {
    return err
  }
  initialized = true
  return nil
})
```

(a util func in `healthcheck.go` like `AddStartupCheck(healthChecker, func)` could do this automatically)

enterprise/server/mockoauth/mockoauth.go
bduffany: +1 for selfauth, a little more concise while still communicating the purpose well

enterprise/server/auth/auth.go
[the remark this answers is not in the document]
bduffany: Chatted offline a bit -- the lazy-init approach is fine (it's simpler than a health-check), but one problem is that if the oidc provider has a blip, the server gets stuck in a broken state while reporting itself as healthy, because we cache the error.

One way around this:
* Do the lazy init only for the mock (selfauth?) case -- otherwise initialize immediately in the constructor. This way, the server will do what it does today if the OIDC provider fails to be configured (it crashes on startup and K8S will CrashLoopBackoff, restarting the server until it succeeds)

* If we are worried about errors in the selfauth case as well (i.e. a temporary internal error), then one way around it is to implement the lazy-init so that it can run as many times as is needed until there is no error (can maybe implement this with a mutex, rather than sync.Once)

enterprise/server/auth/auth.go
```diff
@@ -224,11 +224,13 @@ type authenticator interface {
 }
 
 type oidcAuthenticator struct {
-	oauth2Config *oauth2.Config
-	oidcConfig   *oidc.Config
-	provider     *oidc.Provider
-	issuer       string
-	slug         string
+	oauth2Config  func() (*oauth2.Config, error)
+	_oauth2Config *oauth2.Config
```
bduffany, line 228: uber nit: might be slightly clearer to replace `_` with `cached`

enterprise/server/auth/auth.go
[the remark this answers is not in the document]
bduffany: Seems OK to me since it fixes the problem that the server can get stuck in a broken state.

One small downside is that if people misconfigure their auth provider, they no longer get the behavior that the server crashes on startup and they see the error printed immediately. I'm not sure how common that problem is in practice though, so I'd be fine to keep as-is :shrug:

bduffany (approved):
LGTM (please also get Tyler's approval)

enterprise/server/mockoauth/mockoauth.go
tylerwilliams: I think the idiomatic way to do this in go is to use a raw string. Here's a specific example of a key in the stdlib (albeit in a test): https://cs.opensource.google/go/go/+/master:src/crypto/x509/x509_test.go;l=106?q=%22Key%20%3D%20%60%22&ss=go%2Fgo

But there are many example of embedding content like this: https://cs.opensource.google/search?q=%22%3D%20%60%22&ss=go%2Fgo (usages, examples, programs, etc.)

server/config/config.go
```diff
@@ -735,6 +737,20 @@ func (c *Configurator) GetSAMLConfig() *SAMLConfig {
...
+	u, err := url.Parse(c.GetAppBuildBuddyURL())
+	if err != nil {
+		u = &url.URL{
+			Scheme: "http",
+			Host:   "localhost",
+		}
+	}
+	return u
+}
+
```
tylerwilliams, line 753: nit: most of the code in configurator is just boilerplate that reads a field and returns it. Could you instead return a bool indicating if self auth is enabled and do this url construction in the place you need the url?

server/real_environment/real_environment.go
```diff
@@ -85,6 +88,7 @@ func NewRealEnv(c *config.Configurator, h interfaces.HealthChecker) *RealEnv {
 		healthChecker: h,
 
 		executionClients: make(map[string]*executionClientConfig, 0),
+		mux:              tracing.NewHttpServeMux(http.NewServeMux()),
```
tylerwilliams, line 91: nit: we should keep the environment pretty agnostic to stuff (like, whether or not the mux is a tracing mux or regular mux). Could you add a Set method in env for the mux and then set it in libmain when you create a new env?

tylerwilliams (changes requested):

enterprise/server/mockoauth/mockoauth.go
tempoz: Sure, but this is a string representing a base 10 number, not a standard string representation of an RSA key. `big.Int` won't be able to parse anything with non-digit (in this case, new-line) characters.

To be more clear, I could use a single raw string with backticks, but it would just look like a single raw string with double quotes:
```
n.SetString(
		`23854054546089486752143428681593425828175873941670684122390466251269713159245737985667367468143300414391912523520647401288672254625181245443791922131328655375075086232375550780596731379873494676833765863924320065467041534880370068444461206120370979494670081178693685915972335443995410513383318660804108686793740105181807521526108041735581000243371902164278468459474539325728897031680901033148357912833321597059542124576712419669819030691795237787947940066997209320864327657065503764510504770118773928022493125524472354132671270720851491074770807035708210053594371353758234068435932384388342343398119337427425316157863`,
		10)
```
and, specifically, it can't be:
```
n.SetString(
		`23854054546089486752143428681593425828175873941670684122390466251269713159
245737985667367468143300414391912523520647401288672254625181245443791922
131328655375075086232375550780596731379873494676833765863924320065467041
534880370068444461206120370979494670081178693685915972335443995410513383
318660804108686793740105181807521526108041735581000243371902164278468459
474539325728897031680901033148357912833321597059542124576712419669819030
691795237787947940066997209320864327657065503764510504770118773928022493
125524472354132671270720851491074770807035708210053594371353758234068435
932384388342343398119337427425316157863`,
		10)
```

because that string has new lines, which `big.Int` can't parse.

enterprise/server/auth/auth.go
tempoz: I like this.

enterprise/server/mockoauth/mockoauth.go
tylerwilliams: Can make these things consts at the top (using raw string) and in your code call n.SetString(strings.Join(strings.Fields(myKey), ""))?

tempoz: works for me.

server/config/config.go
tempoz: Fixed.

server/real_environment/real_environment.go
tempoz: Fixed.

tylerwilliams (approved):
LGTM

PR #1246, opened 2021-11-18
Allow the menu bar to scroll away on mobile

siggisim:
This increases available viewport when viewing on small devices like a mobile screen.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1249, opened 2021-11-18
More CSS fixes for build logs

bduffany:
Tested in Chrome (both Linux and macOS) and Safari.

Test log (light / non-dense)

image

Build log (dark / dense)

image

---

**Version bump**: Patch

### DIFF
--- a/app/root/root.css
+++ b/app/root/root.css
@@ -927,8 +927,6 @@ code .comment {
 .terminal {
   height: 300px;
   margin-top: -49px;
-  /* overflow-x: auto;
-  overflow-y: hidden; */
   overflow: visible;
 }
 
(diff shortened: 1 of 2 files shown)


app/root/root.css
```diff
@@ -938,14 +941,23 @@ code .comment {
 }
 
 .terminal-actions {
-  right: 32px;
-  top: 22px;
```
siggisim, line 942: Does not specifying the top work for non-dense mode too?

siggisim (approved):

app/root/root.css
bduffany: Yep, it works due to the new `.terminal-container` wrapper div around the terminal component that has `position: relative`

PR #1251, opened 2021-11-18
Fix MySQL 8 syntax for BEP webhooks

bduffany:
The infamous `Groups` table quoting issue strikes once again. This is a problem in MySQL 8 but not 5.7

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1252, opened 2021-11-22
Add a webdriver test for authenticated invocations (via self-auth)

bduffany:
Uses the new self-auth to log in to the app, gets the bazelrc contents (including the API key) from the setup page, then runs a bazel invocation using those recommended bazelrc flags. Asserts that we can view the invocation if we're still logged in, and that if we're logged out, we see the login page instead.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/test/webdriver/invocation/invocation_test.go
+++ b/enterprise/server/test/webdriver/invocation/invocation_test.go
@@ -1,0 +1,63 @@
+package invocation_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/testutil/buildbuddy_enterprise"
+	"github.com/buildbuddy-io/buildbuddy/server/testutil/testbazel"
+	"github.com/buildbuddy-io/buildbuddy/server/testutil/webtester"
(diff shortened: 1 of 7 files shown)


enterprise/server/test/webdriver/invocation/invocation_test.go
```diff
@@ -0,0 +1,63 @@
...
+	// Log out and make sure we see only the login page when attempting to view
+	// the invocation again
+
+	webtester.Logout(wt)
+
+	wt.Get(app.HTTPURL() + "/invocation/" + result.InvocationID)
+
+	wt.Find(".login-button")
+
+	// TODO(bduffany): Log in as a different self-auth user that is not in the
```
tempoz, line 60: Currently self-auth only supports one user, but it would not be too hard to change it.

tempoz (approved):
LGTM

This is awesome, Brandon! I'm super-happy that we already have tests for self-auth :D

bduffany:
> LGTM
> 
> This is awesome, Brandon! I'm super-happy that we already have tests for self-auth :D

Thanks! And very nice job with the implementation, it worked without any hiccups at all :smiley:

PR #1253, opened 2021-11-22
API improvements

siggisim:
- Allow for target selection by label by adding `label` as a field to `TargetSelector`
- Include digest and file size when returning `File` protos
- Return `language` for targets

---

**Version bump**: Patch


tylerwilliams (approved):
nit: want to add a test for the label querying?

siggisim:
> nit: want to add a test for the label querying?

Added tests for all of the new functionality, and a lot of the old

PR #1254, opened 2021-11-22
Temporarily revert bazel cleanup logic causing workflow_test to be flaky

bduffany:
This was added in #1237 which is not in prod yet so it's safe to revert. Occasionally, it causes workflow_test to fail.

---

**Version bump**: Patch


tylerwilliams (approved):
Lgtm

PR #1255, opened 2021-11-23
Fix workflow_test flakiness and re-enable bare workflow runner shutdown logic

bduffany:
The root cause is that we weren't properly waiting for all executors to shut down before removing the test root workspace containing the data for all the executors.

---

**Version bump**: Patch


tempoz (approved):
LGTM, though feel free to wait for other reviewers if you think it's prudent.

PR #1256, opened 2021-11-23
Change the log download button on the terminal compnent so that it fetches the full log from the server

tempoz:
---

**Version bump**: Minor

### DIFF
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -264,7 +264,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
               value={this.getBuildLogs()}
               loading={this.areBuildLogsLoading()}
               expanded={activeTab == "log"}
-              fullLogsFetcher={async () => {
+              fullLogsFetcher={() => {
                 return rpcService.service
                   .getEventLogChunk(
                     new eventlog.GetEventLogChunkRequest({
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -51,6 +51,7 @@ interface Props {
 
 const largePageSize = 100;
 const smallPageSize = 10;
+const maxInt32 = 2147483647;
 
 export default class InvocationComponent extends React.Component<Props, State> {
   state: State = {
@@ -206,7 +207,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
         .getEventLogChunk(
           new eventlog.GetEventLogChunkRequest({
             invocationId: this.props.invocationId,
-            minLines: 2147483647, // int32 max value: max number of lines we can request.
+            minLines: maxInt32, // int32 max value: max number of lines we can request.
           })
         )
         .then((response: eventlog.GetEventLogChunkResponse) => {
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -264,6 +264,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
               value={this.getBuildLogs()}
               loading={this.areBuildLogsLoading()}
               expanded={activeTab == "log"}
+							invocationId={this.props.invocationId}
             />
           )}
 
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -3,7 +3,6 @@ import React from "react";
 import { Subscription } from "rxjs";
 import { invocation } from "../../proto/invocation_ts_proto";
 import { User } from "../auth/auth_service";
-import capabilities from "../capabilities/capabilities";
 import faviconService from "../favicon/favicon";
 import rpcService from "../service/rpc_service";
 import TargetComponent from "../target/target";
@@ -32,6 +31,7 @@ import InvocationActionCardComponent from "./invocation_action_card";
 import TargetsComponent from "./invocation_targets";
 import { BuildBuddyError } from "../util/errors";
 import UserPreferences from "../preferences/preferences";
+import { eventlog } from "../../proto/eventlog_ts_proto";
 
 interface State {
   loading: boolean;
@@ -264,7 +264,18 @@ export default class InvocationComponent extends React.Component<Props, State> {
               value={this.getBuildLogs()}
               loading={this.areBuildLogsLoading()}
               expanded={activeTab == "log"}
-              invocationId={this.props.invocationId}
+              fullLogsFetcher={async () => {
+                return rpcService.service
+                  .getEventLogChunk(
+                    new eventlog.GetEventLogChunkRequest({
+                      invocationId: this.props.invocationId,
+                      minLines: 2147483647, // int32 max value: max number of lines we can request.
+                    })
+                  )
+                  .then((response: eventlog.GetEventLogChunkResponse) => {
+                    return new TextDecoder().decode(response.buffer || new Uint8Array());
+                  });
+              }}
             />
           )}
 
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -264,7 +264,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
               value={this.getBuildLogs()}
               loading={this.areBuildLogsLoading()}
               expanded={activeTab == "log"}
-							invocationId={this.props.invocationId}
+              invocationId={this.props.invocationId}
             />
           )}
 
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -201,6 +201,18 @@ export default class InvocationComponent extends React.Component<Props, State> {
       denseMode: this.props.preferences.denseModeEnabled,
     });
     const isBazelInvocation = this.state.model.isBazelInvocation();
+    const fetchBuildLogs = () => {
+      return rpcService.service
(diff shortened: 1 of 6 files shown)


app/terminal/terminal.tsx
```diff
@@ -72,6 +75,27 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
   }
 
   handleDownloadClicked() {
+    if (typeof this.props.invocationId === "string") {
```
bduffany, line 78: nit: equivalently, you can write `if (this.props.invocationId)`

app/terminal/terminal.tsx
```diff
@@ -72,6 +75,27 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
...
+      rpcService.service
+        .getEventLogChunk(
+          new eventlog.GetEventLogChunkRequest({
```
bduffany, line 81: nit: while this RPC is loading, we should show a loading spinner so the user gets active feedback that we're working on the request, in case the log is particularly large. Also, if there's an error in the RPC, we can use `errorService` from `app/errors/error_service` to show a temporary error banner.

Something like this should probably work:

```
this.setState({ isLoadingFullLog: true });
rpcService.service.getEventLogChunk(...)
    .then(...)
    .catch((e) => errorService.handleError(BuildBuddyError.parse(e)))
    .finally(() => this.setState({ isLoadingFullLog: false });
```

and when rendering (note the `disabled` state and also the `Spinner` from `app/components/spinner/spinner`):

```
<button ... disabled={this.state.isLoadingFullLog}>
  {this.state.isLoadingFullLog ? <Spinner /> : <img ... />}
</button>
```

Might need some CSS tweaks to make the spinner the same size as the download button, as well as making sure the spinner works in both light and dark terminal theme.

app/terminal/terminal.tsx
```diff
@@ -1,9 +1,12 @@
 import React from "react";
 import { LazyLog } from "react-lazylog";
+import { eventlog } from "../../proto/eventlog_ts_proto";
+import rpcService from "../service/rpc_service";
 
 export interface TerminalProps {
   value?: string;
   lightTheme?: boolean;
+  invocationId?: string;
```
bduffany, line 9: It might give us more flexibility to keep this component agnostic to invocations. Maybe instead pass in a `fullLogFetcher`?

```
fullLogFetcher?: () => Promise<string>;
```

bduffany (changes requested):

app/invocation/invocation.tsx
```diff
@@ -264,6 +264,18 @@ export default class InvocationComponent extends React.Component<Props, State> {
               value={this.getBuildLogs()}
               loading={this.areBuildLogsLoading()}
               expanded={activeTab == "log"}
+              fullLogsFetcher={async () => {
```
bduffany, line 267: uber nit: since this isn't using `await`, and the value being returned already has type `Promise<string>`, there's no need to annotate with `async`

app/terminal/terminal.tsx
```diff
@@ -72,7 +80,33 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
...
+        .catch((e) => errorService.handleError(e))
+        .finally(() => {
+          this.setState({ isLoadingFullLog: false });
+          try {
+            document.body.removeChild(element);
+          } catch (e) {
+            if (!(e instanceof DOMException) || e.name != "NotFoundError") {
+              errorService.handleError(e);
+            }
+          }
```
bduffany, line 105: I don't totally understand this part -- do errors from the download itself only surface when removing the element? (That seems surprising to me -- seems like they should be surfaced regardless of whether or not the element is removed)

It would be nice if the `remove` didn't need special case handling for the `logsFetcher` case, because then the downloading part could be factored into a separate top-level utility function like `downloadFile()` which creates the invisible link, clicks on it, then removes it

app/terminal/terminal.tsx
```diff
@@ -72,7 +80,33 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
...
+          element.setAttribute("download", "build_logs.txt");
+          element.style.display = "none";
+          document.body.appendChild(element);
+          element.click();
+        })
+        .catch((e) => errorService.handleError(e))
+        .finally(() => {
+          this.setState({ isLoadingFullLog: false });
+          try {
+            document.body.removeChild(element);
```
bduffany, line 100: nit: `element.remove()`

bduffany (changes requested):

app/terminal/terminal.tsx
tempoz: Awesome, thanks.

app/terminal/terminal.tsx
tempoz: I didn't know about `element.remove()`, and `removeChild` *could* throw an error if the child doesn't exist, which is to say if any statement before `appendChild` threw an exception. I didn't want to propagate that error. Fortunately, `element.remove()` has no such problems.

app/invocation/invocation.tsx
tempoz: mmm, I see. `tsserver` emitted an "informational" diagnostic when I didn't annotate it, which is why I had, but I'm happy to just ignore it instead.

app/terminal/terminal.tsx
bduffany: Currently, we (mostly) treat empty / zero values, undefined, and null the same, so `typeof` is rarely needed.

Eventually I think it'd be nice to decide on a single value to represent a non-present optional value (either `undefined` or `null`) and enable strict null checks in our `tsconfig.json`. Then, when trying to decide whether a value is present, we can just check `!== null` (or `!== undefined`).

bduffany (approved):
awesome :)

app/invocation/invocation.tsx
```diff
@@ -264,6 +264,18 @@ export default class InvocationComponent extends React.Component<Props, State> {
               value={this.getBuildLogs()}
               loading={this.areBuildLogsLoading()}
               expanded={activeTab == "log"}
+              fullLogsFetcher={() => {
```
siggisim, line 267: Should we pull this out into a function? Seems like embedding this much JS into a render function seems like an antipattern.

app/terminal/terminal.tsx
```diff
@@ -72,7 +80,27 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
...
   handleDownloadClicked() {
-    var element = document.createElement("a");
+    if (this.props.fullLogsFetcher) {
+      this.setState({ isLoadingFullLog: true });
+      const element = document.createElement("a");
+      this.props
+        .fullLogsFetcher()
+        .then((fullLog: string) => {
+          const unstyledLogs = fullLog.replace(ANSI_STYLES_REGEX, "");
+          element.setAttribute("href", "data:text/plain;charset=utf-8," + encodeURIComponent(unstyledLogs));
```
siggisim, line 90: Seems like this is all pretty much the same code as below - should we pull it into a function?

app/terminal/terminal.tsx
```diff
@@ -72,7 +80,27 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
   }
 
   handleDownloadClicked() {
-    var element = document.createElement("a");
+    if (this.props.fullLogsFetcher) {
+      this.setState({ isLoadingFullLog: true });
+      const element = document.createElement("a");
```
siggisim, line 85: This can happen inside of the `then` block

app/invocation/invocation.tsx
```diff
@@ -264,6 +264,18 @@ export default class InvocationComponent extends React.Component<Props, State> {
...
+                return rpcService.service
+                  .getEventLogChunk(
+                    new eventlog.GetEventLogChunkRequest({
+                      invocationId: this.props.invocationId,
+                      minLines: 2147483647, // int32 max value: max number of lines we can request.
```
siggisim, line 272: Should we pull this into a constant?

app/terminal/terminal.tsx
tempoz: I left it out so that it could be referenced in the `finally` block, with the intention of putting the `remove` in the `finally` block being that we could still remove the element if we encountered an exception after it was added. But that's probably overly paranoid and we're not doing it below, so I'll swap it.

app/terminal/terminal.tsx
```diff
@@ -72,7 +80,27 @@ export default class TerminalComponent extends React.Component<TerminalProps, Te
...
+          element.setAttribute("href", "data:text/plain;charset=utf-8," + encodeURIComponent(unstyledLogs));
+          element.setAttribute("download", "build_logs.txt");
+          element.style.display = "none";
+          document.body.appendChild(element);
+          element.click();
+        })
+        .catch((e) => errorService.handleError(e))
+        .finally(() => {
+          this.setState({ isLoadingFullLog: false });
+          element.remove();
```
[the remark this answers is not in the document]
tempoz: Done.

app/terminal/terminal.tsx
tempoz: Done.

siggisim (approved):

PR #1257, opened 2021-11-23
Fix jankiness / network waterfall while loading SVGs during initial page load

bduffany:
Before:

https://user-images.githubusercontent.com/2414826/143095274-c7eb8c8d-e651-4dc7-a8f5-534539a3d785.mp4

After:

https://user-images.githubusercontent.com/2414826/143095288-b6d295da-a034-4932-8f27-dbf58813ab28.mp4

We currently serve SVGs uncached which results in a waterfall of requests while the app is loading, causing jank and network congestion (each SVG request uses up a connection from the client's connection pool and has a fixed overhead due to HTTP headers etc.).

Rather than trying to cache these SVGs, the fix in this PR is to use icons as code (from the `lucide-react` package) rather than using `img` elements, where possible, effectively embedding them in the app bundle / chunks.

This approach has a few advantages compared to `img src`:
* No need to further slow down / complicate the bundling process/logic by including SVGs in the bundle and then figuring out how to append the resulting hash to all SVG URLs
* SVG elements can now be styled directly with CSS, with no need to create duplicate icons for each style variant (color, stroke width, etc.)
* Using a new icon is easier: just import it from `lucide-react` -- no need to download from lucide.dev
* Strong typing means it's not possible to reference an icon that doesn't exist

`lucide-react` is tree-shakeable so this only increases the bundle size according to the icons we actually use. With this PR, it increases the compressed bundle size by 4 KB (0.2% of the total compressed bundle size).

In follow-up PRs I can go through and update the rest of the icons; wanted to try this out as a POC and make sure we're happy with the overall approach here before going too crazy.

---

**Version bump**: Patch

### DIFF
--- a/app/alert/BUILD
+++ b/app/alert/BUILD
@@ -10,6 +10,7 @@ ts_library(
     ]),
     deps = [
         "@npm//@types/react",
+        "@npm//lucide-react",
         "@npm//react",
         "@npm//rxjs",
     ],
--- a/app/alert/alert.tsx
+++ b/app/alert/alert.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { Subscription } from "rxjs";
 import alertService, { Alert, AlertType } from "./alert_service";
+import { XCircle, CheckCircle } from "lucide-react";
 
 interface State {
   alert?: Alert;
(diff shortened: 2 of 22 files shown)


siggisim (approved):
This is cool. Once we've removed them all, we can delete the checked-in svgs that are supported by lucide (which should be most if not all).

tylerwilliams:
Just wanted to say this is really cool, and I really dig the detailed thoughtful PR description too.

PR #1262, opened 2021-11-29
Migrate more icons to lucide-react

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/app/root/root.tsx
+++ b/app/root/root.tsx
@@ -41,6 +41,7 @@ export default class RootComponent extends React.Component {
       next: (user: User) => this.setState({ ...this.state, user }),
     });
     faviconService.setDefaultFavicon();
+    window._preferences = this.state.preferences;
   }
 
   componentDidMount() {
--- a/enterprise/app/root/root.tsx
+++ b/enterprise/app/root/root.tsx
@@ -70,6 +70,7 @@ export default class EnterpriseRootComponent extends React.Component {
     authService.register();
     router.register(this.handlePathChange.bind(this));
     faviconService.setDefaultFavicon();
+    (window as any)._preferences = this.state.preferences;
   }
 
   componentDidMount() {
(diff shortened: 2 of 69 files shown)


app/root/root.tsx
```diff
@@ -41,6 +41,7 @@ export default class RootComponent extends React.Component {
       next: (user: User) => this.setState({ ...this.state, user }),
     });
     faviconService.setDefaultFavicon();
+    window._preferences = this.state.preferences;
```
siggisim, line 44: intentional?

enterprise/app/root/root.tsx
```diff
@@ -70,6 +70,7 @@ export default class EnterpriseRootComponent extends React.Component {
     authService.register();
     router.register(this.handlePathChange.bind(this));
     faviconService.setDefaultFavicon();
+    (window as any)._preferences = this.state.preferences;
```
siggisim, line 73: same as above

siggisim (approved):
Thanks for doing this!

After clicking around in dev, I found a couple of places that look a little funky - not sure if you want to handle that in this PR or a separate one (or maybe they're already taken care of).

Commit and branch icons on the history page are darker than others:
Screen Shot 2021-11-29 at 10 24 56 AM

The filter icon the raw tab is a little funky:
Screen Shot 2021-11-29 at 10 25 15 AM

app/root/root.tsx
bduffany: Ah yes, I was using this to be able to quickly switch between dense and non-dense on a given page. Was calling `_preferences.toggleDenseMode()` in the console

bduffany:
> Thanks for doing this!
> 
> After clicking around in dev, I found a couple of places that look a little funky - not sure if you want to handle that in this PR or a separate one (or maybe they're already taken care of).
> 
> Commit and branch icons on the history page are darker than others: Screen Shot 2021-11-29 at 10 24 56 AM
> 
> The filter icon the raw tab is a little funky: Screen Shot 2021-11-29 at 10 25 15 AM

The filter icons are fixed in this PR -- I'll take another look at the git icons in a separate PR

PR #1263, opened 2021-11-29
Remove unused grafana dashboards

tylerwilliams:
Handled gRPC requests per second by status is not used on either executors or workflow executors.

### DIFF
--- a/tools/metrics/grafana/dashboards/buildbuddy.json
+++ b/tools/metrics/grafana/dashboards/buildbuddy.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": 1,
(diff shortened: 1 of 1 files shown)


bduffany:
Hm I tried previewing locally, and it looks like those dashboards are still there:

image

I'm having a hard time understanding the changes made to the JSON, but it looks like there are some changes to the "selected" state of the "Jobs" selector? Rather than changing the jobs selector, I think maybe we'd want to update that "gRPC" panel to only run for job=buildbuddy-app, rather than repeating for all jobs -- i.e. hard-code `job="buildbuddy-app"` in the prom queries, and disable repeating per-job in the row settings (gear icon)

bduffany (approved):
LGTM

PR #1270, opened 2021-12-01
Graph the avg last evicted item age

tylerwilliams:
Graph the avg last evicted item age.

Looks like this on dev: 
image


bduffany (approved):
LGTM, nice!

PR #1271, opened 2021-12-01
Add support for persistent workers that support the JSON protocol

siggisim:
Adds a remote execution property called `persistentWorkerProtocol` that supports two values (per the remote persistent workers proposal):
- `json`
- `proto`

If unset, defaults to `proto`. If set to a different value, returns an `FailedPrecondition` error `unsupported persistent worker type`.

Adds tests for `json` protocol.

Follows the protocol description describted in bazel persistent worker's documentation: https://docs.bazel.build/versions/main/creating-workers.html#work-responses

Namely:
- Each protocol buffer is preceded by its length in varint format (see MessageLite.writeDelimitedTo().
- JSON requests and responses are not preceded by a size indicator.
- JSON requests uphold the same structure as the protobuf, but use standard JSON.
- In order to maintain the same backward and forward compatibility properties as protobuf, JSON workers must tolerate unknown fields in these messages, and use the protobuf defaults for missing values.
- Bazel stores requests as protobufs and converts them to JSON using protobuf’s JSON format

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -83,6 +85,11 @@ const (
 	hitStatusLabel = "hit"
 	// Label assigned to runner pool request count metric for unfulfilled requests.
 	missStatusLabel = "miss"
+
+	// Value for persisent workers that support the JSON persistent worker protocol.
+	workerProtocolJSONValue = "json"
+	// Value for persisent workers that support the protobuf persistent worker protocol.
+	workerProtocolProtobufValue = "protobuf"
```
brentleyjones, line 92: Since the values for `requires-worker-protocol` are `json` and `proto`, and `WorkerProtocolFormat` is `JSON` and `PROTO`, I say we should go with `proto` here. I know it bucks the current design doc, but afaik no one else has implemented this side of it, and I'll comment in it with this same reasoning.

brentleyjones: I updated my bazel patch here: https://github.com/brentleyjones/bazel/tree/5.0/remote-persistent-workers

siggisim: Agreed - done.

brentleyjones (approved):
I tested this with a couple different JSON based workers and it worked.

PR #1272, opened 2021-12-01
Turn on selfauth for the local enterprise config

tempoz:
Enable self-auth and comment out oauth from google in the local enterprise config to simplify
running enterprise locally.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1275, opened 2021-12-02
Add flag to allow using local images during development

bduffany:
This allows quickly testing local image builds without needing to push to a remote registry.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1279, opened 2021-12-06
Avoid writing duplicate entries to key value store

luluz66:
In EventLogWriter, do not write to key value store if the chunk value
hasn't been changed.

### DIFF
--- a/server/eventlog/eventlog.go
+++ b/server/eventlog/eventlog.go
@@ -324,16 +324,16 @@ func (w *EventLogWriter) writeChunkToKeyValStore(ctx context.Context, writeReque
 		ChunkId: chunkId,
 		Buffer:  append(chunk, volatileTail...),
 	}
-
-	if !proto.Equal(w.lastChunk, curChunk) {
-		keyval.SetProto(
-			ctx,
-			w.keyValueStore,
-			w.eventLogPath,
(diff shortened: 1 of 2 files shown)


server/eventlog/eventlog.go
```diff
@@ -320,6 +305,35 @@ type WriteCloserWithContext interface {
...
+	if chunkId == chunkstore.ChunkIndexAsStringId(math.MaxUint16) {
+		keyval.SetProto(ctx, w.keyValueStore, w.eventLogPath, nil)
+		return
+	}
+	curChunk := &elpb.LiveEventLogChunk{
+		ChunkId: chunkId,
+		Buffer:  append(chunk, volatileTail...),
+	}
+
+	if !proto.Equal(w.lastChunk, curChunk) {
```
tylerwilliams, line 328: uber nit: I think it would be clearer to early return if they are equal (like you do above), then unnest the keyval.SetProto, but up to you.

tylerwilliams (approved):
LGTM

(make sure Zoey gets a chance to look at this before submitting tho)

server/eventlog/eventlog.go
luluz66: Done.

PR #1280, opened 2021-12-06
Show group-owned connected executors even if "use self-owned Linux executors" is not enabled

bduffany:
* Always show all connected executors, even if "use self-owned Linux executors" is not enabled in settings. This fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/980.
* To avoid confusion while also simplifying the UX in the case where you have Linux executors registered, but haven't enabled "use self-owned Linux executors", show an explicit callout in this case to go to the settings page and enable self-hosted executors.
* The page was starting to get noisy and crowded showing both the full executors list and full setup instructions below it -- so this PR also splits the page into separate "Status" and "Setup" tabs. The tabbed UX should not result in any unnecessary extra clicks in most cases, since we default to the "setup" tab if no executors are currently registered.

Misc:
* Fix link handling on executors page (links were causing a full page load instead of router navigation)
* Improve contrast ratio on tabs (text is hard to read) and add stronger signifiers that they are clickable

Screenshots:

---

image

---

image

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/980

### DIFF
--- a/enterprise/app/executors/executors.tsx
+++ b/enterprise/app/executors/executors.tsx
@@ -10,6 +10,7 @@ import { bazel_config } from "../../../proto/bazel_config_ts_proto";
 import { FilledButton } from "../../../app/components/button/button";
 import router from "../../../app/router/router";
 import Select, { Option } from "../../../app/components/select/select";
+import { AlertCircle } from "lucide-react";
 
 enum FetchType {
   Executors,
@@ -17,9 +18,11 @@ enum FetchType {
   BazelConfig,
 }
 
-function onClickLink(href: string, e: React.MouseEvent<HTMLAnchorElement>) {
-  e.preventDefault();
-  router.navigateTo(href);
+function linkHandler(href: string) {
+  return (e: React.MouseEvent) => {
+    e.preventDefault();
+    router.navigateTo(href);
+  };
 }
 
 interface ExecutorDeployProps {
@@ -80,8 +83,19 @@ class ExecutorSetup extends React.Component<ExecutorSetupProps> {
   state = { selectedExecutorKeyIdx: 0 };
 
   render() {
+    // If the user already has executors running, show the "short" version of the setup.
+    if (this.props.user.selectedGroup.useGroupOwnedExecutors && this.props.executors.length) {
+      return (
+        <>
+          <h2>Deploy additional executors</h2>
+          <ExecutorDeploy executorKeys={this.props.executorKeys} schedulerUri={this.props.schedulerUri} />
+        </>
+      );
+    }
+
     return (
       <>
+        <h1>Set up self-hosted executors</h1>
         <hr />
         <h2>1. Create an API key for executor registration</h2>
         {this.props.executorKeys.length == 0 && (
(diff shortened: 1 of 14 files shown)


enterprise/app/executors/executors.tsx
```diff
@@ -288,47 +292,76 @@ export default class ExecutorsComponent extends React.Component<Props, State> {
...
-            <h1>No self-hosted executors connected.</h1>
-            <p>Self-hosted executors are enabled, but there are no executors connected.</p>
-            <p>Follow the instructions below to deploy your executors.</p>
-            <ExecutorSetup
-              user={this.props.user}
-              executorKeys={this.state.executorKeys}
-              executors={this.state.nodes}
-              schedulerUri={this.state.schedulerUri}
-            />
+    const linuxNodes = this.state.nodes.filter((node) => node.os.toLowerCase() === "linux");
```
siggisim, line 301: I don't love the idea of the front-end having to know about special case linux vs mac executor logic.

Also, whether or not this special case logic applies depends on whether or not this config option is set: https://github.com/buildbuddy-io/buildbuddy/blob/master/server/config/config.go#L255

Maybe GetExecutionNodes could return something like whether or not the returned executor is enabled?

siggisim (commented):
I generally like this, but I'm a little worried about the on-prem case where `force_user_owned_darwin_executors` isn't set. More details in my comment.

I'd also like to add instructions on setting up self hosted mac executors, linux executors using the binary (instead of the docker image), and kubernetes executors (via helm) - but we can do that separately.

enterprise/app/executors/executors.tsx
bduffany: SG -- I added an "enabled" bit and surfaced it in a new proto `GetExecutionNodesResponse.Executor` -- I didn't add the bit to the `ExecutionNode` because currently that proto just contains static configuration that gets stored in Redis, and don't want ppl to accidentally rely on that enabled bit for anything other than the executors page.

For now the UI assumes that if at least one executor is disabled, then it's because self-hosted executors are not enabled in settings. If there are other reasons later on, we can maybe change this to a `status` enum or add a `disabled_reason` enum or something like that.

image

enterprise/app/executors/executors.tsx
```diff
@@ -288,47 +292,74 @@ export default class ExecutorsComponent extends React.Component<Props, State> {
...
+                      </a>
+                    </FilledButton>
+                  </div>
+                </div>
+              </div>
+            )}
+            <ExecutorsList executors={this.state.nodes} />
+            {!this.state.nodes.length && this.props.user.selectedGroup.useGroupOwnedExecutors && (
+              <div className="empty-state">
+                <h1>No self-hosted Linux executors are connected.</h1>
```
siggisim, line 341: nit: should we remove the word `Linux` here?

siggisim (approved):
Slick!

vadimberezniker (approved):
Nice changes!

enterprise/app/executors/executors.tsx
bduffany: Good catch - done

PR #1281, opened 2021-12-06
Remove Redis instance selector from dashboard.

vadimberezniker:
There's currently only one Redis instance per region.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1282, opened 2021-12-07
Create digest component and use it on the actions card

siggisim:
This displays (colored) hashes and files sizes for all action inputs, which makes visually diffing action inputs easier.

Also applied these to actions outputs and the executions table.

Screenshot:
Screen Shot 2021-12-06 at 4 28 24 PM

---

**Version bump**: Minor


bduffany (approved):
Nice - this seems handy for comparing two action pages side by side.

PR #1283, opened 2021-12-07
Don't lookup cache or db on empty API key

luluz66:


enterprise/server/auth/auth.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lru"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
-	"github.com/buildbuddy-io/buildbuddy/server/util/request_context"
+	requestcontext "github.com/buildbuddy-io/buildbuddy/server/util/request_context"
```
vadimberezniker, line 23: This would be better saved for a separate PR.
Instead of introducing an alias here, it might be better to change the package name to match the filename.

enterprise/server/auth/auth.go
```diff
@@ -568,6 +568,9 @@ func lookupUserFromSubID(env environment.Env, ctx context.Context, subID string)
 }
 
 func (a *OpenIDAuthenticator) lookupAPIKeyGroupFromAPIKey(ctx context.Context, apiKey string) (interfaces.APIKeyGroup, error) {
+	if apiKey == "" {
+		return nil, status.UnauthenticatedError("missing API keys")
```
vadimberezniker, line 572: keys -> key

vadimberezniker (approved):

enterprise/server/auth/auth.go
luluz66: Thank you for catching this. vim somehow auto adds it. Reverted

enterprise/server/auth/auth.go
luluz66: Done

PR #1284, opened 2021-12-07
Add support for specifying an "session affinity key" in the Run API.

vadimberezniker:
This allows clients to specify preference as to which Bazel instance
their requests get routed.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

Would be nice to add a test for this?

bduffany (commented):
WDYT about implementing this as a suffix on the remote instance name?

vadimberezniker:
> LGTM
> 
> Would be nice to add a test for this?

added tests

vadimberezniker:
> WDYT about implementing this as a suffix on the remote instance name?

I think eventually we'll want to be able to propagate the instance name down the hosted bazel process. 
When that's implemented, relying on instance name for routing would unnecessarily cause cache misses.

PR #1288, opened 2021-12-08
Request `offline_access` scope from non-Google OIDC providers in order to retrieve refresh token

siggisim:
This fixes the issue where Okta users registered through OIDC will be logged out every 60 minutes.

Without this scope, Okta does not send a refresh_token with the authorize response.

Google auth does not support this scope, but instead uses a url parameter `access_type=offline` which we already set thanks to oauth2.AccessTypeOffline on line 91 of auth.go:
https://github.com/coreos/go-oidc/blob/v2.2.1/oidc.go#L30

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1289, opened 2021-12-08
Retrieve ACKs from build event proxy

glukasiknuro:
Found out that when using proxy some streams were not complete. It's
because the build event server to which traffic was proxied was sending
ACKs messages along the way, but buildbuddy was not retrieving it,
so those were accumulated till some buffers were filled up.

---

**Version bump**: Patch

### DIFF
--- a/server/build_event_protocol/build_event_proxy/build_event_proxy.go
+++ b/server/build_event_protocol/build_event_proxy/build_event_proxy.go
@@ -2,7 +2,7 @@ package build_event_proxy
 
 import (
 	"context"
-    "io"
+	"io"
 	"sync"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
LGTM

Hey @glukasiknuro, this looks great, thanks for sending a PR to fix!

tylerwilliams:
@glukasiknuro Looks like the checkstyle presubmit is failing -- would you mind running go fmt on this and then we can merge it? LMK if this is an issue.

glukasiknuro:
> @glukasiknuro Looks like the checkstyle presubmit is failing -- would you mind running go fmt on this and then we can merge it? LMK if this is an issue.

Done. I guess it needs "Approve and run" to re-run checkstyle.

tylerwilliams:
You have good timing; this CL went out in our release (https://github.com/buildbuddy-io/buildbuddy/releases/tag/v2.7.3) today :)

Thanks again for the PR!

PR #1290, opened 2021-12-08
Split out mac release workflow (to enable re-running independently) and pin xcode version

siggisim:
This will allow us to re-run mac workflows without creating a new release each time.

Also pins Xcode version.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1293, opened 2021-12-09
Avoid panic when Xcode is not installed

luluz66:
Return error for NewXCodeLocator when Xcode cannot be found.
If NewXCodeLocator returns an error log.Fatal

### DIFF
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -142,7 +142,11 @@ func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChec
 		log.Infof("No authentication will be configured: %s", err)
 	}
 
-	realEnv.SetXCodeLocator(xcode.NewXcodeLocator())
+	xl, err := xcode.NewXcodeLocator()
+	realEnv.SetXCodeLocator(xl)
+	if err != nil {
+		log.Fatalf("Failed to set XCodeLocator: %s", err)
+	}
 
 	if gcsCacheConfig := configurator.GetCacheGCSConfig(); gcsCacheConfig != nil {
 		opts := make([]option.ClientOption, 0)
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -143,10 +143,10 @@ func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChec
 	}
 
 	xl, err := xcode.NewXcodeLocator()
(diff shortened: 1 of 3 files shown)


enterprise/server/cmd/executor/executor.go
```diff
@@ -142,7 +142,11 @@ func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChec
 		log.Infof("No authentication will be configured: %s", err)
 	}
 
-	realEnv.SetXCodeLocator(xcode.NewXcodeLocator())
+	xl, err := xcode.NewXcodeLocator()
+	realEnv.SetXCodeLocator(xl)
+	if err != nil {
```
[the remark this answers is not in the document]
tylerwilliams: yup -- check, then use

server/xcode/xcode_darwin.go
```diff
@@ -78,11 +79,15 @@ func (x *xcodeLocator) xcodeVersionForVersionString(version string) *xcodeVersio
...
 // Very losely based on https://github.com/bazelbuild/bazel/blob/master/tools/osx/xcode_locator.m
-func (x *xcodeLocator) locate() {
+func (x *xcodeLocator) locate() error {
 	bundleID := stringToCFString(xcodeBundleID)
 	defer C.CFRelease(C.CFTypeRef(bundleID))
 
 	urlsPointer := C.LSCopyApplicationURLsForBundleIdentifier(bundleID, nil)
+
+	if urlsPointer == C.CFArrayRef(unsafe.Pointer(nil)) {
+		return fmt.Errorf("cannot find XCode bundle: %q", xcodeBundleID)
```
tylerwilliams, line 89: Could we use something from status instead of minting a new fmt.Errorf? FailedPreconditionError seems like it would fit

tylerwilliams (changes requested):

server/xcode/xcode_darwin.go
siggisim: could also give the user a hint like - Make sure Xcode is installed

brentleyjones: nit: It's Xcode not XCode

enterprise/server/cmd/executor/executor.go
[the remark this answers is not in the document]
luluz66: Done.

server/xcode/xcode_darwin.go
luluz66: All Done.

tylerwilliams (approved):
LGTM

PR #1295, opened 2021-12-09
Switch Grafana to global prometheus instance.

vadimberezniker:
Adds 'region' filter to all queries.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

Excited for this!

PR #1300, opened 2021-12-10
Quick fix of release.py

bduffany:
Quick fix of release.py using BuildBuddy Code

### DIFF
--- a/release.py
+++ b/release.py
@@ -26,7 +26,7 @@ except NameError:
     pass
 
 """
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -26,7 +26,7 @@
     pass
 
 """
-release.py - A simple script to create a release.
+release.py - A simple script to create a release
```
tylerwilliams, line 29: add a period.

PR #1301, opened 2021-12-10
Revert "Add `dockerUser` platform prop to control Docker user ID and output dir permissions (#1294)"

bduffany:
Decided to go with an alternate approach of specifying a platform prop that controls directory permissions directly.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1303, opened 2021-12-10
Make workspace dirs writable by non-root users, behind platform prop

bduffany:
---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1304, opened 2021-12-11
Add group_id to execution stage duration metric.

vadimberezniker:
Use this data to display RBE prober latency by stage to make it easier
to diagnose RBE prober failures.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1305, opened 2021-12-11
Distinguish between key not set and empty key

luluz66:
1. change ParseAPIKeyFromString to return a string pointer instead of
   a string; and thus allow us to distinguish between when the key is
   not set (i.e. 'x-buildbuddy-api-key=' is not present) vs empty key
   (i.e. the header is present)
2. also print quotes for "Invalid API Key" message, to make it more
   readable when the key is empty.

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -651,12 +651,12 @@ func (a *OpenIDAuthenticator) ParseAPIKeyFromString(input string) (string, error
 		return "", nil
 	}
 	if len(matches) != 2 {
-		return "", status.FailedPreconditionError("invalid input")
+		return "", status.UnauthenticatedError("failed to parse API key: invalid input")
 	}
 	if apiKey := matches[1]; apiKey != "" {
 		return apiKey, nil
 	}
-	return "", status.FailedPreconditionError("missing API Key")
+	return "", status.UnauthenticatedError("failed to parse API key: missing API Key")
 }
 
 func (a *OpenIDAuthenticator) AuthContextFromAPIKey(ctx context.Context, apiKey string) context.Context {
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -89,7 +89,7 @@ const (
 
 var (
 	authCodeOption []oauth2.AuthCodeOption = []oauth2.AuthCodeOption{oauth2.AccessTypeOffline, oauth2.ApprovalForce}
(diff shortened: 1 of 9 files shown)


enterprise/server/auth/auth.go
```diff
@@ -89,7 +89,7 @@ const (
 
 var (
 	authCodeOption []oauth2.AuthCodeOption = []oauth2.AuthCodeOption{oauth2.AccessTypeOffline, oauth2.ApprovalForce}
-	apiKeyRegex                            = regexp.MustCompile(APIKeyHeader + "=([a-zA-Z0-9]+)")
+	apiKeyRegex                            = regexp.MustCompile(APIKeyHeader + "=([a-zA-Z0-9]*)")
```
tylerwilliams, line 92: "*" = 0 or more strings
"+" = 1 or more strings

So this will now match "APIKeyHeader=." where before it did not.

enterprise/server/auth/auth.go
```diff
@@ -644,12 +644,19 @@ func authContextFromClaims(ctx context.Context, claims *Claims, err error) conte
...
+		return "", nil
 	}
-	return ""
+	if len(matches) != 2 {
+		return "", status.FailedPreconditionError("invalid input")
+	}
+	if apiKey := matches[1]; apiKey != "" {
+		return apiKey, nil
+	}
+	return "", status.FailedPreconditionError("missing API Key")
```
tylerwilliams, line 659: nit: do you want to return:

status.UnauthenticatedError("failed to parse API key: invalid input")
and 
status.UnauthenticatedError("failed to parse API key: missing API Key")

here instead, and then just return the err directly in your calling code?


(in general I like to return as detailed of an error as possible at the source, and then just pass ideally without modification in most cases)

enterprise/server/auth/auth_test.go
```diff
@@ -126,6 +127,55 @@ func TestAuthenticateHTTPRequest(t *testing.T) {
 	require.Equal(t, validUserToken, authCtx.Value(contextUserKey), "context user details should match details returned by provider")
 }
 
+func TestParseAPIKeyFromString(t *testing.T) {
```
tylerwilliams, line 130: love this, thanks for adding tests!

tylerwilliams (changes requested):
Couple nits, otherwise LG.

enterprise/server/auth/auth.go
luluz66: Done.

tylerwilliams (approved):
LGTM

PR #1308, opened 2021-12-13
Fix some sidebar styling issues and clean up sidebar styles

bduffany:
Fixes:
* Fix bottom spacing between profile picker (at bottom) and the bottom edge of the sidebar, which was reproducible in dense mode only
* Fix issue that icons in the group picker become squished when the group name is too long

Refactor:
* Avoid specifying margins where possible and instead use the `gap` property which now has full support in all major browsers: https://caniuse.com/gap. The gap property is cleaner than margins because it allows setting a single CSS property on the parent, rather than having to awkwardly apply margins to all children except the last child.
* Avoid specifying negative margins where possible.
* Avoid non-symmetric padding and margins where possible -- make it so that the top padding almost always matches the bottom padding and left padding almost always matches the right padding.
* Remove `.dense` specific styles where possible in favor of setting a CSS variable that controls the base padding.

Tested in Linux/Chrome, macOS/Chrome, macOS/Safari, across all combinations of {mobile, dense mode, sidebar footer expanded}.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/sidebar/sidebar.css
+++ b/enterprise/app/sidebar/sidebar.css
@@ -211,10 +211,6 @@
   opacity: 1;
 }
 
-.sidebar .org-picker-item svg {
-  flex-shrink: 0;
-}
-
 /** Dense **/
 
 .dense .sidebar {
--- a/enterprise/app/sidebar/sidebar.css
+++ b/enterprise/app/sidebar/sidebar.css
@@ -6,15 +6,40 @@
   display: flex;
   flex-direction: column;
   box-sizing: border-box;
-  padding: 32px;
   overflow: auto;
   overflow-x: hidden;
   font-size: 18px;
   user-select: none;
 }
 
+.sidebar {
+  --sidebar-padding: 32px;
+  color: #fff;
+}
+
+.dense .sidebar {
+  --sidebar-padding: 16px;
+}
+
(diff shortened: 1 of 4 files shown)


enterprise/app/sidebar/sidebar.css
```diff
@@ -172,11 +211,14 @@
   opacity: 1;
 }
 
+.sidebar .org-picker-item svg {
```
siggisim, line 214: should we apply this to all icons?

siggisim (approved):

enterprise/app/sidebar/sidebar.tsx
```diff
@@ -280,6 +280,7 @@ export default class SidebarComponent extends React.Component {
                   </div>
                 )}
               </div>
+              <hr />
```
siggisim, line 283: how does this look?

enterprise/app/sidebar/sidebar.css
bduffany: sg, done

enterprise/app/sidebar/sidebar.tsx
bduffany: Same as before -- I added an override to remove the default user agent styles:

image

PR #1314, opened 2021-12-13
Proto changes for compressed blobs support

bduffany:
Will be convenient to have this merged as a first step towards supporting compressed blobs.

Copying the `remote_execution.proto` changes from upstream that are relevant to compression.

This API change is backwards compatible. We already correctly return `supported_compressors = <empty list>, supported_batch_update_compressors = <empty list>` in the capabilities proto, which is equivalent to supporting only the identity compression type. We also already return InvalidArgument when we see resource names for compressed blobs.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1315, opened 2021-12-13
Use pod names instead of IPs to identify servers in golang graphs.

vadimberezniker:
The pod names are slightly more useful than IPs.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1316, opened 2021-12-14
Add user-specific github token

siggisim:
This allows the user to link a user-specific Github token for use in the Code UI instead of using the org-wide one.

---

**Version bump**: Patch

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -220,6 +220,7 @@ type User struct {
 	Email     string
 	ImageURL  string
 
+	// User-specific Github token (if linked).
 	GithubToken string
 
 	// Group roles are used to determine read/write permissions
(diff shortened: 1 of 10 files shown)


server/tables/tables.go
```diff
@@ -220,6 +220,8 @@ type User struct {
 	Email     string
 	ImageURL  string
 
+	GithubToken string
```
tylerwilliams, line 223: nit: add a comment that this is used if set otherwise the group one is used?

tylerwilliams (approved):
LGTM

server/tables/tables.go
siggisim: There's currently no fallback behavior, but I added a comment.

PR #1320, opened 2021-12-14
Integrate some useful kube state metrics into dashboard.

vadimberezniker:
- Show # of ready pods in Status per-job panels.
 - Show autoscaler target in autoscaler graph
 - Add graph of unexpected restarts in Status row.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1323, opened 2021-12-14
Tag proto build targets

tempoz:
Tag proto build targets so they can be built separately to populate the `.pb.go` files for
`go mod tidy`.

---

**Version bump**: None


siggisim:
Not sure exactly what the use-case is here, but instead of tagging these, could we select them with a `bazel query` like:
```
bazel query 'kind(go_proto_.*, //...)'
```
or
```
bazel query 'kind(go_proto_library, //...)'
```

You can then build just these targets with something like 
```
bazel build $(bazel query 'kind(go_proto_library, //...)')
```

PR #1325, opened 2021-12-14
Add libraries to wrap I/O for tracing

tempoz:
Adds (but does not yet use) libraries that wrap common I/O operations with tracing.

---

**Version bump**: Minor

### DIFF
--- a/nogo_config.json
+++ b/nogo_config.json
@@ -55,6 +55,9 @@
     "only_files": {
       "/server/": "Server code",
       "/enterprise/server/": "Enterprise server code"
+    },
+    "exclude_files": {
+      "ctxio": "Intentionally adds context to standard method signatures"
     }
   },
   "stringintconv": {
--- a/server/util/tracing/ctxio/BUILD
+++ b/server/util/tracing/ctxio/BUILD
@@ -1,0 +1,11 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "ctxio",
+    srcs = ["ctxio.go"],
+    importpath = "github.com/buildbuddy-io/buildbuddy/server/util/tracing/ctxio",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//server/util/tracing/span",
+    ],
+)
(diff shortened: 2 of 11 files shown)


tylerwilliams (changes requested):
Thanks for splitting this out, it's definitely clearer now :) Rather than going "nuclear" and rewriting whole stdlibs to trace -- could we start with an mvp that just adds tracing code to the commonly used stuff, like util/disk.{WriteFile,ReadFile,FileWriter,FileReader}?

I think that would cover the cache writes, which is the bulk of them and if we also took a pass through dirtools.go and made sure it was using one of those util functions in disk we'd probably get most of the remote exec ones too.

tempoz:
I agree that that would handle many cases, but we unfortunately rely heavily on `io.Copy` from Readers and to Writers throughout much of the code, which cannot be passed contexts; should we just bundle the context into the cache on creation? I read that contexts should always be passed in to the function and not be member fields if it can be avoided, but maybe this is a place where we cannot reasonably avoid it? Or maybe we can just ignore the `Copy` calls and the use of `Reader`s and `Writer`s for now...

tempoz:
Regardless, I'm working on the tracing for commonly used things now.

tylerwilliams:
For FileReader and FileWriter (and ReadFile and WriteFile), using the context provided when creating the writer is fine because we don't have any super long-lived files where this would present a problem, and if we did and the tracing didn't work for those files, that would be OK*.

For GCS and the like, the interface includes a context, so adding a span in Reader/Writer/Get/Set etc should be doable. For Reader and Writer, again we'll be using the "bundled" context, but those writes are also expected to be short-ish, so I think this is fine for the same reason as above.


* this is kind of a longer diversion, but most of the reason not to use a bundled context like that is because contexts carry deadlines within them, and if you hold onto a context that has expired and try to use it, it won't work. The funny thing about contexts is that they *also* carry k/vs, like the span information, some auth stuff, etc, and this data does not expire with the deadline, it's still there and fine to use. So if you were just using the k/v information from a bundled context, you might not care if its expired. But it's still sort of a code smell we try to avoid if possible.

PR #1328, opened 2021-12-15
Add action_result_digest to Execution proto

luluz66:
When an action succeeds, action result digest is the same as action
digest; but when an action fails, action result digest contains
result details, but action digest doesn't.


tylerwilliams (approved):
LGTM

PR #1331, opened 2021-12-15
Grafana fixes.

vadimberezniker:
- Set proper unit for RBE Prober Latency graph.
 - Fix pod regex for the ready metric to match wider range of pod names

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1332, opened 2021-12-15
Set ActionResultCache in execution proto

luluz66:
1. When remote execution fails, compute action result digest based on
   the original action digest and the invocation id. When if succeeds,
   action result digest is the same as action digest. Upload the action
   result digest
2. When we convert Execution row from db to proto, set action result
   cache using the same logic as above.

### DIFF
--- a/enterprise/server/execution_service/execution_service.go
+++ b/enterprise/server/execution_service/execution_service.go
@@ -17,6 +17,8 @@ import (
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	timestamppb "github.com/golang/protobuf/ptypes/timestamp"
 	statuspb "google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/proto"
 )
 
 type ExecutionService struct {
@@ -73,8 +75,19 @@ func tableExecToProto(in tables.Execution) (*espb.Execution, error) {
 		return nil, err
 	}
 
+	var actionResultDigest *repb.Digest
+	if in.StatusCode == int32(codes.OK) {
+		actionResultDigest = proto.Clone(d).(*repb.Digest)
+	} else {
+		actionResultDigest, err = digest.AddInvocationIDToDigest(d, in.InvocationID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	out := &espb.Execution{
-		ActionDigest: d,
+		ActionDigest:       d,
+		ActionResultDigest: actionResultDigest,
 		Status: &statuspb.Status{
 			Code:    in.StatusCode,
 			Message: in.StatusMessage,
--- a/enterprise/server/execution_service/execution_service.go
+++ b/enterprise/server/execution_service/execution_service.go
@@ -12,13 +12,13 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/query_builder"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/golang/protobuf/ptypes"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/proto"
 
 	espb "github.com/buildbuddy-io/buildbuddy/proto/execution_stats"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	timestamppb "github.com/golang/protobuf/ptypes/timestamp"
 	statuspb "google.golang.org/genproto/googleapis/rpc/status"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/protobuf/proto"
 )
 
 type ExecutionService struct {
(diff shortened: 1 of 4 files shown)


enterprise/server/execution_service/execution_service.go
```diff
@@ -73,8 +75,19 @@ func tableExecToProto(in tables.Execution) (*espb.Execution, error) {
 		return nil, err
 	}
 
+	var actionResultDigest *repb.Digest
+	if in.StatusCode == int32(codes.OK) {
+		actionResultDigest = proto.Clone(d).(*repb.Digest)
+	} else {
+		actionResultDigest, err = digest.AddInvocationIDToDigest(d, in.InvocationID)
```
bduffany, line 82: Would it make sense to instead directly store the ActionResult proto in the CAS when the action fails (as opposed to fudging the hash and storing it in AC)? Then, I think we wouldn't have to worry about Bazel returning the cached action result of a failed action, since cached actions should only ever be served from AC (IIUC?) and it would avoid the need to adding the invocation ID to the digest. I think we would just need to send the ActionDigest as part of the last Operation in the execution stream, so that the ExecutionServer can store it in the table. + @tylerwilliams

bduffany (changes requested):

enterprise/server/execution_service/execution_service.go
luluz66: I think the current solution is simpler. 
1) The current solution doesn't require us to store additional field in the db. if we store the failed action result in CAS, we would somehow need to store the failed action result digest in the execution table. 
2) The FE code is simpler, fetch the action result from action result digest, if it's not set fall back to action digest. If we store the failed action result in CAS. I would imagine the fetch action result cold would be a bit hairy (if we only store the failed action result in CAS; if we store both successful and failed action result in CAS, then there are duplicate stuff between AC and CAS) 
3) With the current solution, bazel would not be able to get the cached action result of a failed action, even though it's in the AC.

tylerwilliams: +1  (for the reasons lulu mentioned)

I prefer the CL as written to trying to store some ActionResults in the CAS and some in the AC. (Which, btw, would also require munging the digests because the Action is already stored under the digest in the CAS...)

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -340,7 +340,14 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, task *repb.E
...
 
-	if !task.GetAction().GetDoNotCache() && cmdResult.Error == nil && cmdResult.ExitCode == 0 {
+	if !task.GetAction().GetDoNotCache() {
+		if cmdResult.Error != nil || cmdResult.ExitCode != 0 {
+			resultDigest, err := digest.AddInvocationIDToDigest(req.GetActionDigest(), task.GetInvocationId())
+			if err != nil {
+				return finishWithErrFn(status.UnavailableErrorf("Error uploading action result: %s", err.Error()))
+			}
+			adInstanceDigest = digest.NewInstanceNameDigest(resultDigest, req.GetInstanceName())
+		}
```
tylerwilliams, line 350: nit: since it's a little mysterious why we'd do this -- could you add a comment explaining it? Maybe something like:

// If the action failed, upload information about the error via a failed ActionResult under an 
// invocation-specific digest, which will not ever be seen by bazel but may be viewed via the BuildBuddy UI.

enterprise/server/execution_service/execution_service.go
```diff
@@ -17,6 +17,8 @@ import (
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	timestamppb "github.com/golang/protobuf/ptypes/timestamp"
 	statuspb "google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/proto"
```
tylerwilliams, line 21: nit: i assume vim did this, but could you please put the "unaliased" imports up above in a separate section from the aliased ones?

tylerwilliams (approved):
LGTM (w/ nits)

enterprise/server/execution_service/execution_service.go
bduffany: Sounds good, thanks for the detailed explanation :)

enterprise/server/remote_execution/executor/executor.go
luluz66: Done

server/remote_cache/digest/digest.go
```diff
@@ -115,6 +115,19 @@ func Compute(in io.Reader) (*repb.Digest, error) {
 	}, nil
 }
 
+func AddInvocationIDToDigest(digest *repb.Digest, invocationID string) (*repb.Digest, error) {
```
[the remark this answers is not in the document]
luluz66: Done

enterprise/server/execution_service/execution_service.go
luluz66: Done

tylerwilliams (approved):
LGTM

bduffany (approved):

PR #1336, opened 2021-12-15
Change restart graph to show delta instead of cumulative restarts.

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1338, opened 2021-12-16
Make remote_Execution_test pass on darwin arm64

luluz66:
1. Pass the runtime os and arch to Command.Platform.Properties
2. Set XcodeLocator in test env
3. set default_xcode_version in executor test config

### DIFF
--- a/enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
+++ b/enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
@@ -581,7 +581,7 @@ func (r *Env) addExecutor(options *ExecutorOptions) *Executor {
 	env.SetAuthenticator(r.newTestAuthenticator())
 	xl, err := xcode.NewXcodeLocator()
 	if err != nil {
-		assert.FailNowf(r.t, "Failed to set XCodeLocator: %s", err.Error())
+		assert.FailNowf(r.t, "Failed to set XcodeLocator: %s", err.Error())
 	}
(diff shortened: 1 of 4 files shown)


enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
```diff
@@ -577,6 +579,11 @@ func (r *Env) addExecutor(options *ExecutorOptions) *Executor {
 	env.SetRemoteExecutionClient(repb.NewExecutionClient(clientConn))
 	env.SetSchedulerClient(scpb.NewSchedulerClient(clientConn))
 	env.SetAuthenticator(r.newTestAuthenticator())
+	xl, err := xcode.NewXcodeLocator()
+	if err != nil {
+		assert.FailNowf(r.t, "Failed to set XCodeLocator: %s", err.Error())
+	}
+	env.SetXcodeLocator(xl)
```
tylerwilliams, line 586: will this cause issue when run on linux?

luluz66: It shouldn't. This is the same code as in enterprise/server/cmd/executor/executor.go. If it's on linux, code in xcode.go will be invoked, and if it's on mac, xcode_darwin.go will be invoked. 

The remote_execution_test succeeds in the workflow; so I assume it works on linux.

bduffany (approved):
LGTM, thanks for fixing this :)

PR #1341, opened 2021-12-16
Add tracing around s3 and gcs cache network activity

tempoz:
Adds tracing for GCS and S3 caches. For reading from GCS with a reader, we depend on the tracing code
already present in google's GCS libs. An alternative strategy (that I am happy to switch to if it is 
desired) would be to wrap the returned reader with a bundled context so we can start a span on each
call to Read.

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/backends/gcs_cache/gcs_cache.go
+++ b/enterprise/server/backends/gcs_cache/gcs_cache.go
@@ -81,6 +81,7 @@ func (g *GCSCache) createBucketIfNotExists(ctx context.Context, bucketName strin
 
 func (g *GCSCache) setBucketTTL(ctx context.Context, bucketName string, ageInDays int64) error {
 	traceCtx, spn := tracing.StartSpan(ctx)
+	spn.SetName("Attrs for GCSCache SetBucketTTL")
 	attrs, err := g.gcsClient.Bucket(bucketName).Attrs(traceCtx)
 	spn.End()
 	if err != nil {
@@ -105,6 +106,7 @@ func (g *GCSCache) setBucketTTL(ctx context.Context, bucketName string, ageInDay
 		},
 	}
 	traceCtx, spn = tracing.StartSpan(ctx)
+	spn.SetName("Update for GCSCache SetBucketTTL")
 	defer spn.End()
 	// Update the bucket TTL, regardless of whatever value is set.
 	_, err = g.gcsClient.Bucket(bucketName).Update(traceCtx, storage.BucketAttrsToUpdate{Lifecycle: &lc})
--- a/enterprise/server/backends/gcs_cache/gcs_cache.go
+++ b/enterprise/server/backends/gcs_cache/gcs_cache.go
@@ -16,6 +16,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/cache_metrics"
 	"github.com/buildbuddy-io/buildbuddy/server/util/prefix"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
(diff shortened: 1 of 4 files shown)


enterprise/server/backends/gcs_cache/gcs_cache.go
```diff
@@ -60,18 +61,28 @@ func NewGCSCache(bucketName, projectID string, ageInDays int64, opts ...option.C
...
 		return g.bucketHandle.Create(ctx, g.projectID, nil)
 	}
 	g.bucketHandle = g.gcsClient.Bucket(bucketName)
 	return nil
 }
 
 func (g *GCSCache) setBucketTTL(ctx context.Context, bucketName string, ageInDays int64) error {
+	_, spn := tracing.StartSpan(ctx)
 	attrs, err := g.gcsClient.Bucket(bucketName).Attrs(ctx)
+	spn.End()
```
tylerwilliams, line 85: can we override ctx with the ctx returned from StartSpan so spans get correctly nested?

	ctx, spn := tracing.StartSpan(ctx)
	attrs, err := g.gcsClient.Bucket(bucketName).Attrs(ctx)
	spn.End()

enterprise/server/backends/gcs_cache/gcs_cache.go
```diff
@@ -93,6 +104,8 @@ func (g *GCSCache) setBucketTTL(ctx context.Context, bucketName string, ageInDay
 			},
 		},
 	}
+	_, spn = tracing.StartSpan(ctx)
+	defer spn.End()
```
tylerwilliams, line 108: I think possibly we have two spans happening inside this function -- since they are named (automagically) after the calling function, you may want to name these two explicitly?

	ctx, spn = tracing.StartSpan(ctx, trace.WithAttributes(... somehow you put name in here....))
	defer spn.End()

tylerwilliams: (maybe doublecheck this isn't happening elsewhere too, 2 spans 1 func)

tylerwilliams (changes requested):
minor nits, otherwise lg

enterprise/server/backends/gcs_cache/gcs_cache.go
tempoz: Done!

enterprise/server/backends/gcs_cache/gcs_cache.go
tempoz: Yep, done.

enterprise/server/backends/gcs_cache/gcs_cache.go
```diff
@@ -60,18 +61,29 @@ func NewGCSCache(bucketName, projectID string, ageInDays int64, opts ...option.C
 	return g, nil
 }
 
+func (g *GCSCache) bucketExists(ctx context.Context, bucketName string) (bool, error) {
+	traceCtx, spn := tracing.StartSpan(ctx)
```
tylerwilliams, line 65: nit: can we override the name "ctx" everywhere instead of creating a new name "traceCtx"?

I think in all cases in this CL we want to use the new ctx, so this prevents someone else from mistakenly using the wrong one.

tempoz: I can change it mostly, but the places where we use two spans in one function would be bad to override the context, no? As we would then be using the context from the previous span (that had ended) to make a new span, which seems wrong?

tempoz: I guess maybe more accurately, I think I should only use the new context so long as the related span is active

tylerwilliams: ah yes, that one i agree, you're right!

enterprise/server/backends/s3_cache/s3_cache.go
```diff
@@ -107,17 +108,30 @@ func NewS3Cache(awsConfig *config.S3CacheConfig) (*S3Cache, error) {
 	return s3c, nil
 }
 
+func (s3c *S3Cache) bucketExists(ctx context.Context, bucketName string) (bool, error) {
+	ctx, spn := tracing.StartSpan(ctx)
+	_, err := s3c.s3.HeadBucketWithContext(ctx, &s3.HeadBucketInput{Bucket: aws.String(bucketName)})
+	spn.End()
+	if err == nil {
+		return true, nil
+	}
```
tylerwilliams, line 117: nit: this might be clearer with a defer?

    ctx, spn := tracing.StartSpan(ctx)
    defer spn.End()
	if _, err := s3c.s3.HeadBucketWithContext(ctx, &s3.HeadBucketInput{Bucket: aws.String(bucketName)}); err == nil {
		return true, nil
	}
    ...

enterprise/server/backends/s3_cache/s3_cache.go
```diff
@@ -423,11 +455,19 @@ func (s3c *S3Cache) Reader(ctx context.Context, d *repb.Digest, offset int64) (i
...
 	finishedWrite chan struct{}
 	timer         *cache_metrics.CacheTimer
 	size          int64
 }
 
+func (w *waitForUploadWriteCloser) Write(p []byte) (int, error) {
+	_, spn := tracing.StartSpan(w.ctx)
+	defer spn.End()
+	return w.WriteCloser.Write(p)
+
```
tylerwilliams, line 468: nit: rm newline

tylerwilliams (approved):
LGTM (w 2 nits)

enterprise/server/backends/s3_cache/s3_cache.go
tempoz: Done.

enterprise/server/backends/s3_cache/s3_cache.go
tempoz: Done.

PR #1342, opened 2021-12-16
Add tracing around azure, gcs, and s3 blobstores

tempoz:
Adds tracing spans around the calls that perform network operations in the blobstores.

---

**Version bump**: Minor

### DIFF
--- a/server/backends/blobstore/blobstore.go
+++ b/server/backends/blobstore/blobstore.go
@@ -28,6 +28,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/disk"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
 	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/api/option"
 
@@ -44,11 +45,11 @@ const (
 )
 
 // Returns whatever blobstore is specified in the config.
-func GetConfiguredBlobstore(c *config.Configurator) (interfaces.Blobstore, error) {
+func GetConfiguredBlobstore(ctx context.Context, c *config.Configurator) (interfaces.Blobstore, error) {
 	log.Debug("Configuring blobstore")
 	if c.GetStorageDiskRootDir() != "" {
 		log.Debug("Disk blobstore configured")
-		return NewDiskBlobStore(c.GetStorageDiskRootDir())
+		return NewDiskBlobStore(ctx, c.GetStorageDiskRootDir())
 	}
 	if gcsConfig := c.GetStorageGCSConfig(); gcsConfig != nil && gcsConfig.Bucket != "" {
 		log.Debug("Configuring GCS blobstore")
@@ -132,7 +133,7 @@ type DiskBlobStore struct {
 	rootDir string
 }
 
-func NewDiskBlobStore(rootDir string) (*DiskBlobStore, error) {
+func NewDiskBlobStore(ctx context.Context, rootDir string) (*DiskBlobStore, error) {
 	if err := disk.EnsureDirectoryExists(rootDir); err != nil {
 		return nil, err
 	}
@@ -205,7 +206,9 @@ func (d *DiskBlobStore) WriteBlob(ctx context.Context, blobName string, data []b
 		return 0, err
 	}
 	start := time.Now()
+	ctx, spn := tracing.StartSpan(ctx)
 	n, err := disk.WriteFile(ctx, fullPath, compressedData)
+	spn.End()
 	recordWriteMetrics(diskLabel, start, n, err)
 	return n, err
 }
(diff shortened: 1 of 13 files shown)


server/backends/blobstore/blobstore.go
```diff
@@ -44,11 +45,11 @@ const (
 )
 
 // Returns whatever blobstore is specified in the config.
-func GetConfiguredBlobstore(c *config.Configurator) (interfaces.Blobstore, error) {
+func GetConfiguredBlobstore(ctx context.Context, c *config.Configurator) (interfaces.Blobstore, error) {
 	log.Debug("Configuring blobstore")
 	if c.GetStorageDiskRootDir() != "" {
 		log.Debug("Disk blobstore configured")
-		return NewDiskBlobStore(c.GetStorageDiskRootDir())
+		return NewDiskBlobStore(ctx, c.GetStorageDiskRootDir())
```
tylerwilliams, line 52: Can we skip this one? I think the server won't init if it's slow / doesn't work, and it only happens once, so don't think it's worth trying to trace on the root context.

server/backends/blobstore/blobstore.go
```diff
@@ -240,6 +247,8 @@ func (d *DiskBlobStore) BlobExists(ctx context.Context, blobName string) (bool,
 	if err != nil {
 		return false, err
 	}
+	ctx, spn := tracing.StartSpan(ctx)
+	defer spn.End()
 	return disk.FileExists(fullPath)
 }
```
tylerwilliams, line 253: you tell me, but maybe we'd want to add a ctx param to disk.FileExists and just do that there?

server/backends/blobstore/blobstore.go
```diff
@@ -287,7 +305,9 @@ func (g *GCSBlobStore) ReadBlob(ctx context.Context, blobName string) ([]byte, e
 		return nil, err
 	}
 	start := time.Now()
+	_, spn := tracing.StartSpan(ctx)
```
tylerwilliams, line 308: nit: i know ctx isn't used here, but does the compiler complain if you do

"ctx, spn := tracing.StartSpan(ctx)"?

It feels like that might be less error-prone going forward, if someone does introduce code that takes a context, they won't need to also know that _ is a ctx we are ignoring and change that too.

server/libmain/libmain.go
```diff
@@ -140,7 +140,7 @@ func configureFilesystemsOrDie(realEnv *real_environment.RealEnv) {
 // the environments used by the open-core version and the enterprise version are
 // not substantially different enough yet to warrant the extra complexity of
 // always updating both main files.
-func GetConfiguredEnvironmentOrDie(configurator *config.Configurator, healthChecker *healthcheck.HealthChecker) *real_environment.RealEnv {
+func GetConfiguredEnvironmentOrDie(ctx context.Context, configurator *config.Configurator, healthChecker *healthcheck.HealthChecker) *real_environment.RealEnv {
```
tylerwilliams, line 143: same as above but i think it'd be fine to just omit this one. Doesn't add much value for the noise

tylerwilliams (changes requested):
Few nits but this mostly LG too.

server/backends/blobstore/blobstore.go
tempoz: np, done.

server/backends/blobstore/blobstore.go
tempoz: Makes sense; I like that.

server/backends/blobstore/blobstore.go
tempoz: Nah, the compiler is cool with it since we aren't declaring ctx, only assigning to it. I think that makes sense - swapped it over.

server/libmain/libmain.go
tempoz: Got it; fixed.

server/backends/blobstore/blobstore.go
```diff
@@ -300,14 +318,18 @@ func (g *GCSBlobStore) WriteBlob(ctx context.Context, blobName string, data []by
 		return 0, err
 	}
 	start := time.Now()
+	_, spn := tracing.StartSpan(ctx)
```
tylerwilliams, line 321: ctx, spn?

server/backends/blobstore/blobstore.go
```diff
@@ -397,19 +421,33 @@ func NewAwsS3BlobStore(awsConfig *config.AwsS3Config) (*AwsS3BlobStore, error) {
 	return awsBlobStore, nil
 }
 
+func (a *AwsS3BlobStore) bucketExists(ctx context.Context, bucketName string) (bool, error) {
+	_, spn := tracing.StartSpan(ctx)
```
tylerwilliams, line 425: ctx, spn?

server/backends/blobstore/blobstore.go
```diff
@@ -555,12 +601,26 @@ func (z *AzureBlobStore) isAzureError(err error, code azblob.ServiceCodeType) bo
 	return false
 }
 
+func (z *AzureBlobStore) containerExists(ctx context.Context) (bool, error) {
+	_, spn := tracing.StartSpan(ctx)
```
tylerwilliams, line 605: ctx, spn?

tylerwilliams (changes requested):

server/backends/blobstore/blobstore.go
tempoz: Done.

server/backends/blobstore/blobstore.go
tempoz: Done.

tylerwilliams (approved):
LGTM (thanks!)

PR #1343, opened 2021-12-17
Fix prettier checkstyle and run prettier

bduffany:
It was outputting the errors to stderr instead of `prettier-errors.txt`. Also `npx` prints to stderr if the binary is not already installed, so we now install prettier separately and run the binary directly.

---

**Version bump**: Patch


siggisim (approved):
I was wondering if the prettier checkstyle was working!

PR #1344, opened 2021-12-17
Add utils for downloading CAS/AC artifacts via the UI

bduffany:
* Make the `invocation_id` param _optional_ in the `/file/download` endpoint, to avoid needing to know the invocation ID just to download a blob from CAS / AC.
  * Before, we would use the `invocation_id` to fetch an API key for the invocation owner. That API key then determines the group ID prefix for the cache.
  * But the group ID prefix comes from AuthenticatedUser, which doesn't care whether it comes from an API key or a JWT. Since Web requests are authenticated with a JWT, it should be possible to use only the JWT and avoid the need for the API key.
  * The only issue is that that we don't properly set `Claims.GroupID` for plain HTTP requests, because that field comes from `RequestContext.selected_group_id`, which is only populated for _protolet_ HTTP requests.
  * To work around this, we can pass the `RequestContext` via URL params, and set the `RequestContext` in the `ctx` via middleware. This way, both plain-HTTP and protolet requests will have the `RequestContext` available in the HTTP authentication path.
  * This also buys us proper impersonation support, since impersonation is based on `impersonating_group_id` in the request context. Specifically, it will allow us to know whether the authenticated user fetching the CAS blobs is an impersonator, to avoid incrementing usage counts when calling these debug endpoints.
* Add util functions for downloading objects from AC (as JSON) and CAS (as raw binary)
  * Expose these in global scope as `_downloadActionResultJSON` and `_downloadBlob` so they can be invoked via devtools. This is a temporary solution until we implement a full UI.

---

Usage:

* Open devtools console (F12)
* To download an action result: `_downloadActionResultJSON(hash: string, size: number, instanceName: string)`
* To download a blob from CAS: `_downloadBlob(hash: string, size: number, instanceName: string)`
* Instance name is optional and defaults to `""`
* Size is required, but you can pass `1` if you don't know it, and it should technically work. (I'm not sure if the server-side behavior will change in the future, but for now this will work).

---

**Version bump**: Patch

### DIFF
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -40,6 +40,8 @@ class RpcService {
       filename,
       bytestream_url: bytestreamURL,
       invocation_id: invocationId,
+      group_id: this.requestContext.groupId || "",
+      impersonating_group_id: this.requestContext.impersonatingGroupId || "",
     })}`;
   }
 
@@ -52,13 +54,8 @@ class RpcService {
     invocationId: string,
     responseType?: "arraybuffer" | "json" | "text" | undefined
   ) {
-    return this.fetchFile(
-      `/file/download?${new URLSearchParams({
-        bytestream_url: bytestreamURL,
-        invocation_id: invocationId,
-      })}`,
-      responseType || ""
-    );
+    const url = this.getBytestreamFileUrl(/*filename=*/ "", bytestreamURL, invocationId);
+    return this.fetchFile(url, responseType || "");
   }
 
   fetchFile(fileURL: string, responseType: "arraybuffer" | "json" | "text" | "") {
--- a/app/terminal/BUILD
+++ b/app/terminal/BUILD
@@ -9,6 +9,7 @@ ts_library(
         "//app/components/spinner",
         "//app/errors",
         "//app/service",
+        "//app/util:download",
         "//proto:eventlog_ts_proto",
         "@npm//@types/react",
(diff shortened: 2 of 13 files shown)


brentleyjones:
This is awesome @bduffany!

One question/request: if I have a gRPC log, I don't readily have the instance name used, but instead have the invitation id. Is there a way to have flavors of these calls that can take the invocation id instead and look that up for me? Example:
```
metadata {
  tool_details {
    tool_name: "bazel"
    tool_version: "5.0.0rc3"
  }
  action_id: "10cca269647f4957e7d94a6a0cbf4c579c26026657eb5633a3d5331d09c24dd6"
  tool_invocation_id: "36be809e-0905-4620-85fb-4084b129d01d"
  correlated_invocations_id: "f95acfde-d451-4fcc-98f7-844ae983d55d"
  action_mnemonic: "SwiftStdlibCopy"
  configuration_id: "b99cb83d8d42d2ed0812bcbc22526a903b7f1c084d37ec50f61e7553c4ced492"
}
method_name: "build.bazel.remote.execution.v2.ActionCache/GetActionResult"
details {
  get_action_result {
    request {
      action_digest {
        hash: "10cca269647f4957e7d94a6a0cbf4c579c26026657eb5633a3d5331d09c24dd6"
        size_bytes: 144
      }
    }
  }
}
```

brentleyjones:
(That can totally be a followup as well instead of holding this up!)

PR #1346, opened 2021-12-17
Trace common disk operations

tempoz:
Adds tracing to common file operations using the disk utility. Avoids changing function signatures.

---

**Version bump**: Minor

### DIFF
--- a/server/util/disk/disk.go
+++ b/server/util/disk/disk.go
@@ -13,6 +13,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
 )
 
 func EnsureDirectoryExists(dir string) error {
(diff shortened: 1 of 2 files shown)


server/util/disk/disk.go
```diff
@@ -76,12 +83,19 @@ func FileExists(fullPath string) (bool, error) {
...
-type SectionReaderCloser struct {
+type readCloser struct {
 	*io.SectionReader
 	io.Closer
+	ctx context.Context
 }
 
-func FileReader(ctx context.Context, fullPath string, offset, length int64) (*SectionReaderCloser, error) {
+func (r *readCloser) Read(p []byte) (int, error) {
+	_, spn := tracing.StartSpan(r.ctx)
```
tylerwilliams, line 93: ctx, spn := tracing.StartSpan(r.ctx)

server/util/disk/disk.go
```diff
@@ -91,16 +105,23 @@ func FileReader(ctx context.Context, fullPath string, offset, length int64) (*Se
...
 }
 
 type writeMover struct {
 	*os.File
+	ctx       context.Context
 	finalPath string
 }
 
+func (w *writeMover) Write(p []byte) (int, error) {
+	_, spn := tracing.StartSpan(w.ctx)
```
tylerwilliams, line 120: ctx, spn := tracing.StartSpan(w.ctx)

tylerwilliams (changes requested):
See inline comments

vadimberezniker (approved):

server/util/disk/disk.go
tempoz: So when I went to do this, I realized there is no context variable referenced in this function, which means I can't name the variable. Same for below.

tylerwilliams: lol i'm a goof, thanks for looking :)

tylerwilliams (approved):
LGTM

PR #1348, opened 2021-12-18
Refactor SDKROOT calculation

brentleyjones:
We now store a map of sdk -> sdkPath (e.g. `iPhoneSimulator15.2` -> `Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator15.2.sdk`, `MacOSX.sdk` -> `SDKs/MacOSX.sdk`). We now also return an error if we can't find the desired SDK in the desired Xcode version, which is more correct than falling back to a version-less SDK.

This change is more performant, with a map lookup instead of iterating over an array. But more importantly, it is needed for a future change to support using the Xcode Command Line Tools as the default "Xcode" version.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/platform/platform_test.go
+++ b/enterprise/server/remote_execution/platform/platform_test.go
@@ -1,6 +1,7 @@
 package platform_test
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
@@ -189,31 +190,31 @@ func TestParse_ApplyOverrides(t *testing.T) {
 		platformProps       []*repb.Platform_Property
 		startingEnvVars     []*repb.Command_EnvironmentVariable
 		expectedEnvVars     []*repb.Command_EnvironmentVariable
+		errorExpected       bool
 		defaultXcodeVersion string
 	}{
 		// Default darwin platform
 		{[]*repb.Platform_Property{
 			{Name: "osfamily", Value: "darwin"},
 		}, []*repb.Command_EnvironmentVariable{
-			{Name: "APPLE_SDK_VERSION_OVERRIDE", Value: "11.1"},
 			{Name: "XCODE_VERSION_OVERRIDE", Value: "12.4.123"},
+		}, []*repb.Command_EnvironmentVariable{
+			{Name: "SDKROOT", Value: "/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"},
+			{Name: "DEVELOPER_DIR", Value: "/Applications/Xcode_12.4.app/Contents/Developer"},
 		},
-			[]*repb.Command_EnvironmentVariable{
-				{Name: "SDKROOT", Value: "/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"},
-				{Name: "DEVELOPER_DIR", Value: "/Applications/Xcode_12.4.app/Contents/Developer"},
-			},
+			false,
 			"12.2",
 		},
 		// Case insensitive darwin platform
 		{[]*repb.Platform_Property{
 			{Name: "OSFAMILY", Value: "dArWiN"},
 		}, []*repb.Command_EnvironmentVariable{
-			{Name: "APPLE_SDK_VERSION_OVERRIDE", Value: "11.1"},
 			{Name: "XCODE_VERSION_OVERRIDE", Value: "12.4.123"},
 		}, []*repb.Command_EnvironmentVariable{
 			{Name: "SDKROOT", Value: "/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"},
 			{Name: "DEVELOPER_DIR", Value: "/Applications/Xcode_12.4.app/Contents/Developer"},
 		},
+			false,
 			"12.2",
 		},
 		// Darwin with no overrides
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/platform/platform_test.go
```diff
@@ -189,31 +190,31 @@ func TestParse_ApplyOverrides(t *testing.T) {
 		platformProps       []*repb.Platform_Property
 		startingEnvVars     []*repb.Command_EnvironmentVariable
 		expectedEnvVars     []*repb.Command_EnvironmentVariable
+		errorExpected       bool
```
tylerwilliams, line 193: don't think it's necessary here, but if you'd like to check for specific errors, you could have this be an *error and compare it for equality if set.

server/interfaces/interfaces.go
```diff
@@ -572,11 +572,9 @@ type HealthChecker interface {
 
 // Locates all Xcode versions installed on the host system.
 type XcodeLocator interface {
-	// Returns the developer directory and SDKs for the given Xcode version.
-	DeveloperDirForVersion(version string) (string, error)
-
-	// Returns true if the given SDK path is present in the given Xcode version.
-	IsSDKPathPresentForVersion(sdkPath, version string) bool
+	// Finds the Xcode that most closely matches the given Xcode version.
```
tylerwilliams, line 575: nit: could we specify in the comment what "closely matches" means -- what counts as a match versus an error?

server/xcode/xcode_darwin.go
```diff
@@ -41,27 +43,20 @@ func NewXcodeLocator() (*xcodeLocator, error) {
...
-	}
-	return xv.developerDirPath, nil
-}
 
-// Return true if the given SDK path is present in the given Xcode version.
-func (x *xcodeLocator) IsSDKPathPresentForVersion(sdkPath, version string) bool {
-	xv := x.xcodeVersionForVersionString(version)
-	if xv == nil {
-		return false
+		return "", "", status.FailedPreconditionErrorf("Xcode version %s not installed on remote executor. Available Xcode versions are %+v", xcodeVersion, x.versions)
```
tylerwilliams, line 52: ubernit, but can we make a nicer string instead of printing the map? (Since this is for users...)

func versionsString(versions map[string]*xcodeVersion) string {
  availableVersionsList := make([]string, 0, len(x.versions))
  for version, _ := range x.versions {
    availableVersions = append(availableVersions, version)
  }
  sort.Strings(availableVersionsList)
  return strings.Join(availableVersionsList, ", ")
}
...
status.FailedPreconditionErrorf("Xcode version %s not installed on remote executor. Available Xcode versions are %s", xcodeVersion, versionsString(x.versions))

tylerwilliams (changes requested):
Couple small nits but otherwise LG, thanks for improving this :)

enterprise/server/remote_execution/platform/platform_test.go
brentleyjones: I had that to start. It felt a little heavy to have to creat an error in the setup, since we control the stubbed function as well.

server/xcode/xcode_darwin.go
```diff
@@ -168,3 +168,25 @@ func stringToCFString(gostr string) C.CFStringRef {
...
 	return C.GoString(C.CFStringGetCStringPtr(cfStr, C.kCFStringEncodingUTF8))
 }
+
+func sdksString(sdks map[string]string) string {
+	availableSDKs := make([]string, 0, len(sdks))
+	i := 0
+	for sdk, _ := range sdks {
+		availableSDKs[i] = sdk
+		i++
+	}
```
tylerwilliams, line 178: [make](https://pkg.go.dev/builtin#make)() (for slices) accepts a length and capacity as the second and third arg.

Typically you want to set one or both if you know the length of slice you'll need ahead of time so that you avoid the extra allocations that would happen to dynamically resize the slice. (Length obviously 'includes' capacity)

So the 3 most common patterns are:

1) no idea how big the slice is, just make a 0 length slice and append to it:
slice := make([]string, 0)
for thing in blah {
  slice = append(slice, thing)
}

2) you know how big the slice is, and you have an index that increases by one for each item
slice := make([]int, count)
for i := 0; i < count; i++ {
  slice[i] = i * i
}

3) you know how big the slice (or roughly how big it is) is but you just want to put items in it
slice := make([]string, 0, count)
for k, v in myMap {
  if v == nil {
    continue
  }
  slice = append(slice, k)
} 


Generally it is clearest to use 3) so you don't need to keep track of i. And since you set the slice capacity you're avoiding the extra allocations already.

So for these loops, here and below, can we remove "i" and use append instead?

tylerwilliams (changes requested):
one nit, otherwise lgtm

server/xcode/xcode_darwin.go
brentleyjones: Based on https://stackoverflow.com/a/27848197 it seemed best to use `i`. I forgot to adjust the `make()`. I can adjust it to use append though.

brentleyjones: (Because of "Although setting the capacity eliminates reallocations, append still has to do extra work to check if you've reached capacity on each append.")

tylerwilliams: Yeah, this is correct (it is slightly faster to use an index like this), but it comes at the cost of having to declare the extra counter variable.

I think for stuff like this (displaying a convenience string to the user) I would prefer the less code / more readable / simpler to refactor version.

tylerwilliams (approved):
LGTM

PR #1350, opened 2021-12-19
Use xcode-select if default_xcode_version isn't set

brentleyjones:
`xcode-select` can even be set to the Command Line Tools, potentially allowing the executor to work without any Xcodes installed at all, as long as no non-macOS platforms are requested.

---

**Version bump**: Patch

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/999.

### DIFF
--- a/server/xcode/xcode_darwin.go
+++ b/server/xcode/xcode_darwin.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -27,6 +28,7 @@ const xcodeBundleID = "com.apple.dt.Xcode"
 const versionPlistPath = "Contents/version.plist"
 const developerDirectoryPath = "Contents/Developer"
 const filePrefix = "file://"
+const defaultXcodeVersion = "default-xcode-version"
 
 type xcodeLocator struct {
 	versions map[string]*xcodeVersion
@@ -53,10 +55,19 @@ func NewXcodeLocator() (*xcodeLocator, error) {
 // Finds the Xcode that matches the given Xcode version.
 // Returns the developer directory for that Xcode and the SDK root for the given SDK.
 func (x *xcodeLocator) PathsForVersionAndSDK(xcodeVersion string, sdk string) (string, string, error) {
+	if xcodeVersion == "" {
+		xcodeVersion = defaultXcodeVersion
+	}
+
 	xv := x.xcodeVersionForVersionString(xcodeVersion)
 	if xv == nil {
-
-		return "", "", status.FailedPreconditionErrorf("Xcode version %s not installed on remote executor. Available Xcode versions are %s", xcodeVersion, versionsString(x.versions))
+		var err error
+		if xcodeVersion == defaultXcodeVersion {
+			err = status.FailedPreconditionErrorf("Default Xcode version not set on remote executor and xcode-select set to an invalid path. Available Xcode versions are %s", versionsString(x.versions))
+		} else {
+			err = status.FailedPreconditionErrorf("Xcode version %s not installed on remote executor. Available Xcode versions are %s", xcodeVersion, versionsString(x.versions))
+		}
+		return "", "", err
 	}
 	sdkPath, ok := xv.sdks[sdk]
 	if !ok {
@@ -102,6 +113,7 @@ func (x *xcodeLocator) locate() error {
 }
 
 func versionMap(urlRefs []C.CFURLRef) map[string]*xcodeVersion {
+	defaultDeveloperDir := xcodeSelectDeveloperDir()
 	versionMap := make(map[string]*xcodeVersion)
 	for _, urlRef := range urlRefs {
 		path := "/" + strings.TrimLeft(stringFromCFString(C.CFURLGetString(C.CFURLRef(urlRef))), filePrefix)
@@ -129,13 +141,34 @@ func versionMap(urlRefs []C.CFURLRef) map[string]*xcodeVersion {
 				continue
 			}
 			log.Debugf("Mapped Xcode Version %s=>%s with SDKs %+v", version, developerDirPath, sdks)
-			versionMap[version] = &xcodeVersion{
+			xv := &xcodeVersion{
 				version:          mostPreciseVersion,
 				developerDirPath: developerDirPath,
 				sdks:             sdks,
 			}
+			versionMap[version] = xv
+			if defaultDeveloperDir == developerDirPath {
+				versionMap[defaultXcodeVersion] = xv
+			}
 		}
 	}
+
+	if xv, ok := versionMap[defaultXcodeVersion]; ok {
+		log.Infof("xcode-select set to Xcode version %s", xv.version)
+		return versionMap
+	}
+
+	// If we are here then defaultDeveloperDir is not a valid Xcode path
+	if defaultDeveloperDir == "/Library/Developer/CommandLineTools" {
+		version := commandLineToolsVersionString()
+		log.Infof("xcode-select set to Command Line Tools (%s)", version)
+		if xv := commandLineToolsVersion(defaultDeveloperDir); xv != nil {
+			versionMap[defaultXcodeVersion] = xv
+		}
+	} else if defaultDeveloperDir != "" {
+		log.Warningf("xcode-select set to an invalid Xcode path: %s", defaultDeveloperDir)
+	}
+
 	return versionMap
 }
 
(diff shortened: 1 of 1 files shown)


siggisim (approved):
This looks reasonable to me, but let's let @tylerwilliams give it a once-over for any golang best practices before merging.

Thanks for doing this!!

server/xcode/xcode_darwin.go
```diff
@@ -90,6 +96,14 @@ func (x *xcodeLocator) locate() error {
...
 	C.CFArrayGetValues(urlsArrayRef, C.CFRange{0, n}, (*unsafe.Pointer)(unsafe.Pointer(&urlRefs[0])))
 
+	var defaultDeveloperDir string
+	out, err := exec.Command("xcode-select", "-p").Output()
+	if err != nil {
+		log.Warningf("Error calling xcode-select: %s", err.Error())
+	} else {
+		defaultDeveloperDir = strings.TrimSpace(string(out))
+	}
+
```
tylerwilliams, line 106: nit: kind of a matter of taste, but given how long "locate()" is I think we should put this in a a function:

// xcodeSelectDir returns the active developer dir according to xcode-select or
// the empty string if xcode select did not run.
func xcodeSelectDir() string {
  out, err := exec.Command("xcode-select", "-p").Output()
  if err != nil {
		log.Warningf("Error calling xcode-select: %s", err.Error())
        return ""
  }
  return strings.TrimSpace(string(out))
}


...

defaultDeveloperDir = x.xcodeSelectDir()

server/xcode/xcode_darwin.go
```diff
@@ -126,13 +140,51 @@ func (x *xcodeLocator) locate() error {
...
 			log.Debugf("Mapped Xcode Version %s=>%s with SDKs %+v", version, developerDirPath, sdks)
-			versionMap[version] = &xcodeVersion{
+			xv := &xcodeVersion{
 				version:          mostPreciseVersion,
 				developerDirPath: developerDirPath,
 				sdks:             sdks,
 			}
+			versionMap[version] = xv
+			if defaultDeveloperDir == developerDirPath {
+				versionMap[""] = xv
```
tylerwilliams, line 150: It's kind of unusual to store something under the empty string like this. This would be more idiomatic / clearer if we stored it under a const known key. Something like:

const defaultVersion = "default-xcode-version"

if defaultDeveloperDir == developerDirPath {
				versionMap[defaultVersion] = xv
}

etc

server/xcode/xcode_darwin.go
```diff
@@ -126,13 +140,51 @@ func (x *xcodeLocator) locate() error {
...
+			}
 		}
 	}
+
+	if xv, ok := versionMap[""]; ok {
+		log.Infof("xcode-select set to Xcode version %s", xv.version)
+	} else {
```
tylerwilliams, line 157: nit: can we return early here and unnest the else? (maybe unnecessary after the other cleanups)

server/xcode/xcode_darwin.go
```diff
@@ -90,6 +96,14 @@ func (x *xcodeLocator) locate() error {
...
+	out, err := exec.Command("xcode-select", "-p").Output()
+	if err != nil {
+		log.Warningf("Error calling xcode-select: %s", err.Error())
+	} else {
+		defaultDeveloperDir = strings.TrimSpace(string(out))
+	}
+
 	versionMap := make(map[string]*xcodeVersion)
 	for _, urlRef := range urlRefs {
 		path := "/" + strings.TrimLeft(stringFromCFString(C.CFURLGetString(C.CFURLRef(urlRef))), filePrefix)
```
tylerwilliams, line 109: This loop would be a lot clearer if we:
 - defined the "xcodePlist" struct up at the top of this file
 - pulled the plist parsing out into a function that takes a path (string) and returns a xcodePlist struct and error

server/xcode/xcode_darwin.go
```diff
@@ -126,13 +140,51 @@ func (x *xcodeLocator) locate() error {
...
+	} else {
+		// If we are here then defaultDeveloperDir is not a valid Xcode path
+		if defaultDeveloperDir == "/Library/Developer/CommandLineTools" {
+			var version string
+			out, err := exec.Command("sh", "-c", "pkgutil --pkg-info=com.apple.pkg.CLTools_Executables | awk '/version:/ {print $2}'").Output()
+			if err == nil {
+				version = strings.TrimSpace(string(out))
+			} else {
+				version = "unknown"
+			}
```
tylerwilliams, line 166: similar to the xcode-select function above, can this be a bare function?

server/xcode/xcode_darwin.go
```diff
@@ -126,13 +140,51 @@ func (x *xcodeLocator) locate() error {
...
+				sdks := make(map[string]string)
+				for _, sdkPath := range sdkPaths {
+					sdks[strings.TrimSuffix(filepath.Base(sdkPath), ".sdk")] = sdkPath
+				}
+				versionMap[""] = &xcodeVersion{
+					version:          "",
+					developerDirPath: defaultDeveloperDir,
+					sdks:             sdks,
+				}
+			}
```
tylerwilliams, line 182: it's [idomatic](https://zchee.github.io/golang-wiki/CodeReviewComments/#indent-error-flow) in go to return early and unnest the else case. If we moved this block to a function, we could do that. Something like

func commandlineToolVersion() *xcodeVersion {
 ...
}

tylerwilliams (changes requested):
few more comments on this one -- no good deed goes unpunished :P lmk if you want to chat about any of these if you have questions / concerns.

server/xcode/xcode_darwin.go
brentleyjones: Yeah, this was relying on the side effect of if there wasn't an override, it would be the empty string. I'm all for making it explicit.

server/xcode/xcode_darwin.go
brentleyjones: Addressed in #1354.

server/xcode/xcode_darwin.go
brentleyjones: Addressed in #1354.

brentleyjones:
* master
  * **PR #1354** 
    * **PR #1350**  (this PR)

server/xcode/xcode_darwin.go
```diff
@@ -151,6 +184,48 @@ func xcodePlistForPath(path string) (*xcodePlist, error) {
...
+	}
+	return strings.TrimSpace(string(out))
+}
+
+// Returns the active developer directory according to xcode-select, or the
+// empty string if xcode-select failed to run.
+func xcodeSelectDeveloperDir() string {
+	out, err := exec.Command("xcode-select", "-p").Output()
+	if err != nil {
+		log.Warningf("Error calling xcode-select: %s", err.Error())
```
brentleyjones, line 223: Should I return an error here instead and let the caller log?

server/xcode/xcode_darwin.go
```diff
@@ -151,6 +184,48 @@ func xcodePlistForPath(path string) (*xcodePlist, error) {
...
+
+// Returns the version of the Xcode Command Line Tools, or "unknown" if it
+// can't be determined.
+func commandLineToolsVersionString() string {
+	out, err := exec.Command(
+		"sh", "-c",
+		"pkgutil --pkg-info=com.apple.pkg.CLTools_Executables | awk '/version:/ {print $2}'",
+	).Output()
+	if err != nil {
+		log.Warningf("Error determining Xcode Command Line Tools version: %s", err.Error())
```
brentleyjones, line 212: Same here, should I send the error up?

server/xcode/xcode_darwin.go
```diff
@@ -129,13 +141,34 @@ func versionMap(urlRefs []C.CFURLRef) map[string]*xcodeVersion {
...
+
+	// If we are here then defaultDeveloperDir is not a valid Xcode path
+	if defaultDeveloperDir == "/Library/Developer/CommandLineTools" {
+		version := commandLineToolsVersionString()
+		log.Infof("xcode-select set to Command Line Tools (%s)", version)
+		if xv := commandLineToolsVersion(defaultDeveloperDir); xv != nil {
+			versionMap[defaultXcodeVersion] = xv
 		}
+	} else if defaultDeveloperDir != "" {
+		log.Warningf("xcode-select set to an invalid Xcode path: %s", defaultDeveloperDir)
```
tylerwilliams, line 169: it's a little weird to have this else case here just to detect the special case when defaultDeveloperDir == "" and then log the error that xcode-select was not found. Could we log this warning inside of xcodeSelectDeveloperDir and then omit this else if block?

server/xcode/xcode_darwin.go
tylerwilliams: Typically yes -- in this case it seems like you just log this version string for info only, so doesn't matter a ton, is that right?

server/xcode/xcode_darwin.go
tylerwilliams: This one matters more, but since you are comparing the returned string against some known good values and looking for a match, if you return an empty string here on failure it seems fine -- you can log an error here if you get one and your calling code will just not find a match the returned empty string against anything.

server/xcode/xcode_darwin.go
brentleyjones: When `defaultDeveloperDir != ""`, which will only happen when `xcode-select` runs correctly, but it's set to a no longer installed Xcode, or some other invalid Xcode path. We only know it's invalid at this point, instead of in `xcodeSelectDeveloperDir`, after we've tried to match the path to all installed Xcode versions detected.

tylerwilliams (approved):
LGTM

(discussed offline, happy with this as is, thanks!)

PR #1351, opened 2021-12-20
Make terminal component more self-contained so it can be used to show action outputs

bduffany:
Updates the terminal component to be more self-contained so that it can render in places other than `.card`.

* Remove external negative margins to avoid unexpected layout shifts
* Give it ownership of the title and subtitle, so that it can manage the layout more directly, without relying on negative margins
* Add a placeholder div which is empty but has the same dimensions as the search bar, so that the title and actions are laid out more naturally and wrap properly on different screen sizes.

Before (layout is shifted upwards, top bar does not have the correct BG color, wrap/download icons are not visible):

image

After (different invocation, so output is not expected to be the same):

image

The terminal could use a bit of padding in this screenshot, but I think that should be managed by a separate tiny helper component, e.g. `TerminalFrame` component which applies padding with the correct color depending on the theme, + slightly rounded borders.

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -117,12 +117,12 @@
   color: #222;
 }
 
-.action-section div {
+.action-section > div {
   flex-grow: 1;
   font-size: 13px;
 }
 
-.action-section div.action-property-title {
+.action-section .action-property-title {
   width: 20%;
   flex-grow: 0;
   flex-shrink: 0;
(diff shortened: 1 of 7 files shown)


siggisim (approved):
What's the downside of adding the padding to this component? Because we want it to be different in different contexts?

bduffany:
> What's the downside of adding the padding to this component?

In the other places where we use the terminal, we have no padding on the terminal itself, and the containing card applies the padding, so I figured it makes sense to stick to that pattern, otherwise we have to add CSS to remove the padding for all of those cases.

PR #1353, opened 2021-12-21
Also check exit code in tableExecToProto func

luluz66:
OK status code doesn't imply that exit code is zero. On linux machine,
the exit code can be 1 and status code is OK.


tylerwilliams (approved):
LGTM

PR #1354, opened 2021-12-30
Refactor XcodeLocator versionMap computation

brentleyjones:
Some cleanup from code review of #1350. It will make that change apply cleaner.

---

**Version bump**: Patch

### DIFF
--- a/server/xcode/xcode_darwin.go
+++ b/server/xcode/xcode_darwin.go
@@ -143,10 +143,10 @@ func xcodePlistForPath(path string) (*xcodePlist, error) {
 	var xcodePlist xcodePlist
 	versionFileReader, err := os.Open(path)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to open %s: %s", path, err.Error())
+		return nil, status.InternalErrorf("Failed to open %s: %s", path, err.Error())
 	}
 	if err := plist.NewXMLDecoder(versionFileReader).Decode(&xcodePlist); err != nil {
-		return nil, fmt.Errorf("Failed to decode %s: %s", path, err.Error())
+		return nil, status.InternalErrorf("Failed to decode %s: %s", path, err.Error())
 	}
 	return &xcodePlist, nil
 }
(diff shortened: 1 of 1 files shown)


server/xcode/xcode_darwin.go
```diff
@@ -134,8 +135,21 @@ func (x *xcodeLocator) locate() error {
 			}
 		}
 	}
-	x.versions = versionMap
-	return nil
+	return versionMap
+}
+
+func xcodePlistForPath(path string) *xcodePlist {
```
tylerwilliams, line 141: Can we have xcodePlistForPath return a (*xcodePlist, error), and then check (and potentially log) the error when calling xcodePlistForPath?

This way it's clear when the returned *xcodePlist is valid (no need to check nil), and you can just have a single log.Warningf()

tylerwilliams (changes requested):
See inline comments

server/xcode/xcode_darwin.go
```diff
@@ -134,8 +136,19 @@ func (x *xcodeLocator) locate() error {
...
+}
+
+func xcodePlistForPath(path string) (*xcodePlist, error) {
+	var xcodePlist xcodePlist
+	versionFileReader, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to open %s: %s", path, err.Error())
+	}
+	if err := plist.NewXMLDecoder(versionFileReader).Decode(&xcodePlist); err != nil {
+		return nil, fmt.Errorf("Failed to decode %s: %s", path, err.Error())
```
tylerwilliams, line 149: In our codebase we prefer to use the errors from the server/util/status package rather than minting our own. They have some benefits:
 - map cleanly to grpc errors
 - can embed a stack trace if a config option is enabled
 - are easily "parseable" with Is*Error, or even into protos

So we could make these:
 >> return nil, status.InternalErrorf("Failed to open %s: %s", path, err.Error())

and:
 >> return nil, status.InternalErrorf("Failed to decode %s: %s", path, err.Error())

brentleyjones:
* master
  * **PR #1354**  (this PR)
    * **PR #1350** 

tylerwilliams (approved):
LGTM

PR #1355, opened 2022-01-03
Add styles for invocation action card terminal

bduffany:
Just adding padding and border-radius to the `.terminal-container`. Moved the background color to `.terminal-container` instead of `.terminal` so that the padding gets the color as well.

image

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -51,7 +51,7 @@
 }
 
 .invocation-action-card .terminal-container {
-  padding: 8px;
+  padding: 8px 32px;
   border-radius: 8px;
 }
 
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -50,6 +50,11 @@
   height: 18px;
 }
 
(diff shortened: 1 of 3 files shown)


app/invocation/invocation.css
```diff
@@ -50,6 +50,11 @@
   height: 18px;
 }
 
+.invocation-action-card .terminal-container {
+  padding: 8px;
```
siggisim, line 54: 8px seems a little tight on the sides.

maybe something more like 
```
padding: 8px 32px;
```

siggisim (approved):

app/invocation/invocation.css
bduffany: sg, done -- the additional padding uncovered an issue that the top bar of the terminal was "ignoring" the padding since it is absolutely positioned -- fixed by adding a wrapper div inside `.terminal-container` and positioning relative to that div instead. Here's what it looks like now:

image

siggisim (approved):

PR #1356, opened 2022-01-03
Add support for displaying artifacts populated via namedSetOfFiles

siggisim:
This prepares us for the impending flip of the `--legacy_important_outputs` flag in future bazel versions: https://github.com/bazelbuild/bazel/pull/14353

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -29,7 +29,6 @@ export default class InvocationModel {
   brokenTest: build_event_stream.BuildEvent[] = [];
   flakyTest: build_event_stream.BuildEvent[] = [];
   timeoutTest: build_event_stream.BuildEvent[] = [];
-  files: Map<string, build_event_stream.IFile[]> = new Map();
   structuredCommandLine: command_line.CommandLine[] = [];
   finished: build_event_stream.BuildFinished;
   aborted: build_event_stream.BuildEvent;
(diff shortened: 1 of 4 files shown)


app/invocation/invocation_model.tsx
```diff
@@ -514,6 +515,16 @@ export default class InvocationModel {
...
   }
 
+  getFiles(event: build_event_stream.IBuildEvent): build_event_stream.IFile[] {
+    if (!event?.completed) {
+      return [];
+    }
+    if (event.completed.importantOutput?.length) {
+      return event.completed.importantOutput;
+    }
+    return event.completed.outputGroup?.flatMap((group) => group.fileSets).flatMap((set) => this.files.get(set.id));
```
bduffany, line 525: nit: maybe change to `this.files.get(set.id) || []` just to be safe?

app/invocation/invocation_model.tsx
```diff
@@ -29,7 +29,7 @@ export default class InvocationModel {
   brokenTest: build_event_stream.BuildEvent[] = [];
   flakyTest: build_event_stream.BuildEvent[] = [];
   timeoutTest: build_event_stream.BuildEvent[] = [];
-  files: build_event_stream.NamedSetOfFiles[] = [];
+  files: Map<string, build_event_stream.IFile[]> = new Map();
```
bduffany, line 32: nit: consider making this `private` so people don't accidentally reference this instead of `getFiles`

Also consider adding a comment describing what the keys represent, or rename to something like `xToFilesMap`

bduffany (approved):

app/invocation/invocation_model.tsx
siggisim: Done

app/invocation/invocation_model.tsx
siggisim: Done

app/invocation/invocation_model.tsx
bduffany: Sry I meant like

```
.flatMap((set) => this.fileSetIDToFilesMap.get(set.id) || [])
```

(`.get()` returns undefined if the entry isn't found in the map)

siggisim: Will send fix in another PR

PR #1358, opened 2022-01-04
Allow parsing resource names referring to zstd-compressed blobs

bduffany:
Allows the bytestream server to parse upload and download resource names for zstd-compressed blobs:
- Upload resource names follow the pattern `{instance_name}/uploads/{uuid}/compressed-blobs/{compressor}/{uncompressed_hash}/{uncompressed_size}{/optional_metadata}`
- Download resource names follow the pattern `{instance_name}/compressed-blobs/{compressor}/{uncompressed_hash}/{uncompressed_size}`

Changes:
* Expand `InstanceNameDigest` to include the blob compression type (either identity or zstd for now)
* Rename `InstanceNameDigest` to `ResourceName`
* Update string parsing methods to return `*ResourceName, error` instead of `string, *repb.Digest, error` and update regex matchers to recognize `compressed-blobs/zstd`
* Add tests

The only functional difference expected to be introduced by this PR is that we no longer immediately return `InvalidArgumentError` when we see Bytestream read/write requests for `compressed-blobs/zstd/DIGEST`. This should not matter to any clients that properly implement the RBE protocol, since we don't yet advertise zstd support via the capabilities API. But if we do happen do get zstd requests for some reason, the worst that can happen is:
* For writes, the checksum will fail upon write completion, since we don't yet decompress the incoming stream to compute the uncompressed digest
* For reads, we will return decompressed blobs even if zstd blobs are requested, and decompression will fail on the client

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -156,7 +156,7 @@ func trimPathPrefix(fullPath, prefix string) string {
 }
 
 type fileToUpload struct {
-	ad           *digest.ResourceName
+	resourceName *digest.ResourceName
 	info         os.FileInfo
 	fullFilePath string
 
@@ -181,7 +181,7 @@ func newDirToUpload(instanceName, parentDir string, info os.FileInfo, dir *repb.
 	return &fileToUpload{
 		fullFilePath: filepath.Join(parentDir, info.Name()),
 		info:         info,
-		ad:           ad,
+		resourceName: ad,
 		data:         data,
 		dir:          dir,
 	}, nil
@@ -196,7 +196,7 @@ func newFileToUpload(instanceName, parentDir string, info os.FileInfo) (*fileToU
 	return &fileToUpload{
 		fullFilePath: fullFilePath,
 		info:         info,
-		ad:           ad,
+		resourceName: ad,
 	}, nil
 }
 
@@ -215,7 +215,7 @@ func (f *fileToUpload) ReadSeekCloser() (io.ReadSeekCloser, error) {
 func (f *fileToUpload) OutputFile(rootDir string) *repb.OutputFile {
 	return &repb.OutputFile{
 		Path:         trimPathPrefix(f.fullFilePath, rootDir),
-		Digest:       f.ad.Digest,
+		Digest:       f.resourceName.GetDigest(),
 		IsExecutable: f.info.Mode()&0111 != 0,
 	}
 }
@@ -223,7 +223,7 @@ func (f *fileToUpload) OutputFile(rootDir string) *repb.OutputFile {
 func (f *fileToUpload) FileNode() *repb.FileNode {
 	return &repb.FileNode{
 		Name:         f.info.Name(),
-		Digest:       f.ad.Digest,
+		Digest:       f.resourceName.GetDigest(),
 		IsExecutable: f.info.Mode()&0111 != 0,
 	}
 }
@@ -231,14 +231,14 @@ func (f *fileToUpload) FileNode() *repb.FileNode {
 func (f *fileToUpload) OutputDirectory(rootDir string) *repb.OutputDirectory {
 	return &repb.OutputDirectory{
 		Path:       trimPathPrefix(f.fullFilePath, rootDir),
-		TreeDigest: f.ad.Digest,
+		TreeDigest: f.resourceName.GetDigest(),
 	}
 }
 
 func (f *fileToUpload) DirNode() *repb.DirectoryNode {
 	return &repb.DirectoryNode{
 		Name:   f.info.Name(),
-		Digest: f.ad.Digest,
+		Digest: f.resourceName.GetDigest(),
 	}
 }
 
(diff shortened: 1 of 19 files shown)


server/remote_cache/digest/digest.go
```diff
@@ -39,25 +39,35 @@ var (
...
 
-func NewInstanceNameDigest(d *repb.Digest, instanceName string) *InstanceNameDigest {
-	return &InstanceNameDigest{
+func NewResourceName(d *repb.Digest, instanceName string, compressor repb.Compressor_Value) *ResourceName {
+	return &ResourceName{
 		Digest:       d,
 		instanceName: instanceName,
+		compressor:   compressor,
 	}
 }
```
tylerwilliams, line 59: I think it's worth being really really careful with the naming of this struct because it's used *everywhere*. Calling this struct a "ResourceName" feels wrong to me because "resource name" already means something: it's the string name for an uploaded or downloaded digest, like "/blobs/foo/my_cool_digest/123".

My first inclination was to call this just Digest, but i think it's a bit confusing because this Digest would not be the same as repb.Digest (which does not include compression or instanceName).

We could call this struct "Resource", which is about as bad as Digest tbh, but works with ResourceName below.

server/remote_cache/digest/digest.go
```diff
@@ -39,25 +39,35 @@ var (
...
 
-func (i *InstanceNameDigest) GetInstanceName() string {
-	return i.instanceName
+func NewInstanceNameDigest(d *repb.Digest, instanceName string) *ResourceName {
```
tylerwilliams, line 61: we should call this something else -- NewUncompressedResource?

tylerwilliams (changes requested):

server/remote_cache/digest/digest.go
bduffany: Resolved offline, the reasoning for naming this `ResourceName` is that the original naming of the proto field "resource name" is likely referring to a URN (Uniform Resource Name), and so it makes sense to follow a similar to convention as `*url.URL` in the go stdlib, where the `*url.URL` refers to a struct, while a `string` refers to a serialized version of the struct, but both are called "URL".

Also sticking with `ResourceName` instead of `Resource` to avoid confusion since this thing is a "pointer" to a resource and not the resource itself.

server/remote_cache/digest/digest.go
bduffany: Resolved offline to call this `NewResourceName` and then have a separate method for setting the compressor, since it's much less common for compression to be set at all.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -156,7 +156,7 @@ func trimPathPrefix(fullPath, prefix string) string {
 }
 
 type fileToUpload struct {
-	ad           *digest.InstanceNameDigest
+	ad           *digest.ResourceName
```
vadimberezniker, line 159: nit: change member to resourceName?

enterprise/server/test/integration/remote_execution/rbeclient/rbeclient.go
```diff
@@ -80,7 +80,7 @@ type Command struct {
 
 	gRPCClientSource GRPCClientSource
 
-	actionDigest *digest.InstanceNameDigest
+	actionDigest *digest.ResourceName
```
vadimberezniker, line 83: nit: rename to resourceName?

vadimberezniker (approved):

server/remote_cache/digest/digest.go
```diff
@@ -39,25 +39,62 @@ var (
...
-	downloadRegex    = regexp.MustCompile("^(?:(?P<instance_name>.*)/)?blobs/(?P<hash>[a-f0-9]{64})/(?P<size>\\d+)")
-	actionCacheRegex = regexp.MustCompile("^(?:(?P<instance_name>.*)/)?blobs/ac/(?P<hash>[a-f0-9]{64})/(?P<size>\\d+)")
+	uploadRegex      = regexp.MustCompile(`^(?:(?:(?P<instance_name>.*)/)?uploads/(?P<uuid>[a-f0-9-]{36})/)?(?P<blob_type>blobs|compressed-blobs/zstd)/(?P<hash>[a-f0-9]{64})/(?P<size>\d+)`)
+	downloadRegex    = regexp.MustCompile(`^(?:(?P<instance_name>.*)/)?(?P<blob_type>blobs|compressed-blobs/zstd)/(?P<hash>[a-f0-9]{64})/(?P<size>\d+)`)
+	actionCacheRegex = regexp.MustCompile(`^(?:(?P<instance_name>.*)/)?(?P<blob_type>blobs|compressed-blobs/zstd)/ac/(?P<hash>[a-f0-9]{64})/(?P<size>\d+)`)
 )
 
-type InstanceNameDigest struct {
+type ResourceName struct {
 	*repb.Digest
```
tylerwilliams, line 48: If you don't have a strong reason for embedding the Digest -- now would be a great time to have callsites stop referring to it directly (r.Digest) and instead use an instance method on ResourceName (r.GetDigest()).

That way if later we decide not to embed it or to store it differently, we don't need to change code.

enterprise/server/remote_execution/dirtools/dirtools.go
bduffany: Done

enterprise/server/test/integration/remote_execution/rbeclient/rbeclient.go
bduffany: Done

server/remote_cache/digest/digest.go
bduffany: SG, done

tylerwilliams (approved):
LGTM

PR #1359, opened 2022-01-04
Add zstd transcoding support to CAS

bduffany:
Behind a flag, accept zstd reads/writes, compressing outgoing blobs and decompressing incoming blobs on the fly. Compressed bytes are not stored in cache.

Transcoding for ByteStream requests to follow in a separate PR.

---

**Version bump**: Patch


server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -266,11 +285,57 @@ func (s *ContentAddressableStorageServer) BatchReadBlobs(ctx context.Context, re
 		} else {
 			blobRsp.Status = &statuspb.Status{Code: int32(codes.OK)}
 		}
+
+		if blobRsp.Status.Code == int32(codes.OK) && s.env.GetConfigurator().GetCacheZstdTranscodingEnabled() && clientAcceptsCompressor(req.AcceptableCompressors, repb.Compressor_ZSTD) {
+			blobRsp = zstdCompressResponse(blobRsp, req.AcceptableCompressors)
```
vadimberezniker, line 290: What do you think of putting this under the existing else clause?

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -266,11 +285,57 @@ func (s *ContentAddressableStorageServer) BatchReadBlobs(ctx context.Context, re
...
+			return true
+		}
+	}
+	return false
+}
+
+func zstdCompressResponse(blobRsp *repb.BatchReadBlobsResponse_Response, acceptableCompressors []repb.Compressor_Value) *repb.BatchReadBlobsResponse_Response {
+	data, err := zstd.Compress(nil, blobRsp.Data)
+	if err != nil {
+		log.Errorf("Failed to compress blob: %s", err.Error())
```
vadimberezniker, line 319: nit: `err.Error()` -> `err`

vadimberezniker (approved):

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
bduffany: I was kinda 50/50 on that since it makes the code more nested, but it's probably clearer to put it in the else clause. Done

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
bduffany: Done

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -265,12 +284,57 @@ func (s *ContentAddressableStorageServer) BatchReadBlobs(ctx context.Context, re
...
+	return false
+}
+
+func zstdCompressResponse(blobRsp *repb.BatchReadBlobsResponse_Response, acceptableCompressors []repb.Compressor_Value) *repb.BatchReadBlobsResponse_Response {
+	data, err := zstd.Compress(nil, blobRsp.Data)
+	if err != nil {
+		log.Errorf("Failed to compress blob: %s", err)
+		return &repb.BatchReadBlobsResponse_Response{
+			Digest: blobRsp.Digest,
+			Status: gstatus.Convert(err).Proto(),
```
tylerwilliams, line 321: is this error coming from zstd? might want to wrap in an InternalErrorf before converting to proto

tylerwilliams (approved):
LGTM (1 nit)

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -265,12 +284,57 @@ func (s *ContentAddressableStorageServer) BatchReadBlobs(ctx context.Context, re
...
+			return true
+		}
+	}
+	return false
+}
+
+func zstdCompressResponse(blobRsp *repb.BatchReadBlobsResponse_Response, acceptableCompressors []repb.Compressor_Value) *repb.BatchReadBlobsResponse_Response {
+	data, err := zstd.Compress(nil, blobRsp.Data)
+	if err != nil {
+		log.Errorf("Failed to compress blob: %s", err)
```
tylerwilliams, line 318: nit: do you really want to log this? It doesn't seem particularly actionable post-development
```suggestion
		log.Errorf("Failed to compress blob: %s", err)
```

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
bduffany: Good catch, fixed

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
bduffany: Since this is never really expected (we should always be able to compress a stream of bytes) an error here could indicate that our compression impl could somehow be broken for some cases we aren't aware of, and users might not tell us about it. We can remove it later if we find that there are cases where it logs errors that seem reasonable / expected, but for now I wanted to be a bit cautious and log them as errors

PR #1361, opened 2022-01-04
Use `ui_actions_shown` to determine number of log lines to buffer for ANSI

tempoz:
Approximates flag parsing by shlex splitting and assuming no options have values of `--curses`
or `--ui_actions_shown`. Actually correct parsing would require reproducing the entirety of all
possible bazel flags in go, which is too much. If we decide correctly parsed options are necessary,
we can instead buffer events until the OptionsParsed event arrives and use the values in that, but
that, too, is cumbersome.

---

**Version bump**: Patch

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -663,11 +663,7 @@ func (e *EventChannel) handleEvent(event *pepb.PublishBuildToolEventStreamReques
 				return err
 			}
 			if e.env.GetConfigurator().GetStorageEnableChunkedEventLogs() {
-				linesToRetain, err := getCursesLinesFromStartedBuildEvent(&bazelBuildEvent)
-				if err != nil {
-					return err
-				}
-				e.logWriter = eventlog.NewEventLogWriter(e.ctx, e.env.GetBlobstore(), e.env.GetKeyValStore(), iid, linesToRetain)
+				e.logWriter = eventlog.NewEventLogWriter(e.ctx, e.env.GetBlobstore(), e.env.GetKeyValStore(), iid, getCursesLinesFromStartedBuildEvent(&bazelBuildEvent))
 			}
 		}
 
@@ -795,26 +791,26 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
 	return "", nil
 }
 
-func getCursesLinesFromStartedBuildEvent(event *build_event_stream.BuildEvent) (int, error) {
+func getCursesLinesFromStartedBuildEvent(event *build_event_stream.BuildEvent) int {
 	// the number of lines curses can overwrite is 2 + the ui_actions shown:
 	// 1 for the progress tracker, 1 for each action, and 1 blank line
 	options, err := extractOptionsFromStartedBuildEvent(event)
 	if err != nil {
-		return 0, err
+		log.Errorf("Could not extract options for ui_actions_show, defaulting to 8: %v", err)
+		return 10
 	}
-	actionsShownValues, err := getValuesForOptionNameFromOptions(options, "ui_actions_shown")
+	optionsList, err := shlex.Split(options)
 	if err != nil {
-		return 0, err
+		log.Errorf("Could not shlex split options for ui_actions_show, defaulting to 8: %v", err)
+		return 10
 	}
-	cursesValues, err := getValuesForOptionNameFromOptions(options, "ui_actions_shown")
-	if err != nil {
-		return 0, err
-	}
+	actionsShownValues := getValuesForOptionNameFromOptionsList(optionsList, "ui_actions_shown")
+	cursesValues := getValuesForOptionNameFromOptionsList(optionsList, "ui_actions_shown")
 	if len(cursesValues) > 1 {
-		log.Errorf("Too many arguments to curses, assuming auto : %v", cursesValues)
+		log.Errorf("Too many arguments to curses, assuming auto: %v", cursesValues)
 	}
 	if len(cursesValues) == 1 && cursesValues[0] == "no" {
-		return 0, nil
+		return 0
 	}
 	if len(actionsShownValues) > 1 {
 		log.Errorf("Too many arguments to ui_actions_shown, defaulting to 8: %v", actionsShownValues)
@@ -824,20 +820,16 @@ func getCursesLinesFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
 		if err != nil {
 			log.Errorf("Invalid argument to ui_actions_shown, defaulting to 8: %v", err)
 		} else if n < 1 {
-			return 3, nil
+			return 3
 		} else {
-			return n + 2, nil
+			return n + 2
 		}
 	}
-	return 10, nil
+	return 10
 }
 
-func getValuesForOptionNameFromOptions(options, value string) ([]string, error) {
+func getValuesForOptionNameFromOptionsList(optionsList []string, value string) []string {
 	values := []string{}
-	optionsList, err := shlex.Split(options)
-	if err != nil {
-		return values, err
-	}
 	flag := "--" + value
 	for i, option := range optionsList {
 		if option == "--" {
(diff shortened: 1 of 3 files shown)


tempoz:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1041

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -661,7 +663,11 @@ func (e *EventChannel) handleEvent(event *pepb.PublishBuildToolEventStreamReques
 				return err
 			}
 			if e.env.GetConfigurator().GetStorageEnableChunkedEventLogs() {
-				e.logWriter = eventlog.NewEventLogWriter(e.ctx, e.env.GetBlobstore(), e.env.GetKeyValStore(), iid)
+				linesToRetain, err := getCursesLinesFromStartedBuildEvent(&bazelBuildEvent)
```
tylerwilliams, line 666: It feels a little scary to me to depend on this certain flag in the build started event, otherwise the whole bes will be an error. Could we have a sane default here (whatever it was before this CL) and then optionally attempt to use the value from the event?

tempoz: Yup, no problem; done.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -789,6 +791,63 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
...
+		log.Errorf("Could not extract options for ui_actions_show, defaulting to 8: %v", err)
+		return 10
+	}
+	optionsList, err := shlex.Split(options)
+	if err != nil {
+		log.Errorf("Could not shlex split options for ui_actions_show, defaulting to 8: %v", err)
+		return 10
+	}
+	actionsShownValues := getValuesForOptionNameFromOptionsList(optionsList, "ui_actions_shown")
+	cursesValues := getValuesForOptionNameFromOptionsList(optionsList, "ui_actions_shown")
```
vadimberezniker, line 808: I'm assuming `ui_actions_shown` here is supposed to be `curses`.

vadimberezniker, line 808: I think it would be cleaner if you split the logic to return the `ui_actions_shown` value into a separate function and then have this function add +2 to the value returned from this new function (and return 0 to mean that no lines should be buffered)

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -789,6 +791,63 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
 	return "", nil
 }
 
+func getCursesLinesFromStartedBuildEvent(event *build_event_stream.BuildEvent) int {
+	// the number of lines curses can overwrite is 2 + the ui_actions shown:
```
vadimberezniker, line 795: nit: getCursesLines -> getNumCursesLines so it's clear that we are returning the number of lines, not lines

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -789,6 +791,63 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
...
+	// 1 for the progress tracker, 1 for each action, and 1 blank line
+	options, err := extractOptionsFromStartedBuildEvent(event)
+	if err != nil {
+		log.Errorf("Could not extract options for ui_actions_show, defaulting to 8: %v", err)
```
vadimberezniker, line 799: nit: `Warningf` would be slightly more appropriate in this function as these errors don't indicate anything particularly severe.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -789,6 +791,63 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
...
+		} else if n < 1 {
+			return 3
+		} else {
+			return n + 2
+		}
+	}
+	return 10
+}
+
+func getValuesForOptionNameFromOptionsList(optionsList []string, value string) []string {
```
vadimberezniker, line 831: What do you think calling this something more concise such as `getOptionValues`?

vadimberezniker, line 831: nit: `optionsList` -> `options`.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -789,6 +791,63 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
...
+	values := []string{}
+	flag := "--" + value
+	for i, option := range optionsList {
+		if option == "--" {
+			break
+		}
+		if strings.HasPrefix(option, flag+"=") {
+			values = append(values, strings.TrimPrefix(option, flag+"="))
+		} else if option == flag {
```
vadimberezniker, line 840: Is this to handle the case where the flags are split like ['--flag', 'value']? I did a quick test and Bazel appears to normalize the flags so that they always come in --flag=value format. Can you double check as well?

server/build_event_protocol/build_event_handler/build_event_handler.go
vadimberezniker, line 831: Should `value` be `name` or `option` or `optionName`?

vadimberezniker (changes requested):

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: Yup, np.

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: Fixed.

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: You are 100% correct.

tempoz: I can do that.

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: Fine with me.

tempoz: I think `value` should be `optionName`? That seems clearest. I don't think `option` should be `optionName` because `option` includes the preceding `--` as well as the subsequent `=<value>`, so I think it's more than the name. I've adjusted the PR to reflect this, let me know what you think.

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: I didn't realize; that makes this much easier. That is what this was for.

tempoz: Just tried it, seems like you're right.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -789,6 +798,55 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
 	return "", nil
 }
 
+func getNumActionsShownFromStartedBuildEvent(event *build_event_stream.BuildEvent) int {
+	options, err := extractOptionsFromStartedBuildEvent(event)
+	if err != nil {
+		log.Warningf("Could not extract options for ui_actions_show, defaulting to 8: %v", err)
+		return 8
```
vadimberezniker, line 805: Can you add a constant for the default # of lines and use that throughout?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -789,6 +798,55 @@ func extractOptionsFromStartedBuildEvent(event *build_event_stream.BuildEvent) (
...
+		} else if n < 1 {
+			return 1
+		} else {
+			return n
+		}
+	}
+	return 8
+}
+
+func getOptionValues(options []string, optionName string) []string {
```
vadimberezniker, line 836: Instead of collecting all the values, it would be simpler to return the value that appears the latest. That should match the Bazel flag semantics.

vadimberezniker (approved):

bduffany:
I am seeing this error get printed during a local build where I'm not changing `ui_actions_shown` from the default value. Does that mean that it might need to be `ui_actions_shown + 1` ?

PR #1362, opened 2022-01-04
Add zstd transcoding support to ByteStream server

bduffany:
---

**Version bump**: Patch


server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -213,46 +232,60 @@ func (s *ByteStreamServer) initStreamState(ctx context.Context, req *bspb.WriteR
...
-	computedDigest := fmt.Sprintf("%x", w.checksum.Sum(nil))
-	if computedDigest != w.d.GetHash() {
-		return status.DataLossErrorf("Uploaded bytes checksum (%q) did not match digest (%q).", computedDigest, w.d.GetHash())
+	if w.decompressor != nil {
+		// Close the decompressor, flushing any currently buffered bytes to the
+		// checksum+cache multi-writer. If this fails, don't bother computing the
+		// checksum or commiting the file to cache, since the incoming data is
+		// likely corrupt anyway.
+		if err := w.decompressor.Close(); err != nil {
+			log.Warning(err.Error())
```
tylerwilliams, line 277: Do we really need to log this?

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -342,3 +380,107 @@ func (s *ByteStreamServer) QueryWriteStatus(ctx context.Context, req *bspb.Query
...
+	}()
+	return d
+}
+
+func (d *zstdDecompressor) Write(p []byte) (int, error) {
+	return d.pw.Write(p)
+}
+
+func (d *zstdDecompressor) Close() error {
+	d.pw.Close()
```
tylerwilliams, line 454: does this return an error that we care about?

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -152,11 +162,17 @@ func (s *ByteStreamServer) Read(req *bspb.ReadRequest, stream bspb.ByteStream_Re
...
 type writeState struct {
-	writer             io.WriteCloser
-	checksum           hash.Hash
+	// Top-level writer that handles incoming bytes.
+	writer io.Writer
+	// Optional writer that decompresses incoming bytes and writes them to both
+	// the checksum and the cache.
+	decompressor io.WriteCloser
+	// Writer which writes uncompressed bytes to the cache.
+	cacheWriter        io.WriteCloser
```
tylerwilliams, line 171: this is confusing -- we went from one WriteCloser to 3 (well, 2 WriteClosers and 1 Writer).

I think the complexity is happening because we're using io.MultiWriter -- could we add a MultiWriteCloser and stick with the one WriteCloser?

hint: https://gist.github.com/wader/d82787b91391b71cf39d4f55d98c6d88

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -342,3 +380,107 @@ func (s *ByteStreamServer) QueryWriteStatus(ctx context.Context, req *bspb.Query
 		Complete:      false,
 	}, nil
 }
+
+type Checksum struct {
```
tylerwilliams, line 384: nit: see above but if we simplify the above -- do we need this class?

tylerwilliams (changes requested):

server/remote_cache/byte_stream_server/byte_stream_server.go
bduffany: The decompressor encloses a Writer so I needed this class to satisfy the Writer interface. I figured it made sense to move the other checksum related stuff in here as well.

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -187,6 +202,9 @@ func (s *ByteStreamServer) initStreamState(ctx context.Context, req *bspb.WriteR
 	if err != nil {
 		return nil, err
 	}
+	if !s.supportsCompressor(r.GetCompressor()) {
+		return nil, status.UnimplementedErrorf("Unsupported compressor %s", r.GetCompressor())
```
tylerwilliams, line 206: above you return:

		return status.InvalidArgumentErrorf("Unsupported compressor %s", r.GetCompressor())
Are you sure these should be different?

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -213,46 +231,63 @@ func (s *ByteStreamServer) initStreamState(ctx context.Context, req *bspb.WriteR
...
+		cacheWriteCloser = devnull.NewWriteCloser()
 	}
-	ws.checksum = sha256.New()
-	ws.writer = wc
-	ws.alreadyExists = exists
-	if exists {
-		ws.bytesWritten = r.GetDigest().GetSizeBytes()
-	} else {
-		ws.bytesWritten = 0
+	ws.cacheCloser = cacheWriteCloser
```
tylerwilliams, line 247: cacheWriteCloser := devnull.NewWriteCloser()
if r.GetDigest().GetHash() != digest.EmptySha256 && !exists {
		cacheWriteCloser, err = cache.Writer(ctx, r.GetDigest())
		if err != nil {
			return nil, err
		}
	}

tylerwilliams (changes requested):
See inline comments

tylerwilliams (approved):

server/remote_cache/byte_stream_server/byte_stream_server.go
bduffany: Nope, changed the above to Unimplemented as well.

server/remote_cache/byte_stream_server/byte_stream_server.go
bduffany: Nice, done

PR #1364, opened 2022-01-05
Add cloud.google.com/go/iam dependency to fix internal build

siggisim:
Not really sure why this only causes an issue in the internal repo, but I believe it has something to do with: https://github.com/bazelbuild/bazel-gazelle/issues/619 and/or https://github.com/bazelbuild/bazel-gazelle/issues/680

The problem is we can't easily add this directive to the internal WORKSPACE file because deps.bzl is in a different repo (the external one):
```
# gazelle:repository_macro deps.bzl%install_buildbuddy_dependencies
```

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1366, opened 2022-01-05
Expose zstd capabilities

bduffany:
To be pushed in a subsequent release after the `--cache.zstd_transcoding_enabled` flag is live on all servers.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1368, opened 2022-01-05
Improve SQL Perf on Test Grid

luluz66:
Use start_time_usec from TargetStatuses for start and end time
condition instead of created_at_usec from Invocations.

MySQL Explain shows
 - query cost is 29324.77 before the change and 5322.02 after
 - for the second join (TargetStatuses), before the change, it produced
   24287 rows and 2698 rows after.

Also, for the end time condition, just check the start time and drop the
duration. If we calculate the finish time, the query cost become
10355.61.

In addition, enable target tracking locally.


tylerwilliams (approved):
LGTM

siggisim:
I'm sure you already gave this a whirl - but throwing this out there just in case. We have an index on Invocations updated_at_usec (but not created_at_usec). Would be interested in how using that (or both?) compares to using created_at_usec.

luluz66:
> I'm sure you already gave this a whirl - but throwing this out there just in case. We have an index on Invocations updated_at_usec (but not created_at_usec). Would be interested in how using that (or both?) compares to using created_at_usec.

Unfortunately neither updated_at_usec or created_at_usec index is used. The execution plan joined the three tables in the following order:
Targets (using group_id as the key) -> TargetStatuses (using target_id as the key) -> Invocations (using invocation_invocation_pk as the key)

PR #1373, opened 2022-01-06
Include service name in gRPC reqs by method dashboard

bduffany:
Multiple services have the same method names (esp. names like "Read" / "Write") so the service name is used to disambiguate.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1374, opened 2022-01-06
Fix impl of WriteResponse.committed_size field

bduffany:
When Bazel actually completes the upload (as opposed to the upload immediately returning because the digest already exists) -- it will check whether the committed_size matches the _raw_ size of the stream written. The previous impl of committed_size returned the number of _uncompressed_ bytes written.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

tylerwilliams:
nit: not in this CL, but could we add an integration test where bazel actually sets this flag and does a small build?

bduffany:
> nit: not in this CL, but could we add an integration test where bazel actually sets this flag and does a small build?

yep, sg

PR #1376, opened 2022-01-07
Show execution server name instead of executor id in UI

tempoz:
Executor IDs are useful internally for running jobs where relevant build artifacts remain cached,
but for users it is generally more meaningful/useful to know the name of the execution server,
so we make that the execution identifier exposed to users in the app.

---

**Version bump**: Minor


app/invocation/invocation_action_card.tsx
```diff
@@ -424,10 +424,8 @@ export default class InvocationActionCardComponent extends React.Component<Props
                       <div className="action-property-title">Execution metadata</div>
                       {this.state.actionResult.executionMetadata ? (
                         <div className="action-list">
-                          <div className="metadata-title">Worker</div>
+                          <div className="metadata-title">Execution Server Name</div>
```
vadimberezniker, line 427: Should we call this the 'Executor Host ID'? 
My thinking:
We use 'executor' throughout our system to refer to the workers so it seems worthwhile to stay consistent. 
Host makes it clearer that we are referring to the machine that is hosting the executor rather than the executor itself.
And ID vs name because this value does not have any meaning to the user, it's ultimately a generated ID.

This is very subjective so please let me know if you disagree with any of this.

proto/scheduler.proto
```diff
@@ -203,6 +203,11 @@ message ExecutionNode {
   //
   // Ex. "34c5cf7e-b3b1-4e20-b43c-3e196b30d983"
   string executor_id = 9;
+
+  // Name of the execution server this executor is running on
+  //
+  // Ex. "8BiY6U0F"
```
vadimberezniker, line 209: I think it would look nicer if we displayed this value as a UUID rather than the base64 encoded value of the UUID. 
base64 encoding can have characters like + / etc which doesn't look very nice. 

@tylerwilliams Do you have any opinions on this?

vadimberezniker (changes requested):

app/invocation/invocation_action_card.tsx
tylerwilliams: +1 good name

proto/scheduler.proto
tylerwilliams: No preference, as long as we can make it clickable in the UI. I don't think these are full UUIDs though, they are just the host part of the ID, right?

(We have already written many of these in the currrent format to the executions table so there's some inertia there...)

app/invocation/invocation_action_card.tsx
tempoz: I like and will implement the name change, but on the topic of why I picked name, I do want to note that I think this ID could be something meaningful in the presence of a name override: https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/executor/executor.go#L81-L84 , though that appears to be an option we don't currently use: https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/cmd/executor/executor.go#L294

proto/scheduler.proto
tempoz: Yep, these are the node IDs used by UUID (which is to say, generally, the MAC addresses) encoded to string by the base64 library. If we want to change how we're formatting the ID, given that it generally is a MAC address, maybe we should format it like one?

app/invocation/invocation_action_card.tsx
tempoz: I like the name change, but I do want to note that I think this ID could be something meaningful in the presence of a name override: https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/executor/executor.go#L81-L84 , though that appears to be an option we don't currently use: https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/cmd/executor/executor.go#L294

tempoz: Oh, actually, we call this entity an execution server extensively in the pre-existing code, should I change those names to `ExecutorHost` to be consistent with the new naming scheme? Otherwise I am concerned this could be confusing later.

proto/scheduler.proto
tempoz: After talking with Vadim and Tyler offline, it looks like the direction we're going with this is to just use GUIDs (formatted normally) that are stored to disk, since MAC addresses can get re-used. I'll address this in a separate PR, since it's a bit beyond the scope of this one.

tylerwilliams: (gonna wait on reviewing this CL till after that one is in)

app/invocation/invocation_action_card.tsx
tempoz: Talked offline; that would be work for a future PR, if ever.

proto/scheduler.proto
tempoz: https://github.com/buildbuddy-io/buildbuddy/pull/1400

tempoz:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/671

enterprise/server/cmd/executor/executor.go
```diff
@@ -332,7 +332,7 @@ func main() {
 	http.Handle("/readyz", env.GetHealthChecker().ReadinessHandler())
 
 	schedulerOpts := &scheduler_client.Options{}
-	reg, err := scheduler_client.NewRegistration(env, taskScheduler, executorID, schedulerOpts)
+	reg, err := scheduler_client.NewRegistration(env, taskScheduler, executorID, executionServer.Name(), schedulerOpts)
```
bduffany, line 335: I think the local variable definition in this function (`executionServer, ... := executor.NewExecutor(...)`) is leading to some confusion -- the local variable would probably be better named `executor`, because we don't really think of executors as servers in the current model (executors act as clients of the app, and the app schedules execution tasks across the registered client executors). We used to use a server model, which is probably why this variable name was originally chosen.

The name `ExecutionServer`, at least internally, typically refers to the gRPC `ExecutionService` implementation (`execution_server.go`) that lives on the app server, which accepts `Execute` requests and then schedules them as tasks to be executed on one of the client executors.

Users don't necessarily know/care about this architecture, but so far we have used the term "executor" to refer to the client executor binaries, both in our docs and in the UI (e.g. "use self-hosted executors" in settings). So I think we should replace the term "execution server name" with "executor name" throughout this PR

bduffany (changes requested):

enterprise/server/cmd/executor/executor.go
tempoz: So the reason I used execution server name instead of executor name is that the instances of the executor binaries already have IDs/names associated with them that are not these names; these names are instead the persistent names associated with the machine an executor runs on. I am concerned that if we call both things "executor names", no one looking at the code will have any idea what the difference between the two is supposed to be.

bduffany: +1 for "executor host" or "executor host ID" like Vadim mentioned in another thread

tempoz: sgtm

tempoz: Fixed.

bduffany (approved):

proto/scheduler.proto
```diff
@@ -203,6 +203,11 @@ message ExecutionNode {
   //
   // Ex. "34c5cf7e-b3b1-4e20-b43c-3e196b30d983"
   string executor_id = 9;
+
+  // ID of the host this executor is running on
+  //
+  // Ex. "8BiY6U0F"
```
vadimberezniker, line 209: nit: update example to match new format

vadimberezniker (approved):

PR #1377, opened 2022-01-07
Ensure zstd decompressor is always closed, even if clients do not send a FinishWrite request

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/byte_stream_server/byte_stream_server.go
+++ b/server/remote_cache/byte_stream_server/byte_stream_server.go
@@ -173,6 +173,7 @@ type writeState struct {
 	activeResourceName string
 	offset             int64
 	alreadyExists      bool
+	closed             bool
 }
 
 func checkInitialPreconditions(req *bspb.WriteRequest) error {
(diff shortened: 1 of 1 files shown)


server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -259,6 +260,11 @@ func (w *writeState) Write(buf []byte) error {
 }
 
 func (w *writeState) Close() error {
+	if w.closed {
+		return nil
+	}
+	w.closed = true
+
```
tylerwilliams, line 267: could you nil out w.decompressorCloser after closing it? and remove this block?

bduffany: lol, I had that before, but was on the fence. Changed it back

tylerwilliams: it's somewhat idiomatic in go to do that afaik. Don't have a source handy tho, lmk if you find one.

tylerwilliams (approved):
LGTM

PR #1378, opened 2022-01-07
use invocation id bytes form as FK for TargetStatus

luluz66:
1. Add a new column Binary (16) in Invocation and TargetStatus to store
   bytes form of invocation id.
2. Add a util function to convert uuid string format to bytes format
3. Set this field when we insert/update TargetStatus and Invocation.
4. Migration code to populate this field. There are two versions of
   migration code, depending on what the dialect is. For SQLite, the rows
   need to be closed before updates are executed; otherwise, it result
   in a deadlock when testing locally.


server/tables/tables.go
```diff
@@ -563,6 +577,61 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
+						return err
+					}
+				}
+				return nil
+			})
+		} else if db.Dialector.Name() == mysqlDialect {
+			postMigrate = append(postMigrate, func() error {
+				if err := db.Exec(`UPDATE Invocations SET invocation_uuid = UUID_TO_BIN(invocation_id)`).Error; err != nil {
+					return err
+				}
```
luluz66, line 623: Tested locally with a fake prod db. It took ~17 min to finish updating Invocations; and ~42 mins to finish updating TargetStatuses. The whole migration took 1hr and 4 min locally.

tylerwilliams: 👍  thanks for testing the performance of this! I know how painful it can be waiting for these things to finish.

What's the plan for rolling this out -- how do you imagine we'll actually apply this update and how do we ensure that rows written during the schema migration also get migrated?

tylerwilliams (changes requested):

server/tables/tables.go
luluz66: I added a pre-migrate check. if the column exists, update invocation uuid if the column is null. The roll will look like this:
1. The first time the migration runs, it will create columns, indices and back fill invocation_uuids. During this time, other servers on the old versions might write the rows without invocation_uuids.
2. Other servers restarted. Then will go through the new migration code I just added; and updated invocation_uuids if they are null.

bduffany: I think with the code as written, the migration will run only once, since the migration is based on `!m.HasColumn(&Invocation{}, "invocation_uuid")` and the column will be created after the first new server instance completes its migration.

One thing you could do is replace that condition with (pseudo-code) `!m.HasColumn(&Invocation{}, "invocation_uuid") || query(SELECT COUNT(*) FROM Invocations WHERE invocation_uuid = NULL) > 0` -- as long as the COUNT runs very fast in the case where the migration is already complete. It'd probably be very fast since it'll see that there are no NULLs in the index, but worth verifying

bduffany: Nevermind, I missed the above case which does exactly this, lol.

tylerwilliams (changes requested):
I'm mostly good with this, would like to make the sqlite thing less of a memory hole tho.

Can you please sync with @vadimberezniker (who is doing the release tomorrow) if you submit before then, so he can make sure this rolls out smoothly?

enterprise/server/api/api_server_test.go
```diff
@@ -21,80 +22,113 @@ import (
 var userMap = testauth.TestUsers("user1", "group1")
 
 func TestGetInvocation(t *testing.T) {
+	testUUID, err := uuid.NewRandom()
+	assert.NoError(t, err)
```
vadimberezniker, line 26: for errors during setup you should use `require.NoError` which immediately terminates the test unlike `assert.NoError` which allows the test to continue even if the test setup failed

server/tables/tables.go
```diff
@@ -569,6 +583,79 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
+				return postMigrate, err
+			}
+			if err := db.Exec(
+				`UPDATE TargetStatuses ts
+			     SET invocation_uuid=(
+				   SELECT invocation_uuid FROM Invocations i 
+				   WHERE i.invocation_pk=ts.invocation_pk)
+				 WHERE invocation_uuid IS NULL  
+			    `).Error; err != nil {
+				return postMigrate, err
```
vadimberezniker, line 599: nit: can we move these migrations into separate functions? this function is getting quite long and there's a lot of nesting

vadimberezniker (approved):

server/tables/tables.go
luluz66: Done.

server/tables/tables.go
```diff
@@ -569,6 +583,79 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
+			}
+
```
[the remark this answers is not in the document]
luluz66: Done.

server/tables/tables.go
```diff
@@ -569,6 +583,79 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
+					if err != nil {
+						return err
+					}
+					ids = append(ids, invocationIDs{
+						invocationID:   invocationID,
+						invocationUUID: invocationUUID,
+						invocationPK:   invocationPK,
+					})
+				}
+				rows.Close()
```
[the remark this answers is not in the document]
luluz66: I used FindInBatches now. Tested locally and it worked.

enterprise/server/api/api_server_test.go
luluz66: Done

tylerwilliams (approved):
LGTM

PR #1379, opened 2022-01-07
Add CPU usage to the per-node graphs.

vadimberezniker:
Aggregate views will be in a separate PR.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1383, opened 2022-01-10
Remove unused containerd code

bduffany:
The containerd impl is unused currently and we can always revert this PR later if we want to explore moving to containerd sometime in the future.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1384, opened 2022-01-10
Reduce cardinality of some common SQL metrics.

vadimberezniker:
Provide ability to specify a custom query name that will override the
labels used for exported metrics.

---

**Version bump**: Patch

### DIFF
--- a/server/util/db/db.go
+++ b/server/util/db/db.go
@@ -6,10 +6,11 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	golog "log"
 	"os"
 	"strings"
 	"time"
+
+	golog "log"
 
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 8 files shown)


server/util/db/db.go
```diff
@@ -52,44 +53,53 @@ type DBHandle struct {
...
-	return &optionsImpl{
-		readOnly:        true,
-		allowStaleReads: true,
+func (dbh *DBHandle) gormHandleForOpts(ctx context.Context, opts *Options) *DB {
+	var db *DB
+	if opts.readOnly && opts.allowStaleReads && dbh.readReplicaDB != nil {
+		db = dbh.readReplicaDB
+	} else {
+		db = dbh.DB
+	}
```
tylerwilliams, line 88: maybe default init db := dbh.DB and kill the else?

tylerwilliams (approved):
LGTM

server/util/db/db.go
vadimberezniker: Done.

PR #1386, opened 2022-01-11
Use invocation proto as source of truth for `repo` `commitSHA` and `branchName`

siggisim:
Previously we were parsing these client side, which is error prone and led to a mismatch in parsing behavior between client and server.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1387, opened 2022-01-11
Use ts.created_at_usec instead of ts.start_time_usec

luluz66:
ts.start_time_usec is the time when the test is first ran. If a test is
cached; then it would be earlier than the created_at_usec.


tylerwilliams (approved):
Does this change the query performance?

luluz66:
> Does this change the query performance?

Nope. The query cost is the same.

PR #1392, opened 2022-01-12
Fix a few minor compression issues

bduffany:
* For very small blobs, fix an issue that we often issue multiple separate responses unnecessarily when the fully compressed blob could have been sent in a single response.
  * More generally, for streaming compression responses, each Read() from the compressor returns some number of bytes that is very different from our desired 1MB read buffer size. Since we directly pipe the compressor to the stream writer, this means we'll write smaller buffers with varying sizes depending on compression ratio as well as the compressor's window size, which determines how often it flushes. We can fix this by staging through an intermediate buffer that we only flush when it reaches 1MB. But the small fix in this PR should fix this issue for 99+% of blobs, and we will still want to keep this optimization anyway to avoid creating a streaming compressor+goroutine for the 99% case.
* Fix an error logged when the compressor tries to write its compressed output to a closed pipe, which just means that the client disconnected, so it can be safely ignored.

---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/byte_stream_server/BUILD
+++ b/server/remote_cache/byte_stream_server/BUILD
@@ -10,6 +10,7 @@ go_library(
         "//proto:remote_execution_go_proto",
         "//server/environment",
         "//server/interfaces",
+        "//server/remote_cache/cachetools",
         "//server/remote_cache/digest",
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):
LGTM

PR #1400, opened 2022-01-12
Generate and store GUID to uniquely identify executors

tempoz:
Falls back to NodeID if generation fails; still returns a GUID if file access fails.

---

**Version bump**: Minor


server/util/uuid/uuid.go
```diff
@@ -43,3 +54,56 @@ func StringToBytes(text string) ([]byte, error) {
...
+func fillHostId() {
+	id, err := guuid.NewRandom()
+	hostID = id.String()
+	if err != nil {
+		hostIDError = err
+		hostID = string(guuid.NodeID())
+	}
+}
+
+func GetHostID() (string, error) {
```
tylerwilliams, line 72: The fallbacks add a lot of complexity, I think we should try to simplify this.

Can we just return errs to the caller if we encounter them?

server/util/uuid/uuid.go
```diff
@@ -43,3 +54,56 @@ func StringToBytes(text string) ([]byte, error) {
...
+	hostIDOnce.Do(
+		func() {
+			userConfigDir, err := os.UserConfigDir()
+			if err != nil {
+				handleHostIDError(err)
+				return
+			}
+			hostIDFilepath := path.Join(userConfigDir, "buildbuddy", hostIDFilename)
```
tylerwilliams, line 80: nit: maybe make buildbuddy a const above and add a comment that that dir will be a subdir in the userconfig dir

server/util/uuid/uuid.go
```diff
@@ -43,3 +54,56 @@ func StringToBytes(text string) ([]byte, error) {
...
+			// try to create the file to write a new ID, if it already exists this will fail
+			hostIDFile, err := os.OpenFile(hostIDFilepath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0644)
```
tylerwilliams, line 82: would WriteFile simplify this?

tylerwilliams (changes requested):

server/util/uuid/uuid.go
tempoz: Yup, done.

server/util/uuid/uuid.go
tempoz: Done.

server/util/uuid/uuid.go
tempoz: Resolved this offline.

server/util/uuid/uuid.go
```diff
@@ -43,3 +58,51 @@ func StringToBytes(text string) ([]byte, error) {
...
+			// we successfully opened the file, generate and record the host id
+			id, hostIDError := guuid.NewRandom()
+			if hostIDError != nil {
+				// read failed from rand.Reader; basically this should never happen
+				return
+			}
+			if _, hostIDError = io.WriteString(hostIDFile, id.String()); hostIDError != nil {
+				return
+			}
+			hostID = id.String()
```
tylerwilliams, line 104: if you pull this logic out into another function you can do away with a lot of the references to hostIDError and write more natural code, then just assign the returned error to hostIDError in GetHostID().

Something like:

func getOrCreateHostID() (string, error) {
  userConfigDir, err := os.UserConfigDir()
  if err != nil {
    return "", err
  }
... etc ...
}


func GetHostID() (string, error) {
  var hostID string
  var err error 
  hostIDOnce.Do(func() {
    hostID, err = getOrCreateHostID()
  })
  return hostID, err
}

tempoz: Done, I think? I didn't move the `hostID` and `hostIDError` declarations out of their `var` block and into the function, since 
I believe that would mean they would return default values every time except the time the `once` runs.

tylerwilliams (approved):
LGTM

PR #1403, opened 2022-01-12
Optimize ByteStream.Read compression and move to klauspost/compress for zstd lib

bduffany:
See https://github.com/buildbuddy-io/buildbuddy/pull/1382 for benchmark details.

TL;DR:
* valyala is the fastest library out of the ones I tested (datadog, klauspost, valyala) but the speed improvements don't directly translate to significant gains in E2E build time that are enough to warrant bringing in a CGO dep (yet). If compression proves to be a major bottleneck as more users start taking advantage of remote cache compression, then we can switch to the valyala implementation in a future PR. There are a few remaining TODOs for that, though: stop depending on the vendored `.a` files and use something like `rules_foreign_cc` to generate them ourselves using `make`. Also need to test the build on Mac + RBE.
* Use a chunked compression approach instead of streaming compression for ByteStream.Read, which is both simpler and more performant at the cost of only a few % compression ratio. In my local tests, chunking reduced E2E build time by about 9%.

---

**Version bump**: Patch


server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -445,20 +445,28 @@ type zstdDecompressor struct {
 // Note that writes are not matched one-to-one, since the compression scheme may
 // require more than one chunk of compressed data in order to write a single
 // chunk of decompressed data.
-func NewZstdDecompressor(writer io.Writer) io.WriteCloser {
```
tylerwilliams, line 448: could this struct be moved to util/compression?

tylerwilliams (changes requested):

server/remote_cache/byte_stream_server/byte_stream_server.go
bduffany: sg, done. Moved the compressor there too, and passing in the intermediate buffers as params since they are bytestream-specific.

tylerwilliams (approved):
LGTM

PR #1417, opened 2022-01-13
Make release script python3 compatible.

vadimberezniker:
universal_newlines causes the pipe to be opened in text mode instead of
binary mode which avoids issues with python3 str vs bytes differences.

---

**Version bump**: None


tylerwilliams (approved):
lGTM

PR #1418, opened 2022-01-13
Only show kube node/pool information for the current region.

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1419, opened 2022-01-13
Give some feedback when clicking submit on forms

siggisim:
---

**Version bump**: None


siggisim:
Gunna submit this TBR because it's late and it's an annoying one.

PR #1420, opened 2022-01-13
Get firecracker_test working

bduffany:
WIP. Currently failing with

`thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CreateDir("/sys/fs/cgroup/cpuset/firecracker/f38217a7-efba-45e0-8212-d6bd1c9a146a", Os { code: 30, kind: Other, message: "Read-only file system" })', src/jailer/src/env.rs:359:34`

---

**Version bump**: Patch


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -47,6 +48,10 @@ import (
 	fcmodels "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 )
 
+var (
+	forceDebugMode = flag.Bool("debug_firecracker_force_debug_mode", false, "Whether to force firecracker debug_mode, streaming firecracker logs to the console. Not intended for production use.")
+)
+
```
tylerwilliams, line 54: do we need to introduce this here? Seems like debugmode can already be set via the opts passed into this class, wondering why we don't set this via a config flag or a flag in vmstart (if that's where you need it)?

bduffany: I found this useful while debugging both vmstart and the unit tests, which would have been a lot of places to set `DebugMode: true`. But this may not be super useful going forward -- removing for now

tylerwilliams (approved):
LGTM

PR #1422, opened 2022-01-13
Backfill invocation uuid take 2

luluz66:
Update invocations and target statuses in batches


server/tables/tables.go
```diff
@@ -578,78 +578,73 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
 	return postMigrate, nil
 }
 
-func backfillInvocationUUIDIfNull(db *gorm.DB) error {
-	if err := db.Exec(`
-		UPDATE Invocations SET invocation_uuid = UUID_TO_BIN(invocation_id) 
-		WHERE invocation_uuid IS NULL
-	`).Error; err != nil {
-		return err
+func updateInBatch(db *gorm.DB, baseQuery string, batchSize int64) error {
```
vadimberezniker, line 598: nit: let's call this updateInBatches since there will be multiple batches

server/tables/tables.go
```diff
@@ -578,78 +578,73 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
-			(SELECT invocation_uuid FROM Invocations i 
-			WHERE i.invocation_pk=ts.invocation_pk)
-		WHERE invocation_uuid IS NULL
-	`).Error
 }
 
 func postMigrateInvocationUUIDForMySQL(db *gorm.DB) error {
-	if err := db.Exec(`UPDATE Invocations SET invocation_uuid = UUID_TO_BIN(invocation_id)`).Error; err != nil {
+	updateInvocationStmt := `UPDATE Invocations SET invocation_uuid = UNHEX(REPLACE(invocation_id, "-","")) WHERE invocation_uuid IS NULL`
+	err := updateInBatch(db, updateInvocationStmt, 10000)
```
vadimberezniker, line 615: nit: inline the updateInBatch call into the if statement below

server/tables/tables.go
```diff
@@ -578,78 +578,73 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
-			invocationUUID, err := uuid.StringToBytes(invocation.InvocationID)
-			if err != nil {
-				return err
-			}
-			if err := db.Exec(`UPDATE TargetStatuses SET invocation_uuid = ? WHERE invocation_pk = ? `, invocationUUID, invocation.InvocationPK).Error; err != nil {
-				return err
-			}
-			if err := db.Exec(`UPDATE Invocations SET invocation_uuid = ? WHERE invocation_id = ? `, invocationUUID, invocation.InvocationID).Error; err != nil {
-				return err
+	res := db.Select("invocation_id", "invocation_pk").
```
vadimberezniker, line 631: nit: these line breaks are a little awkward

you could store the fn into a variable to reduce the lenth of the line

something like 
updateFn := func(tx *gorm.DB, batch int) error { ... }
res := db.Select("invocation_id", "invocation_pk").FindInBatches(&results, 100, updatefn)

vadimberezniker (approved):
LGTM, only minor nits

server/tables/tables.go
luluz66: Done

server/tables/tables.go
luluz66: Done

server/tables/tables.go
luluz66: Done

PR #1424, opened 2022-01-14
Make Xcode matching more percise

brentleyjones:
If the following Xcode versions are installed:

- 12.0.0.12A7209
- 12.1.0.12A7403
- 12.1.1.12A7605b
- 12.2.0.12B45b

Then we map these input versions to these output versions:

- 12 - > 12.2.0.12B45b
- 12.0 -> 12.0.0.12A7209
- 12.0.0 -> 12.0.0.12A7209
- 12.0.0.12A7209 -> 12.0.0.12A7209
- 12.1 -> 12.1.1.12A7605b
- 12.1.0 -> 12.1.0.12A7403
- 12.1.0.12A7403 -> 12.1.0.12A7403
- 12.1.1 -> 12.1.1.12A7605b
- 12.1.1.12A7605b -> 12.1.1.12A7605b
- 12.2 -> 12.2.0.12B45b
- 12.2.0 -> 12.2.0.12B45b
- 12.2.0.12B45b -> 12.2.0.12B45b

To summarize, we map a less percise version to the highest available version it could match. This is good.

We _also_ currently take a fully qualified version, such as  12.5.0.12E262 and if it doesn't match one of the inputs available above, we make it less percise until it potentially matches: 12.5.0, 12.5, and finally 12. This means that 12.5.0.12E262 -> 12 - > 12.2.0.12B45b, which is invalid.

This commit removes that final step, meaning in the example above, 12.5.0.12E262 will return an error saying that Xcode version isn't available.

---

**Version bump**: Patch

### DIFF
--- a/server/xcode/xcode_darwin.go
+++ b/server/xcode/xcode_darwin.go
@@ -59,8 +59,8 @@ func (x *xcodeLocator) PathsForVersionAndSDK(xcodeVersion string, sdk string) (s
 		xcodeVersion = defaultXcodeVersion
 	}
 
-	xv := x.xcodeVersionForVersionString(xcodeVersion)
-	if xv == nil {
+	xv, ok := x.versions[xcodeVersion]
+	if !ok {
 		var err error
 		if xcodeVersion == defaultXcodeVersion {
 			err = status.FailedPreconditionErrorf("Default Xcode version not set on remote executor and xcode-select set to an invalid path. Available Xcode versions are %s", versionsString(x.versions))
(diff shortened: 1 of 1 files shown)


siggisim:
I think this is a reasonable change, but I'm a little worried about this being a breaking change for folks that (knowingly or unknowingly) depend on the current behavior. 

For example teams that don't keep Xcode versions in-sync across their team (I know this isn't necessarily a great practice but it certainly happens).

Or teams that aren't building iOS software, but need Xcode to build some C stuff and don't really care about the point version as much.  

If we really think this is important, maybe we can introduce it behind a flag (could be default for new deployments, so we avoid breaking existing deployments). Or just print out a warning if this fuzzy matching happens.

Happy to chat offline about the actual deployments where I think this could cause problems.

siggisim (changes requested):
See comment

siggisim (approved):
Chatted offline, we should be good to go.

tylerwilliams (approved):
LGTM

PR #1426, opened 2022-01-14
Allow running executor without any Xcode installed

brentleyjones:
I forgot to make this change when adding support for the Command Line Tools. Technically it's not an error to not have any Xcodes installed, but we will still warn when this happens.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1429, opened 2022-01-14
Run ci_runner_test and workflow_test using CI runner image

bduffany:
This ensures that the `ci_runner`, when invoked in these tests, runs in the same environment that it would in production. In particular, this ensures it is using the same version of `git` that we use in prod.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/test/integration/ci_runner/BUILD
+++ b/enterprise/server/test/integration/ci_runner/BUILD
@@ -6,6 +6,11 @@ go_test(
     data = [
         "//enterprise/server/cmd/ci_runner",
     ],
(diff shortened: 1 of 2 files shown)


bduffany:
Here is the execution page showing that the CI runner image was actually used (see platform properties): https://app.buildbuddy.io/invocation/a03cba56-b98d-4a98-8d42-3941e0b9907a?actionDigest=e35b3fb65e754d4fd1dd01162f4dc3154298a61297514f02dd3a240391e4292b/240&actionResultDigest=e35b3fb65e754d4fd1dd01162f4dc3154298a61297514f02dd3a240391e4292b/240#action

vadimberezniker (approved):

PR #1430, opened 2022-01-14
Small fixes for firecracker networking

bduffany:
* Add iptables rule to fix broken `veth1xxxxx <=> defaultDevice` link in the case where packet forwarding is disabled in the root netns. The rule is added per-container and is removed as part of container networking cleanup.
* Avoid adding multiple masquerade rules on each run of the executor

---

**Version bump**: Patch

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,0 +1,25 @@
+# BuildBuddy CLI
+
+Install BuildBuddy CLI
+
+```
+curl -fsSL install.buildbuddy.io | bash
+```
+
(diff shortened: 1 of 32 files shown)


tylerwilliams (approved):
LGTM

bduffany:
Merging https://github.com/buildbuddy-io/buildbuddy/pull/1434/files as part of this PR, because I accidentally merged that one after it got LGTM but before this PR was merged.

PR #1432, opened 2022-01-18
Drop invocation pk column

luluz66:
1. Remove Invocation PK column from Invocations and
   TargetStatuses
2. DB Migration:
   - For MySQL, use ALTER TABLE DROP PRIMARY KEY, ADD PRIMARY KEY
     statement
   - For SQLite, rename the old TargetStatuses table before auto
     migration; therefore, gorm will create TargetStatuses table
     from the new schema; In post migration, use INSERT to backfill
     the data

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -123,7 +123,6 @@ type Invocation struct {
 	TotalUploadUsec                  int64
 	TotalCachedActionExecUsec        int64
 	DownloadThroughputBytesPerSecond int64
-	InvocationPK                     int64  `gorm:"uniqueIndex:invocation_invocation_pk"`
 	InvocationUUID                   []byte `gorm:"size:16;uniqueIndex:invocation_invocation_uuid"`
 	Success                          bool
 }
@@ -394,8 +393,7 @@ func (t *Target) TableName() string {
 type TargetStatus struct {
 	Model
 	TargetID       int64  `gorm:"primaryKey;autoIncrement:false"`
-	InvocationPK   int64  `gorm:"primaryKey;autoIncrement:false;index:target_status_invocation_pk"`
-	InvocationUUID []byte `gorm:"size:16;index:target_status_invocation_uuid"`
+	InvocationUUID []byte `gorm:"primaryKey;autoIncrement:false;size:16"`
 	TargetType     int32
 	TestSize       int32
 	Status         int32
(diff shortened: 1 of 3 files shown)


luluz66:
Tested with fake prod db: 
17 min to alter primary key
46 min to drop invocation_pk on Invocations table
2.7 min to drop invocation_pk on TargetStatuses table.

bduffany:
While the columns are being dropped, is a lock held on the table (can new rows be inserted?)

server/tables/tables.go
```diff
@@ -620,20 +631,28 @@ func postMigrateInvocationUUIDForMySQL(db *gorm.DB) error {
 			(SELECT invocation_uuid FROM Invocations i 
 			WHERE i.invocation_pk=ts.invocation_pk)
 		WHERE invocation_uuid IS NULL`
-	return updateInBatches(db, updateTargetStatusStmt, 10000)
+	if err := updateInBatches(db, updateTargetStatusStmt, 10000); err != nil {
+		return err
+	}
+	return db.Exec("ALTER TABLE TargetStatuses DROP PRIMARY KEY, ADD PRIMARY KEY(target_id, invocation_uuid) ").Error
```
vadimberezniker, line 637: According to the mysql docs [1], you should add `ALGORITHM=INPLACE, LOCK=NONE` to the alter statement to make sure it does not lock. The documentation doesn't make clear what the default locking is if you don't specify LOCK=NONE so it's better to be safe. 

[1] https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html#online-ddl-primary-key-syntax-notes
Search for "Dropping a primary key and adding another"

luluz66: Use Algorithm=Inplace, LOCK=NONE here and also for drop columns. Tested locally with fake prod db , we can write to Invocations, Targets, TargetStatuses during schema change.

luluz66:
> While the columns are being dropped, is a lock held on the table (can new rows be inserted?)

Tested with the new INPLACE and LOCK=NONE; new rows can be inserted.

luluz66 (commented):
Also, put the drop columns behind a flag. This is because drop columns should not be called initially. Otherwise servers with old binary will fail to write to Invocations once the column is dropped. Instead, we should wait one release to drop the column; and even longer for on-prem customers.

server/tables/tables.go
```diff
@@ -672,24 +712,47 @@ func PostAutoMigrate(db *gorm.DB) error {
...
-	}{
+	}
+
+	colsToDelete := []ColRef{
 		// Group.api_key has been migrated to a single row in the APIKeys table.
 		{&Group{}, "api_key"},
-	} {
+	}
+
+	if *dropInvocationPKCol {
```
vadimberezniker, line 725: Can you add a comment here to document why this is hidden behind a flag?

vadimberezniker (approved):

server/tables/tables.go
luluz66: Done

PR #1434, opened 2022-01-18
Make firecracker_test work with `bazel test`

bduffany:
* Disable sandboxing
* Fix error caused by HOME dir not being defined

---

**Version bump**: Patch


tylerwilliams (approved):
nice find

PR #1435, opened 2022-01-18
Update google go dependencies

tempoz:
This may fix our problem where traces include blob names and clutter the namespace; otherwise,
we'll silence the opencensus tracing integration for now.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

tylerwilliams:
With updates that update core libs like https://github.com/grpc/grpc-go/releases, it's probably a good idea to summarize the big changes in the CL description, or at least acknowledge you've looked over at them. (I looked at these and they seem pretty safe)

PR #1436, opened 2022-01-18
Silence opencensus traces since GCS traces are too noisy

tempoz:
The tracing internal to the GCS client creates traces with the blob name in the span name,
which clutters the span namespace unnecessarily in Jaeger, so disable it.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1437, opened 2022-01-18
Firecracker: optionally init dockerd before starting exec server

bduffany:
If the platform prop `init-dockerd` is set to `true`, add a flag to `goinit` that runs `dockerd` before starting the exec server.

Verified using a few different distros (Ubuntu 20.04, Ubuntu 16.04, Fedora) that running `dockerd` without any args is enough to get `dockerd` running. This is great because it means we don't have to rely on any init scripts which would be specific to systemd or sysvinit. The `dockerd` command even auto-starts `containerd` (one of its deps) if needed.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

bduffany:
FYI, built `initrd.cpio` @ commit 40eabcfe5e4735112f2b33d2db7cb7ad3d7b402c

PR #1438, opened 2022-01-19
Allow mixing of self-hosted and cloud-hosted executors

siggisim:
This PR enables users to choose whether or not to use self-hosted executors at the action level.

It does so by introducing a new platform property called `use-self-hosted-executors`.

For folks that only want to use self-hosted executors, there is still an organization-level switch that defaults all invocations to self-hosted executors, and disallows the use of cloud-hosted executors.

The wording of this switch updated slightly to `Use self-hosted executors ` => `Default to self-hosted executors` but the functionality effectively remains the same for existing users.

Also updated the "Enabled" language for executors to "Default" to match this behavior.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/executors/executor_card.tsx
+++ b/enterprise/app/executors/executor_card.tsx
@@ -10,10 +10,10 @@ interface Props {
 export default class ExecutorCardComponent extends React.Component<Props> {
   render() {
     const node = this.props.executor.node;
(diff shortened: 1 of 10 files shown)


tylerwilliams (approved):
LGTM

Could you add a test for this?

siggisim:
> Could you add a test for this?
Done

PR #1440, opened 2022-01-19
Fix error returned by uuid.GetHostID when running some tests locally

bduffany:
For some reason, HOME and XDG_CONFIG_HOME don't seem to be defined when running `bazel test //...` locally, but they _are_ defined when running remotely, at least on my machine. So `os.UserConfigDir` returns an error. Fixed by detecting TEST_TMPDIR and using that instead.

---

**Version bump**: Patch

### DIFF
--- a/server/testutil/testenv/testenv.go
+++ b/server/testutil/testenv/testenv.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"io/ioutil"
 	"net"
-	"os"
 	"path/filepath"
 	"testing"
 	"text/template"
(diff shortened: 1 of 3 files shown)


server/testutil/testenv/testenv.go
```diff
@@ -124,29 +122,11 @@ func writeTmpConfigFile(testRootDir string) (string, error) {
 	return configPath, nil
 }
 
-func ensureHomeDir(t testing.TB) {
```
bduffany, line 127: @tempoz FYI I had added this code in a previous PR because I mistakenly thought a test configuration change I had made was causing this error with HOME / XDG_CONFIG_HOME not being defined. Turns out the error was something else, so this can be reverted.

tempoz (approved):
LGTM

PR #1443, opened 2022-01-19
Don't ignore workspace creation errors

bduffany:
---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

(is there a linter we can enable to catch these?)

vadimberezniker (approved):
Yikes.

bduffany:
This was highlighted in my editor:

```
this value of err is never used (SA4006) go-staticcheck
```

but I am not sure if this triggers a build failure. @tempoz any ideas?

tempoz:
@bduffany the compiler is permissive about re-assigning values as long as they're "used" in the block anywhere, regardless of if that particular assignment is ever used/checked.

PR #1444, opened 2022-01-19
Have Hosted Bazel runs be displayed more naturally in the UI.

vadimberezniker:
The runner will report child invocation information via the new
`ChildInvocationsConfigured` and `ChildInvocationCompleted` events that
can be used for both workflows & hosted bazel invocations.

Invocation history:
invocation_history

Invocation details:
invocation_details

---

**Version bump**: None

### DIFF
--- a/app/invocation/child_invocation_card.tsx
+++ b/app/invocation/child_invocation_card.tsx
@@ -1,9 +1,9 @@
 import React from "react";
 import format from "../format/format";
 import router from "../router/router";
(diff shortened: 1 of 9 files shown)


bduffany (approved):
LGTM -- we might want to consider showing a (Hosted) pill indicator, similar to what we do for workflows, and allow filtering the UI to just hosted bazel runs as well -- but that can be a separate PR

PR #1448, opened 2022-01-20
Remove terminal subtitles

siggisim:
They're not really compatible with the way we render the terminal search box.

On small screens they would wrap and overlap with terminal content.

This PR removes/combines useless subtitles and breaks out useful ones into their own card.

---

**Version bump**: Patch

### DIFF
--- a/app/root/root.css
+++ b/app/root/root.css
@@ -2151,6 +2151,11 @@ code .comment {
   background: #263238;
 }
 
-.loading.loading-dark:before {
+.loading.loading-dark-terminal {
+  background: #222;
+}
+
+.loading.loading-dark:before,
+.loading.loading-dark-terminal:before {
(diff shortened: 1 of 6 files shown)


app/root/root.css
```diff
@@ -1066,6 +1060,7 @@ code .comment {
   background-color: transparent;
   cursor: pointer;
   padding: 12px;
+  margin: -12px 0;
```
bduffany, line 1063: nit: is there a way to do this without using negative margins?

bduffany (approved):

app/root/root.css
siggisim: Its margin to offset the padding, while keeping the click targets large. We can reduce click target size - but I'd rather keep them big for the occasional mobile user.

PR #1445, opened 2022-01-19
Respect executor host root dir flag when not running on k8s

bduffany:
The fix for https://github.com/buildbuddy-io/buildbuddy-internal/issues/1076 involves two parts:

1. Update our recommended command to the following:

```diff
 docker run \
+    --volume /tmp/buildbuddy:/buildbuddy \
     --mount type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock \
     gcr.io/flame-public/buildbuddy-executor-enterprise:latest \
+    --executor.host_executor_root_directory=/tmp/buildbuddy \
     --executor.docker_socket=/var/run/docker.sock \
     --executor.app_target=grpcs://remote.buildbuddy.dev \
     --executor.api_key=XXXX
```

2. (This PR) Respect the `host_executor_root_directory` flag in environments other than k8s. The current behavior of only supporting this flag on k8s is most likely arbitrary (reviewers can correct me if I am wrong :sweat_smile: )

---

**Version bump**: Patch

PR #1449, opened 2022-01-20
Fix docker command for running the executor

bduffany:
Also rename `executor.host_executor_root_directory` to `executor.host_root_directory` while we still can (it's unlikely that anyone is using this flag yet, since we haven't documented it anywhere, and it previously wasn't working as intended -- see https://github.com/buildbuddy-io/buildbuddy/pull/1445)

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/executors/executors.tsx
+++ b/enterprise/app/executors/executors.tsx
@@ -60,9 +60,12 @@ class ExecutorDeploy extends React.Component<ExecutorDeployProps, ExecutorDeploy
         </Select>
         <code>
           <pre>
-            {`docker run --mount type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock \\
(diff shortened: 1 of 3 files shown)


vadimberezniker (approved):
for some context: this flag was introduced during our first attempt to move docker data onto local SSD where ssd was mounted on the gke node at /mnt/ssd0. Then we later moved back to using emptyDIr + ephemeral ssd storage which made the flag no longer relevant for our use case.

bduffany:
> for some context: this flag was introduced during our first attempt to move docker data onto local SSD where ssd was mounted on the gke node at /mnt/ssd0. Then we later moved back to using emptyDIr + ephemeral ssd storage which made the flag no longer relevant for our use case.

makes sense, thanks :smiley:

PR #1455, opened 2022-01-21
Website: Prevent navbar from covering jump link targets

bduffany:
See https://css-tricks.com/fixed-headers-and-jump-links-the-solution-is-scroll-margin-top/ for background

---

**Version bump**: Patch


siggisim (approved):
Thanks for doing this! I noticed this recently and remembered there was some way to fix it but couldn't remember the details

PR #1457, opened 2022-01-21
Add config flag for test grid v2

bduffany:
Define a server flag that decides whether to enable the new V2 test grid query and pipe it up to the client.

---

**Version bump**: Patch

### DIFF
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -53,4 +53,7 @@ message FrontendConfig {
 
   // Whether Darwin (macOS) executors must be self-hosted.
   bool force_user_owned_darwin_executors = 17;
+
+  // Whether test grid V2 is enabled.
+  bool test_grid_v2_enabled = 18;
 }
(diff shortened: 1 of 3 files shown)


siggisim:
Is this one needed (in the config)? You might still hit a backend that doesn't support v2 during rollout right? Can we just keep it isolated to the request/response flag?

bduffany:
> Is this one needed (in the config)? You might still hit a backend that doesn't support v2 during rollout right? Can we just keep it isolated to the request/response flag?

Sure, we can set the request value via a client-only flag that we control via `localStorage` during testing, and then remove the flag in the release following the server rollout (along with the old server code).

siggisim:
Ah I see, if it's just a UI flag I'm cool with it since those are nicer than localStorage. 

I forgot we had to change the UI before getting a response from the server (i.e. to show the list of repos).

siggisim (approved):

PR #1461, opened 2022-01-24
TestGrid:  Proto Change for GetTargets

luluz66:
-- Add pagination token to GetTarget Request/Response
-- Add a version field to GetTarget to indicate which sql query to run.

### DIFF
--- a/proto/target.proto
+++ b/proto/target.proto
@@ -96,7 +96,7 @@ message GetTargetRequest {
   context.RequestContext request_context = 1;
 
   // The filters to apply to this query. Required.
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
LGTM -- let's make sure to remove version from FE/backend after this is released though?

bduffany (approved):
LGTM

luluz66:
> LGTM -- let's make sure to remove version from FE/backend after this is released though?

Sure..btw i changed from an int to bool to make it more readable.

PR #1462, opened 2022-01-24
Firecracker: Update default RBE image to support docker-in-firecracker actions

bduffany:
Verified that our firecracker `goinit` is able to run `dockerd` and that `docker_test` passes using this image + docker-in-firecracker.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1466, opened 2022-01-25
Docs sidebar tweaks

siggisim:
Makes navigating docs easier and require much less clicking by expanding sidebar items by default.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1468, opened 2022-01-25
Add "How Bazel 5.0 Makes Your Builds Faster" blog post

brentleyjones:
---

**Version bump**: None


website/blog/how-bazel-5-0-makes-your-builds-faster.md
```diff
@@ -0,0 +1,133 @@
...
+That said,
+if network bandwidth is a bottleneck in your build times,
+this flag could make those builds significantly faster.
+
+[zstd]: https://en.wikipedia.org/wiki/Zstandard
+
+[^2]:
+    For now.
+    Eventually we might store artifacts in their compressed state,
+    removing all overhead on the cache side.
```
bduffany, line 127: For more context, there will still be some overhead on the cache side because we'll need decompressive transcoding on uploads in order to verify their checksum. Also, for RBE inputs/outputs we'd need both compressive and decompressive transcoding either on the cache or on the executor (most likely, it makes sense to do it on the cache, since CPU resources are more under contention on the executor). So for folks using RBE, it's a bit unclear at the moment whether storing artifacts in their compressed state would be a net benefit.

brentleyjones: :+1:

I'll tweak the wording a bit.

siggisim (approved):

PR #1470, opened 2022-01-25
Clean apt cache after installing deps in default RBE image

bduffany:
According to https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

* It's always recommended to run `rm -rf /var/lib/apt/lists/*` (shaves off about 100MB)
* The official Ubuntu/Debian images should run `apt-get clean` automatically, but I found that there are things in the apt cache that could be cleaned (shaving off around another 100 MB). The explicit `apt-get clean` might be needed because we are on an older ubuntu version (16.04) or because Google's RBE image doesn't do this automatically for some reason.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1472, opened 2022-01-25
Bump default execution image version

bduffany:
The new image version has `docker-ce` installed. Compressed size is about 90MB increased from the previous image (8.6%)

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1482, opened 2022-01-26
Implement server-side pagination for Test Grid

luluz66:
The SQL query can be found [here ](https://docs.google.com/document/d/16fZgzVeaxLV74gvzLqiIWaOjt6fNAy4A33Ku3Yt0M38/edit)


server/target/pagination.go
```diff
@@ -0,0 +1,50 @@
+package target
```
bduffany, line 1: nit: our current convention is to have one Go file per package, so that if you see an identifier without a package prefix, you can tell that it is located in the file you're currently looking at

server/target/target.go
```diff
@@ -55,74 +58,29 @@ func GetTarget(ctx context.Context, env environment.Env, req *trpb.GetTargetRequ
...
-	// Adds user / permissions to targets (t) table.
-	if err := perms.AddPermissionsCheckToQueryWithTableAlias(ctx, env, q, "t"); err != nil {
-		return nil, err
-	}
-	// Adds user / permissions to invocations (i) table.
-	if err := perms.AddPermissionsCheckToQueryWithTableAlias(ctx, env, q, "i"); err != nil {
-		return nil, err
+	if req.GetServerSidePagination() {
+		return readPaginatedTargets(ctx, env, req)
+	} else {
```
bduffany, line 64: nit: remove `else`

server/target/target.go
```diff
@@ -184,8 +151,129 @@ func readTargets(ctx context.Context, env environment.Env, req *trpb.GetTargetRe
...
+		SELECT t.target_id, t.label, t.rule_type, ts.target_type, ts.test_size, ts.status,
+		ts.start_time_usec, ts.duration_usec, i.invocation_id, i.commit_sha, i.branch_name,
+		i.repo_url, i.created_at_usec
+		FROM Targets as t
+		JOIN TargetStatuses as ts ON ts.target_id = t.target_id`)
+	q.AddJoinClause(joinQuery, "i", "ts.invocation_uuid = i.invocation_uuid")
+	return fetchTargetsFromDB(ctx, env, q, repo)
+}
+
+func readTargets(ctx context.Context, env environment.Env, req *trpb.GetTargetRequest, startUsec, endUsec int64) (*trpb.GetTargetResponse, error) {
```
bduffany, line 232: Can you move this above `fetchTargetsFromDB`? I think that will reduce the diff and make the PR a bit easier to review

luluz66: Done

server/target/target.go
luluz66: Done

server/target/pagination.go
luluz66: Moved to target.go

server/tables/tables.go
```diff
@@ -505,6 +506,7 @@ type PostAutoMigrateLogic func() error
 // but some info is needed before the migration takes place in order to know what
 // to do.
 func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
+	log.Info("luluz-debug: PreAutoMigrate")
```
bduffany, line 509: nit: rm debug logs here and below

server/target/target.go
```diff
@@ -118,11 +163,16 @@ func readTargets(ctx context.Context, env environment.Env, req *trpb.GetTargetRe
...
+	return fetchTargetsFromDB(ctx, env, q, tq.GetRepoUrl())
+}
+
+func fetchTargetsFromDB(ctx context.Context, env environment.Env, q *query_builder.Query, repoURL string) (*trpb.GetTargetResponse, error) {
 	queryStr, args := q.Build()
 
 	seenTargets := make(map[string]struct{}, 0)
 	targets := make([]*trpb.Target, 0)
 	statuses := make(map[string][]*trpb.TargetStatus, 0)
+	var paginationToken *PaginationToken
```
bduffany, line 175: nit: consider calling it `nextPageToken` (I got confused thinking this was being used to paginate the query made within this func)

bduffany (changes requested):

server/target/target.go
```diff
@@ -46,6 +53,44 @@ func convertToCommonStatus(in build_event_stream.TestStatus) cmpb.Status {
...
+	if err := gob.NewDecoder(decoder).Decode(t); err != nil {
+		return nil, status.InvalidArgumentErrorf("cannot decode page token %q, %s", strToken, err)
+	}
+	return t, nil
+}
+
+func (t *PaginationToken) Encode() (string, error) {
+	var buf bytes.Buffer
+	encoder := base64.NewEncoder(base64.StdEncoding, &buf)
+	if err := gob.NewEncoder(encoder).Encode(t); err != nil {
```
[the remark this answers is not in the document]
luluz66: The reason I use a go struct instead of proto is because this is server-only, and I want to make obscure from client.

server/target/target.go
```diff
@@ -46,6 +53,44 @@ func convertToCommonStatus(in build_event_stream.TestStatus) cmpb.Status {
 	}
 }
 
+// PaginationToken is a struct to represent pagination for GetTarget RPC
+type PaginationToken struct {
+	InvocationEndUsec int64
+	CommitSHA         string
```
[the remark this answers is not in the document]
luluz66: Added comments

server/target/target.go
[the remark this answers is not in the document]
tylerwilliams: I agree with @bduffany here -- I think we should probably make this a gzipped proto bytes string instead. (We don't use gob anywhere and I like not having to worry about the backwards compatibility of it).

While I agree on making this opaque to the client, we own the client so I don't think we have to make it quite so hidden :)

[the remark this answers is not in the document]
tylerwilliams: s/gzipped/base64d/g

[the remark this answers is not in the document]
luluz66: Done

server/target/target.go
```diff
@@ -12,13 +13,18 @@ import (
...
 	cmpb "github.com/buildbuddy-io/buildbuddy/proto/api/v1/common"
 	trpb "github.com/buildbuddy-io/buildbuddy/proto/target"
+	tppb "github.com/buildbuddy-io/buildbuddy/proto/target_pagination"
+	gitutil "github.com/buildbuddy-io/buildbuddy/server/util/git"
 )
 
 const (
 	sqlite3Dialect = "sqlite3"
+	ciRole         = "CI"
+	targetPageSize = 20
```
tylerwilliams, line 27: nit: can you comment what this field is? (assume some doofus like me comes along and needs to change it, do i have enough info in the comment to do that)

luluz66: Done

proto/target_pagination.proto
```diff
@@ -0,0 +1,16 @@
...
+package target_pagination;
+
+// The pagination token used to render Test Grid.
+// This proto is intended to use by backend only.
+// For the first page, we fetch N distinct most recent commits ordered by the
+// latest created_at_usec for the invocation.
+// For the subsequent pages, we fetch commits older than the provided
+// invocation_end_time_usec; or when created_at_usec = invocation_end_time_usec,
+// we fetch commits that has a commit_sha alphabetically after the provided
+// commit_sha.
```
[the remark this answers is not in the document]
luluz66: Done

bduffany (approved):

PR #1488, opened 2022-01-27
Firecracker: Surface fatal errors from `goinit` when VM exec connection fails

bduffany:
Before, if the VM crashes while we are attempting to connect to the VM exec server, we'd just return `Unknown: context deadline exceeded` to Bazel.

After this PR, we return the fatal error logged by `goinit`: `Internal error: Firecracker VM crashed: write /mnt/vmvfs: copy_file_range: no space left on device`. In this example, the error can be fixed by giving more memory to the Firecracker VM. This error is still not super helpful to users, but is very helpful for debugging issues.

TODO in future PRs:
- Prevent these "no space left on device" errors from happening in the first place by always making sure we have enough memory for goinit
- Surface fatal errors whenever we fail to communicate with the VM, not just when connecting to the VM ExecServer. (I'm not sure what the failure modes look like in those cases though, so we might want to deal with these as they come up.)

---

**Version bump**: Patch


vadimberezniker (approved):

deps.bzl
```diff
@@ -2559,9 +2559,8 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         name = "com_github_lni_dragonboat_v3",
         build_file_proto_mode = "disable_global",
         importpath = "github.com/lni/dragonboat/v3",
-        replace = "github.com/tylerwilliams/dragonboat/v3",
-        sum = "h1:rgT2w+T0OLmZoRSez1eiAwVgZ0ZIBE0zyzry++mUNbM=",
-        version = "v3.3.4-rc2",
+        sum = "h1:hh/XOeTC+fN1xJ5GsidYwKt2oeCe/V9WUjoB9dEKoJY=",
+        version = "v3.3.4",
```
tylerwilliams, line 2563: I think this will probably break my raft stuff. Do you know what made this change / how we can alias this package to my fork of it correctly?

enterprise/server/cmd/goinit/main.go
```diff
@@ -50,6 +50,8 @@ var (
 // die logs the provided error if it is not nil and then terminates the program.
 func die(err error) {
 	if err != nil {
+		// NOTE: do not change this "die: " prefix. We rely on it to parse the fatal
```
tylerwilliams, line 53: lol die die die

https://www.youtube.com/watch?v=yV9f4w9SXEc

tylerwilliams (changes requested):
LGTM (but see my deps.bzl comment)

deps.bzl
bduffany: https://github.com/buildbuddy-io/buildbuddy/pull/1491 will fix the raft issue -- need a `replace` directive in `go.mod`. Gazelle respects the replace directive and will generate the correct `deps.bzl` rule with the `replace` attribute.

tylerwilliams: Ack, thanks!

tylerwilliams (approved):
LGTM

PR #1489, opened 2022-01-27
TargetStatuses PK Migration Take 2

luluz66:
1. Check if primary keys exist before alter table drop primary keys
2. When the primary keys don't exist skip dropping primary keys
3. delete from target statuses where invocation uuid is null

Tested locally with mysql 5.7

Scenario 1:  primary key doesn't exist

1. create the db from v2.8.3 (last week's release)
2. drop primary key
3. set all invocation_uuid to null on TargetStatuses (to trigger invocation uuid migration)
5. drop a row in table Invocations (to trigger the delete statement)
6. run the auto migration

Scenario 2: primary key exist
Similar to Scenario 1; skipping step 2.


luluz66:
PTAL - I changed two things in the last commit: 
1. check whether invocation_uuid is a primary key before invoking migration. The old condition checks the index which might not work if the customer skips releases as the index was added recently.
2. Also add a version check. Only use INPLACE Alogrithm when the version is 5.7 and above. My local test with MySQL 5.6 shows that "Error 1846: ALGORITHM=INPLACE is not supported. Reason: cannot silently convert NULL values, as required in this SQL_MODE. Try ALGORITHM=COPY."

server/tables/tables.go
```diff
@@ -672,24 +763,102 @@ func PostAutoMigrate(db *gorm.DB) error {
...
+
 	return nil
 }
 
+func dropColumnInPlaceForMySQL(db *gorm.DB, table Table, column string) error {
+	return db.Exec("ALTER TABLE " + table.TableName() + " DROP COLUMN " + column + ", ALGORITHM=INPLACE, LOCK=NONE").Error
+}
+
+func hasPrimaryKey(db *gorm.DB, table Table, key string) (bool, error) {
+	count := 0
```
vadimberezniker, line 813: nit: move count variable closer to where it's first used below

server/tables/tables.go
```diff
@@ -672,24 +763,102 @@ func PostAutoMigrate(db *gorm.DB) error {
...
+	checkPrimaryKeyStmt := ""
+	switch dialect := db.Dialector.Name(); dialect {
+	case sqliteDialect:
+		checkPrimaryKeyStmt = `SELECT COUNT(*) FROM PRAGMA_TABLE_INFO(?) WHERE pk > 0 AND name = ?`
+	case mysqlDialect:
+		checkPrimaryKeyStmt = `SELECT COUNT(0) from INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = SCHEMA() AND column_key = 'PRI' and table_name=? and column_name=?`
```
vadimberezniker, line 819: nit: use count(*) here for consistency?

vadimberezniker (approved):

server/tables/tables.go
```diff
@@ -672,24 +763,102 @@ func PostAutoMigrate(db *gorm.DB) error {
...
+		if err != nil {
+			log.Warningf("Failed to drop column %s.%s: %s", ref.table.TableName(), ref.column, err)
 		}
 	}
+
 	return nil
 }
 
+func dropColumnInPlaceForMySQL(db *gorm.DB, table Table, column string) error {
+	return db.Exec("ALTER TABLE " + table.TableName() + " DROP COLUMN " + column + ", ALGORITHM=INPLACE, LOCK=NONE").Error
```
tylerwilliams, line 809: is the "ALGORITHM=INPLACE" safe here or does it need a version guard as well?

tylerwilliams (approved):
LGTM (w/ 1 comment, plus Vadim's comments)

PR #1491, opened 2022-01-27
Add checkstyle check and fix script for go deps

bduffany:
This PR adds a check to enforce the following:
* `go mod tidy -e` would produce no diffs to `go.mod` or `go.sum`
* `go.mod` has exactly 2 `require()` blocks, one for direct imports and one for indirect imports. See https://github.com/golang/go/issues/45965
* Updating `deps.bzl` via Gazelle would produce no diffs

Provides a script to automatically fix most errors detected, or provide instructions on how to manually fix (`tools/fix_go_deps.sh`)

Note: the `-e` flag to `go mod tidy` lets us avoid the need to add the generated protos to the repo before running `go mod tidy`. This flag is not supported in Go 1.15 (the version currently used in GH actions), so this PR also adds a step to upgrade the Go version to v1.17.

Example failing run (before running `tools/fix_go_deps.sh` in this PR): https://github.com/buildbuddy-io/buildbuddy/runs/4960976363?check_suite_focus=true -- the diff is huge here because `go.mod` is particularly whacky these days, but the diff should be much smaller in the future.

---

**Version bump**: Patch


vadimberezniker (approved):
nice :)

.github/workflows/checkstyle.yaml
```diff
@@ -17,6 +17,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v1
 
+      - name: Setup go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.17.2"
```
tylerwilliams, line 23: do you want to leave a comment in WORKSPACE that points to this file and mentions these versions should be kept in sync?

tools/fix_go_deps.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
tylerwilliams, line 1: can you add an option to ./buildfix.sh that does this so we continue to have one script that will "ready" your build for review?

.github/workflows/checkstyle.yaml
bduffany: Done

tools/fix_go_deps.sh
bduffany: Done, added a `--fix-go-deps` flag for this. Also added an `--all` flag that is the same as `--gazelle --fix-go-deps`.

bduffany:
FYI, verified that this works on Mac.

tempoz (commented):
This looks great! Thanks so much for doing this; when `go.mod` got out-of-date, `gopls` would give up and my IDE functionality would stop working until I went and ran go mod tidy again, so this is super helpful.

go.mod
```diff
@@ -5,26 +5,27 @@ go 1.17
 replace (
 	github.com/firecracker-microvm/firecracker-go-sdk => github.com/tylerwilliams/firecracker-go-sdk v0.22.1
 	github.com/go-redsync/redsync/v4 v4.4.1 => github.com/bduffany/redsync/v4 v4.4.1-minimal
+	github.com/lni/dragonboat/v3 => github.com/tylerwilliams/dragonboat/v3 v3.3.4-rc2
```
tylerwilliams, line 8: thankyou!

tylerwilliams (approved):
LGTM

PR #1492, opened 2022-01-27
Fix incremental Firecracker snapshots.

vadimberezniker:
Use kernel API to correctly find dirty data blocks in the incremental snapshot.

Updates firecracker & jailer binaries to latest release version,
required for enabling diff snapshots.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1493, opened 2022-01-27
Add workaround for Bazel committed_size check when uploading compressed blobs

bduffany:
Context: https://github.com/bazelbuild/bazel/issues/14654

---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/byte_stream_server/byte_stream_server.go
+++ b/server/remote_cache/byte_stream_server/byte_stream_server.go
@@ -180,7 +180,6 @@ type writeState struct {
 	d                  *repb.Digest
 	activeResourceName string
 	offset             int64
-	alreadyExists      bool
 }
 
 func checkInitialPreconditions(req *bspb.WriteRequest) error {
@@ -239,7 +238,9 @@ func (s *ByteStreamServer) initStreamState(ctx context.Context, req *bspb.WriteR
 	if err != nil {
 		return nil, err
 	}
-	ws.alreadyExists = exists
+	if exists {
+		return nil, status.AlreadyExistsError("Already exists")
+	}
 
 	ws.checksum = NewChecksum()
 	ws.writer = ws.checksum
@@ -322,16 +323,12 @@ func (s *ByteStreamServer) Write(stream bspb.ByteStream_WriteServer) error {
 
 			// If the API key is read-only, pretend the object already exists.
 			if !canWrite {
-				r, err := digest.ParseUploadResourceName(req.ResourceName)
-				if err != nil {
-					return err
-				}
-				return stream.SendAndClose(&bspb.WriteResponse{
-					CommittedSize: r.GetDigest().GetSizeBytes(),
-				})
+				return handleAlreadyExists(stream, req)
 			}
-
 			streamState, err = s.initStreamState(ctx, req)
+			if status.IsAlreadyExistsError(err) {
+				return handleAlreadyExists(stream, req)
+			}
 			if err != nil {
 				return err
 			}
@@ -340,11 +337,6 @@ func (s *ByteStreamServer) Write(stream bspb.ByteStream_WriteServer) error {
 					log.Error(err.Error())
 				}
 			}()
-			if streamState.alreadyExists {
-				return stream.SendAndClose(&bspb.WriteResponse{
-					CommittedSize: streamState.d.GetSizeBytes(),
-				})
-			}
 			ht := hit_tracker.NewHitTracker(ctx, s.env, false)
 			uploadTracker := ht.TrackUpload(streamState.d)
 			defer uploadTracker.Close()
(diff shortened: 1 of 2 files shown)


server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -322,16 +323,12 @@ func (s *ByteStreamServer) Write(stream bspb.ByteStream_WriteServer) error {
...
-				return stream.SendAndClose(&bspb.WriteResponse{
-					CommittedSize: r.GetDigest().GetSizeBytes(),
-				})
+				return handleAlreadyExists(stream, req)
 			}
-
 			streamState, err = s.initStreamState(ctx, req)
+			if status.IsAlreadyExistsError(err) {
+				return handleAlreadyExists(stream, req)
+			}
```
tylerwilliams, line 331: does this jive well  with the uploadTracker stuff below or does this mean we're missing data?

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -402,6 +394,59 @@ func (s *ByteStreamServer) QueryWriteStatus(ctx context.Context, req *bspb.Query
...
+			if err != nil {
+				return err
+			}
+		}
+		committedSize = int64(len(firstRequest.Data)) + remainingSize
+	}
+	return stream.SendAndClose(&bspb.WriteResponse{CommittedSize: committedSize})
+}
+
+func remainingUploadSize(stream bspb.ByteStream_WriteServer) (int64, error) {
```
tylerwilliams, line 436: nit: comment what this function does?

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -402,6 +394,59 @@ func (s *ByteStreamServer) QueryWriteStatus(ctx context.Context, req *bspb.Query
...
+	// check whether the committed_size matches the number of bytes it has
+	// uploaded, but only if it has uploaded all bytes. But because Bazel doesn't
+	// wait for an ack before it uploads each chunk, we (generally) have no way of
+	// knowing whether Bazel has uploaded all bytes at this point. So we are
+	// forced to wait until bazel uploads all chunks, so that we know for sure
+	// that the committed_size we send back will match Bazel's expectation.
+	//
+	// There are two cases where we can short-circuit, though:
+	//
+	// - If this is an uncompressed stream, we assume that the committed_size will
```
tylerwilliams, line 415: probably worth comment what's happening in the biggest part of the function below: if this is a compressed stream?

server/remote_cache/byte_stream_server/byte_stream_server.go
bduffany: For compressed uploads requiring more than 1 message to upload, this does mean that now we are potentially letting the client upload a significant amount of data that is not tracked as an upload (cache write).

Fixed by creating a separate UploadTracker in handleAlreadyExists, but only in the case where we have to download the remaining stream.

Note that it's still tracking uncompressed size, but we have an existing issue to track the raw number of bytes sent by the client as opposed to uncompressed bytes.

Side note, I think we can do better in the case where the client sends only a single message that contains the entire payload, but then we short-circuit because the object already exists in cache. Currently (and also before this PR) we don't track those payloads at all, so we might present an inaccurate view of upload rate because it excludes these payloads from the total uploaded bytes. We also probably need to decide whether to present those as "cache writes" to the user. However, it's probably not worth worrying about too much, since those short-circuiting cases should be relatively rare.

server/remote_cache/byte_stream_server/byte_stream_server.go
bduffany: Reworded and shortened -- hopefully it's clearer now.

server/remote_cache/byte_stream_server/byte_stream_server.go
bduffany: Done, and renamed to make it more clear that it's doing IO

server/remote_cache/byte_stream_server/byte_stream_server.go
brentleyjones: > We also probably need to decide whether to present those as "cache writes" to the user

If the user uploaded, even if it's a duplicate, I feel we should show it. It's confusing to see the number lower than what they could see with the gRPC log. Similar to when it's a read-only key, I feel we should show the number of items uploaded. Maybe we split those in two in the UI though? Uploaded versus written to cache?

brentleyjones: Also, I think showing a warning when they upload but it's not written because of a read-only key can remove a lot of confusion as well.

tylerwilliams (approved):
LGTM

nice job coming up with a workaround for this on the spot :)

PR #1498, opened 2022-01-28
Move go deps fix before gazelle

bduffany:
This makes sure that `deps.bzl` is up to date before running Gazelle, which updates BUILD files that can reference the repositories in `deps.bzl`.

The checkstyle change isn't strictly needed, but this applies the same ordering there just to keep things consistent.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1512, opened 2022-02-01
Pin arm64 executor build to Xcode 12.4 as well

brentleyjones:
We already do this for the x86_64 build.

---

**Version bump**: Patch.


tylerwilliams (approved):
LGTM

(thanks for making these consistent)

PR #1513, opened 2022-02-01
Fix goroutine leak in zstd decoder pool

bduffany:
The problem is that the pool waits for decoders to be garbage collected before calling `Close`, but decoders can't be garbage collected until `Close` is called (circular logic).

Fixed by adding a lightweight reference around the decoder that can be garbage collected before the decoder can (since it doesn't start any goroutines itself), and when finalized it closes the underlying decoder.

Before (running a bazel build in a loop that uploads a bunch of compressed artifacts to cache, clearing the local cache in between runs and uploading via a new remote instance name):

image

After (same scenario as above):

image

Also, save even more memory by setting `WithDecoderConcurrency(1)` -- we don't need the default concurrency of GOMAXPROCS because we don't use these decoders concurrently (yet at least). Each decoder compresses one stream at a time (or one chunk, in the `DecodeAll` case). By setting concurrency=1 we're effectively using 8X fewer goroutines.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1131

### DIFF
--- a/server/util/compression/compression.go
+++ b/server/util/compression/compression.go
@@ -152,6 +152,12 @@ func NewZstdChunkingCompressor(reader io.Reader, readBuf []byte, compressBuf []b
 	return pr, nil
 }
 
+// DecoderRef wraps a *zstd.Decoder. Since it does not directly start any
+// goroutines, it can be garbage collected before the wrapped decoder can.
+// When garbage collected, a finalizer automatically closes the wrapped decoder,
+// thus allowing the decoder to be garbage collected as well.
+type DecoderRef struct{ *zstd.Decoder }
+
 // ZstdDecoderPool allows reusing zstd decoders to avoid excessive allocations.
 type ZstdDecoderPool struct {
 	pool sync.Pool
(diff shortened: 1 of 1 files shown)


vadimberezniker (approved):
To be honest, I don't understand why this fix works but since your testing shows that it does we should go ahead and merge it. It would be good to get a better understanding of why it works though for educational purposes.

bduffany:
> It would be good to get a better understanding of why it works though for educational purposes.

SG, when I find some time, I will try and see if I can put together a minimal example so that we don't need to completely grok the compression lib to understand the root issue.

server/util/compression/compression.go
```diff
@@ -161,12 +167,15 @@ func NewZstdDecoderPool() *ZstdDecoderPool {
...
 				if err != nil {
 					return err
 				}
-				runtime.SetFinalizer(dc, (*zstd.Decoder).Close)
-				return dc
+				ref := &DecoderRef{dc}
+				runtime.SetFinalizer(ref, func(ref *DecoderRef) {
+					ref.Decoder.Close()
+				})
+				return ref
```
tylerwilliams, line 178: nit: might be less error prone if you had a NewDecoderRef() function that did this and setup the finalizer internally?

tylerwilliams: oh i guess this only happens in NewZstdDecoderPool, ignore me :)

PR #1515, opened 2022-02-01
Move default workflow args to /etc/bazel.bazelrc

bduffany:
Makes the config story consistent between default and non-default (i.e. `buildbuddy.yaml`-configured) workflows, and makes the command line rendered in the UI for the default workflow a bit less messy.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/835

### DIFF
--- a/docs/workflows-config.md
+++ b/docs/workflows-config.md
@@ -29,19 +29,14 @@ actions:
         branches:
           - "main"
     bazel_commands:
-      - "test //... --build_metadata=ROLE=CI --bes_backend=grpcs://cloud.buildbuddy.io --bes_results_url=https://app.buildbuddy.io/invocation/"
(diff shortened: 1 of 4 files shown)


vadimberezniker (approved):

tylerwilliams (approved):
LGTM

bduffany:
FYI, changed the path from ~/.bazelrc to /etc/bazel.bazelrc, since while testing I noticed that `~/.bazelrc` takes precedence over the workspace `.bazelrc` (we want the workspace `.bazelrc` to take precedence). https://docs.bazel.build/versions/main/guide.html#where-are-the-bazelrc-files

PR #1519, opened 2022-02-02
Temporarily set --nosystem_rc to unbork dev workflows

bduffany:
This will prevent the default API key from being applied to dev workflows, which is undesired because our workflows point at BB prod (since we only have a single `buildbuddy.yaml` file) so the dev API key won't work.

We probably want a way for the "inner" invocation to be authenticated and to use env-specific endpoints instead of hard-coding prod in our buildbuddy.yaml (maybe by providing some opt-in configs via `/etc/bazel.bazelrc` and then referencing those via `--config=buildbuddy-bes-backend`, `--config=buildbuddy-cache-backend` etc).

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

If we were using firecracker for workflows and we fix the resume-after-restart bug, dev hitting dev would be fine again right?

bduffany:
> LGTM
> 
> If we were using firecracker for workflows and we fix the resume-after-restart bug, dev hitting dev would be fine again right?

Will follow-up offline because I'm not totally sure I understand, but yeah, we need to be able to reliably restart workflow tasks as well as the "outer" workflow invocation being retried, and hopefully the firecracker snapshotting stuff can help with latency when the executors are restarted, so the workflows don't have to be re-run from a completely cold runner (assuming we can store snapshots in the CAS or at least on a disk that persists across executor restarts)

PR #1521, opened 2022-02-03
Drop invocation_pk index and set default value

luluz66:
invocation_pk is a primary key field so it's not nullable. We need to
set a default value otherwise insert without specifying this value will
fail

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -610,6 +610,18 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
 			log.Warningf("Unsupported sql dialect: %q", db.Dialector.Name())
 		}
 	}
+	if m.HasTable("TargetStatuses") && m.HasIndex("TargetStatuses", "target_status_invocation_pk") {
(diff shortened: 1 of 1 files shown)


server/tables/tables.go
```diff
@@ -610,6 +610,18 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
 			log.Warningf("Unsupported sql dialect: %q", db.Dialector.Name())
 		}
 	}
+	if m.HasTable("TargetStatuses") && m.HasIndex("TargetStatuses", "target_status_invocation_pk") {
+		// Drop the invocation_pk index; and set default to invocation_pk column.
+		postMigrate = append(postMigrate, func() error {
```
tylerwilliams, line 615: do you want to condition the default column value on the index existing or treat them separately?

tylerwilliams (approved):
LGTM

server/tables/tables.go
luluz66: I think it's fine. As long as on-prem users don't drop the index by hand.

vadimberezniker (approved):

PR #1522, opened 2022-02-03
Use correct path in release-m1.yaml

brentleyjones:
This Xcode 12.4 was installed to a different path.

---

**Version bump**: None.


tylerwilliams (approved):
LGTM

PR #1527, opened 2022-02-03
Firecracker: Fix error propagation when copying outputs fails

bduffany:
---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

Is it easy to test this?

bduffany:
> Is it easy to test this?

In order to test this meaningfully, I think we'd have to write a test exercising firecracker itself, which would mean we'd need to support firecracker-in-firecracker if we want these to run on CI. I haven't tried that yet, but it's something I'd like to try at some point

PR #1528, opened 2022-02-03
Firecracker: Improve default task sizing

bduffany:
- Account for firecracker VM memory overhead in memory estimate. Experimentally, I found that firecracker needs at least 104 MB of memory to be able to run a trivial task (`pwd`). This PR sets the overhead to 150MB to give us some wiggle room. I think this might depend on the size of `initrd.cpio` + `vmlinux` since those get loaded into memory (?) so in the future we might want to calculate the sizes of those files at build time and derive the estimate from those sizes.
- Increase memory and disk estimates when using docker-in-firecracker. I found that running a `mysql:8.0` image inside a test requires at least 400MB additional memory (on top of the default estimate for a `small` test) and about 4GB additional disk space. I derived the estimate values by multiplying those by 2x and 3x, respectively, since customers may run more images / heavier images in their tests.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1094

### DIFF
--- a/enterprise/server/tasksize/tasksize.go
+++ b/enterprise/server/tasksize/tasksize.go
@@ -26,8 +26,8 @@ const (
 	// Additional resources needed depending on task characteristics
 
 	FirecrackerAdditionalMemEstimateBytes = int64(150 * 1e6) // 150 MB
-	DockerAdditionalMemEstimateBytes      = int64(400 * 1e6) // 400 MB
-	DockerAdditionalDiskEstimateBytes     = int64(3 * 1e9)   // 3 GB
+	DockerAdditionalMemEstimateBytes      = int64(800 * 1e6) // 800 MB
+	DockerAdditionalDiskEstimateBytes     = int64(12 * 1e9)  // 12 GB
 
 	MaxEstimatedFreeDisk = int64(20 * 1e9) // 20GB
 
@@ -82,7 +82,7 @@ func Estimate(task *repb.ExecutionTask) *scpb.TaskSize {
 		// Note: props.InitDockerd is only supported for docker-in-firecracker.
 		if props.InitDockerd {
 			freeDiskEstimate += DockerAdditionalDiskEstimateBytes
-			memEstimate += DockerAdditionalDiskEstimateBytes
+			memEstimate += DockerAdditionalMemEstimateBytes
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
LGTM

This seems OK to me, (a little sketch), but I'm ok with it if @vadimberezniker doesn't see any issues.

enterprise/server/tasksize/tasksize.go
```diff
@@ -23,6 +23,12 @@ const (
 	DefaultCPUEstimate      = int64(600)
 	DefaultFreeDiskEstimate = int64(100 * 1e6) // 100 MB
 
+	// Additional resources needed depending on task characteristics
+
+	FirecrackerAdditionalMemEstimateBytes = int64(150 * 1e6) // 150 MB
```
vadimberezniker, line 28: Can you document what this overhead is?

enterprise/server/tasksize/tasksize.go
```diff
@@ -23,6 +23,12 @@ const (
...
+	DockerAdditionalMemEstimateBytes      = int64(800 * 1e6) // 800 MB
```
vadimberezniker, line 29: Can you clarify (either via comment or more explicit variable name) that this is the additional memory for running Docker-in-firecracker than anything to do with the the Docker container type.

enterprise/server/tasksize/tasksize.go
```diff
@@ -23,6 +23,12 @@ const (
...
+	DockerAdditionalDiskEstimateBytes     = int64(12 * 1e9)  // 12 GB
```
vadimberezniker, line 30: We should look into making the default free space we allocate for firecracker larger. Need to check if there are any perf implications. I think we have enough local SSD on the executors that we can afford to be generous with default size as long as there isn't significant performance overhead.

vadimberezniker (approved):

enterprise/server/tasksize/tasksize.go
bduffany: Done

enterprise/server/tasksize/tasksize.go
bduffany: Done

PR #1530, opened 2022-02-03
Firecracker: Fix error when passing more than one init flag

bduffany:
It looks like as of firecracker v0.25, passing more than one `--` to bootArgs is not allowed, even if `--` is used as a prefix rather than a standalone arg. It causes the error

```
Internal: Failed starting machine: [PUT /actions][400] createSyncActionBadRequest  &{FaultMessage:Invalid kernel command line: Too many `--` in kernel cmdline.}
```

Was able to repro the error locally after upgrading from v0.24 -> v0.25, and verified that passing the args with a single `-` instead of `--` fixes the issue.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1532, opened 2022-02-04
Implement commit grouping for test grid v2

bduffany:
Note that this PR does not deal with paging, only grouping into ordered columns by commit. (Paging will be in a subsequent PR)

Screenshot (2 commits shown, ran only "//:pass" on the older of the two commits and ran only "//:fail" on the newer. Dotted square means "Target status not reported at commit"):

image

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/tap/tap.tsx
+++ b/enterprise/app/tap/tap.tsx
@@ -3,12 +3,14 @@ import Long from "long";
 import moment from "moment";
 import rpcService from "../../../app/service/rpc_service";
 import { User } from "../../../app/auth/auth_service";
-import { target } from "../../../proto/target_ts_proto";
+import { api, target } from "../../../proto/target_ts_proto";
 import { Subscription } from "rxjs";
 import router from "../../../app/router/router";
 import format from "../../../app/format/format";
 import Select, { Option } from "../../../app/components/select/select";
 import { Filter } from "lucide-react";
+import capabilities from "../../../app/capabilities/capabilities";
+import { timestampToUnixNanos, durationToNanos } from "../../../app/util/proto";
 
 interface Props {
   user: User;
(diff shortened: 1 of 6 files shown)


siggisim (approved):

enterprise/app/tap/tap.tsx
```diff
@@ -120,12 +132,49 @@ export default class TapComponent extends React.Component<Props> {
...
+        // we show in the UI is the one that ended the latest.
+        // TODO(bduffany): Keep track of per-target count by commit, in
+        // case the same target was executed multiple times for a given
+        // commit. Otherwise the count stat looks incorrect.
+        const existing = targetIdToStatus.get(history.target.id);
+        if (!existing || timestamp > endTimestampUnixNanos(existing.timing)) {
+          targetIdToStatus.set(history.target.id, targetStatus);
+        }
+      }
+    }
```
bduffany, line 170: @luluz66 FYI I'm re-ordering commits by latest `target.end_time_usec`, even though the server orders commits by `max(invocation.created_at_usec)`, since the server doesn't return the ordered commits list. LMK what you think -- I don't have a strong preference between the client or the server controlling the order of the rendered commits

siggisim: If we end up doing pagination server side, we probably want the ordering to match

bduffany: True, if the client just "concatenates" target histories and then recomputes these indexes from scratch each time, then the commits could wind up getting swapped around as new pages are fetched. It's probably better if the server returns an explicit list of commits

luluz66: I believe start_time_usec and end_time_usec of target might be wrong. They can be before invocation created_at _usec if the tests are cached. How do you feel about I pass the invocation's created_at_usec in the Target? When there are multiple invocations per commit, you also need the created_at_usec to show the last one right?

bduffany: > pass the invocation's created_at_usec in the Target

Discussed offline -- this is the option we decided to go with. Once it's implemented on the backend I'll update the ordering in this PR

PR #1533, opened 2022-02-04
Abandon old invocation attempts in the presence of new ones

tempoz:
- Update gorm to get a bugfix
- add attempt to database, add register method

The complicated gorm in the registerInvocation function will be replaced with raw SQL before
this is merged.

---

**Version bump**: Minor


tempoz:
This addresses https://github.com/buildbuddy-io/buildbuddy-internal/issues/1121, https://github.com/buildbuddy-io/buildbuddy-internal/issues/1039, and I'm pretty sure also fixes the problem causing https://github.com/buildbuddy-io/buildbuddy-internal/issues/1038.

server/eventlog/eventlog.go
```diff
@@ -32,8 +33,16 @@ const (
...
-	return invocationId + "/chunks/log/eventlog"
+var (
+	//Id of an empty log
+	EmptyId = chunkstore.ChunkIndexAsStringId(chunkstore.EmptyIndex)
+)
+
+func GetEventLogPathFromInvocationIdAndAttempt(invocationId string, attempt uint64) string {
+	if attempt == 0 {
+		return invocationId + "/chunks/log/eventlog"
+	}
```
tylerwilliams, line 44: The special-casing of the 0 functionality is to allow for this function to work with old invocations as well?

(if so, can you add a comment so we don't accidentally mess with this?)

deps.bzl
```diff
@@ -4156,20 +4156,20 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
     go_repository(
         name = "io_gorm_driver_mysql",
         importpath = "gorm.io/driver/mysql",
-        sum = "h1:TATTzt+kR+IV0+h3iUB3dHUe8omCvQ0rOkmfCsUBohk=",
-        version = "v1.0.4",
+        sum = "h1:l8+9VwjjyzEkw0PNPBOr2JHhLOGVk7XEnl5hk42bcvs=",
```
tylerwilliams, line 4159: nit: can you send the gorm upgrade CL separately? I can approve it now, and it will also reveal if that upgrade alone were to break any tests.

tempoz: Yes, will do.

server/eventlog/eventlog.go
tempoz: Yep, and will do.

server/eventlog/eventlog.go
```diff
@@ -32,8 +33,16 @@ const (
 	numReadWorkers = 16
 )
 
-func GetEventLogPathFromInvocationId(invocationId string) string {
-	return invocationId + "/chunks/log/eventlog"
+var (
+	//Id of an empty log
+	EmptyId = chunkstore.ChunkIndexAsStringId(chunkstore.EmptyIndex)
```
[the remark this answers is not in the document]
tempoz: Yeah, it gets used in `build_event_handler.go`. I think it would also be nice to use it in the other places where I was doing this conversion manually, but I didn't want to include that in this CL.

tempoz:
Switched out for simpler gorm.

server/util/db/db.go
```diff
@@ -494,6 +497,32 @@ func (h *DBHandle) InsertIgnoreModifier() string {
...
+	return "FOR UPDATE"
+}
+
 func (h *DBHandle) SetNowFunc(now func() time.Time) {
 	h.db.Config.NowFunc = now
 }
+
+func (h *DBHandle) IsDuplicateKeyError(err error) bool {
+	var mysqlErr *gomysql.MySQLError
+	if errors.As(err, &mysqlErr) && mysqlErr.Number == 1062 {
```
tylerwilliams, line 520: nit: add a comment linking to where these error codes are defined?

server/backends/invocationdb/invocationdb.go
```diff
@@ -44,36 +96,25 @@ func (d *InvocationDB) createInvocation(tx *db.DB, ctx context.Context, ti *tabl
...
-			created = true
-			return nil
-		}
-		return tx.Model(&existing).Where("invocation_id = ?", ti.InvocationID).Updates(ti).Error
+// UpdateInvocation updates an existing invocation with the given
+// id and attempt number. It returns whether a row was updated.
+func (d *InvocationDB) UpdateInvocation(ctx context.Context, ti *tables.Invocation) (bool, error) {
+	updated := false
+	err := d.h.TransactionWithOptions(ctx, db.Opts().WithQueryName("update_invocation"), func(tx *db.DB) error {
+		result := tx.Where("`attempt` = ?", ti.Attempt).Updates(ti)
```
tylerwilliams, line 113: nit: this is a little confusing because it relies on gorm magic. I think it would be clearer if it had invocation_id = ? and atttempt = ? so it was clear what was going on?

tylerwilliams (approved):
This LGTM (few nits) but I think you are right that we should let it soak rather than trying to commit it right before the release.

PR #1538, opened 2022-02-07
Fix some runner pool shutdown issues

bduffany:
* Fix "unexpected runner state" errors in both Run and Add when the pool is shutdown mid-task
* Fix NPE due to unsynchronized concurrent access to `p.runners` when the pool is shutdown mid-task

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1158

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -775,7 +775,7 @@ func (p *Pool) Get(ctx context.Context, task *repb.ExecutionTask) (*CommandRunne
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	if p.isShuttingDown {
-		return nil, status.UnavailableErrorf("Could not get new task runner since executor is shutting down.")
+		return nil, status.UnavailableErrorf("Could not get a new task runner because the executor is shutting down.")
 	}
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -750,6 +772,11 @@ func (p *Pool) Get(ctx context.Context, task *repb.ExecutionTask) (*CommandRunne
 		VFS:                fs,
 		VFSServer:          vfsServer,
 	}
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.isShuttingDown {
+		return nil, status.UnavailableErrorf("Could not get new task runner since executor is shutting down.")
```
tylerwilliams, line 778: s/since/because/g

tylerwilliams (approved):
LGTM for fix

(when we have a good opportunity we should have another look at this code. I don't think it's modeling the state transition stuff in a particularly safe way and I'm not super confident there aren't other issues we're missing here?)

enterprise/server/remote_execution/runner/runner.go
bduffany: Done

PR #1547, opened 2022-02-08
Decouple db and executor binary

tempoz:
`db.go` needs access to the sqlite3 driver library (the direct one, not the gorm one), and that dependency causes the executor to fail to build. Per @vadimberezniker's recommendation, this CL decouples the two files so that we can check sqlite errors in `db.go`.

We accomplish this by introducing interfaces for `DBHandle` and `DBOptions` primarily so that things which make use of the `DBHandle` in the environment don't have to depend, even indirectly, on `db.go`. We also move the direct database querying in `auth.go` into `authdb.go`. Because we don't want to have the interface include literally every method in `gorm.DB`, which was previously embedded in the `DBHandler`, we instead expose only a few of the calls and then expose the DB via a function call. This additionally lets us access the fields of the DB in the few places where we do that. At some point in the future I would like to remove the calls we are forwarding to the `db` field and instead use the `DB()` method to call those methods directly in the places where they are used.

---

**Version bump**: Minor


vadimberezniker (approved):
These changes LGTM.
It took me a moment to realize that the reason you need to do this is to break the environment -> db dependency. That would have been good to mention in the description.

tylerwilliams (approved):
LGTM

PR #1549, opened 2022-02-08
Set TestStatus to FAILED_TO_BUILD when success is false

luluz66:
https://github.com/buildbuddy-io/buildbuddy-internal/issues/1160


tylerwilliams (approved):
LGTM

PR #1551, opened 2022-02-09
[Code editor] Clean up state handling

siggisim:
- Store currently selected file path in url
- Always get owner, repo, and file path from URL, not state
- Save state every time state is updated, not ad-hoc

This minimizes the amount of state we have to worry about saving, and saves it more reliably - making it easier to save the important state server-side.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1556, opened 2022-02-09
Implement page-by-page fetching for test grid V2

bduffany:
- Use `response.next_page_token` to determine whether more invocations can be loaded (x-axis "load more" button) and pass the current page token along in each request
- Merge target histories from subsequent pages into the current list of target histories

TODO in subsequent PRs:
- Improve the page layout so that target names remain visible while scrolling horizontally
- I noticed that the last ~2 fetches or so fetch a very small number of commits -- need to work out whether that's a bug on the client or server. But it shouldn't block this PR, since the V1 UI should still be the same after this PR is merged.

Demo (outlined squares mean "not executed")

https://user-images.githubusercontent.com/2414826/153279082-eef99792-8d65-4970-80bc-456ceb594cda.mp4

<details>
<summary>Script that was used to generate this test dashboard</summary>

```bash
#!/bin/bash
set -e
COMMITS=100

flip_coin() {
  echo "$(($RANDOM % 2))"
}

for i in $(seq "$COMMITS"); do
  # Create a new commit
  echo "$RANDOM" >RANDOM.txt
  git add RANDOM.txt
  git commit -m "Commit #$i"

  # Write a test that only runs at this commit
  unique_test_name="commit_${i}_only"
  echo "sh_test(name = \"$unique_test_name\", srcs = [\":pass.sh\"])" >>BUILD

  # Run tests at that commit
  TARGETS=(//:pass //:fail //:flake "//:$unique_test_name")
  if [[ $(flip_coin) == 1 ]]; then
    TARGETS+=(//:maybe_execute)
  fi
  bazel test "${TARGETS[@]}" \
    --config=ci \
    --config=local-bes \
    --config=local-auth \
    --nocache_test_results || true
  git restore BUILD

  sleep 1
done
```
</details>

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/tap/tap.tsx
+++ b/enterprise/app/tap/tap.tsx
@@ -21,6 +21,7 @@ interface State {
   targetHistory: target.ITargetHistory[];
   commits: string[] | null;
   commitToTargetIdToStatus: Map<string, Map<string, target.ITargetStatus>> | null;
+  nextPageToken: string;
   loading: boolean;
   coloring: "status" | "timing";
   sort: "target" | "count" | "pass" | "avgDuration" | "maxDuration";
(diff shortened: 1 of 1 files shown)


siggisim (approved):
Love this video

PR #1558, opened 2022-02-09
Fix small alert icon

bduffany:
Before

image

After

image

---

**Version bump**: Patch

### DIFF
--- a/app/alert/alert.tsx
+++ b/app/alert/alert.tsx
@@ -11,8 +11,8 @@ interface State {
 const DISPLAY_DURATION_MS = 4000;
 
 const ICONS: Record<AlertType, JSX.Element> = {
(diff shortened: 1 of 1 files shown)


siggisim (approved):
The alignment looks a little funky, should we `align-items: center`?

bduffany:
> The alignment looks a little funky, should we `align-items: center`?

I agree, but I also think it looks funky with it aligned middle:

image

The upper left icon seems to work well for other cards because there is a title next to it:

image

Maybe we could add an "Error" title in another PR? (maybe for errors only, so that we still have the slim notification for success messages, which are usually short)

siggisim:
Yeah, you're right - let's go with this. LGTM

PR #1559, opened 2022-02-09
Cleanup: Address TODOs in test grid component

bduffany:
- Consistent x-axis TODO will be addressed by test grid V2, so remove the commented code
- Use `api.v1.Status` enum instead of raw numbers

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/tap/tap.tsx
+++ b/enterprise/app/tap/tap.tsx
@@ -3,7 +3,7 @@ import Long from "long";
 import moment from "moment";
 import rpcService from "../../../app/service/rpc_service";
 import { User } from "../../../app/auth/auth_service";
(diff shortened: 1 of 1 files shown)


enterprise/app/tap/tap.tsx
```diff
@@ -382,25 +383,6 @@ export default class TapComponent extends React.Component<Props, State> {
 
           <div className="tap-grid-container">
             <div className="tap-grid">
-              {/* 
-            // TODO(siggisim): Figure out the best way to give a consistent x-axis
```
siggisim, line 386: Still think we should render an x-axis

siggisim (approved):

PR #1564, opened 2022-02-10
Don't bundle the event log with the invocation

tempoz:
The event logs have been stored and served in chunks for some time now, so we
can safely shift away from bundling them with the InvocationResponse when
serving. This will cut down on load times for invocations with massive logs.

This change also strips the console output from the build event protos that have
chunked logs not only when we serve them, but also when they get stored to disk,
so that we're not storing the same information twice. This change remains
conditional on the `StorageEnableChunkedEventLogs` flag, but this does mean that
invocations that are streamed to a server when that flag is enabled will not
display logs to the UI if the server is later restarted with the flag disabled.

---

**Version bump**: Major


siggisim:
We still use this consoleBuffer in a few places on the front-end: https://cs.github.com/buildbuddy-io/buildbuddy?q=consolebuffer+path%3Aapp

siggisim:
Also ubernit regarding the PR title/description: I'd call this the "build log" or the console output, not the "event log" - when I think of the event log, I think of the Raw build events tab.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -761,9 +764,8 @@ func (e *EventChannel) processSingleEvent(event *inpb.InvocationEvent, iid strin
 		if e.logWriter != nil {
 			e.logWriter.Write(e.ctx, []byte(p.Progress.Stderr))
 			e.logWriter.Write(e.ctx, []byte(p.Progress.Stdout))
-			// For now, write logs to both chunks and the invocation proto
-			// p.Progress.Stderr = ""
-			// p.Progress.Stdout = ""
+			p.Progress.Stderr = ""
+			p.Progress.Stdout = ""
```
bduffany, line 768: nit: It looks like we do this in event_parser.go as well -- would it be fine to remove it from here?

bduffany:
UI no longer reads `consoleBuffer` for invocations that support chunking: https://cs.github.com/?scope=org%3Abuildbuddy-io&scopeName=buildbuddy-io&q=consoleBuffer+path%3A**%2F*.ts*

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: No, we need both: this one prevents us from writing the logs into the event proto stream, while the other call stops us from transmitting a console buffer for older invocations that have both a chunked log and a console buffer.

bduffany: sg -- mind adding a comment as to why these are here?

server/build_event_protocol/event_parser/event_parser.go
```diff
@@ -206,8 +214,10 @@ func (sep *StreamingEventParser) FillInvocation(invocation *inpb.Invocation) {
 		buildDuration = time.Duration((sep.endTimeMillis - sep.startTimeMillis) * int64(time.Millisecond))
 	}
 	invocation.DurationUsec = buildDuration.Microseconds()
-	// TODO(siggisim): Do this rendering once on write, rather than on every read.
-	invocation.ConsoleBuffer = string(sep.screenWriter.RenderAsANSI())
+	if sep.screenWriter != nil {
+		// TODO(siggisim): Do this rendering once on write, rather than on every read.
+		invocation.ConsoleBuffer = string(sep.screenWriter.RenderAsANSI())
+	}
```
bduffany, line 220: In order for screenWriter to have nonempty contents at this point, it looks like this `FillInvocation` call needs to happen twice, once initially, and again after parsing all events, which is what we currently do when calling `fillInvocationFromEvents`.

IMO this is a fairly error-prone API, and ideally we'd refactor at some point in the future so that `FillInvocation` only needs to be called once after parsing all events. So, if possible, it'd be nice to not add more dependencies on calling `FillInvocation` twice. Would it make sense to instead pass an optional screenWriter into `NewEventParser`?

bduffany (changes requested):

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: Done!

server/build_event_protocol/event_parser/event_parser.go
bduffany: What I mean is, in `fillInvocationFromEvents`, we do this (pseudo-code):

```go

parser := NewParser()
parser.FillInvocation(inv) // screenWriter is initialized here but nothing is written to it
for event := events {
  parser.Parse(event) // stuff gets written to screenWriter
}
parser.FillInvocation(inv) // contents of the existing screenWriter get written to consoleBuffer

```

I think it would be more clear if screenWriter were initialized in NewParser(), if possible. Currently we depend on `FillInvocation` being called before all events are parsed, but I think eventually it would be nice to remove that dependency.

tempoz: Oh! Now I understand. Yeah, you're right, that is a problem. Fixed!

bduffany (approved):

PR #1565, opened 2022-02-10
Firecracker: Use docker pull for images

bduffany:
If `docker_socket` is configured, then when pulling images for firecracker, run `docker pull` followed by `docker save`, rather than `skopeo copy docker://...`. This allows us to reuse locally cached layers from the Docker cache, as well as de-duping across pulls.

This should reduce executor warmup time, since on startup we try to pull images for both Docker and Firecracker concurrently. 

It should also improve cases where several tasks hit the executor at once with the same container-image that has not yet been pulled, and the tasks would execute extremely slowly (several minutes) due to the same image being downloaded multiple times in parallel, without deduplication. It's unclear how common this will be in production, but I have already run into this several times locally (executor grinding to a halt because it is pulling a bunch of the same image in parallel).

Note that we can't use skopeo to copy directly from the Docker daemon --> OCI image via the `docker-daemon:` source protocol, due to https://github.com/containers/image/issues/1049. As a workaround, we use `docker save` then convert from Docker format to OCI format using `skopeo copy docker-archive:/path/to/docker/save/output.tar oci:/path/to/oci/image:label`

This PR is a partial fix for https://github.com/buildbuddy-io/buildbuddy-internal/issues/1097. In future PRs, we would also benefit from de-duping the image conversion (to OCI format), unpacking (umoci), and ext4 image creation as well (mke2fs), since those are all expensive operations (~tens of seconds total).

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1097

### DIFF
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -331,6 +331,9 @@ func PullImage(ctx context.Context, client *dockerclient.Client, image string, c
 	return nil
 }
 
+// SaveImage saves an image from the Docker cache to a tarball file at the given
+// path. The image must have already been pulled, otherwise the operation will
+// fail.
 func SaveImage(ctx context.Context, client *dockerclient.Client, imageRef, path string) error {
 	r, err := client.ImageSave(ctx, []string{imageRef})
 	if err != nil {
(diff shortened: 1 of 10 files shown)


vadimberezniker (approved):
Really nice.

PR #1567, opened 2022-02-10
1/6 Reduce db dependencies

tempoz:
This is actually cleanup of the initial run for reducing dependencies on `db.go`. This CL
removes the `Raw` call from the `DBHandle` interface. This call was initially added to cut
down on the complexity of the CL that initially reduced dependencies.

The motivation for reducing depencies on the `db.go` file is that that file is forced to
use structs from an external library containing cgo, and that was breaking when we built
the executor. To avoid a recurrence of that extremely puzzling error, only files that
absolutely *must* depend on `db.go` will do so (`libmain.go` and `testenv.go`).

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1568, opened 2022-02-10
2/6 Reduce db dependencies

tempoz:
- Remove `Raw` from DBHandle interface
- Let the DBHandle determine the type of Options

This CL moves the function call for DBOptions into the DBHandle interface to reduce
dependencies on the `db.go` file.

The motivation for reducing depencies on the `db.go` file is that that file is forced to
use structs from an external library containing cgo, and that was breaking when we built
the executor. To avoid a recurrence of that extremely puzzling error, only files that
absolutely *must* depend on `db.go` will do so (`libmain.go` and `testenv.go`).

---

**Version bump**: TODO

### DIFF
--- a/enterprise/server/api/api_server.go
+++ b/enterprise/server/api/api_server.go
@@ -73,7 +73,7 @@ func (s *APIServer) GetInvocation(ctx context.Context, req *apipb.GetInvocationR
 	}
 	queryStr, args := q.Build()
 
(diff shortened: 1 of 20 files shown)


tylerwilliams (commented):
Could you remove the Options changes from this CL and I'll approve the Raw() part?

PR #1574, opened 2022-02-11
Remove UI deps on inline console buffer

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_suggestion_card.tsx
+++ b/app/invocation/invocation_suggestion_card.tsx
@@ -1,24 +1,11 @@
 import React from "react";
-import InvocationModel from "./invocation_model";
 import { Radio } from "lucide-react";
 
 interface Props {
-  model: InvocationModel;
+  buildLogs: string;
 }
 
-interface State {
-  suggestion: React.ReactElement;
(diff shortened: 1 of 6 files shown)


app/invocation/invocation_suggestion_card.tsx
```diff
@@ -56,43 +43,34 @@ export default class SuggestionCardComponent extends React.Component {
     },
   ];
 
-  componentDidMount() {
-    this.updateSuggestion();
-  }
-
-  componentDidUpdate(prevProps: Props) {
-    if (prevProps.model !== this.props.model) {
-      this.updateSuggestion();
```
siggisim, line 65: This was to prevent us from having to run a regency max over the build logs on every render. Can be expensive for large logs.

siggisim (approved):
Looks good, just one comment

app/invocation/invocation_suggestion_card.tsx
bduffany: gotcha -- added a `shouldComponentUpdate` so that it doesn't re-render unless new logs come in.

siggisim (approved):

PR #1575, opened 2022-02-11
Add BUILD file to the marketing website so it can be built and developed with Bazel

siggisim:
Adds a really dumb Yarn rule that runs a yarn command or bundles the `build` directory into a tar.

Updated website README to document the 3 workflows:
- Development with live reload
- Previewing production build locally
- Bundling static tar
---

**Version bump**: None

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -2,7 +2,7 @@
 
 ## Development
 
-To run the marketing website locally, run:
+To run the marketing website locally for development, run:
 
 ```bash
 bazel run website:start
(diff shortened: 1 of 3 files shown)


website/README.md
```diff
@@ -1,13 +1,37 @@
...
+
+To preview a static production build locally, run:
+
+```bash
+bazel run website:serve
+```
+
+View the marketing site at http://localhost:3000/
+
+You'll need to rebuild this target for it to pick up any changes.
```
tylerwilliams, line 25: could you add a line about what the difference between these is?

tylerwilliams (approved):
LGTM

neato

website/README.md
siggisim: Done

PR #1576, opened 2022-02-11
Use first_start_time and total_run_duration

luluz66:
1. only use deprecated fields as a fallback
2. remove result struct -- it's set but not used. As a result,
   deprecated fields: test_attempt_start_millis_epoch and
   test_attempt_duration_millis are removed

https://github.com/buildbuddy-io/buildbuddy-internal/issues/1162

### DIFF
--- a/server/util/timeutil/timeutil.go
+++ b/server/util/timeutil/timeutil.go
@@ -9,7 +9,7 @@ import (
 )
 
 // Returns time.Time from a timestamp proto field. When the given ts is nil, uses fallbackMillis
-func GetTime(ts *timestamppb.Timestamp, fallbackMillis int64) time.Time {
+func GetTimeWithFallback(ts *timestamppb.Timestamp, fallbackMillis int64) time.Time {
 	if ts != nil {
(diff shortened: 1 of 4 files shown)


server/util/timeutil/timeutil.go
```diff
@@ -3,8 +3,29 @@ package timeutil
 import (
 	"fmt"
 	"time"
+
+	"google.golang.org/protobuf/types/known/durationpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// Returns time.Time from a timestamp proto field. When the given ts is nil, uses fallbackMillis
+func GetTime(ts *timestamppb.Timestamp, fallbackMillis int64) time.Time {
```
tylerwilliams, line 12: nit: I hate to advocate for longer names, but I think calling these GetTimeWithFallback and GetDurationWithFallback might make the code clearer at the callsites. WDYT?

luluz66: Done

tylerwilliams (approved):
LGTM

PR #1581, opened 2022-02-14
Firecracker: Bump vsock dial timeout (again)

bduffany:
Running `firecracker_test` locally with `--runs_per_test=20 --test_filter=Docker` consistently fails at least once with the current timeout. Bumping an additional 5 sec so that the test will at least pass consistently. In prod, timeouts will be retried, so we probably shouldn't bump this much further.

---

**Version bump**: Patch


vadimberezniker (approved):

bduffany:
Will close this for now since it's only affecting manual testing -- if tests start timing out in prod after several attempts, we can increase this timeout only for docker actions.

PR #1586, opened 2022-02-14
Disable docker test workflow

bduffany:
Can be re-enabled once docker-in-firecracker works in prod

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1587, opened 2022-02-14
Add a mysql variant of usage_test using docker-in-firecracker

bduffany:
This PR serves a few purposes:
* Provide a more realistic test/example for docker-in-firecracker functionality, which runs as part of our CI and will give us a better feel for what needs to be optimized. Our current set of docker-in-firecracker tests (just `docker_test.go`) is limited to very lightweight images, like busybox.
* Start adding some automated mysql tests, which might be nice to have for some of our core functionality (such as the DB migration logic, maybe a basic e2e test with bazel running against a BB instance w/ a MySQL backend, etc.).

---

**Version bump**: Patch


server/testutil/testmysql/testmysql.go
```diff
@@ -0,0 +1,101 @@
...
+	log.Debug("Starting mysql DB...")
+
+	const dbName = "buildbuddy-test"
+	cmd := exec.Command(
+		"docker", "run", "--rm", "--detach",
+		"--env", "MYSQL_ROOT_PASSWORD=root",
+		"--env", "MYSQL_DATABASE="+dbName,
+		"--publish", fmt.Sprintf("%d:3306", port),
+		"--name", containerName,
+		"mysql:8.0",
```
vadimberezniker, line 55: Could we add variants for other mysql versions like 5.6 and 5.7?
This would give us some coverage for older mysql versions which we currently lack.

vadimberezniker (approved):
This is really cool. This will finally allow us to get mysql coverage which we were lacking before.

server/testutil/testmysql/testmysql.go
bduffany: SG, I will look into that once we are a bit more confident in these tests.

PR #1588, opened 2022-02-14
Migrate finish_time_millis and start_time_millis

luluz66:
Migrate finish_time_millis and start_time_millis from event_parser, and
build_logs_test

https://github.com/buildbuddy-io/buildbuddy-internal/issues/1162

### DIFF
--- a/enterprise/server/test/integration/build_logs/build_logs_test.go
+++ b/enterprise/server/test/integration/build_logs/build_logs_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	bespb "github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
 	bbspb "github.com/buildbuddy-io/buildbuddy/proto/buildbuddy_service"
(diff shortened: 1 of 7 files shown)


enterprise/server/test/integration/build_logs/build_logs_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/timestamppb"
```
bduffany, line 16: We usually use `"github.com/golang/protobuf/ptypes"` (from `        "@com_github_golang_protobuf//ptypes:go_default_library_gen"`) and then `ptypes.TimestampProto` to get an instance of timestamp from a `time.Time`.

There's an open issue to migrate to `google.golang.org/protobuf`, I think, but for now we probably want to use the old package everywhere.

bduffany (changes requested):

enterprise/server/test/integration/build_logs/build_logs_test.go
luluz66: I already introduced timestamppb and durationpb in earlier PRs. I think we can at least start to migrate to timestamppb and durationpb from ptypes.

bduffany (approved):

PR #1589, opened 2022-02-14
Add ability to configure Redis Ring (sharded) client instead of single server-client.

vadimberezniker:
The Ring client does not support all the same options that the simple
client does which means we can't create a ring client everywhere.
Notably we support unix sockets in redis target which are not supported
by the ring client.

The go-redis library also does not have a common option type that works
for both the simple and ring clients. It has a a UniversalOptions type
but it works for everything EXCEPT the Ring client so we introduce our
own in redisutil.

---

**Version bump**: Minor

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -34,37 +34,38 @@ type generalConfig struct {
 }
 
 type appConfig struct {
-	BuildBuddyURL             string   `yaml:"build_buddy_url" usage:"The external URL where your BuildBuddy instance can be found."`
-	EventsAPIURL              string   `yaml:"events_api_url" usage:"Overrides the default build event protocol gRPC address shown by BuildBuddy on the configuration screen."`
-	CacheAPIURL               string   `yaml:"cache_api_url" usage:"Overrides the default remote cache protocol gRPC address shown by BuildBuddy on the configuration screen."`
-	RemoteExecutionAPIURL     string   `yaml:"remote_execution_api_url" usage:"Overrides the default remote execution protocol gRPC address shown by BuildBuddy on the configuration screen."`
-	LogLevel                  string   `yaml:"log_level" usage:"The desired log level. Logs with a level >= this level will be emitted. One of {'fatal', 'error', 'warn', 'info', 'debug'}"`
-	GRPCMaxRecvMsgSizeBytes   int      `yaml:"grpc_max_recv_msg_size_bytes" usage:"Configures the max GRPC receive message size [bytes]"`
-	GRPCOverHTTPPortEnabled   bool     `yaml:"grpc_over_http_port_enabled" usage:"Cloud-Only"`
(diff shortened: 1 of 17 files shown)


server/config/config.go
```diff
@@ -251,21 +253,29 @@ type SSLConfig struct {
 	DefaultHost      string   `yaml:"default_host" usage:"Host name to use for ACME generated cert if TLS request does not contain SNI."`
 }
 
+type ShardedRedisConfig struct {
+	Shards   []string `yaml:"shards" usage:"Ordered list of Redis shard addresses."`
+	Username string   `yaml:"username" usage:"Redis username"`
+	Password string   `yaml:"password" usage:"Redis password"`
```
bduffany, line 259: would it ever make sense for shards to have their own unique passwords/usernames? (wondering if ShardedRedisConfig could instead be replaced with a list of targets)

vadimberezniker: The redis ring client in go-redis does not support specifying different usernames/passwords for different shards.

server/config/config.go
```diff
@@ -34,37 +34,38 @@ type generalConfig struct {
...
+	GlobalFilterEnabled       bool               `yaml:"global_filter_enabled" usage:"If set, the global filter will be enabled in the UI."`
+	TestGridV2Enabled         bool               `yaml:"test_grid_v2_enabled" usage:"Whether to enable test grid V2"`
+	UsageEnabled              bool               `yaml:"usage_enabled" usage:"If set, the usage page will be enabled in the UI."`
+	UsageStartDate            string             `yaml:"usage_start_date" usage:"If set, usage data will only be viewable on or after this timestamp. Specified in RFC3339 format, like 2021-10-01T00:00:00Z"`
+	UsageTrackingEnabled      bool               `yaml:"usage_tracking_enabled" usage:"If set, enable usage data collection."`
+	DefaultRedisTarget        string             `yaml:"default_redis_target" usage:"A Redis target for storing remote shared state. To ease migration, the redis target from the remote execution config will be used if this value is not specified."`
+	DefaultShardedRedis       ShardedRedisConfig `yaml:"default_sharded_redis" usage:"Configuration for storing ephemeral state across multiple Redis instances. Mutually exclusive with default_redis_target."`
```
bduffany, line 67: nit: wdyt about calling this `default_redis_config`? (same for other config fields)

bduffany (approved):
LGTM, very clean

tylerwilliams (approved):
LGTM

server/config/config.go
vadimberezniker: I think we should keep the current name to make it clear that this is for setting up sharding.
I expect almost every on-prem customer to keep running BB using a non-sharded Redis setup and hopefully this naming will make it clearer that they shouldn't touch the sharded config.

PR #1591, opened 2022-02-14
Add ability to restart Redis in tests.

vadimberezniker:
---

**Version bump**: None

### DIFF
--- a/enterprise/server/testutil/testredis/testredis.go
+++ b/enterprise/server/testutil/testredis/testredis.go
@@ -6,7 +6,6 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


enterprise/server/testutil/testredis/testredis.go
```diff
@@ -64,27 +70,48 @@ func Start(t testing.TB) string {
...
+
+// Start spawns a Redis server for the given test and returns a handle to the running process.
+func Start(t testing.TB) *Handle {
+	// redis socket must be in /tmp, redis won't read socket files in arbitrary locations
+	socketPath := testfs.MakeSocket(t, "redis.sock")
+	target := fmt.Sprintf("unix://%s", socketPath)
+
+	handle := &Handle{
+		Target:     target,
+		socketPath: socketPath,
```
bduffany, line 109: Does this need to pass `t: t,` ?

Otherwise LGTM

bduffany (approved):

enterprise/server/testutil/testredis/testredis.go
vadimberezniker: Woops, good catch!

bduffany (approved):

PR #1594, opened 2022-02-15
Add workflow systemrc configs for env-specific endpoints

bduffany:
This will allow us to run dev workflows against dev and prod workflows against prod, by adding the following to our .bazelrc:

```
# Use env-specific targets. These configs are populated by the workflow runner in `/etc/bazel.bazelrc`.
build:workflows --config=buildbuddy_bes_backend
build:workflows --config=buildbuddy_bes_results_url
build:workflows --config=buildbuddy_remote_cache
build:workflows --config=buildbuddy_remote_executor
```

---

**Version bump**: Patch


vadimberezniker (approved):
Clever!

siggisim (approved):

PR #1597, opened 2022-02-15
Check sql modes instead of version

luluz66:
1. log mysql version and sql versions
2. use sql modes to determine whether to use INPLACE or COPY algorithm


tylerwilliams (approved):
LGTM

PR #1600, opened 2022-02-16
Add `Show all events` for `Raw` tab

tempoz:
Adds a `Show all events` button in the same vein as the `Show more events` button to the `RAW`
tab for invocations.

---

**Version bump**: Minor


enterprise/config/buildbuddy.local.yaml
```diff
@@ -1,5 +1,5 @@
 app:
-  build_buddy_url: "http://localhost:8080"
+  build_buddy_url: "http://delubrum.pw:8080"
```
brentleyjones, line 2:
```suggestion
  build_buddy_url: "http://localhost:8080"
```

tempoz: Ugh, finally accidentally commited this

tempoz:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/809

app/invocation/invocation_raw_logs_card.tsx
```diff
@@ -92,6 +96,13 @@ export default class RawLogsCardComponent extends React.Component {
                   See more events
                 </div>
               )}
+            {this.props.pageSize &&
+              filteredEvents.length > this.props.pageSize * this.state.numPages &&
+              !!this.state.numPages && (
+                <div className="all" onClick={this.handleAllClicked.bind(this)}>
```
bduffany, line 102: nit: since this renders under the same conditions as the above div, you can instead write this as

```
            {this.props.pageSize &&
              filteredEvents.length > this.props.pageSize * this.state.numPages &&
              !!this.state.numPages && (
                <>
                <div className="more" onClick={this.handleMoreClicked.bind(this)}>
                  See more events
                </div>
                <div className="more" onClick={this.handleAllClicked.bind(this)}>
                  See all events
                </div>
                </>
              )}
```

The `<>...</>` is short-hand for `<React.Fragment>...</React.Fragment>` which lets you group together elements without an explicit DOM wrapper such as `<div>`.

See https://reactjs.org/docs/fragments.html

app/root/root.css
```diff
@@ -601,6 +601,14 @@ svg.icon.orange {
   user-select: none;
 }
 
+.card .all {
```
bduffany, line 604: nit: You can reuse the existing className (`more`) to avoid introducing new styles.

app/invocation/invocation_raw_logs_card.tsx
```diff
@@ -32,6 +32,10 @@ export default class RawLogsCardComponent extends React.Component {
     this.setState({ ...this.state, numPages: this.state.numPages + 1 });
   }
 
+  handleAllClicked() {
+    this.setState({ ...this.state, numPages: Number.MAX_SAFE_INTEGER });
```
bduffany, line 36: This is probably non-obvious, but writing `this.setState({ ...this.state, foo: bar })` is redundant because `setState` (somewhat confusingly) updates only the provided state keys, and any keys which are not present in the passed in state updates object are preserved from the existing state.

So you can remove the `...this.state` and just write `this.setState({ numPages: Number.MAX_SAFE_INTEGER })`. Bonus points if you would like to clean up the other places in this file where we do this as well -- `this.setState(this.state)` in `handleEventClicked` can also be replaced with `this.setState({})`.

bduffany (changes requested):

app/invocation/invocation_raw_logs_card.tsx
tempoz: Thanks; done!

app/invocation/invocation_raw_logs_card.tsx
tempoz: Done!

app/root/root.css
tempoz: Done!

bduffany (approved):
LGTM, nice :+1:

PR #1601, opened 2022-02-16
Firecracker: add basic tracing

bduffany:
Tracing is very basic at the moment to give us a rough idea of where all the time is being spent.

Initial trace results show that a good chunk of time is spent setting up networking and connecting to the VM exec server (which may partly be due to init time), so we might want to spend some time optimizing those.

Example trace:

image

---

**Version bump**: Patch


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1075,29 +1107,38 @@ func (c *FirecrackerContainer) Create(ctx context.Context, actionWorkingDir stri
...
 	}
-	if err := m.Start(vmCtx); err != nil {
+	if err := c.startMachine(ctx, vmCtx, m); err != nil {
 		return status.InternalErrorf("Failed starting machine: %s", err)
 	}
 	c.machine = m
 	return nil
 }
 
+func (c *FirecrackerContainer) startMachine(ctx context.Context, startCtx context.Context, m *fcclient.Machine) error {
```
tylerwilliams, line 1117: might be clearer to not put this in a separate method like this. The two contexts coming from somewhere else are a bit mind boggling.

(also, not relevant if you remove this, but when you have more than one sequential arguments to a function in go that have the same type, you can omit the type information for all but the last one to save space)

bduffany: Moved the tracing back to the original callsite, with the help of a util func that wraps an anonymous func with a span. wdyt about this approach?

bduffany: Resolved offline, using an inline closure instead

tylerwilliams (approved):
LGTM

(this is the way)

PR #1607, opened 2022-02-16
Add one more line to ansi buffer

tempoz:
Some invocations still indicate that the cursor is trying to move 1 line above the top of the
screen, so let's add 1 more line. I think this may be trying to erase a newline at the end of
the first line outside the buffer and not actually affecting the output display, since I
haven't seen malformed logs since we last incremented the ANSI buffer, but this should
silence complaints in the log.

---

**Version bump**: Patch


tempoz:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/962.

siggisim (approved):

PR #1610, opened 2022-02-16
Don't allow redirects that are not on this server

tylerwilliams:
Don't allow redirects that are not on this server.

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
 
+    burl "github.com/buildbuddy-io/buildbuddy/server/util/url"
 	akpb "github.com/buildbuddy-io/buildbuddy/proto/api_key"
 	oidc "github.com/coreos/go-oidc"
 )
@@ -94,6 +95,15 @@ var (
 
 func jwtKeyFunc(token *jwt.Token) (interface{}, error) {
 	return jwtKey, nil
+}
+
+func sameHostname(urlStringA, urlStringB string) bool {
+	if urlA, err := url.Parse(urlStringA); err == nil {
+		if urlB, err := url.Parse(urlStringB); err == nil {
+			return urlA.Hostname() == urlB.Hostname()
+		}
+	}
+	return false
 }
 
 type Claims struct {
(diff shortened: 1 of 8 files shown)


server/backends/github/github.go
```diff
@@ -107,6 +109,12 @@ func (c *GithubClient) Link(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	redirectURL := r.FormValue("redirect_uri")
```
bduffany, line 112: nit: replace `r.FormValue("redirect_uri")` on line 125 below with `redirectURL`, to minimize the possibility of accidentally using an unvalidated value in case this code gets shifted around in the future

enterprise/server/auth/auth.go
```diff
@@ -484,29 +485,13 @@ func NewOpenIDAuthenticator(ctx context.Context, env environment.Env, authConfig
...
-			return urlA.Hostname() == urlB.Hostname()
-		}
-	}
-	return false
-}
-
 func (a *OpenIDAuthenticator) validateRedirectURL(redirectURL string) error {
-	if a.myURL.Host == "" {
-		return status.FailedPreconditionError("You must specify a build_buddy_url in your config to enable authentication. For more information, see: https://www.buildbuddy.io/docs/config-app")
-	}
```
bduffany, line 499: nit: do we want to keep this check? Seems to give a nicer error message if the user just forgot to specify build_buddy_url

server/util/url/url_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/buildbuddy-io/buildbuddy/server/config"
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
+	"github.com/buildbuddy-io/buildbuddy/server/util/healthcheck"
+	"github.com/buildbuddy-io/buildbuddy/server/util/url"
+)
+
+func envWithAppURL(t *testing.T, appUrl string) environment.Env {
+	c, err := config.NewConfiguratorFromData([]byte(fmt.Sprintf("app:\n  build_buddy_url: %s\n", appUrl)))
```
bduffany, line 15: nice, I like this new pattern for getting a minimal testenv :+1: will start using this in future tests

bduffany (approved):

server/util/url/url_test.go
tylerwilliams: @tempoz is going to fix everything™ and make all of these config vars flags soon which will make this much easier -- you just set the flag value for the lifespan of the test.

enterprise/server/auth/auth.go
tylerwilliams: Done.

server/backends/github/github.go
tylerwilliams: Done.

PR #1612, opened 2022-02-17
Clean up executor test

luluz66:
executeCommand is deleted from Executor; so executor_test.go no longer
compiles. I moved the test for BareCommandContainer to bare_test.go; and
deleted tests for DockerCommandContainer since we already have
docker_test.go


tylerwilliams (approved):
LGTM

(thanks for cleaning this up!)

PR #1613, opened 2022-02-17
Replace github.com/dgrijalva/jwt-go with github.com/golang-jwt/jwt.

tylerwilliams:
Should fix https://github.com/buildbuddy-io/buildbuddy/security/dependabot/13


bduffany:
I recall these 2 libs handle the audience field in the JWT differently -- the old lib only supports single-valued audiences while the new lib supports more than one. Any idea whether this will cause issues?

deps.bzl
```diff
@@ -826,13 +826,13 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         sum = "h1:DshxFxZWXUcO0xX476VJC07Xsr6ZCBVRHKZ93Oh7Evo=",
         version = "v1.6.0",
     )
-
     go_repository(
```
bduffany, line 829: nit: can this be removed? (I think gazelle might not delete stuff from `deps.bzl`, so may be worth deleting and rerunning buildfix)

bduffany (approved):

PR #1614, opened 2022-02-17
Firecracker: Don't ignore errors from `Remove`

bduffany:
When recycling a runner (returning it to the runner pool) we first call `Pause`, and if `Pause` returns an error, we abort the recycle attempt and don't add the runner to the pool.

Since firecracker calls `Remove` as part of `Pause`, this PR ensures that `Pause` returns an error if `Remove` fails -- otherwise, the container might be in a weird / unexpected state when we later try to resume, due to the removal failure.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1615, opened 2022-02-17
Update mongodb dependency version (sigh).

tylerwilliams:
Should fix https://github.com/buildbuddy-io/buildbuddy/security/dependabot/15


vadimberezniker (approved):
hah... mongodb?!?

tylerwilliams:
> hah... mongodb?!?

First I tried to delete the dep but it's used by something :/

PR #1616, opened 2022-02-17
Don't return from stats recorder goroutine when invocations are retried

bduffany:
Attempt at fixing the error: `"Failed to write cache stats: stats recorder task buffer is full"`, which we believe is happening when all stats recorder workers have returned.

Also replace some `log.Errorf` occurrences with `alert.UnexpectedEvent` so we can be alerted of this more quickly in the future.

---

**Version bump**: Patch

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -261,7 +261,7 @@ func (r *statsRecorder) Start() {
 				// for any other purpose).
 				hit_tracker.CleanupCacheStats(ctx, r.env, task.invocationJWT.id)
 				if !updated {
-					log.Debugf("Attempt %d of invocation %s pre-empted by more recent attempt, no cache stats flushed.", task.invocationJWT.attempt, task.invocationJWT.id)
+					log.Warningf("Attempt %d of invocation %s pre-empted by more recent attempt, no cache stats flushed.", task.invocationJWT.attempt, task.invocationJWT.id)
 					// Don't notify the webhook; the more recent attempt should trigger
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
LGTM (but get @vadimberezniker's OK too please since we're going to cherry-pick this)

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -259,7 +261,10 @@ func (r *statsRecorder) Start() {
 				// for any other purpose).
 				hit_tracker.CleanupCacheStats(ctx, r.env, task.invocationJWT.id)
 				if !updated {
-					return status.CanceledErrorf("Attempt %d of invocation %s pre-empted by more recent attempt, no cache stats flushed.", task.invocationJWT.attempt, task.invocationJWT.id)
+					log.Debugf("Attempt %d of invocation %s pre-empted by more recent attempt, no cache stats flushed.", task.invocationJWT.attempt, task.invocationJWT.id)
```
vadimberezniker, line 264: We don't have Debug logs enabled in prod. If you want to know whether it's happening in prod, you'd need to raise to at least Info.

vadimberezniker (approved):

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: It looks like we use `Warningf` in other cases where invocations are pre-empted, so I changed this to Warning level. (Though we might want to reconsider downgrading these to Debug level once we are more confident that we handle retries correctly)

PR #1617, opened 2022-02-17
Refactor invocation finalization workers and add prometheus metrics

bduffany:
* Refactor workers so that it's more obvious that they should run forever, so we don't accidentally terminate them.
* Add metrics so we can monitor/alert on the status of these finalizers.
* Use more than one invocation lookup worker for webhook notifications (previously all webhook notifications were blocked on a single worker looking up invocations in serial).

image

---

**Version bump**: Patch

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -261,7 +261,7 @@ func (r *statsRecorder) Start() {
 				// for any other purpose).
 				hit_tracker.CleanupCacheStats(ctx, r.env, task.invocationJWT.id)
 				if !updated {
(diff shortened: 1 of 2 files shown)


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -233,53 +236,62 @@ func readScoreCard(ctx context.Context, env environment.Env, invocationID string
...
 
 func (r *statsRecorder) Start() {
-	r.eg = errgroup.Group{}
 	ctx := context.Background()
 	for i := 0; i < numStatsRecorderWorkers; i++ {
-		r.eg.Go(func() error {
+		metrics.StatsRecorderWorkers.Inc()
+		go func() {
+			defer metrics.StatsRecorderWorkers.Dec()
 			for task := range r.tasks {
```
vadimberezniker, line 244: If we had Go generics, I think we could write a nice little wrapper for creating and monitoring these types of worker pools. Some day :)

vadimberezniker (approved):

PR #1619, opened 2022-02-17
Raise open file descriptor limit for VMs.

vadimberezniker:
The current limit (4096) is not sufficient for large Bazel builds.

The default process hard limit in the kernel is 4096, but it is modifiable by any
privileged process. The system hard limit is much higher.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1624, opened 2022-02-18
Firecracker: Use correct context when cleaning up forwarding rule

bduffany:
This was preventing the firecracker container from being cleaned up properly (it was failing with "context canceled")

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1625, opened 2022-02-18
Firecracker: Improve EOF handling in mergeDiffSnapshot

bduffany:
Some data can be read when at EOF -- make sure to write the data before breaking out of the loop.

---

**Version bump**: Patch


vadimberezniker:
Can this happen in practice? The file evenly divides into the block size so I can't see a case where you would see a read that returns both data and EOF.

bduffany:
@vadimberezniker Makes sense -- in that case, I don't think it'll happen in practice. wdyt about merging this anyway just to make the code less surprising? (Personally I just have alarm bells going off whenever I see an EOF check immediately after the read)

bduffany:
Closing since I don't feel super strongly about getting it merged and it's not actually causing any problems.

PR #1626, opened 2022-02-18
Firecracker: Handle ENXIO errors in mergeDiffSnapshot

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1184

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -455,9 +455,6 @@ func mergeDiffSnapshot(ctx context.Context, baseSnapshotPath string, diffSnapsho
 					return err
 				}
 				offset += int64(n)
-				if offset >= regionEnd {
-					break
-				}
 			}
 			return nil
 		})
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/json"
+	"errors"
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -431,6 +433,11 @@ func mergeDiffSnapshot(ctx context.Context, baseSnapshotPath string, diffSnapsho
 				// 3 is the Linux constant for the SEEK_DATA option to lseek.
 				newOffset, err := gin.Seek(offset, 3)
 				if err != nil {
+					// ENXIO is expected when the offset is within a hole at the end of
+					// the file.
+					if err := errors.Unwrap(err); err == syscall.ENXIO {
```
vadimberezniker, line 438: Since we're checking for syscall errors, I think we might as well call `syscall.Seek` directly instead of going through the file wrapper.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -431,6 +433,11 @@ func mergeDiffSnapshot(ctx context.Context, baseSnapshotPath string, diffSnapsho
 				// 3 is the Linux constant for the SEEK_DATA option to lseek.
 				newOffset, err := gin.Seek(offset, 3)
 				if err != nil {
+					// ENXIO is expected when the offset is within a hole at the end of
```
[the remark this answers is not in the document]
vadimberezniker: That's correct, but that extra read of a single block doesn't matter in the grand scheme of things. I'd rather make the code simpler.

vadimberezniker (approved):

PR #1627, opened 2022-02-19
Add option to mount workspace ext4 fs instead of extracting contents.

vadimberezniker:
Mounting the filesystem file directly is significantly faster.

This is behind a flag for now because it doesn't work well locally due
to needing root. I tried using `udisksctl` to get around this
requirement but ran into the following issues:
   1) On local Ubuntu, at least on my
      installation, something automatically mounts the filesystem as soon as I
      set up the loop device even before I call "mount".
   2) On a deployed image, we would need to have DBus & udisk daemons
      installed and running for this to work.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/BUILD
+++ b/enterprise/server/remote_execution/containers/firecracker/BUILD
@@ -37,6 +37,7 @@ go_library(
             "//proto:vmvfs_go_proto",
             "//server/environment",
             "//server/interfaces",
+            "//server/util/alert",
             "//server/util/disk",
             "//server/util/hash",
             "//server/util/log",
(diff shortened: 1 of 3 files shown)


bduffany (commented):
Seems like a nice perf boost, but will let Tyler LGTM so he has a chance to review.

BTW, for hosted bazel specifically, I think the proposal in https://github.com/buildbuddy-io/buildbuddy-internal/issues/1188 would also help with the perf of `copyOutputsToWorkspace` (creating 2 separate volumes, one for the workspace and one for a persistent root disk that is not updated across container re-runs, and then running hosted bazel in the root disk rather than the workspace disk, and also ensuring that the bazel cache dir is somewhere in the persistent root disk).

tylerwilliams (approved):

PR #1630, opened 2022-02-22
Add option to disable requesting refresh token scope

siggisim:
Some auth providers don't support this / don't enable this.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1634, opened 2022-02-22
Don't track usage stats for `file/download`

tempoz:
Adds a value to the context for requests coming from the `file/download` endpoint and then
checks that value in the hit tracker to see if usage should be tracked.

---

**Version bump**: Minor

### DIFF
--- a/server/http/filters/filters.go
+++ b/server/http/filters/filters.go
@@ -2,6 +2,7 @@ package filters
 
 import (
 	"compress/gzip"
+	"context"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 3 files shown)


tempoz:
Fixes buildbuddy-io/buildbuddy-internal#1051.

tylerwilliams:
Could we generalize this a tiny bit by:
 - setting a header for any request coming from the frontend in rpc_service.ts
 - pass that through to the context in server/rpc/filters/filters.go

Then we can use this more broadly if we want to exempt UI stats or track them separately.

tempoz:
@tylerwilliams Sounds good! I'll get right on changing it over.

PR #1644, opened 2022-02-24
Fix some bugs with test grid V2

bduffany:
* Don't show the "load more" button after the server returns `paginationToken=""` on the last page -- otherwise we'd start re-fetching commits from the beginning
* Show a loading spinner in the "load more" button while loading
* Fix padding on commit timeline header
* Prevent commit timeline from shrinking in height after the "load more" button disappears
* Update wording in empty state page to mention that commit SHA and repo URL are required (inline the instructions for configuring `workspace_status.sh`, and de-emphasize the "view build metadata guide" CTA since all the relevant info is now inlined in the current page)

---

**Version bump**: Patch


enterprise/app/tap/tap.tsx
```diff
@@ -394,18 +396,54 @@ export default class TapComponent extends React.Component<Props, State> {
...
-              </p>
+              {this.isV2 ? (
+                <p>
+                  To see a CI test grid, make sure your CI tests are configured as follows:
+                  <ul>
+                    <li>
+                      Add <code className="inline-code">--build_metadata=ROLE=CI</code> to your CI bazel test command
+                    </li>
+                    <li>
+                      Add a file called <code className="inline-code">workspace_status.sh</code> to the root of your
```
siggisim, line 407: Not sure how I feel about being so explicit about this. Another totally reasonable approach is to use something like Buildkite where all of the environment variables are already set.

siggisim (approved):

enterprise/app/tap/tap.tsx
bduffany: Ah right. Replaced these with links to the build metadata guide

PR #1649, opened 2022-02-25
Add tests for Invocation API in `buildbuddy_server`

tempoz:
Add smoke tests `GetInvocation`, `UpdateInvocation`, `SearchInvocation`, and `DeleteInvocation`.

---

**Version bump**: Patch

### DIFF
--- a/server/buildbuddy_server/buildbuddy_server_test.go
+++ b/server/buildbuddy_server/buildbuddy_server_test.go
@@ -1,0 +1,202 @@
+package buildbuddy_server_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/buildbuddy-io/buildbuddy/server/build_event_protocol/build_event_handler"
+	"github.com/buildbuddy-io/buildbuddy/server/buildbuddy_server"
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/tables"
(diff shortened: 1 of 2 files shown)


tempoz:
Related: buildbuddy-io/buildbuddy-internal#931

server/buildbuddy_server/buildbuddy_server_test.go
```diff
@@ -0,0 +1,202 @@
...
+	return testInvocationID, err
+}
+
+func TestGetInvocation(t *testing.T) {
+	te := testenv.GetTestEnv(t)
+	auth := testauth.NewTestAuthenticator(testauth.TestUsers(user1, group1, user2, group2))
+	te.SetAuthenticator(auth)
+
+	iid, err := createInvocationForTesting(te, user1)
+	assert.NoError(t, err)
```
vadimberezniker, line 75: You should use `require.NoError` and other `require` functions over their `assert` counterparts when it's not useful to continue the test after an assertion failure. The `assert` functions don't stop the test execution so the test can report multiple errors or terminate incorrectly because of precondition failures.

vadimberezniker (approved):
Thanks for adding these tests. Please take a look at my comment re: require vs assert

server/buildbuddy_server/buildbuddy_server_test.go
tempoz: Awesome; thanks for the information as always, Vadim!

PR #1650, opened 2022-02-25
Remove need for trailing slash for some URLs

tempoz:
Paths served by the static file server required a trailing slash, which was potentially confusing.

---

**Version bump**: Patch

### DIFF
--- a/server/libmain/libmain.go
+++ b/server/libmain/libmain.go
@@ -398,6 +398,7 @@ func StartAndRunServices(env environment.Env) {
 	// Register all of our HTTP handlers on the default mux.
 	mux.Handle("/", httpfilters.WrapExternalHandler(env, staticFileServer))
 	for _, appRoute := range(appRoutes) {
+		// this causes the muxer to handle redirects from e. g. /path -> /path/
 		mux.Handle(appRoute, httpfilters.WrapExternalHandler(env, staticFileServer))
 	}
 	mux.Handle("/app/", httpfilters.WrapExternalHandler(env, http.StripPrefix("/app", afs)))
--- a/server/libmain/libmain.go
+++ b/server/libmain/libmain.go
@@ -397,7 +397,7 @@ func StartAndRunServices(env environment.Env) {
 	mux := env.GetMux()
 	// Register all of our HTTP handlers on the default mux.
 	mux.Handle("/", httpfilters.WrapExternalHandler(env, staticFileServer))
-	for _, appRoute := range(appRoutes) {
+	for _, appRoute := range appRoutes {
 		// this causes the muxer to handle redirects from e. g. /path -> /path/
 		mux.Handle(appRoute, httpfilters.WrapExternalHandler(env, staticFileServer))
 	}
--- a/server/libmain/libmain.go
+++ b/server/libmain/libmain.go
@@ -397,6 +397,9 @@ func StartAndRunServices(env environment.Env) {
 	mux := env.GetMux()
 	// Register all of our HTTP handlers on the default mux.
 	mux.Handle("/", httpfilters.WrapExternalHandler(env, staticFileServer))
+	for _, appRoute := range(appRoutes) {
+		mux.Handle(appRoute, httpfilters.WrapExternalHandler(env, staticFileServer))
+	}
 	mux.Handle("/app/", httpfilters.WrapExternalHandler(env, http.StripPrefix("/app", afs)))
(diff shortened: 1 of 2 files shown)


tempoz:
Fixes buildbuddy-io/buildbuddy-internal#303

siggisim:
Thoughts on removing the slashes here in stead so we don't have to do this string manipulation on every http request? https://github.com/buildbuddy-io/buildbuddy/blob/d4470dc4f8efdd6c4fda2c5a9967ee994bfce13f/server/libmain/libmain.go#L79

siggisim:
You'll also probably need remove some slashes here in order for the front-end routing to work properly, but you'd need to test to make sure:
https://github.com/buildbuddy-io/buildbuddy/blob/master/app/router/router.tsx#L415

tempoz:
The problem with that is that then those paths aren't being served *with* a trailing slash. `http.ServeMux` automatically redirects paths that lack a trailing slash to ones that have one, if they exist, but it doesn't work in reverse.

tempoz:
The frontend URLs are still fine; all URLs that previously worked still work in the same way, this just makes URLs that were missing a trailing slash also work.

siggisim:
> The problem with that is that then those paths aren't being served _with_ a trailing slash. `http.ServeMux` automatically redirects paths that lack a trailing slash to ones that have one, if they exist, but it doesn't work in reverse.

Not sure how that would affect this line - rootPath is just looping over that hardcoded list of paths:
```
   if strings.HasPrefix(r.URL.Path, strings.TrimSuffix(rootPath, "/")) {
```

tempoz:
mmmm, I see, I had mistakenly believed the static file server was using that list to determine which URLs it handles via a mux, but it just serves everything from "/".

tempoz:
The concern I have about removing the trailing slash from that list is that then e. g. `code` and `codenames` would redirect to the same thing, I believe?

tempoz:
New code should only serve to `/path/.*` and `/path`

siggisim:
> New code should only serve to `/path/.*` and `/path`

Ah cool, this is nice!


> The frontend URLs are still fine; all URLs that previously worked still work in the same way, this just makes URLs that were missing a trailing slash also work.

The issue here is that if we use those paths (with slashes) to do any sort of prefix matching on the frontend - the page will be served by the server and load correctly, but the front-end will show the wrong data. 

A few examples that I found while perusing the code:
- We look for `/usage/` with a slash to decide whether or not to show the usage page. This means if you visit `/usage` the page will load, but likely show something else: https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/app/root/root.tsx#L119
- To decide whether or not an item is selected in the sidebar, we check for the slash. This means that if you visit `/trends`, you will see the correct content, but the *Trends* sidebar item won't be lit up: https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/app/sidebar/sidebar.tsx#L66

My ideal fix would be to use these path constants https://github.com/buildbuddy-io/buildbuddy/blob/master/app/router/router.tsx#L415 anywhere we're trying to match the url to a "place". But deciding whether or not those constants should have trailing slashes is a bit tricky, because sometimes you want them (like `/history/user/` where the path is actually `/history/user/tempoz`) and other times they're optional (like `/trends`).

bduffany:
> My ideal fix would be to use these path constants https://github.com/buildbuddy-io/buildbuddy/blob/master/app/router/router.tsx#L415 anywhere we're trying to match the url to a "place". But deciding whether or not those constants should have trailing slashes is a bit tricky, because sometimes you want them (like `/history/user/` where the path is actually `/history/user/tempoz`) and other times they're optional (like `/trends`).

One way to slightly simplify the logic is to have the client automatically strip any trailing slashes in the URL path using `window.history.replaceState`, or have the server do it via an HTTP redirect. That way, you have a guarantee that `path === "/foo"` if and only if you are at the `foo` page, and `path.startsWith('/foo/`)` if and only if you're on a "sub-page" of foo

tempoz:
One option is to use a submux for this on the backend, since that would automatically redirect `/path` to `/path/`, if `/path/` exists, which I think should avoid any need for frontend changes?

tempoz:
Just put the routes in the normal muxer, no reason to introduce a submuxer.

siggisim (approved):
Thanks for doing this!

bduffany (approved):

PR #1651, opened 2022-02-25
Refresh auth token every 30 min if necessary

siggisim:
Does so by calling GetUser every 30 min.

If we get a Unauthenticated or PermissionDenied error from this call, we show a yellow "Logged out" banner and redirect to the login screen.

---

**Version bump**: Patch

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -14,7 +14,7 @@ export { User };
 
 const SELECTED_GROUP_ID_LOCAL_STORAGE_KEY = "selected_group_id";
 const IMPERSONATING_GROUP_ID_SESSION_STORAGE_KEY = "impersonating_group_id";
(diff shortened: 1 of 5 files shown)


app/auth/auth_service.ts
```diff
@@ -44,6 +46,21 @@ export class AuthService {
           this.onUserRpcError(error);
         }
       });
+
+    setInterval(() => {
+      if (this.user) this.refreshToken();
+    }, TOKEN_REFRESH_INTERVAL_SEC * 1000);
```
tylerwilliams, line 52: wdyt about defining the const in ms and skipping the conversion here? Since you're giving it a 30 min value anyway, cut out the middle man.

tylerwilliams (approved):
LGTM

app/auth/auth_service.ts
siggisim: done

PR #1656, opened 2022-02-27
Migrate API GetLog to new chunked event log handler

siggisim:
Also updates test to check log contents - this will prevent the API from getting out of sync again in the future.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1660, opened 2022-02-28
Return p.buildRoot when docker is not enabled

luluz66:
This is to support podman.

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -568,15 +568,16 @@ func (p *Pool) hostBuildRoot() string {
 	if hd := p.env.GetConfigurator().GetExecutorConfig().HostRootDirectory; hd != "" {
 		return filepath.Join(hd, "remotebuilds")
 	}
-	if p.podID == "" {
-		// Probably running on bare metal -- return the build root directly.
-		return p.buildRoot
+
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
This LGTM

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -568,15 +568,16 @@ func (p *Pool) hostBuildRoot() string {
...
-		// Probably running on bare metal -- return the build root directly.
-		return p.buildRoot
-	}
-	// Running on k8s -- return the path to the build root on the *host* node.
-	// TODO(bduffany): Make this configurable in YAML, populating {{.PodID}} via template.
-	// People might have conventions other than executor-data for the volume name + remotebuilds
-	// for the build root dir.
-	return fmt.Sprintf("/var/lib/kubelet/pods/%s/volumes/kubernetes.io~empty-dir/executor-data/remotebuilds", p.podID)
+
+	if p.dockerClient != nil && p.podID != "" {
```
bduffany, line 572: @luluz66 IMO it's a little confusing to have `hostBuildRoot` return something other than the host build root directory. To support podman, could you instead pass the executor root directory to podman, rather than the host build root? (Sorry I didn't catch this in the earlier PR -- if I understand correctly, podman doesn't need to know the host build root directory)

PR #1663, opened 2022-02-28
Attempt to auto-relogin when sessions expire

siggisim:
This should greatly reduce the frequency that users with {no refresh token, short-lived refresh tokens, SAML auth} see the login screen. It should now only ever be shown if they manually log out or if access is revoked on the auth provider side.

On the client side: 
 - Attempt to refresh the token if we get a "session expired" error message on the initial page load
 - If token refresh fails, try to auto-relogin once by redirecting to the /login/ page

On the server side:
- Return a "session expired" error message:
  - if we are attempting to refresh the OIDC token, and the refresh fails
  - if the slug cookie is set, but we don't have a valid SAML session
- Only force the OIDC consent screen to be showed if a session ID isn't set (i.e. the user has logged out).
- Fallback to cookies for issuer / slug on the `/login/` page if no url params are set. 
- Plumb through some error message that were previously being swallowed

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -91,10 +91,10 @@ const (
 	// WARNING: app/auth/auth_service.ts depends on these messages matching.
 	userNotFoundMsg = "User not found"
 	loggedOutMsg    = "User logged out"
+	ExpiredSessionMsg = "User session expired"
 )
 
 var (
-	authCodeOption []oauth2.AuthCodeOption = []oauth2.AuthCodeOption{oauth2.AccessTypeOffline, oauth2.ApprovalForce}
 	apiKeyRegex                            = regexp.MustCompile(APIKeyHeader + "=([a-zA-Z0-9]*)")
 	jwtKey                                 = []byte("set_the_jwt_in_config") // set via config.
 )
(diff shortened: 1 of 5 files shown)


enterprise/server/saml/saml.go
```diff
@@ -88,6 +92,11 @@ func (a *SAMLAuthenticator) AuthenticatedHTTPContext(w http.ResponseWriter, r *h
 			return ctx
 		}
 	}
+
+	if slug := auth.GetCookie(r, slugCookie); err != nil && slug != "" {
+		return auth.AuthContextWithError(ctx, status.PermissionDeniedErrorf("Error getting service provider for slug %s: %s", slug, err.Error()))
+	}
```
tylerwilliams, line 98: i think this might be clearer as if err == nil {
.... block ...
} else {
  if slug := auth.GetCookie(r, slugCookie); slug != "" {
    ... do something
  } 
}

fallback case()

enterprise/server/auth/auth.go
```diff
@@ -532,6 +536,19 @@ func (a *OpenIDAuthenticator) getAuthConfigForSlug(slug string) authenticator {
...
+func (a *OpenIDAuthenticator) getAuthCodeOptions(r *http.Request) []oauth2.AuthCodeOption {
+	options := []oauth2.AuthCodeOption{}
+	if !a.env.GetConfigurator().GetDisableRefreshToken() {
+		options = append(options, oauth2.AccessTypeOffline)
+	}
+	sessionID := GetCookie(r, sessionIDCookie)
+	// If a session doesn't already exist, force a consent screen.
+	if sessionID == "" {
+		options = append(options, oauth2.ApprovalForce)
+	}
```
tylerwilliams, line 548: is this necessary?

tylerwilliams (approved):
LGTM

enterprise/server/auth/auth.go
siggisim: Not strictly, but I find it useful for being able to switch between Google accounts if necessary. Kinda hate when you log, out and then try to re-login with another account and you get quickly auto-logged in without a consent screen.

tylerwilliams: OK, wfm.

enterprise/server/saml/saml.go
siggisim: Done

PR #1668, opened 2022-03-02
De-flake runner recycling integration test

bduffany:
Task routing works just fine, but the executor is sometimes not able to recycle the runner by the time the second `stat` task was issued.

Fix the issue by waiting for the runner to be recycled by polling the Prometheus metric.

Before, the test would consistently fail when executed with `--config=remote --runs_per_test=100`. After, the test consistently passes under the same conditions.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1200


enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -89,6 +89,8 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
 	preferredSet := map[string]struct{}{}
 	ranked := make([]interfaces.ExecutionNode, 0, len(nodes))
 
+	log.Debugf("Preferred executor IDs for %q: %v", key, preferredNodeIDs)
```
bduffany, line 92: FYI, these debug logs are intentional.

enterprise/server/test/integration/remote_execution/remote_execution_test.go
```diff
@@ -352,13 +359,37 @@ func TestSimpleCommand_RunnerReuse_PoolSelectionViaHeader_RoutesCommandToSameExe
 
 	require.Equal(t, 0, res.ExitCode)
 
-	cmd = rbe.Execute(&repb.Command{
-		Arguments: []string{"stat", "foo.txt"},
-		Platform:  platform,
-	}, opts)
-	res = cmd.Wait()
+	// Run another command that tries to read "foo.txt" written by a previous
```
vadimberezniker, line 362: Instead of blindly retying, what do you think about adding a prometheus metric for successful recycles and waiting on that?

bduffany: SG -- done

vadimberezniker (approved):

PR #1671, opened 2022-03-02
Refactor flag handling to allow config flags to be explicitly defined in code

tempoz:
- app changes for moving config into flags defined in code
- update tests for change to flag handling

`config.go` now registers only missing flags, and thus can no longer register them in its `init` function,
as not all flags can be guaranteed to be defined when config is imported. The code which was pointing
the default FlagSet at the `sharedGeneralConfig` has been repurposed to allow us to define missing flags
or define flags whose values point to a given `generalConfig` for ease of assignment to and from the
default flags. Explicitly defined flags take precedence over explicitly defined config values, which
in turn take precedence over default flag values. On the initial flag parse, the explicitly set flags
are saved so that they may be referenced for that precedence later. The string slices are also saved,
to prevent any future config reconciliation from receiving values contained in a past config. A map
version of the yaml config file is used to distinguish between zero values in the config and absent values
in the config in order to determine the aforementioned precedence.

The `ParseAndReconcileFlagsAndConfig` function is meant to be the new way to get a configurator in most
cases, and it will parse the flags, parse the config file, and use the flag > config > default precedence to
resolve conflicts and render the resulting `Configurator` and the default FlagSet consistent.

The `config_file` flag was moved into `config.go`, since the flag would otherwise need to be parsed before calling `ParseAndReconcileFlagsAndConfig` in order to dereference it to pass it in, complicating the construction of a configurator.

The PR is split into two commits for ease of viewing: first the changes to the production code, and then
the test changes. It was unfortunately not realistic to split across multiple PRs, as the tests would not
pass until all the code changes were complete.

---

**Version bump**: Minor

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -382,10 +382,12 @@ type OrgConfig struct {
 	Domain string `yaml:"domain" usage:"Your organization's email domain. If this is set, only users with email addresses in this domain will be able to register for a BuildBuddy account."`
 }
 
-var sharedGeneralConfig = &generalConfig{}
-var originalStringSlices = make(map[string][]string)
-var defineFlagsOnce sync.Once
-var originalSetFlags = make(map[string]bool)
+var (
+	sharedGeneralConfig = &generalConfig{}
+	originalStringSlices = make(map[string][]string)
+	defineFlagsOnce sync.Once
+	originalSetFlags = make(map[string]bool)
+)
 
 func RegisterAndParseFlags() {
 	defineFlagsOnce.Do(func() {
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -390,7 +390,7 @@ var (
 	defineFlagsOnce      sync.Once
 
 	// Set of the flags that were explicitly set on the command line
-	originalSetFlags     = make(map[string]struct{})
+	originalSetFlags = make(map[string]struct{})
 )
 
 func RegisterAndParseFlags() {
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -404,8 +404,14 @@ func RegisterAndParseFlags() {
 }
 
 // FOR TESTING PURPOSES ONLY!!!
-func GetOriginalSetFlags() map[string]struct{} {
-	return originalSetFlags
+func TestOnlySetFlag(flagName string, set bool) bool {
+	_, wasSet := originalSetFlags[flagName]
+	if set {
+		originalSetFlags[flagName] = struct{}{}
+	} else {
+		delete(originalSetFlags, flagName)
+	}
+	return wasSet
 }
 
 type structSliceFlag struct {
@@ -593,6 +599,9 @@ func (cc *RedisClientConfig) String() string {
 	return cc.SimpleTarget
 }
 
+// After calling this function, the generalConfig in the Configurator and the
+// flags in the default flag set (flag.Commandline) with a corresponding config
+// value will be consistent.
 func (c *Configurator) ReconcileFlagsAndConfig() {
 	flagSet := flag.NewFlagSet("", flag.ContinueOnError)
 	defineFlagsForMembers([]string{}, reflect.ValueOf(c.gc).Elem(), flagSet)
@@ -618,9 +627,10 @@ func (c *Configurator) ReconcileFlagsAndConfig() {
 		_, setInFlags := originalSetFlags[flg.Name]
 		_, presentInYaml := c.presenceMap[flg.Name]
 		// If the flag was set on the command line or there is no value specified in
-		// the config, we use the value defined by the flags. Otherwise (which is to
-		// say, if there was no flag explicitly set on the command line and there is
-		// a value present in the config), use the config value.
+		// the config, we use the value defined by the flags and update the config
+		// with that value. Otherwise (which is to say, if there was no flag
+		// explicitly set on the command line and there is a value present in the
+		// config), use the config value and update the flags with that value.
 		if setInFlags || !presentInYaml {
 			flg.Value.Set(flag.Lookup(flg.Name).Value.String())
 			return
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -383,10 +383,10 @@ type OrgConfig struct {
 }
 
 var (
-	sharedGeneralConfig = &generalConfig{}
+	sharedGeneralConfig  = &generalConfig{}
 	originalStringSlices = make(map[string][]string)
-	defineFlagsOnce sync.Once
-	originalSetFlags = make(map[string]bool)
+	defineFlagsOnce      sync.Once
+	originalSetFlags     = make(map[string]bool)
 )
 
 func RegisterAndParseFlags() {
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -391,7 +391,7 @@ var (
 
 func RegisterAndParseFlags() {
 	defineFlagsOnce.Do(func() {
-		defineFlagsForMembers([]string{}, reflect.ValueOf(&generalConfig{}).Elem(), nil)
+		defineFlagsForMembers([]string{}, reflect.ValueOf(&generalConfig{}).Elem(), flag.CommandLine)
 		flag.Parse()
 		flag.Visit(func(flg *flag.Flag) {
 			originalSetFlags[flg.Name] = struct{}{}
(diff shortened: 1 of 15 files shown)


tempoz:
Thought the `workflow_test` was a solely local failure; the fix is unlikely to substantially change the PR, so feel free to review (or wait if you prefer)

tylerwilliams (changes requested):
See inline comments

server/config/config.go
```diff
@@ -474,19 +544,24 @@ func parseConfig(fileBytes []byte) (*generalConfig, error) {
 		log.Warningf("Unknown fields in config: %s", err)
 	}
 
-	if err := yaml.Unmarshal([]byte(expandedFileBytes), &sharedGeneralConfig); err != nil {
+	gc := &generalConfig{}
+	if err := yaml.Unmarshal([]byte(expandedFileBytes), gc); err != nil {
 		return nil, fmt.Errorf("Error parsing config file: %s", err)
 	}
+	sharedGeneralConfig = gc
```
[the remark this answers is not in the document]
tempoz: Good point.
```suggestion
	// The shared config caches the last config parsed from a file so that a call to
	// `readConfig` with an empty `configFile` can return the cached config.
	sharedGeneralConfig = gc
```

server/config/config.go
```diff
@@ -376,7 +382,25 @@ type OrgConfig struct {
 	Domain string `yaml:"domain" usage:"Your organization's email domain. If this is set, only users with email addresses in this domain will be able to register for a BuildBuddy account."`
 }
 
-var sharedGeneralConfig generalConfig
+var sharedGeneralConfig = &generalConfig{}
+var originalStringSlices = make(map[string][]string)
+var defineFlagsOnce sync.Once
+var originalSetFlags = make(map[string]bool)
```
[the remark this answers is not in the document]
tempoz: Oh, right.
```suggestion
var (
	sharedGeneralConfig = &generalConfig{}
	originalStringSlices = make(map[string][]string)
	defineFlagsOnce sync.Once
	originalSetFlags = make(map[string]bool)
)
```

server/config/config.go
```diff
@@ -459,12 +516,25 @@ func defineFlagsForMembers(parentStructNames []string, T reflect.Value) {
 	}
 }
 
-// Register flags too.
-func init() {
-	defineFlagsForMembers([]string{}, reflect.ValueOf(&sharedGeneralConfig).Elem())
+func populateYamlPresenceMap(yamlMap map[interface{}]interface{}, presenceMap map[string]bool) {
```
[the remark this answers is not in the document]
tempoz: Fixed!

server/config/config.go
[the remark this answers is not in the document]
tempoz: Done!

server/config/config.go
```diff
@@ -409,42 +435,75 @@ func (f *structSliceFlag) Set(value string) error {
...
+			} else if flag.Lookup(fqFieldName) == nil {
+				flag.String(fqFieldName, "", docString)
+			}
 		case reflect.Int:
-			flag.IntVar(f.Addr().Interface().(*int), fqFieldName, int(f.Int()), docString)
+			if configSet != nil {
+				configSet.IntVar(f.Addr().Interface().(*int), fqFieldName, int(f.Int()), docString)
+			} else if flag.Lookup(fqFieldName) == nil {
+				flag.Int(fqFieldName, 0, docString)
+			}
```
tylerwilliams, line 472: flag.CommandLine exposes the "default" flag set (which flag.Int defines, for example).

I think you could significantly simplify this code by removing all the else if cases by assuming a flag set is *always* provided. In the case where you want to set the default flags, you can pass in flag.CommandLine (intead of nil)

tylerwilliams (changes requested):
I noticed one small thing that might simplify this a bit

server/config/config.go
```diff
@@ -376,7 +382,27 @@ type OrgConfig struct {
 	Domain string `yaml:"domain" usage:"Your organization's email domain. If this is set, only users with email addresses in this domain will be able to register for a BuildBuddy account."`
 }
 
-var sharedGeneralConfig generalConfig
+var (
+	sharedGeneralConfig  = &generalConfig{}
+	originalStringSlices = make(map[string][]string)
```
vadimberezniker, line 387: Can you add a short comment describing what originalStringSlices and originalSetFlags contain?

server/config/config.go
```diff
@@ -474,19 +546,26 @@ func parseConfig(fileBytes []byte) (*generalConfig, error) {
...
+		return nil, fmt.Errorf("Error parsing config file: %s", err)
+	}
+	populateYamlPresenceMap(generalConfigMap, presenceMap)
 
-	return &sharedGeneralConfig, nil
+	return gc, nil
 }
 
-func readConfig(fullConfigPath string) (*generalConfig, error) {
+func readConfig(fullConfigPath string, presenceMap map[string]struct{}) (*generalConfig, error) {
```
vadimberezniker, line 566: Is it necessary to pass the presence map as an out parameter? Could you instantiate the empty map inside the function and return it instead?

server/config/config.go
```diff
@@ -459,12 +518,25 @@ func defineFlagsForMembers(parentStructNames []string, T reflect.Value) {
...
+			}
+			return
+		}
+		presenceMap[strings.Join(key, ".")] = struct{}{}
+	}
+	check([]string{}, yamlMap)
 }
 
-func parseConfig(fileBytes []byte) (*generalConfig, error) {
+func parseConfig(fileBytes []byte, presenceMap map[string]struct{}) (*generalConfig, error) {
```
vadimberezniker, line 539: Is it necessary to pass the presence map as an out parameter? Could you instantiate the empty map inside the function and return it instead?

server/config/config.go
```diff
@@ -541,6 +638,38 @@ func (cc *RedisClientConfig) String() string {
...
+				// aforementioned append operation, as opposed to setting anything.
+				flg.Value.Set(strings.Join(originalSlice, ","))
+				*flagSlice = *configSlice
+				return
+			}
+			log.Warningf("yaml defines %s as []string, but flags do not.", flg.Name)
+		}
+		_, setInFlags := originalSetFlags[flg.Name]
+		_, presentInYaml := c.presenceMap[flg.Name]
+		if setInFlags || !presentInYaml {
```
vadimberezniker, line 665: Can you add a comment here describing this logic since it's not straightforward.

vadimberezniker (changes requested):

server/config/config.go
```diff
@@ -409,124 +439,146 @@ func (f *structSliceFlag) Set(value string) error {
...
-
-func NewConfigurator(configFilePath string) (*Configurator, error) {
-	conf, err := readConfig(configFilePath)
-	if err != nil {
-		return nil, err
+func ParseAndReconcileFlagsAndConfig(configFilePath string) (*Configurator, error) {
+	RegisterAndParseFlags()
+	if configFilePath == "" {
+		configFilePath = *configFile
 	}
```
tempoz, line 575: I moved the use of the `config_file` flag into `ParseAndReconcileFlagsAndConfig`, so now a caller can still call `NewConfigurator` with an empty string and get a configurator using the previously parsed config, regardless of the status of the `config_file` flag. Additionally, if we wanted, we could move this function and the `config_file` flag to their own package and then only define the `config_file` flag when we wanted to use this function specifically, rather than every time we import the config.

tylerwilliams (approved):
LGTM

server/config/config.go
```diff
@@ -541,6 +593,42 @@ func (cc *RedisClientConfig) String() string {
...
+				return
+			}
+			log.Warningf("yaml defines %s as []string, but flags do not.", flg.Name)
+		}
+		_, setInFlags := originalSetFlags[flg.Name]
+		_, presentInYaml := c.presenceMap[flg.Name]
+		// If the flag was set on the command line or there is no value specified in
+		// the config, we use the value defined by the flags. Otherwise (which is to
+		// say, if there was no flag explicitly set on the command line and there is
+		// a value present in the config), use the config value.
```
vadimberezniker, line 623: I think it's important to mention that setting the flag here indirectly updates the value within the config struct.

server/util/testing/flags/flags.go
```diff
@@ -3,18 +3,27 @@ package flags
...
 // Set a flag value and register a cleanup function to restore the flag
 // to its original value after the given test is complete.
 func Set(t testing.TB, name, value string) {
+	config.RegisterAndParseFlags()
 	f := flag.Lookup(name)
 	if f == nil {
 		t.Fatalf("Undefined flag: %s", name)
 	}
 	original := f.Value.String()
+	originalSetMap := config.GetOriginalSetFlags()
```
vadimberezniker, line 19: nit: it's bad hygiene to return a mutable structure owned by a different object and manipulate it outside of that object. 
i think it would be slightly better to have explicit `TestOnlySetFlag` and `TestOnlyUnsetFlag` methods to do this manipulation. `TestOnlySetFlag` could return a bool to tell you whether it's a new flag.

vadimberezniker (approved):

PR #1672, opened 2022-03-03
Firecracker: Add support for hot-swapping workspaces

bduffany:
- Split the mounted RW drive into 2 drives: `scratchfs` (persists across container reuse, not hot-swapped for now) and `workspacefs` (hot-swappable)
- Add a `SyncWorkspace` function to allow hot-swapping the workspace device

The `SyncWorkspace` function is not actually called yet because it would wind up doing unnecessary extra work in the hosted bazel case, since hosted bazel actions have no explicit inputs or outputs that they care about. Specifically, we'd unnecessarily nuke the workspace and then re-create an ext4 image with the same contents. Workspaces can be rather large since they can contain an entire git repo, so this would likely slow down hosted bazel too much.

PRs to follow:
* Update the CI runner to write its contents to the scratch disk (at `/root`) instead of the workspace disk.
* Add a config option to specify the `EstimatedFreeDisk` fraction that gets allocated to the scratch disk vs. workspace disk.
* Call `SyncWorkspace` after `DownloadInputs` to get firecracker working for regular bazel actions that want to use runner recycling (specifically, docker-in-firecracker test actions).
* Once that's done, runner-recycling will work for docker-in-firecracker actions, speeding them up significantly since they can reuse the docker image cache from previous tasks and not have to restart the Docker daemon. We'll be able to set `recycle-runner=true` on our Docker workflow actions to make Docker test workflows faster.

Future optimizations to be made:
* Reuse the vmexec client instead of dialing it every time

TODO before submitting:
- [x] Rebuild `initrd.cpio`

---

**Version bump**: Patch 

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -162,6 +162,7 @@ func openFile(ctx context.Context, env environment.Env, fileName string) (io.Rea
 
 // putFileIntoDir finds "fileName" on the local filesystem, in runfiles, or
 // in the bundle. It then puts that file into destdir (via hardlink or copying)
+// and returns a path to the file in the new location. Files are written in
 // a content-addressable-storage-based location, so when files are updated they
 // will be put into new paths.
 func putFileIntoDir(ctx context.Context, env environment.Env, fileName, destDir string, mode fs.FileMode) (string, error) {
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -711,6 +711,9 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
 // createWorkspaceImage creates a new ext4 image from the action working dir
 // and returns the chroot-relative path to the created image.
 func (c *FirecrackerContainer) createWorkspaceImage(ctx context.Context, workspacePath string) (string, error) {
+	ctx, span := tracing.StartSpan(ctx)
+	defer span.End()
+
 	c.workspaceGeneration++
 	relativePath := fmt.Sprintf("%s.gen_%d.ext4", workspaceFSName, c.workspaceGeneration)
 	hostPath := filepath.Join(c.getChroot(), relativePath)
@@ -729,7 +732,9 @@ func (c *FirecrackerContainer) createWorkspaceImage(ctx context.Context, workspa
 // container, updates the workspace block device to an ext4 image pointed to
 // by chrootRelativeImagePath, and re-mounts the drive.
 func (c *FirecrackerContainer) hotSwapWorkspace(ctx context.Context, execClient vmxpb.ExecClient, chrootRelativeImagePath string) error {
-	log.Infof("Hot-swapping guest workspace drive")
+	ctx, span := tracing.StartSpan(ctx)
+	defer span.End()
+
 	if _, err := execClient.UnmountWorkspace(ctx, &vmxpb.UnmountWorkspaceRequest{}); err != nil {
 		return status.WrapError(err, "Failed to unmount workspace")
 	}
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -91,6 +91,12 @@ const (
 	workspaceFSName  = "workspacefs.ext4"
 	workspaceDriveID = "workspacefs"
 
+	// The scratchfs image name and drive ID.
+	scratchFSName  = "scratchfs.ext4"
+	scratchDriveID = "scratchfs"
+	// TODO(bduffany): Make this configurable
+	scratchDiskSizeBytes = 2e9
+
 	// The containerfs drive ID.
 	containerFSName  = "containerfs.ext4"
 	containerDriveID = "containerfs"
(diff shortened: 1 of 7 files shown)


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -31,6 +31,10 @@ type LoadSnapshotOptions struct {
 	InitrdImagePath     string
 	ContainerFSPath     string
 
+	// This field is optional -- a snapshot may have a scratch filesystem
+	// attached or it may have on attached at runtime.
```
tylerwilliams, line 35: s/on/one/g

enterprise/server/vmexec/vmexec.go
```diff
@@ -76,6 +76,20 @@ func (x *execServer) Initialize(ctx context.Context, req *vmxpb.InitializeReques
 	return &vmxpb.InitializeResponse{}, nil
 }
 
+func (x *execServer) UnmountWorkspace(ctx context.Context, req *vmxpb.UnmountWorkspaceRequest) (*vmxpb.UnmountWorkspaceResponse, error) {
```
tylerwilliams, line 79: nit: would be nice to send the vmexec changes in a separate cl

tylerwilliams (commented):
still looking at this and gotta run get food but a few quick comments

enterprise/server/vmexec/vmexec.go
bduffany: Done: https://github.com/buildbuddy-io/buildbuddy/pull/1674

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -694,6 +698,49 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
 	if err != nil {
 		return status.WrapError(err, "Failed to initialize firecracker VM exec client")
 	}
+
+	if workspaceDirOverride != "" {
+		// If the snapshot is being loaded with a different workspaceFS
+		// then handle that now.
+		if err := c.hotSwapWorkspace(ctx, execClient, wsImgRelPath); err != nil {
```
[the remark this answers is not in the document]
bduffany: good call -- done

enterprise/server/remote_execution/snaploader/snaploader.go

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -700,6 +692,53 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
...
+	}
+	c.workspaceFSPath = hostPath
+	return relativePath, nil
+}
+
+// hotSwapWorkspace unmounts the workspace drive from a running firecracker
+// container, updates the workspace block device to an ext4 image pointed to
+// by chrootRelativeImagePath, and re-mounts the drive.
+func (c *FirecrackerContainer) hotSwapWorkspace(ctx context.Context, execClient vmxpb.ExecClient, chrootRelativeImagePath string) error {
+	log.Infof("Hot-swapping guest workspace drive")
```
vadimberezniker, line 732: Did you mean to leave this in? w/o additional container instance information it's not a very useful log line

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -700,6 +692,53 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
...
+// and returns the chroot-relative path to the created image.
+func (c *FirecrackerContainer) createWorkspaceImage(ctx context.Context, workspacePath string) (string, error) {
+	c.workspaceGeneration++
+	relativePath := fmt.Sprintf("%s.gen_%d.ext4", workspaceFSName, c.workspaceGeneration)
+	hostPath := filepath.Join(c.getChroot(), relativePath)
+	sizeBytes, err := disk.DirSize(workspacePath)
+	if err != nil {
+		return "", err
+	}
+	if err := ext4.DirectoryToImage(ctx, workspacePath, hostPath, sizeBytes+c.constants.DiskSlackSpaceMB*1e6); err != nil {
```
vadimberezniker, line 721: I'm assuming you will move the configurable "slack space" to the scratch disk with the ci_runner changes to use the scratch disk?

vadimberezniker (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: Yep, that's the plan :)

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: Woops, removed. Also added tracing

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -700,6 +692,58 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
...
+	defer span.End()
+
+	if _, err := execClient.UnmountWorkspace(ctx, &vmxpb.UnmountWorkspaceRequest{}); err != nil {
+		return status.WrapError(err, "Failed to unmount workspace")
+	}
+	if err := c.machine.UpdateGuestDrive(ctx, workspaceDriveID, chrootRelativeImagePath); err != nil {
+		return status.InternalErrorf("error updating workspace drive attached to snapshot: %s", err)
+	}
+	if _, err := execClient.MountWorkspace(ctx, &vmxpb.MountWorkspaceRequest{}); err != nil {
+		return status.WrapError(err, "Failed to remount workspace after update")
```
tylerwilliams, line 745: We are terrible at this, and I plan to audit it everywhere at some point, but:

errors message should begin with a lower case letter (because they can be wrapped by other errors higher up)
log lines can begin with an upper case letter (because they are immediately logged and not wrapped)

tylerwilliams (approved):
LGTM

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: Makes sense, I'll keep this in mind for the future. Fixed the capitalization in the errors added by this PR.

PR #1673, opened 2022-03-03
Firecracker: Ensure containers are only removed once

bduffany:
This is for 2 reasons:
- When adding a container to the runner pool, if `Pause` fails, then the pool will call `Remove`. Since `FirecrackerContainer.Pause` internally calls `FirecrackerContainer.Remove`, `Remove` winds up getting called twice.
- When tests fail between `Create` and `Remove`, the networks are left hanging around on the machine, causing the next test execution to fail. Now, we can register `Remove` as a cleanup func just after calling `Create`, without worrying about whether the test has already called `Pause` or `Remove` by the end of the test.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1674, opened 2022-03-03
Firecracker: Add vmexec RPCs for mounting/unmounting a workspace

bduffany:
These will be used for workspace hot-swap support, which is needed to get firecracker working for normal bazel actions that want to use runner recycling.

---

**Version bump**: Patch 

### DIFF
--- a/enterprise/server/vmexec/vmexec.go
+++ b/enterprise/server/vmexec/vmexec.go
@@ -76,6 +76,20 @@ func (x *execServer) Initialize(ctx context.Context, req *vmxpb.InitializeReques
 	return &vmxpb.InitializeResponse{}, nil
 }
 
+func (x *execServer) UnmountWorkspace(ctx context.Context, req *vmxpb.UnmountWorkspaceRequest) (*vmxpb.UnmountWorkspaceResponse, error) {
+	if err := syscall.Unmount("/workspace", 0); err != nil {
+		return nil, status.InternalErrorf("unmount failed: %s", err)
+	}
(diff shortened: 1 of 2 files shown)


proto/vmexec.proto
```diff
@@ -46,9 +46,18 @@ message InitializeResponse {
   // This page intentionally left empty.
 }
 
+message UnmountWorkspaceRequest {}
```
vadimberezniker, line 49: Can you add brief docs for these APIs?

vadimberezniker (approved):

enterprise/server/vmexec/vmexec.go
```diff
@@ -17,6 +17,14 @@ import (
 	vmxpb "github.com/buildbuddy-io/buildbuddy/proto/vmexec"
 )
 
+const (
+	// workspaceDevice is the path to the hot-swappable workspace block device.
+	workspaceDevice = "/dev/vdc"
+	// workspaceMountPath is the path where the hot-swappable workspace block
+	// device is mounted.
+	workspaceMountPath = "/workspace"
```
tylerwilliams, line 25: If these need to match what's in init, I would note that and link to the source.

proto/vmexec.proto
tylerwilliams: +1

bduffany: Done, and moved the service to the top of the file so that the docs come first.

enterprise/server/vmexec/vmexec.go
bduffany: Done

PR #1675, opened 2022-03-03
Remove storage global options for podman

luluz66:
Once we mounted /var/lib/containers, we don't need to use /fuse/overlay
mount program.


tylerwilliams (approved):
LGTM

PR #1677, opened 2022-03-04
Enable network for podman

luluz66:
Use the default network mode: bridge for podman

Tested with a sh_test that curl www.google.com


tylerwilliams (approved):
LGTM

Does this change performance any?

luluz66:
> LGTM
> 
> Does this change performance any?

I haven't tested. I could test building buildbuddy after it's pushed to dev.

PR #1678, opened 2022-03-05
Implement podman lifecycle methods

luluz66:
Also remove two global options "--events-backend=file", "--cgroup-manager=cgroupfs". They are no longer needed in podman 3.4.

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -123,7 +123,7 @@ func (c *podmanCommandContainer) Create(ctx context.Context, workDir string) err
 }
 
 func (c *podmanCommandContainer) Exec(ctx context.Context, cmd *repb.Command, stdin io.Reader, stdout io.Writer) *interfaces.CommandResult {
-	podmanRunArgs := make([]string, 0, len(2*cmd.GetEnvironmentVariables())+len(cmd.Arguments)+1)
+	podmanRunArgs := make([]string, 0, 2*len(cmd.GetEnvironmentVariables())+len(cmd.Arguments)+1)
 	for _, envVar := range cmd.GetEnvironmentVariables() {
 		podmanRunArgs = append(podmanRunArgs, "--env", fmt.Sprintf("%s=%s", envVar.GetName(), envVar.GetValue()))
 	}
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -123,7 +123,7 @@ func (c *podmanCommandContainer) Create(ctx context.Context, workDir string) err
 }
 
 func (c *podmanCommandContainer) Exec(ctx context.Context, cmd *repb.Command, stdin io.Reader, stdout io.Writer) *interfaces.CommandResult {
-	podmanRunArgs := make([]string, 0, len(cmd.GetEnvironmentVariables())+len(cmd.Arguments)+1)
+	podmanRunArgs := make([]string, 0, len(2*cmd.GetEnvironmentVariables())+len(cmd.Arguments)+1)
 	for _, envVar := range cmd.GetEnvironmentVariables() {
 		podmanRunArgs = append(podmanRunArgs, "--env", fmt.Sprintf("%s=%s", envVar.GetName(), envVar.GetValue()))
 	}
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -33,8 +33,9 @@ type podmanCommandContainer struct {
 
 	image     string
 	buildRoot string
-	// workDir is the path to the workspace directory mounted to the container.
-	workDir string
+
+	// name is the container name.
+	name string
 }
 
 func NewPodmanCommandContainer(env environment.Env, imageCacheAuth *container.ImageCacheAuthenticator, image, buildRoot string) container.CommandContainer {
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -93,17 +102,37 @@ func (c *podmanCommandContainer) Run(ctx context.Context, command *repb.Command,
...
+		return status.UnavailableErrorf("failed to create container: %s", err)
+	}
+
+	startResult := runPodman(ctx, "start", nil, nil, c.name)
+	return startResult.Error
 }
 
 func (c *podmanCommandContainer) Exec(ctx context.Context, cmd *repb.Command, stdin io.Reader, stdout io.Writer) *interfaces.CommandResult {
-	return runPodman(ctx, "run" /*workDir=*/, "", stdin, stdout, cmd.Arguments...)
+	podmanRunArgs := []string{}
```
tylerwilliams, line 124: you can make this the right capacity ahead of time and avoid extra allocation work:

podmanRunArgs := make([]string, 0, len(cmd.GetEnvironmentVariables()))

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -116,31 +145,40 @@ func (c *podmanCommandContainer) IsImageCached(ctx context.Context) (bool, error
...
+}
+
+func (c *podmanCommandContainer) Pause(ctx context.Context) error {
+	res := runPodman(ctx, "pause", nil, nil, c.name)
+	return res.Error
+}
+
+func (c *podmanCommandContainer) Unpause(ctx context.Context) error {
+	res := runPodman(ctx, "unpause", nil, nil, c.name)
+	return res.Error
```
tylerwilliams, line 167: nit: would be nice to micro-comment the nil, nils in here

runPodman(ctx, "unpause", nil /*=stdin*/, nil/*=stdout*/, c.name)

tylerwilliams (approved):
LGTM (2 nits)

This looks pretty good to me -- maybe have Brandon give it a look too? Another way to improve it might be to add a test that pauses a container and resumes it, or uses the stdin/stdout pipes.

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -93,17 +102,37 @@ func (c *podmanCommandContainer) Run(ctx context.Context, command *repb.Command,
...
+	}
+	podmanRunArgs = append(podmanRunArgs, c.name)
+	podmanRunArgs = append(podmanRunArgs, cmd.Arguments...)
+	return runPodman(ctx, "exec", stdin, stdout, podmanRunArgs...)
 }
 
 func (c *podmanCommandContainer) IsImageCached(ctx context.Context) (bool, error) {
 	// Try to avoid the `pull` command which results in a network roundtrip.
-	listResult := runPodman(ctx, "images", "", nil, nil, "--filter=reference="+c.image, "--format={{.ID}}")
+	listResult := runPodman(ctx, "images", nil, nil, "--filter=reference="+c.image, "--format={{.ID}}")
```
bduffany, line 135: nit: I think the `--filter` here does a string-contains check, so if there is a cached image tagged `v1.1` and we check whether `v1` is cached, it will incorrectly return true -- might want to split the output lines into columns and then return whether at least one line matches an `==` check on the image ref column.

bduffany (approved):
Looks great, just a couple of nits

enterprise/server/remote_execution/containers/podman/podman.go
luluz66: Done. I actually use podman image inspect. inspect command can be used to inspect containers, volumes, network, and pods.

enterprise/server/remote_execution/containers/podman/podman.go
luluz66: Done

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -93,17 +102,37 @@ func (c *podmanCommandContainer) Run(ctx context.Context, command *repb.Command,
...
 func (c *podmanCommandContainer) Create(ctx context.Context, workDir string) error {
-	c.workDir = workDir
-	return nil
+	containerName, err := generateContainerName()
+	if err != nil {
+		return status.UnavailableErrorf("failed to generate podman container name: %s", err)
+	}
+	c.name = containerName
+
+	podmanRunArgs := c.getPodmanRunArgs(workDir, false /*=autoRemove*/)
```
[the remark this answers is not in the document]
luluz66: Done

enterprise/server/remote_execution/containers/podman/podman.go
luluz66: Done

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -121,7 +123,7 @@ func (c *podmanCommandContainer) Create(ctx context.Context, workDir string) err
 }
 
 func (c *podmanCommandContainer) Exec(ctx context.Context, cmd *repb.Command, stdin io.Reader, stdout io.Writer) *interfaces.CommandResult {
-	podmanRunArgs := []string{}
+	podmanRunArgs := make([]string, 0, len(cmd.GetEnvironmentVariables())+len(cmd.Arguments)+1)
```
bduffany, line 126: ubernit: `2 * len(cmd.GetEnvironmentVariables())` here, or use the combined form like `fmt.Sprintf("--env=%s=%s")`

bduffany (approved):

enterprise/server/remote_execution/containers/podman/podman.go
luluz66: Done

PR #1679, opened 2022-03-07
Firecracker: Fix some issues with runner recycling

bduffany:
Fixes:
* Fix VM exec server timeout due to forgetting to close the gRPC client connection from a previous VM run. This was introduced in #1672. I suspect that gRPC was trying to reuse the previous connection, which was broken since the VM was paused.
* Fix incorrect `workspaceFSPath` introduced in #1672. Except when getting the initial firecracker config, the workspace FS path should always point to the path in the chroot dir. That PR introduced a change where `workspaceFSPath` pointed to the path under the temp dir `fc-container-*`, which caused a few issues related to cleanup, copying outputs, etc.

Cleanup:
* Clean up some unnecessary/error-prone state related to paths -- workspace FS path in chroot can be computed from `c.getChroot()` and `c.workspaceGeneration`, and scratch FS path can be computed from `c.getChroot()`
* Remove the logic for nulling out `workspaceFSPath` and restoring it after `SaveSnapshot` -- this appears vestigial

Testing/debugging:
* Disable `AllowSnapshotStart` tests; these are flaky because the snapshot gets clobbered due to the snapshot artifacts getting mutably hard linked into the file cache, rather than storing an immutable copy. Additionally, we don't wait for the init setup to complete before taking the snapshot. With these 2 issues combined, what winds up happening is that the first snapshot can be taken before `goinit` writes `/vfs`, then when resuming from snapshot, `goinit` will try to write `/vfs` but that path is already written, since the snapshot disk image has been clobbered by the first run.
* Extract `busyboxImage` constant in tests, since it's convenient to swap out this image when getting rate limited by docker hub. I didn't change this image to the GCR image permanently because the `ping` test fails with the GCR image for some reason.
* Extract `minMemSizeMB` constant in tests, since this is useful to rule out mem size as the reason for a test failure
* Add a `tree` util in tests which is very helpful for visualizing the state of the chroot at various points
* Add a `String()` method to the snaploader manifest, which also aids in debugging snapshot issues

---

**Version bump**: Patch 

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -286,10 +286,9 @@ type FirecrackerContainer struct {
 	constants           Constants
 	containerImage      string // the OCI container image. ex "alpine:latest"
 	actionWorkingDir    string // the action directory with inputs / outputs
-	workspaceFSPath     string // the path to the workspace ext4 image
 	workspaceGeneration int    // the number of times the workspace has been re-mounted into the guest VM
-	scratchFSPath       string // the path fo the scratch ext4 image
 	containerFSPath     string // the path to the container ext4 image
+	tempDir             string // path for writing disk images before the chroot is created
 
 	rmOnce *sync.Once
 	rmErr  error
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1508,7 +1516,7 @@ func (c *FirecrackerContainer) remove(ctx context.Context) error {
 		log.Errorf("Error cleaning up networking: %s", err)
 		lastErr = err
 	}
-	if err := os.RemoveAll(filepath.Dir(c.workspaceFSPath)); err != nil {
+	if err := os.RemoveAll(c.tempDir); err != nil {
```
vadimberezniker, line 1519: paranoia: check that c.tempDir is not ""?

vadimberezniker (approved):

PR #1680, opened 2022-03-07
Firecracker: Move slack space allocation to scratch disk

bduffany:
Instead of hard-coding the scratch disk size as 2GB, hard-code the workspace size as 2GB and allocate the EstimatedFreeDiskBytes to scratch space instead. This makes sense both for hosted bazel (once we move its working directory to the scratch disk) and for docker-in-firecracker tests (which need most of the disk space for cached image layers).

---

**Version bump**: Patch 

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/containeropts.go
+++ b/enterprise/server/remote_execution/containers/firecracker/containeropts.go
@@ -24,10 +24,6 @@ type ContainerOpts struct {
 	// The amount of RAM, in MB, to allocate to this VM.
 	MemSizeMB int64
 
-	// The amount of slack space to allocate to the workspace drive (for writing
-	// action outputs, test temp files, etc.)
-	WorkspaceDiskSlackSpaceMB int64
-
 	// The size of the scratch disk to allocate (for writing files anywhere
 	// outside of the workspace directory, such as /tmp or ~/.cache).
 	ScratchDiskSizeMB int64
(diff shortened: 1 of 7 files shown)


bduffany:
> Does the workspace slack need to be adjustable or could we get away with using a fixed value?

Seems like only a matter of time before someone hits the limit and we have to implement this anyway (either that, or we provision a very large workspace by default that users are unlikely to hit the limits of -- but then I suspect that would not be good for filecache health even if we up the filecache size to 1TB)

vadimberezniker:
> > Does the workspace slack need to be adjustable or could we get away with using a fixed value?
> 
> Seems like only a matter of time before someone hits the limit and we have to implement this anyway (either that, or we provision a very large workspace by default that users are unlikely to hit the limits of -- but then I suspect that would not be good for filecache health even if we up the filecache size to 1TB)
> 
> This might be a bit too speculative though, wdyt?

Since we don't really know which way things are going to go, it seems simpler to implement a fixed slack amount for the workspace and revisit it later if that turns out to be a problem. Let me know if you feel strongly about making it adjustable from the start.

bduffany:
> Since we don't really know which way things are going to go, it seems simpler to implement a fixed slack amount for the workspace and revisit it later if that turns out to be a problem.

SG, hard coding it as 2GB for now.

vadimberezniker (approved):

PR #1683, opened 2022-03-07
Firecracker: Sync workspace after downloading inputs

bduffany:
This allows firecracker actions that deal with inputs/outputs to work properly when using runner recyling.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1188

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -246,12 +246,9 @@ func (r *CommandRunner) Run(ctx context.Context) *interfaces.CommandResult {
 			return commandutil.ErrorResult(err)
 		}
 		r.state = ready
+		break
 	case ready:
-		// When re-using a container, need to synchronize workspace contents before
-		// executing.
-		if err := r.syncWorkspace(ctx); err != nil {
-			return commandutil.ErrorResult(err)
-		}
+		break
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -262,6 +265,26 @@ func (r *CommandRunner) Run(ctx context.Context) *interfaces.CommandResult {
...
+//
+// This is also an NOP for configurations in which the runner shares the same
+// view of the workspace as the host, such as the workspace is mounted to a
+// container or made directly accessible to a bare runner.
+func (r *CommandRunner) syncWorkspace(ctx context.Context) error {
+	if r.VFS != nil {
+		return nil
+	}
+	if fc, ok := r.Container.Delegate.(*firecracker.FirecrackerContainer); ok {
+		return fc.SyncWorkspace(ctx)
```
vadimberezniker, line 283: Would it work if we always called `SyncWorkspace` inside of firecracker `Exec` func? I think it would work if we changed `Create` to create a VM with a dummy workspace fs and then swap it out when `Exec` is called.

bduffany: Good call -- done

PR #1684, opened 2022-03-08
Implement ForceRoot option in podman

luluz66:
Add unit tests

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman_test.go
+++ b/enterprise/server/remote_execution/containers/podman/podman_test.go
@@ -165,14 +165,17 @@ func TestForceRoot(t *testing.T) {
 	tests := []struct {
 		desc      string
 		forceRoot bool
+		wantUID   int
 	}{
 		{
 			desc:      "forceRoot",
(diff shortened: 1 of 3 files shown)


bduffany (approved):

enterprise/server/remote_execution/containers/podman/podman_test.go
```diff
@@ -142,5 +145,47 @@ func TestIsImageCached(t *testing.T) {
...
+	cacheAuth := container.NewImageCacheAuthenticator(container.ImageCacheAuthenticatorOpts{})
+	image := "gcr.io/flame-public/test-nonroot:test-enterprise-v1.5.4"
+
+	cmd := &repb.Command{
+		Arguments: []string{"id", "-u"},
+	}
 
+	tests := []struct {
+		desc      string
+		forceRoot bool
```
bduffany, line 167: nit: you could add the expected UID here to avoid conditional logic in the test

luluz66: Done

PR #1686, opened 2022-03-08
Add filecache last evicted age metric

bduffany:
This will help us monitor the health of the filecache as we roll out firecracker, which adds large artifacts to the filecache and depends on the filecache for restoring VMs from snapshot.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1214


tylerwilliams (approved):
LGTM

I like the simplicity of this -- if it were me I probably would have complicated things and used the file stat ctime instead of modifying `entry`.

PR #1687, opened 2022-03-08
Add filecache file size metric

bduffany:
This will allow us to understand whether an unhealthy file cache is due to a large amount of small/moderately sized files being added, vs. a small/moderate amount of very large files.

---

**Version bump**: Patch

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -575,10 +575,10 @@ var (
 		Help:      "Age of the last entry evicted from the executor's local file cache (relative to when it was added to the cache), in **microseconds**.",
 	})
 
-	FileCacheFileSizeBytes = promauto.NewHistogram(prometheus.HistogramOpts{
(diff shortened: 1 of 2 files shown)


server/metrics/metrics.go
```diff
@@ -563,6 +563,14 @@ var (
 		Help:      "Age of the last entry evicted from the executor's local file cache (relative to when it was added to the cache), in **microseconds**.",
 	})
 
+	FileCacheFileSizeBytes = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: bbNamespace,
+		Subsystem: "remote_execution",
+		Name:      "file_cache_file_size_bytes",
```
vadimberezniker, line 569: Should we call the metric & variable file_cache_added_file_size_bytes to be more clear that this represents files being added?

vadimberezniker (approved):

server/metrics/metrics.go
bduffany: Good call -- done

PR #1690, opened 2022-03-08
add a default workload support type

luluz66:


server/config/config.go
```diff
@@ -297,6 +297,7 @@ type ExecutorConfig struct {
 	DockerSiblingContainers       bool                      `yaml:"docker_sibling_containers" usage:"If set, mount the configured Docker socket to containers spawned for each action, to enable Docker-out-of-Docker (DooD). Takes effect only if docker_socket is also set. Should not be set by executors that can run untrusted code."`
 	DockerInheritUserIDs          bool                      `yaml:"docker_inherit_user_ids" usage:"If set, run docker containers using the same uid and gid as the user running the executor process."`
 	DefaultXcodeVersion           string                    `yaml:"default_xcode_version" usage:"Sets the default Xcode version number to use if an action doesn't specify one. If not set, /Applications/Xcode.app/ is used."`
+	DefaultIsolationType          string                    `yaml:"default_isolation_type" usage:"The default workload isolation type when no type is specified in an action. If not set,we use the first of the following that is set: docker, firecracker, podman, barerunner"`
```
brentleyjones, line 300:
```suggestion
	DefaultIsolationType          string                    `yaml:"default_isolation_type" usage:"The default workload isolation type when no type is specified in an action. If not set, we use the first of the following that is set: docker, firecracker, podman, or barerunner"`
```

tylerwilliams (approved):
LGTM

PR #1692, opened 2022-03-09
Treat docker exec exit code 137 (killed) as an incomplete execution

bduffany:
The docker exec result contains error code 137 when SIGTERM is sent to the executor on shutdown, which (indirectly) causes the exec process to be killed.

We'd previously treat this case as though the action ran to completion, uploading an ActionResult with exit code 137. We now return `Unavailable` instead, indicating that the action should be retried. Once #1691 is submitted, this will short-circuit properly and the action will actually be retried as intended.

Podman does not seem to have this problem, at least with our current CLI-based implementation. Not sure if things would change if we were to switch to the Go bindings.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1183

### DIFF
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -32,7 +32,11 @@ import (
 	gstatus "google.golang.org/grpc/status"
 )
 
-var (
+const (
+	// killedExitCode is the exit code returned when a docker exec process exits
+	// due to receiving SIGKILL.
+	killedExitCode = 137
+
 	dockerDaemonErrorCode        = 125
 	containerFinalizationTimeout = 10 * time.Second
 	defaultDockerUlimit          = int64(65535)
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -401,6 +405,15 @@ func (r *dockerCommandContainer) Exec(ctx context.Context, command *repb.Command
 		res = r.exec(ctx, command, stdin, stdout)
 		return res.Error
 	})
+	// Translate killedExitCode and populate the error field to match the
+	// commandutil.Run semantics.
+	// TODO(bduffany): Find a way to detect whether the exec process was killed
```
vadimberezniker, line 410: Could we set a variable in `Remove` to indicate that we are killing the container and then check that value here to limit this to only the shutdown scenario?

PR #1697, opened 2022-03-10
Don't fail if a valid JWT doesn't contain (optional) at_hash

siggisim:
At this point the JWT has already been validated.

We undergo the extra step of making sure the token's at_hash matches - but the at_hash parameter is optional and not set by all auth providers: https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken

The go-oidc library mentions: "It is the caller's responsibility to ensure that the optional access token hash is present for the ID token before calling this method":
https://github.com/coreos/go-oidc/blob/d42db69c79f2fa664fd4156e939bf27bba0d2f68/oidc/oidc.go#L371

Also improves error logging to help debug situations like these in the future.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -307,6 +307,11 @@ func (a *oidcAuthenticator) checkAccessToken(ctx context.Context, jwt, accessTok
 	if err != nil {
 		return err
 	}
+	// at_hash is optional:
+	// https://github.com/coreos/go-oidc/blob/d42db69c79f2fa664fd4156e939bf27bba0d2f68/oidc/oidc.go#L373
+	if validToken.AccessTokenHash == "" {
+		return nil
+	}
 	return validToken.VerifyAccessToken(accessToken)
 }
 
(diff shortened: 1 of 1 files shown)


bduffany (approved):

enterprise/server/auth/auth.go
```diff
@@ -833,11 +838,11 @@ func (a *OpenIDAuthenticator) authenticateUser(w http.ResponseWriter, r *http.Re
 	// If the session is not found, bail.
 	sesh, err := authDB.ReadSession(ctx, sessionID)
 	if err != nil {
-		return nil, ut, status.PermissionDeniedError(loggedOutMsg)
+		return nil, ut, status.PermissionDeniedErrorf("%s: session not found: %s", loggedOutMsg, err.Error())
```
vadimberezniker, line 841: I'm a little wary about returning the raw error message in auth to the user as it's hard to predict what information we are exposing to a potential attacker.

vadimberezniker (approved):

enterprise/server/auth/auth.go
siggisim: Good point, I've switched to returning a static error and debug logging the actual error.

This gives us the best of both worlds in that we know which error is being hit, and debug logging can be enabled to see the error message.

enterprise/server/auth/auth.go
```diff
@@ -833,11 +838,13 @@ func (a *OpenIDAuthenticator) authenticateUser(w http.ResponseWriter, r *http.Re
 	// If the session is not found, bail.
 	sesh, err := authDB.ReadSession(ctx, sessionID)
 	if err != nil {
-		return nil, ut, status.PermissionDeniedError(loggedOutMsg)
+		log.Debugf("Session not found: %s", err.Error())
```
vadimberezniker, line 841: nit: %s understands errors so there's no need to call .Error()

siggisim: Done

PR #1700, opened 2022-03-11
Lower internal jwt lifespan from 24h=>6h and always use that

siggisim:
Should decouple max execution duration from --remote_timeout when using --experimental_remote_execution_keepalive

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

vadimberezniker (approved):

PR #1702, opened 2022-03-11
Retry failed workflow invocation streams

bduffany:
This fixes an issue that a workflow invocation can be disconnected if the build event stream client is connected to a BB app that is shut down while the stream is in progress and the workflow invocation takes longer than the grace period.

Tested locally.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1183

### DIFF
--- a/enterprise/server/build_event_publisher/build_event_publisher.go
+++ b/enterprise/server/build_event_publisher/build_event_publisher.go
@@ -210,10 +210,15 @@ func (b *EventBuffer) Subscribe() (<-chan *pepb.OrderedBuildEvent, func()) {
 			// Send all events which haven't been sent yet
 			for i < len(buffer) {
 				buildEvent := buffer[i]
-				events <- &pepb.OrderedBuildEvent{
+				obe := &pepb.OrderedBuildEvent{
 					StreamId:       b.streamID,
 					SequenceNumber: int64(i) + 1,
 					Event:          buildEvent,
+				}
+				select {
+				case <-cancel:
+					return
+				case events <- obe:
 				}
 				if _, ok := buildEvent.Event.(*bepb.BuildEvent_ComponentStreamFinished); ok {
 					return
--- a/enterprise/server/build_event_publisher/build_event_publisher.go
+++ b/enterprise/server/build_event_publisher/build_event_publisher.go
@@ -119,18 +119,12 @@ func (p *Publisher) run(ctx context.Context) error {
 		if err := stream.Send(req); err != nil {
 			return status.UnavailableErrorf("send failed: %s", err)
 		}
-		_, finished := obe.Event.Event.(*bepb.BuildEvent_ComponentStreamFinished)
-		if !finished {
-			continue
-		}
-		// After successfully transmitting all events, close our side of the stream
-		// and wait for server ACKs before closing the connection.
-		if err := stream.CloseSend(); err != nil {
-			return status.UnavailableErrorf("close send failed: %s", err)
-		}
-		break
 	}
-
+	// After successfully transmitting all events, close our side of the stream
+	// and wait for server ACKs before closing the connection.
+	if err := stream.CloseSend(); err != nil {
+		return status.UnavailableErrorf("close send failed: %s", err)
+	}
 	// Return any error from receiving ACKs
 	return <-doneReceiving
 }
--- a/enterprise/server/build_event_publisher/build_event_publisher.go
+++ b/enterprise/server/build_event_publisher/build_event_publisher.go
@@ -9,6 +9,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/auth"
 	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/retry"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/google/uuid"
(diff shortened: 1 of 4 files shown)


enterprise/server/build_event_publisher/build_event_publisher.go
```diff
@@ -88,98 +93,133 @@ func (p *Publisher) run(ctx context.Context) {
...
 }
 
-func (p *Publisher) Publish(e *bespb.BuildEvent) error {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-	if p.err != nil {
-		return status.WrapError(p.err, "cannot publish event due to previous error")
+// Wait publishes the Finished event and then waits for the server to ACK
+// all published events.
+func (p *Publisher) Wait() error {
```
vadimberezniker, line 174: `Finish` might be a better name for this func since it's not only waiting.

enterprise/server/build_event_publisher/build_event_publisher.go
```diff
@@ -88,98 +93,133 @@ func (p *Publisher) run(ctx context.Context) {
...
+	if err := p.Publish(nil); err != nil {
```
vadimberezniker, line 175: Instead of treating `nil` as a special value, why don't we have this function create the finished event?

enterprise/server/build_event_publisher/build_event_publisher.go
```diff
@@ -88,98 +93,133 @@ func (p *Publisher) run(ctx context.Context) {
...
+	// notify is used to signal that an event has been added to the buffer.
+	notify chan struct{}
+}
+
+func NewEventBuffer() *EventBuffer {
+	return &EventBuffer{notify: make(chan struct{}, 1)}
+}
+
+// Get returns the currently buffered items.
+func (b *EventBuffer) Get() []*bepb.BuildEvent {
```
vadimberezniker, line 204: It would be a little cleaner to have a `Subscribe()`-like method that returns a new channel that streams the data from `items` (existing and new) instead of putting that responsibility on the client.

vadimberezniker (changes requested):

enterprise/server/build_event_publisher/build_event_publisher.go
bduffany: Done -- I agree that this is cleaner :)

enterprise/server/build_event_publisher/build_event_publisher.go
```diff
@@ -88,98 +92,153 @@ func (p *Publisher) run(ctx context.Context) {
...
+}
+
+// Subscribe returns a channel that streams the currently buffered events as
+// well as any subsequently added events, up until the ComponentStreamFinished
+// event.
+//
+// The returned callback function cancels the subscription. It must be called,
+// or else a goroutine leak may occur. Once cancel is called, it is safe for
+// another goroutine to call Subscribe() again.
+func (b *EventBuffer) Subscribe() (<-chan *pepb.OrderedBuildEvent, func()) {
```
tylerwilliams, line 193: it feels a little strange to return a channel to read from (read) and a cancel func (write/cancel) instead of having the producer close the channel. Since you're just deferring the cancel at the Subscribe() callsite -- could we instead clean things up another way?

tylerwilliams (changes requested):

enterprise/server/build_event_publisher/build_event_publisher.go
```diff
@@ -19,67 +20,70 @@ import (
...
 		besBackend: besBackend,
-
-		// We probably won't ever saturate this buffer since we only need to
-		// publish a few events for the actions themselves and progress events
-		// are rate-limited. Also, events are sent to the server with low
-		// latency compared to the rate limiting interval.
-		events: make(chan *bespb.BuildEvent, 256),
-		done:   make(chan struct{}, 1),
+		events:     NewEventBuffer(streamID),
+		done:       make(chan error, 1),
```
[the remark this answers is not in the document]
bduffany: sg, renamed to `errCh`

enterprise/server/build_event_publisher/build_event_publisher.go
```diff
@@ -19,67 +20,70 @@ import (
...
 func (p *Publisher) Start(ctx context.Context) {
-	go p.run(ctx)
-}
-
-func (p *Publisher) run(ctx context.Context) {
-	defer func() {
-		p.done <- struct{}{}
+	go func() {
+		defer close(p.done)
+		r := retry.New(ctx, &retry.Options{
```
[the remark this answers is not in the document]
bduffany: The default is time-based, and I wasn't sure there was a sane value to use for the timeout, so I decided to be consistent with what Bazel does and use a count-based max.

enterprise/server/build_event_publisher/build_event_publisher.go
```diff
@@ -88,98 +92,153 @@ func (p *Publisher) run(ctx context.Context) {
...
+// It supports a single consumer and multiple producers: multiple goroutines can
+// call Add(), and a single goroutine can call Subscribe() to get the currently
+// buffered events and subsequently added events.
+type EventBuffer struct {
+	streamID *bepb.StreamId
+
+	mu sync.Mutex
+	// TODO(bduffany): Store these on disk since the buffer can get quite large if
+	// there is a lot of log output.
+	items []*bepb.BuildEvent
```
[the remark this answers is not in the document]
bduffany: Yep, saving this for a followup -- I'll file an issue to track

enterprise/server/build_event_publisher/build_event_publisher.go
bduffany: The producer does close the channel (`defer close(events)` at the start of the goroutine). But the goroutine needs a way to get notified that the receiver has stopped receiving values, otherwise it will get blocked indefinitely when trying to send on the channel. I could be missing something though -- maybe there is a cleaner way to do this?

tylerwilliams (approved):
LGTM

PR #1704, opened 2022-03-14
Allow trusted workflow execution on PR approval events

bduffany:
When receiving an approving pull request review from an OWNER, MEMBER, or COLLABORATOR of the repo into which the PR is being merged, treat the commit as trusted. The trust only applies to the approved commit and not subsequent commits (error statuses will be generated on subsequent commits, until another approval is granted).

This allows Mac workflows to run on PRs from OSS contributors, such as: https://github.com/buildbuddy-io/rules_xcodeproj/pull/123

---

**Version bump**: Patch


bduffany:
@brentleyjones This is a simpler approach than the one I mentioned in https://github.com/buildbuddy-io/buildbuddy-internal/issues/1213 -- instead of having an explicit approval flow on our end, the approval is implicit, via an approving review from an owner or member of the repo on GitHub. Let me know what you think about this approach

brentleyjones:
This works as long as the workflow check is a requirement to merge the PR. Some setups might have them be non-blocking, and the reviewer might want to see a passing check for some PRs or authors before approving.

brentleyjones:
Also, needing to approve multiple times, for each commit, can be a pain, and breaks some peoples processes.

enterprise/server/webhooks/github/github.go
```diff
@@ -137,38 +137,68 @@ func (*githubGitProvider) ParseWebhookData(r *http.Request) (*interfaces.Webhook
...
+func isTrustedAssociation(association string) bool {
+	return association == "OWNER" || association == "MEMBER" || association == "COLLABORATOR"
+}
+
+// parsePullRequestOrReview extracts WebhookData from a pull_request or
+// pull_request_review event.
+func parsePullRequestOrReview(event interface{}) (*interfaces.WebhookData, error) {
+	v, err := fieldgetter.ExtractValues(
+		event,
+		"Action",
```
tylerwilliams, line 180: is this used below?

enterprise/server/webhooks/github/github.go
```diff
@@ -137,38 +137,68 @@ func (*githubGitProvider) ParseWebhookData(r *http.Request) (*interfaces.Webhook
...
+		return nil, err
+	}
+	isFork := v["PullRequest.Base.Repo.CloneURL"] != v["PullRequest.Head.Repo.CloneURL"]
+	return &interfaces.WebhookData{
+		EventName:     webhook_data.EventName.PullRequest,
+		PushedRepoURL: v["PullRequest.Head.Repo.CloneURL"],
+		PushedBranch:  v["PullRequest.Head.Ref"],
+		SHA:           v["PullRequest.Head.SHA"],
+		TargetRepoURL: v["PullRequest.Base.Repo.CloneURL"],
+		TargetBranch:  v["PullRequest.Base.Ref"],
```
tylerwilliams, line 197: minor thing, but it would be nice to make the list of fields you're using and referencing match in order. That way a reader can see you're requesting 5 things and then using those same 5, without having to play matchmaker in their head.

tylerwilliams (changes requested):

tylerwilliams (approved):

PR #1705, opened 2022-03-14
Upgrade go-github to v43

bduffany:
The upgrade gives us access to some generated getters that aren't available in our current version, particularly `PullRequestReview.GetAuthorAssociation()` used in https://github.com/buildbuddy-io/buildbuddy/pull/1704

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #1707, opened 2022-03-14
Enable web tracing

tylerwilliams:
This CL allows appending "debug=1" or "debug=true" to a URL in the UI to request the backend to force tracing.

My intent with this CL is to make it easier to debug performance issues in prod: currently it's very difficult to see why queries are slow for a particular page load without setting up a mirror environment and running the same request with debug logging on. After this change, it should only require loading the page with debug=1 and then fishing around in jaeger.

The app enables this functionality by setting an HTTP header in requests to the backend when the debug=1 flag is present in the URL. For simplicity, the header is only set on RPC (via protolet) requests.

The backend protolet previously did not set any gRPC metadata before calling handler implementations -- now it does. Only whitelisted headers are put into the metadata.

### DIFF
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -35,6 +35,21 @@ class RpcService {
     (window as any)._rpcService = this;
   }
 
+  debuggingEnabled(): boolean {
+    const url = new URL(window.location.href);
+    let sp = url.searchParams.get("debug");
+    if (sp == null) {
+      let match = url.hash.match("debug=(.*)");
+      if (match != null) {
+        sp = match[1];
+      }
+    }
+    if (sp == "1" || sp == "true" || sp == "True") {
+      return true;
+    }
+    return false;
+  }
+
   getBytestreamUrl(bytestreamURL: string, invocationId: string, { filename = "" } = {}): string {
     const encodedRequestContext = uint8ArrayToBase64(context.RequestContext.encode(this.requestContext).finish());
     const params: Record<string, string> = {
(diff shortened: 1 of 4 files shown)


app/service/rpc_service.ts
```diff
@@ -35,6 +35,21 @@ class RpcService {
...
+  debuggingEnabled(): boolean {
+    const url = new URL(window.location.href);
+    let sp = url.searchParams.get("debug");
+    if (sp == null) {
+      let match = url.hash.match("debug=(.*)");
+      if (match != null) {
+        sp = match[1];
+      }
+    }
+    if (sp == "1" || sp == "true" || sp == "True") {
```
bduffany, line 47:
```suggestion
    if (sp === "1" || sp === "true" || sp === "True") {
```

app/service/rpc_service.ts
```diff
@@ -35,6 +35,21 @@ class RpcService {
     (window as any)._rpcService = this;
   }
 
+  debuggingEnabled(): boolean {
+    const url = new URL(window.location.href);
+    let sp = url.searchParams.get("debug");
+    if (sp == null) {
```
bduffany, line 41:
```suggestion
    if (sp === null) {
```

Using strict equality checks to avoid unwanted type coercion

app/service/rpc_service.ts
```diff
@@ -35,6 +35,21 @@ class RpcService {
...
+      let match = url.hash.match("debug=(.*)");
+      if (match != null) {
```
bduffany, line 43:
```suggestion
      if (match !== null) {
```

bduffany (approved):

app/service/rpc_service.ts
```diff
@@ -35,6 +35,21 @@ class RpcService {
     (window as any)._rpcService = this;
   }
 
+  debuggingEnabled(): boolean {
+    const url = new URL(window.location.href);
+    let sp = url.searchParams.get("debug");
+    if (sp === null) {
+      let match = url.hash.match("debug=(.*)");
```
bduffany, line 42: on second though, I'm not sure this hash matching logic will work -- pages that check for `hash == <something>` will not render as expected, I think

tylerwilliams: I can remove this bit and we can rely on the searchParams. Once we switch all routes to use /paths instead of fragments, this should be a non-issue. wdyt?

PR #1708, opened 2022-03-14
Website polish

siggisim:
- Make image loading smoother
- Enable lazy image loading in dev so we catch these kinds of issues earlier
- Set alt tags for images

---

**Version bump**: None


tylerwilliams (approved):
LGTM

PR #1710, opened 2022-03-15
Add package for starting test executor binary.

vadimberezniker:
The new `testexecutor` package is similiar to the existing `app` package
for starting the app binary.

The common behavior has been moved to the `testserver` package.

I also moved the `app.FreePort` function into a new `testport` package
since it doesn't logically belong to the app package and I want to use
it from the testexecutor package.

---

**Version bump**: None


tylerwilliams (approved):
LGTM

server/testutil/testserver/testserver.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	cmd := exec.Command(runfile(t, opts.BinaryPath), opts.Args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() {
+		cmd.Process.Kill() // ignore errors
```
bduffany, line 63: nit: should we change this to `cmd.Process.Signal(syscall.SIGTERM)` to allow shutdown logic to run?

bduffany (approved):
nice!

server/testutil/testserver/testserver.go
vadimberezniker: > nit: should we change this to `cmd.Process.Signal(syscall.SIGTERM)` to allow shutdown logic to run?

I don't see the value. Since this is after the test is done, it doesn't matter if the binary shuts down gracefully or not.

PR #1712, opened 2022-03-15
Remove extraneous continue

tempoz:
Don't need to continue here, switch breaks from cases by default in golang and only falls through if it hits the `fallthrough` keyword

---

**Version bump**: Patch


bduffany (approved):

tylerwilliams (approved):
LGTM

PR #1713, opened 2022-03-15
Don't report commit statuses for inner workflow invocations

bduffany:
This prevents the `bazel test //...` statuses from showing up, which are redundant with the action statuses (e.g. "Test all") and can be easily clobbered if multiple actions execute the same bazel command (e.g. "Test all" and "Test all (Mac)" which both run `bazel test //...`).

Tested locally.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1229

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -1206,6 +1206,12 @@ func writeBazelrc(path string) error {
 
 	lines := []string{
 		"build --build_metadata=ROLE=CI",
+		// Don't report commit statuses for individual bazel commands, since the
+		// overall status of all bazel commands is reflected in the status reported
+		// for the workflow invocation. In addition, for PRs, we first merge with
+		// the target branch which causes the HEAD commit SHA to change, and this
+		// SHA won't actually exist on GitHub.
+		"build --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true",
 		"build --bes_backend=" + *besBackend,
 		"build --bes_results_url=" + *besResultsURL,
 	}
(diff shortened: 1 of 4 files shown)


siggisim (approved):
Seems like the clobbering could be improved by either prefixing or suffixing the status with either the platform or the workflow action name.

I don't think these are that redundant since they link to different places in the BuildBuddy UI (maybe we just need to label them better, or fix that). What happens if a workflow kicks off multiple invocations?

Fine with this going in, just wanted to give my $0.02

bduffany:
@siggisim I agree that it might be nice to have a status reported for each `bazel_command` in the workflow if there are multiple invocations - I am a bit on the fence about the case where there's only one invocation though, since it clutters the GitHub UI a bit, and I suspect it might confuse people since our workflow UI and invocation UI look very similar today.

Will punt on figuring out the details of that for now, and address in a future PR

siggisim:
> @siggisim I agree that it might be nice to have a status reported for each `bazel_command` in the workflow if there are multiple invocations - I am a bit on the fence about the case where there's only one invocation though, since it clutters the GitHub UI a bit, and I suspect it might confuse people since our workflow invocation UI and bazel invocation UI look very similar today and we don't link to the parent workflow, which allows the workflow to be re-run etc.

If anything it seems like killing the workflow link seems closer to what we want. All we really want is the invocation links + a re-run button.

bduffany:
> If anything it seems like killing the workflow link seems closer to what we want. All we really want is the invocation links + a re-run button.

That's a good option too, I think. The workflow link does surface some info that isn't in the invocations, such as the status/logs from pulling the repo, which can fail. If we can surface that directly in the invocation somehow, that might be a good way to effectively "hide" the outer invocation while still keeping all the useful info around

PR #1714, opened 2022-03-15
Revert "Add command to test grid query"

siggisim:
Reverts buildbuddy-io/buildbuddy#1696


siggisim:
Reverted because we're seeing this error in dev:
```
"Database auto-migration failed: Error 1091: Can't DROP 'invocations_test_grid_query_index'; check that column/key exists"
```

Looks like the check and the drop are acting on different tables.

tylerwilliams (approved):
LGTM

PR #1718, opened 2022-03-16
When creating / editing API keys allow admins to chose whether they're visible to developers

siggisim:
This allows admins to choose whether or not individual API keys are visible to developers when creating / updating keys.

All existing API keys will only be visible to Admins.

Admin:
Screen Shot 2022-03-15 at 5 08 30 PM

User:
Screen Shot 2022-03-15 at 5 09 02 PM

**Version bump**: Patch

### DIFF
--- a/enterprise/app/api_keys/api_keys.tsx
+++ b/enterprise/app/api_keys/api_keys.tsx
@@ -241,8 +241,7 @@ export default class ApiKeysComponent extends React.Component<ApiKeysComponentPr
     onChange: (name: string, value: any) => any,
     e: React.ChangeEvent<HTMLInputElement>
   ) {
-    request.visibleToDevelopers = e.target.checked;
-    onChange("visibleToDevelopers", request.visibleToDevelopers);
+    onChange("visibleToDevelopers", e.target.checked);
   }
 
   private onChangeRegisterExecutor<T extends ApiKeyFields>(
@@ -489,7 +488,7 @@ function describeCapabilities<T extends ApiKeyFields>(apiKey: T | null) {
     capabilities += "+Executor";
   }
   if (isVisibleToDevelopers(apiKey)) {
-    capabilities += " [Devs]";
+    capabilities += " [D]";
   }
   return capabilities;
 }
--- a/enterprise/app/api_keys/api_keys.tsx
+++ b/enterprise/app/api_keys/api_keys.tsx
@@ -144,6 +144,7 @@ export default class ApiKeysComponent extends React.Component<ApiKeysComponentPr
           id: apiKey.id,
           label: apiKey.label,
           capability: [...apiKey.capability],
+          visibleToDevelopers: apiKey.visibleToDevelopers,
         }),
       },
     });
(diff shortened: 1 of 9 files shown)


enterprise/app/api_keys/api_keys.tsx
```diff
@@ -235,6 +236,15 @@ export default class ApiKeysComponent extends React.Component<ApiKeysComponentPr
     this.onChangeCapability(request, api_key.ApiKey.Capability.CACHE_WRITE_CAPABILITY, !e.target.checked, onChange);
   }
 
+  private onChangeVisibility<T extends ApiKeyFields>(
+    request: T,
+    onChange: (name: string, value: any) => any,
+    e: React.ChangeEvent<HTMLInputElement>
+  ) {
+    request.visibleToDevelopers = e.target.checked;
+    onChange("visibleToDevelopers", request.visibleToDevelopers);
```
bduffany, line 245:
```suggestion
    onChange("visibleToDevelopers", e.target.checked);
```

(`onChange` winds up calling `setState` with a new request where the provided field is equal to the value passed in)

server/tables/tables.go
```diff
@@ -305,7 +305,8 @@ type APIKey struct {
 	//
 	// NOTE: If the default is changed, a DB migration may be required to
 	// migrate old DB rows to reflect the new default.
-	Capabilities int32 `gorm:"default:1"`
+	Capabilities        int32 `gorm:"default:1"`
+	VisibleToDevelopers bool
```
bduffany, line 309: This field will be NULL by default -- dealing with nulls can sometimes be a pain, so IMO it'd be better to add the annotation `gorm:"not null"` and GORM will make this value `false` for existing rows instead

Also just double-checking, do you want this to be set to false for existing rows?

enterprise/server/backends/userdb/userdb.go
```diff
@@ -330,7 +344,7 @@ func (d *UserDB) InsertOrUpdateGroup(ctx context.Context, g *tables.Group) (stri
 			if err := tx.Create(&newGroup).Error; err != nil {
 				return err
 			}
-			_, err = createAPIKey(tx, groupID, newAPIKeyToken(), defaultAPIKeyLabel, defaultAPIKeyCapabilities)
+			_, err = createAPIKey(tx, groupID, newAPIKeyToken(), defaultAPIKeyLabel, defaultAPIKeyCapabilities, false)
```
tylerwilliams, line 347: ```..., false /*visibleToDevelopers*/)``` here and below

enterprise/server/backends/userdb/userdb.go
```diff
@@ -155,10 +156,21 @@ func (d *UserDB) GetAPIKeys(ctx context.Context, groupID string) ([]*tables.APIK
...
+	}
+
 	query := d.h.DB(ctx).Raw(`
-		SELECT api_key_id, value, label, perms, capabilities
+		SELECT api_key_id, value, label, perms, capabilities, visible_to_developers
 		FROM APIKeys
 		WHERE group_id = ?
+		`+filter+`
 		ORDER BY label ASC
 	`, groupID)
```
tylerwilliams, line 175: this is old code, pre-query builder, but would benefit from it. something like:

```
q := query_builder.NewQuery(`SELECT api_key_id, value, label, perms, capabilities, visible_to_developers FROM APIKeys`)
q.AddWhereClause("group_id = ?", groupID)
q.SetOrderBy("label", true /*ascending*/)
if err := authutil.AuthorizeGroupRole(u, groupID, role.Admin); err == nil {
		q.AddWhereClause("visible_to_developers = ?", true)
}
query := d.h.DB(ctx).Raw(q.Build())
```

tylerwilliams (approved):
LGTM w/ a few minor nits

this is sweet

server/tables/tables.go
siggisim: Yeah, I definitely want all API keys that are visible to developers to be manually marked as such.

I've added `gorm:"not null;default:false"` to make the default value false.

PR #1719, opened 2022-03-16
Roll forward #1696: Add command to test grid query

bduffany:
Commit containing just the fixes (tested locally): https://github.com/buildbuddy-io/buildbuddy/pull/1719/commits/38bea110281ff7d4373fb7de9b062f063bb7671e

* Fix erroneous reference to "TargetStatuses" instead of "Invocations"
* Fix sqlite migration bug that occurs because we check existence of the index pre-migration but defer the drop statement execution until post-automigration. The deferred DROP INDEX fails because the index gets dropped automatically as part of GORM auto-migration (more accurately, GORM recreates the entire table from scratch, copying values from the old table). This behavior only appears to happen for sqlite and not mysql.

---

**Version bump**: Patch

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -637,11 +637,6 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
 			return nil
 		})
 	}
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):
LGTM

PR #1720, opened 2022-03-16
Make web tracing actually work.

tylerwilliams:
This allows the debug=1 param to work correctly. When set, an HTTP header is passed from app -> backend, then passed from http request -> grpc metadata, and used to force tracing. This change also adds tracing to the gorm library, so that sql queries will be traced on a request span.

- upgrade gorm and myqsl driver
- add a dep on https://github.com/uptrace/opentelemetry-go-extra/tree/main/otelgorm to enable gorm tracing
- move grpc metadata injection to tracing so it happens first


siggisim (approved):

server/util/tracing/tracing.go
```diff
@@ -246,7 +247,32 @@ func (m *HttpServeMux) Handle(pattern string, handler http.Handler) {
...
+	}
+	m[k] = make([]string, len(h[k]))
+	copy(m[k], h[k])
+}
+
+func headersToContext(h http.Header) metadata.MD {
+	m := make(metadata.MD, 0)
+	for key, _ := range h {
+		switch strings.ToLower(key) {
+		case "x-buildbuddy-trace":
```
vadimberezniker, line 262: you can use the traceHeader constant in the same file

vadimberezniker (approved):
nice

server/util/db/db.go
```diff
@@ -247,7 +248,9 @@ func openDB(configurator *config.Configurator, dialect string, connString string
 	}
 
 	instrumentGORM(gdb)
-
+	if err := gdb.Use(otelgorm.NewPlugin()); err != nil {
```
vadimberezniker, line 251: it looks like by default it will store variable values in the span, which can contain sensitive data. 
there's an WithoutQueryVariables option to disable this.

PR #1724, opened 2022-03-17
Add some tests for RBE signal handling

bduffany:
I plan on changing some of the error handling logic in RBE so that we can retry errors caused by failing to unpause runners, and make sure we don't return runners back to the pool.

The purpose of this PR is to add some tests to demonstrate and solidify the current behavior of the system, so that it will be more clear what behavior is changing as part of this upcoming error handling PR.

---

**Version bump**: Patch


vadimberezniker (approved):
nice, thank you for adding coverage for this

PR #1734, opened 2022-03-18
Explictly define flags previously implicitly defined by the config

tempoz:
- explicitly define database flags
- explicitly define app flags
- explicitly define build_event_proxy flags
- explicitly define storage flags
- explicitly define cache flags
- explicitly define flags for executor

First set of config flags to be explicitly defined. Separated from others both for size
considerations and to ensure that the correct direction is being taken.

---

**Version bump**: Patch

### DIFF
--- a/server/backends/blobstore/blobstore.go
+++ b/server/backends/blobstore/blobstore.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"flag"
 	"fmt"
 	"io/ioutil"
 	"net/url"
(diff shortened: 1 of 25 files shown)


server/backends/blobstore/blobstore.go
```diff
@@ -35,6 +36,32 @@ import (
...
+	awsS3RoleSessionName          = flag.String("storage.aws_s3.role_session_name", "", "The role session name to use for web identity auth.")
+	awsS3Endpoint                 = flag.String("storage.aws_s3.endpoint", "", "The AWS endpoint to use, useful for configuring the use of MinIO.")
+	awsS3StaticCredentialsID      = flag.String("storage.aws_s3.static_credentials_id", "", "Static credentials ID to use, useful for configuring the use of MinIO.")
+	awsS3StaticCredentialsSecret  = flag.String("storage.aws_s3.static_credentials_secret", "", "Static credentials secret to use, useful for configuring the use of MinIO.")
+	awsS3StaticCredentialsToken   = flag.String("storage.aws_s3.static_credentials_token", "", "Static credentials token to use, useful for configuring the use of MinIO.")
+	awsS3DisableSSL               = flag.Bool("storage.aws_s3.disable_ssl", false, "Disables the use of SSL, useful for configuring the use of MinIO.")
+	awsS3ForcePathStyle         = flag.Bool("storage.aws_s3.s3_force_path_style", false, "Force path style urls for objects, useful for configuring the use of MinIO.")
```
tempoz, line 57:
```suggestion
	awsS3ForcePathStyle           = flag.Bool("storage.aws_s3.s3_force_path_style", false, "Force path style urls for objects, useful for configuring the use of MinIO.")
```

tempoz:
Taking a different direction with this change.

PR #1736, opened 2022-03-18
Implement UnlinkGitHubAccount

bduffany:
* Clear `Group.github_token`
* Delete workflows registered with the GitHub token, and report any issues unregistering webhooks as warnings

This does not actually revoke app access, but just clears our references to the access token. We instead advise the user to revoke access via GitHub's UI.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

Kind on the fence for this CL but I think in general we should strive to write all queries with the query_builder so it's easier to do broad db-level refactorings down the line (say, if we wanted to support postgresql, or something like that).

PR #1738, opened 2022-03-18
Reuse MySQL server instances across tests

bduffany:
This gets `usage_test_mysql` running on CI in about ~6.6s total (for warm runs)

---

**Version bump**: Patch


bduffany:
FYI the failure in prod is expected -- it won't work until firecracker runner recycling is fixed in prod (in this week's release)

tylerwilliams (approved):

vadimberezniker (approved):

PR #1741, opened 2022-03-21
Don't return NO_STATUS rows from TargetStatuses table

bduffany:
These NO_STATUS rows are stored when running a test with `--test_tag_filters` and test targets are matched by the pattern (e.g. `//...`) but excluded by the tag filters. For example, when running our "Docker tests" workflow we see that a NO_STATUS gets reported for all targets which aren't tagged with `docker`.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1244

### DIFF
--- a/server/build_event_protocol/target_tracker/target_tracker.go
+++ b/server/build_event_protocol/target_tracker/target_tracker.go
@@ -234,7 +234,7 @@ func (t *TargetTracker) writeTestTargetStatuses(ctx context.Context, permissions
 	}
 	newTargetStatuses := make([]*tables.TargetStatus, 0)
 	for _, target := range t.targets {
-		if !isTest(target) || target.overallStatus == build_event_stream.TestStatus_NO_STATUS {
+		if !isTest(target) {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


server/build_event_protocol/target_tracker/target_tracker.go
```diff
@@ -234,7 +234,7 @@ func (t *TargetTracker) writeTestTargetStatuses(ctx context.Context, permissions
 	}
 	newTargetStatuses := make([]*tables.TargetStatus, 0)
 	for _, target := range t.targets {
-		if !isTest(target) {
+		if !isTest(target) || target.overallStatus == build_event_stream.TestStatus_NO_STATUS {
 			continue
 		}
```
luluz66, line 239: Discussed offline. One benefit about writing no status here is that if there are cases where the test failed and we mislabeled them as NO_STATUS (for example, "ABORTED" or "FAILED TO BUILD". It will be easier to find them compared to they don't appear in the db in the first place.

luluz66 (approved):

PR #1753, opened 2022-03-22
Use type parameters for component props and state

bduffany:
Motivation:
* Specifying `props: Props` instead of `React.Component<Props>` prevents the compiler from understanding the expected types of `shouldComponentUpdate` and other lifecycle methods, which can be helpful for catching certain types of bugs. See https://codesandbox.io/s/withered-fog-pq8o3x?file=/src/App.tsx
* Specifying `props: Props` is also incompatible with TS `strict` mode, which would be nice to enable at some point (for better null-safety, etc.)
* When not using type parameters for `State`, the compiler is unable to detect issues with `this.setState`. See https://codesandbox.io/s/dazzling-taussig-0r4s1y?file=/src/App.tsx:0-339

This PR refactors

```
class Foo extends React.Component {
  props: Props;
  state: State = { ... };
}
```

to

```
class Foo extends React.Component<Props, State> {
  state: State = { ... }
}
```

and fix any compile-time errors that result from the change.

The only real bugs caught by this refactoring were in `action_card.tsx`, and the remaining build errors are fixed by fiddling with types to make the compiler happy.

Unfortunately `this.setState` requires some awkward casting in a few places, and there is not really a good workaround -- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18365 has more context

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_execution_card.tsx
+++ b/app/invocation/invocation_execution_card.tsx
@@ -19,7 +19,7 @@ interface Props {
 
 interface State {
   loading: boolean;
-  executions: execution_stats.Execution[];
+  executions: execution_stats.IExecution[];
   sort: string;
   direction: "asc" | "desc";
   statusFilter: string;
@@ -45,7 +45,7 @@ const GRPC_STATUS_LABEL_BY_CODE: Record<number, string> = Object.fromEntries(
   Object.entries(google.rpc.Code).map(([name, value]) => [value, name])
 );
 
-function getExecutionStatus(execution: execution_stats.Execution): ExecutionStatus {
+function getExecutionStatus(execution: execution_stats.IExecution): ExecutionStatus {
   if (execution.stage === ExecutionStage.Value.COMPLETED) {
     if (execution.status.code !== 0) {
       return {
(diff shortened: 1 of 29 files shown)


app/invocation/invocation_execution_card.tsx
```diff
@@ -218,27 +216,27 @@ export default class ExecutionCardComponent extends React.Component {
     }
   }
 
-  handleInputChange(event: any) {
-    const target = event.target;
+  handleInputChange(event: React.ChangeEvent) {
```
siggisim, line 219: I think you could eliminate the casting here and below with something like
```
React.ChangeEvent<HTMLInputElement>
```

enterprise/app/code/code.tsx
```diff
@@ -617,7 +615,7 @@ export default class CodeComponent extends React.Component<Props> {
   }
 
   updateState(newState: Partial<State>, callback?: VoidFunction) {
-    this.setState(newState, () => {
+    this.setState(newState as State, () => {
```
siggisim, line 618: Surprised this cast is needed, I though it took State|Partial<State>|(a bunch of other stuff)

siggisim (approved):
Thanks for cleaning all of this up

app/invocation/invocation_execution_card.tsx
bduffany: Good catch, I was confusing this with `MouseEvent` where `target` doesn't wind up having the expected type even when a type param is passed. Fixed

enterprise/app/code/code.tsx
siggisim: 👍

PR #1754, opened 2022-03-22
Include workflow action name in runner routing key

bduffany:
This allows workflow actions to use different build configurations (e.g. `--define` flags) without discarding the analysis cache for other workflow actions. It also allows actions to run `bazel clean` as the first `bazel_command` without affecting the bazel cache from other actions.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1251

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -60,10 +60,15 @@ func generateWebhookID() (string, error) {
 	return strings.ToLower(u.String()), nil
 }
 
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
I was struggling with this until I realized that actionName is the *workflow* action name, not a name for the remote-execution action.

bduffany:
> I was struggling with this until I realized that actionName is the _workflow_ action name, not a name for the remote-execution action.

I can see how that confusion might arise -- updated some var names / comment wording to hopefully make it a bit more clear

PR #1769, opened 2022-03-24
Make redis metrics delays configurable via flags

bduffany:
This will allow us to set `--redis_command_buffer_flush_period=0` and `--cache_stats_finalization_delay=0` to avoid long delays while waiting for redis stats to be flushed during integration tests.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
+++ b/enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
@@ -62,15 +62,3 @@ func (c *collector) ReadCounts(ctx context.Context, key string) (map[string]int6
 func (c *collector) Delete(ctx context.Context, key string) error {
 	return c.rdb.Del(ctx, key).Err()
 }
-
-func (c *collector) ReadbackDelay() time.Duration {
-	if *redisutil.CommandBufferFlushPeriod == 0 {
-		// Command buffer is operating in synchronous mode, meaning any call to
-		// IncrementCount() is reflected in ReadCounts() immediately afterwards.
-		// So, there is no readback delay.
-		return 0
-	}
-	// Readback delay is the buffer period plus max expected amount of time
-	// it takes to flush all buffered values to Redis.
-	return *redisutil.CommandBufferFlushPeriod + 250*time.Millisecond
-}
--- a/enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
+++ b/enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
@@ -63,13 +63,14 @@ func (c *collector) Delete(ctx context.Context, key string) error {
 	return c.rdb.Del(ctx, key).Err()
 }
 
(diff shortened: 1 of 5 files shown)


enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
```diff
@@ -62,3 +62,14 @@ func (c *collector) ReadCounts(ctx context.Context, key string) (map[string]int6
...
+
+func (c *collector) BufferDelay() time.Duration {
+	// Buffer is operating in synchronous mode; don't report a buffer delay.
+	if *redisutil.CommandBufferFlushPeriod == 0 {
+		return 0
+	}
+	// Report the buffer delay as slightly more than the command buffer flush
+	// period, to allow some time to actually execute the Redis commands, plus
+	// a little extra to account for clock skew across apps.
+	return *redisutil.CommandBufferFlushPeriod + 250*time.Millisecond
```
tylerwilliams, line 74: rather than "polluting" your implementation with this logic to make the build_event_handler code work -- can you add the delay there at that callsite and return the true value here?

tylerwilliams (changes requested):

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -19,9 +20,8 @@ import (
 	redsync_goredis "github.com/go-redsync/redsync/v4/redis/goredis/v8"
 )
 
-const (
-	// How long to wait between flushing CommandBuffer instances to Redis.
-	commandBufferFlushInterval = 250 * time.Millisecond
+var (
+	CommandBufferFlushPeriod = flag.Duration("redis_command_buffer_flush_period", 250*time.Millisecond, "How long to wait between flushing buffered redis commands. Setting to 0 will disable buffering.")
```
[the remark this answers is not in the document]
bduffany: Done

enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
bduffany: I think I see what you mean -- with a name like `BufferDelay`, this ought to reflect the actual amount of time that we're buffering.

My intent was actually to reflect the time that a caller has to wait between calling `IncrementCounts()` and the time that the counts can be "read back" via `ReadCounts()`. So I renamed this to `ReadbackDelay()`, and since this delay depends both on the buffering delay as well as the storage mechanism used, I think it makes sense for the 250ms value to be an implementation detail here. The 250ms is a worst case upperbound that is specific to Redis -- other implementations could have higher or lower values here.

Side note, I removed the comment about clock skew, since that is irrelevant as we're dealing with relative durations and not absolute timestamps

bduffany: Resolved offline to just make these constants flags, instead of adding an interface

tylerwilliams (approved):

PR #1776, opened 2022-03-24
Use flags for UI elements

tempoz:
---

**Version bump**: Minor

PR #1774, opened 2022-03-24
Use flags for URLs

tempoz:
Introduces URLFlags to allow URL parameters to be parsed on startup, which means we don't have
error check them everytime we parse them in the future.

These URLs are used in many places, so we introduce packages for the flags and assocaited code
to reside in, to give everything that uses them (and only things that use them) common
definitions of the flags.

---

**Version bump**: Minor

PR #1772, opened 2022-03-24
Use flags for log settings

tempoz:
Part of encapsulating these settings in the log lib is making Configure use the settings instead of other args.

---

**Version bump**: Minor

PR #1771, opened 2022-03-24
Use flags for database settings

tempoz:
---

**Version bump**: Minor

PR #1777, opened 2022-03-24
Use flags for usage service

tempoz:
Encapsulate usage options in the usage and usage service package; give them a common place
to get the usage enabled flag from.

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/cmd/server/main.go
+++ b/enterprise/server/cmd/server/main.go
@@ -143,7 +143,7 @@ func convertToProdOrDie(ctx context.Context, env *real_environment.RealEnv) {
 	search := invocation_search_service.NewInvocationSearchService(env, env.GetDBHandle())
 	env.SetInvocationSearchService(search)
 
-	usage_service.Register(env, env.SetUsageService)
+	usage_service.Register(env)
 
 	apiServer := api.NewAPIServer(env)
 	env.SetAPIService(apiServer)
--- a/enterprise/server/cmd/server/main.go
+++ b/enterprise/server/cmd/server/main.go
@@ -61,7 +61,6 @@ import (
 )
 
 var (
-
 	serverType = flag.String("server_type", "buildbuddy-server", "The server type to match on health checks")
 )
 
@@ -145,9 +144,7 @@ func convertToProdOrDie(ctx context.Context, env *real_environment.RealEnv) {
 	search := invocation_search_service.NewInvocationSearchService(env, env.GetDBHandle())
 	env.SetInvocationSearchService(search)
 
-	if env.GetConfigurator().GetAppUsageEnabled() {
-		env.SetUsageService(usage_service.New(env))
-	}
+	env.SetUsageService(usage_service.New(env))
 
 	apiServer := api.NewAPIServer(env)
 	env.SetAPIService(apiServer)
@@ -214,6 +211,7 @@ func main() {
 		realEnv.SetCacheRedisClient(redisClient)
 	}
 
+	var rbuf *redisutil.CommandBuffer
 	if redisConfig := configurator.GetDefaultRedisClientConfig(); redisConfig != nil {
 		rdb, err := redisutil.NewClientFromConfig(redisConfig, healthChecker, "default_redis")
 		if err != nil {
@@ -221,7 +219,7 @@ func main() {
 		}
 		realEnv.SetDefaultRedisClient(rdb)
 
-		rbuf := redisutil.NewCommandBuffer(rdb)
+		rbuf = redisutil.NewCommandBuffer(rdb)
 		rbuf.StartPeriodicFlush(context.Background())
 		realEnv.GetHealthChecker().RegisterShutdownFunction(rbuf.StopPeriodicFlush)
 
@@ -229,27 +227,16 @@ func main() {
 		realEnv.SetKeyValStore(rkv)
 		rmc := redis_metrics_collector.New(rdb, rbuf)
 		realEnv.SetMetricsCollector(rmc)
+	}
 
-		if configurator.GetAppUsageTrackingEnabled() {
-			region := realEnv.GetConfigurator().GetAppRegion()
-			if region == "" {
-				log.Fatalf("Usage tracking requires app.region to be configured.")
-			}
-			ut, err := usage.NewTracker(
-				realEnv, timeutil.NewClock(), usage.NewFlushLock(realEnv))
-			if err != nil {
-				log.Fatalf("Error creating usage tracker: %s", err)
-			}
-			realEnv.SetUsageTracker(ut)
-
-			ut.StartDBFlush()
-			healthChecker.RegisterShutdownFunction(func(ctx context.Context) error {
-				ut.StopDBFlush()
-				return nil
-			})
-		}
-	} else if configurator.GetAppUsageTrackingEnabled() {
-		log.Fatalf("Usage tracking is enabled, but no Redis client is configured.")
+	if ut, err := usage.NewTracker(realEnv, timeutil.NewClock(), usage.NewFlushLock(realEnv)); err != nil {
+		log.Fatalf("%v", err)
+	} else if ut != nil {
+		ut.StartDBFlush()
+		healthChecker.RegisterShutdownFunction(func(ctx context.Context) error {
+			ut.StopDBFlush()
+			return nil
+		})
 	}
 
 	if redisConfig := configurator.GetRemoteExecutionRedisClientConfig(); redisConfig != nil {
--- a/enterprise/server/cmd/server/main.go
+++ b/enterprise/server/cmd/server/main.go
@@ -61,6 +61,7 @@ import (
 )
 
 var (
+
 	serverType = flag.String("server_type", "buildbuddy-server", "The server type to match on health checks")
 )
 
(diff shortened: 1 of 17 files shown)


tempoz:
* master
  * **PR #1771** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1771" target="_blank">Graphite</a> 
    * **PR #1772** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1772" target="_blank">Graphite</a> 
      * **PR #1773** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1773" target="_blank">Graphite</a> 
        * **PR #1774** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1774" target="_blank">Graphite</a> 
          * **PR #1775** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1775" target="_blank">Graphite</a> 
            * **PR #1776** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1776" target="_blank">Graphite</a> 
              * **PR #1777** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1777" target="_blank">Graphite</a>  (this PR)
                * **PR #1778** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1778" target="_blank">Graphite</a> 
                  * **PR #1779** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1779" target="_blank">Graphite</a> 
                    * **PR #1780** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1780" target="_blank">Graphite</a> 

enterprise/server/usage/usage_config/usage_config.go
```diff
@@ -0,0 +1,9 @@
+package usage_config
+
+import "flag"
+
+var usageTrackingEnabled = flag.Bool("app.usage_tracking_enabled", false, "If set, enable usage data collection.")
+
+func UsageTrackingEnabled() bool {
+	return *usageTrackingEnabled
+}
```
bduffany, line 9: would it work to move this to `usage.go` and have `usage_service` depend on `usage.go` instead? Would be nice to remove this extra pkg if possible

bduffany (changes requested):

enterprise/server/usage/usage_config/usage_config.go
tempoz: We could, but it would create an otherwise unnecessary dependency on `usage.go`, which could mean that `app.region` could end up defined for a tool that only makes use of `usage_service`, which would make it a meaningless flag. I tried to avoid this as a design goal with all of these changes, but if we feel that I'm being over-cautious, I can switch it here.

enterprise/server/cmd/server/main.go
```diff
@@ -61,6 +61,7 @@ import (
 )
 
 var (
+
```
bduffany, line 64: nit: rm added blankline

enterprise/server/cmd/server/main.go
```diff
@@ -213,40 +212,32 @@ func main() {
...
-			healthChecker.RegisterShutdownFunction(func(ctx context.Context) error {
-				ut.StopDBFlush()
-				return nil
-			})
-		}
-	} else if configurator.GetAppUsageTrackingEnabled() {
-		log.Fatalf("Usage tracking is enabled, but no Redis client is configured.")
+	if ut, err := usage.NewTracker(realEnv, timeutil.NewClock(), usage.NewFlushLock(realEnv), rbuf); err != nil {
+		log.Fatalf("%v", err)
+	} else if ut != nil {
```
bduffany, line 235: nit: rm `ut != nil` check

enterprise/server/usage/usage.go
```diff
@@ -101,16 +99,25 @@ type tracker struct {
 	stopFlush chan struct{}
 }
 
-func NewTracker(env environment.Env, clock timeutil.Clock, flushLock interfaces.DistributedLock, rbuf *redisutil.CommandBuffer, opts *TrackerOpts) *tracker {
+func NewTracker(env environment.Env, clock timeutil.Clock, flushLock interfaces.DistributedLock, rbuf *redisutil.CommandBuffer) (*tracker, error) {
+	if !usage_config.UsageTrackingEnabled() {
```
bduffany, line 103: nit: it feels a bit unconventional for a `New*` func to return `nil, nil` -- would it work to move this enabled check to the callsite instead?

enterprise/server/usage/usage_config/usage_config.go
bduffany: Gotcha, I hadn't considered the effects of pulling in unused flags as deps. sgtm

enterprise/server/usage_service/usage_service.go
```diff
@@ -27,6 +31,9 @@ type usageService struct {
 }
 
 func New(env environment.Env) *usageService {
+	if !usage_config.UsageTrackingEnabled() {
```
bduffany, line 34: nit: same issue here re. `New*` returning nil -- would be slightly less surprising to do this check at the call site IMO

enterprise/server/usage/usage_config/usage_config.go
```diff
@@ -0,0 +1,9 @@
+package usage_config
```
tylerwilliams, line 1: can we call this config so the path names don't repeat?

(callers can choose to import as 'usage_config /path/to/usage/config' if they choose)

enterprise/server/usage_service/usage_service.go
tylerwilliams: The way I've seen this handled before is to instead call this method something like:

```
func Register(env environment.Env) {
  if enabled {
    env.SetUsageTracker(New(blah))
  }
}
```

This mostly works well -- the cases where it starts to break down are when you have many order-dependent backends and you need some set before or with others, then callers need to be aware of what's registered and not. That said, it forces you to break many of these dependencies and clarify things, which is kind of nice.

tempoz:
Switched usage tracker to using the metrics collector instead of rbuf directly, as this simplified the construction/registration of the usage tracker and also allowed the usage tracker to use the common code path for incrementing redis / setting expiry. If we'd rather see this change in a different PR (or not at all), feel free to lmk.

tylerwilliams:
Agree with you -- would you mind splitting the usage tracker changes into another CL?

tempoz:
Moved those changes out.

tylerwilliams (approved):
this looks really good to me; the only nit I have is that I think we should modify the env interface to contain the same setters realEnv does and have the register functions take the env interface rather than a realEnv. It's fine if you want to make that change as part of moving env to interfaces later on tho?

tylerwilliams:
Also thanks for bearing with me through all the back and forth!

bduffany (approved):

tempoz:
> this looks really good to me; the only nit I have is that I think we should modify the env interface to contain the same setters realEnv does and have the register functions take the env interface rather than a realEnv. It's fine if you want to make that change as part of moving env to interfaces later on tho?

Sounds good; feels a bit outside the scope of this PR :P

tempoz:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1777?utm_source=stack-comment).


Job Reference: Qg74nVRI1LSaCBRkxMdN

tempoz:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1777?utm_source=stack-comment).


Job Reference: IzK4IJkw70q5gig9wHik

tempoz:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1777?utm_source=stack-comment).


Job Reference: 6MoHHD1cAEi13ZYwKoui

tempoz:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1777?utm_source=stack-comment).


Job Reference: YXJ0IDeHv03gACfZZh33

tempoz:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1777?utm_source=stack-comment).


Job Reference: majh5r6PUbPED6xbQE58

tempoz:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1777?utm_source=stack-comment).


Job Reference: kON6XZCaQAfaQXroYZ1x

PR #1781, opened 2022-03-24
Don't attempt to modify workspace if using remote bazel

siggisim:
---

**Version bump**: Patch


vadimberezniker:
--remote=bazel implicitly enables remote execution as well. we do want to add the toolchain stuff to support rbe, right?

siggisim:
Because we set:
```
--remote_default_exec_properties=container-image=docker://gcr.io/flame-public/buildbuddy-ci-runner:latest
```

We shouldn't need to configure a toolchain.

vadimberezniker (approved):

PR #1782, opened 2022-03-24
Remove multiple action support from ci_runner.

vadimberezniker:
We no longer invoke it with multiple actions. This allows some of the
code to be simplified.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -88,24 +88,23 @@ const (
 )
 
 var (
-	besBackend                  = flag.String("bes_backend", "", "gRPC endpoint for BuildBuddy's BES backend.")
-	cacheBackend                = flag.String("cache_backend", "", "gRPC endpoint for BuildBuddy Cache.")
-	rbeBackend                  = flag.String("rbe_backend", "", "gRPC endpoint for BuildBuddy RBE.")
-	besResultsURL               = flag.String("bes_results_url", "", "URL prefix for BuildBuddy invocation URLs.")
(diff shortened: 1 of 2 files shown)


bduffany (approved):
Nice! This will improve workflow UX quite a bit, since we'll start publishing events/commit statuses immediately, instead of waiting for the repo to be cloned

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -489,127 +492,78 @@ func run() error {
...
-	if *reportLiveRepoSetupProgress {
-		ber, err := newBuildEventReporter(ctx, *besBackend, ws.buildbuddyAPIKey, *invocationID, *workflowID != "" /*=isWorkflow*/)
-		if err != nil {
-			return err
-		}
-		buildEventReporter = ber
-		if err := buildEventReporter.Start(ws.startTime); err != nil {
-			return status.WrapError(err, "could not publish started event")
-		}
+	if err := buildEventReporter.Start(ws.startTime); err != nil {
```
bduffany, line 495: unrelated to this PR, but I wonder if we should publish the `Started` event before we publish any progress events during repo setup? Wondering if this could lead to the same issue as https://github.com/buildbuddy-io/buildbuddy-internal/issues/1259

vadimberezniker: After this PR, the `Started` event will always be sent as the first event. `buildEventReporter.Start` publishes the `Started` event. The repo setup and everything else happens after this.

bduffany: ah, I thought that was the case at first, but then somehow confused myself. Nice :+1:

PR #1785, opened 2022-03-25
Speed up remote bazel checkouts when base commit doesn't change.

vadimberezniker:
ci_runner arg changes:
  --target_commit_sha can be specified instead of --target_branch to
checkout the target at a specific commit
  --push_repo_* args can be omitted

---

**Version bump**: None

### DIFF
--- a/enterprise/server/hostedrunner/hostedrunner.go
+++ b/enterprise/server/hostedrunner/hostedrunner.go
@@ -132,14 +132,14 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest, 
 		"--bes_backend=" + conf.GetAppEventsAPIURL(),
 		"--cache_backend=" + conf.GetAppCacheAPIURL(),
 		"--bes_results_url=" + conf.GetAppBuildBuddyURL() + "/invocation/",
-		"--commit_sha=" + req.GetRepoState().GetCommitSha(),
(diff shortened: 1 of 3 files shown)


enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -132,14 +132,14 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest,
...
 		"--bazel_sub_command=" + req.GetBazelCommand(),
-		"--pushed_branch=master",
-		"--target_branch=master",
 		"--invocation_id=" + invocationID,
-		"--report_live_repo_setup_progress",
+	}
+	if req.GetRepoState().GetCommitSha() != "" {
+		args = append(args, "--target_commit_sha="+req.GetRepoState().GetCommitSha())
+	} else {
+		args = append(args, "--target_branch=master")
```
tylerwilliams, line 142: do we need some smarts to pick master/main?

tylerwilliams (approved):

enterprise/server/hostedrunner/hostedrunner.go
vadimberezniker: We should modify the API to make the caller specify either the branch name or the SHA. The caller should know better. I will send a separate PR.

PR #1797, opened 2022-03-28
Add platform prop to disable networking

siggisim:
Using the Google RBE property defined here: https://github.com/bazelbuild/bazel-toolchains/blob/v5.1.0/rules/exec_properties/exec_properties.bzl#L156

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -50,6 +50,7 @@ type DockerOptions struct {
 	ForceRoot               bool
 	DockerMountMode         string
 	InheritUserIDs          bool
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -50,6 +51,7 @@ type DockerOptions struct {
 	ForceRoot               bool
 	DockerMountMode         string
 	InheritUserIDs          bool
+	DockerNetwork           string
```
bduffany, line 54: nit: maybe rename to `Network` to be consistent w/ podman

bduffany (approved):

PR #1800, opened 2022-03-29
Build Podman4 binary

luluz66:
Cannot find Podman 4 from an official repo; so have to build from
source.

Tested podman4 in dev and verified that tensorflow can be built with
network enabled; and there is no obvious regression performance wise
with podman 4.


dockerfiles/executor_image/Dockerfile
```diff
@@ -15,9 +15,32 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - && \
...
+RUN cd conmon-2.1.0 && make
+RUN cd conmon-2.1.0 && make podman
+# Install cni-plugins
+RUN mkdir -p /opt/cni/bin
+RUN curl -fsSL https://github.com/containernetworking/plugins/releases/download/v1.1.1/cni-plugins-linux-amd64-v1.1.1.tgz | tar xz -C /opt/cni/bin/
+
+# Add configuration
+RUN mkdir -p /etc/containers
+RUN curl -L -o /etc/containers/registries.conf https://src.fedoraproject.org/rpms/containers-common/raw/main/f/registries.conf
+RUN curl -L -o /etc/containers/policy.json https://src.fedoraproject.org/rpms/containers-common/raw/main/f/default-policy.json
```
tylerwilliams, line 43: I think each time you add a "RUN" command docker is going to make a new layer. Is it possible to add the --squash flag to the build step (in the BUILD file?)

PR #1801, opened 2022-03-29
Expand cache scorecard proto and add dedicated RPC for cache scorecard

bduffany:
Expanding the cache scorecard result message with more details, and adding a new field `result` which will contain a flat results list. This flattened list is easiest for the client to slice and dice.

Might need to iterate on this in subsequent PRs if we see that this approach doesn't scale well, but this flat list works well for development purposes. The plan is to only populate the new fields behind a flag.

---

**Version bump**: Patch

### DIFF
--- a/proto/cache.proto
+++ b/proto/cache.proto
@@ -42,6 +42,8 @@ message GetCacheScoreCardRequest {
 
   // The invocation ID for which to look up cache stats.
   string invocation_id = 2;
+
+  // TODO(bduffany): Add paging
 }
 
 message GetCacheScoreCardResponse {
--- a/proto/cache.proto
+++ b/proto/cache.proto
@@ -2,6 +2,11 @@ syntax = "proto3";
 
 package cache;
 
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "google/rpc/status.proto";
+import "proto/remote_execution.proto";
+
 // Next Tag: 14
 message CacheStats {
   // Server-side Action-cache stats.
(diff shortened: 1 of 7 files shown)


proto/cache.proto
```diff
@@ -30,13 +36,85 @@ message CacheStats {
...
 }
 
+// Request to retrieve detailed per-request cache stats.
+message GetCacheScoreCardRequest {
+  context.RequestContext request_context = 1;
+
+  // The invocation ID for which to look up cache stats.
+  string invocation_id = 2;
+
+  // TODO(bduffany): Add paging
```
tylerwilliams, line 46: nit: i think we have page token protos -- can you just copy those from elsewhere? should be the same.

proto/cache.proto
```diff
@@ -30,13 +36,85 @@ message CacheStats {
...
+    // Whether the transferred payload was compressed.
+    bool compressed = 10;
+
+    // The compressed size of the transferred payload. Should be ignored if
+    // `compressed` is false.
+    int64 compressed_size_bytes = 11;
   }
 
   // In the interest of saving space, we only show cache misses.
+  // TODO(bduffany): use flat `results` list and deprecate this
```
tylerwilliams, line 116: Something to consider: in the interest of performance we might want to do something in the future where by default we just show misses, and then if you hit a button in the UI, we go re-fetch the full scorecard.

Just putting this here as a note so we can both think about what that might look like -- if this list contained only miss results.

tylerwilliams (changes requested):

proto/cache.proto
bduffany: The only page token proto I could find was in `target_pagination.proto` and the fields will be different. Added a basic page token here

proto/cache.proto
bduffany: Ack, sg to return only misses by default. Will think about that when adding the sorting/filtering fields

proto/cache.proto
```diff
@@ -30,11 +36,96 @@ message CacheStats {
...
+
+message GetCacheScoreCardResponse {
+  context.ResponseContext response_context = 1;
+
+  // The cache results for the current page.
+  repeated ScoreCard.Result results = 2;
+
+  // A token used to fetch more results. If empty, there are no results
+  // remaining.
+  string page_token = 3;
```
tylerwilliams, line 69: nit: call this next_page_token (to allow us to add prev_page_token later if we want)

tylerwilliams (changes requested):

proto/cache.proto
bduffany: Done

proto/cache.proto
```diff
@@ -30,11 +36,96 @@ message CacheStats {
...
+  // The invocation ID for which to look up cache stats.
+  string invocation_id = 2;
+
+  // Pagination token. Not intended for direct client usage.
+  message PageToken {
+    // The start index in the scorecard results list, after sorting and
+    // filtering results. Number of results to fetch is determined by the
+    // server.
+    int64 index = 1;
+  }
```
tylerwilliams, line 52: Can we put this in context.proto (or make pagination.proto)?

tylerwilliams (changes requested):

proto/cache.proto
bduffany: Done

tylerwilliams (approved):

PR #1802, opened 2022-03-29
Use the metrics collector in the UsageTracker

tempoz:
Change UsageTracker to use MetricsCollector in preparation for flag changes.

This means that registration of the UsageTracker will no longer require access
to the bare `rbuf` (`redisutil.CommandBuffer`), and can instead depend solely
on the `env`.

---

**Version bump**: Minor


server/util/testing/flags/flags.go
```diff
@@ -7,10 +7,21 @@ import (
...
 
 	"github.com/buildbuddy-io/buildbuddy/server/config"
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+// Set a flag value and register a cleanup function to restore the flag
+// to its original value after the given test is complete.
+func SetAndReconcile(t testing.TB, name, value string, env environment.Env) {
```
tylerwilliams, line 17: nit: add a TODO that this will be removed in final cleanup?

tylerwilliams (approved):
LGTM as long as @bduffany is ok with it

server/util/testing/flags/flags.go
tempoz:
```suggestion
// TODO: Remove this function when we no longer need to reconcile flags
func SetAndReconcile(t testing.TB, name, value string, env environment.Env) {
```

server/backends/memory_metrics_collector/memory_metrics_collector.go
```diff
@@ -32,6 +33,35 @@ func NewMemoryMetricsCollector() (*MemoryMetricsCollector, error) {
 	}, nil
 }
 
+func (m *MemoryMetricsCollector) IncrementCounts(ctx context.Context, key string, counts map[string]int64) error {
```
bduffany, line 36: mind adding unit tests for IncrementCounts and IncrementCountsWithExpiry? We don't have any use cases that would directly exercise these new codepaths so it'd be nice to have some coverage

bduffany (changes requested):

server/backends/memory_metrics_collector/memory_metrics_collector.go
tempoz: Sure, sounds good.

PR #1806, opened 2022-03-30
Allow remote Run requests to specify OS & Arch.

vadimberezniker:
Default CLI to current OS & Arch.

---

**Version bump**: None

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -37,8 +37,8 @@ const (
 )
 
 var (
-	execOs            = flag.String("os", "", "If set, requests execution on a specific OS.")
-	execArch          = flag.String("arch", "", "If set, requests execution on a specific CPU architecture.")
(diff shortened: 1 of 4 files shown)


cli/remotebazel/remotebazel.go
```diff
@@ -33,6 +35,8 @@ const (
 )
 
 var (
+	execOs            = flag.String("os", "", "If set, requests execution on a specific OS.")
+	execArch          = flag.String("arch", "", "If set, requests execution on a specific CPU architecture.")
```
tylerwilliams, line 39: can we make linux x86_64 the flag default here since this is what we fully support? This will allow easily doing linux builds from mac and make this work out of the box more easily.

If/when we support mac and arm and other stuff we can flip the defaults back to the client's arch/os.

vadimberezniker: Done.

tylerwilliams (approved):

PR #1814, opened 2022-03-31
Gossip: Set multiple tags at once by default

tylerwilliams:
Because currently setting tags is synchronous (waits for gossip to broadcast the tag before returning) this makes setting multiple tags much faster (which we do a lot in our store tests).


siggisim (approved):
gossipman

PR #1815, opened 2022-04-01
Add bazel RBE integration tests

bduffany:
Motivation: This will give us greater confidence that any proposed changes to address https://github.com/buildbuddy-io/buildbuddy-internal/issues/1279 won't break the existing retry behavior.

This also buys us test coverage for the following Bazel <-> RBE integrations:
* Bazel's retry behavior based on the error codes that we are returning
* Bazel's stdout/stderr output displayed to the user based on the status codes that we return

Since these two behaviors are directly visible when using bazel with BB RBE, it seems worthwhile to have integration tests in place to enforce that these are working as intended.

---

**Version bump**: Patch


tylerwilliams (approved):
nice

PR #1818, opened 2022-04-01
Fix handling of action timeouts for docker actions

bduffany:
:point_right: **Easier to review with whitespace diff disabled https://github.com/buildbuddy-io/buildbuddy/pull/1818/files?diff=split&w=1**

---

Part of addressing https://github.com/buildbuddy-io/buildbuddy-internal/issues/1273

* Fix bug that stdout/stderr are not populated when using docker `Run()` and the action times out -- these are helpful for debugging the cause of the timeout, particularly when the timeout is caused by a problem with the test code.
* Fix bug that we get stuck indefinitely in `copyOutputs()` when using docker `Exec()` and the action times out.
* Add docker unit tests for both `Run()` and `Exec()`.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1273

### DIFF
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -33,7 +33,7 @@ import (
 	gstatus "google.golang.org/grpc/status"
 )
 
-var (
+const (
 	// dockerExecSIGKILLExitCode is returned in the ExitCode field of a docker
 	// exec inspect result when an exec process is terminated due to receiving
 	// SIGKILL.
@@ -479,6 +479,12 @@ func (r *dockerCommandContainer) exec(ctx context.Context, command *repb.Command
 		attachResp.Close()
 	}()
 	if err := copyOutputs(responseReader, result); err != nil {
+		// If we timed out, ignore the "closed connection" error from copying
+		// outputs
+		if ctx.Err() == context.DeadlineExceeded {
+			result.Error = status.DeadlineExceededError("command timed out")
+			return result
+		}
 		result.Error = wrapDockerErr(err, "failed to get output of exec process")
 		return result
 	}
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -154,6 +154,13 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		}()
 	}()
 
+	r.containerLogs(ctx, cid, result)
+	if result.Error != nil {
+		return result
+	}
+
+	// The container should not be running at this point since we're done copying
+	// logs, but we need to call ContainerWait to get the exit code.
 	statusCh, errCh := r.client.ContainerWait(ctx, cid, dockercontainer.WaitConditionNotRunning)
 
 	select {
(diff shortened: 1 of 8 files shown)


server/testutil/testenv/testenv.go
```diff
@@ -65,6 +65,9 @@ executor:
   # If we don't actually have the memory, we'll OOM, which is OK
   # for testing purposes.
   memory_bytes: 10_000_000_000
+  # Make sure Bazel can clean up the sandbox when running docker tests locally.
+  # Without this, containers run as root, so outputs can't be cleaned up.
+  docker_inherit_user_ids: true
```
tylerwilliams, line 70: @luluz66 @bduffany do we need to/will we support this in podman too?

enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
```diff
@@ -90,6 +91,10 @@ const (
 	defaultWaitTimeout = 20 * time.Second
 )
 
+var (
+	enableDocker = flag.Bool("rbetest.enable_docker", false, "Whether to enable docker for executors created via the rbetest framework")
```
tylerwilliams, line 95: since we're on the verge of switching to podman, can we find a way to use that instead?

tylerwilliams (changes requested):

server/testutil/testenv/testenv.go
bduffany: It looks like podman inherits user IDs by default, so we probably don't need to add an explicit option for it:

```shell
mkdir -p /tmp/foo
podman run -v /tmp/foo:/tmp/foo ubuntu:20.04 touch /tmp/foo/bar.txt
stat /tmp/foo/bar.txt
# Output: Access: (0644/-rw-r--r--)  Uid: ( 1000/       b)   Gid: ( 1000/       b)
```

vs docker

```shell
rm -rf /tmp/foo
mkdir -p /tmp/foo
docker run -v /tmp/foo:/tmp/foo ubuntu:20.04 touch /tmp/foo/bar.txt
stat /tmp/foo/bar.txt
# Output: Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)

```

enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
bduffany: Removed this flag for now since it's just for convenience when manual testing. Will handle podman as a separate PR, to keep this one small

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -166,27 +173,30 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
...
-		}
-		if err != nil {
-			result.Error = wrapDockerErr(err, "failed to read docker container logs")
-			return result
-		}
 	}
 	return result
 }
 
+func (r *dockerCommandContainer) containerLogs(ctx context.Context, cid string, result *interfaces.CommandResult) {
```
tylerwilliams, line 180: nit: would maybe be nice to call this something like copyContainerLogs or addContainerLogs to make it clear they are being read and added to the result?

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -33,7 +33,7 @@ import (
 	gstatus "google.golang.org/grpc/status"
 )
 
-var (
+const (
```
tylerwilliams, line 36: nice

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -462,7 +472,19 @@ func (r *dockerCommandContainer) exec(ctx context.Context, command *repb.Command
 	}
 
 	defer attachResp.Close() // note: Close() doesn't return an error.
+	go func() {
+		// If the context times out before we return from this func, close the
+		// attachResp -- otherwise copyOutputs() hangs.
+		<-ctx.Done()
+		attachResp.Close()
```
tylerwilliams, line 479: is it safe to call attachResp when the func returns before ctx.Done hits?

server/testutil/testenv/testenv.go
tylerwilliams: do we still need this here?

enterprise/server/remote_execution/containers/docker/docker.go
bduffany: good point, I mindlessly copied the docker API name. Done

enterprise/server/remote_execution/containers/docker/docker.go
bduffany: I didn't notice any problems in practice, but wrapped with a `sync.Once` just to be safe

server/testutil/testenv/testenv.go
bduffany: not really -- removed

enterprise/server/remote_execution/containers/docker/docker.go
bduffany: resolved offline - attachResp.Close() ultimately calls this code https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/http/httputil/persist.go;l=286-292;drc=cbf90b0028265a6195257e31a03aae42d402dd64 which is safe to call more than once

tylerwilliams (approved):

PR #1821, opened 2022-04-02
Fix handling of action timeouts for bare actions

bduffany:
In some cases, child processes of bare commands aren't killed when the context times out, and if the child process' output is wired up to stdout then we are stuck waiting for it to terminate, which can take much longer than the context deadline.

The fix is to kill by pgid when the context times out, which kills not only the top level process but also all child processes. The previous code relied on the default behavior of `CommandContext`, which is to kill the command by pid when the context times out, but this can leave child processes sticking around.

---

**Version bump**: Patch


enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -99,6 +104,42 @@ func Run(ctx context.Context, command *repb.Command, workDir string, stdin io.Re
...
+// exec.CommandContext(). Unlike exec.CommandContext.Run(), it kills the process
+// group when the context is done, instead of just killing the top-level
+// process. This helps ensure that orphaned processes aren't left running after
+// the command completes.
+//
+// It returns a FailedPrecondition error if cmd.SysProcAttr.Setpgid is not set
+// to true.
+func RunWithProcessGroupCleanup(ctx context.Context, cmd *exec.Cmd) error {
+	if cmd.SysProcAttr == nil || !cmd.SysProcAttr.Setpgid {
+		return status.FailedPreconditionError("process group cleanup requires SysProcAttr.Setpgid to be set")
```
tylerwilliams, line 120: is this ever true? when?

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -99,6 +104,42 @@ func Run(ctx context.Context, command *repb.Command, workDir string, stdin io.Re
 	}
 }
 
+// RunWithProcessGroupCleanup runs the given command, ensuring that child
+// processes are killed if the command times out.
+//
+// It is intended to be used with a command created via exec.Command(), not
```
tylerwilliams, line 110: nit: point to constructExecCommand?

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -99,6 +104,42 @@ func Run(ctx context.Context, command *repb.Command, workDir string, stdin io.Re
...
+// exec.CommandContext(). Unlike exec.CommandContext.Run(), it kills the process
```
tylerwilliams, line 111: just for my edification; what happens if a caller does run a exec.CommandContext cmd with this?

tylerwilliams (changes requested):

enterprise/server/remote_execution/commandutil/commandutil.go
bduffany: Done

enterprise/server/remote_execution/commandutil/commandutil.go
bduffany: It's not true in our usage, but I figured it's better to check preconditions since this func won't work as documented if Setpgid is not set, and the error does not surface in any other way.

I also considered having this func force-set Setpgid=true, but I figured it's more error prone for this thing to be silently mutating the command.

For more context, I'm also using this func here: https://github.com/buildbuddy-io/buildbuddy/pull/1822/files#diff-edc095b5def6cdcdacd2ca035eb690ff8ad3c3391d2eaeadfbe0e0fbc2f70b24R151

enterprise/server/remote_execution/commandutil/commandutil.go
bduffany: If that happens, there's a race between the `exec` lib and our goroutine to run `syscall.Kill(pid, syscall.SIGKILL)` and `syscall.Kill(-pgid, syscall.SIGKILL)` respectively. No matter the order, the Wait() call will exit with a `signal: killed` error, but child processes may or may not be cleaned up properly.

tylerwilliams (approved):

PR #1822, opened 2022-04-04
Fix handling of action timeouts for firecracker actions

bduffany:
* Run `vmexec.Exec()` commands with an explicit timeout that is less than the context deadline, so that we have enough time to send stdout/stderr back to the client, as well as copy outputs from the workspace
* Kill the executed command's process _group_ when the context is canceled, not just the process. Otherwise, orphaned child processes may prevent the `exec.Cmd.Wait()` call from completing, since the exec impl may be stuck trying to read all remaining stdout/stderr from those child processes.
* TODO (in a future PR): If the context deadline is exceeded, we can't guarantee that the VM is a good state (at least with the current design), so we need to ensure that the VM does not get added back to the runner pool in this case.

* [ ] TODO (before submitting): Rebuild initrd.cpio

---

**Version bump**: Patch


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -133,6 +136,16 @@ const (
 
 	// The path in the guest where VFS is mounted.
 	guestVFSMountDir = "/vfs"
+
+	// How much of the context deadline to allocate towards actually executing
+	// the command. The remaining fraction is dedicated allocated towards
+	// collecting outputs.
+	execDeadlineFraction = 0.85
```
tylerwilliams, line 143: Seems like a small fixed timeout would be a little simpler to calcuate / work with -- is a second or two enough time to budget for collecting outputs?

tylerwilliams (changes requested):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: Resolved offline -- it's probably less surprising to the user if we use a 1s timeout (changed to 1s).

tylerwilliams (approved):

PR #1828, opened 2022-04-04
Cancel RBE actions when invocation is interrupted.

vadimberezniker:
When a BES interrupted event is received, we ask the execution server to
cancel any pending executions for that invocation.

The execution server uses the data in the database to retrieve the list
of actions that are currently executing.

The execution server uses this list to ask the scheduler server to
cancel these tasks.

The scheduler server cancels the tasks by deleting them from Redis. The
task lease mechanism is updated to verify that the task still exists
when the lease is renewed.

---

**Version bump**: Patch


server/tables/tables.go
```diff
@@ -322,14 +322,14 @@ type Execution struct {
 	Worker      string
 	// Command Snippet
 	CommandSnippet          string
-	InvocationID            string `gorm:"index:execution_invocation_id"`
+	InvocationID            string `gorm:"index:executions_invocation_id_stage"`
```
tylerwilliams, line 325: we'll have to watch this during the release, this is our biggest table

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -679,6 +682,14 @@ func (e *EventChannel) handleEvent(event *pepb.PublishBuildToolEventStreamReques
 		SequenceNumber: event.OrderedBuildEvent.SequenceNumber,
 	}
 
+	if f, ok := bazelBuildEvent.Payload.(*build_event_stream.BuildEvent_Finished); ok {
```
tylerwilliams, line 685: nit: add a comment?

enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
```diff
@@ -381,6 +381,71 @@ func (s *BuildBuddyServer) GRPCPort() int {
...
+	opChan       chan *retpb.ControllerToCommandRequest
+	state        int
+	stateChanged chan struct{}
+}
+
+func (s *testCommandState) SetState(state int) {
+	s.mu.Lock()
+	s.state = state
+	if s.stateChanged != nil {
+		s.stateChanged <- struct{}{}
```
tylerwilliams, line 405: would it make sense to remove teh mutex and just shove the state down the pipe to your waiter (which can update the member variable if you need that?)

server/build_event_protocol/build_event_handler/build_event_handler.go
vadimberezniker: Done.

enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
vadimberezniker: Making the channel optional means we don't need to worry about whether or not there is currently a waiter on the channel. We could make the channel buffered, but then we need to worry about setting the capacity to make sure the channel doesn't get filled. Don't think we should get too caught up on this given this is test code.

server/tables/tables.go
vadimberezniker: Yeah, it's just slightly scary but should be fine. I will keep an eye on it during the release. I plan to merge this after tomorrow's release to give this at least a week to soak in dev.

tylerwilliams (approved):

PR #1830, opened 2022-04-05
Tune runner pool for Mac workflow workspace reuse

bduffany:
Mac workflow executors were only allowing up to 2 workspaces in the pool at once. Since rules_xcodeproj has 3 workflow actions, this was causing a new workspace to be created on every workflow run, resulting in slow workflows.

Instead of using the task size estimate to get pooled runner memory usage (8GB for workflow tasks), use the baseline mem estimate (2.6GB). Assuming 16GB memory, this increases the memory-based max pool count from 2 to 6. In practice, I was seeing bazel resident memory at about 1.2GB (after building buildbuddy enterprise server) and 1.8GB (after building bazel itself), so this 2.6GB estimate seems much more reasonable than 8GB.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1277

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -79,7 +79,7 @@ const (
 	defaultRunnerMemoryLimitBytes = tasksize.WorkflowMemEstimate
 	// Memory usage estimate multiplier for pooled runners, relative to the
 	// default memory estimate for execution tasks.
-	runnerMemUsageEstimateMultiplierBytes = 6.5
+	runnerMemUsageEstimateMultiplierBytes = 2
 
 	// Label assigned to runner pool request count metric for fulfilled requests.
 	hitStatusLabel = "hit"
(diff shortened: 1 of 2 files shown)


vadimberezniker:
What project was the `160MB` number for? That seems low for a more complex project (let's say the buildbuddy one). 
Want to double check that these changes work across different project types.

bduffany:
> What project was the `160MB` number for?

I must've been looking at the wrong process -- I tried building BB enterprise server and saw resident memory at about 1.2GB, and for building bazel itself, I was seeing 1.8GB.

I reverted the change to the default mem estimate and kept only the change to use the default mem estimate instead of using the task size estimate.

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -471,11 +471,10 @@ func (p *Pool) add(ctx context.Context, r *CommandRunner) *labeledError {
 			"stats_failed",
 		}
 	}
-	// If memory usage stats are not implemented, fall back to the task size
-	// estimate.
+	// If memory usage stats are not implemented, fall back to the default task
+	// size estimate.
 	if stats.MemoryUsageBytes == 0 {
-		estimate := tasksize.Estimate(r.task)
```
vadimberezniker, line 477: I'm having a hard time convincing myself that the newer code is more correct. What is the rationale for using the default task memory estimate as the basis of all task memory estimates?

If we think a workflow can use 8GB, then isn't it correct to say that we shouldn't allow more than 2 workflows in the pool? If the answer is that a workflow is not going to use 8GB then shouldn't we lower the workflow memory estimate so that we can fit more workflows in the pool? 

Sorry, not trying to be difficult. I know these memory estimates are a pain and perhaps I don't understand them well enough.

bduffany: I don't remember exactly, but I think the 8GB number accounts for the memory that bazel can be using while a build is in progress (e.g. local compile actions) while the effective 2.6GB "baseline" value is the memory used by bazel while it is sleeping between actions.

The name "default" is not great -- it's more of an "upperbound" estimate of the memory used by a paused, bare runner (for Docker, we can compute the memory of paused runners directly, and for Firecracker, the memory usage is zero since all resources are on disk, I think? Need to tweak that at some point, too)

vadimberezniker (approved):

PR #1832, opened 2022-04-05
Add firecracker support for workflows

bduffany:
Tested locally (with both cold and warm runs).

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"flag"
 	"fmt"
 	"io"
 	"net/http"
@@ -49,6 +50,8 @@ const (
 )
 
 var (
+	enableFirecracker = flag.Bool("remote_execution.workflows_enable_firecracker", false, "Whether to enable firecracker for Linux workflow actions.")
+
 	workflowURLMatcher = regexp.MustCompile(`^.*/webhooks/workflow/(?P<instance_name>.*)$`)
 
 	// ApprovalRequired is an error indicating that a workflow action could not be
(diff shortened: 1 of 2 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -694,7 +708,11 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 	return actionDigest, err
 }
 
-func (ws *workflowService) workflowsPoolName() string {
+func (ws *workflowService) workflowsPoolName(os string) string {
+	// When using firecracker, don't use the workflows pool.
+	if *enableFirecracker && (os == "" || os == platform.LinuxOperatingSystemName) {
+		return platform.DefaultPoolValue
```
vadimberezniker, line 714: I think it's still useful to have a separate pool for workflows+hosted bazel due to auto-scaling on the main pool. 
Any workflow or hosted bazel instances that ends up on a VM created after a scale-up will be lost if the pool scales down and will have to go through a cold start.

bduffany: Reverted this change

enterprise/server/workflow/service/service.go
```diff
@@ -627,10 +630,18 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
...
 	// Use the CI runner image if the OS supports containerized actions.
 	if os == "" || os == platform.LinuxOperatingSystemName {
 		containerImage = ws.workflowsImage()
+		if *enableFirecracker {
+			isolationType = string(platform.FirecrackerContainerType)
+			// When using Firecracker, write all outputs to the scratch disk, which
+			// has more space than the workspace disk and doesn't need to be extracted
+			// to the executor between action runs.
+			envVars = append(envVars, &repb.Command_EnvironmentVariable{Name: "WORKDIR_OVERRIDE", Value: "/root/workspace"})
+		}
```
tylerwilliams, line 644: would this be better to happen on the executor side, or do we not modify the action there?

bduffany: We don't modify the action on the executor, so it's probably cleaner to keep this logic here to avoid more special cases for workflow actions

tylerwilliams: mostly just thinking it's a tad annoying to have to know ahead of time, in the app, that we're going to be using firecracker in the executor and override something like the workdir which we have no context about. Feels like if this were doable on the executor, right before firecracker execution, that might be better. But I may be misunderstanding!

tylerwilliams: (discussed offline, i can live with this)

vadimberezniker (approved):

PR #1839, opened 2022-04-06
Use flags for auth settings

tempoz:
Moves the auth config into auth, the selfauth flag into selfauth, and the saml config into saml.
Biggest change is that auth config is now managed and retrieved through the Authenticator object,
which required interface changes, and that as a result, a nil authenticator is no longer a valid
state. Adds Register functions to move implementation details out of main.

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -17,6 +18,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
 	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
 	"github.com/buildbuddy-io/buildbuddy/server/util/capabilities"
+	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lru"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
@@ -34,6 +36,20 @@ import (
 	burl "github.com/buildbuddy-io/buildbuddy/server/util/url"
 	oidc "github.com/coreos/go-oidc"
 )
+
+var (
+	adminGroupID = flag.String("auth.admin_group_id", "", "ID of a group whose members can perform actions only accessible to server admins.")
+	enableAnonymousUsage = flag.Bool("auth.enable_anonymous_usage", false, "If true, unauthenticated build uploads will still be allowed but won't be associated with your organization.")
+	oauthProviders       = []config.OauthProvider{}
+	jwtKey               = flag.String("auth.jwt_key", "set_the_jwt_in_config", "The key to use when signing JWT tokens.")
+	apiKeyGroupCacheTTL  = flag.Duration("auth.api_key_group_cache_ttl", 0, "Override for the TTL for API Key to Group caching. Set to '0' to disable cache.")
+	httpsOnlyCookies     = flag.Bool("auth.https_only_cookies", false, "If true, cookies will only be set over https connections.")
+	disableRefreshToken  = flag.Bool("auth.disable_refresh_token", false, "If true, the offline_access scope which requests refresh tokens will not be requested.")
+)
+
+func init() {
+	flagutil.StructSliceVar(&oauthProviders, "auth.oauth_providers", "")
+}
 
 const (
 	// The key that the user object is stored under in the
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -11,10 +11,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/selfauth"
 	"github.com/buildbuddy-io/buildbuddy/server/config"
 	"github.com/buildbuddy-io/buildbuddy/server/endpoint_urls/build_buddy_url"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
+	"github.com/buildbuddy-io/buildbuddy/server/nullauth"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
 	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
 	"github.com/buildbuddy-io/buildbuddy/server/util/capabilities"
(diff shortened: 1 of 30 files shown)


enterprise/server/auth/auth.go
```diff
@@ -469,23 +484,24 @@ func newOpenIDAuthenticator(ctx context.Context, env environment.Env, oauthProvi
 		return nil, err
 	}
 
-	// Set the JWT key.
-	jwtKey = []byte(env.GetConfigurator().GetAuthJWTKey())
-
 	// Initialize API Key -> Group cache unless it's disabled by config.
 	var akgCache *apiKeyGroupCache
-	if env.GetConfigurator().GetAuthAPIKeyGroupCacheTTL() != "0" {
+	if *apiKeyGroupCacheTTL != time.Duration(0) {
```
vadimberezniker, line 489: This is an (unintentional) change in behavior. 
The cache is on by default and disabled if the config value is set to "0" (vs the default config value of "").
Now that we have default flag values you could set the default value of apiKeyGroupCacheTTL to 5 minutes and get rid of defaultAPIKeyGroupCacheTTL

enterprise/server/auth/auth.go
```diff
@@ -35,6 +39,20 @@ import (
...
+	enableAnonymousUsage = flag.Bool("auth.enable_anonymous_usage", false, "If true, unauthenticated build uploads will still be allowed but won't be associated with your organization.")
+	oauthProviders       = []config.OauthProvider{}
+	jwtKey               = flag.String("auth.jwt_key", "set_the_jwt_in_config", "The key to use when signing JWT tokens.")
+	apiKeyGroupCacheTTL  = flag.Duration("auth.api_key_group_cache_ttl", 0, "Override for the TTL for API Key to Group caching. Set to '0' to disable cache.")
+	httpsOnlyCookies     = flag.Bool("auth.https_only_cookies", false, "If true, cookies will only be set over https connections.")
+	disableRefreshToken  = flag.Bool("auth.disable_refresh_token", false, "If true, the offline_access scope which requests refresh tokens will not be requested.")
+)
+
+func init() {
+	flagutil.StructSliceVar(&oauthProviders, "auth.oauth_providers", "")
```
vadimberezniker, line 53: Can you add a doc string?

enterprise/server/auth/auth.go
```diff
@@ -704,9 +782,8 @@ func ClaimsFromSubID(env environment.Env, ctx context.Context, subID string) (*C
 	// role within the configured admin group, set their authenticated user to
 	// *only* have access to the org being impersonated.
 	if c := requestcontext.ProtoRequestContextFromContext(ctx); c != nil && c.GetImpersonatingGroupId() != "" {
-		adminGroupID := env.GetConfigurator().GetAuthAdminGroupID()
 		for _, membership := range claims.GetGroupMemberships() {
-			if membership.GroupID != adminGroupID || membership.Role != role.Admin {
+			if membership.GroupID != env.GetAuthenticator().AdminGroupID() || membership.Role != role.Admin {
```
vadimberezniker, line 786: Can we use `a.adminGroupID` here?

server/nullauth/nullauth.go
```diff
@@ -9,7 +9,45 @@ import (
...
+
+func (a *NullAuthenticator) AdminGroupID() string {
+	return a.adminGroupID
+}
+
+func (a *NullAuthenticator) AnonymousUsageEnabled() bool {
+	return !a.anonymousUsageDisabled
+}
+
+func (a *NullAuthenticator) EnableAnonymousUsage() {
```
vadimberezniker, line 32: Are these setters (enable, disable, set) used anywhere?

vadimberezniker (changes requested):

enterprise/server/auth/auth.go
tempoz: Done!

server/nullauth/nullauth.go
tempoz: Not anymore, apparently; removed.
Thanks for the catch!

enterprise/server/auth/auth.go
```diff
@@ -35,6 +39,21 @@ import (
 	oidc "github.com/coreos/go-oidc"
 )
 
+var (
+	adminGroupID         = flag.String("auth.admin_group_id", "", "ID of a group whose members can perform actions only accessible to server admins.")
+	enableAnonymousUsage = flag.Bool("auth.enable_anonymous_usage", false, "If true, unauthenticated build uploads will still be allowed but won't be associated with your organization.")
+	oauthProviders       = []config.OauthProvider{}
+	jwtKey               = flag.String("auth.jwt_key", "set_the_jwt_in_config", "The key to use when signing JWT tokens.")
+	apiKeyGroupCacheTTL  = flag.Duration("auth.api_key_group_cache_ttl", time.Duration(5)*time.Minute, "Override for the TTL for API Key to Group caching. Set to '0' to disable cache.")
```
vadimberezniker, line 47: 5 * time.Minute should be sufficient

vadimberezniker, line 47: Let's change "Override for the TTL for..." to "TTL for...".

vadimberezniker (approved):

enterprise/server/auth/auth.go
tempoz:
```suggestion
	apiKeyGroupCacheTTL  = flag.Duration("auth.api_key_group_cache_ttl", 5*time.Minute, "Override for the TTL for API Key to Group caching. Set to '0' to disable cache.")
```

tempoz:
```suggestion
	apiKeyGroupCacheTTL  = flag.Duration("auth.api_key_group_cache_ttl", time.Duration(5)*time.Minute, "TTL for API Key to Group caching. Set to '0' to disable cache.")
```

server/config/config.go
```diff
@@ -230,8 +230,8 @@ type cacheConfig struct {
 
 type authConfig struct {
 	JWTKey               string          `yaml:"jwt_key" usage:"The key to use when signing JWT tokens."`
-	APIKeyGroupCacheTTL  string          `yaml:"api_key_group_cache_ttl" usage:"Override for the TTL for API Key to Group caching. Set to '0' to disable cache."`
-	OauthProviders       []OauthProvider `yaml:"oauth_providers"`
+	APIKeyGroupCacheTTL  time.Duration   `yaml:"api_key_group_cache_ttl" usage:"Override for the TTL for API Key to Group caching. Set to '0' to disable cache."`
```
tempoz, line 233:
```suggestion
	APIKeyGroupCacheTTL  time.Duration   `yaml:"api_key_group_cache_ttl" usage:"TTL for API Key to Group caching. Set to '0' to disable cache."`
```

enterprise/server/auth/auth.go
tempoz:
```suggestion
	apiKeyGroupCacheTTL  = flag.Duration("auth.api_key_group_cache_ttl", time.Duration(5)*time.Minute, "TTL for API Key to Group caching. Set to '0' to disable cache.")
```

PR #1843, opened 2022-04-06
Add a CLI tool to fetch objects from cache and print their text representation

bduffany:
This is useful for fetching objects not associated with an invocation, for example workflows that failed to start, or when replaying actions in dev.

I've had this sitting in a branch for a while and have found a few uses for it, and a few other ppl have used it as well. Figured it's worth checking in at this point

---

**Version bump**: Patch

### DIFF
--- a/tools/cas/cas.go
+++ b/tools/cas/cas.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"flag"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/golang/protobuf/proto"
+	"google.golang.org/grpc/metadata"
+
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
+)
+
+var (
+	blobDigest   = flag.String("digest", "", "CAS digest")
+	instanceName = flag.String("remote_instance_name", "", "Remote instance name")
+	target       = flag.String("target", "", "CAS grpc target")
+	apiKey       = flag.String("api_key", "", "API key to attach to the outgoing context")
+
+	invocationID = flag.String("invocation_id", "", "Invocation ID. This is required when fetching the result of a failed action.")
+
+	blobType = flag.String("type", "", "Type of proto to inspect.")
+)
+
+func main() {
+	flag.Parse()
+	parts := strings.Split(*blobDigest, "/")
+	if len(parts) != 2 {
+		log.Fatalf("Invalid digest format: expecting {hash}/{size}, got %q", *blobDigest)
+	}
+	sizeBytes, err := strconv.Atoi(parts[1])
+	if err != nil {
+		log.Fatalf("Error parsing digest: %s", err)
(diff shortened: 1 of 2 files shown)


tools/cas/cas.go
```diff
@@ -0,0 +1,98 @@
...
+	blobDigest   = flag.String("digest", "", "Digest of the blob to fetch, in HASH/SIZE format.")
+	blobType     = flag.String("type", "", "Type of proto to inspect: Action, ActionResult, Command, stdout, or stderr.")
+	instanceName = flag.String("remote_instance_name", "", "Remote instance name")
+	target       = flag.String("target", "", "Cache grpc target, such as grpcs://remote.buildbuddy.io")
+	apiKey       = flag.String("api_key", "", "API key to attach to the outgoing context")
+
+	invocationID = flag.String("invocation_id", "", "Invocation ID. This is required when fetching the result of a failed action. Otherwise, it's optional.")
+)
+
+func main() {
```
luluz66, line 32: would be nice to have a comment showing example usage

bduffany: Done

tylerwilliams:
It'd be cool to find a sensible way to slot this into the bb cli tool. Maybe we can do that in a future cl.

tools/cas/cas.go
```diff
@@ -0,0 +1,120 @@
...
+//     bazel run //tools/cas -- -target=grpcs://remote.buildbuddy.dev -digest=HASH/SIZE -type=stderr
+//
+// Show a failed action result proto (requires invocation ID):
+//     bazel run //tools/cas -- -target=grpcs://remote.buildbuddy.dev -digest=HASH/SIZE -type=ActionResult -invocation_id=IID
+func main() {
+	flag.Parse()
+	if *target == "" {
+		log.Fatalf("Missing --target")
+	}
+	parts := strings.Split(*blobDigest, "/")
```
vadimberezniker, line 54: Use `digest.Parse`

tools/cas/cas.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		log.Fatalf("Error parsing digest: %s", err)
+	}
+	ind := digest.NewResourceName(&repb.Digest{
+		Hash:      parts[0],
+		SizeBytes: int64(sizeBytes),
+	}, *instanceName)
+	conn, err := grpc_client.DialTarget(*target)
+	if err != nil {
+		log.Fatalf("Error dialing CAS target: %s", err.Error())
```
vadimberezniker, line 68: just `err` (%s understands errors)

tools/cas/cas.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	bsClient := bspb.NewByteStreamClient(conn)
+	acClient := repb.NewActionCacheClient(conn)
+
+	ctx := context.Background()
+	if *apiKey != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-api-key", *apiKey)
```
vadimberezniker, line 75: Would be nice to add a utility function for this. We've been doing this in a bunch of different tools.

vadimberezniker (approved):
Awesome tool to have!

luluz66 (approved):

tools/cas/cas.go
bduffany: I agree -- would prefer do this as a separate PR though, probably at a later date when I have more time to work on CLI tools :)

tools/cas/cas.go
bduffany: Done

tools/cas/cas.go
bduffany: Done

tools/cas/cas.go
vadimberezniker: :+1:

PR #1848, opened 2022-04-07
Strip remote headers in cmd lines

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/server/util/redact/redact.go
+++ b/server/util/redact/redact.go
@@ -53,7 +53,7 @@ func stripURLSecrets(input string) string {
 	return urlSecretRegex.ReplaceAllString(input, "")
 }
 
-func stripURLSecretsFromCmdLine(tokens []string) []string {
+func stripURLSecretsFromCmdLine(tokens []string) {
 	for i, token := range tokens {
 		// Prevent flag name from being included in redaction pattern.
 		if strings.HasPrefix(token, "--") {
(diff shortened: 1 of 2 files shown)


siggisim (approved):

server/util/redact/redact.go
```diff
@@ -67,7 +67,20 @@ func stripURLSecretsFromCmdLine(tokens []string) []string {
...
 		tokens[i] = stripURLSecrets(token)
 	}
-	return tokens
+}
+
+func stripRemoteHeadersFromCmdLine(tokens []string) {
+	for i, token := range tokens {
+		// Prevent flag name from being included in redaction pattern.
+		if strings.HasPrefix(token, "--remote_header=") {
+			tokens[i] = "--remote_header=<REDACTED>"
```
luluz66, line 76: is it possible to just strip x-buildbuddy-platform.container-registry-username=_json_key --remote_header=x-buildbuddy-platform.container-registry-password? Currently, x-buildbuddy-api-key is already stripped. Some of the fields in remote_header are helpful in debugging. e.g. x-buildbuddy-platform.workload-isolation-type.

bduffany: I did it this way to be consistent with the structuredCommandLine redaction -- I can see the value in having an allowed list of remote headers to keep, though. Can chat more offline

PR #1849, opened 2022-04-07
More fixes for firecracker timeout handling

bduffany:
* We fail to collect stdout/stderr from certain bazel tests because the bazel `test-setup.sh` script runs the test child process as its own process group leader. So, [the fix in the last PR to kill processes by pgid](https://github.com/buildbuddy-io/buildbuddy/pull/1822/files#diff-ce918767a53b70309881b30d88900177b4791e488cee8d4449e1c1a01cab920aR113) was not sufficient. Fix this by explicitly killing the process tree by recursively inspecting child pids of the root command process. This fix also benefits the bare runner impl, not just firecracker.
* Move the action timeout calculation closer to when we actually send the exec request (after dialing the exec server), since dialing the exec server can eat up a significant amount of the timeout that would otherwise be allocated to collecting outputs.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/commandutil/BUILD
+++ b/enterprise/server/remote_execution/commandutil/BUILD
@@ -8,13 +8,16 @@ go_library(
     deps = [
         "//proto:remote_execution_go_proto",
         "//server/interfaces",
+        "//server/util/log",
         "//server/util/status",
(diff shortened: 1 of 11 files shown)


vadimberezniker (approved):

bduffany:
Note to self: rebuild initrd.cpio before submitting

PR #1858, opened 2022-04-08
Associate ci_runner execution with the invocation it's creating.

vadimberezniker:
It's a slightly odd link since the invocation is created by the
ci_runner, but it does give us an easy way to find the ci_runner
execution linked to a workflow run.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -24,6 +24,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/bazel_request"
 	"github.com/buildbuddy-io/buildbuddy/server/util/db"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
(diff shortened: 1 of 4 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -843,31 +849,40 @@ func (ws *workflowService) startWorkflow(webhookID string, r *http.Request) erro
...
+		return "", err
+	}
+
+	execCtx, err := bazel_request.WithRequestMetadata(ctx, &repb.RequestMetadata{ToolInvocationId: invocationID})
 	if err != nil {
 		return "", err
 	}
 
-	executionID, err := ws.env.GetRemoteExecutionService().Dispatch(ctx, &repb.ExecuteRequest{
+	opStream, err := ws.env.GetRemoteExecutionClient().Execute(execCtx, &repb.ExecuteRequest{
```
bduffany, line 869: does this stream need to be closed or will it get closed automatically?

bduffany (approved):
nice, I think this may also help with https://github.com/buildbuddy-io/buildbuddy-internal/issues/967

enterprise/server/workflow/service/service.go
vadimberezniker: Unfortunately since it's a one-directional receive stream it's not possible to close it from the client so it will hang around until the invocation is complete. This doesn't seem like something we need to worry about though, what do you think?

bduffany: Would it work to cancel the execCtx once we've recv'd the first msg? If not, seems OK to leave it running in the background.

vadimberezniker: Oh that's a great idea. Completely forgot that was an option.

PR #1861, opened 2022-04-08
Use mirror.gcr.io as a mirror and fallback to docker.io

luluz66:
The default registries.conf from podman will always use mirror.gcr.io
for docker.io. This may result in error when the image is on docker.io
but not on mirror.gcr.io

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1292


dockerfiles/executor_image/data/registries.conf
```diff
@@ -0,0 +1,21 @@
+# Note that changing the order here may break tests.
```
luluz66, line 1: The original file can be found at: https://github.com/containers/podman/blob/main/test/registries.conf

tylerwilliams (approved):

PR #1872, opened 2022-04-11
Add initial framework for driver (read-only to start)

tylerwilliams:
This also removes some of the old FindNodes code that lived in driver.go and associated tests.

### DIFF
--- a/enterprise/server/raft/driver/driver.go
+++ b/enterprise/server/raft/driver/driver.go
@@ -1,0 +1,435 @@
+package driver
+
+import (
+	"context"
+	"sort"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/raft/constants"
+	"github.com/buildbuddy-io/buildbuddy/server/gossip"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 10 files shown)


enterprise/server/raft/driver/driver.go
```diff
@@ -0,0 +1,435 @@
...
+	log.Printf("Dead replicas: %+v, my clusters: %+v", deadReplicas, leasedClusterIDs)
+	// AddClusterNode()
+
+	overloadedReplicas := d.clusterMap.OverloadedReplicas(leasedClusterIDs, d.opts.MaxReplicaSizeBytes)
+	log.Printf("Overloaded replicas: %+v, my clusters: %+v", overloadedReplicas, leasedClusterIDs)
+	// SplitCluster()
+
+	excessReplicas := d.clusterMap.ExcessReplicas(leasedClusterIDs, rsp.GetNode().GetNhid())
+	log.Printf("Excess replicas: %+v, my clusters: %+v", excessReplicas, leasedClusterIDs)
+	// AddClusterNode && RemoveClusterNode
```
siggisim, line 432: intentional? and above

siggisim (approved):

enterprise/server/raft/driver/driver.go
tylerwilliams: yeah in a followup that's where I'm going to actually call those methods, just is a reminder for me.

PR #1875, opened 2022-04-12
Sample before incurring the allocs

tylerwilliams:
In https://github.com/buildbuddy-io/buildbuddy/pull/1873 I noticed that the interceptors added significant overhead, in terms of # of allocs, to each request. (80% more bytes and 61% more allocs, even for untraced requests). Most of that I found to be coming from otelgrpc.

This CL improves that, by only enabling the otelgrpc tracing interceptor on requests that will be traced. A request should be traced if it is already being traced, it falls into the sampling fraction, or tracing was forced. 

With this change, the overall interceptor overhead drops to (20% more bytes and 15% more allocs).

BenchmarkBare-32                  	   10000	    103698 ns/op	    8675 B/op	     172 allocs/op
BenchmarkInstrumentedClient-32    	    9474	    107735 ns/op	    9055 B/op	     181 allocs/op
BenchmarkInstrumentedServer-32    	    9475	    115181 ns/op	   10081 B/op	     190 allocs/op
BenchmarkInstrumentedAll-32       	    9328	    119766 ns/op	   10461 B/op	     199 allocs/op


server/util/tracing/tracing.go
```diff
@@ -309,3 +245,53 @@ func AddStringAttributeToCurrentSpan(ctx context.Context, key, value string) {
...
+	}
+	return nil
+}
+
+// shouldTrace returns a boolean indicating if request should be traced because:
+//  - tracing was forced
+//  - some other service is tracing this request already
+//  - sampling indicates this request should be traced
+func ShouldTrace(ctx context.Context, method string) bool {
+	grpcMD, ok := metadata.FromIncomingContext(ctx)
```
tylerwilliams, line 270: @vadimberezniker i'm a little unsure if this is right for the client interceptors -- should those be using FromOutgoingContext?

tylerwilliams: Fixed.

vadimberezniker (approved):

PR #1884, opened 2022-04-13
Small release script improvements

bduffany:
* Only support python3 (I tested today with python3 and it works, so let's just stick with one python version so we don't have to worry about multiple)
* Stream live command output
* Update messaging at the end of the script

---

**Version bump**: Patch

### DIFF
--- a/release.py
+++ b/release.py
@@ -1,29 +1,14 @@
-#!/usr/bin/env python
-from __future__ import print_function
-
+#!/usr/bin/env python3
 import argparse
-import glob
 import json
 import os
-import re
 import subprocess
 import sys
 import tempfile
 import time
-
-# Don't care about py2/py3 differences too much.
-try:
-    from urllib.parse import urlencode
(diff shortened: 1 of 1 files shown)


vadimberezniker (approved):

release.py
```diff
@@ -270,15 +254,17 @@ def main():
...
     update_docker_image(new_version, update_latest_tag)
+    print("Pushed docker image for new version %s" % new_version)
 
     ## Don't need this because github automatically creates a source archive when we
     ## make a new tag. Useful when we have artifacts to upload.
     #artifacts = build_artifacts(repo_name, new_version)
     #create_release_and_upload_artifacts("/".join([org_name, repo_name]), new_version, artifacts)
 
-    print("Release (%s) complete. Go enjoy a cold one!" % new_version)
+    print("Done -- proceed with the release guide!")
```
tylerwilliams, line 267: 🍻

tylerwilliams (approved):

PR #1885, opened 2022-04-14
Update protobuf libraries, as old ones are deprecated

tempoz:
As stated in https://pkg.go.dev/github.com/golang/protobuf, github.com/golang/protobuf is deprecated.
In moving to go 1.18, we are encountering some difficulty with dependencies and trying to modernize
the dependency graph where we can.

The largest difference here is the missing `proto.Buffer` struct, which was used in the runner_test
to write protos prefixed with sizes. This has been updated to just use binary.PutUvarint.

Other than that, many of the places where errors were possible in converting from proto, errors are
no longer an option, which allowed streamlining and simplifying of some code.

---

**Version bump**: Minor


tempoz:
* master
  * **PR #1885** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1885" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1885?utm_source=stack-comment).

tylerwilliams (approved):

PR #1887, opened 2022-04-14
Fix missing docker action stderr/stdout

bduffany:
---

**Version bump**: Patch


tylerwilliams (approved):
LG for fixing prod -- can we add a test in a followup?

bduffany:
> LG for fixing prod -- can we add a test in a followup?

Yep, will do

PR #1889, opened 2022-04-14
Add regression tests for container stdout/stderr collection

bduffany:
* Verified that the docker test fails when `Follow: true` in `ContainerLogsOptions` is commented out (only the first line, "Hello world" is displayed, but the "Hello again" printed after a delay is missing)
* Ported the docker test to podman as well (even though podman didn't have any issues) since it is a nice test to have.

---

**Version bump**: Patch


tylerwilliams (approved):
nice, thanks for adding these

PR #1893, opened 2022-04-15
Allow specifying--dev flag for exec containers

AleksanderGondek:
This change introduces the capability to pass devices from host machine to execution containers, which are being spawned by the build-buddy.

The rationale: adding support for `/dev/fuse` or any other host-bound devices during Bazel RBE execution.

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -18,6 +18,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 
+	config "github.com/buildbuddy-io/buildbuddy/server/config"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 )
 
@@ -38,6 +39,7 @@ type PodmanOptions struct {
 	ForceRoot bool
 	Network   string
 	CapAdd    string
+	Devices   []config.DockerDeviceMapping
 	Runtime   string
 }
 
@@ -95,6 +97,9 @@ func (c *podmanCommandContainer) getPodmanRunArgs(workDir string) []string {
 	}
 	if c.options.CapAdd != "" {
 		args = append(args, "--cap-add="+c.options.CapAdd)
+	}
+	for _, device := range c.options.Devices {
+		args = append(args, "--device="+device.PathOnHost)
 	}
 	if c.options.Runtime != "" {
 		args = append(args, "--runtime="+c.options.Runtime)
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -18,8 +18,8 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 
-	config "github.com/buildbuddy-io/buildbuddy/server/config"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	config "github.com/buildbuddy-io/buildbuddy/server/config"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


server/config/config.go
```diff
@@ -288,6 +288,12 @@ type RemoteExecutionConfig struct {
 	EnableRedisAvailabilityMonitoring bool               `yaml:"enable_redis_availability_monitoring" usage:"If enabled, the execution server will detect if Redis has lost state and will ask Bazel to retry executions."`
 }
 
+type DockerDeviceMapping struct {
+	PathOnHost         string  `yaml:"path_on_host" usage:"path to device that should be mapped from the host."`
+	PathInContainer    string  `yaml:"path_in_container" usage:"path under which the device will be present in container."`
+	CgroupPermissions  string  `yaml:"cgroup_permissions" usage:"cgroup permissions that should be assinged to device."`
```
tylerwilliams, line 294: s/assinged/assigned/g

tylerwilliams (changes requested):
Hey @AleksanderGondek, thanks for the CL!

I noticed a small typo, but if you fix that we're good to go :)

server/config/config.go
AleksanderGondek: Thank you for catching this - typos are bane of my existence! Fixed.

AleksanderGondek:
@tylerwilliams thank you for welcoming me to bbbuddy contribution world ;) 

Any tips on how to approach those failing tests? I see that raft testing suite is having a worse day and there is an error message about “runner recycling being not supported for anonymous builder”.

tylerwilliams (approved):

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -96,6 +98,9 @@ func (c *podmanCommandContainer) getPodmanRunArgs(workDir string) []string {
 	if c.options.CapAdd != "" {
 		args = append(args, "--cap-add="+c.options.CapAdd)
 	}
+	for _, device := range c.options.Devices {
+		args = append(args, "--device="+device.PathOnHost)
```
bduffany, line 102: I think podman (like docker) uses the `hostpath:containerpath:perms` flag format -- something like this should probably work?

```suggestion
		deviceFlag := device.PathOnHost
		if device.PathInContainer != "" {
				deviceFlag += ":" + device.PathInContainer
				if device.CgroupPermissions != "" {
						deviceFlag += ":" + device.CgroupPermissions
				}
		}
		args = append(args, "--device="+deviceFlag)
```

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -96,6 +98,19 @@ func (c *podmanCommandContainer) getPodmanRunArgs(workDir string) []string {
...
 		args = append(args, "--cap-add="+c.options.CapAdd)
 	}
+	for _, device := range c.options.Devices {
+		deviceSpecs := make([]string, 0)
+		if device.PathOnHost != "" {
+			deviceSpecs = append(deviceSpecs, device.PathOnHost)
+		}
+		if device.PathInContainer != "" {
+			deviceSpecs = append(deviceSpecs, device.PathInContainer)
+		}
```
bduffany, line 108: If PathInContainer is unspecified, we should set it to match the path on host, otherwise if cgroup permissions are specified, the flag will wind up as `/host/path:cgroupperms` and the `cgroupperms` will be parsed as the container path:

```suggestion
		if device.PathInContainer != "" {
			deviceSpecs = append(deviceSpecs, device.PathInContainer)
		} else {
			deviceSpecs = append(deviceSpecs, device.PathOnHost)
		}
```

bduffany:
> Any tips on how to approach those failing tests? I see that raft testing suite is having a worse day and there is an error message about “runner recycling being not supported for anonymous builder”.

This is happening because we currently run workflows on PRs from external contributors as anonymous builds. The PR looks solid and unlikely to break docker tests, but I'll double check by running the workflows manually before submitting :)

enterprise/server/remote_execution/containers/podman/podman.go
AleksanderGondek: Before implementation, I have took a look at podman testing suite: [relevant link](https://github.com/containers/podman/blob/418ab2e5e154cb0df13f2e901cf4aea31a5cd0a9/test/e2e/run_device_test.go#L63)
It is my understanding that without PathInContainer, podman automatically maps it to same path as PathInHost and that writting CGroupPerms at second place (instead of PathInContainer) is supported. 

Please let me know if that works or should I implement the execution flow accordingly to your suggestion :)

bduffany: Thank you for looking more deeply into that! LGTM as-is

bduffany (approved):

bduffany:
Docker tests LGTM locally, so this is good to merge :+1:

PR #1896, opened 2022-04-15
Runner side changes to extract run target metadata.

vadimberezniker:
If the `--record_run_metadata` flag is passed to the runner, we perform
the bazel run with the `--script_path` argument which does all the
normal run preparation followed by writing a run script without actually
running it.

We then parse the generated script to extract metadata necessary for the
CLI to be able to run the binary:
 - the default arguments
 - walk the runtime files and make sure they are in the CAS

This information is then recorded in the build event stream using the
new RunTargetAnalyzed event.

---

**Version bump**: None


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -109,6 +113,7 @@ var (
 	visibility         = flag.String("visibility", "", "If set, use the specified value for VISIBILITY build metadata for the workflow invocation.")
 	bazelSubCommand    = flag.String("bazel_sub_command", "", "If set, run the bazel command specified by these args and ignore all triggering and configured actions.")
 	patchDigests       = flagutil.StringSlice("patch_digest", []string{}, "Digests of patches to apply to the repo after checkout. Can be specified multiple times to apply multiple patches.")
+	recordRunMetadata  = flag.Bool("record_run_metadata", false, "Instead of running a target, extract metadata about it and store it in the build event stream.")
```
tylerwilliams, line 116: nit: report it in the build event stream

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -751,12 +756,47 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
...
+		// extract run options (i.e. args, runfile information) from the generated run script.
+		runScript := ""
+		if *recordRunMetadata {
+			tmpDir, err := os.MkdirTemp("", "bazel-run-script-*")
+			if err != nil {
+				return err
+			}
+			defer func() {
+				_ = os.RemoveAll(tmpDir)
+			}()
```
tylerwilliams, line 769: defer os.RemoveAll(tmpDir) ?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -808,6 +848,169 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 	return nil
 }
 
+type runInfo struct {
+	args         []string
+	runfiles     []*bespb.File
+	runfilesRoot string
+}
+
+func uploadRunfiles(ctx context.Context, workspaceRoot, runfilesDir string) ([]*bespb.File, error) {
```
tylerwilliams, line 857: we've collectively spent quite a lot of time writing code to upload files to the CAS. Is there a way we could re-use some code we've written before to do this rather than writing this out? Can we use something like UploadDirectoryToCas(/path/to/dir)?

proto/build_event_stream.proto
```diff
@@ -1248,6 +1253,16 @@ message ChildInvocationCompleted {
   google.protobuf.Duration duration = 5;
 }
 
+// Event describing the runtime properties of a runnable target.
+message RunTargetAnalyzed {
+  // Default arguments configured for the target, typically in the BUILD file.
+  repeated string arguments = 1;
```
tylerwilliams, line 1259: nit: comment add example form of an argument?

tylerwilliams (changes requested):

tylerwilliams (approved):
LGTM (can you break up uploadRunfiles into two parts maybe?)

vadimberezniker:
> LGTM (can you break up uploadRunfiles into two parts maybe?)

split up. i will look into refactoring the code to reuse some of the code.

PR #1907, opened 2022-04-20
Scorecard: Implement GetCacheScoreCard RPC

bduffany:
This new RPC will allow sorting/filtering/paging cache results for the new scorecard UI, and also will be used to fetch the initial page of scorecard results once the invocation is complete (we won't continuously fetch while the invocation is in progress, to avoid too much load on Redis).

---

**Version bump**: Patch


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -69,6 +73,9 @@ const (
 
 	// Exit code in Finished event indicating that the build was interrupted (i.e. killed by user).
 	InterruptedExitCode = 8
+
+	// How many results to return at a time for the cache scorecard.
+	scoreCardPageSize = 250
```
vadimberezniker, line 78: Call this defaultScoreCardPageSize since the page size can be specified in request?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -1034,6 +1048,82 @@ func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*in
 	return invocation, nil
 }
 
+func GetCacheScoreCard(ctx context.Context, env environment.Env, req *capb.GetCacheScoreCardRequest) (*capb.GetCacheScoreCardResponse, error) {
+	page := &pgpb.OffsetLimit{Offset: 0, Limit: scoreCardPageSize}
+	if req.PageToken != "" {
+		var err error
+		page, err = paging.DecodeOffsetLimit(req.PageToken)
```
vadimberezniker, line 1055: nit: I would create a local var and re-assign to page as it makes it more clear what's going on. i.e.
reqPage, err := ...
if err != nil {}
page = reqPage

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -1034,6 +1048,82 @@ func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*in
...
+		}
+	}
+
+	scorecard, err := readScoreCard(ctx, env, req.InvocationId)
+	if err != nil {
+		return nil, err
+	}
+
+	end := page.Offset + page.Limit
+	scorecard.Results = scorecard.Results[int(page.Offset):end]
```
vadimberezniker, line 1067: you're validating the end value **after** subsetting the slice
it appears you are also not validating page.Offset index

vadimberezniker (changes requested):

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -1034,6 +1048,82 @@ func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*in
 	return invocation, nil
 }
 
+func GetCacheScoreCard(ctx context.Context, env environment.Env, req *capb.GetCacheScoreCardRequest) (*capb.GetCacheScoreCardResponse, error) {
```
tylerwilliams, line 1051: can we move this scorecard searching / sorting logic to its own file (called scorecard.go?)

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Done

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Done

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: sg, done

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: woops, messed this up when refactoring. Fixed

server/remote_cache/scorecard/scorecard.go
```diff
@@ -0,0 +1,139 @@
...
+		NextPageToken: nextPageToken,
+	}, nil
+}
+
+func isTimestampLess(a, b *timestamppb.Timestamp) bool {
+	if a.Seconds != b.Seconds {
+		return a.Seconds < b.Seconds
+	}
+	return a.Nanos < b.Nanos
+}
```
tylerwilliams, line 76: Methods like these are kinda tricky -- I never know if a is less than b or b is less than a.

Can you instead do:

a.AsTime().Before(b.AsTime())

*inline* in all of the callsites rather than using this helper?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -979,13 +954,17 @@ func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*in
...
+		// invocation.
+		if !hit_tracker.DetailedStatsEnabled() {
+			// The cache ScoreCard is not stored in the table invocation, so we do this lookup
+			// after converting the table invocation to a proto invocation.
+			if ti.InvocationStatus == int64(inpb.Invocation_PARTIAL_INVOCATION_STATUS) {
+				scoreCard = hit_tracker.ScoreCard(ctx, env, iid)
+			} else {
+				if sc, err := scorecard.Read(ctx, env, iid); err == nil {
+					scoreCard = sc
+				}
```
tylerwilliams, line 967: we probably want to log this error rather than swallowing it

server/remote_cache/scorecard/scorecard.go
```diff
@@ -0,0 +1,139 @@
...
+
+// SortResults sorts scorecard results, grouping by actions sorted by min
+// request start timestamp, and sorting the requests within each action by start
+// time.
+// TODO(bduffany): More sorting options
+func SortResults(results []*capb.ScoreCard_Result) {
+	minStartTimeByActionID := map[string]*timestamppb.Timestamp{}
```
tylerwilliams, line 83: minStartTimeByActionID := make(map[string]*timestamppb.Timestamp{}, len(results))

(my hot take is that it's *always* good to use make)

server/remote_cache/scorecard/scorecard.go
```diff
@@ -0,0 +1,139 @@
...
+		if isTimestampLess(ti, tj) {
+			return true
+		}
+		// If two different actions have the same start time, break the tie using
+		// their action ID, so that results stay grouped by action ID.
+		if !isTimestampLess(tj, ti) && results[i].ActionId != results[j].ActionId {
+			return results[i].ActionId < results[j].ActionId
+		}
+		// Within a single action, sort by start time.
+		return isTimestampLess(results[i].StartTime, results[j].StartTime)
```
tylerwilliams, line 107: seems like this logic could be simplified

server/remote_cache/scorecard/scorecard.go
tylerwilliams, line 83: nit: maybe just call this minStartTime?

(I know I set a poor example here and often write wordy dict names)

server/remote_cache/scorecard/scorecard.go
```diff
@@ -0,0 +1,139 @@
...
+	})
+}
+
+func blobName(invocationID string) string {
```
tylerwilliams, line 111: Add a comment warning that things will break if this is changed because data is stored under this name?

tylerwilliams (changes requested):

server/remote_cache/scorecard/scorecard.go
bduffany: Simplified it a bit -- I think we do need three comparisons after all though (need to break ties between two actions that happen to have the same start time).

server/remote_cache/scorecard/scorecard.go
bduffany: Done

server/remote_cache/scorecard/scorecard.go
bduffany: Done

bduffany: I didn't set a capacity here because the number of actions can be significantly less than the number of results. Switched to `make()`.

server/remote_cache/scorecard/scorecard.go
bduffany: Done -- also switched the map to use `time.Time` so we don't have to deal with the protos as much.

server/build_event_protocol/build_event_handler/build_event_handler.go
bduffany: Done

tylerwilliams (approved):

vadimberezniker (approved):

PR #1908, opened 2022-04-20
handle podman error properly

luluz66:
Need to check exit code in addition to commndResult.Error in some of the
APIs in podman.go

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -247,17 +260,32 @@ func (c *podmanCommandContainer) Remove(ctx context.Context) error {
...
+		return res.Error
+	}
+	if res.ExitCode == 0 || strings.Contains(string(res.Stderr), "no such container") {
+		return nil
+	}
+	return status.UnknownErrorf("podman remove failed: exit code %d, stderr: %s", res.ExitCode, string(res.Stderr))
 }
 
 func (c *podmanCommandContainer) Pause(ctx context.Context) error {
 	res := runPodman(ctx, "pause", nil /*=stdin*/, nil /*=stdout*/, c.name)
```
bduffany, line 273: nit: `if res.Error!=nil` then return the existing error? (like in Unpause)

bduffany (approved):
LGTM, just 1 nit

enterprise/server/remote_execution/containers/podman/podman.go
luluz66: done

tylerwilliams (approved):

PR #1909, opened 2022-04-20
handle podman error properly

luluz66:
Need to check exit code in addition to commndResult.Error in some of the
APIs in podman.go

Same as
https://github.com/buildbuddy-io/buildbuddy/pull/1908


luluz66:
For some reasons, the same code cannot compile in https://github.com/buildbuddy-io/buildbuddy/pull/1908 CIs because it thinks some functions are declared twice. Create a new PR

tylerwilliams (approved):

PR #1914, opened 2022-04-21
Fix error code reported for context errors

bduffany:
---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1321

### DIFF
--- a/enterprise/server/remote_execution/executor/executor.go
+++ b/enterprise/server/remote_execution/executor/executor.go
@@ -266,7 +266,7 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, task *repb.E
 	// Exit codes < 0 mean that the command either never started or was killed.
 	// Make sure we return an error in this case.
 	if cmdResult.ExitCode < 0 {
-		cmdResult.Error = incompleteExecutionError(cmdResult.ExitCode, cmdResult.Error)
+		cmdResult.Error = incompleteExecutionError(ctx, cmdResult.ExitCode, cmdResult.Error)
 	}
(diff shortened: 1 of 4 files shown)


siggisim:
I'm a little nervous that we're currently retrying a lot of these errors (because they return to bazel as Unavailable) but now we won't be retrying them (because we return DeadlineExceeded).

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -349,11 +349,26 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, task *repb.E
...
 	if err == nil {
 		alert.UnexpectedEvent("incomplete_command_with_nil_error")
 		return status.UnknownErrorf("Command did not complete, for unknown reasons (internal status %d)", exitCode)
 	}
+	// If the context timed out or was cancelled, ignore any error returned from
+	// the command since it's most likely caused by the context error, and we
+	// don't want to require container implementations to have to handle these
+	// cases explicitly.
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		log.Infof("Ignoring command error likely caused by %s: %s", ctxErr, err)
```
tylerwilliams, line 362: can you make sure this is a sufficient log level for debugging this in prod?

tylerwilliams (approved):
This LGTM -- make sure @siggisim is cool with it before submitting.

We can babysit it carefully / yank if necessary.

vadimberezniker (approved):

siggisim (approved):
Thanks!!

PR #1915, opened 2022-04-21
Pass through extra args to `bazel run` to the downloaded binary.

vadimberezniker:
If the user executes `bazel run //target -- --foo=bar`, we need to pass
`--foo=bar` to the downloaded binar.

When parsing the command line args, we split them into regular &
passthrough slices (stored in `BazelArgs` struct). 

This also fixes a missing `defer` in the ci_runner (introduced in the
last PR).

---

**Version bump**: None

### DIFF
--- a/cli/autoconfig/autoconfig.go
+++ b/cli/autoconfig/autoconfig.go
@@ -36,6 +36,15 @@ type BazelOpts struct {
 	WorkspaceFilePath  string
 	EnableRemoteBazel  bool
 	SidecarSocket      string
+
+	// If the user is trying to use build events, remote caching or execution but did not specify the endpoints on the
+	// command line, these fields will contain the automatically configured values that should be passed to Bazel.
+	BESBackendOverride     string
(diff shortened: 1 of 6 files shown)


cli/autoconfig/autoconfig.go
```diff
@@ -36,6 +36,15 @@ type BazelOpts struct {
 	WorkspaceFilePath  string
 	EnableRemoteBazel  bool
 	SidecarSocket      string
+
+	// If the user is trying to use build events, remote caching or execution but did not specify the endpoints on the
+	// command line, these fields will contain the automatically configured values that should be passed to Bazel.
```
tylerwilliams, line 41: nit: can you make these 80 chars?

tylerwilliams (approved):

cli/autoconfig/autoconfig.go
vadimberezniker: Done.

PR #1920, opened 2022-04-22
Add sorting, filtering, and grouping options to scorecard proto

bduffany:
---

**Version bump**: TODO

### DIFF
--- a/proto/cache.proto
+++ b/proto/cache.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 package cache;
 
 import "google/protobuf/duration.proto";
+import "google/protobuf/field_mask.proto";
 import "google/protobuf/timestamp.proto";
 import "google/rpc/status.proto";
 import "proto/context.proto";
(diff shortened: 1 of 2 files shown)


vadimberezniker (approved):

tylerwilliams (approved):

proto/cache.proto
```diff
@@ -46,6 +47,64 @@ message GetCacheScoreCardRequest {
...
+    // ResponseType represents a category of RPC response codes.
+    enum ResponseType {
+      UNKNOWN_RESPONSE_TYPE = 0;
+      // Show only successful responses (OK response).
+      OK = 1;
+      // Show only NOT_FOUND responses.
+      NOT_FOUND = 2;
+      // Show only error responses (excluding NOT_FOUND).
+      ERROR = 3;
+    }
```
tylerwilliams, line 70: Should this be at the top level too?

Could we use a builtin grpc status proto for this...?

tylerwilliams (approved):

proto/cache.proto
bduffany: moved to the top-level -- resolved offline, but for future readers: this proto is used because we want the user to be able to filter NOT_FOUND and "other" errors separately.

PR #1922, opened 2022-04-22
Bump timeout in TestRedisRestart test.

vadimberezniker:
In the failing cases seen so far, it takes Bazel more than 10 seconds to
send any remote execution requests causing us to fail to find the remote
execution task information.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1324

---

**Version bump**: None


bduffany (approved):
Thanks for fixing this!

PR #1924, opened 2022-04-22
Add initrd.cpio to build using pure Go cpio lib

bduffany:
This keeps initrd.cpio in sync with its transitive deps, and avoids adding 40MB to the repo history on each update.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/tools/cpio/cpio.go
+++ b/enterprise/tools/cpio/cpio.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"flag"
+	"io"
+	"os"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/cavaliergopher/cpio"
+)
+
+// This tool allows creating cpio archives in SVR4 (newc) format.
+// Example:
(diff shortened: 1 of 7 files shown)


enterprise/tools/cpio/cpio.go
```diff
@@ -0,0 +1,60 @@
...
+		check(err)
+		hdr := &cpio.Header{Name: path, Mode: cpio.FileMode(s.Mode()), Size: s.Size()}
+		err = w.WriteHeader(hdr)
+		check(err)
+		f, err := os.Open(path)
+		check(err)
+		defer f.Close()
+		b, err := io.ReadAll(f)
+		check(err)
+		_, err = w.Write(b)
```
tylerwilliams, line 57: _, err := copy(w, f) would be better, so we don't need all of this data in memory

tylerwilliams (approved):
nice

(1 nit)

enterprise/tools/cpio/cpio.go
bduffany: good call, done

vadimberezniker (approved):

PR #1925, opened 2022-04-22
Keep sidecar alive while CLI is running.

vadimberezniker:
We lean on the sidecar in remote bazel for caching artifacts but the
sidecar can terminate due to inactivity during a long remote bazel
build.

This PR avoids this by continuously pinging the sidecar while the CLI is
running.

---

**Version bump**: None

### DIFF
--- a/cli/cmd/bb/bb.go
+++ b/cli/cmd/bb/bb.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"time"
 
 	"github.com/bazelbuild/bazelisk/core"
 	"github.com/bazelbuild/bazelisk/repositories"
@@ -16,9 +17,11 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/cli/parser"
 	"github.com/buildbuddy-io/buildbuddy/cli/remotebazel"
 	"github.com/buildbuddy-io/buildbuddy/cli/sidecar"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
 	"github.com/buildbuddy-io/buildbuddy/server/version"
 
 	bblog "github.com/buildbuddy-io/buildbuddy/cli/logging"
+	scpb "github.com/buildbuddy-io/buildbuddy/proto/sidecar"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


cli/cmd/sidecar/sidecar.go
```diff
@@ -178,6 +179,17 @@ func registerCacheProxy(ctx context.Context, env *real_environment.RealEnv, grpc
...
+}
+
+func (s *sidecarService) Ping(ctx context.Context, req *scpb.PingRequest) (*scpb.PingResponse, error) {
+	return &scpb.PingResponse{}, nil
+}
+
+func registerSidecarService(grpcServer *grpc.Server) {
+	scpb.RegisterSidecarServer(grpcServer, &sidecarService{})
+}
+
```
tylerwilliams, line 192: any reason this is a different method?

cli/cmd/sidecar/sidecar.go
```diff
@@ -178,6 +179,17 @@ func registerCacheProxy(ctx context.Context, env *real_environment.RealEnv, grpc
 	repb.RegisterCapabilitiesServer(grpcServer, cacheProxy)
 }
 
+type sidecarService struct {
+}
```
tylerwilliams, line 183: nit: type sidecarService struct {}

cli/cmd/bb/bb.go
```diff
@@ -88,6 +91,43 @@ func parseBazelRCs(bazelFlags *commandline.BazelFlags) []*parser.BazelOption {
...
+		return err
+	}
+	s := scpb.NewSidecarClient(conn)
+	connected := make(chan struct{})
+	go func() {
+		connectionValidated := false
+		for {
+			_, err := s.Ping(ctx, &scpb.PingRequest{})
+			if connectionValidated && err != nil {
+				fmt.Printf("sidecar did not response to ping request: %s\n", err)
```
tylerwilliams, line 109: s/response/respond/g

can we use some form of log here instead of fmt?

cli/cmd/bb/bb.go
```diff
@@ -88,6 +91,43 @@ func parseBazelRCs(bazelFlags *commandline.BazelFlags) []*parser.BazelOption {
...
+			case <-time.After(1 * time.Second):
+			}
+		}
+	}()
+	select {
+	case <-connected:
+		return nil
+	case <-time.After(5 * time.Second):
+		return fmt.Errorf("could not connect to sidecar")
+	}
```
tylerwilliams, line 128: <i was gonna suggest using WithTimeout but then i realized that api is bad.>

tylerwilliams (changes requested):

tylerwilliams (approved):

PR #1927, opened 2022-04-23
Use SWC to transpile TypeScript

bduffany:
* Clean builds of the app (and incremental builds of targets with lots of transitive deps, such as format.tsx or router.tsx) are ~20X faster. For example, an incremental build of the app after modifying format.tsx decreased from 39.7 seconds to 1.7 seconds.
* Type checking is removed from development builds. Instead, type checking is moved to separate `{name}_typecheck_test` targets for each TS library target.
* The TS language server will still do type checking, so most type checking during development still works. (Type checking for protos still requires the usual workarounds, such as running [sgp](https://github.com/tempoz/sgp) to symlink generated protos)

---

**Version bump**: TODO


bduffany:
@siggisim this is ready for review now -- since we are now migrated to ts_project I think the diff is small enough that we could roll this back easily if needed.

bduffany:
@siggisim this is ready for review now -- since we are now migrated to ts_project I think the diff is small enough that we could roll this back easily if needed.

tsconfig.json
```diff
@@ -13,7 +13,16 @@
...
-    "rootDirs": [".", "bazel-bin"]
+    "rootDirs": [
+      ".",
+      "bazel-bin",
+      "bazel-out/k8-opt/bin",
+      "bazel-out/k8-fastbuild/bin",
+      "bazel-out/darwin-opt/bin",
+      "bazel-out/darwin-fastbuild/bin",
+      "bazel-out/darwin_arm64-opt/bin",
+      "bazel-out/darwin_arm64-fastbuild/bin"
```
siggisim, line 24: I thought I remembered seeing something here that we were doing some other workaround thing so that we didn't have to list all of these in our tsconfig: https://github.com/buildbuddy-io/buildbuddy/pull/1956/files#diff-2085a400b69ac0ec4291b70e84b04a90a144e10b8cea4b4f082d70dce50ce015R11

rules/typescript/index.bzl
```diff
@@ -1,26 +1,76 @@
...
+        transpiler = _swc,
+        srcs = srcs,
+        **kwargs
+    )
+
+def ts_jasmine_node_test(name, srcs, deps = [], size = "small", **kwargs):
+    if len(srcs) != 1:
+        fail("srcs must contain exactly one TS source file")
+
+    # jasmine_node_test can't run TS tests directly, so compile the TS to JS
```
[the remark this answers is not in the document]
bduffany: I'd be fine to put this on hold and wait till 1.0 -- frontend development is getting pretty slow though, so in the meantime I might spend some time on splitting up some targets so that they don't invalidate so much of the build graph when changed (mainly the invocation target).

tsconfig.json
bduffany: There was some issue I ran into when omitting these, and I meant to dig into it and see what the root issue was. I'll dig more into it next time I revisit this PR

bduffany: I removed these and it builds just fine -- maybe I needed these during some intermediate state as I was migrating things, but everything builds fine without these now.

bduffany: jk, the typecheck tests fail without these. I would like to dig into it at some point and see why these are needed, but would like to decouple that from this PR

PR #1928, opened 2022-04-25
Add basic support for macOS sandboxing.

tylerwilliams:
This CL adds macOS sandboxing support for remotely executed actions by generating a sandbox-config file per-action and then executing the action using sandbox-exec.

A generated sandbox config looks like this:

```
(version 1)
(debug deny)
(allow default)
(allow process-exec (with no-sandbox) (literal "/bin/ps"))
(deny file-read*
    (subpath "/tmp/remote_build")
)
(allow file-read*
    (subpath "/tmp/remote_build/9d054f1d-26bd-4ba0-a203-e2925869b279")
)
(deny file-write*)
(allow file-write*
    (subpath "/Users/tylerw/Library/Caches")
    (subpath "/Users/tylerw/Library/Developer")
    (subpath "/Users/tylerw/Library/Logs")
    (subpath "/dev")
    (subpath "/private/tmp")
    (subpath "/private/var/folders/2f/wy_q59s572b5fsv3tryx38d00000gn/C")
    (subpath "/private/var/folders/2f/wy_q59s572b5fsv3tryx38d00000gn/T")
    (subpath "/private/var/tmp")
    (subpath "/tmp/remote_build/9d054f1d-26bd-4ba0-a203-e2925869b279")
)
```

The above profile:
 - allows reads from everywhere, except anything in the build root that is not the currently executing action
 - disallows writes everywhere except the remote action directory, tmp dirs, and some system locations needed by Xcode.

I've only tested this so far on a very simple project: //examples/macos/HelloWorld:HelloWorld from rules_apple.

### DIFF
--- a/enterprise/server/remote_execution/containers/sandbox/sandbox.go
+++ b/enterprise/server/remote_execution/containers/sandbox/sandbox.go
@@ -148,9 +148,9 @@ func resolveAlwaysWritableDirs(ctx context.Context) ([]sbxPath, error) {
 	dirs = append(dirs, darwinUserCache)
 
 	// Add some user specific dirs as well.
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		return nil, err
+	homeDir := "/"
+	if h, err := os.UserHomeDir(); err == nil {
+		homeDir = h
 	}
 	dirs = append(dirs, filepath.Join(homeDir, "Library/Caches"))
 	dirs = append(dirs, filepath.Join(homeDir, "Library/Logs"))
@@ -272,7 +272,7 @@ func (c *sandbox) runCmdInSandbox(ctx context.Context, command *repb.Command, wo
 	// slash, so make sure it does not!
 	inaccessiblePaths := []sbxPath{NewSubPath(filepath.Dir(strings.TrimRight(workDir, "/")))}
 	writablePaths := append(alwaysWritableDirs, NewSubPath(workDir))
-	
+
 	sandboxConfigPath := filepath.Join(workDir, "sandbox.sb")
 	sandboxConfigBuf := makeSandboxConfig(writablePaths, inaccessiblePaths, c.enableNetwork)
 	if err := os.WriteFile(sandboxConfigPath, sandboxConfigBuf, 0660); err != nil {
--- a/enterprise/server/remote_execution/containers/sandbox/sandbox.go
+++ b/enterprise/server/remote_execution/containers/sandbox/sandbox.go
@@ -49,6 +49,9 @@ const (
 	subPath
 )
 
+// sbxPath represents a path and the associated sandbox path type. Calling
+// String() on a sbxPath will return a tinyScheme sandbox statement that can
+// be used in a sandbox config.
 type sbxPath struct {
 	expression string
 	pathType   sbxPathType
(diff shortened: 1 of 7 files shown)


enterprise/server/remote_execution/containers/sandbox/sandbox.go
```diff
@@ -0,0 +1,311 @@
...
+	// Three different modes are supported:
+	// 1) regular expression
+	// 2) literal
+	// 3) subpath
+	regexPath sbxPathType = iota
+	literalPath
+	subPath
+)
+
+type sbxPath struct {
```
bduffany, line 52:
```suggestion
// sbxPath represents a set of paths to be allowed or blocked for access by the sandbox.
type sbxPath struct {
```

enterprise/server/remote_execution/containers/sandbox/sandbox.go
```diff
@@ -0,0 +1,311 @@
...
+	result := runSimpleCommand(ctx, command)
+	if result.Error != nil {
+		return "", result.Error
+	}
+	return strings.TrimSpace(string(result.Stdout)), nil
+}
+
+func resolveSymlink(path string, fileInfo os.FileInfo) string {
+	if fileInfo.Mode()&os.ModeSymlink != 0 {
+		if absPath, err := os.Readlink(path); err == nil {
```
bduffany, line 111: nit: would it make sense to return the error here if there is one?

enterprise/server/remote_execution/containers/sandbox/sandbox.go
```diff
@@ -0,0 +1,311 @@
...
+	homeDir := "/"
+	if h, err := os.UserHomeDir(); err == nil {
+		homeDir = h
+	}
+	dirs = append(dirs, filepath.Join(homeDir, "Library/Caches"))
+	dirs = append(dirs, filepath.Join(homeDir, "Library/Logs"))
+	dirs = append(dirs, filepath.Join(homeDir, "Library/Developer"))
+
+	for _, path := range dirs {
+		if fileInfo, err := os.Lstat(path); !os.IsNotExist(err) {
```
bduffany, line 161: nit: I don't understand the logic for checking `!os.IsNotExist(err)` here, maybe add a comment?

tylerwilliams: This is a somewhat confusing API, but this is just checking the file exists:

https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go
(you need to read the whole thing because it contains some wrong answers lol)

enterprise/server/remote_execution/containers/sandbox/sandbox.go
```diff
@@ -0,0 +1,330 @@
...
+func (s *sbxPath) typeString() string {
+	switch s.pathType {
+	case regexPath:
+		return "regex"
+	case literalPath:
+		return "literal"
+	case subPath:
+		return "subpath"
+	default:
+		return "unknwon-sbx-path-type"
```
siggisim, line 70: ubernit: `unknwon`=>`unknown`

siggisim (approved):

enterprise/server/remote_execution/containers/sandbox/sandbox.go
bduffany: It seems like `os.IsNotExist(someRandomIOError)` would return `false`, therefore `!os.IsNotExist(err)` would return true, and you would unintentionally execute the `if {}` body

tylerwilliams: I had to write a test to see it but of course you're right :) https://go.dev/play/p/bDmgptfJCly

Sent https://github.com/buildbuddy-io/buildbuddy/pull/1932 to fix.

PR #1929, opened 2022-04-25
Upgrade rules_nodejs to 5.4.0

bduffany:
* Bump rules_nodejs to v5.4.0
* Set params on `yarn_install` to override the new defaults in rules_nodejs 5.x, and instead stick with our old settings for now (to ease the migration)
* Since ts_library has been moved to the concatjs package in rules_nodejs 5.x, and we eventually want to avoid depending on ts_library, move ts_library to its own macro in `//rules/typescript:index.bzl` rather than referencing concatjs directly, to avoid further churn.

---

**Version bump**: Patch


siggisim (approved):

bduffany:
Running into various issues with this. Putting this on hold for now since I need to work on getting firecracker enabled for workflows, but will revisit soon.

PR #1936, opened 2022-04-26
Add functionality to flagutil.go in preparation for switching flag-based configuration

tempoz:
Adds the functionality to `flagutil.go` necessary to swap the config machinery fully
over to being flag-based.

---

**Version bump**: Minor

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -1392,6 +1392,12 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         sum = "h1:5SgMzNM5HxrEjV0ww2lTmX6E2Izsfxas4+YHWRs3Lsk=",
         version = "v1.8.0",
     )
+    go_repository(
+        name = "com_github_go_test_deep",
+        importpath = "github.com/go-test/deep",
+        sum = "h1:TDsG77qcSprGbC6vTN8OuXp5g+J+b5Pcguhf7Zt61VM=",
+        version = "v1.0.8",
+    )
 
     go_repository(
         name = "com_github_go_toolsmith_astcast",
--- a/deps.bzl
+++ b/deps.bzl
@@ -1392,12 +1392,6 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         sum = "h1:5SgMzNM5HxrEjV0ww2lTmX6E2Izsfxas4+YHWRs3Lsk=",
         version = "v1.8.0",
     )
-    go_repository(
-        name = "com_github_go_test_deep",
-        importpath = "github.com/go-test/deep",
-        sum = "h1:TDsG77qcSprGbC6vTN8OuXp5g+J+b5Pcguhf7Zt61VM=",
-        version = "v1.0.8",
-    )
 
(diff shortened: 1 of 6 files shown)


tylerwilliams (changes requested):
Code looks reasonable, but could you add tests for this?

tempoz:
Tests added, bugs fixed.

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/go-redis/redis/v8 v8.10.0
 	github.com/go-redsync/redsync/v4 v4.4.1
 	github.com/go-sql-driver/mysql v1.6.0
+	github.com/go-test/deep v1.0.8
```
bduffany, line 39: would go-cmp work instead? (since we already depend on it and use it in a few places)

server/util/flagutil/flagutil.go
```diff
@@ -240,3 +268,126 @@ func (f *URLFlag) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	_, err = url.Parse(string(*f))
 	return err
 }
+
+func GenerateYAMLMapFromFlags() (map[interface{}]interface{}, error) {
```
tylerwilliams, line 272: can you comment all the newly exported methods in this file?

tempoz:
it ended up being the case that the unmarshal implementation for yaml with a map was blithely overwriting my values at a given index with no regard for what was there before; added a new function to marshal things to the correct types.

server/util/flagutil/flagutil.go
```diff
@@ -240,3 +269,176 @@ func (f *URLFlag) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	_, err = url.Parse(string(*f))
 	return err
 }
+
+// Generates a map of the type that should be marshaled from YAML for each flag
+// name at the corresponding nested map index.
+func GenerateYAMLTypeMapFromFlags() (map[string]interface{}, error) {
```
tylerwilliams, line 275: In go, the official way to write comments is to start them with the function name. So this would be:

// GenerateYAMLTypeMapFromFlags generates a map of the type that should be
// marshaled from YAML for each flag name at the corresponding nested map index.


(can you double check the rest of the comments in here too?)

tylerwilliams (approved):
LGTM (1 nit)

deps.bzl
```diff
@@ -1392,6 +1392,12 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         sum = "h1:5SgMzNM5HxrEjV0ww2lTmX6E2Izsfxas4+YHWRs3Lsk=",
         version = "v1.8.0",
     )
+    go_repository(
```
bduffany, line 1395: nit: rm

bduffany (approved):

server/util/flagutil/flagutil.go
tempoz: Thanks for the info! Done.

deps.bzl
tempoz: Done.

PR #1937, opened 2022-04-26
Change tests to use new flag-based configuration

tempoz:
Update the tests to use the new flag-based configuration machinery.

---

**Version bump**: Minor


enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
```diff
@@ -80,6 +79,9 @@ import (
 	scpb "github.com/buildbuddy-io/buildbuddy/proto/scheduler"
 	guuid "github.com/google/uuid"
 	bspb "google.golang.org/genproto/googleapis/bytestream"
+
+	// included to define flags
+	_ "github.com/buildbuddy-io/buildbuddy/enterprise/server/cmd/executor"
```
bduffany, line 84: Hm, wondering why this is this needed? Seems like if the flag is not transitively imported already, then setting the flag one way or the other would have no effect

server/config/config.go
```diff
@@ -551,6 +551,9 @@ func (cc *RedisClientConfig) String() string {
 // value will be consistent.
 func (c *Configurator) ReconcileFlagsAndConfig() {
 	c.GenerateFlagSet().VisitAll(func(flg *flag.Flag) {
+		if flag.Lookup(flg.Name) == nil {
```
bduffany, line 554: What's the purpose of this check? (maybe add a comment)

server/util/testing/flags/flags.go
```diff
@@ -2,51 +2,42 @@ package flags
...
-// CheckFlagsAgainstConfig checks that all defined flags containing a `.` are
-// present in the config struct.
-func CheckFlagsAgainstConfig(t *testing.T) {
-	configurator, err := config.NewConfiguratorFromData([]byte{})
+// Set a flag value and register a cleanup function to restore the flag
+// to its original value after the given test is complete.
+func Set(t testing.TB, name string, value interface{}) {
+	origValue, err := flagutil.DereferencedValueFromFlagName(name)
+	if status.IsNotFoundError(err) {
+		log.Warningf("Attempted to set %s to %v, but %s was not found.", name, value, name)
```
bduffany, line 32: If we make this a test failure instead of just logging, I think it would prevent these warnings from building up over time, and also prevent people from setting flags that don't exist and then a test passes even though it shouldn't. wdyt?

tempoz: I made this an error before and it made testenv sad, but I now think there's a better way to fix this: manually guard in testenv against setting a nonexistent flag. Not sure why I didn't just do that in the first place. Fixing this also revealed many places where we were changing options that are never referenced.

bduffany (approved):

tempoz:
* master
  * **PR #1937** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1937" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1937?utm_source=stack-comment).

PR #1938, opened 2022-04-26
Switch over fully to flag-based configuration

tempoz:
Removes the configurator entirely, as it is no longer a relevant structure. `flags.Parse`
can now be called safely directly from a main function without "registering" flags first.
We now only need to populate the flags from a given YAML config, as opposed to reconciling,
and the new flagutils way of populating the flags means that we don't need to externally
keep track of which flags were set initially via the command line.

---

**Version bump**: Minor

### DIFF
--- a/cli/cmd/sidecar/sidecar.go
+++ b/cli/cmd/sidecar/sidecar.go
@@ -115,9 +115,9 @@ func propagateAPIKeyStreamInterceptor() grpc.StreamServerInterceptor {
 	return rpcfilters.ContextReplacingStreamServerInterceptor(propagateAPIKeyFromIncomingToOutgoing)
 }
 
-func initializeEnv(configurator *config.Configurator) *real_environment.RealEnv {
+func initializeEnv() *real_environment.RealEnv {
 	healthChecker := healthcheck.NewHealthChecker(*serverType)
-	env := real_environment.NewRealEnv(configurator, healthChecker)
+	env := real_environment.NewRealEnv(healthChecker)
 	env.SetAuthenticator(&nullauth.NullAuthenticator{})
 	env.SetBuildEventHandler(&devnull.BuildEventHandler{})
 	return env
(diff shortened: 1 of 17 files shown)


cli/cmd/sidecar/sidecar.go
```diff
@@ -207,16 +207,14 @@ func initializeDiskCache(env *real_environment.RealEnv) {
 func main() {
 	// Can remove all this configurator stuff once all flags the sidecar uses are
 	// defined outside the configurator.
-	config.RegisterAndParseFlags()
-	configurator, err := config.NewConfiguratorFromData([]byte{})
-	if err != nil {
+	flag.Parse()
+	if err := config.PopulateFlagsFromData([]byte{}); err != nil {
```
vadimberezniker, line 211: Do we still need this? Does calling PopulateFlagsFromData with an empty byte slice have any side-effects?

cli/cmd/sidecar/sidecar.go
```diff
@@ -207,16 +207,14 @@ func initializeDiskCache(env *real_environment.RealEnv) {
 func main() {
 	// Can remove all this configurator stuff once all flags the sidecar uses are
```
vadimberezniker, line 208: delete comment

vadimberezniker (approved):
this is super exciting!!!

cli/cmd/sidecar/sidecar.go
tempoz: nah, good point. this is totally unnecessary.

tempoz:
* master
  * **PR #1937** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1937" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/1938?utm_source=stack-comment).

PR #1951, opened 2022-04-28
Scorecard: Implement server-side sorting, searching, filtering, and grouping

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/scorecard/scorecard.go
+++ b/server/remote_cache/scorecard/scorecard.go
@@ -117,12 +117,12 @@ func filterResults(results []*capb.ScoreCard_Result, req *capb.GetCacheScoreCard
 				return nil, status.InvalidArgumentErrorf("invalid response type %d", req.GetFilter().GetResponseType())
 			}
 		case "search":
+			s := strings.ToLower(req.GetFilter().GetSearch())
 			predicates = append(predicates, func(result *capb.ScoreCard_Result) bool {
-				s := req.GetFilter().GetSearch()
-				return strings.Contains(result.GetActionId(), s) ||
-					strings.Contains(result.GetActionMnemonic(), s) ||
-					strings.Contains(result.GetTargetId(), s) ||
-					strings.Contains(result.GetDigest().GetHash(), s)
+				return strings.Contains(strings.ToLower(result.GetActionId()), s) ||
+					strings.Contains(strings.ToLower(result.GetActionMnemonic()), s) ||
+					strings.Contains(strings.ToLower(result.GetTargetId()), s) ||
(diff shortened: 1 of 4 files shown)


server/remote_cache/scorecard/scorecard.go
```diff
@@ -62,46 +68,156 @@ func GetCacheScoreCard(ctx context.Context, env environment.Env, req *capb.GetCa
...
+					strings.Contains(result.GetActionMnemonic(), s) ||
+					strings.Contains(result.GetTargetId(), s) ||
+					strings.Contains(result.GetDigest().GetHash(), s)
+			})
+		default:
+			return nil, status.InvalidArgumentErrorf("invalid field path %q", path)
+		}
+	}
+
+	out := make([]*capb.ScoreCard_Result, 0, len(results))
```
vadimberezniker, line 132: nit: i think it would look slightly cleaner if you defined a separate func to check the predicates. something like
matches := func(r *capb.ScoreCard) bool {...}

then the loop below would become
```
for _, result := range results {
  if(matches(result)) {
    out = append(out, result)
  }
}
```

server/remote_cache/scorecard/scorecard_test.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	res, err := scorecard.GetCacheScoreCard(ctx, env, req)
+	require.NoError(t, err)
+
+	assertResults(t, res, besUpload, casUpload, acMiss, casDownload)
+}
+
+func assertResults(t *testing.T, res *capb.GetCacheScoreCardResponse, msg ...*capb.ScoreCard_Result) {
+	// Note: not asserting directly on the protos because the diff is too hard to read.
```
vadimberezniker, line 188: take a look at protocmp and consider whether it's useful here: https://github.com/buildbuddy-io/buildbuddy/blob/4d9c9965dcfe967822a1269648f95c976c263fda/enterprise/server/workflow/service/service_test.go#L289

vadimberezniker (approved):

server/remote_cache/scorecard/scorecard_test.go
bduffany: Thanks for pointing me to that -- I didn't know about go-cmp before, and it looks super useful.

I took a look at the diff that it outputs, and it has the same problem as the testify diff -- it does field-level diffs which are a bit too fine-grained for this particular test where the protos just being out of order causes a lot of diffs. Ideally, it would say something like "these lists have the same protos, but in the wrong order," but alas.

server/remote_cache/scorecard/scorecard.go
bduffany: Much cleaner -- done

PR #1953, opened 2022-04-29
Scorecard: Implement UI for sorting, searching, filtering, and grouping

bduffany:
https://user-images.githubusercontent.com/2414826/165875813-c8de456f-f3be-4fe4-b9f2-639459e03e02.mp4

* Implement sorting, searching, filtering, grouping
* Fix bug that timeline bars were rendering with inconsistent positioning. It seems either `flex-grow` does not give us enough precision, causing the position to be a bit erratic -- or there's something about `flex-grow` I don't understand. Either way, switched to `margin-left` to set the timeline bar position, which is much more sane.
* Add a new FilterInput component since we have 3 other places where we duplicate the same pattern. Polish it a bit by making it clickable anywhere and having a separate UI state when you're actively filtering. Planning on migrating the existing filter components in a follow-up PR.

---

**Version bump**: Patch


app/invocation/cache_requests_card.tsx
```diff
@@ -113,29 +151,162 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
...
+          <Select value={this.getResponseType()} onChange={this.onChangeResponseType.bind(this)}>
+            <Option value={cache.ResponseType.UNKNOWN_RESPONSE_TYPE}>Any status</Option>
+            <Option value={cache.ResponseType.OK}>OK only</Option>
+            <Option value={cache.ResponseType.NOT_FOUND}>NotFound only</Option>
+            <Option value={cache.ResponseType.ERROR}>Errors only</Option>
+          </Select>
+          {/* Grouping controls */}
+          <div className="separator" />
+          <CheckboxButton checked={this.getGroupByAction()} onClick={this.onToggleGroupByAction.bind(this)}>
+            Group by action
```
siggisim, line 254: Should this be "Group by target"?

siggisim (approved):
This is really great, just a couple of comments about the language we use in the UI

app/invocation/cache_requests_card.tsx
```diff
@@ -113,29 +151,162 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
...
+            {this.getDescending() ? <SortDesc className="icon" /> : <SortAsc className="icon" />}
+          </OutlinedButton>
+          {/* Filtering controls */}
+          <div className="separator" />
+          <Select value={this.getCacheType()} onChange={this.onChangeCacheType.bind(this)}>
+            <Option value={cache.CacheType.UNKNOWN_CACHE_TYPE}>CAS and AC</Option>
+            <Option value={cache.CacheType.CAS}>CAS only</Option>
+            <Option value={cache.CacheType.AC}>AC only</Option>
+          </Select>
+          <Select value={this.getRequestType()} onChange={this.onChangeRequestType.bind(this)}>
```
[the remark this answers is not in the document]
bduffany: People can see the same data that we show in the current UI by selecting cache="AC only", request="Read", response="NotFound". One thing we could do to make it easier is to make those the default filters.

That said, even if those are made the default settings, there is some room for improvement in how we present these filtering options to the user -- for example, to see all CAS transfers starting from those default filters, you'd have to change AC => CAS, Read => Read+Write, NotFound => All (a bit cumbersome IMO).

Maybe we could condense these 3 filters into a single "View" filter, with the options:
- All: cache=any, req=any, res=any
- Misses (default option): cache=AC, req=read, res=NotFound
- Hits: cache=AC, req=read, res=OK
- Downloads: cache=CAS, req=read, res=OK
- Uploads: cache=CAS, req=write, res=OK
- Errors: cache=any, req=any, res=(!NotFound && !OK)

But then I wonder whether people want to be able to see the AC requests and CAS requests side-by-side in the same view... the "All" option lets you do that, but then you don't get the option of filtering by Uploads vs. Downloads, or see only the CAS transfers involved in AC misses, etc.

Maybe a compromise between the 1-dropdown design (above) and 3-dropdown design (current), is to have two separate dropdowns, one for AC and one for CAS? Like:

AC (maybe labeled "Actions"?):
- All, Misses (default), Hits, Errors
CAS (maybe labeled "Artifacts"?):
- All, Hidden (default), Uploads, Downloads, Errors

Haven't thought this through a ton though -- would be interested to hear more about what you think

app/invocation/cache_requests_card.tsx
siggisim: Ah okay, this makes sense - I thought we were just grouping by the target name that we got in the metadata header.

app/invocation/cache_requests_card.tsx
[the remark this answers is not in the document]
siggisim: I kinda like the single dropdown because that's how I think about it in my head (I feel like I understand this stuff pretty deeply, but I wasn't sure which options to select to see cache misses).

If people ask for more than those 6 views, we could always add new ones (or add advanced filtering options that gives them all 3).

app/invocation/cache_requests_card.tsx
siggisim: Yeah that seems reasonable

bduffany: I'll do this in another PR since it'll need some backend changes too.

app/invocation/cache_requests_card.tsx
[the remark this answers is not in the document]
bduffany: Combined into a single dropdown

bduffany:
Will go ahead and merge since I'd like to enable this in dev so we can start testing it out -- feel free to leave more comments and I can address in a follow-up

PR #1954, opened 2022-04-29
Increase max recv msg size for vmexec.Exec

bduffany:
Bumping to 50MB since we are seeing tasks returning 20MB of stdout/stderr.

---

**Version bump**: Patch


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -149,6 +149,10 @@ const (
 	// This deadline isn't used for the Exec() codepath because removal is handled
 	// in the background by the runner pool.
 	finalizationTimeout = 10 * time.Second
+
+	// Max recv message size for Exec() requests, which can have lots of stderr/
+	// stdout in the response.
+	grpcMaxRecvMsgSize = 50_000_000
```
tylerwilliams, line 155: could you add the units in the const name?

vadimberezniker (approved):

PR #1963, opened 2022-05-03
Record transferred size in scorecard results

bduffany:
This will let us show compression ratio in the scorecard.

We also want to show overall bytes saved from compression - that can be in a separate PR, but this PR does most of the work needed for that.

---

**Version bump**: Patch


bduffany:
@tylerwilliams fyi, made the refactor from `SetBytesTransferred/Close` -> `CloseWithBytesTransferred`

vadimberezniker (approved):

PR #1964, opened 2022-05-03
Fixup option name broken on refactor

layus:
s/--docker_devices/--executor.docker_devices/

The old code had all the options unprefixed, this one instance must have been missed.

Also, added some hints about the right format. Took me an insane amount of time to figure out, so I hope it will help others.

---

**Version bump**: Patch

-->

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -79,7 +79,8 @@ var (
 )
 
 func init() {
-	flagutil.StructSliceVar(&dockerDevices, "docker_devices", "Sets --device= on the docker command.")
+	flagutil.StructSliceVar(&dockerDevices, "executor.docker_devices", `Configure (docker) devices that will be available inside the sandbox container. Format is --executor.docker_devices='[{"PathOnHost":"/dev/foo","PathInContaine
+r":"/some/dest","CgroupPermissions":"see,docker,docs"}]'`)
 }
 
 const (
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):
@layus Thank you so much for catching this and sending a PR. Much appreciated!

bduffany:
The test failure is because we're undergoing a migration from config structs to flags, and the test detected that the config struct usage string here needs an update: https://github.com/buildbuddy-io/buildbuddy/blob/9dced96dd860fbde4668a1990847f7c6caef9a21/server/config/config.go#L314

Changing the usage string there should fix the broken test :+1:

layus:
Having fun with github live edits. But if there is still an issue I will have to redo it locally :grimacing: . Also, please squash before merging.

layus:
That new error is... interesting :-). I just tested locally and the patch passed the tests. I wonder what happened ?

layus:
Having bazel crash on these issues is quite unexpected. Did you ever report bazel crashing on remote executor errors like this one https://app.buildbuddy.io/invocation/a0a86180-62c4-4d0b-a503-4314578f6884 ?

siggisim:
That failure is due to us having a different (not great) workflow for untrusted PRs:
```
Invalid Argument: error creating runner for command: runner recycling is not supported for anonymous builds (recycling was requested via platform property "recycle-runner=true")
```

@bduffany or @tylerwilliams should be able to run the tests and merge.

bduffany (approved):

bduffany:
Thanks again for the contribution!

PR #1973, opened 2022-05-04
Inital implementation of volumes

layus:
Initial implementation of volumes/mounts/bind mounts support for docker sandboxes and podman, albeit I did not test podman at all, and docker only works in my single tests case. 

---

**Version bump**: Minor


enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -56,6 +56,7 @@ type DockerOptions struct {
 	DockerNetwork           string
 	DockerCapAdd            string
 	DockerDevices           []config.DockerDeviceMapping
+	Binds                   []string
```
bduffany, line 59: It looks like [`Binds`](https://cs.github.com/moby/moby/blob/61404de7df1a6c75c2cbdc2c3ce226c95bebcaad/api/types/container/host_config.go#L411) specifies only volumes, while [`Mounts`](https://cs.github.com/moby/moby/blob/61404de7df1a6c75c2cbdc2c3ce226c95bebcaad/api/types/container/host_config.go#L411) can specify other types of mounts as well, including volumes.

What do you think about changing this to `Mounts` and then passing it through as `HostConfig.Mounts`?

layus:
Replying from my phone right now, but it seems to me that binds allows to mount volumes (a relative source part) and bind mounts (an absolute host path). Mounts also allows to mount tmpfs and pipes, but that seems less generic. I love the fact that this syntax can be mapped to several backends as-is. Using Mounts requires either advanced parsing or structured attrs. Not sure it is worth the loss of simplicity for now.

Le 4 mai 2022 16:41:53 GMT+02:00, Brandon Duffany ***@***.***> a écrit :
***@***.*** commented on this pull request.

bduffany:
@layus after a bit more thought, I think I see what you mean regarding the simplicity. We would need to parse the mount flag and pass it as a struct. So I think it's fine to use volumes. Thanks for bringing up that concern :) I'll review the PR as-is now.

server/config/config.go
```diff
@@ -312,6 +312,7 @@ type ExecutorConfig struct {
 	DockerNetHost                 bool                      `yaml:"docker_net_host" usage:"Sets --net=host on the docker command. Intended for local development only."`
 	DockerCapAdd                  string                    `yaml:"docker_cap_add" usage:"Sets --cap-add= on the docker command. Comma separated."`
 	DockerDevices                 []DockerDeviceMapping     `yaml:"docker_devices" usage:"Configure (docker) devices that will be available inside the sandbox container. Format is --executor.docker_devices='[{\"PathOnHost\":\"/dev/foo\",\"PathInContaine\nr\":\"/some/dest\",\"CgroupPermissions\":\"see,docker,docs\"}]'"`
+	Mounts                        []string                  `yaml:"mounts" usage:"Sets --mount= (podman), --volume= (docker) or similar one-line mountpoint config int the action sandbox."`
```
bduffany, line 315:
```suggestion
	DockerVolumes                 []string                  `yaml:"docker_volumes" usage:"Additional --volume arguments to be passed to docker or podman."`
```

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -126,6 +127,9 @@ func (c *podmanCommandContainer) getPodmanRunArgs(workDir string) []string {
 		}
 		args = append(args, "--device="+strings.Join(deviceSpecs, ":"))
 	}
+	for _, mount := range c.options.Mounts {
+		args = append(args, "--mount="+mount)
+	}
```
bduffany, line 132: Podman supports volumes too, so I think we'd want to use the same flag here to be consistent with Docker.

```suggestion
	for _, volume := range c.options.Volumes {
		args = append(args, "--volume="+volume)
	}
```

layus:
@bduffany Alternatively, The parsing is readily available upstream: https://github.com/docker/cli/blob/21c5391cee93f90ca17fb841b60a8aa72373723c/opts/mount.go
Would it be easy to integrate here ?

bduffany:
> @bduffany Alternatively, The parsing is readily available upstream: https://github.com/docker/cli/blob/21c5391cee93f90ca17fb841b60a8aa72373723c/opts/mount.go Would it be easy to integrate here ?

We don't currently depend on that docker/cli package, so it would probably be easier to stick with just volumes for now

bduffany (approved):

PR #1976, opened 2022-05-04
Speed up image building in release.py

bduffany:
Group all the `update_latest_tag` builds together, since those all have the same `--define` flags. (Changing `--define` flags invalidates the Bazel analysis cache, causing the builds to be much slower)

---

**Version bump**: Patch

### DIFF
--- a/release.py
+++ b/release.py
@@ -94,28 +94,20 @@ def update_docker_image(new_version, update_latest_tag):
     clean_cmd = 'bazel clean --expunge'
     run_or_die(clean_cmd)
 
-    # build the open source app
-    version_build_cmd = 'bazel run -c opt --stamp --define version=%s --define release=true deployment:release_onprem' % new_version
-    run_or_die(version_build_cmd)
+    build_oss_app_cmd = 'bazel run -c opt --stamp --define version=%s --define release=true deployment:release_onprem' % new_version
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -97,21 +97,23 @@ def update_docker_image(new_version, update_latest_tag):
...
     version_build_cmd = 'bazel run -c opt --stamp --define version=enterprise-%s --define release=true enterprise/deployment:release_executor_enterprise' % new_version
     run_or_die(version_build_cmd)
+
+    # update "latest" tags
     if update_latest_tag:
+        latest_build_cmd = 'bazel run -c opt --stamp --define version=latest --define release=true deployment:release_onprem'
+        run_or_die(latest_build_cmd)
+
+        latest_build_cmd = 'bazel run -c opt --stamp --define version=latest --define release=true enterprise/deployment:release_enterprise'
+        run_or_die(latest_build_cmd)
```
tylerwilliams, line 115: nit: call these vars 2 diff things to avoid silly typo errors later

tylerwilliams (approved):

release.py
bduffany: fixed here and above

tylerwilliams:
Thanks for improving this!

PR #1984, opened 2022-05-05
Store transferred size in invocation table instead of only compressed sizes

bduffany:
Rename "compressed size" to "transferred size" and include IDENTITY requests in the total transferred size (e.g. AC requests) so that the total bytes transferred shown in the UI is more accurate. For example, if we have 1 AC request uncompressed @ 100KB and 1 CAS request @ 10KB compressed, 30KB uncompressed, we want to show it in the UI as 130KB downloaded (110KB w/ compression). But before this PR we'd show 130KB downloaded (10KB w/ compression) which is inaccurate and makes the compression ratio seem higher than it is.

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_cache_card.tsx
+++ b/app/invocation/invocation_cache_card.tsx
@@ -83,13 +83,13 @@ export default class CacheCardComponent extends React.Component<Props> {
                             "downloaded",
                             "download-color-swatch",
                             Number(cacheStat.totalDownloadSizeBytes),
-                            Number(cacheStat.totalCompressedDownloadSizeBytes)
+                            Number(cacheStat.totalDownloadTransferredSizeBytes)
                           )}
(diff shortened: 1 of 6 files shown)


tylerwilliams (approved):
These are pretty similarly named so might want to add a comment so callers use the right one.

	TotalDownloadSizeBytes            int64
	TotalUploadSizeBytes              int64
	TotalDownloadTransferredSizeBytes int64
	TotalUploadTransferredSizeBytes   int64

bduffany:
> These are pretty similarly named so might want to add a comment so callers use the right one.
> 
> ```
> TotalDownloadSizeBytes            int64
> TotalUploadSizeBytes              int64
> TotalDownloadTransferredSizeBytes int64
> TotalUploadTransferredSizeBytes   int64
> ```

Done

PR #1985, opened 2022-05-06
Show more data in the scorecard

bduffany:
* Show compressed size when not filtering to AC (AC requests results are always uncompressed so it's useless when filtering to AC)
* Hide waterfall chart when screen is too small
* Allocate more of the table width to the "name" column (target and/or action mnemonic)

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/cache_requests_card.tsx
+++ b/app/invocation/cache_requests_card.tsx
@@ -1,7 +1,17 @@
 import React from "react";
 import router from "../router/router";
 import InvocationModel from "./invocation_model";
-import { X, ArrowUp, ArrowDown, ArrowLeftRight, ChevronRight, Check, SortAsc, SortDesc } from "lucide-react";
+import {
+  X,
+  ArrowUp,
+  ArrowDown,
+  ArrowLeftRight,
+  ChevronRight,
+  Check,
+  SortAsc,
+  SortDesc,
+  ArrowRight,
+} from "lucide-react";
 import { cache } from "../../proto/cache_ts_proto";
 import rpc_service from "../service/rpc_service";
 import DigestComponent from "../components/digest/digest";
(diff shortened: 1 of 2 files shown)


app/invocation/cache_requests_card.tsx
```diff
@@ -477,6 +502,16 @@ function cacheTypeTitle(cacheType: cache.CacheType): string | undefined {
   }
 }
 
+function renderCompressionSavings(result: cache.ScoreCard.IResult) {
+  const compressionSavings = 1 - Number(result.transferredSizeBytes) / Number(result.digest.sizeBytes);
+  return (
+    <span className={`compression-savings ${compressionSavings > 0 ? "positive" : "negative"}`}>
+      {compressionSavings <= 0 ? "+" : ""}
+      {(-compressionSavings * 100).toPrecision(3)}%
```
siggisim, line 510: Thoughts on just showing percentage here? (You could hover if you wanted to see the new size).

app/invocation/cache_requests_card.tsx
```diff
@@ -304,6 +314,20 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
         <div>
           <DigestComponent hashWidth="96px" sizeWidth="72px" digest={result.digest} expandOnHover={false} />
         </div>
+        {this.isCompressedSizeColumnVisible() && (
```
siggisim, line 317: Are there any column headers? Might be nice as there are a lot more numbers now.

siggisim (approved):

app/invocation/cache_requests_card.tsx
siggisim: Also trying to think through the order of the columns.

Seems like in order of importance it'd be something like:
Hit/Miss, Name, Digest/Size, Time, Compression savings, Waterfall

bduffany: Reordered the columns :+1: (tweaked the order slightly)

app/invocation/cache_requests_card.tsx
bduffany: I was on the fence, but it's probably cleaner to show just the percentage. Done :+1:

siggisim (approved):
Looks great!

PR #1987, opened 2022-05-06
Delete unused flag

bduffany:
There are no refs to this flag, so it's safe to delete.

---

**Version bump**: TODO


tempoz (approved):
LGTM; thanks for finding this!!!

PR #1993, opened 2022-05-09
Add full request metadata to executor cache requests

bduffany:
Forward the complete request metadata proto with requests from executor -> cache, not just the invocation ID metadata.

This fixes an issue that the scorecard is missing the target ID, action ID, and action mnemonic for RBE requests.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
+++ b/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
@@ -262,8 +262,6 @@ func (q *PriorityTaskScheduler) EnqueueTaskReservation(ctx context.Context, req 
 func (q *PriorityTaskScheduler) propagateExecutionTaskValuesToContext(ctx context.Context, execTask *repb.ExecutionTask) context.Context {
 	ctx = context.WithValue(ctx, "x-buildbuddy-jwt", execTask.GetJwt())
 	rmd := execTask.GetRequestMetadata()
-	// TODO(bduffany): remove this fallback once all apps populate request
-	// metadata in each ExecutionTask.
 	if rmd == nil {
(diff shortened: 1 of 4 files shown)


enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -258,13 +259,16 @@ func (q *PriorityTaskScheduler) EnqueueTaskReservation(ctx context.Context, req
 	return &scpb.EnqueueTaskReservationResponse{}, nil
 }
 
-func propagateExecutionTaskValuesToContext(ctx context.Context, execTask *repb.ExecutionTask) context.Context {
+func (q *PriorityTaskScheduler) propagateExecutionTaskValuesToContext(ctx context.Context, execTask *repb.ExecutionTask) context.Context {
 	ctx = context.WithValue(ctx, "x-buildbuddy-jwt", execTask.GetJwt())
-	rmd := &repb.RequestMetadata{
-		ToolInvocationId: execTask.GetInvocationId(),
+	rmd := execTask.GetRequestMetadata()
+	// TODO(bduffany): remove this fallback once all apps populate request
```
vadimberezniker, line 265: I think this can still be nil for the remote bazel synthetic Execute path. 
we recently updated the workflow code to populate the request metadata key, but the remote bazel path has not been updated yet.

vadimberezniker (approved):

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
bduffany: Thanks -- with that context, I think adding a hard dependency on RequestMetadata seems a bit much. Removed the TODO for now, so we can gracefully handle unexpected cases like this.

PR #1994, opened 2022-05-09
Don't store executor requests in cache scorecard

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/proto/remote_execution.proto
+++ b/proto/remote_execution.proto
@@ -1727,6 +1727,12 @@ message RequestMetadata {
   // or equality across invocations, though some client tools may offer these
   // guarantees.
   string configuration_id = 7;
+
+  // BUILDBUDDY EXTENSIONS BELOW
+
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):

proto/remote_execution.proto
```diff
@@ -1750,3 +1756,9 @@ message ExecutionTask {
   Platform platform_overrides = 8;
   RequestMetadata request_metadata = 9;
 }
+
+message ExecutorDetails {
+  // ID of the host this executor is running on.
+  // Ex. "8BiY6U0F"
```
vadimberezniker, line 1762: nit: this is actually a GUID

vadimberezniker (approved):

proto/remote_execution.proto
bduffany: Removed this line

PR #1995, opened 2022-05-09
Add client column to scorecard

bduffany:
Add a column indicating whether a request came from Bazel or an executor.

image

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/cache_requests_card.tsx
+++ b/app/invocation/cache_requests_card.tsx
@@ -342,6 +342,9 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
             )}
           </div>
         )}
+        {this.isClientColumnVisible() && (
+          <div className="client-column">{result.executorHostId ? "Executor" : "Bazel"}</div>
+        )}
         <div className="cache-type-column" title={cacheTypeTitle(result.cacheType)}>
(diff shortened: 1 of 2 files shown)


siggisim:
Do we want to show executor requests at all here? Seems like we'd just want to show Bazel requests.

bduffany:
> Do we want to show executor requests at all here? Seems like we'd just want to show Bazel requests.

I could see the AC misses getting a bit noisy. But it might be interesting to see slow CAS transfers in this graph -- it'd be a quick way to see "oh, this build was slow because the executor was uploading a huge artifact."

For now it's probably simplest to hide all executor requests and then maybe later add a filtering option to see executor CAS transfers. wdyt?

siggisim:
> For now it's probably simplest to hide all executor requests and then maybe later add a filtering option to see executor CAS transfers. wdyt?

Seems reasonable, but the executors just make *so* many CAS requests that I'm not sure how valuable it would be to see them in a view like this (can send you an invocation with over 1 million executor CAS requests).

My fear would be collecting these would slow down serving the useful requests (from Bazel) and potentially cause us to hit ~500k scorecard request cutoff - which would leave us with partial data for both the useful requests and the executor requests.

PR #2003, opened 2022-05-11
Re-enable code splitting

bduffany:
```
BEFORE
size    11323 KB
gzipped 2174 KB

AFTER
size    4093 KB
gzipped 742 KB

DIFF
size    -7230 KB
gzipped -1431 KB

DIFF%
size    -63.85%
gzipped -65.84%
```

---

**Version bump**: Patch

### DIFF
--- a/app/BUILD.bazel
+++ b/app/BUILD.bazel
@@ -26,6 +26,7 @@ esbuild(
         ":fastbuild": False,
         "//conditions:default": True,
     }),
+    splitting = True,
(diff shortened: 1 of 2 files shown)


siggisim (approved):
LGTM, but we wait to merge til after the release so we can test in dev for a week?

bduffany:
> LGTM, but we wait to merge til after the release so we can test in dev for a week?

yep -- sg

PR #2004, opened 2022-05-11
Better / safer edit instance template script.

tylerwilliams:
This was basically a clean rewrite of this script attempting to make it a bit safer and more intuitive to use. It's probably easier to review as that rather than in diff mode.

It is safe to run multiple times -- it will do nothing once modifications have been made.
It supports a dry run mode so you can see what changes will be made.
It uses a smarter template naming approach to better support changes in the future.
It modifies all instance groups in the region, not just one at a time.

TODO: I might still want to delete the old instance templates, I only delete the old groups now. WDYT?

### DIFF
--- a/tools/edit_instance_template/edit_instance_template.go
+++ b/tools/edit_instance_template/edit_instance_template.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"path"
+	"strconv"
 	"strings"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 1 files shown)


tools/edit_instance_template/edit_instance_template.go
```diff
@@ -16,152 +17,276 @@ import (
...
 
 var (
-	project    = flag.String("project", "", "GCP project that hosts the template")
-	template   = flag.String("template", "", "The name of the base template")
-	suffix     = flag.String("suffix", "ssd-network", "Suffix to append to instance template name")
-	size       = flag.Int("size", 5, "The initial number of instances in the group")
-	region     = flag.String("region", "us-west1", "The region of the subnetwork")
+	project       = flag.String("project", "", "GCP project that hosts the template")
+	region        = flag.String("region", "us-west1", "The region of the subnetwork")
+	templateRegex = flag.String("template_regex", "", "The regex to use when matching instance templates")
```
vadimberezniker, line 22: It would be good to have a safer default for the template regex. I don't like the idea of modifying *all* MIGs globally by default, even with the dry-run option. 
What do you think of making the following changes:
1) add --gke_cluster flag and ignore any templates/MIGs that do not start with gke-{cluster}
2) change default for template_regex to .*executor.*
3) only modify instance templates used by MIGs in --region
the last one is probably the most annoying because you can't tell in which region a template is being used w/o looking at MIGs first. 

This might be a little paranoid, but I've read too many postmortems about people blowing things up with scripts like this :)

vadimberezniker (changes requested):

tools/edit_instance_template/edit_instance_template.go
```diff
@@ -16,152 +17,276 @@ import (
 )
 
 var (
-	project    = flag.String("project", "", "GCP project that hosts the template")
-	template   = flag.String("template", "", "The name of the base template")
-	suffix     = flag.String("suffix", "ssd-network", "Suffix to append to instance template name")
-	size       = flag.Int("size", 5, "The initial number of instances in the group")
-	region     = flag.String("region", "us-west1", "The region of the subnetwork")
+	project       = flag.String("project", "", "GCP project that hosts the template")
+	region        = flag.String("region", "us-west1", "The region of the subnetwork")
```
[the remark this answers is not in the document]
tylerwilliams: Done

tools/edit_instance_template/edit_instance_template.go
tylerwilliams: I added a cluster flag and a pool flag, and I filter igms to only those in the specified region.

tools/edit_instance_template/edit_instance_template.go
```diff
@@ -16,152 +17,285 @@ import (
...
-	if n, err := getTemplateURL(ctx, c, newTemplateName); err != nil {
-		if !strings.Contains(err.Error(), "Error 404") {
-			log.Fatalf("Could not check for existence of new template: %s", err)
+	migc, err := compute.NewInstanceGroupManagersRESTClient(ctx)
+	if err != nil {
+		log.Fatalf("Could not create MIG client: %s", err)
+	}
+	defer migc.Close()
+
+	// Find all instance templates that match our regex. Typically
```
vadimberezniker, line 218: outdated comment as there's no more regex

tools/edit_instance_template/edit_instance_template.go
```diff
@@ -16,152 +17,285 @@ import (
...
+	return op.Wait(ctx)
 }
 
 func main() {
 	flag.Parse()
 
+	if !*apply {
+		defer log.Printf("Run this script again with --apply=true to execute these changes!")
+	}
+
```
vadimberezniker, line 204: can you bail the script early if `*region == ""` or `*cluster==""`?

vadimberezniker (approved):

tools/edit_instance_template/edit_instance_template.go
tylerwilliams: Done.

tools/edit_instance_template/edit_instance_template.go
tylerwilliams: Done.

PR #2008, opened 2022-05-12
Delete and create instance group managers with the same name

tylerwilliams:


tools/edit_instance_template/edit_instance_template.go
```diff
@@ -278,26 +278,15 @@ func main() {
...
-					if err := createIGMWithTemplate(ctx, migc, name, shortZone, modifiedTemplateURL); err != nil {
-						log.Fatalf("Error creating instance group manager: %s", err)
-					}
-				}
 				if err := deleteIGM(ctx, migc, igm); err != nil {
 					log.Fatalf("Error deleting instance group manager: %s", err)
 				}
-
+				shortZone := path.Base(igm.GetZone())
+				if err := createIGMWithTemplate(ctx, migc, igm.GetName(), shortZone, modifiedTemplateURL); err != nil {
```
tylerwilliams, line 287: here we use the old igm name when creating the new igm

vadimberezniker (approved):

PR #2011, opened 2022-05-12
Add config option to specify additional environment variables that should be passed to remote actions

siggisim:
---

**Version bump**: None


enterprise/server/remote_execution/platform/platform.go
```diff
@@ -26,6 +26,7 @@ var (
 	enableFirecracker    = flag.Bool("executor.enable_firecracker", false, "Enables running execution commands inside of firecracker VMs")
 	defaultImage         = flag.String("executor.default_image", "gcr.io/flame-public/executor-docker-default:enterprise-v1.6.0", "The default docker image to use to warm up executors or if no platform property is set. Ex: gcr.io/flame-public/executor-docker-default:enterprise-v1.5.4")
 	enableVFS            = flag.Bool("executor.enable_vfs", false, "Whether FUSE based filesystem is enabled.")
+	extraEnvVars         = flag.String("executor.extra_env_vars", "", "Additional environment variables to pass to remotely executed actions, comma separate. i.e. MY_ENV_VAR=foo,ANOTHER_ONE=bar")
```
tylerwilliams, line 29: Could we use a string slice flag instead to allow for env var values that contain commas?

siggisim: Done!

tylerwilliams (approved):

PR #2012, opened 2022-05-12
Refactor flagutil to use generics

tempoz:
- Removed reflection-based `structSliceFlag`
- Removed `stringSliceFlag`
- Unified above two use cases with `SliceFlag[T]`
- Removed the `SliceFlag` interface and associated methods
- Added interfaces and methods to help reolve type of parameterized flags (and custom flags in general)
- Removed `URLFlag` from `flagTypeMap` in favor of using above-mentioned interfaces/methods
- Refactored flagutil to use `any` instead of `interface{}`
- `DereferencedValueByFlagName` now accepts a type to attempt to convert the underlying value to, and returns a value of that type
- Updated tests to reflect changes
- Removed references to `flag.Lookup` outside of flagutil in favor of `DereferencedValueByFlagName`
- Updated slice-based flags in code to use new interface, preferring `flagutil.Slice` over `flagutil.SliceVar`
- Diabled `nilness` analyzer for `nogo` as methods of templated types anywhere in the code caused it to panic, even when ignoring the files containing the templates
- Copied `oauthProviders` instead of assigning to avoid possibly changing the value when appending to the former assignee

---

**Version bump**: Patch


tempoz:
* master
  * **PR #2012** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2012" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2012?utm_source=stack-comment).

enterprise/server/auth/auth.go
```diff
@@ -517,15 +512,16 @@ func newForTesting(ctx context.Context, env environment.Env, testAuthenticator a
...
+			*enableAnonymousUsage || (len(*oauthProviders) == 0 && !selfauth.Enabled()),
 			*adminGroupID,
 		),
 	)
 	return nil
 }
 
 func Register(ctx context.Context, env environment.Env) error {
-	authConfigs := oauthProviders
+	authConfigs := make([]OauthProvider, len(*oauthProviders))
```
tempoz, line 523:
```suggestion
	// if we append to authConfigs, don't want to incidentally change oauthProviders
	authConfigs := make([]OauthProvider, len(*oauthProviders))
```

enterprise/server/remote_execution/redis_client/redis_client.go
```diff
@@ -24,7 +24,11 @@ func RegisterRemoteExecutionClient(env environment.Env) error {
 	}
 
 	// Fulfill internal remote execution requests locally.
-	conn, err := grpc_client.DialTarget(fmt.Sprintf("grpc://localhost:%d", reflect.ValueOf(flag.Lookup("grpc_port").Value).Convert(reflect.TypeOf((*int)(nil))).Elem().Int()))
+	grpc_port, err := flagutil.DereferencedValueFromFlagName[int]("grpc_port")
+	if err != nil {
+		return status.InternalErrorf("Error initializing remote execution client: %s", err)
+	}
+	conn, err := grpc_client.DialTarget(fmt.Sprintf("grpc://localhost:%d", grpc_port))
```
tempoz, line 31:
```suggestion
	grpcPort, err := flagutil.DereferencedValueFromFlagName[int]("grpc_port")
	if err != nil {
		return status.InternalErrorf("Error initializing remote execution client: %s", err)
	}
	conn, err := grpc_client.DialTarget(fmt.Sprintf("grpc://localhost:%d", grpcPort))
```

tempoz, line 31:
```suggestion
	grpcPort, err := flagutil.DereferencedValueFromFlagName[int]("grpc_port")
	if err != nil {
		return status.InternalErrorf("Error initializing remote execution client: %s", err)
	}
	conn, err := grpc_client.DialTarget(fmt.Sprintf("grpc://localhost:%d", grpcPort))
```

tylerwilliams (approved):

BUILD
```diff
@@ -28,7 +28,7 @@ nogo(
         "@org_golang_x_tools//go/analysis/passes/loopclosure:go_default_library",
         "@org_golang_x_tools//go/analysis/passes/lostcancel:go_default_library",
         # "@org_golang_x_tools//go/analysis/passes/nilfunc:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/nilness:go_default_library",
+        # "@org_golang_x_tools//go/analysis/passes/nilness:go_default_library", # template methods currently cause this analyzer to panic
```
vadimberezniker, line 31: have you tried updating the org_golang_x_tools version? it looks like they addressed some generics related issues in newer versions

enterprise/server/cmd/server/main.go
```diff
@@ -60,23 +60,19 @@ func configureFilesystemsOrDie(realEnv *real_environment.RealEnv) {
...
-	if staticDirectoryFlag := flag.Lookup("static_directory"); staticDirectoryFlag != nil {
-		if staticDirectory := staticDirectoryFlag.Value.String(); staticDirectory != "" {
-			staticFS, err := static.FSFromRelPath(staticDirectory)
-			if err != nil {
-				log.Fatalf("Error getting static FS from relPath: %q: %s", staticDirectory, err)
-			}
-			realEnv.SetStaticFilesystem(staticFS)
+	if staticDirectory, err := flagutil.DereferencedValueFromFlagName[string]("static_directory"); err == nil && staticDirectory != "" {
+		staticFS, err := static.FSFromRelPath(staticDirectory)
+		if err != nil {
```
vadimberezniker, line 65: The name is a mouthful, could be something short like flagutil.GetValue()?

server/backends/blobstore/blobstore.go
```diff
@@ -39,10 +39,12 @@ import (
 
 var (
 	// Disk flags
-	rootDirectory     = flag.String("storage.disk.root_directory", "/tmp/buildbuddy", "The root directory to store all blobs in, if using disk based storage.")
-	partitions        []disk.Partition
-	partitionMappings []disk.PartitionMapping
-	useV2Layout       = flag.Bool("storage.disk.use_v2_layout", false, "If enabled, files will be stored using the v2 layout. See disk_cache.MigrateToV2Layout for a description.")
+	rootDirectory = flag.String("storage.disk.root_directory", "/tmp/buildbuddy", "The root directory to store all blobs in, if using disk based storage.")
+	useV2Layout   = flag.Bool("storage.disk.use_v2_layout", false, "If enabled, files will be stored using the v2 layout. See disk_cache.MigrateToV2Layout for a description.")
```
vadimberezniker, line 43: feel free to delete these 3 flags (use_v2_layout, partitions, partition_mappings), they are only applicable to cache storage

server/util/flagutil/flagutil.go
```diff
@@ -275,16 +182,42 @@ func (f *URLFlag) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return err
 }
 
+func (f *URLFlag) AliasedType() reflect.Type {
```
vadimberezniker, line 185: do we need both AliasedType and YAMLTypeAlias()? the url flag is defined as a string so i think if you were to return *URLFlag in AliasedType it would still work. in what cases would they be different?

vadimberezniker (changes requested):

server/util/flagutil/flagutil.go
tempoz: We do; I use it in https://github.com/buildbuddy-io/buildbuddy/pull/2014, for example. The `AliasedType` is the "native" type of the flag in go (the type returned to the user when creating the flag), while the `YAMLTypeAlias` is the type used by YAML, which will importantly have the `MarshalYAML` and `UnmarshalYAML` methods.

server/backends/blobstore/blobstore.go
tempoz: Done!

enterprise/server/cmd/server/main.go
tempoz: I think it's important to clarify that it's the dereferenced `Value`, since `flag.Value` is a pointer. How's `GetDereferencedValue` sound?

BUILD
tempoz: Good idea, I hadn't tried that. Unfortunately, now I have, and the newer version still panics :/

enterprise/server/cmd/server/main.go
vadimberezniker: If the caller wants the flag value, in what cases do they care whether it's a pointer or not? 
It's not a big deal, so if you want to call it `GetDereferencedValue` then it's fine.

vadimberezniker (approved):

enterprise/server/cmd/server/main.go
tempoz: It's mostly only important because `SetValueForFlagName` takes a non-pointer value, and if you don't specify the type (i.e., `T` is `any`), you can't then pass that value to `SetValueForFlagName`, which is the use case we have in `server/util/testing/flags/flags.go`'s `Set` function.

PR #2015, opened 2022-05-13
Ameliorate slow GetTree calls by adding some simple caching.

tylerwilliams:
Allow for faster GetTree calls when there are large or deep directories of dependencies that do not change often.

Example (partial tree hit):

GetTree fetched 4094 dirs from cache across 2047 calls in cumulative 1.050845167s (total time: 15.200223ms)
GetTree fetched 2049 dirs from cache across 3 calls in cumulative 21.27µs (total time: 1.869137ms)

<>Recursive code always makes my head explode<>

I refactored the 'fetch' function so that it creates its own errgroup and returns all descendent directories under the specified dir instead of just appending everything to the output stream as it comes. This allows the fetch call to cache a partial tree result after it has fetched everything underneath it, and then it can check for a cache hit before doing work.

To keep things simple, store the cached partial trees in the AC namespace. Only cache partial trees that have > 10 children which are not the root node (we can play with this later).


siggisim (approved):
❤️

vadimberezniker (approved):
awesome

PR #2014, opened 2022-05-12
Change URLFlag so that its underlying type is URL, not string

tempoz:
The motivation for this is mainly to stop having to re-parse a URL over and over again, as well
as to remove the associated error handling code, which was needlessly complicating.

---

**Version bump**: Patch

PR #2017, opened 2022-05-13
Add ability to create flags that are aliases for existing flags

tempoz:
This allows us to, without breaking backward-compatibilty, put flags that are currently
just top-level values (no dots) into sections which make sense for them. e. g., `port`
could now be `app.port`, perhaps, while still allowing the original name to be used
completely normally. Syntax looks something like:

```
var appPort = flag.Int("app.port", 8080, "Port for buildbuddy server to use.")
```

and then sometime later, perhaps in a different file:

```
var port = flagutil.Alias[int]("port", "app.port")
```

The pointers `port` and `app.port` are completely equivalent and thus point to
the same data.

---

**Version bump**: Minor


tempoz:
* master
  * **PR #2012** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2012" target="_blank">Graphite</a> 
    * **PR #2014** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2014" target="_blank">Graphite</a> 
    * **PR #2017** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2017" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2017?utm_source=stack-comment).

server/util/flagutil/flagutil.go
```diff
@@ -56,6 +56,10 @@ type TypeAliased interface {
 	AliasedType() reflect.Type
 }
 
+type IsNameAliasing interface {
```
tylerwilliams, line 59: does this need to be public?

server/util/flagutil/flagutil.go
```diff
@@ -190,6 +194,63 @@ func (f *URLFlag) YAMLTypeAlias() reflect.Type {
 	return reflect.TypeOf((*URLFlag)(nil))
 }
 
+type FlagAlias struct {
+	name string
+}
+
+func Alias[T any](newName, name string) *T {
```
tylerwilliams, line 201: can you double check what the "override" behavior is for normal flags before submitting this?

tylerwilliams (approved):

server/util/flagutil/flagutil.go
tempoz: Flags are last one wins, like we thought, which makes sense with how slice flags work.

server/util/flagutil/flagutil.go
tempoz: No, good point.

PR #2019, opened 2022-05-16
Periodically refresh executor pool metadata in redis

bduffany:
Periodically update `executorPools/...` in Redis so that the executors page does not suddenly stop working due to this key being evicted.

Tested locally using 5s refresh interval, since I made the memory limit super low to make triggering eviction easier. In prod, 5 min should be short enough of a refresh interval since we have much more memory so the max key age should be much longer than 5 min.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1367


tylerwilliams (approved):
LGTM but please get Vadim's OK before submitting.

bduffany:
Closing since a lot of other things break when we reach maxmemory, and this doesn't buy us much.

PR #2023, opened 2022-05-16
Add flag to disable local peer shortcut for distributed writes.

vadimberezniker:
Enabled by default.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/distributed/distributed.go
+++ b/enterprise/server/backends/distributed/distributed.go
@@ -29,12 +29,13 @@ import (
 )
 
 var (
-	listenAddr        = flag.String("cache.distributed_cache.listen_addr", "", "The address to listen for local BuildBuddy distributed cache traffic on.")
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/distributed/distributed.go
```diff
@@ -29,12 +29,13 @@ import (
...
+	groupName          = flag.String("cache.distributed_cache.group_name", "", "A unique name for this distributed cache group. ** Enterprise only **")
+	nodes              = flagutil.StringSlice("cache.distributed_cache.nodes", []string{}, "The hardcoded list of peer distributed cache nodes. If this is set, redis_target will be ignored. ** Enterprise only **")
+	replicationFactor  = flag.Int("cache.distributed_cache.replication_factor", 0, "How many total servers the data should be replicated to. Must be >= 1. ** Enterprise only **")
+	clusterSize        = flag.Int("cache.distributed_cache.cluster_size", 0, "The total number of nodes in this cluster. Required for health checking. ** Enterprise only **")
+	disableLocalWrites = flag.Bool("cache.distributed_cache.disable_local_writes", true, "If enabled, does not shortcut distributed writes that belong to the local shard.")
```
tylerwilliams, line 38: nit: would be clearer as a + boolean?

enableLocalWrites = flag.Bool("cache.distributed_cache.enable_local_writes", false, "If enabled,  shortcut distributed writes that belong to the local shard instead of making an RPC.")

tylerwilliams (approved):

PR #2026, opened 2022-05-17
Setup secondary network when starting executor

luluz66:
When --route_prefix is not "default", i.e. we are using a secondary network, we need to set up policy based routing. 

When executor starts, we need to set up the secondary network: 
1. add a routing table. `echo "1 rt1" | sudo tee -a /etc/iproute2/rt_tables`
2. add ip rules. 
```
ip rule add from 172.24.0.24 table rt1
ip rule add to 172.24.0.24  table rt1
```
3. add ip routes
```
ip route add 172.24.0.1 src 172.24.0.24 dev ens6 table rt1
ip route add default via 172.24.0.1 dev ens6 table rt1
```

After the above steps, we need to add rules/routes for podman:
1. register the podman default bridge network by running `podman run --rm busybox sh`
2. add ip rules
```
ip rule add from 10.88.0.0/16 lookup rt1
ip rule add  to 10.88.0.0/16 lookup rt1
```
3. add ip route
```
ip route add 10.88.0.0/16 via 10.88.0.1 dev cni-podman0
```

Test:
Push the executor to executor-lulu-pool; and 
1. verified the custom rules and routes on the executor and on the vm
2. Run `tcpdump -n -i any "host 8.8.8.8"` and in another terminal run `podman run --rm busybox ping -c3 8.8.8.8". Confirmed that the ping succeeded and it went through ens6 (the secondary network interface)
3. Confirmed that `podman run --rm busybox ping -c3 <internal_ip_address on default network>` failed.


tylerwilliams (approved):
There's a lot here, nice job digging in and figuring this out :)

PR #2031, opened 2022-05-17
Add more invocation suggestions (round 1)

bduffany:
Behind a flag `--app.expanded_suggestions_enabled`

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_suggestion_card.tsx
+++ b/app/invocation/invocation_suggestion_card.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { AlertCircle, HelpCircle } from "lucide-react";
 import { TextLink } from "../components/link/link";
 import InvocationModel from "./invocation_model";
+import capabilities from "../capabilities/capabilities";
 
 interface Props {
   suggestions: Suggestion[];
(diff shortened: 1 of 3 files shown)


app/invocation/invocation_suggestion_card.tsx
```diff
@@ -70,6 +71,108 @@ const matchers: SuggestionMatcher[] = [
...
+        flags.push(flag);
+      }
+    }
+    if (!flags.length) return null;
+    return {
+      level: SuggestionLevel.INFO,
+      message: (
+        <>
+          This build is using the old BuildBuddy endpoint, <span className="inline-code">cloud.buildbuddy.io</span>.
+          Migrate to <span className="inline-code">remote.buildbuddy.io</span> for improved performance.
```
bduffany, line 93: @siggisim I wasn't sure if we wanted to add more details here about why it's faster, wdyt?

app/invocation/invocation_suggestion_card.tsx
```diff
@@ -70,6 +71,108 @@ const matchers: SuggestionMatcher[] = [
...
+    if (!capabilities.config.expandedSuggestionsEnabled) return null;
+
+    if (!model.optionsMap.get("remote_cache") && !model.optionsMap.get("remote_executor")) return null;
+    if (model.optionsMap.get("remote_download_outputs")) return null;
+
+    return {
+      level: SuggestionLevel.INFO,
+      message: (
+        <>
+          Adding the Bazel flag <span className="inline-code">--remote_download_minimal</span> can significantly improve
```
siggisim, line 164: you might want to suggest (and look for) either --remote_download_minimal or --remote_download_toplevel  (the latter is needed instead in some scenarios)

bduffany, line 164: @siggisim do you know of any legitimate cases where `--remote_download_minimal` can cause problems? wasn't sure whether we want to recommend this across the board or not.

siggisim (approved):
lgtm, just one nit

app/invocation/invocation_suggestion_card.tsx
siggisim: I think this is fine!

app/invocation/invocation_suggestion_card.tsx
siggisim: yeah - i think in some cases you have to use toplevel instead

bduffany: Done

bduffany: kk, I think we should be covered since we're mentioning both flags now :+1:

app/invocation/invocation_suggestion_card.tsx
```diff
@@ -70,6 +71,109 @@ const matchers: SuggestionMatcher[] = [
...
+      message: (
+        <>
+          Adding the Bazel flag <span className="inline-code">--remote_download_minimal</span> or{" "}
+          <span className="inline-code">--remote_download_toplevel</span> can significantly improve performance for
+          builds with a large number of intermediate results.
+        </>
+      ),
+      reason: (
+        <>
+          Shown because this build is cache-enabled and the effective flag{" "}
```
brentleyjones, line 171: Each of these starts with "Shown because". Seems a bit redundant.

bduffany: do you mean redundant from a code perspective ("DRY") or redundant from the user's perspective?

If the latter, I think the redundancy serves to reinforce that the bottom "footnote" is always just explaining the technical reason that the suggestion is displayed, and that it can be ignored when skimming over these suggestions unless you are confused why this suggestion is being displayed:

image

brentleyjones: From a user's perspective. Seeing it in your screenshot though, I think "Reason: " would read better than "Shown because".

PR #2032, opened 2022-05-18
Add more invocation suggestions (round 2)

bduffany:
- ~~Suggest using `--incompatible_strict_action_env` if not set~~ -- removed this one, since Brentley pointed out it's not desirable for iOS builds, and there may be other cases I'm not aware of.
- Suggest adding `COMMIT_SHA` and `REPO_URL` metadata if this is a CI test and those pieces of metadata are missing
- Polish up a few other suggestions (add flag links, make wording more consistent)

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1357

### DIFF
--- a/app/invocation/invocation_suggestion_card.tsx
+++ b/app/invocation/invocation_suggestion_card.tsx
@@ -174,28 +174,6 @@ const matchers: SuggestionMatcher[] = [
       ),
     };
   },
-  // Suggest using --incompatible_strict_action_env
-  ({ model }) => {
-    if (!capabilities.config.expandedSuggestionsEnabled) return null;
-
-    if (model.optionsMap.get("incompatible_strict_action_env") !== undefined) return null;
-
-    return {
-      level: SuggestionLevel.INFO,
-      message: (
-        <>
-          Consider adding the Bazel flag <BazelFlag>--incompatible_strict_action_env</BazelFlag> to help avoid cache
-          misses due to slight differences in environment variables across builds.
-        </>
-      ),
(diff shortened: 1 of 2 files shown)


app/invocation/invocation_suggestion_card.tsx
```diff
@@ -161,15 +161,68 @@ const matchers: SuggestionMatcher[] = [
...
+          <span className="inline-code">--remote_download_outputs</span> is not explicitly set.
+        </>
+      ),
+    };
+  },
+  // Suggest using --incompatible_strict_action_env
+  ({ model }) => {
+    if (!capabilities.config.expandedSuggestionsEnabled) return null;
+
+    if (model.optionsMap.get("incompatible_strict_action_env") !== undefined) return null;
```
bduffany, line 181: @siggisim I wasn't sure whether we want to show this across the board, or only if remote cache is enabled. wdyt?

brentleyjones: I don't think we should suggest it until we can mark suggestions as "I don't want to do this". Most iOS projects shouldn't actually set this, and should instead filter `PATH` in `tools/bazel`, since `--incompatible_strict_action_env` adds `/usr/local/bin`, which picks up Homebrew on Intel Macs.

bduffany: Thanks for pointing this out -- removed from this PR.

siggisim (approved):

PR #2037, opened 2022-05-18
Extend context before copying container logs

siggisim:
I see quite a few of these errors in server logs:
```
Command execution returned error: rpc error: code = DeadlineExceeded desc = failed to read docker container logs: context deadline exceeded
```

This seems to indicate we're relying on test timeout context to uploads logs (which can be large), but we should probably give it some additional time.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -158,6 +158,8 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		}()
 	}()
 
+	ctx, cancel := background.ExtendContextForFinalization(ctx, containerFinalizationTimeout)
+	defer cancel()
 	r.copyContainerLogs(ctx, cid, result)
 	if result.Error != nil {
 		return result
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -158,6 +159,8 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		}()
 	}()
 
+	ctx, cancel := background.ExtendContextForFinalization(ctx, containerFinalizationTimeout)
```
bduffany, line 162: I think this will effectively allow the task forever, since `copyContainerLogs` blocks until the task is complete.

Maybe the logged errors are WAI? (or we could silence them)

bduffany: more context: `ContainerStart` (above) starts the task, but doesn't wait for it to finish

bduffany: err... it doesn't allow tasks to run forever, but rather restricts all tasks to the 10s timeout

bduffany: err... actually ExtendContextForFinalization will only add the 10s if there is less than 10s remaining -- so I think in most cases, extending the context here might not change the timeout

siggisim: Ah, so it would extend all execution deadlines by 10s?

I'm seeing some really large test logs (1.5GB+) and seeing some funky behavior. Thought this might come into play since I was seeing these in the logs but I might be wrong.

siggisim: Just saw your follow up - that makes sense. Back to the drawing board lol.

PR #2038, opened 2022-05-18
Update bazelrc with some bazel 5.0 cache flags

bduffany:
This will give us some experience using these flags, since we are recommending them with the suggestions cards.

- Enable `--experimental_remote_cache_compression` when running with `--config=remote` or `--config=cache` (does not apply to workflows since it's likely to be detrimental, and not to github CI either since it's unclear whether the network is a bottleneck there)
- Enable `--experimental_remote_cache_async` and `--incompatible_remote_build_event_upload_respect_no_cache` unconditionally

---

**Version bump**: Patch

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,12 +13,14 @@ build:cache-dev --bes_results_url=https://app.buildbuddy.dev/invocation/
 build:cache-dev --bes_backend=grpcs://remote.buildbuddy.dev
 build:cache-dev --remote_cache=grpcs://remote.buildbuddy.dev
 build:cache-dev --remote_upload_local_results
+build:cache-dev --experimental_remote_cache_compression
 
 # Build with --config=cache to send build logs to the production server with cache
 build:cache --bes_results_url=https://app.buildbuddy.io/invocation/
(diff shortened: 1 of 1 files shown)


siggisim (approved):

vadimberezniker:
It looks like `--incompatible_remote_build_event_upload_respect_no_cache` also prevents logs from being uploaded when `--noremote_upload_local_results` is set. Is that intentional? I use `--noremote_upload_local_results` in my user bazelrc.

brentleyjones:
I believe so: https://github.com/bazelbuild/bazel/pull/14338#issuecomment-983657277

> test logs are treated as action outputs. i.e. they are uploaded if the result of test actions can be uploaded. Or not uploaded if tagged with "no-cache".

bduffany:
Seems a bit unfortunate that the only way to upload test logs is to disable `--noremote_upload_local_results`, which I guess may upload a bunch of other unwanted stuff?

brentleyjones:
Correct. It's probably worth a Bazel GitHub issue.

PR #2041, opened 2022-05-19
Show file names for BES uploads in cache scorecard

bduffany:
Instead of showing "bes-upload" for all of these names, show the actual file name by cross-referencing File metadata that we have readily available in the BEP.

Computing these on the server side rather than client side, and storing them in the scorecard proto so that they are searchable.

The server side implementation extracts file metadata from the buffered BES events that we already have in memory during FinalizeInvocation. The downside is that this metadata sits in memory for 500ms for each invocation while we wait for stats buffers across apps to be flushed. This will slightly increase memory usage, but to a very small degree. Also, the alternatives seemed a bit overkill (either storing this metadata in Redis or re-reading the full invocation in the stats recorder just to get file name metadata).

image

---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/scorecard/scorecard.go
+++ b/server/remote_cache/scorecard/scorecard.go
@@ -131,7 +131,7 @@ func filterResults(results []*capb.ScoreCard_Result, req *capb.GetCacheScoreCard
 		case "search":
 			s := strings.ToLower(req.GetFilter().GetSearch())
 			predicates = append(predicates, func(result *capb.ScoreCard_Result) bool {
-				filePath := strings.Join([]string{result.GetPathPrefix(), result.GetName()}, "/")
+				filePath := filepath.Join(result.GetPathPrefix(), result.GetName())
 				return strings.Contains(strings.ToLower(result.GetActionId()), s) ||
 					strings.Contains(strings.ToLower(result.GetActionMnemonic()), s) ||
 					strings.Contains(strings.ToLower(result.GetTargetId()), s) ||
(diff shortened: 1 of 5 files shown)


server/remote_cache/scorecard/scorecard.go
```diff
@@ -124,10 +131,12 @@ func filterResults(results []*capb.ScoreCard_Result, req *capb.GetCacheScoreCard
 		case "search":
 			s := strings.ToLower(req.GetFilter().GetSearch())
 			predicates = append(predicates, func(result *capb.ScoreCard_Result) bool {
+				filePath := strings.Join([]string{result.GetPathPrefix(), result.GetName()}, "/")
```
tylerwilliams, line 134: nit: use filepath.Join for paths?

server/remote_cache/scorecard/scorecard.go
```diff
@@ -257,6 +266,53 @@ func blobName(invocationID string) string {
 	return filepath.Join(invocationID, blobFileName)
 }
 
+// ExtractFiles extracts files from invocation BES events, returning a mapping
+// of digest to file.
```
tylerwilliams, line 270: nit: could you clarify which files are extracted?

server/remote_cache/scorecard/scorecard.go
```diff
@@ -257,6 +266,53 @@ func blobName(invocationID string) string {
...
+	for _, result := range sc.Results {
+		if result.ActionId != "bes-upload" {
+			continue
+		}
+		f := files[result.Digest.GetHash()]
+		if f == nil {
+			continue
+		}
+		result.Name = f.Name
+		result.PathPrefix = strings.Join(f.PathPrefix, "/")
```
tylerwilliams, line 312: filepath.Join

tylerwilliams (changes requested):
few nits, this LG otherwise. And neat that we can figure this stuff out :)

server/remote_cache/scorecard/scorecard.go
bduffany: Done

server/remote_cache/scorecard/scorecard.go
bduffany: Done

server/remote_cache/scorecard/scorecard.go
bduffany: Done

PR #2043, opened 2022-05-19
set up policy based routing for firecracker

luluz66:


server/util/networking/networking.go
```diff
@@ -154,6 +154,15 @@ func DeleteRoute(ctx context.Context, vmIdx int) error {
 	return runCommand(ctx, "ip", "route", "delete", cloneIP)
 }
 
+func DeleteRuleIfSecondaryNetworkEnabled(ctx context.Context, vmIdx int) error {
+	if !IsSecondaryNetworkEnabled() {
+		// IP rule is only added when the secondary network is enabled
+		return nil
+	}
+	cloneIP := fmt.Sprintf("192.168.%d.%d", vmIdx/30, ((vmIdx%30)*8)+3)
```
tylerwilliams, line 162: can you make a "getCloneIP(vmIdx int)" func since we have this in a couple places?

tylerwilliams (approved):
1 small nit, otherwise lgtm

server/util/networking/networking.go
luluz66: Done

PR #2045, opened 2022-05-19
Add suggestion preference to Groups table

bduffany:
It always starts as ENABLED when creating a group, and can be disabled in the settings UI only when editing a group.

---

**Version bump**: Patch


tylerwilliams (approved):
We should add a GroupPreferences table soon...

PR #2051, opened 2022-05-20
Route ddisk read requests to zone peers when possible.

tylerwilliams:
Pretty much does what the title says.

Each app uses a library to figure out its own GKE zone, but it can also be overridden with a flag.

The ddisk node acting as coordinator (the one that is serving the external-facing RPC) computes the list of nodes that have the file and prefers to read from itself or a peer in its zone, if one exists.

This is to help improve latency and cut down on inter-zone egress costs.


server/resources/resources.go
```diff
@@ -152,3 +158,10 @@ func GetMyPort() (int32, error) {
 	}
 	return int32(i), nil
 }
+
+func GetZone() (string, error) {
+	if *zoneOverride != "" {
+		return *zoneOverride, nil
+	}
+	return metadata.Zone()
```
vadimberezniker, line 166: You can check `metadata.OnGCE()` before calling `metadata.Zone()` to avoid trying to contact the metadata server if the app is not running on GCE. `metadata.OnGCE()` does a cached one time check to detect whether the binary is running on GCE.

vadimberezniker (approved):

server/resources/resources.go
vadimberezniker: Oh, I guess auto-merge doesn't care if you leave a comment welp.

PR #2054, opened 2022-05-20
Use a different range for firecracker host interfaces

luluz66:
vmIdx is mapped the following ip range:
     * `192.168.{vmIdx/30}.{(vmIdx%30) * 8) + 5}/30` on the host
     * `192.168.{vmIdx/30}.{(vmIdx%30) * 8) + 6}/30` in the namespace
      * `192.168.{vmIdx/30}.{(vmIdx%30) * 8) + 3}` as the host IP

We are not going to assign ip's in range `192.168.241.0/29` (tap addresses/vm addresses)

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1379

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -122,15 +122,6 @@ func createRandomVethPair(ctx context.Context, netNamespace string) (string, str
 	return veth0, veth1, err
 }
 
-func incrementIP(ip net.IP) {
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
LGTM!

(but see what @vadimberezniker thinks too)

vadimberezniker (approved):

PR #2055, opened 2022-05-20
Make public invocation executions effectively public

bduffany:
Checked the performance with invocations with large executions, looks unaffected by the additional join with the invocations table.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/execution_service/execution_service.go
+++ b/enterprise/server/execution_service/execution_service.go
@@ -40,9 +40,20 @@ func checkPreconditions(req *espb.GetExecutionRequest) error {
 func (es *ExecutionService) queryExecutions(ctx context.Context, baseQuery *query_builder.Query) ([]tables.Execution, error) {
 	dbh := es.env.GetDBHandle()
 	q := baseQuery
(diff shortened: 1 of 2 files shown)


enterprise/server/execution_service/execution_service.go
```diff
@@ -40,9 +40,16 @@ func checkPreconditions(req *espb.GetExecutionRequest) error {
 func (es *ExecutionService) queryExecutions(ctx context.Context, baseQuery *query_builder.Query) ([]tables.Execution, error) {
 	dbh := es.env.GetDBHandle()
 	q := baseQuery
-	if err := perms.AddPermissionsCheckToQueryWithTableAlias(ctx, es.env, q, "e"); err != nil {
+
+	permClauses, err := perms.GetPermissionsCheckClauses(ctx, es.env, q, "e")
+	if err != nil {
 		return nil, err
 	}
+	// If the invocation is public, allow viewing its execution metadata.
```
tylerwilliams, line 48: could you add a more detailed comment explaining why this is necessary?

bduffany: Done

tylerwilliams (approved):

PR #2059, opened 2022-05-20
Support partial reads in ByteStream API.

vadimberezniker:
Used to serve container layer ranges.

---

**Version bump**: None


tylerwilliams (changes requested):
This looks good -- could you add a few *simple* tests to check that offset / limit are respected? (fine if you want to write one and just c/p for each backend cache impl)

tylerwilliams (approved):

tylerwilliams:
Thank you!

PR #2063, opened 2022-05-23
Fix data races in runner_test

bduffany:
Adding even more locking to the runner pool (to protect `r.state`, `r.PlatformProperties`, and `r.task`) ... we can remove the need for locking once we have a proper fix for https://github.com/buildbuddy-io/buildbuddy-internal/issues/1159 (e.g., canceling the context on shutdown, instead of doing a concurrent call to `r.Remove()` while the task is executing)

I chose to lock around the runner pool's lock, rather than introducing per-runner locks, for (a) simplicity, and (b) to avoid introducing the possibility of deadlocks (i.e., having to worry about acquiring `p.mu` and `r.mu` in a consistent order)

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -158,6 +158,7 @@ type state int
 type commandRunner struct {
 	env            environment.Env
 	imageCacheAuth *container.ImageCacheAuthenticator
+	p              *pool
 
 	// ACL controls who can use this runner.
 	ACL *aclpb.ACL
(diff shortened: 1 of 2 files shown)


vadimberezniker (approved):
This seems ok as a stop-gap measure. 
Are you planning on tackling https://github.com/buildbuddy-io/buildbuddy-internal/issues/1159 soon?

bduffany:
> Are you planning on tackling [buildbuddy-io/buildbuddy-internal#1159](https://github.com/buildbuddy-io/buildbuddy-internal/issues/1159) soon?

Yes, I'd like to spend some time this week working on it and start testing this week.

PR #2064, opened 2022-05-23
Fix data race in scheduler server

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -1361,10 +1361,12 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 				// (in subsequent loop iterations) if the preferred node probe fails.
 				preferredNode = nil
 			} else {
+				nodeBalancer.mu.Lock()
 				nodes = nodeBalancer.nodes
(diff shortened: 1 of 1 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1361,10 +1361,12 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 				// (in subsequent loop iterations) if the preferred node probe fails.
 				preferredNode = nil
 			} else {
+				nodeBalancer.mu.Lock()
```
vadimberezniker, line 1364: I think it would be better to add methods to nodePool to retrieve the nodes under lock rather than accessing the mutex directly.

vadimberezniker (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
bduffany: Done, wdyt?

vadimberezniker: looks good!

vadimberezniker: thanks for fixing this

PR #2070, opened 2022-05-25
Remote persistent workers: Fix work request encoding

bduffany:
During the protobuf migration we switched from `buffer.EncodeMessage` to just `proto.Marshal`, which inadvertently dropped length-delimiting. This PR adds back the length delimiter, by mimicking the old implementation of `EncodeMessage` as closely as possible.

Tests to follow.

---

**Version bump**: Patch


tylerwilliams (approved):

tempoz (approved):
LGTM, Brandon; nice job.

PR #2073, opened 2022-05-26
Pebble disk cache

tylerwilliams:
Add a pebble cache (based off of raft storage work)

Want to get this in and iterate on it. The goal is to replace the existing in-memory LRU with a pebble database. Performance compares well with what we have today.

// TODO:
//  - add a pacer that deletes files as new ones are written when the cache is
//    above a certain size? TBD if this is necessary
//  - add a flag that ingests a disk cache


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -0,0 +1,801 @@
...
+	}
+	if p.isolation.GetCacheType() == rfpb.Isolation_ACTION_CACHE {
+		// for AC items, we need to examine the remote_instance_name as
+		// well and make sure there is a match.
+		if fileMetadata.GetFileRecord().GetIsolation().GetRemoteInstanceName() != p.isolation.GetRemoteInstanceName() {
+			return nil, status.NotFoundErrorf("file %q not found", fileMetadataKey)
+		}
+	}
+	rc, err := filestore.FileReader(ctx, p.blobDir(), fileMetadata.GetStorageMetadata().GetFileMetadata(), offset, limit)
+	if err != nil {
```
vadimberezniker, line 427: `if err == nil`?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -0,0 +1,801 @@
...
+	e.acCount = acCount
+	return nil
+}
+
+func (e *partitionEvictor) ttl() error {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	maxAllowedSize := int64(janitorCutoffThreshold * float64(e.part.MaxSizeBytes))
+	for {
+		if err := e.computeSize(); err != nil {
```
vadimberezniker, line 747: What's the overhead of recomputing this size every second?

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: Whoops good catch, thanks!

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: higher than I want it to eventually be, but I think it's fine for a mvp.

PR #2076, opened 2022-05-27
Bug fix: Populate flags in the same order they are specified in the YAML

tempoz:
This fixes a bug with `FlagAlias` in which if a flag and an alias flag were both
specified in YAML with different values, the behavior was non-deterministic. Now
the flags are guaranteed to be populated in the order specified in the YAML, so
for assignment-based flags the value will be the last occurence of a flag or an
alias flag, and append-based flags will have their contents appended in order.

---

**Version bump**: Patch


tempoz:
* master
  * **PR #2076** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2076" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2076?utm_source=stack-comment).

tylerwilliams (approved):

PR #2079, opened 2022-05-31
Don't parse and fill output files for actions that were not requested

siggisim:
This contains an expensive regex, which gets slow for invocations with a lot of output files.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/api/api_server.go
+++ b/enterprise/server/api/api_server.go
@@ -184,6 +184,7 @@ func (s *APIServer) GetAction(ctx context.Context, req *apipb.GetActionRequest) 
 
 		// Filter to only selected actions.
 		if action != nil && actionMatchesActionSelector(action, req.GetSelector()) {
+			action = fillActionOutputFilesFromBuildEvent(action, event.BuildEvent)
 			actions = append(actions, action)
 		}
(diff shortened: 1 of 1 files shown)


enterprise/server/api/api_server.go
```diff
@@ -377,6 +377,21 @@ func fillActionFromBuildEvent(action *apipb.Action, event *build_event_stream.Bu
 			action.Id.TargetId = encodeID(event.GetId().GetTestResult().GetLabel())
 			action.Id.ConfigurationId = event.GetId().GetTestResult().GetConfiguration().Id
 			action.Id.ActionId = encodeID(fmt.Sprintf("test-S_%d-R_%d-A_%d", testResultID.Shard, testResultID.Run, testResultID.Attempt))
+			return action
+		}
+	}
+	return nil
+}
+
+func fillActionOutputFilesFromBuildEvent(action *apipb.Action, event *build_event_stream.BuildEvent) *apipb.Action {
```
tylerwilliams, line 386: typically the "output" param comes last in functions where you take both inputs and outputs as pointers

tylerwilliams (approved):
LGTM

PR #2089, opened 2022-06-02
Fix prometheus label docs

bduffany:
* Make the "Label" suffix for label consts optional (it was preventing the documentation from being generated correctly)
* Clean up a few comments to make them more suitable for the docs page

---

**Version bump**: Patch


tempoz (approved):
LGTM

PR #2094, opened 2022-06-03
Fix explicit/effective command lines with multiple pattern components

siggisim:
Currently these render without spaces, i.e.
```
bazel build targetonetargettwotargetthree --config=local
```

With this change we join the pattern components with spaces, i.e.
```
bazel build targetone targettwo targetthree --config=local
```

---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation_details_card.tsx
+++ b/app/invocation/invocation_details_card.tsx
@@ -215,15 +215,18 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
                       className="copy-icon"
                       onClick={this.handleCopyClicked.bind(
                         this,
-                        `bazel ${this.props.model.started?.command} ${
-                          this.props.model.expanded?.id?.pattern?.pattern
-                        } ${this.props.model.optionsParsed?.explicitCmdLine.join(" ")}`
+                        `bazel ${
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_details_card.tsx
```diff
@@ -215,15 +215,18 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
                       className="copy-icon"
                       onClick={this.handleCopyClicked.bind(
                         this,
-                        `bazel ${this.props.model.started?.command} ${
-                          this.props.model.expanded?.id?.pattern?.pattern
-                        } ${this.props.model.optionsParsed?.explicitCmdLine.join(" ")}`
+                        `bazel ${
+                          this.props.model.started?.command
```
tylerwilliams, line 219: could we just have a "formatCommand" func that returned the pretty and correct version?

tylerwilliams (approved):

app/invocation/invocation_details_card.tsx
siggisim: Pulled out the common bits into a function

tylerwilliams (approved):

tylerwilliams (approved):

PR #2095, opened 2022-06-03
Remove docker dep vendoring.

vadimberezniker:
I need to update the docker library version and it looks like the
loading performance issue is no longer there (likely fixed in one of the
gazelle updates.)

---

**Version bump**: None


tylerwilliams (approved):

siggisim (approved):
Woot! Can we remove the Google Cloud one too?

vadimberezniker:
> Woot! Can we remove the Google Cloud one too?

Good point, I'll test it out and send a separate PR if it doesn't have issues anymore.

PR #2097, opened 2022-06-03
Registry for serving optimized container images.

vadimberezniker:
Two standard endpoints:
 - /v2/\<image>/manifests/\<ref> for serving manifests or image indices
 - /v2/\<image>/blobs/\<ref> for serving full or partial layers

Non-standard endpoint for getting the optimized manifest reference of a
given image:
  - /v2/\<image>/optimizedManifests/\<ref>
  This endpoint allows us to rewrite image requests for specific
digests. Normally if a client requests image at ref sha256:foo, they
will error out if the returned manifest does not have the correct
digest.

Optimized manifest information is stored in a blob store whereas the
layer data itself is stored in the CAS.

---

**Version bump**: None


tylerwilliams (approved):
this is pretty dope :)

PR #2102, opened 2022-06-06
Retry should be limited by the max backoff if no maximum number of tries is specified

tempoz:
Also changes invocation db update to use the default retry object and adds some functions
useful for logging.

---

**Version bump**: Patch


tempoz:
* master
  * **PR #2102** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2102" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2102?utm_source=stack-comment).

server/util/retry/retry.go
```diff
@@ -12,21 +12,28 @@ type Options struct {
...
 
 	currentAttempt int
+	maxAttempts    int
 	isReset        bool
 }
 
 func New(ctx context.Context, opts *Options) *Retry {
+	maxAttempts := opts.MaxRetries
+	if maxAttempts <= 0 {
+		maxAttempts = int(math.Ceil(math.Log(float64(opts.MaxBackoff)/float64(opts.InitialBackoff)) / math.Log(opts.Multiplier)))
```
bduffany, line 30: this seems difficult to modify down the road if we want to add things like random jitter to this util -- wdyt about directly detecting when we're at the max backoff instead?

bduffany, line 30: Since this logic is separate from the backoff logic, it seems more prone to going out of sync. Mind adding a test for a few different cases to verify that the formula works as intended?

tempoz: Done!

bduffany (approved):
LGTM - looks like the only non-test usages of `retry.DefaultWithContext` are under `enterprise/server/raft/...` which I don't have much context for, so it'd be good to double-check w/ Tyler that the behavior change is OK

tylerwilliams (approved):

server/util/retry/retry_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/retry"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMaxRetryFromBackoff(t *testing.T) {
```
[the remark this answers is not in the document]
tempoz: Oooooooh, I like this style a lot! Done.

PR #2103, opened 2022-06-06
A little locking goes a long way: prefer Rlock for contains checks.

tylerwilliams:
This permits substantially more QPS on the FindMissing endpoint by reducing contention.

In my local testing this increased /FindMissingBlobs QPS (where each batch contained 100 digests) from ~3K to ~12K.

### DIFF
--- a/server/backends/disk_cache/disk_cache.go
+++ b/server/backends/disk_cache/disk_cache.go
@@ -446,8 +446,8 @@ func (p *partition) WaitUntilMapped() {
 }
 
 func (p *partition) Statusz(ctx context.Context) string {
-	p.mu.Lock()
-	defer p.mu.Unlock()
+	p.mu.RLock()
+	defer p.mu.RUnlock()
 	buf := "<br>"
 	buf += fmt.Sprintf("<div>Partition %q</div>", p.id)
 	buf += fmt.Sprintf("<div>Root directory: %s</div>", p.rootDir)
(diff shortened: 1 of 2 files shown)


siggisim (approved):

vadimberezniker (approved):

server/backends/disk_cache/disk_cache.go
```diff
@@ -718,15 +718,18 @@ func (p *partition) contains(ctx context.Context, cacheType interfaces.CacheType
...
+	isMapped := p.diskIsMapped
+	p.mu.RUnlock()
 
-	if !ok && !p.diskIsMapped {
+	if !ok && !isMapped {
 		// OK if we're here it means the disk contents are still being loaded
 		// into the LRU. But we still need to return an answer! So we'll go
 		// check the FS, and if the file is there we'll add it to the LRU.
+		p.mu.Lock()
 		ok = p.addFileToLRUIfExists(k)
```
tempoz, line 731: Are we concerned at all that the state of the var `p.diskIsMapped` may change between when we assign it to `isMapped` and when we actually call `p.addFileToLRUIfExists`?

tylerwilliams: Yes, I did check this -- since addFileToLRUIfExists checks that var again, I think this code is safe?

tylerwilliams: (That said, the test is currently failing so there might be some state change wonkiness here still to debug)

tylerwilliams: slight tweak -- same lift and test passes.

PR #2104, opened 2022-06-06
Revert "A little locking goes a long way: prefer Rlock for contains checks."

tylerwilliams:
Reverts buildbuddy-io/buildbuddy#2103

fixing. a. race. condition.


tempoz (approved):
LGTM

PR #2106, opened 2022-06-07
Enable a simple metrics-collector API cache

tylerwilliams:
- Slight refactor: move some common API functionality to server/api/common
- Extract Actions and their associated files as the build stream is received, store in metrics collector list, expiry 1 hr
- When enabled, read from the metrics-collected cache in the API. If any cached value is found, it is assumed to be the full response.


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -826,6 +832,10 @@ func (e *EventChannel) processSingleEvent(event *inpb.InvocationEvent, iid strin
 	e.targetTracker.TrackTargetsForEvent(e.ctx, event.BuildEvent)
 	e.statusReporter.ReportStatusForEvent(e.ctx, event.BuildEvent)
 
+	if err := e.collectAPIFacets(iid, event.BuildEvent); err != nil {
```
siggisim, line 835: Should we hate this behind the cache enabled flag so we can turn it off easily if it causes problems?

siggisim (approved):
Thanks for doing this so quick!

enterprise/server/api/api_server.go
```diff
@@ -166,32 +164,50 @@ func (s *APIServer) GetAction(ctx context.Context, req *apipb.GetActionRequest)
...
+	}
+
+	// This limit is a safeguard against buffering too many results in memory.
+	// We expect to hit this rarely (if ever) in production usage.
+	const limit = 100_000
+	if *enableCache && s.env.GetMetricsCollector() != nil {
+		serializedResults, err := s.env.GetMetricsCollector().ListRange(ctx, api_common.ActionsKey(iid), 0, limit-1)
+		if err == nil {
+			a := &apipb.Action{}
+			for _, serializedResult := range serializedResults {
```
siggisim, line 179: Do we need to filter this down to just actions that match the selector?

siggisim (approved):

server/build_event_protocol/build_event_handler/build_event_handler.go
tylerwilliams: Done.

enterprise/server/api/api_server.go
tylerwilliams: Done.

PR #2108, opened 2022-06-07
Update images

tempoz:
Updates the container base versions to pick up fixes for vulnerabilities.

---

**Version bump**: Patch


tempoz:
Addresses buildbuddy-io/buildbuddy-internal#1420

PR #2111, opened 2022-06-07
Compress more better faster stronger

tylerwilliams:


siggisim (approved):

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -883,18 +898,32 @@ func (e *EventChannel) collectAPIFacets(iid string, event *build_event_stream.Bu
...
+		wf := writerFunc(func(data []byte) (int, error) {
+			if err := e.collector.ListAppend(e.ctx, key, string(data)); err != nil {
+				return 0, err
+			}
+			if err := e.collector.Expire(e.ctx, key, apiFacetsExpiration); err != nil {
+				return 0, err
+			}
+			return len(data), nil
+		})
+		w, err := zstd.NewWriter(wf)
```
bduffany, line 916: If each of these writers processes only a small amount of data (less than 1MB or so) then it might be worth benchmarking this vs. chunk-by-chunk compression (e.g. `ListAppend(..., compression.CompressZstd(data))`) -- I remember experiencing some issues with the NewWriter API for small chunks, and getting significantly better performance by having the NewZstdChunkingCompressor reuse a single Writer instance.

enterprise/server/api/api_server.go
```diff
@@ -147,6 +148,42 @@ func (s *APIServer) GetTarget(ctx context.Context, req *apipb.GetTargetRequest)
...
+	if err != nil {
+		return nil, err
+	}
+
+	actions := make([]*apipb.Action, 0)
+	a := &apipb.Action{}
+	buf := make([]byte, 100)
+
+	for _, compressed := range serializedResults {
+		raw, err := dec.DecodeAll([]byte(compressed), buf[:0])
```
bduffany, line 173: might make sense to do `compression.DecompressZstd` here so that you go through the DecoderPool

PR #2112, opened 2022-06-07
Move registration of app-specific services out of general gRPC utils.

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):
Thank you

PR #2114, opened 2022-06-08
Add RPC endpoint to registry to retrieve optimized image ref.

vadimberezniker:
This RPC is also responsible for validating that the client has access
to the target image.

---

**Version bump**: None


enterprise/server/registry/registry.go
```diff
@@ -413,8 +422,8 @@ func (r *registry) writeManifest(ctx context.Context, oldManifest partial.Descri
 	return nil
 }
 
-func (r *registry) convert(ctx context.Context, req *request, remoteDesc *remote.Descriptor) (*regpb.Manifest, error) {
-	req.logger.Infof("Converting %q", remoteDesc.Ref)
+func (r *registry) convert(ctx context.Context, reqLogger log.Logger, remoteDesc *remote.Descriptor) (*regpb.Manifest, error) {
```
tylerwilliams, line 425: I'd prefer it if we didn't pass loggers around -- it's more allocations and less readability for not a ton of gain. I understand the need to sometimes have a custom logger that logs additional stuff, but the idea of doing it per-request is too much.

But if you wanted to log request ID in the logger (like I think you're doing here?), I'd def be into variants of log.Printf/log.Warningf/etc that take a context and add a UUID tag?

enterprise/server/registry/registry.go
```diff
@@ -737,10 +718,155 @@ type registry struct {
...
+			if err != nil {
+				return err
+			}
+			_, err = l.Size()
+			return err
+		})
+	}
+	if err := eg.Wait(); err != nil {
+		if t, ok := err.(*transport.Error); ok && t.StatusCode == http.StatusUnauthorized {
+			return status.PermissionDeniedErrorf("could not retrieve image layer from remote: %s", err)
```
tylerwilliams, line 753: nit: retrieve

enterprise/server/registry/registry.go
```diff
@@ -737,10 +718,155 @@ type registry struct {
...
+			}
+			layerRef := imgRef.Context().Digest(d.String())
+			l, err := remote.Layer(layerRef, remoteOpts...)
+			if err != nil {
+				return err
+			}
+			_, err = l.Size()
+			return err
+		})
+	}
```
tylerwilliams, line 750: i don't totally understand how many remote requests this is making -- but is it really necessary to fetch all the layers in order to check access?

enterprise/server/registry/registry.go
```diff
@@ -813,7 +928,13 @@ func main() {
 	flag.Parse()
 
 	if err := flagutil.PopulateFlagsFromFile(config.Path()); err != nil {
-		log.Fatalf("Error loading config from file: %s", err)
+		fmt.Printf("Error loading config from file: %s", err)
+		os.Exit(1)
```
tylerwilliams, line 932: I see you changed this from log.Fatalf -- what's the advantage of doing it this way?

proto/registry.proto
```diff
@@ -17,3 +17,37 @@ message Manifest {
...
+message Credentials {
+  string username = 1;
+  string password = 2;
+}
+
+message Platform {
+  string arch = 1;
+  string os = 2;
+  string variant = 3;
+}
```
tylerwilliams, line 30: nit: would be nice to comment these and add examples of each

proto/registry.proto
```diff
@@ -17,3 +17,37 @@ message Manifest {
...
+  // If an image has multiple variants, this selector is used
+  // to select a single variant.
+  Platform platform = 2;
+
+  // Optional credentials for accessing the target image.
+  Credentials image_credentials = 3;
+}
+
+message GetOptimizedImageResponse {
+  string optimized_image = 1;
```
tylerwilliams, line 45: same here -- it's nice to have examples of each field to help the reader understand what they're getting

tylerwilliams (changes requested):
only minor complaint is the logger stuff, otherwise looks nice :)

enterprise/server/registry/registry.go
tylerwilliams: lol whoops meant to delete this i can't read

tylerwilliams (approved):

PR #2117, opened 2022-06-08
add schema for quota configs

luluz66:
QuotaBucket is to define the sustained rate and max burst.
QuotaGroup is to define the relationship between group ID / IP and the
QuotaBucket.

Note:
To save the space, for each namespace, we will define a default bucket. If a group id / ip is not in the QuotaGroup table, then quota manager will apply the default bucket.

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -548,8 +548,12 @@ func (*Usage) TableName() string {
 
 type QuotaBucket struct {
 	Model
+	// The namespace indicates a single resource to be protected from abusive
+	// usage.
 	Namespace string `gorm:"primarykey"`
-	Name      string `gorm:"primarykey"`
+
+	// The name of the bucket, like "default", "banned", or "restricted".
+	Name string `gorm:"primarykey"`
 
 	// The maximum sustained rate of requests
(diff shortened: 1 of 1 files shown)


tylerwilliams (changes requested):
This is the nomenclature as I understand it:

 - Limit: How many requests can be used in a time period
 - Period: The length of time over which the limit applies
 - Burst: The number of requests that will be allowed to exceed the rate in a single burst.
 - Namespace: A single "resource", like 1 RPC. 
 - QuotaKey: The GroupID or IP Address (for anon clients). Used to count quota for a single "user".
 - Bucket: A set quota limit+period+burst applicable to a single namespace.
 - Name: The descriptive name for a particular bucket, like "default", "banned", or "restricted"
 - QuotaGroup: A matching of a QuotaBucket to a QuotaKey (basically user:X is in bucket:restricted)

Which I think makes sense. Namespace will probably be auto-generated by using the RPC name or something.

Can you link these tables with an ID or fk relationship so that we can support renaming buckets without having to recreate rows in the quota_groups table?

Otherwise lg

tylerwilliams (approved):

PR #2129, opened 2022-06-09
Tree cache safety

tylerwilliams:
Be more cautious and verify TreeCache integrity before caching or using a cached value. Also verify Directory protos are non-empty.


siggisim:
Still haven't managed to repro, but found a hypothetical case that would be fixed by: https://github.com/buildbuddy-io/buildbuddy/pull/2133

siggisim (approved):

PR #2068, opened 2022-05-24
Separate yaml-handling and flag types out from flagutil.go

tempoz:
`flagutil.go` was getting very large, and the yaml-handling was semantically
very different from modifying the flags directly or defining new flag types, so we
separate it out. We also take this opportunity to move the flag types to their own
file.

---

**Version bump**: Patch

PR #2130, opened 2022-06-10
Enable deprecation of flags

tempoz:
---

**Version bump**: Minor

### DIFF
--- a/server/util/flagutil/common/common.go
+++ b/server/util/flagutil/common/common.go
@@ -38,28 +38,36 @@ func flagTypeFromFlagFuncName(name string) reflect.Type {
 }
 
 type TypeAliased interface {
+	// AliasedType returns the type this flag.Value aliases.
 	AliasedType() reflect.Type
 }
 
 type IsNameAliasing interface {
+	// AliasedName returns the flag name this flag.Value aliases.
 	AliasedName() string
 }
 
 type WrappingValue interface {
+	// WrappedValue returns the value this flag.Value wraps.
 	WrappedValue() flag.Value
 }
 
 type Appendable interface {
+	// AppendSlice appends the passed slice to this flag.Value.
 	AppendSlice(any) error
 }
 
 type DocumentNodeOption interface {
+	// Transform transforms the passed yaml.Node in place.
 	Transform(in any, n *yaml.Node)
+	// Passthrough returns whether this option should be passed to child nodes.
 	Passthrough() bool
 }
 
-type SetValueHooked interface {
-	SetValueHook()
+type SetValueForFlagNameHooked interface {
+	// SetValueForFlagNameHooked is the hook for flags that is called when the
+	// flag.Value is set by name.
+	SetValueForFlagNameHook()
 }
 
 // GetTypeForFlagValue returns the (pointer) Type this flag aliases; this is the same
@@ -77,7 +85,7 @@ func GetTypeForFlagValue(value flag.Value) (reflect.Type, error) {
 }
 
 // SetValueForFlagName sets the value for a flag by name.
-func SetValueForFlagName(name string, i any, setFlags map[string]struct{}, appendSlice bool, strict bool) error {
+func SetValueForFlagName(name string, i any, setFlags map[string]struct{}, appendSlice bool, force bool, strict bool) error {
 	flg := DefaultFlagSet.Lookup(name)
 	if flg == nil {
 		if strict {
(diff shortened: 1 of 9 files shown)


tempoz:
* master
  * **PR #2068** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2068" target="_blank">Graphite</a> 
    * **PR #2130** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2130" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2130?utm_source=stack-comment).

server/util/flagutil/types/types.go
```diff
@@ -195,20 +221,29 @@ func (f *FlagAlias) AliasedName() string {
...
+	log.Warningf("Flag %s is deprecated and was set on the command line.", d.name)
+	return d.Value.Set(value)
+}
+
+func (d *DeprecatedFlag) WrappedValue() flag.Value {
+	return d.Value
+}
+
+func (d *DeprecatedFlag) SetValueHook() {
+	log.Warningf("Flag %s is deprecated and was set programmatically, likely through the YAML config.", d.name)
```
bduffany, line 248: nbd if it's too complicated w/ the current design, but is there a way to differentiate whether this was set via YAML vs. something else? Thinking the error message would be  more clear for end users if we could remove the "and was set programmatically" part.

Also, can we print the flag help string here so that users know what the migration path is? e.g.

`log.Warningf("Flag %q (set via YAML config) is deprecated: %s", name, migrationPath)`

e.g.

`Flag "foo" (set via YAML config) is deprecated: no longer has any effect, please remove`

server/util/flagutil/types/types.go
```diff
@@ -195,20 +221,29 @@ func (f *FlagAlias) AliasedName() string {
...
-	if err != nil {
-		return reflect.TypeOf(flg.Value)
-	}
-	return t
+type DeprecatedFlag struct {
+	flag.Value
+	name string
+}
+
+func DeprecatedVar[T any](value flag.Value, name string, usage string) *T {
```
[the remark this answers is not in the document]
bduffany: (to provide an even stronger forcing function you could also require a struct, like  `type MigrationPlan struct { message string }` and take that as a param)

server/util/flagutil/common/common.go
```diff
@@ -38,64 +38,111 @@ func flagTypeFromFlagFuncName(name string) reflect.Type {
...
+	} else if v, ok := value.(TypeAliased); ok {
 		return v.AliasedType(), nil
 	}
-	return nil, status.UnimplementedErrorf("Unsupported flag type at %s: %T", flg.Name, flg.Value)
+	return nil, status.UnimplementedErrorf("Unsupported flag type : %T", value)
 }
 
 // SetValueForFlagName sets the value for a flag by name.
-func SetValueForFlagName(name string, i any, setFlags map[string]struct{}, appendSlice bool, strict bool) error {
+func SetValueForFlagName(name string, i any, setFlags map[string]struct{}, appendSlice bool, force bool, strict bool) error {
```
bduffany, line 88: nit: document what these bool params do?

server/util/flagutil/common/common.go
```diff
@@ -38,64 +38,111 @@ func flagTypeFromFlagFuncName(name string) reflect.Type {
...
+func setValueFromFlagName(value flag.Value, name string, i any, setFlags map[string]struct{}, appendSlice bool, force bool, setHooks ...func()) error {
+	if v, ok := value.(SetValueForFlagNameHooked); ok {
+		setHooks = append(setHooks, v.SetValueForFlagNameHook)
+	}
+	return SetValueWithCustomIndirectBehavior(value, name, i, setFlags, appendSlice, force, setValueFromFlagName, setHooks...)
+}
+
+type SetValueForIndirectFxn func(value flag.Value, name string, i any, setFlags map[string]struct{}, appendSlice bool, force bool, setHooks ...func()) error
+
+func SetValueWithCustomIndirectBehavior(value flag.Value, name string, i any, setFlags map[string]struct{}, appendSlice bool, force bool, setValueForIndirect SetValueForIndirectFxn, setHooks ...func()) error {
```
bduffany, line 108: nit: document

bduffany (approved):

server/util/flagutil/common/common.go
tempoz: Documented one and removed the other two, they're not actually necessary anymore.

server/util/flagutil/common/common.go
tempoz: Done!

PR #2133, opened 2022-06-10
Validate directory has contents before returning it in GetTree

siggisim:
Haven't managed to trigger this case in testing yet (only artificially by randomly setting a blob in `rspMap` to empty bytes / empty proto) - but this would fix a hypothetical issue that would cause a cached:
```
Failed Precondition: Digest hash:"60430d8c02074632d6ca79aa65d7b261dce5b87798c5424fe061625bae630644" size_bytes:157 not found
  Precondition Failure:
    (MISSING) blobs/60430d8c02074632d6ca79aa65d7b261dce5b87798c5424fe061625bae630644/157:
```

Investigating further to see if I can figure out how to convince GetMulti to return empty bytes (searched server logs for "returned a zero-length response for" but got no hits).

---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
+++ b/server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
@@ -500,6 +500,10 @@ func (s *ContentAddressableStorageServer) fetchDirectory(ctx context.Context, ca
 		if err := proto.Unmarshal(blob, subDir); err != nil {
 			return nil, err
 		}
+		if len(subDir.Directories) == 0 && len(subDir.Files) == 0 {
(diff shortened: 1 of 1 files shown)


server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -500,6 +500,10 @@ func (s *ContentAddressableStorageServer) fetchDirectory(ctx context.Context, ca
 		if err := proto.Unmarshal(blob, subDir); err != nil {
 			return nil, err
 		}
+		if len(subDir.Directories) == 0 && len(subDir.Files) == 0 {
```
tylerwilliams, line 503: do we need to check symlink nodes too?

PR #2135, opened 2022-06-13
Use IP as fallback when GroupID is not set in the context

luluz66:


server/quota/quota_manager.go
```diff
@@ -246,13 +247,26 @@ func (qm *QuotaManager) getGroupID(ctx context.Context) string {
...
+
 func (qm *QuotaManager) getKey(ctx context.Context) string {
-	groupID := qm.getGroupID(ctx)
-	if groupID != "" {
+	if groupID := qm.getGroupID(ctx); groupID != "" {
 		return groupID
 	}
-	//TODO(lulu): find IP address
-	return ""
+	return getIP(ctx)
```
tylerwilliams, line 269: we should check getKey hasn't returned an empty string (in Allow or wherever else) just so that we know we're not doing something funky

luluz66: Done

tylerwilliams (approved):

PR #2136, opened 2022-06-14
Podman changes to support image streaming.

vadimberezniker:
On executor startup, we start the stargz store as a child process and
configured podman to look for image information in the store.

When a new pull is requested, we consult the registry to obtain the
optimized image name. This pull is also used for revalidating the
container credentials.

This change support registry auth, but the CAS data is still not
segregated by group. That will require another change.

---

**Version bump**: Minor


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -43,13 +54,114 @@ const (
...
+
+	return &Provider{
+		env:                   env,
+		imageCacheAuth:        imageCacheAuthenticator,
+		buildRoot:             buildRoot,
+		imageStreamingEnabled: imageStreamingEnabled,
+		regClient:             regClient,
+		optImageCache:         c,
+	}, nil
+}
```
tylerwilliams, line 150: Because most of the methods we use are defined on podmanCommandContainer, it seems like we could move most of this code to an initX() error type function on commandContainer and run it behind a sync.Once. That would then let us remove the Provider struct entirely and leave the runner code as-is, which I think would be a little cleaner / simpler?

tylerwilliams (changes requested):

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -15,15 +19,22 @@ import (
...
+	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 
+	regpb "github.com/buildbuddy-io/buildbuddy/proto/registry"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 )
 
 var (
+	imageStreamingRegistryGRPCTarget = flag.String("executor.podman_image_streaming.registry_grpc_target", "", "gRPC endpoint of BuildBuddy registry")
```
bduffany, line 35: wdyt about `s/podman_/podman./` in the flag name? (I added a flag `executor.podman.enable_stats` in separate PR, so that podman namespace exists now)

enterprise/server/remote_execution/containers/podman/podman.go
vadimberezniker: I think having another type responsible for the overall podman configuration is cleaner than hiding the logic behind `sync.Once`.

tylerwilliams (approved):

PR #2137, opened 2022-06-14
Remove stale executors if the registration is older than 10 minutes.

vadimberezniker:
Executors will now continously update their registration instead of only
when idle. Scheduler will remove any executors that have not updated
their registration within the last 10 minutes.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1419

---

**Version bump**: Patch


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -41,6 +42,7 @@ var (
 	defaultPoolName              = flag.String("remote_execution.default_pool_name", "", "The default executor pool to use if one is not specified.")
 	sharedExecutorPoolGroupID    = flag.String("remote_execution.shared_executor_pool_group_id", "", "Group ID that owns the shared executor pool.")
 	requireExecutorAuthorization = flag.Bool("remote_execution.require_executor_authorization", false, "If true, executors connecting to this server must provide a valid executor API key.")
+	removeStaleExecutors         = flag.Bool("remote_execution.remove_stale_executors", false, "If true, executors are removed if they are not heard from for a prolonged amount of time.")
```
tylerwilliams, line 45: do you want to default this to true?

tylerwilliams (approved):

siggisim:
Any chance we can get this in before tomorrow's release? Not sure who the release master is 
cc @vadimberezniker @bduffany @luluz66

bduffany:
@siggisim Based on Vadim's comment:

> The executors need to be updated first otherwise they can be removed prematurely. This is especially true for BYOR.

 it sounds like we need to wait for older executor versions to be updated before we can realistically enable this? or am I misunderstanding?

bduffany (approved):

siggisim:
Yeah we can commit with flag set to false, and then flip to true once we're sure executors have updated.

This way on-prem folks can start using it sooner.

PR #2138, opened 2022-06-14
Task sizing: Add an option to enable cgroup-based podman stats

bduffany:
Adds an option `EnableStats` to the podman API that periodically collects CPU usage (nanos) and mem usage (bytes) while the container is running, and reports these in the `CommandResult` (both for `Run` and `Exec`). It also allows the `Stats()` func to work properly, which is used for runner pool resource management.

The implementation is based on polling the cgroupfs. Locally, each cgroupfs read takes ~10 microseconds, so we should be able to sample these stats at a fairly high rate.

Alternatives considered:
* Using the `getrusage` system call - I found this to be unreliable. In particular, `maxrss` only reports the max resident set size amongst all child processes individually, as in `max(child1_maxrss, ..., childN_maxrss)` and doesn't represent the max total memory usage across all child processes. This is problematic for any actions that use process-based concurrency. It also would introduce complexity and performance overhead since we'd have to add a wrapper binary for every task which calls `getrusage(RUSAGE_CHILDREN)` and writes it to a file.
* Using `podman stats` - I found `podman stats` to yield an error in dev - I haven't yet dug in as to why, but also `podman stats` requires some overhead of spinning up a process and parsing JSON output (the output uses human-readable abbreviations so the parsing would be pretty ugly and imprecise). We would also be limited to a 1s polling interval since the podman CLI throws an error if you try to specify anything less than that.
 
Overall I figured since `podman stats` is getting its data from the cgroupfs anyway, it's not too bad to read from the cgroup files directly, save for the configuration hairiness.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -46,13 +46,16 @@ type DockerDeviceMapping struct {
 	CgroupPermissions string `yaml:"cgroup_permissions" usage:"cgroup permissions that should be assigned to device."`
 }
 
-// Stats holds represents a container's held resources.
+// Stats represents a container's resource usage.
 type Stats struct {
+	// MemoryUsageBytes is the most recent memory usage value sampled from the
+	// container, in bytes.
 	MemoryUsageBytes int64
-
-	// TODO: add CPU usage once we have a reliable way to measure it.
-	// CPU only applies to bare execution, since Docker containers
-	// are frozen when not in use, reducing their CPU usage to 0.
+	// CPUNanos is the total CPU usage used by a task, in CPU-nanoseconds.
+	CPUNanos int64
+	// PeakMemoryUsageBytes is the highest memory usage sampled during the
+	// execution of a task, in bytes.
+	PeakMemoryUsageBytes int64
 }
 
 type FileSystemLayout struct {
--- a/enterprise/server/remote_execution/containers/podman/BUILD
+++ b/enterprise/server/remote_execution/containers/podman/BUILD
@@ -13,9 +13,11 @@ go_library(
         "//server/interfaces",
         "//server/util/alert",
         "//server/util/background",
+        "//server/util/disk",
         "//server/util/log",
         "//server/util/networking",
         "//server/util/random",
+        "//server/util/retry",
         "//server/util/status",
     ],
 )
@@ -23,6 +25,13 @@ go_library(
 go_test(
     name = "podman_test",
     srcs = ["podman_test.go"],
+    args = [
+        # Note: these are used only for the Stats tests, which require running as root.
+        # These paths were tested on Ubuntu 20.04 + podman 3.4.2, but may be different for other setups.
+        # See the comment in podman.go for more info.
+        "--executor.podman.memory_usage_path_template=/sys/fs/cgroup/memory/machine.slice/libpod-{{.ContainerID}}.scope/memory.usage_in_bytes",
+        "--executor.podman.cpu_usage_path_template=/sys/fs/cgroup/cpu,cpuacct/machine.slice/libpod-{{.ContainerID}}.scope/cpuacct.usage",
+    ],
     tags = [
         "manual",
         "no-sandbox",
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -2,11 +2,16 @@ package podman
 
 import (
 	"context"
+	"flag"
 	"fmt"
+	"io"
+	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/commandutil"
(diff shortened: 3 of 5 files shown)


enterprise/server/remote_execution/containers/podman/BUILD
```diff
@@ -13,16 +13,25 @@ go_library(
...
         "//server/util/status",
     ],
 )
 
 go_test(
     name = "podman_test",
     srcs = ["podman_test.go"],
+    args = [
+        # Note: these are used only for the Stats tests, which require running as root.
+        # These paths were tested on Ubuntu 20.04 + podman 3.4.2, but may be different for other setups.
```
luluz66, line 30: We are using podman 4 in dev/prod. Maybe worth testing on podman 4.

enterprise/server/remote_execution/containers/podman/podman_test.go
```diff
@@ -349,3 +351,43 @@ func TestPodmanRun_LongRunningProcess_CanGetAllLogs(t *testing.T) {
...
+	// Note: This test requires root. Under cgroup v2, root is not required, but
+	// some devs' machines are running Ubuntu 20.04 currently, which only has
+	// cgroup v1 enabled (enabling cgroup v2 requires modifying kernel boot
+	// params).
+	u, err := user.Current()
+	require.NoError(t, err)
+	if u.Uid != "0" {
+		t.Skip("Test requires root")
+	}
+	// podman needs iptables which is in /usr/sbin.
```
luluz66, line 365: I thought we only call iptables when the second network is enabled.

enterprise/server/remote_execution/containers/podman/BUILD
bduffany: Tested e2e with podman 4 in dev -- here is a chart using the metrics from #2139
 
image

enterprise/server/remote_execution/container/container.go
```diff
@@ -46,13 +46,16 @@ type DockerDeviceMapping struct {
...
 	MemoryUsageBytes int64
-
-	// TODO: add CPU usage once we have a reliable way to measure it.
-	// CPU only applies to bare execution, since Docker containers
-	// are frozen when not in use, reducing their CPU usage to 0.
+	// CPUNanos is the total CPU usage used by a task, in CPU-nanoseconds.
+	CPUNanos int64
+	// PeakMemoryUsageBytes is the highest memory usage sampled during the
+	// execution of a task, in bytes.
+	PeakMemoryUsageBytes int64
```
tylerwilliams, line 58: Do we have a use case in mind for MemoryUsageBytes? It seems like we'll always need to alloc at least PeakMemoryUsageBytes anyway to avoid OOMs, so I'm unclear on why we need both.

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -15,15 +20,25 @@ import (
...
 
 var (
+	// Note: to get these cgroup paths, run a podman container like
+	//     podman run --rm --name sleepy busybox sleep infinity
+	// then look at the output of
+	//     find /sys/fs/cgroup | grep libpod-$(podman container inspect sleepy | jq -r '.[0].Id')
+
+	memUsagePathTemplate = flag.String("executor.podman.memory_usage_path_template", "/sys/fs/cgroup/memory/libpod_parent/libpod-{{.ContainerID}}/memory.usage_in_bytes", "Go template specifying a path pointing to a container's current memory usage, in bytes. Templated with `ContainerID`.")
+	cpuUsagePathTemplate = flag.String("executor.podman.cpu_usage_path_template", "/sys/fs/cgroup/cpuacct/libpod_parent/libpod-{{.ContainerID}}/cpuacct.usage", "Go template specifying a path pointing to a container's total CPU usage, in CPU nanoseconds. Templated with `ContainerID`.")
+
```
tylerwilliams, line 41: Could we have a setup function that determines these paths when podman first runs (sync.Once style)?

I get a little scared of these super longs paths like this that need to be specified and what happens when they don't work on some on-prem system that is different than ours.

Or if you are pretty sure they will never change, I'm happy to roll with this.

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -43,6 +58,9 @@ const (
 	podmanDefaultNetworkIPRange = "10.88.0.0/16"
 	podmanDefaultNetworkGateway = "10.88.0.1"
 	podmanDefaultNetworkBridge  = "cni-podman0"
+
+	pollCIDTimeout    = 15 * time.Second
+	statsPollInterval = 50 * time.Millisecond
```
tylerwilliams, line 63: please add comments about what this control

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -288,6 +341,90 @@ func (c *podmanCommandContainer) PullImage(ctx context.Context, creds container.
...
+		if !c.options.EnableStats {
+			return
+		}
+		var last *container.Stats
+		var lastErr error
+
+		start := time.Now()
+		defer func() {
+			// Only log an error if the task ran long enough that we could reasonably
+			// expect to sample stats at least once while it was executing.
```
tylerwilliams, line 373: nit: can you make sure these fall < 80 chars?

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -288,6 +341,90 @@ func (c *podmanCommandContainer) PullImage(ctx context.Context, creds container.
...
+			dur := time.Since(start)
+			if last == nil && dur > 1*time.Second && lastErr != nil {
```
tylerwilliams, line 375: I'm trying to understand this logic -- we poll every 50ms (so we'd expect any task that took over 50ms and ran in cgroups to have some stats) but we don't log an err unless we've polled for over a second? Is this warning useful / necessary?

tylerwilliams (approved):
Couple nits but overall this is pretty nice. Grabbing this data from cgroupfs is a cool idea :)

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: I wasn't super thrilled with these either. I thought about getting these directly from podman, but podman doesn't expose the cgroup paths via any of the CLI commands as far as I can tell (e.g. `podman container inspect`).

One thing I can think to do would be to run a small container like busybox, and then run the Go-equivalent of a command like

`find /sys/fs/cgroup | grep libpod-${CONTAINER_ID} | grep -P '(/cpu.usage|/memory.usage_in_bytes|/memory.current)'`

Then derive the cgroup paths from those paths. I'm not sure it's worth the complexity so I figured we could hard code for now then maybe do something smarter later on -- wdyt?

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: Done -- I had my editor's tab size at 2. Changed it to 4 to match our repo's `.editorconfig`.

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: This includes enough time to account for the overhead of creating the container, which can be a few hundred ms and possibly a little longer if the executor is heavily loaded. I just set this to 1s so that we shouldn't expect to see false positives.

It's helpful to have these so that we know something is misconfigured, e.g. incorrect cgroup paths.

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: Done

enterprise/server/remote_execution/containers/podman/podman.go
tylerwilliams: sgtm

tylerwilliams (approved):

PR #2139, opened 2022-06-14
Add prometheus metrics for podman container resource usage

bduffany:
:point_right: Depends on https://github.com/buildbuddy-io/buildbuddy/pull/2138 (please review that one first)

Enables charts like this:

image

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -11,6 +11,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/platform"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
+	"github.com/buildbuddy-io/buildbuddy/server/metrics"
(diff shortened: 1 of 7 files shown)


enterprise/server/remote_execution/container/container.go
```diff
@@ -58,6 +63,43 @@ type Stats struct {
...
+	if usageBytes == 0 {
+		delete(m.memoryUsage, c)
+	} else {
+		m.memoryUsage[c] = usageBytes
+	}
+	var total int64
+	for _, v := range m.memoryUsage {
+		total += v
+	}
+	metrics.RemoteExecutionMemoryUsageBytes.Set(float64(total))
```
tylerwilliams, line 100: would using a different metric type allow you to avoid this locking? It seems like kind of a bummer to have to lock when recording stats. (even though it's probably not an issue in this case)

bduffany: The other metric types are not really suitable since the intent is for this to reflect the current memory usage, so that we can compare to the current "estimated" memory usage.

tylerwilliams (approved):

PR #2140, opened 2022-06-14
Bump github.com/prometheus/client_golang to latest version.

vadimberezniker:
According https://github.com/prometheus/client_golang/pull/955, the
newer version uses a more efficient API to get data from the go runtime.

---

**Version bump**: None


tylerwilliams (approved):
sweet!

PR #2141, opened 2022-06-14
add APIs to modify quota config

luluz66:


tylerwilliams:
(gonna look at these after we get schema all nailed down)

proto/quota.proto
```diff
@@ -0,0 +1,123 @@
...
+message Namespace {
+  // The name of the namespace
+  string name = 1;
+
+  // All the buckets in the namespace
+  repeated Bucket buckets = 2;
+
+  // Each group indicates that quota_key will be rate limited by the max_rate
+  // and max_burst defined in the bucket identified by the bucket_name.
+  repeated QuotaGroup groups = 3;
```
tylerwilliams, line 49: maybe it's better to define a new proto like AssignedBucket that contains a QuotaKey and a Bucket and then return a list of those? Then clients don't need to cross-reference

proto/quota.proto
```diff
@@ -0,0 +1,123 @@
...
+
+message AddNamespace {
+  // The name of the namespace
+  string name = 1;
+
+  // The default bucket to be used when there is no explicit quota groups
+  // defined for a user.
+  Bucket default_bucket = 2;
+
+  repeated Bucket buckets = 3;
```
tylerwilliams, line 60: nit: can you add comments specifying if fields or Optional or Required for any Create/Add type calls?

proto/quota.proto
```diff
@@ -0,0 +1,123 @@
...
+
+  // All the buckets in the namespace
+  repeated Bucket buckets = 2;
+
+  // Each group indicates that quota_key will be rate limited by the max_rate
+  // and max_burst defined in the bucket identified by the bucket_name.
+  repeated QuotaGroup groups = 3;
+}
+
+message AddNamespace {
```
tylerwilliams, line 52: not sure why AddNamespace is its own proto if it's only used in CreateNamespaceRequest?

proto/quota.proto
```diff
@@ -0,0 +1,123 @@
...
+message RemoveNamespaceRequest {
+  context.RequestContext request_context = 1;
+  string namespace = 2;
+}
+
+message RemoveNamespaceResponse {
+  context.ResponseContext response_context = 1;
+}
+
+message ModifyBucketRequest {
```
bduffany, line 81: nit: wdyt about calling this RPC `ModifyNamespace`? since from a UI perspective, it's changing the buckets which exist in a namespace

proto/quota.proto
```diff
@@ -0,0 +1,123 @@
...
+
+message ApplyBucketRequest {
+  context.RequestContext request_context = 1;
+
+  QuotaKey key = 2;
+  string namespace = 3;
+
+  // When new_bucket_name is empty, will apply default bucket for the specified
+  // key.
+  string new_bucket_name = 4;
```
bduffany, line 107: nit: maybe just call this bucket_name, since it's inferred from `ApplyBucketRequest` that this is the bucket which should be newly applied

luluz66: Done

tylerwilliams (approved):

proto/quota.proto
```diff
@@ -0,0 +1,107 @@
...
+  // The name of the namespace
+  string name = 1;
+
+  // All the buckets in the namespace
+  repeated AssignedBucket assigned_buckets = 2;
+}
+
+message RemoveNamespaceRequest {
+  context.RequestContext request_context = 1;
+  string namespace = 2;
```
bduffany, line 56: nit

```suggestion
  // The namespace to remove. All buckets that constitute the namespace will be
  // removed as well as any quota groups associated with those buckets.
  string namespace = 2;
```

(feel free to reject or modify if my understanding is incorrect :sweat_smile:)

bduffany (approved):
nice, just nits. I like how minimal the API is now :)

PR #2143, opened 2022-06-16
refactor: migrate to rules_js

thesayyn:
Overview of changes;

- Migrate away from rules_nodejs to rules_js
- Introduce pnpm lockfile

---

**Version bump**: TODO


tylerwilliams:
Hey @thesayyn, @alexeagle, thanks for the PR!!! This is super cool :) Just to set expectations a little bit, and make sure you don't think we're ignoring this PR, we're gonna do some performance and other RBE testing of this vs our existing setup and give it a little time before we decide to merge this. But we're really excited about rules_js and the direction this is going!

(@bduffany can help point you at some docs to make sure rbe is working)

proto/defs.bzl
```diff
@@ -64,9 +40,11 @@ def ts_proto_library(name, proto, **kwargs):
     )
 
     # Transform .proto files to a single _pb.js file named after the macro
-    pbjs(
+    bin.pbjs(
         name = js_target,
-        data = [":" + proto_target] + _PROTOBUFJS_CLI_DEPS,
+        srcs = [":" + proto_target],
+        copy_srcs_to_bin = False,
+        chdir = "../../../",
```
gregmagolan, line 47: could pass

```
env = {
   "BAZEL_BIN": "."
},
```

instead of the `chdir` I believe; tho I think we should add an explicit attribute for this situation such as `run_out_of_execroot = True` or make it an opt-out `run_out_of_bazel_bin = False`

proto/defs.bzl
```diff
@@ -75,15 +53,17 @@ def ts_proto_library(name, proto, **kwargs):
             "--root=%s" % name,
             "--strict-long",  # Force usage of Long type with int64 fields
             "--out=$@",
-            "$(execpaths %s)" % proto_target,
+            "$(locations %s)" % proto_target,
```
gregmagolan, line 56: why locations? perhaps you want `rootpaths`?

tsconfig.json
```diff
@@ -11,18 +11,7 @@
...
     "composite": true,
     "target": "ES2016",
-    "baseUrl": ".",
-    "lib": ["es2017", "es2019", "dom", "dom.iterable"],
-    "rootDirs": [
-      ".",
-      "bazel-bin",
-      "bazel-out/k8-opt/bin",
-      "bazel-out/k8-fastbuild/bin",
-      "bazel-out/darwin-opt/bin",
```
gregmagolan, line 21: 🎉

proto/defs.bzl
thesayyn: Ah right. Thanks!

proto/defs.bzl
thesayyn: It fails for external sources when I use `rootpaths` for a strange reason.

bduffany:
:wave: Thanks for kicking this off!

The CI test failures are a blocker for us, since we want to ensure that our app builds on our RBE service.

To build the app on RBE and reproduce those failures, you can try building the app with `--config=remote`, like `bazel build //app:app_bundle --config=remote` or `bazel build //enterprise/app:app_bundle --config=remote`. If you want to dig deeper into the RBE actions you can check out the `Executions` page in the BB invocation to see metadata about each individual action, including the inputs and outputs to each action.

If you want to dig deeper and debug with a local BB stack, I can also send over some instructions for doing that. Feel free to also reach out over in the Bazel Slack if it's easier!

alexeagle:
rules_js actions are known to not work on RBE due to https://github.com/bazelbuild/bazel/issues/10298
We rely on the symlinks being laid out in the execroot intact.

lberki:
@alexeagle why does rules_js not work? After a quick reading of the discussion thread on #10298 the only thing I see is the requirement for `--experimental_allow_unresolved_symlinks`.

alexeagle:
Right, it's not just unresolved symlinks that are broken - the RBE protocol doesn't send symlinks to the remote worker *at all*. They are all resolved to regular files by the time the remote execroot is laid out, which breaks the pnpm package manager (the nodejs require algorithm must walk a symlink into the virtual package store for transitive resolutions to work)

lberki:
Thanks! I thought that we had fixed that in the past but I apparently remembered it wrong.

alexeagle:
Hey @bduffany (and maybe @brentleyjones will be curious) - rules_js can work with RBE now: https://github.com/aspect-build/rules_js/issues/308#issuecomment-1225130194
Would be good to verify that on buildbuddy (we only tested on Engflow)

juanpmarin:
https://github.com/buildbuddy-io/buildbuddy/blob/f6efe0975cf16f429f0275eee127c8c7120b692f/WORKSPACE#L3
shouldn't this line be removed? As it is necessary by rules_nodejs only

alexeagle:
FYI Bazel 6.0.0-pre.20220922.1 now includes all the symlink fixes.

It seems there may be something broken on Buildbuddy with the symlinks however, based on recent reports on Bazel slack
e.g. https://bazelbuild.slack.com/archives/CULGHFKCK/p1665079241579239

maggie-lou:
* master
  * **PR #2143** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2143" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2143?utm_source=stack-comment).

bduffany:
@thesayyn I've rebased this PR and fixed merge conflicts in https://github.com/buildbuddy-io/buildbuddy/pull/2688

I haven't quite managed to get it building yet; see https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1671033211592849

PR #2146, opened 2022-06-17
Task sizing: Store estimated and measured compute usage in the Executions table

bduffany:
Note: this change is intended to be backwards compatible and not rely on the order in which apps/executors are deployed.

---

**Version bump**: Patch


proto/remote_execution.proto
```diff
@@ -1766,6 +1767,13 @@ message ExecutionTask {
   RequestMetadata request_metadata = 9;
 }
 
+// An ExecutionTask along with its scheduling metadata computed by the execution
+// service.
+message ScheduledTask {
```
vadimberezniker, line 1772: Does this need to be a proto or can it be a type internal to the executor?

server/interfaces/interfaces.go
```diff
@@ -532,6 +532,13 @@ type TaskRouter interface {
 	MarkComplete(ctx context.Context, cmd *repb.Command, remoteInstanceName, executorInstanceID string)
 }
 
+// ScheduledTask represents an execution task along with its scheduling metadata
```
vadimberezniker, line 535: Can this be removed now?

vadimberezniker (approved):

tylerwilliams (commented):
(whoops, thought i'd sent these comments but they were pending)

PR #2147, opened 2022-06-17
Utility to dedupe expensive work across servers.

vadimberezniker:
Every client for a given work key will try to obtain a lock in Redis.
The first client to get the lock will perform the work and publish the
result to a stream. If the owner does not refresh the lock within 5
seconds, another worker will have a chance to claim the lock and do the
work.

---

**Version bump**: None

### DIFF
--- a/server/util/dsingleflight/dsingleflight.go
+++ b/server/util/dsingleflight/dsingleflight.go
@@ -1,0 +1,195 @@
+// Package dsingleflight is a distributed version of the singleflight package.
+// Work is coordinated across servers using Redis.
+package dsingleflight
+
+import (
+	"context"
+	"fmt"
+	"time"
+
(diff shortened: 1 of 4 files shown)


server/util/dsingleflight/dsingleflight.go
```diff
@@ -0,0 +1,195 @@
+// Package dsingleflight is a distributed version of the singleflight package.
+// Work is coordinated across servers using Redis.
+package dsingleflight
+
```
tylerwilliams, line 4: nit: if you wanted to get really fancy you could add a comment with an example of how to use this

tylerwilliams (approved):

server/util/dsingleflight/dsingleflight.go
vadimberezniker: done

PR #2148, opened 2022-06-17
Dedupe image conversions across registry instances.

vadimberezniker:
This will allow us to run multiple replicas of the registry.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/registry/registry.go
+++ b/enterprise/server/registry/registry.go
@@ -17,6 +17,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/redis_client"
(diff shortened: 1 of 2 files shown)


enterprise/server/registry/registry.go
```diff
@@ -274,27 +309,44 @@ func (r *registry) convertImage(ctx context.Context, img v1.Image) (manifestLike
 		Image:           newImg,
 		casDependencies: casDependencies,
 	}
-	if err := r.writeManifest(ctx, img, ci); err != nil {
+	mfProto, err := manifestProto(ci)
+	if err != nil {
+		return nil, err
+	}
+	if err := r.writeManifest(ctx, img, mfProto); err != nil {
```
tylerwilliams, line 316: mfProto

image

tylerwilliams (approved):

PR #2153, opened 2022-06-22
Add missing single quote

tempoz:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -235,7 +235,7 @@ func (r *buildEventReporter) Start(startTime time.Time) error {
 
 	options := []string{}
 	if *besBackend != "" {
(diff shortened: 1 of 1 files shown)


tempoz:
Fixes buildbuddy-io/buildbuddy-internal#1447.

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -235,7 +235,7 @@ func (r *buildEventReporter) Start(startTime time.Time) error {
 
 	options := []string{}
 	if *besBackend != "" {
-		options = append(options, fmt.Sprintf("--bes_backend=%s'", *besBackend))
+		options = append(options, fmt.Sprintf("--bes_backend=%s", *besBackend))
```
bduffany, line 238: I think we want to wrap the `%s` in single-quotes, similar to the `--remote_header` arg below

PR #2154, opened 2022-06-22
Task sizing: Add flag to estimate based on most recent execution stats

bduffany:
When `--remote_execution.use_measured_task_sizes` is set:
* Store task sizing info in Redis
* Estimate task size based on the last recorded size

Note: This flag is not intended to be enabled until we have scheduler logic in place to retry tasks with increased memory allowance when they hit OOM errors; otherwise firecracker actions may get into a permanently bad state. (We could also unblock ourselves here by disabling adaptive task sizing when firecracker isolation type is in use.)

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -53,6 +53,7 @@ import (
 	aclpb "github.com/buildbuddy-io/buildbuddy/proto/acl"
 	espb "github.com/buildbuddy-io/buildbuddy/proto/execution_stats"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	scpb "github.com/buildbuddy-io/buildbuddy/proto/scheduler"
 	uidpb "github.com/buildbuddy-io/buildbuddy/proto/user_id"
 	vfspb "github.com/buildbuddy-io/buildbuddy/proto/vfs"
 	wkpb "github.com/buildbuddy-io/buildbuddy/proto/worker"
@@ -756,13 +757,21 @@ func (p *pool) warmupImage(ctx context.Context, containerType platform.Container
 			{Name: "workload-isolation-type", Value: string(containerType)},
 		},
 	}
-	task := &repb.ExecutionTask{
+	execTask := &repb.ExecutionTask{
 		Command: &repb.Command{
 			Arguments: []string{"echo", "'warmup'"},
 			Platform:  plat,
 		},
 	}
-	platProps := platform.ParseProperties(task)
+	task := &interfaces.ScheduledTask{
+		SchedulingMetadata: &scpb.SchedulingMetadata{
+			// Note: this will use the default task size estimates and not
+			// adaptive task sizing, which requires the app.
+			TaskSize: tasksize.Estimate(execTask),
+		},
+		ExecutionTask: execTask,
+	}
+	platProps := platform.ParseProperties(execTask)
 	c, err := p.newContainer(ctx, platProps, task)
 	if err != nil {
 		log.Errorf("Error warming up %q: %s", containerType, err)
@@ -819,8 +828,9 @@ func (p *pool) Warmup(ctx context.Context) {
 //
 // The returned runner is considered "active" and will be killed if the
 // executor is shut down.
-func (p *pool) Get(ctx context.Context, task *repb.ExecutionTask) (interfaces.Runner, error) {
+func (p *pool) Get(ctx context.Context, st *interfaces.ScheduledTask) (interfaces.Runner, error) {
 	executorProps := platform.GetExecutorProperties()
+	task := st.ExecutionTask
 	props := platform.ParseProperties(task)
 	// TODO: This mutates the task; find a cleaner way to do this.
 	if err := platform.ApplyOverrides(p.env, executorProps, props, task.GetCommand()); err != nil {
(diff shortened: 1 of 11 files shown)


bduffany:
Converted to draft because some executor changes got into this PR unintentionally. Will move those to a separate PR

enterprise/server/tasksize/tasksize.go
```diff
@@ -64,20 +93,141 @@ func Register(env environment.Env) error {
...
+
+func commandKey(cmd *repb.Command) (string, error) {
+	// Include the command executable name in the key for easier debugging.
+	arg0 := "?"
+	if len(cmd.Arguments) > 0 {
+		arg0 = cmd.Arguments[0]
+		if len(arg0) > 64 {
+			arg0 = arg0[:64] + "..."
+		}
+	}
```
tylerwilliams, line 225: if we're hashing these, do we still need to truncate?

enterprise/server/tasksize/tasksize.go
```diff
@@ -64,20 +93,141 @@ func Register(env environment.Env) error {
...
+	// For now, associate stats with the exact command, including the full
+	// command line, env vars, and platform.
+	// Note: This doesn't account for platform overrides for now
+	// (--remote_header=x-buildbuddy-platform.NAME=VALUE).
+	cmdKey, err := commandKey(cmd)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s/%s/%s", redisKeyPrefix, groupKey, cmdKey), nil
```
tylerwilliams, line 214: So this will look like:

"taskSize/GR1231231241412//usr/bin/gccSHA12312312314" ?

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -756,13 +757,21 @@ func (p *pool) warmupImage(ctx context.Context, containerType platform.Container
...
-	platProps := platform.ParseProperties(task)
+	task := &interfaces.ScheduledTask{
+		SchedulingMetadata: &scpb.SchedulingMetadata{
+			// Note: this will use the default task size estimates and not
+			// adaptive task sizing, which requires the app.
+			TaskSize: tasksize.Estimate(execTask),
+		},
+		ExecutionTask: execTask,
+	}
+	platProps := platform.ParseProperties(execTask)
```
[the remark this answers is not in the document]
bduffany: (moved this to a separate PR -- will address there)

enterprise/server/tasksize/tasksize.go
```diff
@@ -52,6 +63,17 @@ const (
 
 	// The fraction of an executor's allocatable resources to make available for task sizing.
 	MaxResourceCapacityRatio = 0.8
+
+	// The expiration for all task-sizing keys stored in Redis.
+	redisKeyExpiration = 5 * 24 * time.Hour
+	// Redis key prefix used for holding current task size estimates.
```
tylerwilliams, line 69: newline before //

enterprise/server/tasksize/tasksize.go
```diff
@@ -64,20 +86,140 @@ func Register(env environment.Env) error {
...
+	if *useMeasuredSizes {
+		rdb = env.GetRemoteExecutionRedisClient()
+		if rdb == nil {
+			return nil, status.FailedPreconditionError("missing Redis client configuration")
+		}
+	}
+	return &taskSizer{
+		env: env,
+		rdb: rdb,
+	}, nil
```
tylerwilliams, line 105: idk if this is better but you could skip the rdb var if you just assign directly into your taskSizer struct

	ts := &taskSizer{
		env: env,
	}
	if *useMeasuredSizes {
		if env.GetRemoteExecutionRedisClient() == nil {
			return nil, status.FailedPreconditionError("missing Redis client configuration")
		}
        ts.rdb = env.GetRemoteExecutionRedisClient()
	}
	return ts, nil

enterprise/server/tasksize/tasksize.go
```diff
@@ -64,20 +86,140 @@ func Register(env environment.Env) error {
...
+	if !*useMeasuredSizes {
+		return defaultSize
+	}
+	recordedSize, err := s.lastRecordedSize(ctx, task)
+	if err != nil {
+		log.CtxWarningf(ctx, "Failed to read task size from Redis; falling back to default size estimate: %s", err)
+		return defaultSize
+	}
+	if recordedSize == nil {
+		return defaultSize
```
tylerwilliams, line 119: add a //TODO that this is where we'd eventually return unsized() and let the executor run this task once to determine the size?

enterprise/server/tasksize/tasksize.go
```diff
@@ -52,6 +63,17 @@ const (
 
 	// The fraction of an executor's allocatable resources to make available for task sizing.
 	MaxResourceCapacityRatio = 0.8
+
+	// The expiration for all task-sizing keys stored in Redis.
+	redisKeyExpiration = 5 * 24 * time.Hour
```
tylerwilliams, line 68: nit: sizeEstimateExpiry or sizeEstimateExpiration?

vadimberezniker (approved):

enterprise/server/tasksize/tasksize.go
tylerwilliams: k, just want to make sure these don't get too big (keys take space in redis too, i have seen them get... wildly out of hand haha)

tylerwilliams (approved):

enterprise/server/tasksize/tasksize.go
bduffany: Done

PR #2156, opened 2022-06-22
Allow pebble to migrate from an existing disk cache

tylerwilliams:


server/backends/disk_cache/disk_cache.go
```diff
@@ -597,49 +588,158 @@ func (fk *fileKey) FromPartitionAndPath(part *partition, fullPath string) error
...
 
 	if len(parts) > 0 {
-		fk.remoteInstanceName = fk.part.internString(strings.Join(parts, "/"))
+		remoteInstanceName = strings.Join(parts, "/")
 	}
+	return
+}
+
+func ScanDiskDirectory(scanDir string) <-chan *rfpb.FileMetadata {
+	scanned := make(chan *rfpb.FileMetadata)
```
vadimberezniker, line 643: I think you want this to be a bufferred channel for better throughput.

vadimberezniker (approved):

server/backends/disk_cache/disk_cache.go
tylerwilliams: Done, thanks!

PR #2157, opened 2022-06-22
Task sizing: Use size computed by scheduler when creating runners

bduffany:
* Use the scheduler as the source of truth, rather than using the default estimate, when running scheduled tasks.
* When running warmup tasks, just use the default estimates.

---

**Version bump**: Patch


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -969,7 +979,7 @@ func (p *pool) newContainer(ctx context.Context, props *platform.Properties, tas
 		}
 		ctr = p.podmanProvider.NewContainer(props.ContainerImage, opts)
 	case platform.FirecrackerContainerType:
-		sizeEstimate := tasksize.Estimate(task)
+		sizeEstimate := task.SchedulingMetadata.GetTaskSize()
```
tylerwilliams, line 982: @bduffany @vadimberezniker

nit: I personally would make SchedulingMetadata a proto for consistency. I know it's not really going over the wire / to disk yet, but it's made up of two protos.

My reasoning is:
 - it would let you remove a struct from interfaces (which is nice)
 - it would allow for safer more consistent treatment of nil vals, ie, you could do:
     ```task.GetSchedulingMetadata().GetTaskSize()``` and if either scheduling_metadata or task_size were nil you'd be fine and get a nil back (instead of a panic like you'd get today).
 - it's basically just a tuple/pair<> of your two protos

there was some rule of thumb about this I vaguely remember from when working on android, and all I can remember was POJO vs PROTO, but I think the gist of it was that these things tend to either grow class logic (in which case you want a class) or grow serialization logic (in which case you want a proto). This feels closer to the latter to me.

tylerwilliams (changes requested):

enterprise/server/remote_execution/runner/runner.go
bduffany: sg, done

tylerwilliams (approved):

PR #2158, opened 2022-06-22
Move ExecutionSummary properties into the ExecutedActionMetadata

tempoz:
---

**Version bump**: Minor

### DIFF
--- a/app/invocation/invocation_execution_card.tsx
+++ b/app/invocation/invocation_execution_card.tsx
@@ -181,7 +181,7 @@ export default class ExecutionCardComponent extends React.Component<Props, State
         );
       case "files-uploaded":
         return (
-          +first?.executedActionMetadata.ioStats?.fileUploadCount -
+          +first?.executedActionMetadata?.ioStats?.fileUploadCount -
           +second?.executedActionMetadata?.ioStats?.fileUploadCount
         );
       case "file-size-downloaded":
--- a/app/invocation/invocation_execution_card.tsx
+++ b/app/invocation/invocation_execution_card.tsx
@@ -175,13 +175,25 @@ export default class ExecutionCardComponent extends React.Component<Props, State
       case "upload-duration":
         return this.uploadDuration(first) - this.uploadDuration(second);
       case "files-downloaded":
-        return +first?.ioStats?.fileDownloadCount - +second?.ioStats?.fileDownloadCount;
+        return (
+          +first?.executedActionMetadata?.ioStats?.fileDownloadCount -
+          +second?.executedActionMetadata?.ioStats?.fileDownloadCount
+        );
       case "files-uploaded":
-        return +first?.ioStats?.fileUploadCount - +second?.ioStats?.fileUploadCount;
+        return (
+          +first?.executedActionMetadata.ioStats?.fileUploadCount -
+          +second?.executedActionMetadata?.ioStats?.fileUploadCount
+        );
       case "file-size-downloaded":
-        return +first?.ioStats?.fileDownloadSizeBytes - +second?.ioStats?.fileDownloadSizeBytes;
+        return (
+          +first?.executedActionMetadata?.ioStats?.fileDownloadSizeBytes -
+          +second?.executedActionMetadata?.ioStats?.fileDownloadSizeBytes
+        );
       case "file-size-uploaded":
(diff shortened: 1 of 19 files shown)


tempoz:
* master
  * **PR #2158** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2158" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2158?utm_source=stack-comment).

tylerwilliams:
@bduffany for typescript

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -53,30 +52,30 @@ var (
 	enableActionMerging               = flag.Bool("remote_execution.enable_action_merging", true, "If enabled, identical actions being executed concurrently are merged into a single execution.")
 )
 
-func fillExecutionFromSummary(summary *espb.ExecutionSummary, execution *tables.Execution) {
+func fillExecutionFromActionMetadata(md *repb.ExecutedActionMetadata, execution *tables.Execution) {
```
bduffany, line 55: we might want to keep around the old method for backwards compatibility, in case customers are running old executor versions that are still setting ExecutionSummary. These are just used for timing data in the UI though, currently, so I don't feel super strongly about it. @tylerwilliams wdyt?

proto/remote_execution.proto
```diff
@@ -860,6 +861,35 @@ message Digest {
...
+}
+
+// Compute usage sampled throughout a task's execution.
+message UsageStats {
+  // Maximum amount of memory used throughout the task's execution.
+  int64 peak_memory_bytes = 1;
+
+  // Total number of CPU-nanoseconds consumed by the task.
+  int64 cpu_nanos = 2;
+}
```
bduffany, line 891: nit: could you move these to the bottom of the file? It's not super well-delineated currently but some of the protos in this file were added by us so we'd probably want to keep those separate. I think everything from `ExecutionTask` and below was added by us, if you wouldn't mind also adding a delineating comment while you're there :)

app/invocation/invocation_execution_card.tsx
```diff
@@ -175,13 +175,25 @@ export default class ExecutionCardComponent extends React.Component<Props, State
...
       case "files-downloaded":
-        return +first?.ioStats?.fileDownloadCount - +second?.ioStats?.fileDownloadCount;
+        return (
+          +first?.executedActionMetadata?.ioStats?.fileDownloadCount -
+          +second?.executedActionMetadata?.ioStats?.fileDownloadCount
+        );
       case "files-uploaded":
-        return +first?.ioStats?.fileUploadCount - +second?.ioStats?.fileUploadCount;
+        return (
+          +first?.executedActionMetadata.ioStats?.fileUploadCount -
```
bduffany, line 184: probably need a `?` after executedActionMetadata here

bduffany (commented):
mostly LG, will let Tyler approve since he might have more thoughts on some of the proto changes

app/invocation/invocation_execution_card.tsx
tempoz:
```suggestion
          +first?.executedActionMetadata?.ioStats?.fileUploadCount -
```

proto/remote_execution.proto
```diff
@@ -1753,6 +1763,12 @@ message TreeToken {
   repeated SizedDirectory sized_directories = 1;
 }
 
+/******************************************************************************/
+/*                                                                            */
+/*                BEGIN BUILDBUDDY-SPECIFIC PROTO DECLARATIONS                */
+/*                                                                            */
+/******************************************************************************/
```
tempoz, line 1770: wdyt of this style of comment? it seemed to me like if we're trying to delineate blocks, it should be very visible?

enterprise/server/remote_execution/execution_server/execution_server.go
tylerwilliams: +1 to making this backwards compatible even though we expect it to be short lived

tempoz: Done!

tylerwilliams (approved):

bduffany (approved):

PR #2160, opened 2022-06-23
Don't return corrupt tree cache content

tylerwilliams:


server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -682,12 +682,17 @@ func (s *ContentAddressableStorageServer) GetTree(req *repb.GetTreeRequest, stre
...
-	for _, child := range treeCache.Children {
+func isComplete(children []*repb.DirectoryWithDigest) bool {
+	allDigests := make(map[string]struct{}, len(children))
+	for _, child := range children {
 		allDigests[child.GetDigest().GetHash()] = struct{}{}
 	}
-	for _, child := range treeCache.Children {
+	for _, child := range children {
+		dir := child.Directory
+		if len(dir.Directories) == 0 && len(dir.Files) == 0 && len(dir.Symlinks) == 0 {
```
siggisim, line 692: any chance an actual zero digest directory could make it here? do we need to check that we're not dealing with the empty hash or something?

siggisim (approved):

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
tylerwilliams: I don't think empty directories should come through here (we check against empty hash and just skip them on ~483). Since this is the same check we're doing when reading dirs anyway, now just applied to cache content, it should be safe.

PR #2161, opened 2022-06-23
Allow faster refresh rate for local Grafana development

bduffany:
Reduce the min refresh rate from 5s -> 1s for local development only. This helps visualize metrics which change very quickly, such as executor memory and CPU.

---

**Version bump**: Patch

### DIFF
--- a/tools/metrics/grafana/dashboards/buildbuddy.json
+++ b/tools/metrics/grafana/dashboards/buildbuddy.json
@@ -8517,6 +8517,7 @@
   },
   "timepicker": {
     "refresh_intervals": [
+      "1s",
       "5s",
       "10s",
       "15s",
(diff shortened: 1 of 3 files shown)


tools/metrics/grafana/dashboards/buildbuddy.json
```diff
@@ -8517,6 +8517,7 @@
   },
   "timepicker": {
     "refresh_intervals": [
+      "1s",
```
vadimberezniker, line 8520: Does the refresh interval have any effect on the granularity of the queried data? I thought this only affected how frequently Grafana re-queried data. The datasources.yml seems like the only setting that affects it. 

The change seems fine if it does what you say it does, just wanted to double check.

vadimberezniker (approved):

tools/metrics/grafana/dashboards/buildbuddy.json
bduffany: The missing piece of the puzzle here is that we already had a prometheus scrape_interval of 1s configured in `tools/metrics/prometheus/prometheus.yml` (not shown in this PR because there are no changes to it). This PR is just telling Grafana that it is allowed to query Prometheus at that same rate.

bduffany: The `GF_DASHBOARDS_MIN_REFRESH_INTERVAL=1s` is needed because by default the min refresh interval is 5s. I think the intent of this 5s default is that it's a safeguard against folks setting a refresh interval too high (whether via URL param or dashboard preset) and then overloading Prometheus with queries (which would just keep returning the same data if the scrape_interval is lower than the refresh interval anyway)

PR #2164, opened 2022-06-23
Integrate registry serving into app.

vadimberezniker:
This will allow us to start the service as part of the app.

Note that the registry runs on its own http/grpc ports to allow us to
treat them as internal services.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/cmd/server/BUILD
+++ b/enterprise/server/cmd/server/BUILD
@@ -28,6 +28,7 @@ go_library(
         "//enterprise/server/invocation_stat_service",
         "//enterprise/server/quota",
         "//enterprise/server/raft/cache",
+        "//enterprise/server/registry:registry_lib",
         "//enterprise/server/remote_execution/execution_server",
         "//enterprise/server/remote_execution/redis_client",
         "//enterprise/server/saml",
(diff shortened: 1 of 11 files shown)


enterprise/server/registry/registry_service/registry_service.go
```diff
@@ -0,0 +1,980 @@
...
+	regRegistryServices := func(server *grpc.Server, env environment.Env) {
+		regpb.RegisterRegistryServer(server, r)
+	}
+
+	if err := grpc_server.RegisterGRPCServer(env, *gRPCPort, regRegistryServices); err != nil {
+		return status.UnknownErrorf("Could not setup GRPC server: %s", err)
+	}
+	if err := grpc_server.RegisterGRPCSServer(env, *gRPCSPort, regRegistryServices); err != nil {
+		return status.UnknownErrorf("Could not setup GRPCS server: %s", err)
+	}
```
tylerwilliams, line 933: Can we run this on 5151 (the existing internal grpc service used by ddisk?) and then add a new port for internal HTTP traffic?

tylerwilliams: https://github.com/buildbuddy-io/buildbuddy-internal/blob/master/enterprise/deployment/dev/us-west1-b/buildbuddy-app.dev.yaml#L13

vadimberezniker: I've added new "standard" ports for internal gRPC & http services.
Migrating distributed cache over to the standard port is more involved. For one, we still need to support the distributed_cache.listen_addr flag because onprem customers may already be using it and it's not a just a simple port flag. We could additionally serve the distributed cache RPCs on the new ports and migrate over, but I don't see a simple way of getting rid of the separate serving code for existing users.

tylerwilliams (approved):

PR #2167, opened 2022-06-23
Ensure dirs exist before using them

tylerwilliams:


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -103,9 +103,8 @@ func Register(env environment.Env) error {
 	if *rootDirectory == "" {
 		return nil
 	}
-	if env.GetCache() != nil {
-		log.Warning("A cache has already been registered, skipping registering pebble_cache.")
```
vadimberezniker, line 107: Was this removal intentional?

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: yup!

PR #2170, opened 2022-06-24
Task sizing: Add a platform prop to disable measurement-based sizing

bduffany:
This will make it easier to test and debug measurement-based sizing without needing to redeploy the app each time. Planning to remove it later once it is no longer needed.

---

**Version bump**: Patch


vadimberezniker (approved):

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -142,6 +143,11 @@ type Properties struct {
 	WorkflowID               string
 	HostedBazelAffinityKey   string
 	UseSelfHostedExecutors   bool
+	// DisableMeasuredTaskSize disables measurement-based task sizing, even if
+	// it is enabled via flag, and instead uses the default / platform based
+	// sizing. Intended for debugging purposes only and should not generally
+	// be used.
+	DisableMeasuredTaskSize bool
```
tylerwilliams, line 150: nit: add a TODO to remove it

tylerwilliams (approved):

enterprise/server/remote_execution/platform/platform.go
bduffany: Done

PR #2172, opened 2022-06-24
Be explicity about cache mutual exclusion / wrapping dependencies

tylerwilliams:
Some caches cannot co-exist with another, so bail if multiple of (mem ^ disk ^ pebble ^ s3 ^ gcs) are set.
Other caches require a base cache, so bail if one is not set: (redis, memcache, distributed).
Stop enabling the disk_cache by default.


enterprise/server/backends/distributed/distributed.go
```diff
@@ -97,6 +97,9 @@ func Register(env environment.Env) error {
 	if *listenAddr == "" {
 		return nil
 	}
+	if env.GetCache() == nil {
+		return status.FailedPreconditionErrorf("Distributed Cache requires a base cache; but one was not configured; please also enable a disk/pebble/memory cache")
```
tempoz, line 101: Can we make all of these consistent?
https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/backends/redis_cache/redis_cache.go#L58
https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/backends/memcache/memcache.go#L53

tempoz (changes requested):

tempoz:
Otherwise, LGTM

tempoz (approved):
LGTM

PR #2182, opened 2022-06-25
Extend Cache API to expose blob size.

vadimberezniker:
---

**Version bump**: None

### DIFF
--- a/proto/raft.proto
+++ b/proto/raft.proto
@@ -32,7 +32,6 @@ message Isolation {
 message FileRecord {
   Isolation isolation = 1;
   build.bazel.remote.execution.v2.Digest digest = 2;
-  int64 size_bytes = 3;
 }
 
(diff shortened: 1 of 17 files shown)


proto/raft.proto
```diff
@@ -32,6 +32,7 @@ message Isolation {
 message FileRecord {
   Isolation isolation = 1;
   build.bazel.remote.execution.v2.Digest digest = 2;
+  int64 size_bytes = 3;
```
tylerwilliams, line 35: Could we put this in FileMetadata instead?

(otherwise it's a little confusing whether or not to look at digest.GetSizeBytes() or fileRecord.GetSizeBytes())

vadimberezniker: Done.

tylerwilliams (approved):

PR #2185, opened 2022-06-27
Task sizing: Run tasks unsized if measurements haven't yet been recorded

bduffany:
When we haven't yet recorded a task size, run it with generous memory limits (up to the executor resource limits so that we can always give it a trial run on one of the registered executors).

Note, this is all still behind the flag `remote_execution.use_measured_task_sizes`.

Screenshot showing a task that consumes 4GB of memory first running in "unsized" mode, then running with a more accurate estimate once we know the size (ignore the spike in the second run -- a few extra test actions crept in):

image

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -368,6 +368,33 @@ func (en *executionNode) GetExecutorID() string {
 	return en.executorID
 }
 
+// FillTaskSizeLimits returns the request with the task size adjusted to account
+// for unsized tasks if applicable. Unsized tasks are granted a generous
+// resource limit so that we can take an initial measurement and see how many
+// resources are actually used, but capped to the resource limits of this
+// execution node.
+func (en *executionNode) FillTaskSizeLimits(req *scpb.EnqueueTaskReservationRequest) *scpb.EnqueueTaskReservationRequest {
+	if req.TaskSize == nil || !tasksize.IsUnsized(req.TaskSize) {
+		return req
+	}
+	req = proto.Clone(req).(*scpb.EnqueueTaskReservationRequest)
+
+	// TODO(bduffany): Incorporate knowledge of OOM-retry count here. If an
+	// unsized task hit an OOM error, we should request even more memory than
+	// our default limits on the second attempt.
+	ts := req.TaskSize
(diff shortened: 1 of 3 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -500,6 +527,12 @@ func (np *nodePool) NodeCount(ctx context.Context, taskSize *scpb.TaskSize) (int
 	if len(np.nodes) == 0 {
 		return 0, status.UnavailableErrorf("No registered executors in pool %q with os %q with arch %q.", np.key.pool, np.key.os, np.key.arch)
 	}
+	// If the task is unsized, any node can execute the task, since we cap
+	// unsized task limits so that they can always fit on any registered
+	// executor.
```
tylerwilliams, line 532: this may not always be the case right? imagine if we had a bunch of executors of size 1 and size 10, we might want to run it at size 10 and see how much it takes, not assume it's max size of 1?

enterprise/server/tasksize/tasksize.go
```diff
@@ -38,6 +38,12 @@ const (
 	DefaultCPUEstimate      = int64(600)
 	DefaultFreeDiskEstimate = int64(100 * 1e6) // 100 MB
 
+	// Unsized task estimates (when using measured task sizes and task sizes
+	// have not yet been measured, these estimates are used).
+
```
tylerwilliams, line 43: nit: rm newline

enterprise/server/tasksize/tasksize.go
```diff
@@ -230,6 +238,19 @@ func commandKey(cmd *repb.Command) (string, error) {
 	return fmt.Sprintf("%s/%x", arg0, sha256.Sum256(b)), nil
 }
 
+// Unsized returns a TaskSize indicating that the size is not yet known.
+func Unsized() *scpb.TaskSize {
```
tylerwilliams, line 242: i'm a little confused by this Unsized being -1, -1, but the above consts UnsizedTask{MemoryBytes,MilliCPU} being 5600 / 4000. What is the difference

bduffany:
We resolved offline to keep the current behavior of running tasks with the default size, since (1) we decided to enforce hard-limits as part of a later effort, and (2) to avoid worrying about firecracker (which does have hard limits today) we'll instead disable the measurement-based sizing on firecracker tasks for now.

PR #2192, opened 2022-06-28
Put prometheus time histograms behind a flag (default enabled)

tylerwilliams:


server/util/grpc_server/grpc_server.go
```diff
@@ -97,7 +99,9 @@ func NewGRPCServer(env environment.Env, port int, credentialOption grpc.ServerOp
 
 	// DISABLED in prod: enabling these causes unnecessary allocations
```
vadimberezniker, line 100: remove/update this comment?

vadimberezniker (approved):

server/util/grpc_server/grpc_server.go
tylerwilliams: Done.

PR #2183, opened 2022-06-27
Add YAML autodocs to website documentation

tempoz:
Have to create `.mdx` versions of the YAML files so we can transclude them in the markdown.

---

**Version bump**: Patch

PR #2194, opened 2022-06-28
Build website with bazel in gitub workflows

tempoz:
The target didn't build with `replaceAll`, so I changed instances of that function to `replace` and a regex with the global flag set.

---

**Version bump**: Patch

### DIFF
--- a/.github/workflows/website.yaml
+++ b/.github/workflows/website.yaml
@@ -18,7 +18,11 @@ jobs:
         uses: actions/checkout@v1
 
       - name: Build Website
-        run: bazel build //website:website
+        run: |
+          bazel build //website:website
+          mkdir website/build
+          cd website/build
+          tar -xf ../../bazel-bin/website/website.tar
 
       - name: Deploy Website 🚀
         uses: JamesIves/github-pages-deploy-action@4.1.0
--- a/.github/workflows/website.yaml
+++ b/.github/workflows/website.yaml
@@ -18,9 +18,7 @@ jobs:
         uses: actions/checkout@v1
 
       - name: Build Website
-        run: |
-          cd website
-          yarn install && yarn build
+        run: bazel build //website:website
 
       - name: Deploy Website 🚀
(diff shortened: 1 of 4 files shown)


tempoz:
* master
  * **PR #2194** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2194" target="_blank">Graphite</a>  👈
    * **PR #2183** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2183" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2194?utm_source=stack-comment).

.github/workflows/website.yaml
```diff
@@ -18,9 +18,7 @@ jobs:
         uses: actions/checkout@v1
 
       - name: Build Website
-        run: |
-          cd website
-          yarn install && yarn build
+        run: bazel build //website:website
 
       - name: Deploy Website 🚀
         uses: JamesIves/github-pages-deploy-action@4.1.0
```
siggisim, line 24: Is this deploy action going to work here? I feel like it's probably not going to look in the right spot for the website to deploy.

tempoz: Mmm, I see. I'm not really familiar with the github actions or how yarn does things.

I'll figure out where it needs to run to deploy.

bduffany: I think we might just need to run the bazel build before the yarn build, so that `bazel-bin` gets created. Then the `yarn build` will be able to resolve the paths from `bazel-bin` properly.

tempoz: I changed it to unpack the tar that bazel create in `website/build`, which should fix everything, I think.

siggisim (approved):
LGTM if the outputs for the tar roughly match what we were previously deploying

PR #2199, opened 2022-06-29
Task sizing: Implement usage stats for bare execution

bduffany:
When `--executor.bare.enable_stats` is set, populate `UsageStats` in `CommandResult` returned by `Run()` and `Exec()`.

We still return zeros for `Stats()` since there are no processes actually running when we call Stats() just before attempting to add a bare runner to the pool, except for Mac workflows (bazel server process is kept running in the BG) + bare persistent workers (worker process is kept running in the BG). Handling those will introduce some complexity, so I'm saving those for a later date.

The approach taken here is to poll CPU & mem usage using platform-specific APIs, handled via the `gopsutil` library.

Tested on both Linux and Mac (Apple Silicon).

---

**Version bump**: Patch


enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -136,25 +157,65 @@ func Run(ctx context.Context, command *repb.Command, workDir string, opts *conta
...
+func collectStats(pid int, done <-chan struct{}) *container.Stats {
+	ts := NewTreeStats(pid)
+	// Most processes are short-lived so we need a fast poll rate if we want
+	// to increase the probability of getting at least one sample. But
+	// polling is expensive: a 50ms poll rate consumes around 10% of a CPU
+	// on Linux (and it's even worse on macOS). So we take a hybrid approach
+	// here: start off polling fast but slow down over time.
+	pollInterval := 25 * time.Millisecond
+	const pollBackoff = 1.25
+	const maxPollInterval = 1000 * time.Millisecond
```
bduffany, line 206: @tylerwilliams I couldn't beat the performance of `gopsutil` even using the raw `procfs` (looking at `/proc/[pid]/stat`), so I came up with this compromise for reducing CPU overhead. LMK what you think

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -28,6 +29,14 @@ const (
 	// never started, or its actual exit code could not be determined because of an
 	// error.
 	NoExitCode = -2
+
+	// Parameters for stats polling. The poll interval starts out at
+	// statsInitialPollInterval and is multiplied by statsPollBackoff after each
+	// attempt, up to statsMaxPollInterval.
+
```
tylerwilliams, line 36: nit: rm newline

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -136,25 +165,63 @@ func Run(ctx context.Context, command *repb.Command, workDir string, opts *conta
...
+		return nil, err
 	}
 	pid := cmd.Process.Pid
-	done := make(chan struct{})
-	defer close(done)
+	processTerminated := make(chan struct{})
+	wait := func() error {
+		defer close(processTerminated)
+		return cmd.Wait()
+	}
```
tylerwilliams, line 177: could you move this down to where you use it?

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -215,6 +282,126 @@ func ChildPids(pid int) ([]int, error) {
 	return out, nil
 }
 
+// TreeStats records stats across the lifecycle of a process tree, accounting
```
tylerwilliams, line 285: it feels like there is enough code here that maybe it warrants moving to its own package. wdyt?

Something like util/stats?

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -98,33 +110,50 @@ func RetryIfTextFileBusy(fn func() error) error {
...
 	}
 }
 
+type RunWithCleanupOpts struct {
+	// EnableStats specifies whether to monitor the command while it is running
+	// and collect approximate usage stats. This is not turned on by default
+	// because it incurs some overhead.
+	EnableStats bool
+}
+
```
tylerwilliams, line 156: could we just make this a bool to RunWithProcessTreeCleanup for now?

tylerwilliams (changes requested):

enterprise/server/remote_execution/commandutil/commandutil.go
bduffany: done

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -98,11 +102,15 @@ func RetryIfTextFileBusy(fn func() error) error {
...
 }
 
-// Run a command, retrying "text file busy" errors and killing the process group
+// Run a command, retrying "text file busy" errors and killing the process tree
 // when the context is cancelled.
-func Run(ctx context.Context, command *repb.Command, workDir string, stdio *container.Stdio) *interfaces.CommandResult {
+//
+// Note: enableStats incurs some overhead throughout process execution, so only
+// use it if stats are needed.
+func Run(ctx context.Context, command *repb.Command, workDir string, enableStats bool, stdio *container.Stdio) *interfaces.CommandResult {
```
bduffany, line 110: @tylerwilliams I wound up not creating the "simple" version of this func because there were only 1 or 2 callsites (excluding `commandutil_test.go`) that would've been able to take advantage of it. I figured we can revisit if there are more callsites and the need for a simpler API becomes more clear.

tylerwilliams (approved):

PR #2202, opened 2022-06-29
Ingest in parallel

tylerwilliams:


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -282,45 +282,55 @@ func (p *PebbleCache) processSizeUpdates(quitChan chan struct{}) {
...
+				inserted += 1
+
+				if inserted%10000 == 0 {
+					if err := batch.Commit(&pebble.WriteOptions{Sync: true}); err != nil {
+						return err
+					}
+					batch = p.db.NewBatch()
+				}
+				if inserted%1e6 == 0 {
+					log.Printf("Pebble Cache: migration progress [%d files in %s]...", inserted, time.Since(start))
```
vadimberezniker, line 315: can you add the goroutine # to the log? otherwise the logs are going to be very confusing.

vadimberezniker: and the log below

PR #2208, opened 2022-06-30
Generate prometheus metrics docs at build time

bduffany:
Removes the need to manually run `server/metrics/generate_docs.py` after updating `metrics.go`.

---

**Version bump**: Patch


siggisim (approved):
So cool!

PR #2218, opened 2022-07-05
Executor: Force-remove workspace directories

bduffany:
If we fail to delete a workspace directory, fall back to force-removal (golang equivalent of `chmod -R +w workspace && rm -rf workspace`.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/workspace/workspace.go
+++ b/enterprise/server/remote_execution/workspace/workspace.go
@@ -285,7 +285,16 @@ func (ws *Workspace) Remove() error {
 	// immediately fail since we've set the removing bit.
 	ws.mu.Unlock()
 
-	return os.RemoveAll(ws.rootDir)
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/workspace/workspace.go
```diff
@@ -354,6 +361,21 @@ func (ws *Workspace) Clean() error {
...
 }
 
+// forceRemove attempts to remove a directory by first making the entire tree
+// writable.
+func forceRemove(path string) error {
+	err := filepath.WalkDir(path, func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		return os.Chmod(path, 0777)
```
tylerwilliams, line 371: could we make this chmod +w?

tylerwilliams (approved):

PR #2223, opened 2022-07-05
inline

tylerwilliams:
Support file inlining when the file contents are < 1204 bytes.

Inlining stores small files directly in the pebble metadata as a []byte rather than on-disk in a file. The motivation is basically that for small files we incur a large overhead by storing a filename in the metadata and writing a file.

For the GetMulti benchmark, this increases throughput by 87%.

Before:
    1225	    960599 ns/op	  340951 B/op	    3042 allocs/op
After:
    2292	    503721 ns/op	  308332 B/op	    2026 allocs/op


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1124,8 +1132,18 @@ func (e *partitionEvictor) deleteFile(sample *evictionPoolEntry) error {
...
 	}
 	e.atimes.Delete(xxhash.Sum64(sample.fileMetadataKey))
-	if err := disk.DeleteFile(context.TODO(), sample.filePath); err != nil {
-		return err
+
+	md := sample.fileMetadata.GetStorageMetadata()
+	switch {
+	case md.GetFileMetadata() != nil:
+		fp := filestore.FilePath(e.blobDir, md.GetFileMetadata())
+		if err := disk.DeleteFile(context.TODO(), fp); err != nil {
```
siggisim, line 1140: You want to use background context? Or leave as TODO?

siggisim (approved):

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: same same (i copied this from disk_cache.go)

PR #2228, opened 2022-07-06
Fix startup scan

tylerwilliams:
Previously we were skipping some records when scanning the AC ranges on startup.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -972,11 +972,11 @@ func (e *partitionEvictor) computeSize() (int64, int64, int64, error) {
 	}
 
 	acPrefixRange := func(start, end []byte) ([]byte, []byte) {
-		left := make([]byte, 0)
+		left := make([]byte, 0, len(e.acPrefix)+len(start))
 		left = append(left, e.acPrefix...)
 		left = append(left, start...)
 
-		right := make([]byte, 0)
+		right := make([]byte, 0, len(e.acPrefix)+len(end))
 		right = append(right, e.acPrefix...)
 		right = append(right, end...)
 		return left, right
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -971,13 +971,25 @@ func (e *partitionEvictor) computeSize() (int64, int64, int64, error) {
...
 
+	acPrefixRange := func(start, end []byte) ([]byte, []byte) {
+		left := make([]byte, 0)
+		left = append(left, e.acPrefix...)
+		left = append(left, start...)
+
+		right := make([]byte, 0)
+		right = append(right, e.acPrefix...)
+		right = append(right, end...)
+		return left, right
```
bduffany, line 982: nit: do you want to pre-compute the capacity for left and right?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -971,13 +971,25 @@ func (e *partitionEvictor) computeSize() (int64, int64, int64, error) {
...
+	}
+
 	// Start scanning the AC.
 	// AC keys look like /partitionID/ac/12312312313(crc-32)/digesthash
-	for i := 0; i < 100; i++ {
-		kr := append(e.acPrefix, []byte(fmt.Sprintf("%.2d", i))...)
-		start, end := keyRange(kr)
-		goScanRange(start, end)
+	for i := 10; i < 99; i++ {
```
bduffany, line 987: uber nit: maybe add a comment why it starts @ 10 here, and why 99 is handled as a special case

vadimberezniker (approved):

bduffany (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -919,7 +919,7 @@ func (e *partitionEvictor) computeSizeInRange(start, end []byte) (int64, int64,
 		LowerBound: start,
 		UpperBound: end,
 	})
-	iter.SeekGE(start)
+	iter.SeekLT(start)
```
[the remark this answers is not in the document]
tylerwilliams: Still looking where I am going wrong, but when we use LT here it does not hit everything.

enterprise/server/backends/pebble_cache/pebble_cache.go

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: Done.

PR #2235, opened 2022-07-06
Only log once when dropping progress events in BES stream

maggie-lou:


maggie-lou:
Is it fine that we're no longer logging each event that we're dropping? Checked the logs and it's still firing thousands of times for a single invocation

tylerwilliams:
Ideally we'd not drop these events! But this at least looks good as a fix for the logs spew.

tylerwilliams (approved):

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -782,7 +783,10 @@ func (e *EventChannel) handleEvent(event *pepb.PublishBuildToolEventStreamReques
 	} else if !e.hasReceivedEventWithOptions {
 		e.eventsBeforeOptions = append(e.eventsBeforeOptions, invocationEvent)
 		if len(e.eventsBeforeOptions) > 100 {
-			log.Warningf("We got over 100 build events before an event with options for invocation %s, dropping %+v", iid, e.eventsBeforeOptions[0])
+			if !e.hasDroppedEventsBeforeOptions {
+				e.hasDroppedEventsBeforeOptions = true
+				log.Warningf("We got over 100 build events before an event with options for invocation %s. Beginning to drop events. Earliest dropped event: %+v", iid, e.eventsBeforeOptions[0])
+			}
```
tylerwilliams, line 789: thinking about this a little more -- is there a way to log the index of the last event we had to drop?

It'd be nice to know basically, how big of a buffer we'd need to keep in memory to not drop anything.

maggie-lou: Done

PR #2236, opened 2022-07-07
Increase bucket granularity and max value for duration metrics

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -1,6 +1,8 @@
 package metrics
 
 import (
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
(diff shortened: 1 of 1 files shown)


server/metrics/metrics.go
```diff
@@ -351,7 +365,7 @@ var (
 		Namespace: bbNamespace,
 		Subsystem: "remote_cache",
 		Name:      "disk_cache_seconds_since_last_access",
-		Buckets:   prometheus.ExponentialBuckets(1, 10, 9),
+		Buckets:   durationSecBuckets,
```
bduffany, line 368: @tylerwilliams I missed this in the last PR, but it looks like the existing metrics all use usec. Mind if I change this metric + impl to usec as well?

tylerwilliams: Ah yeah I changed it in that last CL because the buckets were reasonable for seconds but not for usec and I figured I only needed second granularity for this metric.

But yeah, OK w/ me if you want to go back to usec and change the buckets too.

server/metrics/metrics.go
```diff
@@ -127,6 +129,18 @@ const (
 	bbNamespace = "buildbuddy"
 )
 
+var (
+	// Bucket constants.
+
+	// durationUsecBuckets is a reasonable bucket setting for microsecond-valued
+	// duration metrics (1 usec to 30 days).
+	durationUsecBuckets = exponentialBucketRange(1, 30*24*float64(time.Hour)/1e3, 2)
```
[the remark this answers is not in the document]
bduffany: much cleaner. Done!

PR #2239, opened 2022-07-07
Move GetKey to server/util from quota manager

luluz66:
So this can be added to metrics

### DIFF
--- a/server/util/quota/quota.go
+++ b/server/util/quota/quota.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"google.golang.org/grpc/metadata"
 )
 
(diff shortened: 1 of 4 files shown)


server/util/quota/quota.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if len(vals) == 0 {
+		return ""
+	}
+	ips := strings.Split(vals[0], ",")
+	return ips[0]
+}
+
+// GetKey get the key for quota accounting from the context.
+// Use group_id if is available from the context, otherwise, use the ip address.
```
tylerwilliams, line 39: nit:

// GetKey gets the key for quota accounting from the context.
// If available, group_id is used, otherwise the key falls back to the ip address.

server/util/quota/quota.go
```diff
@@ -0,0 +1,45 @@
...
+func GetKey(ctx context.Context, env environment.Env) string {
```
tylerwilliams, line 40: do you want this function to return an error if no key is found in the context? What should happen in that case?

tylerwilliams (approved):

PR #2245, opened 2022-07-07
Write cache scorecard under attempt-specific directory

maggie-lou:
Fixes buildbuddy-io/buildbuddy-internal#1287

### DIFF
--- a/server/remote_cache/scorecard/scorecard.go
+++ b/server/remote_cache/scorecard/scorecard.go
@@ -324,9 +324,7 @@ func FillBESMetadata(sc *capb.ScoreCard, files map[string]*bespb.File) {
 }
 
 // Read reads the invocation cache scorecard from the configured blobstore.
-func Read(
-	ctx context.Context, env environment.Env, invocationID string, invocationAttempt uint64,
-) (*capb.ScoreCard, error) {
+func Read(ctx context.Context, env environment.Env, invocationID string, invocationAttempt uint64) (*capb.ScoreCard, error) {
 	blobStore := env.GetBlobstore()
 	buf, err := blobStore.ReadBlob(ctx, blobName(invocationID, invocationAttempt))
 	if err != nil && status.IsNotFoundError(err) {
@@ -347,13 +345,7 @@ func Read(
 }
 
 // Write writes the invocation cache scorecard to the configured blobstore.
-func Write(
-	ctx context.Context,
-	env environment.Env,
-	invocationID string,
-	invocationAttempt uint64,
-	scoreCard *capb.ScoreCard,
-) error {
+func Write(ctx context.Context, env environment.Env, invocationID string, invocationAttempt uint64, scoreCard *capb.ScoreCard) error {
 	scoreCardBuf, err := proto.Marshal(scoreCard)
 	if err != nil {
 		return err
--- a/server/remote_cache/scorecard/scorecard.go
+++ b/server/remote_cache/scorecard/scorecard.go
@@ -2,6 +2,7 @@ package scorecard
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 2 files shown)


server/remote_cache/scorecard/scorecard.go
```diff
@@ -330,12 +347,18 @@ func Read(ctx context.Context, env environment.Env, invocationID string) (*capb.
...
 
 // Write writes the invocation cache scorecard to the configured blobstore.
-func Write(ctx context.Context, env environment.Env, invocationID string, scoreCard *capb.ScoreCard) error {
+func Write(
+	ctx context.Context,
+	env environment.Env,
+	invocationID string,
+	invocationAttempt uint64,
+	scoreCard *capb.ScoreCard,
+) error {
```
tylerwilliams, line 356: nit: the formatting for Read and Write got a little wild but can we leave them as long single-string functions (that's what we prefer and have elsewhere)

server/remote_cache/scorecard/scorecard.go
```diff
@@ -259,9 +260,16 @@ func sortResults(results []*capb.ScoreCard_Result, opts *sortOpts) {
 	})
 }
 
-func blobName(invocationID string) string {
+func blobName(invocationID string, invocationAttempt uint64) string {
```
tylerwilliams, line 263: Maybe peep what zoey did in GetStreamIdFromInvocationIdAndAttempt in server/build_event_protocol/build_event_handler/build_event_handler.go?

[I'm not opposed to doing a double-read if we have to, but one thing I like about the other approach is keeping the scorecard library agnostic of "attempt" -- by making the attempt number part of the invocation ID. Then it becomes an implementation detail of the build_event_handler, which it is, rather than complicating the scorecard library]

tylerwilliams (changes requested):

server/remote_cache/scorecard/scorecard.go
maggie-lou: If attempt is an implementation detail of the build_event_handler, does it really make sense for us to add attempt number to the directory naming structure? For example if new clients in the future need to read scorecards, they'll also need the concept of an attempt

Also can you explain more how we would make the attempt number part of the invocation ID? Esp re: 
1. Old scorecards are still saved under just the invocation ID, so at some point we still need the two separated. We could move the double read into build_event_handler, but Read() is and could be called in other places that would need to re-implement the double read logic
2. scorecard::GetCacheScoreCard() will also need an attempt number to read the new directory naming, so we may not be able to completely keep the scorecard library "attempt"-agnostic.

tylerwilliams: ahhhh i'd forgot we had GetCacheScoreCard in here too. Seeing that, I think I agree with what you're saying so maybe we keep attempt as an arg to Read.

RE double read, are invocation attempts 1-indexed? Can we detect if we're reading old style / new style based on the attempt number? That is sort of what i assumed this code was doing: 

```
func GetStreamIdFromInvocationIdAndAttempt(iid string, attempt uint64) string {
	if attempt == 0 {
                // This invocation predates the attempt-tracking functionality, so its
                // streamId does not contain the attempt number.
                return iid
        }
        return iid + "/" + strconv.FormatUint(attempt, 10)
}
```

maggie-lou: My understanding is that we can't automatically detect if a given invocation is using the old / new style, because attempt numbers already exist now on invocations, we just aren't using them currently. It looks like Zoe's code applies to a situation where there weren't attempt numbers previously

Ex. If we try to read an invocation from yesterday, it would've used old naming, and would've had an attempt number (not in naming, but it would've existed). We'd have no way to differentiate that from an invocation from after this change, that would also have an attempt number

maggie-lou: Correct me if I'm wrong but it seems like supporting old cached scorecards is non-critical? We could also double write to the new naming for like a week, and then in a week only support reading from the new naming and old builds won't be able to pull cached scorecards

tylerwilliams: Ack, you're right re attempt number.

I think just eating the cost of the double read is better -- esp since you're reading new-style first, we sort of expect the amortized cost of the double read -> 1.

So I guess I'm convinced lol :P

tylerwilliams (approved):

server/remote_cache/scorecard/scorecard.go
maggie-lou: Done

server/remote_cache/scorecard/scorecard.go
maggie-lou: Ack

bduffany (approved):
LGTM

PR #2248, opened 2022-07-07
executor dl compression

tylerwilliams:
- Executor support for compressed downloads

(not in this CL, but NewBatchFileFetcher should probably move to cachetools.go)


enterprise/config/buildbuddy.local.yaml
```diff
@@ -3,7 +3,7 @@ app:
   create_group_per_user: true
   add_user_to_domain_group: true
   code_editor_enabled: true
-  default_redis_target: "localhost:6379"
+  default_redis_target: "localhost:7777"
```
vadimberezniker, line 6: did you intend to commit this?

enterprise/config/executor.local.yaml
```diff
@@ -3,6 +3,6 @@ executor:
   docker_socket: /var/run/docker.sock
   docker_inherit_user_ids: true
   enable_firecracker: true
-  app_target: "grpc://localhost:1985"
+  app_target: "grpc://localhost:1996"
```
vadimberezniker, line 6: accident?

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -572,10 +631,7 @@ func (ff *BatchFileFetcher) FetchFiles(filesToFetch FileMap, opts *DownloadTreeO
 					return ff.batchDownloadFiles(ctx, req, filesToFetch, opts)
 				})
 			}(req)
-			req = &repb.BatchReadBlobsRequest{
-				InstanceName: ff.instanceName,
-			}
-			currentBatchRequestSize = 0
```
vadimberezniker, line 578: was this an accidental removal? shouldn't the size be reset?

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -447,34 +455,71 @@ type FileMap map[digest.Key][]*FilePointer
...
 	}
 
-	var rsp, err = ff.casClient.BatchReadBlobs(ctx, req)
+	rsp, err := casClient.BatchReadBlobs(ctx, req)
 	if err != nil {
 		return err
 	}
+
+	var decodeBuf []byte
+	if ff.supportsCompression() {
```
vadimberezniker, line 512: why don't you pass supportsCompression from the caller? would make the code easier to follow (instead of having two seemingly independent calls to supportsCompression())

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -608,14 +669,19 @@ func (ff *BatchFileFetcher) bytestreamReadFiles(ctx context.Context, instanceNam
 	if err != nil {
 		return err
 	}
-	if err := cachetools.GetBlob(ctx, ff.bsClient, digest.NewResourceName(fp.FileNode.Digest, instanceName), f); err != nil {
+	resourceName := digest.NewResourceName(fp.FileNode.Digest, instanceName)
+	if ff.supportsCompression() {
```
vadimberezniker, line 673: same here, why not pass supportsCompression from the caller?

vadimberezniker (changes requested):

enterprise/server/remote_execution/dirtools/dirtools.go
tylerwilliams: I removed the other one / double checking that was was happening before.

I'd prefer to leave this one because the call is only done once, beyond that it's just referencing an instance var, and bytestreamReadFiles already has uh... 5 args.

enterprise/server/test/integration/remote_execution/remote_execution_test.go
```diff
@@ -939,6 +939,121 @@ func TestComplexActionIO(t *testing.T) {
 	assert.Empty(t, missing)
 }
 
+func TestComplexActionIOWithCompression(t *testing.T) {
```
tylerwilliams, line 942: @bduffany here is the test you recommended adding

tylerwilliams (commented):
(removed accidentally committed config file)

vadimberezniker (approved):

PR #2252, opened 2022-07-08
add a graph for RPCs by Quota Key

luluz66:


tylerwilliams (approved):

tools/metrics/grafana/dashboards/buildbuddy.json
```diff
@@ -8525,219 +8516,290 @@
       "type": "row"
     },
     {
-      "collapsed": true,
+      "collapsed": false,
```
bduffany, line 8519: nit: revert so that all panels stay collapsed on startup

```suggestion
      "collapsed": true,
```

(I'd like to have the script take care of this automatically at some point)

bduffany (approved):

PR #2258, opened 2022-07-11
Improve pebble logging and error handling; remove extraneous checks'

tylerwilliams:


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -742,21 +749,12 @@ func (p *PebbleCache) Reader(ctx context.Context, d *repb.Digest, offset, limit
 	if err != nil {
 		return nil, err
 	}
-	if p.isolation.GetCacheType() == rfpb.Isolation_ACTION_CACHE {
```
vadimberezniker, line 745: Why the remove the remote instance name check?

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: I think it's unnecessary because the remote instance name is stored (well, a crc32 of it) in the file metadata key

PR #2260, opened 2022-07-11
Fix data race in RemoveConnectedExecutor

bduffany:
`GetNodes()` returns a reference to `np.connectedExecutors` which is then read in `enqueueTaskReservations` (e.g. `toNodeInterfaces(nodes)`. `RemoveConnectedExecutor()` can concurrently mutate that slice though in order to remove an executor. The fix is to remove executors by creating a new slice, rather than mutating its contents.

---

**Version bump**: Patch


vadimberezniker (approved):
nice find

bduffany:
> nice find

the race detector caught this one -- don't know what we'd do without it!

PR #2261, opened 2022-07-11
Add protos for stream-based Firecracker execution

bduffany:
This will be useful for:
- Stats (immediate use case): allows us to stream stats and live-update prometheus metrics while the task is running.
- Stdio: allows us to stream stdin and stdout/stderr, so we no longer have to worry about grpc max size restricting the amount of stdout/stderr we can receive.
- Persistent workers: firecracker-based persistent workers require stdin streaming, since we need to stream tasks one-by-one as each task comes in.

---

**Version bump**: Patch

### DIFF
--- a/proto/vmexec.proto
+++ b/proto/vmexec.proto
@@ -4,12 +4,19 @@ package vmexec;
 
 import "google/protobuf/duration.proto";
 import "google/rpc/status.proto";
+import "proto/remote_execution.proto";
 
 // The exec service is run from inside a VM. The host uses this service to
 // execute commands as well as prepare the VM for command execution.
 service Exec {
   // Executes a command in the VM and returns the result of execution.
   rpc Exec(ExecRequest) returns (ExecResponse);
+
+  // Executes a command in the VM and streams back command progress.
+  // TODO(bduffany): Once this is implemented, rename this to Exec
+  // and delete the old Exec() function.
+  rpc ExecStreamed(stream ExecStreamedRequest)
+      returns (stream ExecStreamedResponse);
 
   // Prepares the VM for command execution after a fresh start or a resume.
   rpc Initialize(InitializeRequest) returns (InitializeResponse);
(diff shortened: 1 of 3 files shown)


proto/vmexec.proto
```diff
@@ -62,6 +67,49 @@ message ExecResponse {
...
+    // this field, `open_stdin` must be set in the initial start request,
+    // otherwise execution will be canceled and an INTERNAL error will be
+    // returned. Clients should close the send-side of the stream once there is
+    // no more input to be sent (this will close the stdin pipe).
+    bytes stdin = 2;
+  }
+}
+
+message ExecStreamedResponse {
+  oneof Message {
```
bduffany, line 87: @tylerwilliams I started thinking about the approach where we call Start() and it returns an ID that you can use to attach stdin, get stdout/stderr, and get Stats(), but I realized it might be more complex than this single streaming approach because it would involve 6 separate RPCs, like Start(), ReadStdout(), ReadStderr(), WriteStdin(), PollStats(), Wait(), and it seemed tricky to get all those separate RPCs to talk to each other and coordinate state, and each one would have a different ctx etc. This single RPC approach seemed simpler but maybe there's other factors I'm not considering

tylerwilliams (approved):
This LGTM but please see what Vadim thinks too

proto/vmexec.proto
```diff
@@ -62,6 +69,50 @@ message ExecResponse {
   google.rpc.Status status = 4;
 }
 
+message ExecStreamedRequest {
+  oneof Message {
```
vadimberezniker, line 73: @tylerwilliams I thought you didn't like oneofs?
I don't have a strong feeling either way, but we should have a consistent approach on whether oneofs should or should not be used.

vadimberezniker (approved):

proto/vmexec.proto
bduffany: I am happy to not use a `oneof`, since the Go API is gnarly. Looks like we don't use `oneof` for `RegisterAndStreamWorkRequest` so I'll update this to match.

tylerwilliams: +1 to not using oneof, I missed this

PR #2262, opened 2022-07-11
Improve QuotaManager Performance

luluz66:
1) Use sync.RWMutex instead of Mutex because quota manager will be read
heavy.
2) Remove the warning log where the namespace is not found. This is
   acceptable now that we assume when namespace or default bucket
   doesn't exist, we are not rate-limiting that RPC. Adding this
   warning will pollute our logs.


enterprise/server/quota/quota_manager.go
```diff
@@ -322,11 +322,10 @@ func (qm *QuotaManager) createNamespace(env environment.Env, name string, config
 // default bucket. Returns nil if the namespace is not found or the default bucket
 // is not defined.
 func (qm *QuotaManager) findBucket(namespace string, key string) Bucket {
-	qm.mu.Lock()
+	qm.mu.RLock()
```
tylerwilliams, line 325: is this better than a https://pkg.go.dev/sync#Map

luluz66: Used sync.Map instead

tylerwilliams (approved):

PR #2265, opened 2022-07-11
Task sizing: Export bare runner stats to prometheus

bduffany:
image

---

**Version bump**: Patch


enterprise/server/remote_execution/containers/bare/bare.go
```diff
@@ -28,7 +29,8 @@ func NewBareCommandContainer(opts *Opts) container.CommandContainer {
 }
 
 func (c *bareCommandContainer) Run(ctx context.Context, command *repb.Command, workDir string, creds container.PullCredentials) *interfaces.CommandResult {
-	return commandutil.Run(ctx, command, workDir, c.opts.EnableStats, &container.Stdio{})
+	c.WorkDir = workDir
```
tylerwilliams, line 32: 👀

tylerwilliams (approved):
LGTM but I think it would be nice to remove container.Stats and make stats.Listener an interface because now this stuff is used in a bunch of places.

PR #2267, opened 2022-07-11
Add messages recommending move to remote from cloud

tempoz:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"syscall"
 	"time"
@@ -86,7 +87,7 @@ func InitializeCacheClientsOrDie(cacheTarget string, realEnv *real_environment.R
 	} else {
 		if cacheTarget == "" {
 			log.Fatalf("No cache target was set. Run a local cache or specify one in the config")
-		} else if cacheTarget == "grpcs://cloud.buildbuddy.io" {
+		} else if u, err := url.Parse(cacheTarget); err == nil && u.Hostname() == "cloud.buildbuddy.io" {
 			log.Warning("You are using the old BuildBuddy endpoint, cloud.buildbuddy.io. Migrate `executor.app_target` to remote.buildbuddy.io for improved performance.")
 		}
 		conn, err = grpc_client.DialTarget(cacheTarget)
--- a/enterprise/server/remote_execution/operation/operation.go
+++ b/enterprise/server/remote_execution/operation/operation.go
@@ -3,6 +3,7 @@ package operation
 import (
 	"context"
 	"encoding/base64"
+	"net/url"
 
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
 	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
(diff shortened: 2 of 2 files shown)


tempoz:
* master
  * **PR #2267** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2267" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2267?utm_source=stack-comment).

enterprise/server/cmd/executor/executor.go
```diff
@@ -86,6 +86,8 @@ func InitializeCacheClientsOrDie(cacheTarget string, realEnv *real_environment.R
 	} else {
 		if cacheTarget == "" {
 			log.Fatalf("No cache target was set. Run a local cache or specify one in the config")
+		} else if cacheTarget == "grpcs://cloud.buildbuddy.io" {
```
bduffany, line 89: do we also allow specifying this as `cloud.buildbuddy.io`? wonder if we should check for that here too

enterprise/server/remote_execution/operation/operation.go
```diff
@@ -110,6 +111,12 @@ func ExecuteResponseWithResult(ar *repb.ActionResult, err error) *repb.ExecuteRe
 	if ar != nil {
 		rsp.Result = ar
 	}
+	if target, err := flagutil.GetDereferencedValue[string]("executor.app_target"); err == nil && target == "grpcs://cloud.buildbuddy.io" {
+		if rsp.GetMessage() != "" {
+			rsp.Message = rsp.GetMessage() + "\n"
+		}
+		rsp.Message = rsp.GetMessage() + "This executor is using the old BuildBuddy endpoint, cloud.buildbuddy.io. Migrate `executor.app_target` to remote.buildbuddy.io for improved performance."
+	}
```
bduffany, line 119: in practice, when does this message get displayed? does it show up once per action that is executed, or only once per invocation? (or only when there are other errors with the invocation?)

tempoz: I think this shows up once per action, which is probably too much. I'll change it to show up just at the end.

enterprise/server/cmd/executor/executor.go
tempoz: I'm sorry, I don't think I understand. Check for what here? `cloud.buildbuddy.io` sans the scheme?

bduffany: yeah, checking for the string `"cloud.buildbuddy.io"`, scheme-less. I think if it's configured that way, we'll accept it and it'll default to grpcs://, but I'm not totally sure

enterprise/server/remote_execution/operation/operation.go
```diff
@@ -59,6 +61,19 @@ type FinishWithErrorFunc func(finalErr error) error
...
 func GetStateChangeFunc(stream StreamLike, taskID string, adInstanceDigest *digest.ResourceName) StateChangeFunc {
 	return func(stage repb.ExecutionStage_Value, execResponse *repb.ExecuteResponse) error {
+		if stage == repb.ExecutionStage_COMPLETED {
+			if target, err := flagutil.GetDereferencedValue[string]("executor.app_target"); err == nil {
+				if u, err := url.Parse(target); err == nil && u.Hostname() == "cloud.buildbuddy.io" {
+					if execResponse.GetMessage() != "" {
+						execResponse.Message = execResponse.GetMessage() + "\n"
+					}
+					execResponse.Message = (execResponse.GetMessage() +
+						"The build used the old BuildBuddy endpoint, cloud.buildbuddy.io. " +
```
bduffany, line 71: nit, when I ran a build locally, the log message sort of blends in, because Bazel doesn't prefix it with "message from executor" or anything like that:

image


 Maybe we can add our own prefix? like:

```suggestion
						"NOTE: this BuildBuddy executor is configured to use the old BuildBuddy endpoint, cloud.buildbuddy.io. " +
```

bduffany (approved):

PR #2269, opened 2022-07-12
Add support for CAS-only API keys

siggisim:
This API key type allows users of this key to upload to the CAS, but not the Action Cache.

This allows users to use an API key that restricts them from poisoning / tampering with the remote cache via the Action Cache, but allows them to still upload timing profiles and test logs. CAS uploading is less prone to issues because we validate the SHAs of the content on upload, and artifacts are stored keyed on their SHA.

This also pairs nicely with remote execution, where users can upload inputs to the CAS (where again their SHAs are validated) - but only the trusted remote executors can populate the AC. 

Screen Shot 2022-07-12 at 2 52 18 PM

---

**Version bump**: Patch

### DIFF
--- a/server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
+++ b/server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
@@ -154,7 +154,7 @@ func (s *ContentAddressableStorageServer) BatchUpdateBlobs(ctx context.Context, 
 		return nil, err
 	}
 
-	canWrite, err := capabilities.IsGranted(ctx, s.env, akpb.ApiKey_CACHE_WRITE_CAPABILITY | akpb.ApiKey_CAS_WRITE_CAPABILITY)
+	canWrite, err := capabilities.IsGranted(ctx, s.env, akpb.ApiKey_CACHE_WRITE_CAPABILITY|akpb.ApiKey_CAS_WRITE_CAPABILITY)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


siggisim:
Fixes: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1484
Progress towards: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1068

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -154,11 +154,15 @@ func (s *ContentAddressableStorageServer) BatchUpdateBlobs(ctx context.Context,
 		return nil, err
 	}
 
-	canWrite, err := capabilities.IsGranted(ctx, s.env, akpb.ApiKey_CACHE_WRITE_CAPABILITY)
+	canWriteAll, err := capabilities.IsGranted(ctx, s.env, akpb.ApiKey_CACHE_WRITE_CAPABILITY)
```
tylerwilliams, line 157: might be cleaner if you modified capabilities.IsGranted to take both at once? then you could leave the code as-is for both bytestream.go and cas.go.

This might even just work if you did capabilities.IsGranted(ctx, s.env, akpb.ApiKey_CACHE_WRITE_CAPABILITY | akpb.ApiKey_CAS_WRITE_CAPABILITY)?

tylerwilliams (approved):

PR #2270, opened 2022-07-13
Update splash screen to use `app.build_buddy_url`

tempoz:
Also account for different lengths of URLs / ports for the pretty formatting in the splash screen, which incidentally fixes a subtle bug where the port being a different number of digits than 4 would misalign the right border.

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/splash/splash.go
+++ b/enterprise/server/splash/splash.go
@@ -1,31 +1,61 @@
 package splash
 
 import (
-	"log"
+	"fmt"
+	"math"
+	"net"
+	"net/url"
+	"strconv"
+	"strings"
 
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/logrusorgru/aurora"
 )
 
 type Printer struct{}
 
-func (p *Printer) PrintSplashScreen(port, gRPCPort int) {
-	// Print out a nice welcome message with getting started instructions.
-	log.Printf("")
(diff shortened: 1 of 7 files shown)


tempoz:
* master
  * **PR #2270** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2270" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2270?utm_source=stack-comment).

enterprise/server/splash/splash.go
```diff
@@ -1,31 +1,61 @@
...
+			au.Sprintf(au.Reset("")),
+			au.Sprintf(au.Reset("   You can now view Buildbuddy in the browser:")),
+			au.Sprintf(au.Reset("        %s").Blue().Bold(), viewURL.String()),
+			au.Sprintf(au.Reset("")),
+		}
+	}
+
+	maxLineLength := 0
+	for _, l := range linesColored[false] {
+		maxLineLength = int(math.Max(float64(maxLineLength), float64(len(l))))
```
luluz66, line 47: why we need to convert to float64 then back to int?

tempoz: math.Max takes `float64`s and then returns a `float64`. Hopefully at some point the standard library will have a generic `max` function, but they don't yet :/

PR #2276, opened 2022-07-13
Add API to delete artifacts from the cache

maggie-lou:


enterprise/server/api/api_server.go
```diff
@@ -339,6 +339,11 @@ func (s *APIServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *APIServer) DeleteFile(ctx context.Context, request *apipb.DeleteFileRequest) (*apipb.DeleteFileResponse, error) {
+	//TODO implement me
+	panic("implement me")
```
tylerwilliams, line 344: instead of using panic (we almost never use it in server code) you can use:
```
return nil, status.UnimplementedError("not yet implemented")
```

proto/api/v1/file.proto
```diff
@@ -21,3 +21,11 @@ message File {
   string hash = 3;
   int64 size_bytes = 4;
 }
+
+// Request object for DeleteFile
+message DeleteFileRequest {
+  string uri = 1;
```
tylerwilliams, line 27: nit: Can you add examples of URIs for:

a CAS artifact
a CAS artifact with a remote_instance_name
an Action Cache artifact
an Action Cache artifact with a remote_instance_name

tylerwilliams (changes requested):

proto/api/v1/file.proto
maggie-lou: Done

enterprise/server/api/api_server.go
maggie-lou: Done

proto/api/v1/file.proto
tylerwilliams: I think we want to remove the `blah://cloud.buildbuddy.io/` prefix from these examples because this API is implicitly *for this server*, and we don't want to allow it to try to delete stuff from other, remote, endpoints.

maggie-lou: Ack

tylerwilliams (approved):

PR #2281, opened 2022-07-14
Implement API to delete artifacts from the cache

maggie-lou:


maggie-lou:
Messed up git history - making a new pr

PR #2282, opened 2022-07-15
Fix get group id for quota interceptor

luluz66:
Rename record metrics interceptor to quota b/c we are also checking
quota in that interceptor.

Put the quota interceptor after the auth one.


server/rpc/filters/filters.go
```diff
@@ -213,6 +213,7 @@ func recordRequestMetricsUnaryServerInterceptor(env environment.Env) grpc.UnaryS
 		}
 		if *enableGRPCMetricsByGroupID {
 			if key, err := quota.GetKey(ctx, env); err == nil {
+				log.Info("quota key is %q, key")
```
bduffany, line 216: rm

bduffany (approved):

server/rpc/filters/filters.go
luluz66: Done

PR #2291, opened 2022-07-16
Clear sessions with invalid refresh tokens

bduffany:
If the session access token is expired and the refresh token is invalid (empty, expired, invalidated by oauth provider, etc.), then the session is no longer useful, since both the access token and refresh token cannot be used anymore. As soon as we detect this, delete the session from both client cookies and the DB, so that the user is forced to go through the consent screen and request a new refresh token. (Due to the check [here](https://cs.github.com/buildbuddy-io/buildbuddy/blob/28c93f863d22951d4ac5c6a87e281428b9b20fd6/enterprise/server/auth/auth.go#L629), we only get an access token when the Session-ID cookie is set, not a refresh token, because a refresh token is only returned when authenticating via the consent screen.)

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1515


tylerwilliams (approved):
LGTM, but can we submit after this week's release?

PR #2292, opened 2022-07-16
Implement terminal V2

bduffany:
https://user-images.githubusercontent.com/2414826/179368728-53ae3c5b-dc03-43bc-a7e4-40fcf86ea279.mp4

Replaces the react-lazylog implementation of the terminal component with
one based on react-window for list rendering, plus the ANSI parsing code
from react-lazylog (converted to TypeScript, with some tweaks). Fixes some
bugs and includes usability and performance improvements.

I wound up using the `FixedSizeList` and sticking with our current
approach of wrapping the text manually, rather than using
`VariableSizeList` and letting the browser do the wrapping, because:

- We still have to compute wrapping manually anyway for
  `VariableSizeList` since it requires passing in a function that
  returns the height of each line, and we don't know the height unless we
  know the wrapped line length.
- The implementation becomes error prone because we have to ensure our
  character width (in pixels) and wrapping computations exactly match
  what the browser would compute when rendering the text.
- Scrolling in VariableSizeList is janky since react-window dynamically
  resizes the scroll height as you scroll. When dragging the scrollbar,
  the mouse goes out of sync with the scrollbar, and it feels a bit
  non-intuitive.

Migration approach:

- All new code is in terminal2.css and terminal2.tsx
- All existing terminal styles are moved to terminal1.css, and
  terminal.tsx is renamed to terminal1.tsx
- When TERMINAL_V2_ENABLED is set to True (in app/terminal/defs.bzl), swap
  out terminal1.css and include terminal2.css instead.
- `terminal.tsx` is now generated, and it just exports the appropriate
  component from `terminal1.tsx` or `terminal2.tsx` depending on the
  value of TERMINAL_V2_ENABLED.

---

Bugs fixed:

- Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1216
  - [x] Search, don't filter when pressing enter in the search box.
  - [x] Automatically jump to the first match when searching, or (more
        preferably) any match that is currently in view, or (even more
        preferably) the current match, if it still matches the current
        search term.
  - [x] Add next/prev navigation with Enter/Shift+Enter, or with arrow
        buttons. Next/prev navigation does the minimal scrolling needed to
        get a match in view and also does smooth scrolling, to avoid
        disorienting the user while they navigate through results.
  - [x] Add clear button to clear the search box.
- Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1139
  - [x] Scroll position is now respected while builds are in progress.
- [x] Fixed issue that searching for a phrase that spans multiple lines
      (due to wrapping) would not be matched.
- [x] Fixed issue that searching across ANSI spans did not work. For
      example, searching for "Hello world" did not match
      `<green>Hello</green> world`.
- [x] Fixed issue that wrapping an ANSI span would result in the style
      information being lost for the wrapped line.

Other improvements:

- [x] Reduced memory usage:
  - Test: Loaded the page and let it sit there until the GC kicked in, and
    memory stopped going down any further. The log had ~100K lines.
  - Results: reduced memory from 343MB in TerminalV1 to 275MB in
    TerminalV2 (Chrome/Linux), 149MB to 107MB (Chrome/macOS). On
    Safari/macOS the memory usage was much more variable so it was hard to
    tell, but the difference in memory usage looked neutral to slightly
    improved.
- [x] Increase size of search input and top bar controls (easier click
      target).
- [x] Increase scrollbar size (easier click target) and increase color
      contrast.
- [x] Increase log size on "All" tab even more (2 more rows of logs).
- [x] Wrapping works like a regular terminal (hard break at viewport
      width, regardless of whitespace). This was done partly to simplify
      implementation, but I think it's also an improvement, since when I
      have wrapping enabled, my preference is to not have to scroll
      horizontally.
- [x] Debounce searching to prevent jank when there are lots of results.
- [x] Reduced app JS bundle size (-3%)

Tested:

- Logs tab looks good
  - [x] Non-dense
  - [x] Dense
- [x] Small screen: search is hidden
- [x] Test log (two title rows)
  - One issue I noticed is that when scrolling continuously, the scrolling
    gets "snagged" each time you advance the scroll position by one full
    viewport height, and you have to let go of the mouse for a second to
    unsnag it and be able to scroll again. This only happens on the test
    log for some reason. It happens with the old terminal as well, so I
    don't think it's a blocker for this PR.
- [x] Action page (no title rows)
- [x] Light theme overall looks good
- Log downloading
  - [x] Shows spinner
  - [x] Downloads logs
- Wrapping
  - [x] Wrapping onto 2 lines
  - [x] Wrapping onto 3+ lines
  - [x] Search that spans multiple lines
  - [x] Wrapping within ANSI spans preserves the styling
- ANSI styles
  - [x] Supports all the same styles as terminal V1
  - [x] Sufficient color contrast in light theme
- Chrome/Linux
  - [x] Small log
  - [x] Huge log (100K lines)
- Firefox/Linux
  - [x] Small log
  - [x] Huge log (100K lines)
- Chrome/macOS
  - [x] Small log
  - [x] Huge log (100K lines)
- Safari/macOS
  - [x] Small log
  - [x] Huge log (100K lines)

---

**Version bump**: Patch


app/terminal/BUILD
```diff
@@ -1,10 +1,33 @@
...
+genrule(
+    name = "index",
+    outs = ["terminal.tsx"],
+    cmd_bash = """
+        echo '
+import { TerminalComponent } from "./%s";
+
+export { TerminalComponent };
+export default TerminalComponent;' > $@
+    """ % ("terminal2" if TERMINAL_V2_ENABLED else "terminal1"),
```
siggisim, line 19: I'd rather just include both terminals and have larger js bundles for a bit during the transition.

siggisim (approved):
Just a few nits.

I know you tested with a 100k line log, but can we test with a really really big log? Like 100-300MB? I've seen these in the wild.

app/BUILD.bazel
```diff
@@ -63,6 +64,8 @@ genrule(
         "//app/components/radio:radio.css",
         "//app/components/checkbox:checkbox.css",
         "//app/components/spinner:spinner.css",
+    ] + [
```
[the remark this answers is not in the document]
bduffany: Resolved offline, the issue is scoping, but it feels a bit error prone to ensure each V1 style is properly scoped, since it's not as simple as prepending a `.terminal1` prefix to all rules. Will keep as-is

app/terminal/BUILD
bduffany: Resolved offline, will keep this as-is.

bduffany:
> can we test with a really really big log? Like 100-300MB? I've seen these in the wild.

I tested with the largest log I could find -- I did run into an issue with the text alignment being off, which I believe is because of floating point rounding errors when the number of lines is very large. I resolved by applying a cap of 835K lines. react-lazylog applies this same limit, so there is no behavioral change.

From a performance point of view, the new log handles the largest logs much better - the log loads in only a few seconds without any issues, but with react-lazylog it takes 30s to a minute or so, and the app becomes unresponsive several times while it is loading.

siggisim (approved):

siggisim:
> > can we test with a really really big log? Like 100-300MB? I've seen these in the wild.
> 
> I tested with the largest log I could find -- I did run into an issue with the text alignment being off near the bottom of the logs, which I think is because of floating point rounding errors when the number of lines is very large. I resolved this by applying a max of 835K rows. react-lazylog applies this same limit, so there is no behavioral change.
> 
> From a performance point of view, the new terminal handles the largest logs much better - the log loads in only a few seconds without any issues, but with react-lazylog it takes 30s to a minute or so, and the app becomes unresponsive several times while it is loading.

Excellent - let's do it!

PR #2293, opened 2022-07-18
Scan and delete

tylerwilliams:
Add code to pebble_cache which will scan and delete files from the blobDir which do not have corresponding records in the pebble DB. This is to remove corrupted data which is effectively invisible and taking up space on the partition but not being used.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -378,7 +378,7 @@ func (p *PebbleCache) deleteOrphanedFiles(quitChan chan struct{}) {
 					if err != nil {
 						return err
 					}
-					log.Printf("Would delete orphaned file: %s (last modified: %s) which is not in cache", path, fi.ModTime())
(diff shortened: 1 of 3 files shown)


vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -52,6 +54,7 @@ var (
 	forceAllowMigration       = flag.Bool("cache.pebble.force_allow_migration", false, "If set, allow migrating into an existing pebble cache")
 	clearCacheBeforeMigration = flag.Bool("cache.pebble.clear_cache_before_migration", false, "If set, clear any existing cache content before migrating")
 	mirrorActiveDiskCache     = flagtypes.Alias[bool]("cache.disk.enable_live_updates", "cache.pebble.mirror_active_disk_cache")
+	orphanDeleteDryRun        = flag.Bool("cache.disk.orphan_delete_dry_run", true, "If set, log orphaned files instead of deleting them")
```
vadimberezniker, line 57: NBD, but should this be cache.pebble and not cache.disk?

vadimberezniker (approved):

PR #2295, opened 2022-07-18
Update NPM bazelisk to v1.12.0

bduffany:
Update the bazelisk version in node_modules to one that supports M1 Macs.

---

**Version bump**: Patch


tempoz (approved):
LGTM

PR #2296, opened 2022-07-18
Task sizing: Add more dimensions to prometheus metrics

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/tasksize/tasksize.go
+++ b/enterprise/server/tasksize/tasksize.go
@@ -120,9 +120,13 @@ func (s *taskSizer) Estimate(ctx context.Context, task *repb.ExecutionTask) *scp
 	}
 	statusLabel := "hit"
 	defer func() {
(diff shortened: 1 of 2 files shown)


vadimberezniker (approved):
I'm fine with these changes, but I know @tylerwilliams has previously expressed concerns about using group ID as a dimension since it can cause these metrics to have high cardinality. I personally think the trade-off is worth it. Please get his LGTM before merging.

bduffany:
For more context, I plan on using the group ID label to check whether certain groups are running into errors with task sizer misses more often than other groups, and then that will allow looking at the actions for those groups to see what is going on.

bduffany:
@vadimberezniker I am going to merge this to get it in this week's release -- I don't expect that the cardinality will be an issue for this metric since we already have other metrics using group ID. If we need to optimize metric size later then we can always revisit.

PR #2297, opened 2022-07-18
Allow users to cancel invocation from UI

maggie-lou:


app/invocation/invocation_buttons.tsx
```diff
@@ -26,6 +27,9 @@ export default class InvocationButtons extends React.Component<InvocationButtons
         {this.props.model.isWorkflowInvocation() && this.canRerunWorkflow() && (
           <WorkflowRerunButton model={this.props.model} />
         )}
+        {this.props.model.getIsRBEEnabled() && this.props.model.isInProgress() && (
```
bduffany, line 30: I'm not sure if this will work for regular Bazel invocations -- we probably want to change the condition here from `getIsRBEEnabled` to `isWorkflowInvocation`

app/invocation/invocation_cancel_button.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  render() {
+    return (
+      <div className="invocation-cancel-button-container">
+        <OutlinedButton onClick={this.onClick.bind(this)}>
+          <SlashIcon className="icon" />
+          <div>Cancel</div>
+        </OutlinedButton>
+      </div>
+    );
```
bduffany, line 26: nit: add a loading state, and also set the `disabled` prop on the button while the RPC is in progress, so that the user can't fire duplicate RPCs, and also so they can get visual feedback as to whether the RPC has completed or not.

You can copy some of the logic from `workflow_rerun_button` for that, mainly the `isLoading` state logic in the function that makes the RPC, and also showing the `<Spinner />` component instead of the SlashIcon while it's loading

to test the loading state locally, you can set the network speed in chrome devtools to something like Fast 3G so that the local RPC is slower, since usually they are almost instant. (I'd recommend only slowing down the network after the app is loaded, since it takes forever to load the full app on Fast 3G settings)

image

app/invocation/invocation_cancel_button.tsx
```diff
@@ -0,0 +1,28 @@
...
+import rpcService from "../service/rpc_service";
+import { invocation } from "../../proto/invocation_ts_proto";
+
+export interface InvocationCancelButtonComponentProps {
+  invocationId: string;
+}
+
+export default class InvocationCancelButtonComponent extends React.Component<InvocationCancelButtonComponentProps> {
+  private onClick() {
+    rpcService.service.cancelInvocation(
```
bduffany, line 13: nit: add an error handler in case the RPC fails (`.catch(e => errorService.handleError(e))`). This will show a red banner with the error message.

(We would like to make this error handling more automatic instead of having to do it manually for every RPC, but aren't quite there yet)

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -142,6 +142,15 @@ func (s *BuildBuddyServer) DeleteInvocation(ctx context.Context, req *inpb.Delet
 	return &inpb.DeleteInvocationResponse{}, nil
 }
 
+func (s *BuildBuddyServer) CancelInvocation(ctx context.Context, req *inpb.CancelInvocationRequest) (*inpb.CancelInvocationResponse, error) {
+	res := s.env.GetRemoteExecutionService()
```
bduffany, line 146: we'll want a perms check here to check that the user has write access to the invocation, since it looks like execution_server doesn't check those permissions for us. The write access check is important because workflow invocations can be public, but we don't want any old user to be able to cancel a workflow invocation.

You could use env.GetInvocationDB() for looking up the invocation, which checks read access for the authenticated user. Then you could use perms.AuthorizeWrite() to check write access as well.

bduffany (changes requested):

app/invocation/invocation_buttons.tsx
maggie-lou: @vadimberezniker [This Cancel method](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/execution_server/execution_server.go#L934) should work for both regular Bazel + workflow invocations right?

Based on the code it looks generalizable but when I try to cancel a Bazel build locally it continues to completion. Having trouble verifying if that's just bc the build is too fast, so almost everything has already been executed by the time I'm able to cancel

vadimberezniker: Are you testing with "remote execution" enabled? The Cancel API cancels all the remote actions associated with an invocation, it does not cancel the invocation itself. So if you are manually running bazel w/o remote execution enabled nothing should happen.

bduffany: Since https://github.com/buildbuddy-io/buildbuddy/pull/1858, I think the "outer" workflow invocation is now associated with the CI runner action, so cancelling the workflow invocation should cancel the CI runner execution and as a result the nested Bazel process should be killed -- which I guess would result in a disconnected inner invocation

maggie-lou: Done

app/invocation/invocation_cancel_button.tsx
maggie-lou: Done

app/invocation/invocation_cancel_button.tsx
maggie-lou: Done

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: Done

maggie-lou:
Sorry messed up the git history when trying to pull in recent changes. Made a new pr here: https://github.com/buildbuddy-io/buildbuddy/pull/2329

PR #2300, opened 2022-07-18
Move QueryWriteStatus short-circuiting before the cache lookup

bduffany:
D'oh.

---

**Version bump**: Patch


vadimberezniker:
This is to avoid an extra cache lookup, correct? There's no change in behavior.

vadimberezniker (approved):

bduffany:
> This is to avoid an extra cache lookup, correct? There's no change in behavior.

yep

PR #2302, opened 2022-07-19
inline atime

tylerwilliams:
- Store atime in metadata
- Update atime when reading files without one

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -298,7 +298,7 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 			if err := disk.EnsureDirectoryExists(blobDir); err != nil {
 				return err
 			}
-			pe, err := newPartitionEvictor(part, blobDir, pc.db, pc.atimes)
+			pe, err := newPartitionEvictor(part, blobDir, pc.db, pc.atimes, pc.accesses)
 			if err != nil {
 				return err
 			}
@@ -667,7 +667,7 @@ func (p *PebbleCache) blobDir() string {
 }
 
 func (p *PebbleCache) updateAtime(fileMetadataKey []byte) {
-	p.sendAtimeUpdate(fileMetadataKey)
+	sendAtimeUpdate(p.accesses, fileMetadataKey)
 	nowNanos := time.Now().UnixNano()
 
 	fmkHash := xxhash.Sum64(fileMetadataKey)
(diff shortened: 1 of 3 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -315,6 +330,70 @@ func partitionLimits(partID string) ([]byte, []byte) {
...
+				log.Warningf("Error unmarshaling data for %q: %s", accessTimeUpdate.fileMetadataKey, err)
+				continue
+			}
+			if err := batchEditAtime(batch, accessTimeUpdate.fileMetadataKey, md); err != nil {
+				log.Warningf("Error updating atime: %s", err)
+			}
+			if int(batch.Count()) >= *atimeWriteBatchSize {
+				flush()
+			}
+		case <-time.After(atimeFlushPeriod):
```
vadimberezniker, line 383: We'll never get to this case if there's a constant trickle of updates. Would it be better to initialize a timer and reset it every time a new batch is created?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -315,6 +330,70 @@ func partitionLimits(partID string) ([]byte, []byte) {
...
+			flush()
+		case <-time.After(time.Second):
+			exitMu.Lock()
+			done := exiting
```
vadimberezniker, line 387: Why not add a case for quitChan directly instead of this roundabout check?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1353,22 +1480,30 @@ func (e *partitionEvictor) resampleK(k int) error {
...
-		}
+		additions = append(additions, entries...)
 	}
 
-	for _, sample := range e.samplePool {
+	// refresh all the entries already in the pool.
+	for i, sample := range e.samplePool {
+		fm, err := readFileMetadata(e.reader, sample.fileMetadataKey)
+		if err != nil {
+			e.samplePool = append(e.samplePool[:i], e.samplePool[i+1:]...)
```
vadimberezniker, line 1497: is it safe to modify the slice while you're iterating over it?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1353,22 +1480,30 @@ func (e *partitionEvictor) resampleK(k int) error {
...
+		}
```
vadimberezniker, line 1498: think you are missing a `continue` here

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: Doing it this way allows us to drain the atime queue before which I wanted

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: Done.

enterprise/server/backends/pebble_cache/pebble_cache.go

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1353,22 +1484,33 @@ func (e *partitionEvictor) resampleK(k int) error {
...
 			return err
 		}
-		for _, entry := range entries {
-			e.samplePool = append(e.samplePool, entry)
-		}
+		additions = append(additions, entries...)
+
 	}
 
+	filtered := e.samplePool[:0]
```
vadimberezniker, line 1498: can you make a new slice here instead of reusing samplePool? appending to filtered will modify the array backing samplePool which you are iterating over.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go

PR #2303, opened 2022-07-19
Pretty-print protos in cas tool

bduffany:
---

**Version bump**: TODO

### DIFF
--- a/tools/cas/cas.go
+++ b/tools/cas/cas.go
@@ -98,8 +98,7 @@ func main() {
 				log.Fatal(err.Error())
 			}
 		}
-		out, _ := prototext.Marshal(ar)
-		fmt.Println(string(out))
+		printMessage(ar)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


tools/cas/cas.go
```diff
@@ -127,6 +125,10 @@ func main() {
 	if err := cachetools.GetBlobAsProto(ctx, bsClient, ind, msg); err != nil {
 		log.Fatal(err.Error())
 	}
-	out, _ := prototext.Marshal(msg)
+	printMessage(msg)
+}
+
+func printMessage(msg proto.Message) {
+	out, _ := prototext.MarshalOptions{Multiline: true}.Marshal(msg)
```
vadimberezniker, line 132: Is this different from proto.MarshalTextString()?

vadimberezniker (approved):

tools/cas/cas.go
bduffany: I think since we migrated from the github.com protobuf package to google.golang.org package, `proto.MarshalTextString()` is no longer available and prototext.Marshal is the new way to do it

PR #2306, opened 2022-07-20
Fix a few more data races in remote_execution_test

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1521

### DIFF
--- a/server/testutil/testenv/testenv.go
+++ b/server/testutil/testenv/testenv.go
@@ -5,7 +5,6 @@ import (
 	"flag"
 	"fmt"
 	"net"
-	"sync"
 	"testing"
 
 	"github.com/buildbuddy-io/buildbuddy/server/backends/blobstore"
@@ -28,9 +27,13 @@ import (
 
 var (
 	useMySQL = flag.Bool("testenv.use_mysql", false, "Whether to use MySQL instead of sqlite for tests.")
-
-	configureLoggerOnce sync.Once
 )
+
+func init() {
+	*log.LogLevel = "debug"
+	*log.IncludeShortFileName = true
+	log.Configure()
+}
 
 type ConfigTemplateParams struct {
 	TestRootDir string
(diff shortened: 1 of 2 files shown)


server/testutil/testenv/testenv.go
```diff
@@ -104,11 +102,7 @@ func (te *TestEnv) GRPCServer(lis net.Listener) (*grpc.Server, func()) {
 }
 
 func GetTestEnv(t testing.TB) *TestEnv {
-	configureLoggerOnce.Do(func() {
-		*log.LogLevel = "debug"
-		*log.IncludeShortFileName = true
-		log.Configure()
-	})
+	testlog.Configure()
```
vadimberezniker, line 105: Tyler might kill me for suggesting this, but what if we put this in a `init` func in testenv instead? 
Would avoid having to introduce this separate testlog utility.

vadimberezniker (approved):

server/testutil/testenv/testenv.go
bduffany: Now that I think about it, it actually used to be in an init() func -- Tyler was even the one who put it there IIRC :smile: 

@tempoz it looks like this was moved this to a `sync.Once` intentionally https://github.com/buildbuddy-io/buildbuddy/commit/3f08babfe01fa55b43c84477b1a6f564aa0fcd70 any chance you remember the context for that?

tempoz: I think this might be vestigial from the interstitial period when we still had to parse and reconcile flags; very possibly it would be fine to revert to the init function. I am 90% sure that the change was motivated by very obvious failures, so it should be pretty clear whether or not it's okay to change it.

bduffany: Thanks for looking at this - changed it back to an `init()` for now

PR #2313, opened 2022-07-20
Reduce the work done on invocation lookup

tempoz:
A lot of redundant work was happening on invocation lookup, and we were compiling regexes every time we used them in the redactor.
- Combined two similar regexes and compiled the redaction regexes in the var block
- Changed redaction flag check to compare against the flags instead of 0 so redaction now runs if any flag is missing rather than only if all are.
- Removed parser in lookup; we only need to address accumulating the events, the structured command line, and the console buffer if the console log isn't chunked.

---

**Version bump**: Patch


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -1080,16 +1085,34 @@ func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*in
...
+				// progress output in the event so they don't eat up
+				// memory.
+				p.Progress.Stderr = ""
+				p.Progress.Stdout = ""
+			case *build_event_stream.BuildEvent_StructuredCommandLine:
+				structuredCommandLines = append(structuredCommandLines, p.StructuredCommandLine)
+			}
 		}
 		invocationMu.Lock()
-		parser.FillInvocation(invocation)
```
bduffany, line 1092: If I'm following correctly, the idea here is that since we already do this when before writing the invocation to the DB, all the invocation fields should already be filled by TableInvocationToProto.

For in-progress invocations though, I think we flush to blobstore more frequently than we flush to mysql. I think we only update the mysql row like 3 times in total throughout the invocation, but blobstore is updated every time we hit the chunk size limit in protofile.go

So I think if we remove this logic to go through the event parser, invocations might feel a bit more "stale"

tempoz: I have now addressed this by ensuring all the "starting events" (direct children of the `BuildStarted` event that contain metadata) are processed prior to the second db write, which happens very early during processing. The goal here is to eventually not read the events protofile at all during invocation lookup.

bduffany: This looks like a reasonably safe change but would you mind putting it behind a flag in case we find that it is causing the metadata to be written later than we expect and we need a way to go back to the old behavior?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -695,6 +709,25 @@ func (e *EventChannel) handleEvent(event *pepb.PublishBuildToolEventStreamReques
 		log.Debugf("First event! sequence: %d invocation_id: %s, project_id: %s, notification_keywords: %s", seqNo, iid, event.ProjectId, event.NotificationKeywords)
 	}
 
+	if e.isFirstStartedEvent(&bazelBuildEvent) {
+		e.hasReceivedStartedEvent = true
+		e.unprocessedStartingEvents[bazelBuildEvent.Id.String()] = struct{}{}
+		log.Info(bazelBuildEvent.Id.String())
```
bduffany, line 715: rm

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: yeah no problem

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz:
```suggestion
```

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: Done!

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -75,8 +76,9 @@ const (
 )
 
 var (
-	chunkFileSizeBytes     = flag.Int("storage.chunk_file_size_bytes", 3_000_000 /* 3 MB */, "How many bytes to buffer in memory before flushing a chunk of build protocol data to disk.")
-	enableChunkedEventLogs = flag.Bool("storage.enable_chunked_event_logs", false, "If true, Event logs will be stored separately from the invocation proto in chunks.")
+	chunkFileSizeBytes           = flag.Int("storage.chunk_file_size_bytes", 3_000_000 /* 3 MB */, "How many bytes to buffer in memory before flushing a chunk of build protocol data to disk.")
+	enableChunkedEventLogs       = flag.Bool("storage.enable_chunked_event_logs", false, "If true, Event logs will be stored separately from the invocation proto in chunks.")
+	enableFastInvocationResponse = flag.Bool("app.enable_fast_invocation_response", false, "If true, invocation responses will be filled from database values and skip parsing the events on read.")
```
bduffany, line 81: nit: consider renaming to `app.disable_invocation_event_parse_on_read`. Also consider defaulting this to true so that you don't need to explicitly configure it -- if we run into any issues we can always change the default to false and redeploy.

bduffany (approved):

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: fair enough, though I think we want the default for booleans to be false because of how flags work, so i'll use `app.require_invocation_event_parse_on_read`?

bduffany: > we want the default for booleans to be false because of how flags work

hm would that mean that flipping the default to `true` wouldn't do anything? seems a bit surprising - but if that is not the case then lgtm. just want to make sure we have a way to revert this with a one-line change if needed.

tempoz: No, that works, it's just the convention that boolean flags are default false so you can flip the behavior from the default with `--flag` instead of `--flag=false`. NO big deal either way, though.

PR #2315, opened 2022-07-21
Update task sizing semantics in scheduling API

bduffany:
This PR is just adding new fields / comments as a proposal for addressing a scheduling issue with task sizing.

With the current task sizing implementation, there's an issue in which, after an executor has executed a task, the task may no longer be schedulable on that executor due to the measured task size exceeding the executor's resource limits. This is not an intended outcome at least with this first iteration of task sizing, since it breaks a lot of tasks that currently schedule and execute just fine.

The proposed fix is to never use measured task sizes for scheduler -> executor probing, and instead always use the "original" task size estimate (either the one from platform props, or the one based on our heuristics). Then, on the executor, we can use the measured_task_size, but capped to the limits of the priority_task_scheduler.

This way, we still get most of the benefits of task sizing, while still using the scheduler's current scheduling decisions.

If this proposal sounds good then I plan to implement it as follow-up PRs to the scheduler & executor.

---

**Version bump**: Patch

### DIFF
--- a/proto/scheduler.proto
+++ b/proto/scheduler.proto
@@ -58,8 +58,27 @@ message TaskSize {
   int64 estimated_free_disk_bytes = 3;
 }
 
+// Next ID: 8
 message SchedulingMetadata {
+  // Task size either explicitly requested via platform properties, or estimated
+  // by BuildBuddy (using basic heuristics).
+  //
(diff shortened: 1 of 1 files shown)


vadimberezniker (approved):
Your proposed fix SGTM.

PR #2316, opened 2022-07-21
Fix deadlock in execution_server::Cancel

maggie-lou:
- Unnest transactions in Cancel API, which cause a deadlock
- Add unit test for Cancel API


maggie-lou:
Context for this fix: https://buildbuddy-corp.slack.com/archives/C01D5GHRJ59/p1658355918346579

enterprise/server/remote_execution/execution_server/execution_server_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func createExecution(t *testing.T, db *gorm.DB, execution *tables.Execution) {
+	err := db.Create(execution).Error
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func createExecutionRedis(t *testing.T, redisClient redis.UniversalClient, executionID string) {
```
vadimberezniker, line 60: I don't love the idea of manually crafting synthetic data in redis in the test. We have a slightly higher level test in remote_execution_test that tests cancellation (TestInvocationCancellation). I wonder if it doesn't trigger the issue because that test only creates a single execution. What do you think of modifying that test to create multiple executions?

vadimberezniker (changes requested):

enterprise/server/remote_execution/execution_server/execution_server_test.go
maggie-lou: Is the concern that the unit test shouldn't actually be writing to the db?

Can (and will) definitely modify that test too. Verified that adding a second execution would've caused that test to timeout. 

Personally I like having another unit test for execution_server::Cancel. If it fails, it becomes really clear where the bug is. Even if remote_execution_test::TestInvocationCancellation had failed, there's more ambiguity whether the bug is how the server handles Interrupted BES events for example. But can look into mocking out redis if that addresses the concern

vadimberezniker: Writing to the DB is fine, but I don't like that the unit test is crafting data (redis structure) that is normally an internal detail of the scheduler service. One way to avoid this is to replace the real scheduler implementation with a fake in your test so that the test is only testing the behavior of execution_server and not of scheduler_server.

maggie-lou: I see - good call. Done

enterprise/server/remote_execution/execution_server/execution_server_test.go
```diff
@@ -0,0 +1,147 @@
+package execution_server
```
bduffany, line 1: nit: we usually prefer writing tests as part of a separate package, like `execution_server_test`, since it usually leads to higher quality tests as you're testing only using the public API of the package, which is how it will actually be used by other parts of the system in practice.

maggie-lou: Done

enterprise/server/scheduling/scheduler_server/scheduler_server_mock.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+
+	scpb "github.com/buildbuddy-io/buildbuddy/proto/scheduler"
+)
+
+type SchedulerServerMock struct{}
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/server/scheduling/scheduler_server/scheduler_server_mock.go
```diff
@@ -0,0 +1,51 @@
...
+	return nil, status.UnimplementedError("not implemented")
+}
+
+func (s *SchedulerServerMock) GetExecutionNodes(ctx context.Context, req *scpb.GetExecutionNodesRequest) (*scpb.GetExecutionNodesResponse, error) {
+	return nil, status.UnimplementedError("not implemented")
+}
+
+func (s *SchedulerServerMock) GetGroupIDAndDefaultPoolForUser(ctx context.Context, os string, useSelfHosted bool) (string, string, error) {
+	return "", "", status.UnimplementedError("not implemented")
+}
```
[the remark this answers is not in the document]
maggie-lou: Nice - done

enterprise/server/remote_execution/execution_server/execution_server_test.go
```diff
@@ -0,0 +1,157 @@
...
+func executionServer(t *testing.T, env *testenv.TestEnv) *execution_server.ExecutionServer {
+	redisTarget := testredis.Start(t).Target
+	rdb := redis.NewClient(redisutil.TargetToOptions(redisTarget))
+	env.SetRemoteExecutionRedisClient(rdb)
+	env.SetRemoteExecutionRedisPubSubClient(rdb)
+
+	router, err := task_router.New(env)
+	if err != nil {
+		t.Fatal(err)
+	}
```
bduffany, line 40: fyi, you can also use `require.NoError(t, err)` here -- a bit more concise but no need to change it if you prefer this style, just throwing it out there as an option

enterprise/server/remote_execution/execution_server/execution_server_test.go
```diff
@@ -0,0 +1,157 @@
...
+	executionID := "blobs/1111111111111111111111111111111111111111111111111111111111111111/100"
+	execution := &tables.Execution{
+		ExecutionID:  executionID,
+		InvocationID: testInvocationID,
+		Stage:        int64(remote_execution.ExecutionStage_EXECUTING),
+	}
+	createExecution(t, db, execution)
+
+	numCanceled, err := s.Cancel(ctx, testInvocationID)
+	assert.Nil(t, err)
```
bduffany, line 82: nit, it gives a clearer error message if you use `assert.NoError`

I also would recommend leaning towards `require` from the testify package instead of `assert`, since require() will fail the test and not move on to future assertions. In this case if there is an error then the assert on numCanceled is expected, and the failed assertion will add more noise to the test.

bduffany (approved):
one more round of just nits, otherwise lgtm

enterprise/server/remote_execution/execution_server/execution_server_test.go
maggie-lou: Done

enterprise/server/remote_execution/execution_server/execution_server_test.go
maggie-lou: Done

enterprise/server/remote_execution/execution_server/execution_server_test.go
```diff
@@ -0,0 +1,157 @@
...
+	assert.Equal(t, 1, numCanceled)
+}
+
+func TestCancel_SkipCompletedExecution(t *testing.T) {
+	env := testenv.GetTestEnv(t)
+	ctx := context.Background()
+	s := executionServer(t, env)
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/server/remote_execution/execution_server/execution_server_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	executionID := "blobs/1111111111111111111111111111111111111111111111111111111111111111/100"
+	execution := &tables.Execution{
+		ExecutionID:  executionID,
+		InvocationID: testInvocationID,
+		Stage:        int64(remote_execution.ExecutionStage_EXECUTING),
+	}
+	createExecution(t, db, execution)
+
+	numCanceled, err := s.Cancel(ctx, testInvocationID)
```
vadimberezniker, line 72: You could have your fake scheduler server count the # of cancellations and avoid the need to return the # of cancelled executions for the sake of the test.

vadimberezniker (approved):

enterprise/server/remote_execution/execution_server/execution_server_test.go
maggie-lou: Good call -done

PR #2319, opened 2022-07-25
Task sizing: Fix measurements causing tasks to become unschedulable

bduffany:
The scheduler will now apply task size measurements just before enqueueing a task onto an executor, limiting the task size to the executor's resource limits. This means that task sizing should no longer affect schedulability at all.

This approach is different than the original proposal in https://github.com/buildbuddy-io/buildbuddy/pull/2315, which was to have executors apply task size measurements locally. I found that it was cleaner to have the scheduler apply the task size measurements just before enqueueing a task onto the executor, because there were 3 different places where the executor was looking at the task_size field and we'd have to be careful to apply the measured task size in all 3 places as well as any future places. By having the scheduler apply the size adjustment, we don't have to make any changes to the executor behavior at all -- the executor can still just look at the task_size field. This keep's the executor's scheduling logic lightweight and keeps the scheduler doing most of the heavyweight scheduling logic.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -153,7 +153,9 @@ type executorHandle struct {
 	requireAuthorization bool
 	stream               scpb.Scheduler_RegisterAndStreamWorkServer
 	groupID              string
-	registration         *scpb.ExecutionNode
+
+	registrationMu sync.Mutex
+	registration   *scpb.ExecutionNode
 
 	mu       sync.RWMutex
 	requests chan enqueueTaskReservationRequest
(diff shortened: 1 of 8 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -205,13 +206,12 @@ func (h *executorHandle) Serve(ctx context.Context) error {
 	}
 	h.groupID = groupID
 
-	var registeredNode *scpb.ExecutionNode
 	removeConnectedExecutor := func() {
-		if registeredNode == nil {
+		if h.registration == nil {
```
vadimberezniker, line 210: I think you need a lock around h.registration

vadimberezniker (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
bduffany: Done, thanks

PR #2320, opened 2022-07-25
Task sizing: Add initial impl of streaming vmexec API

bduffany:
- Implement `ExecStreamed` RPC in `vmexec.go`
- Add a client library that correctly implements the client streaming protocol.
- A subsequent PR will replace our existing usage of `Exec()` in `firecracker.go` with `ExecStreamed()`.

---

**Version bump**: Patch


vadimberezniker (approved):
looks great

PR #2328, opened 2022-07-26
Use clickhouse when querying invocation stats/trend

luluz66:
This is protected behind a flag.

Note: code to flush invocation stats to clickhouse will be in a followup
PR.


enterprise/config/buildbuddy.local.yaml
```diff
@@ -7,6 +7,8 @@ app:
   enable_target_tracking: true
 database:
   data_source: "sqlite3:///tmp/buildbuddy-enterprise.db"
+clickhouse:
```
vadimberezniker, line 10: I don't think we should tie the naming to a specific implementation. What if we call this 'olap_database' as an analog to 'database'?

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -22,12 +23,14 @@ import (
 type InvocationStatService struct {
 	env environment.Env
 	h   interfaces.DBHandle
+	ch  interfaces.ClickHouseDBHandle
```
vadimberezniker, line 26: I'm not a big fan of these short names for struct members. `ch` is not very obvious. 
I realize there's already an existing member called `h`, but I don't think that's much better.
What we rename `ch` to `olapdb` and separately rename `h` to `db`?

server/interfaces/interfaces.go
```diff
@@ -272,6 +272,11 @@ type DBHandle interface {
 	IsDeadlockError(err error) bool
 }
 
+type ClickHouseDBHandle interface {
```
vadimberezniker, line 275: Is our implementation strongly tied to Clickhouse or can we theoretically switch to a different column DB? Would it make sense to call this OLAPDBHandle?

vadimberezniker (changes requested):

enterprise/server/invocation_stat_service/invocation_stat_service.go
luluz66: Done

vadimberezniker (approved):

PR #2329, opened 2022-07-26
Allow users to cancel workflows from UI

maggie-lou:
- Add cancel button to invocation UI
- Add RPC to role_filter
- Address PR comments
- Improve UI and integration test

Continuing off of [this PR](https://github.com/buildbuddy-io/buildbuddy/pull/2297), due to messed up git history
Fixes buildbuddy-io/buildbuddy-internal#1247

### DIFF
--- a/app/invocation/invocation_cancel_button.tsx
+++ b/app/invocation/invocation_cancel_button.tsx
@@ -1,25 +1,36 @@
 import React from "react";
 import { Slash as SlashIcon } from "lucide-react";
+import Spinner from "../components/spinner/spinner";
 import { OutlinedButton } from "../components/button/button";
 import rpcService from "../service/rpc_service";
 import { invocation } from "../../proto/invocation_ts_proto";
+import errorService from "../errors/error_service";
 
 export interface InvocationCancelButtonComponentProps {
   invocationId: string;
 }
 
+type State = {
+  isLoading?: boolean;
+};
+
 export default class InvocationCancelButtonComponent extends React.Component<InvocationCancelButtonComponentProps> {
+  state: State = {};
+
   private onClick() {
-    rpcService.service.cancelInvocation(
-      new invocation.CancelInvocationRequest({ invocationId: this.props.invocationId })
-    );
+    this.setState({ isLoading: true });
+    rpcService.service
+      .cancelInvocation(new invocation.CancelInvocationRequest({ invocationId: this.props.invocationId }))
+      .catch((e) => errorService.handleError(e))
+      .finally(() => this.setState({ isLoading: false }));
   }
 
   render() {
+    const isLoading = this.state.isLoading;
     return (
       <div className="invocation-cancel-button-container">
-        <OutlinedButton onClick={this.onClick.bind(this)}>
-          <SlashIcon className="icon" />
+        <OutlinedButton disabled={isLoading} onClick={this.onClick.bind(this)}>
+          {isLoading ? <Spinner className="icon" /> : <SlashIcon className="icon" />}
           <div>Cancel</div>
         </OutlinedButton>
       </div>
--- a/app/invocation/invocation_cancel_button.tsx
+++ b/app/invocation/invocation_cancel_button.tsx
@@ -12,30 +12,32 @@ export interface InvocationCancelButtonComponentProps {
 
 type State = {
   isLoading?: boolean;
+  cancelled?: boolean;
 };
 
 export default class InvocationCancelButtonComponent extends React.Component<InvocationCancelButtonComponentProps> {
   state: State = {};
 
   private onClick() {
-    this.setState({ isLoading: true });
+    this.setState({ isLoading: true, cancelled: true });
     rpcService.service
-      .cancelInvocation(new invocation.CancelInvocationRequest({ invocationId: this.props.invocationId }))
-      .catch((e) => errorService.handleError(e))
-      .finally(() => {
-        // Delay updating button loading state because after the Cancel RPC completes, can take some time for the
-        // invocation to be marked as disconnected
-        setTimeout(() => {
-          this.setState({ isLoading: false });
-        }, 2000);
-      });
+      .cancelExecutions(new invocation.CancelExecutionsRequest({ invocationId: this.props.invocationId }))
+      .catch((e) => {
+        errorService.handleError(e);
+        this.setState({ cancelled: false });
+      })
+      .finally(() => this.setState({ isLoading: false }));
   }
 
   render() {
     const isLoading = this.state.isLoading;
+    const alreadyCancelled = this.state.cancelled;
     return (
       <div className="invocation-cancel-button-container">
-        <OutlinedButton disabled={isLoading} onClick={this.onClick.bind(this)}>
+        <OutlinedButton
+          disabled={isLoading || alreadyCancelled}
+          onClick={this.onClick.bind(this)}
+          title={alreadyCancelled ? "Invocation has already been cancelled and is now being cleaned up." : undefined}>
           {isLoading ? <Spinner className="icon" /> : <SlashIcon className="icon" />}
           <div>Cancel</div>
         </OutlinedButton>
--- a/app/invocation/invocation_cancel_button.tsx
+++ b/app/invocation/invocation_cancel_button.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+import { Slash as SlashIcon } from "lucide-react";
+import { OutlinedButton } from "../components/button/button";
+import rpcService from "../service/rpc_service";
+import { invocation } from "../../proto/invocation_ts_proto";
+
+export interface InvocationCancelButtonComponentProps {
+  invocationId: string;
+}
+
+export default class InvocationCancelButtonComponent extends React.Component<InvocationCancelButtonComponentProps> {
+  private onClick() {
+    rpcService.service.cancelInvocation(
+      new invocation.CancelInvocationRequest({ invocationId: this.props.invocationId })
+    );
+  }
+
+  render() {
+    return (
+      <div className="invocation-cancel-button-container">
+        <OutlinedButton onClick={this.onClick.bind(this)}>
+          <SlashIcon className="icon" />
+          <div>Cancel</div>
+        </OutlinedButton>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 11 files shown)


enterprise/server/test/integration/workflow/workflow_test.go
```diff
@@ -46,9 +52,39 @@ actions:
   - name: "Test action"
     triggers: { push: { branches: [ master ] } }
     bazel_commands: [ "build //:nop" ]
-`,
+` + osStr + archStr,
```
bduffany, line 55: nit: it is somewhat idiomatic to build the string inline, like this:

```suggestion
    os: "`+runtime.GOOS+`"
    arch: "`+runtime.GOARCH+`"
`
```

imo the whitespace is a bit clearer this way. (same comment for below)

enterprise/server/test/integration/workflow/workflow_test.go
```diff
@@ -46,9 +52,39 @@ actions:
...
 	}
+
+	return fileNameToFileContentsMap
+}
+
+// repoContentsSlowScript simulates a test repo with the config files required to run a workflow
+// It sets up a slow sample script that takes a while to run so the CI runner does not return immediately,
+// giving tests that need to modify the workflow (Ex. for testing cancellation) time to complete
+func repoContentsSlowScript() map[string]string {
```
bduffany, line 64: nit: `repoContentsWithSlowBazelTest` might be a bit clearer

enterprise/server/test/integration/workflow/workflow_test.go
```diff
@@ -86,6 +122,26 @@ func setup(t *testing.T, gp interfaces.GitProvider) (*rbetest.Env, interfaces.Wo
 	return env, workflowService
 }
 
+func triggerWebhook(t *testing.T, gitProvider *testgit.FakeProvider, workflowService interfaces.WorkflowService, repoURL string, commitSHA string, webhookURL string) {
+	// Set up the fake git provider so that GetFileContents can return the
+	// buildbuddy.yaml from our test repo
+	gitProvider.FileContents = repoContents()
```
bduffany, line 128: nit: it would be less error prone to configure this in each test -- since for some tests you want the "slow" repo contents to be returned here

enterprise/server/test/integration/workflow/workflow_test.go
```diff
@@ -31,8 +32,13 @@ import (
 	wfpb "github.com/buildbuddy-io/buildbuddy/proto/workflow"
 )
 
-var (
-	simpleRepoContents = map[string]string{
+// repoContents simulates a test repo with the config files required to run a workflow
+func repoContents() map[string]string {
```
vadimberezniker, line 36: `repoContents` is a bit vague. What do you think about calling this `simpleRepo` and renaming repoContentsSlowScript to `slowScriptRepo` or `repoWithslowScript`?

proto/buildbuddy_service.proto
```diff
@@ -32,6 +32,8 @@ service BuildBuddyService {
       returns (invocation.UpdateInvocationResponse);
   rpc DeleteInvocation(invocation.DeleteInvocationRequest)
       returns (invocation.DeleteInvocationResponse);
+  rpc CancelInvocation(invocation.CancelInvocationRequest)
```
vadimberezniker, line 35: I think we would be better off calling this `CancelExecutions` as this API does not affect the invocation directly.

app/invocation/invocation_cancel_button.tsx
```diff
@@ -0,0 +1,45 @@
...
+    this.setState({ isLoading: true });
+    rpcService.service
+      .cancelInvocation(new invocation.CancelInvocationRequest({ invocationId: this.props.invocationId }))
+      .catch((e) => errorService.handleError(e))
+      .finally(() => {
+        // Delay updating button loading state because after the Cancel RPC completes, can take some time for the
+        // invocation to be marked as disconnected
+        setTimeout(() => {
+          this.setState({ isLoading: false });
+        }, 2000);
```
bduffany, line 30: instead of this delay, you could have a state like "cancelled" and then set it to true when the RPC succeeds, and when in cancelled state the button can stay disabled so the user doesn't try to cancel again. So the `disabled` prop on the button would be like `disabled={this.state.isLoading || this.state.cancelled}`

You could also maybe add a `title` prop to the button which explains the reason that the button is still disabled even though it's no longer loading, e.g. `title={this.state.cancelled ? 'Invocation has already been cancelled and is now being cleaned up.' : undefined}`

The UI should automatically refresh once the invocation is disconnected and then the logic in the `InvocationButtons` component should kick in and render the re-run button instead.

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	uidpb "github.com/buildbuddy-io/buildbuddy/proto/user_id"
```
vadimberezniker, line 7: our style is to separate system imports, unaliased project imports and aliased project imports into separate groups

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -142,6 +143,20 @@ func (s *BuildBuddyServer) DeleteInvocation(ctx context.Context, req *inpb.Delet
 	return &inpb.DeleteInvocationResponse{}, nil
 }
 
+func (s *BuildBuddyServer) CancelInvocation(ctx context.Context, req *inpb.CancelInvocationRequest) (*inpb.CancelInvocationResponse, error) {
+	err := s.authorizeInvocationWrite(ctx, req.InvocationId)
+	if err != nil {
+		return nil, err
+	}
+
+	res := s.env.GetRemoteExecutionService()
```
vadimberezniker, line 152: Remote execution is an optional (enterprise) feature which means GetRemoteExecutionService could theoretically be nil. It's unlikely CancelInvocation would be invoked, but we should be defensive here and check whether it's nil.

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -142,6 +143,20 @@ func (s *BuildBuddyServer) DeleteInvocation(ctx context.Context, req *inpb.Delet
 	return &inpb.DeleteInvocationResponse{}, nil
 }
 
+func (s *BuildBuddyServer) CancelInvocation(ctx context.Context, req *inpb.CancelInvocationRequest) (*inpb.CancelInvocationResponse, error) {
+	err := s.authorizeInvocationWrite(ctx, req.InvocationId)
```
vadimberezniker, line 147: it's a good habit to use the getter (i.e. GetInvocationId) methods on protos as they have additional nil checking.
For example req.GetFoo().GetBar() will not panic even if foo is nil whereas req.Foo.Bar will. 
It's a top level field here so technically it doesn't matter, but it will help avoid issues in more complex code.

bduffany (commented):
looking great! just a few more nits

vadimberezniker (changes requested):

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: Done

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: Done

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: Done

proto/buildbuddy_service.proto
maggie-lou: Done

enterprise/server/test/integration/workflow/workflow_test.go
maggie-lou: Done

app/invocation/invocation_cancel_button.tsx
maggie-lou: Done

enterprise/server/test/integration/workflow/workflow_test.go
maggie-lou: Good catch. Modified the helper function to take file contents as input. Thought it would be nice to have it in a helper function so we don't have to duplicate the descriptive comments in every test

enterprise/server/test/integration/workflow/workflow_test.go
maggie-lou: Done

enterprise/server/test/integration/workflow/workflow_test.go
maggie-lou: Oh nice couldn't figure out how to interpolate a variable

bduffany (approved):
nice! :tada:

server/buildbuddy_server/buildbuddy_server.go
vadimberezniker: I see you updated the use below but not the one on this line.

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -142,6 +143,24 @@ func (s *BuildBuddyServer) DeleteInvocation(ctx context.Context, req *inpb.Delet
...
 
+func (s *BuildBuddyServer) CancelExecutions(ctx context.Context, req *inpb.CancelExecutionsRequest) (*inpb.CancelExecutionsResponse, error) {
+	err := s.authorizeInvocationWrite(ctx, req.InvocationId)
+	if err != nil {
+		return nil, err
+	}
+
+	res := s.env.GetRemoteExecutionService()
+	if res == nil {
+		return nil, status.UnimplementedError("Not Implemented")
```
vadimberezniker, line 154: Would be good to use a more appropriate error here, something like:
status.FailedPreconditionError("remote execution not enabled")

vadimberezniker (approved):

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: Oops done

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: Oops done

PR #2333, opened 2022-07-26
Fix 'invalid nil Timestamp' message in logs when tasks have an error status set

bduffany:
Confirmed via the added logging that this is what's going on.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1508

### DIFF
--- a/enterprise/server/remote_execution/execution_server/execution_server.go
+++ b/enterprise/server/remote_execution/execution_server/execution_server.go
@@ -894,6 +894,15 @@ func (s *ExecutionServer) updateUsage(ctx context.Context, cmd *repb.Command, ex
 	}
 	dur, err := executionDuration(executeResponse.GetResult().GetExecutionMetadata())
 	if err != nil {
+		// If the task encountered an error, it's somewhat expected that the
+		// execution duration will be unset, so don't return an error. For
+		// example, we may have failed to pull the image, so execution could not
+		// even begin. Note that an error doesn't necessarily imply a missing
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -894,6 +894,13 @@ func (s *ExecutionServer) updateUsage(ctx context.Context, cmd *repb.Command, ex
 	}
 	dur, err := executionDuration(executeResponse.GetResult().GetExecutionMetadata())
 	if err != nil {
+		// If the task encountered an error, it's somewhat expected that the
+		// execution duration will be unset, so don't return an error. For
+		// example, we may have failed to pull the image, so execution could not
+		// even begin.
+		if execErr := gstatus.ErrorProto(executeResponse.GetStatus()); execErr != nil {
```
vadimberezniker, line 901: Are there cases where a status would not be OK and we'd still be able to compute the duration? 
i.e. why does this check need to be inside the error check and not before trying to compute duration?

vadimberezniker (approved):

enterprise/server/remote_execution/execution_server/execution_server.go
bduffany: Added a comment - if a task timed out then I believe we'll still get an exec duration, and we still want to collect usage data for those.

PR #2336, opened 2022-07-27
Dial the image converter for each layer.

vadimberezniker:
This will make it more likely that layers are spread out amongst
converter instances.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/image_converter/image_converter.go
+++ b/enterprise/server/image_converter/image_converter.go
@@ -60,9 +60,8 @@ var (
 )
 
 type imageConverter struct {
-	casClient            repb.ContentAddressableStorageClient
-	bsClient             bspb.ByteStreamClient
-	imageConverterClient rgpb.ImageConverterClient
+	casClient repb.ContentAddressableStorageClient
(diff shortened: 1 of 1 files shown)


enterprise/server/image_converter/image_converter.go
```diff
@@ -189,7 +188,12 @@ func (c *imageConverter) convertImage(ctx context.Context, req *rgpb.ConvertImag
 		layerDesc := layerDesc
 
 		eg.Go(func() error {
-			newLayer, err := c.imageConverterClient.ConvertLayer(egCtx, &rgpb.ConvertLayerRequest{
+			convConn, err := grpc_client.DialTarget(*imageConverterBackend)
```
bduffany, line 191: do we need a `convConn.Close()` somewhere?

bduffany (approved):

enterprise/server/image_converter/image_converter.go
vadimberezniker: Good catch!

PR #2347, opened 2022-07-29
Add flag type registry

tempoz:
This introduces `autoflags`, which greatly simplifies the use of `DeprecatedFlag`s. It also allows us to disambiguate the special case of string-based slice flags from the general JSON-based slice flags, which were previously fairly muddled together, while still not requiring the user of the library to make the distinction themselves. Finally, it allows for easy extension of our flag types. `autoflags.New` and `autoflags.Deprecated` can be used anywhere a normal `flag`-style flag declaration, like `flag.Bool` is used, and `autoflags.Var` and `autoflags.DeprecatedVar` can be used anywhere a `flag`-style `Var` flag declaration is used, like `flag.BoolVar`.

---

**Version bump**: Minor

### DIFF
--- a/server/util/flagutil/types/types.go
+++ b/server/util/flagutil/types/types.go
@@ -1,5 +1,14 @@
 package types
 
+// For user-defined flag value types. New flag value types should be declared
+// either as a type definition of the type they contain (see `StringSliceFlag`
+// for an example) or a type definition of a `reflect.Value` which will itself
+// wrap the desired value (see `JSONSliceFlag` for an example). Any new type
+// should also be added to the `Var` function in the `autoflags` subpackage of
+// this package.
+//
+// New flag value types should implement `common.TypeAliased` and `flag.Value`.
+
 import (
 	"encoding/json"
 	"flag"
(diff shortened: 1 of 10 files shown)


tempoz:
* master
  * **PR #2347** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2347" target="_blank">Graphite</a>  👈
    * **PR #2348** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2348" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2347?utm_source=stack-comment).

server/util/flagutil/types/types.go
```diff
@@ -45,95 +54,156 @@ func NewPrimitiveFlag[T bool | time.Duration | float64 | int | int64 | uint | ui
 	return NewPrimitiveFlagVar(&value)
 }
 
-type SliceFlag[T any] []T
+func Slice[T any](name string, defaultValue []T, usage string) *[]T {
```
bduffany, line 57: nit: could you add a comment to Slice and JSONSlice clarifying the difference between the two?

server/util/flagutil/types/types.go
```diff
@@ -258,16 +328,47 @@ type DeprecatedFlag[T any] struct {
...
+	} else if _, ok := converted.(*T); !ok {
+		log.Fatalf("Error creating deprecated flag %s: could not coerce flag of type %T to type %T.", name, converted, (*T)(nil))
+	}
+	flg.Value = &DeprecatedFlag[T]{
+		Value:         flg.Value,
+		name:          flg.Name,
+		MigrationPlan: migrationPlan,
+	}
+	flg.Usage = flg.Usage + " **DEPRECATED** " + migrationPlan
 }
```
bduffany, line 368: nit: are there any cases where `Deprecate` would be used, where `Deprecated` or `DeprecatedVar` could not? wondering if we could make do with just `Deprecated` and `DeprecatedVar`, so that things are more consistent

bduffany (commented):
(accidentally commented on the wrong PR, copying here)

This seems like a nice change overall. I feel that these new funcs would fit better higher up in flagutil.go though, since it is a less deeply nested package, and the name gives an unambiguous signal like "these funcs have high utility and aren't just implementation details -- please do use them for your flag-related needs." wdyt?

server/util/flagutil/types/types.go
tempoz: I remove Slice in the next PR; is that okay?

server/util/flagutil/types/types.go
tempoz: `Deprecate` lets us reach in and deprecate any flag, including flags from other libraries, if we ever start using libs that define their own flags, which could be quite useful. It also makes the definitions in `autoflags` much simpler. I think what I should maybe do, actually, is define `Deprecated` and `DeprecatedVar` in terms of `Deprecate` to remove code duplication.

tempoz: Oh, wait, there is no `Deprecated`, because I couldn't make that without `autoflags`, that's right. Well, I'll define `DeprecatedVar` in terms of `Deprecate`, at least.

tempoz:
So my motivation for making `autoflags` its own package is twofold:

1. It very clearly names exactly what its functions are doing, so readers of the code will not be confused.
1. (This is the ridiculous one): if we imagine the `flagutil` library as its own library, the way to extend it with user-created types would be to write one's own `autoflags` package and use that as a drop-in replacement for the one provided in `flagutil`, which is also part of why I want to keep `autoflags` very simple.

I could duplicate these functions in the top-level flagutil package, though, if you think that's better?

bduffany (approved):
LGTM - maybe get another LGTM as well since this will be used in a bunch of places and it'd be nice to have more input.

I'm cool with keeping the package name as is, but just wanted to expand on my reasoning for moving to "flagutil" a bit more -

> 1. It very clearly names exactly what its functions are doing, so readers of the code will not be confused.

I found the name "autoflags" confusing at first glance. After reading the package, I guess the "auto" is a reference to "automatically" inferring the type of the flag? But as generics become more commonplace I'm not sure the distinction is worth making. golang puts all of its flag stuff in a single "flag" package which is very handy, that's why I think it'd be nice to do the analogous thing and put our own stuff in "flagutil". So the pattern is like, for basic flags, use "flag", for everything else, use "flagutil."

>  if we imagine the flagutil library as its own library, the way to extend it with user-created types would be to write one's own autoflags package and use that as a drop-in replacement for the one provided in flagutil

Sorry but I'm not sure I follow?

luluz66 (approved):

PR #2353, opened 2022-07-29
Use streaming exec RPC for Firecracker containers

bduffany:
Uses the new ExecStreamed API added in https://github.com/buildbuddy-io/buildbuddy/pull/2320 for the Firecracker Exec() implementation.

Effects:
- Populates `UsageStats` in the returned `CommandResult`, which is a pre-requisite for Firecracker task sizing.
  - Note: The task sizer currently ignores Firecracker tasks, so we do not need to worry about Firecracker tasks not being re-schedulable due to these newly measured task sizes.
- Adds support for Firecracker-based persistent workers, since the new streaming API supports stdin.
- Cleans up gRPC max size as well as the extra finalization delay for collecting stdout/stderr, which aren't needed since stdout/stderr are now streamed.

TODO in follow-up PRs: add prometheus metrics; clean up old Exec() RPC.

---

**Version bump**: Patch


server/util/status/status.go
```diff
@@ -219,3 +220,19 @@ func Message(err error) string {
 	}
 	return err.Error()
 }
+
+// FromContextError converts ctx.Err() to the equivalent gRPC status error from
+// this package.
+func FromContextError(ctx context.Context) error {
+	err := ctx.Err()
```
vadimberezniker, line 227: The grpc status library seems to already have this implemented, why don't we call status.FromContextError() here?

vadimberezniker (approved):

server/util/status/status.go
bduffany: Done, thanks

PR #2357, opened 2022-08-01
Prevent shutdown race in pebble_cache

tylerwilliams:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1532


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1674,5 +1818,13 @@ func (p *PebbleCache) Stop() error {
 		return err
 	}
 
+	p.closedMu.Lock()
+	defer p.closedMu.Unlock()
+	if p.closed {
+		return nil
+	}
+	p.closed = true
+	p.dbWaiters.Wait() // wait for all db users to finish up.
```
vadimberezniker, line 1827: Can you add a log before & after the wait so we can tell when the shutdown began & ended?

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: PTAL

PR #2361, opened 2022-08-02
Associate hosted bazel execution with invocation ID.

vadimberezniker:
Prerequisite for being able to cancel the executions.

---

**Version bump**: None


siggisim (approved):
Thanks for doing this!

PR #2365, opened 2022-08-03
Bump up log level for distributed writes.

vadimberezniker:
This should make it easier to root cause the "replication factor"
errors.

---

**Version bump**: None


tylerwilliams (approved):
Thank you!

PR #2373, opened 2022-08-04
Remove disk-cache live migration functionality

tylerwilliams:


tylerwilliams:
Going to keep this around in case it's useful for migrating other large clusters. Will rework so adds / removes are ordered.

PR #2375, opened 2022-08-04
Make partition evictor use db leases too

tylerwilliams:
This is to prevent potential panics / race conditions.

Also, compute some simple summary stats across all partitions by summing the cas, ac, and bytes totals and showing them in /Statusz. (I find myself adding these often when debugging things, so I figured I'd let a computer do it)

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -307,7 +307,7 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 			if err := disk.EnsureDirectoryExists(blobDir); err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -576,12 +581,16 @@ func (p *PebbleCache) scanForBrokenFiles(quitChan chan struct{}) error {
...
+		close(p.brokenFilesDone)
+	}()
+
 	mismatchCount := 0
 	for iter.Next() {
 		// Check if we're shutting down; exit if so.
 		select {
 		case <-quitChan:
-			break
+			return nil
```
tylerwilliams, line 592: this was a bug before -- previously break was only exiting the select loop and this goroutine was not stopping

vadimberezniker (approved):

PR #2377, opened 2022-08-05
Allow running workflows from a subdirectory

bduffany:
---

**Version bump**: TODO


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -499,12 +526,20 @@ func run() error {
 		if err := os.Chdir(repoDirName); err != nil {
 			return err
 		}
```
vadimberezniker, line 528: Do we need to read the config both here & below? Why don't we read it unconditionally prior to checking *shutdownAndExit?

vadimberezniker (approved):

enterprise/server/cmd/ci_runner/main.go
bduffany: The error handling is different below, since the error is reported via BES if we fail to read the config, which is intentional (we want to report this sort of error to the user). I'd prefer to keep the error handling close to where the error is returned, so imo it's best left as-is.

PR #2378, opened 2022-08-05
Executor: Log attempts to enqueue unschedulable tasks

bduffany:
This is to rule out task sizing as a possible cause of tasks hanging.

This should never happen, but if it does, it's better to return an error to the scheduler than enqueue an unschedulable task, which can (a) cause tasks to hang if all 3 probes go to executors which cannot fit the task, and (b) inflate the queue length metric causing a scale-up that can only be reversed by killing the executors with the inflated queues.

---

**Version bump**: Patch


tylerwilliams (approved):

bduffany:
I updated the PR to just log instead of returning an error, because while testing this I noticed that returning an error terminates the executor registration and the scheduler does not cancel the task upon error (which seems by design).

PR #2382, opened 2022-08-09
Add platform property for enabling image streaming.

vadimberezniker:
This will give us a way to measure streaming peformance without having
to enable it by default.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -131,15 +131,6 @@ func NewProvider(env environment.Env, imageCacheAuthenticator *container.ImageCa
 		return nil, err
 	}
 
-	if *imageStreamingEnabled {
-		if *imageStreamingRegistryGRPCTarget == "" {
-			return nil, status.FailedPreconditionError("registry gRPC target must be set to use image streaming")
-		}
-		if *imageStreamingRegistryHTTPTarget == "" {
-			return nil, status.FailedPreconditionErrorf("registry HTTP target must be set to use image streaming")
-		}
-	}
-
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -113,19 +117,29 @@ type pullStatus struct {
...
+
+	if *imageStreamingEnabled {
+		if *imageStreamingRegistryGRPCTarget == "" {
+			return nil, status.FailedPreconditionError("registry gRPC target must be set to use image streaming")
+		}
+		if *imageStreamingRegistryHTTPTarget == "" {
+			return nil, status.FailedPreconditionErrorf("registry HTTP target must be set to use image streaming")
+		}
+	}
+
```
tylerwilliams, line 142: could you do something like setting a flag once you've checked things so the check only happens once?
```
    imageStreamingSupported := false
	if *imageStreamingEnabled {
		if *imageStreamingRegistryGRPCTarget == "" {
			return nil, status.FailedPreconditionError("registry gRPC target must be set to use image streaming")
		}
		if *imageStreamingRegistryHTTPTarget == "" {
			return nil, status.FailedPreconditionErrorf("registry HTTP target must be set to use image streaming")
		}
        imageStreamingSupported = true
	}
```


Then below you can remove the:

```
imageStreamingSupported := *imageStreamingRegistryHTTPTarget != "" && *imageStreamingRegistryGRPCTarget != ""
```

vadimberezniker: Done.

tylerwilliams (approved):

PR #2388, opened 2022-08-10
Exclude filecache-linked files from download metrics

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1559

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -459,6 +459,9 @@ type BatchFileFetcher struct {
 	instanceName string
 	once         *sync.Once
 	compress     bool
+
+	statsMu sync.Mutex
+	stats   repb.IOStats
 }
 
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -459,6 +459,9 @@ type BatchFileFetcher struct {
 	instanceName string
 	once         *sync.Once
 	compress     bool
+
+	statsMu sync.Mutex
+	stats   *repb.IOStats
```
vadimberezniker, line 464: nit: this doesn't need to be a pointer

vadimberezniker (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
bduffany: Done

PR #2397, opened 2022-08-15
Don't store lastUse field in disk_cache's LRU

maggie-lou:
On startup disk cache creates a LRU ledger using last access data from the filesystem to sort entries. Previously we were storing this timestamp in the LRU. Remove this field because it is never updated, even when files are accessed. This may cause ambiguity when we start to expose the actual last access time, which may be different than whatever is stored in this field

### DIFF
--- a/server/backends/disk_cache/disk_cache.go
+++ b/server/backends/disk_cache/disk_cache.go
@@ -418,9 +418,8 @@ type fileRecord struct {
 // fileRecordWrapper is a wrapper for fileRecord that contains additional metadata that does not need to be
 // written in the value field to the LRU cache
 type fileRecordWrapper struct {
-	fileRecord          *fileRecord
-	lastUseNanos        int64
-	lastModifyTimeNanos int64
+	fileRecord   *fileRecord
+	lastUseNanos int64
 }
 
 func (fr *fileRecord) FullPath() string {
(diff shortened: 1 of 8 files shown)


maggie-lou:
* master
  * **PR #2397** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2397" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2397?utm_source=stack-comment).

vadimberezniker:
I think it would be simpler to return the access time & modify time of the underlying file rather than adding the extra bookkeeping to the LRU. This would not reflect any non-read requests (e.g. contains) but it should be good enough to reflect the time of the last  Read attempt.

maggie-lou:
That's a good point on non-read requests. Was trying to make this re-usable by memory_cache, which also uses LRU, but agree that it makes sense to just write something for that use case

What do you think about still removing lastUse from fileRecord, which is the value the disk cache stores in the LRU cache. We never update that field, so was trying to reduce ambiguity if that field has a different value than the actual last access time from the underlying file

vadimberezniker:
> What do you think about still removing lastUse from fileRecord, which is the value the disk cache stores in the LRU cache. We never update that field, so was trying to reduce ambiguity if that field has a different value than the actual last access time from the underlying file

That sounds good.

server/backends/disk_cache/disk_cache.go
```diff
@@ -409,12 +409,19 @@ func newPartition(id string, rootDir string, maxSizeBytes int64, useV2Layout boo
...
+// fileRecord is the data struct we store in the LRU cache
 type fileRecord struct {
 	key       *fileKey
-	lastUse   int64
 	sizeBytes int64
 }
 
+// fileRecordWrapper is a wrapper for fileRecord that contains additional metadata that does not need to be
+// written in the value field to the LRU cache
+type fileRecordWrapper struct {
```
vadimberezniker, line 420: The `wrapper` suffix does not convey any information. What about calling this something like `timestampedFileRecord`? Similarly for the functions below.

vadimberezniker, line 420: golang has a concept of "struct embedding" that makes composing types more convenient by making their fields/funcs accessible without having to reference the embedded type via name. All you need to do is remove the `fileRecord` name so the struct becomes 
type timestampedFileRecord struct {
  *fileRecord
  lastUseNanos int64
}

vadimberezniker (changes requested):

server/backends/disk_cache/disk_cache.go
maggie-lou: Cool - thanks! Still had to do a lot of explicit "timestampedRecord.fileRecord" because certain functions expect the embedded object itself, as opposed to one of its field/method

maggie-lou: Good call - done

vadimberezniker: The `timestampedRecord` is only necessary in `initializeCache`. 
For the other functions, you can create a fileRecord directly. i.e. you can split `makeTimestampedRecord` into `makeTimestampedRecord` and `makeRecord` and use `makeRecord` where you are currently using `makeTimestampedRecord` w/o the lastUse field.

maggie-lou: Done

vadimberezniker (approved):

PR #2402, opened 2022-08-16
Make SQL slow query threshold configurable

bduffany:
Useful for batch tools that interact with the DB directly with more expensive queries, and don't want the console to be spammed with slow query warnings.

---

**Version bump**: Patch


vadimberezniker (approved):

tylerwilliams (approved):
nice!

PR #2403, opened 2022-08-16
Fix double JanitorCutoffThreshold multiplication in pebble cache

maggie-lou:
- Bug was causing items to be evicted prematurely
- Fix flaky unit tests due to truncating floats instead of rounding up. Was also causing items to be evicted prematurely, causing test failures
- Fix TestingWaitForGC. If the cache held exactly the max number of bytes, the test would wait forever and timeout


tylerwilliams (approved):
This is great, thanks for catching and fixing this!

PR #2406, opened 2022-08-17
Changes to update the firecracker sdk

tempoz:
Had to save the client so that we could wait for socket. Should be straightforward otherwise.

---

**Version bump**: Minor


tempoz:
* master
  * **PR #2406** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2406" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2406?utm_source=stack-comment).

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -588,6 +621,12 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
 			ExecFile:       firecrackerBinPath,
 			ChrootStrategy: fcclient.NewNaiveChrootStrategy(""),
 		},
+		Snapshot: fcclient.SnapshotConfig{
+			MemFilePath:         fullMemSnapshotName,
+			SnapshotPath:        vmStateSnapshotName,
+			EnableDiffSnapshots: true,
+			ResumeVM:            false,
```
tylerwilliams, line 628: Without knowing what this does -- are we still able to suspend a vm, snapshot it, reload the snapshot into a new VM and then resume it?

tylerwilliams (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
tempoz: Manually tested it; works 👍

PR #2415, opened 2022-08-19
Fix duplicate target rows in test grid

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1584

### DIFF
--- a/server/build_event_protocol/accumulator/accumulator.go
+++ b/server/build_event_protocol/accumulator/accumulator.go
@@ -101,8 +101,15 @@ func (v *BEValues) InvocationID() string {
 func (v *BEValues) StartTime() time.Time {
 	return v.buildStartTime
 }
+
+// RepoURL returns the normalized repo URL.
 func (v *BEValues) RepoURL() string {
-	return v.getStringValue(repoURLFieldName)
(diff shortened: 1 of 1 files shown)


server/build_event_protocol/accumulator/accumulator.go
```diff
@@ -101,8 +101,15 @@ func (v *BEValues) InvocationID() string {
...
+// RepoURL returns the normalized repo URL.
 func (v *BEValues) RepoURL() string {
-	return v.getStringValue(repoURLFieldName)
+	rawURL := v.getStringValue(repoURLFieldName)
+	normalizedURL, err := gitutil.NormalizeRepoURL(rawURL)
+	if err != nil {
+		return rawURL
+	}
+	return normalizedURL.String()
 }
```
bduffany, line 113: I think this should be a safe change because it is only used in two places:

- target_tracker.go, which is where we need this bugfix
- build_status_reporter.go, which already normalizes this repo URL (indirectly) before it consumes it. Normalizing a repo URL twice should be the same as normalizing it once

siggisim (approved):

luluz66 (approved):

PR #197, opened 2021-02-02
workflows: Move workflow proto into its own package

bduffany:
Also rename `ListWorkflow` to `GetWorkflows` to be consistent with `Get` verbs / plurality of other RPC methods

---

**Version bump**: Patch

PR #2416, opened 2022-08-19
Add rough outline for cache migration proposal

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/enterprise/config/buildbuddy.local.yaml
+++ b/enterprise/config/buildbuddy.local.yaml
@@ -36,7 +36,7 @@ cache:
 # Because `migration` is not set, config/migration.yaml will be ignored
 cache:
   pebble:
-    root_directory: /disk_dir
+    root_directory: /pebble_dir
 
 auth:
   enable_anonymous_usage: true
--- a/enterprise/config/buildbuddy.local.yaml
+++ b/enterprise/config/buildbuddy.local.yaml
@@ -13,8 +13,58 @@ storage:
     root_directory: /tmp/buildbuddy_enterprise
   enable_chunked_event_logs: true
 cache:
+  # Sample flow if migration disk -> pebble
+
+  # Begin double writing, and still return data from disk
   disk:
     root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_src: true
+  pebble:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_dest: true
+
+  # Copy all data from disk -> pebble
+  disk:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_src: true
+  pebble:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_dest: true
+  migration_cache:
+    migrate_and_exit: true
+
+  # Double read for a period, fix any bugs, still use disk as source of truth
+  disk:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_src: true
+  pebble:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_dest: true
+  migration_cache:
+      double_read_percentage: 0.1
+
+  # Once double reads are stable, keep double writing, but now return data from pebble
+  disk:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_dest: true
+  pebble:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_src: true
+
+  # Something goes wrong with pebble - i.e. bug with write logic and we want to stop double writes
+  # and clean up corrupted data in pebble
+  disk:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+  migration:
+     clear_cache_before_migration: true
+
+  # Fully confident with pebble, want to stop double writing to disk
+  pebble:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+
+
+
+
 auth:
   enable_anonymous_usage: true
   enable_self_auth: true
--- a/enterprise/config/buildbuddy.local.yaml
+++ b/enterprise/config/buildbuddy.local.yaml
@@ -12,58 +12,31 @@ storage:
   disk:
     root_directory: /tmp/buildbuddy_enterprise
   enable_chunked_event_logs: true
+
+# No migration - just using disk cache
 cache:
-  # Sample flow if migration disk -> pebble
-
-  # Begin double writing, and still return data from disk
   disk:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_src: true
-  pebble:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_dest: true
+    root_directory: /disk_dir
 
-  # Copy all data from disk -> pebble
+# Start migration - will read migration details from config/migration.yaml
+cache:
+  migration: true
+
+# Something goes wrong with migration - stop double writes and go back to only using disk cache
+# Because `migration` is not set, config/migration.yaml will be ignored
+cache:
   disk:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_src: true
+    root_directory: /disk_dir
+
+# Restart migration after fixing bug - will read migration details from config/migration.yaml
+cache:
+  migration: true
+
+# Fully confident with new cache - stop double writing, only use new pebble cache
+# Because `migration` is not set, config/migration.yaml will be ignored
+cache:
   pebble:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_dest: true
-  migration_cache:
-    migrate_and_exit: true
-
-  # Double read for a period, fix any bugs, still use disk as source of truth
-  disk:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_src: true
-  pebble:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_dest: true
-  migration_cache:
-      double_read_percentage: 0.1
-
-  # Once double reads are stable, keep double writing, but now return data from pebble
-  disk:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_dest: true
-  pebble:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-    migration_src: true
-
-  # Something goes wrong with pebble - i.e. bug with write logic and we want to stop double writes
-  # and clean up corrupted data in pebble
-  disk:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-  migration:
-     clear_cache_before_migration: true
-
-  # Fully confident with pebble, want to stop double writing to disk
-  pebble:
-    root_directory: /tmp/buildbuddy-enterprise-cache
-
-
-
+    root_directory: /disk_dir
 
 auth:
   enable_anonymous_usage: true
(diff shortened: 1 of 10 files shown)


maggie-lou:
* master
  * **PR #2416** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2416" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2416?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -0,0 +1,157 @@
...
+
+	mu             sync.RWMutex
+	eg             *errgroup.Group
+	doubleReadChan chan *repb.Digest
+}
+
+type MigrationDestCache interface {
+	interfaces.Cache
+
+	// Implementation should be async, so that the double write doesn't slow down operations on the primary db
```
tylerwilliams, line 39: With double reads and double writes -- one way I've seen it done is that both take place in parallel, synchronously in the request, and then if the result differs for some reason, a warning is logged. 

If they happen async this isn't possible. What are the tradeoffs involved with doing this async vs sync?

enterprise/config/buildbuddy.local.yaml
```diff
@@ -13,8 +13,58 @@ storage:
     root_directory: /tmp/buildbuddy_enterprise
   enable_chunked_event_logs: true
 cache:
+  # Sample flow if migration disk -> pebble
```
tylerwilliams, line 16: I really liked reading this example config with your comments to get an idea of how this would be used

enterprise/config/buildbuddy.local.yaml
```diff
@@ -13,8 +13,58 @@ storage:
...
+
+  # Double read for a period, fix any bugs, still use disk as source of truth
+  disk:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_src: true
+  pebble:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_dest: true
+  migration_cache:
+      double_read_percentage: 0.1
```
tylerwilliams, line 44: Could you explain what this would mean in more detail? (I have a guess but just want to make sure it's what you're thinking: basically that for 10% of reads we would read from both caches and if there was an error or diff between the results we'd report it?)

enterprise/config/buildbuddy.local.yaml
```diff
@@ -13,8 +13,58 @@ storage:
     root_directory: /tmp/buildbuddy_enterprise
   enable_chunked_event_logs: true
 cache:
+  # Sample flow if migration disk -> pebble
+
```
tylerwilliams, line 17: Since in the config all of this is parented under the "cache:" section -- would we just always create a cache migrator wrapper around whatever cache is configured?

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -0,0 +1,157 @@
...
+			//
+			if *autoFixDoubleReadErrs {
+				c.Dest.Set(ctx, digest, primaryData)
+			}
+		})
+	}
+}
+
+// Migrate copies all data from the src cache to the dest cache
+func (c *MigrationCache) Migrate(ctx context.Context, srcRootDir string) error {
```
tylerwilliams, line 130: One issue we've seen in the past is that Migrating a single app's worth of data can take quite a long time. With 16 apps I think we predicted a full serial migration, app by app, would take 2+ days. During that time a single app would be down.

What do you think about having an "online" migration mode where files are copied between caches in the background while the app runs? This would allow all of the caches to migrate simultaneously while serving traffic.

enterprise/config/buildbuddy.local.yaml
maggie-lou: Yep! Pseudocode is in Get and ProcessDoubleReads

enterprise/config/buildbuddy.local.yaml
maggie-lou: Glad it was helpful!

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Don't know if you saw ProcessDoubleReads() but thought that could be a good way to have sync double reads running in the background, without affecting performance of Get() on the src db

As for double writes: 
Async
+ Performance of source db is not affected. Ideally customers should not know migration is happening, if there's a performance problem with writes to the dest db
- Harder to debug if there's some lag propagating changes to dest db
- Depending on implementation of async write log, might lose data when server restarts. Could be messy to fix those discrepancies
- Switching over source of truth to destination db might be a little off if not all the updates have completed double writing (though probably minor)

Sync
+ Simplicity
- If the writes run in parallel and the src write fails but dest write succeeds, will need to "rollback" changes to the dest bc the src db should be the source of truth. Will need an additional sync delete on the dest db
- If there's a performance problem writing to the new db, could cause propagating performance problems bc we need to hold the lock when writing to both dbs

Listing them out, sync approach might actually be better. We can add another flag to only double write a proportion of writes and slowly ramp up, so we can identify and fix any double write performance problems before affecting all write traffic

enterprise/config/buildbuddy.local.yaml
maggie-lou: My thought was that we always initialize an empty MigrationCache object on the env, but it there's no active migration happening and its src and dest fields aren't set, main.go would just ignore it and continue setting a cache (with no wrapper) like it currently does

maggie-lou: See Line #197 in disk_cache.go for kind of a demonstration of what I mean

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Don't know if you saw ProcessDoubleReads() but thought that could be a good way to have sync double reads running in the background, without affecting performance of Get() on the src db

As for double writes:
Async
/+ Performance of source db is not affected. Ideally customers should not know migration is happening, if there's a performance problem with writes to the dest db
/- Harder to debug if there's some lag propagating changes to dest db
/- Depending on implementation of async write log, might lose data when server restarts. Could be messy to fix those discrepancies
/- Switching over source of truth to destination db might be a little off if not all the updates have completed double writing (though probably minor)

Sync
/+ Simplicity
/- If the writes run in parallel and the src write fails but dest write succeeds, will need to "rollback" changes to the dest bc the src db should be the source of truth. Will need an additional sync delete on the dest db
/- If there's a performance problem writing to the new db, could cause propagating performance problems bc we need to hold the lock when writing to both dbs

Listing them out, sync approach might actually be better. We can add another flag to only double write a proportion of writes and slowly ramp up, so we can identify and fix any double write performance problems before affecting all write traffic

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Side note that doesn't affect your point - can you clarify what you mean by "app" here? 

Yeah that's a good idea. And we could make sure to buffer the online migration so it doesn't monopolize resources or the lock while we're trying to serve traffic

maggie-lou: We can apply different configs to each server, right? filepath.WalkDir walks in lexical order, so it really only makes sense to have 1 server copying data at a time

tylerwilliams: By "app" I mean one cache server.

I agree with your point about ensuring the online migration does not hog resources. We've seen this be an issue in the past, so having a builtin way to ratelimit it will be important.

Will ping you re filepath issue, but I think it's not relevant here because each server stores different data.

enterprise/server/backends/migration_cache/migration_cache.go
tylerwilliams: I like the simplicity of sync.

One idea @vadimberezniker suggested was we could also enqueue stuff for migration when it is read

maggie-lou: Updated the proposal to copy data asynchronously during reads, as opposed to trying to copy everything from the src cache on startup. However that means we will need to do async double writes, otherwise we can't guarantee that the ordering will be correct between the copying and new writes

enterprise/config/buildbuddy.local.yaml
```diff
@@ -13,8 +13,58 @@ storage:
...
+  # Copy all data from disk -> pebble
+  disk:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_src: true
+  pebble:
+    root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_dest: true
+  migration_cache:
+    migrate_and_exit: true
```
siggisim, line 34: Super minor idea - I wonder if there's some nice way we could keep all of the migration stuff in the migration block. That way you wouldn't have to add this (and corresponding flags) to each cache implementation. It also would mean it would be harder to misconfigure by setting multiple caches as src/dest.

Something like:
```
migration:
  source: disk
  dest: pebble
  double_read_percentage: 0.1
```

Then you could switch on these in the migration code and set the right src/dest.

Also wondering how we would handle a pebble => pebble migration - say with a different filesystem layout.

Just some thoughts / ideas. I like the overall design!

enterprise/config/buildbuddy.local.yaml
```diff
@@ -13,8 +13,58 @@ storage:
     root_directory: /tmp/buildbuddy_enterprise
   enable_chunked_event_logs: true
 cache:
+  # Sample flow if migration disk -> pebble
+
+  # Begin double writing, and still return data from disk
```
bduffany, line 18: Do we also want to be able to configure a double write percentage? For example, if double-write would be too expensive, maybe it makes sense to do it only for digests starting with `ff` or something like that.

enterprise/config/buildbuddy.local.yaml
```diff
@@ -13,8 +13,58 @@ storage:
...
   disk:
     root_directory: /tmp/buildbuddy-enterprise-cache
+    migration_src: true
```
vadimberezniker, line 21: Overall, I like this proposal. Siggi's comment about same-cache-type migration needs consideration.
We have some same-cache-type migrations in mind that we want to do (e.g. moving existing pebble cache to smaller disks) and we might have more in the future.

enterprise/config/buildbuddy.local.yaml
maggie-lou: I think a slow rollout of the apps would be cleaner. If we only double write a %, the dest cache may hold some stale data

Ex. 
T0: 
- Write 5 to src
- <Exceeds double write threshold> Write 5 to dest

T1:
- Write 7 to src
- <Does not exceed double write threshold> No write to dest, dest still holds 5

It may be complicated to detect that dest cache is holding stale data (Ex. if no one reads that key for a while, so the src value never gets copied over or double read)

maggie-lou: If we wanted a double write % we could implement a stable way to ensure that if one key is double written, it is always double written, but that feels a bit complicated

bduffany: Slow rollout sgtm

enterprise/config/buildbuddy.local.yaml
maggie-lou: Good call out! Updated the initialization mechanism if you want to take another look @siggisim @tylerwilliams @vadimberezniker @bduffany

PR #2423, opened 2022-08-22
Add last access and modify time to cache requests card

maggie-lou:
buildbuddy-io/buildbuddy-internal#1485


maggie-lou:
* master
  * **PR #2423** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2423" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2423?utm_source=stack-comment).

proto/resource.proto
```diff
@@ -27,5 +26,14 @@ message ResourceName {
   // The cache isolation type of this resource. For now, that can
   // be either CAS or AC. Other cache types may exist in the future.
   // Ex. CAS, AC
-  cache.CacheType cache_type = 4;
+  CacheType cache_type = 4;
+}
+
+// CacheType represents the type of cache being written to.
+enum CacheType {
```
maggie-lou, line 33: There will be an import cycle if we import the cache file. Once we finish the ResourceName refactor the cache package will probably frequently need to reference the resource package, so makes sense to me to prevent this here

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -886,6 +888,48 @@ func (s *BuildBuddyServer) GetUsage(ctx context.Context, req *usagepb.GetUsageRe
...
+	cacheType, err := ProtoCacheTypeToCacheType(resourceName.GetCacheType())
+	if err != nil {
+		return nil, err
+	}
+	cache, err := s.env.GetCache().WithIsolation(ctx, cacheType, resourceName.GetInstanceName())
+	if err != nil {
+		return nil, err
+	}
+
+	metadata, err := cache.Metadata(ctx, resourceName.GetDigest())
```
bduffany, line 907: does calling Metadata() update the last access time? (if so, do we need a flag in the request that tells it not to update?)

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -886,6 +888,48 @@ func (s *BuildBuddyServer) GetUsage(ctx context.Context, req *usagepb.GetUsageRe
...
+func ProtoCacheTypeToCacheType(cacheType resource.CacheType) (interfaces.CacheType, error) {
+	switch cacheType {
+	case resource.CacheType_AC:
+		return interfaces.ActionCacheType, nil
+	case resource.CacheType_CAS:
+		return interfaces.CASCacheType, nil
+	default:
+		return interfaces.UnknownCacheType, status.InvalidArgumentErrorf("unknown cache type %v", cacheType)
+	}
+}
```
bduffany, line 931: can be a separate PR, but it would be cleaner if we removed the cache type enums from interfaces.go and used the protos everywhere

app/invocation/invocation.css
```diff
@@ -360,7 +360,7 @@
 
   display: grid;
   grid-template-columns: min-content auto;
-  grid-gap: 0 8px;
+  grid-gap: 0 18px;
```
bduffany, line 363: uber nit: we usually keep things in multiples of 8px -- would this look OK at 16px?

```suggestion
  grid-gap: 0 16px;
```

app/invocation/cache_requests_card.tsx
```diff
@@ -368,7 +371,62 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
     ));
   }
 
+  private getCacheMetadata(scorecardResult: cache.ScoreCard.IResult) {
+    const digest = scorecardResult.digest;
+    const remoteInstanceName = this.props.model.getRemoteInstanceName();
+    rpc_service.service
+      .getCacheMetadata(
+        new cache.GetCacheMetadataRequest({
```
bduffany, line 379: nit: you can omit `new cache.GetCacheMetadataRequest(...)` since the proto library accepts a regular old object, and it will still perform the same type-checking.

app/invocation/cache_requests_card.tsx
```diff
@@ -368,7 +371,62 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
...
+          return {
+            digestToCacheMetadata: prevState.digestToCacheMetadata,
+          };
+        });
+      });
+  }
+
   private renderResultHovercard(result: cache.ScoreCard.IResult, startTimeMillis: number) {
+    const digest = result.digest.hash;
+    if (!this.state.digestToCacheMetadata.has(digest)) {
```
bduffany, line 410: It's usually a best practice for `render()` to not have side effects, but there's not really a good alternative atm. Mind adding this TODO?

```suggestion
    // TODO(bduffany): Add an `onHover` prop to <Tooltip> and move this logic there
    if (!this.state.digestToCacheMetadata.has(digest)) {
```

app/invocation/cache_requests_card.tsx
```diff
@@ -368,7 +371,62 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
...
+          },
+        })
+      )
+      .then((response) => {
+        this.setState(function (prevState, props) {
+          prevState.digestToCacheMetadata.set(digest.hash, response);
+          return {
+            digestToCacheMetadata: prevState.digestToCacheMetadata,
+          };
+        });
```
bduffany, line 393: optional nit: for this sort of state update (where you aren't changing a state key but still want the component to re-render), I usually prefer to mutate the state then call `forceUpdate()` to inform React about the state update. I feel like the alternative of calling `setState()` and setting a state key equal to itself is less clear.

```suggestion
        this.state.digestToCacheMetadata.set(digest.hash, response);
        this.forceUpdate();
```

app/invocation/cache_requests_card.tsx
```diff
@@ -368,7 +371,62 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
...
+      .catch((e) => {
+        console.log("Could not fetch metadata: " + BuildBuddyError.parse(e));
+
+        // Set an empty struct in the map, so FE doesn't keep trying to fetch an invalid result
+        this.setState(function (prevState, props) {
+          prevState.digestToCacheMetadata.set(digest.hash, null);
+          return {
+            digestToCacheMetadata: prevState.digestToCacheMetadata,
+          };
+        });
```
bduffany, line 404: nit: if you move this logic to just before you begin the fetch, that will also fix another issue that if you hover again while a request is in progress, you could potentially fire a duplicate request for the same digest.

also same comment as above re. setState() vs. forceUpdate()

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: No it doesn't

server/buildbuddy_server/buildbuddy_server.go
maggie-lou: Yeah when I refactor things to use the ResourceName struct I'll clean that up too. We also have a third file that has a cache type enum but the numbers are switched... Very confusing...

app/invocation/cache_requests_card.tsx
maggie-lou: Nice - done

app/invocation/cache_requests_card.tsx
maggie-lou: Done

app/invocation/cache_requests_card.tsx
maggie-lou: Done

app/invocation/cache_requests_card.tsx
maggie-lou: Done

app/invocation/invocation.css
maggie-lou: Oh yeah totally

maggie-lou:
I will call out though that because the RPC calls are happening in real time, the timestamps will give the current last access / modify time, not the timestamps at the point the invocation was created

bduffany:
> I will call out though that because the RPC calls are happening in real time, the timestamps will give the current last access / modify time, not the timestamps at the point the invocation was created

I think eventually we'd want to move this information to a separate "digest" page which shows info about an arbitrary digest (not tied to an invocation), and then we can have a section which shows the current cache metadata for the digest (if it still exists in cache).

But for now it's nice to at least surface this somewhere

bduffany (approved):
nice!

vadimberezniker (approved):

PR #2425, opened 2022-08-22
filestory

tylerwilliams:
- Refactor: straight up move of file functions from constants -> filestore.
- Include groupID in fileRecord proto
- Allow configuring groupID in path or not
- Fixup filestore callsites in pebble_cache
- Fixup filestore callsites in raft_cache
- Buildfix everything
- Remove build target for dead test
- Fix store test callsites.


enterprise/server/raft/filestore/filestore.go
```diff
@@ -35,7 +34,111 @@ type WriteCloserMetadata interface {
...
+
+	FileReader(ctx context.Context, fileDir string, f *rfpb.StorageMetadata_FileMetadata, offset, limit int64) (io.ReadCloser, error)
+	FileWriter(ctx context.Context, fileDir string, fileRecord *rfpb.FileRecord) (WriteCloserMetadata, error)
+}
+
+type fileStorer struct {
+	includeGroupIDInFilePaths bool
+}
+
+func New(includeGroupIDInFilePaths bool) Store {
```
vadimberezniker, line 85: What do you think about calling it just `useGroupIDs` or `isolateByGroupIDs` as this opt affects keys as well?

enterprise/server/raft/filestore/filestore.go
```diff
@@ -35,7 +34,111 @@ type WriteCloserMetadata interface {
 	MetadataWriter
 }
 
-func NewWriter(ctx context.Context, fileDir string, wb pebble.Writer, fileRecord *rfpb.FileRecord) (WriteCloserMetadata, error) {
+// returns partitionID, isolation, hash or
```
vadimberezniker, line 37: could use a a better comment. should make it clear that we always return the same elements but one of them can be empty depending on isolation type.

vadimberezniker (approved):

enterprise/server/raft/filestore/filestore.go
tylerwilliams: Done.

enterprise/server/raft/filestore/filestore.go
tylerwilliams: Done.

PR #2426, opened 2022-08-23
Make random digest generator usable outside of tests.

vadimberezniker:
Once in a while I need to generate random digests in a one-off testing
tool and it'd be nice to have this functionality easily accessible.

---

**Version bump**: None

### DIFF
--- a/server/testutil/testdigest/testdigest.go
+++ b/server/testutil/testdigest/testdigest.go
@@ -3,7 +3,6 @@ package testdigest
 import (
 	"bytes"
 	"io"
-	"io/ioutil"
 	"math/rand"
 	"sync"
 	"testing"
@@ -15,7 +14,7 @@ import (
 
 var (
 	randomSeedOnce sync.Once
-	randomSrc      io.Reader
+	randomGen      *Generator
 )
 
 type randomDataMaker struct {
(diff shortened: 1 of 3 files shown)


server/testutil/testdigest/testdigest.go
```diff
@@ -39,28 +38,58 @@ func (r *randomDataMaker) Read(p []byte) (n int, err error) {
 	}
 }
 
-func NewRandomDigestReader(t testing.TB, sizeBytes int64) (*repb.Digest, io.ReadSeeker) {
-	randomSeedOnce.Do(func() {
-		randomSrc = &randomDataMaker{rand.NewSource(time.Now().Unix())}
-	})
+type Generator struct {
+	randMaker *randomDataMaker
+}
```
tylerwilliams, line 43: can we move this to a package that is not in a testing package? I'd like to keep those testonly while also being able to use this generator in other tools (like smash, which I think has a copy of this same code?)

tylerwilliams (approved):

server/testutil/testdigest/testdigest.go
vadimberezniker: Done.

tylerwilliams (approved):

PR #2428, opened 2022-08-23
Task sizing: Add support for estimating task sizes using Tensorflow model

bduffany:
Adds a package that will enable using a basic linear regression model for task sizing.

The package is not used anywhere yet.

---

**Version bump**: Patch


enterprise/server/tasksize_model/tasksize_model.go
```diff
@@ -0,0 +1,245 @@
+package tasksize_model
```
tylerwilliams, line 1: This is *really* cool, but I'd much prefer to just use a straight up tensorflow lib for doing this rather than our own so that if we make this model better later we are well supported -- can we try that?

bduffany: Done - switched to a TF model. Each prediction takes around 10 usec. I will check in the tooling for generating the TF repo (currently hosted at https://github.com/bduffany/go-tensorflow) in a separate PR

tylerwilliams (approved):

enterprise/server/tasksize_model/tasksize_model_darwin.go
```diff
@@ -0,0 +1,12 @@
+//go:build darwin && !ios
+// +build darwin,!ios
+
+package tasksize_model
+
+import (
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+)
+
+func New() (Model, error) {
```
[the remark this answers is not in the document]
bduffany: Done

tylerwilliams (approved):

PR #2430, opened 2022-08-23
Verify range generation when processing data writes on replicas.

vadimberezniker:
---

**Version bump**: None


enterprise/server/raft/sender/sender.go
```diff
@@ -218,6 +205,42 @@ func (s *Sender) Run(ctx context.Context, key []byte, fn runFunc) error {
 	return status.UnavailableError("sender.Run retries exceeded")
 }
 
+func (s *Sender) RunAll(ctx context.Context, key []byte, fn runAllFunc) error {
```
tylerwilliams, line 208: Could you add a comment to this about what it is and what it's useful for?

enterprise/server/raft/store/store.go
```diff
@@ -299,6 +299,37 @@ func (s *Store) RemoveRange(rd *rfpb.RangeDescriptor, r *replica.Replica) {
...
+	// force client to go back and re-pull the rangeDescriptor from the meta
+	// range.
+	if rd.GetGeneration() != header.GetGeneration() {
+		return nil, status.OutOfRangeErrorf("%s: generation: %d requested: %d", constants.RangeNotCurrentMsg, rd.GetGeneration(), header.GetGeneration())
+	}
+
+	return rd, nil
+}
+
+func (s *Store) rangeIsValid(header *rfpb.Header) error {
```
tylerwilliams, line 328: nit: could you add comments explaining what Valid is (in contrast to Active) and when it should be used?

tylerwilliams (changes requested):

tylerwilliams:
(This LGTM if you add a couple of comments)

tylerwilliams (approved):

PR #2432, opened 2022-08-23
Fix replica usage calculation.

vadimberezniker:
---

**Version bump**: None

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -117,7 +117,13 @@ func batchLookup(wb *pebble.Batch, query []byte) ([]byte, error) {
 	return val, nil
 }
 
-func sizeOf(val []byte) (int64, error) {
+func sizeOf(key []byte, val []byte) (int64, error) {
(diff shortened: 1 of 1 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -147,8 +153,8 @@ func (sm *Replica) Usage() (*rfpb.ReplicaUsage, error) {
 	defer iter.Close()
 
 	estimatedBytesUsed := int64(0)
-	for iter.Next() {
-		sizeBytes, err := sizeOf(iter.Value())
+	for iter.First(); iter.Valid(); iter.Next() {
```
tylerwilliams, line 156: are there other places we should fix this?

tylerwilliams (approved):

enterprise/server/raft/replica/replica.go
vadimberezniker: https://github.com/buildbuddy-io/buildbuddy/pull/2433 was the only other misuse I found

PR #2439, opened 2022-08-24
Add a script to automate dev QA testing

bduffany:
Runs the test builds that we usually run, but also handles cloning the repos if not already cloned, as well as printing all the invocation links after all build commands are run so that they can be easily pasted into Slack.

---

**Version bump**: Patch


tylerwilliams (approved):
neat :)

luluz66 (approved):

PR #2441, opened 2022-08-24
Firecracker test fixes

tempoz:
Changed test jailer root to be a symbolic link into the bazel test temp directory. This accomplishes two things:

1. Allows hard links between the jailer root and other files in the bazel test temp directory in the case that `/tmp` and the bazel test temp directory are not on the same filesystem.

2. Uses the bazel test temp directory for the files in the jailer root, and thus not potentially risking running the temp filesystem out of disk space.

Had to slightly alter the locateBinaries logic to accommodate this, as it did not previously account for the possibility of different `JailerRoot` directories.

Fixed `dialVMExecServer` to correctly forward a `DeadlineExceededError` when the error it hits is an exceeded deadline.

---

**Version bump**: Patch


tempoz:
* master
  * **PR #2441** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2441" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2441?utm_source=stack-comment).

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1321,6 +1322,7 @@ func (c *FirecrackerContainer) dialVMExecServer(ctx context.Context) (*grpc.Clie
 			if err := c.parseFatalInitError(); err != nil {
 				return nil, err
 			}
+			return nil, status.DeadlineExceededErrorf("Failed to connect to firecracker VM exec server: %s", err)
```
bduffany, line 1325:
```suggestion
			// Intentionally not returning DeadlineExceededError here since it is not
			// a Bazel-retryable error, but this particular timeout should be retryable.
```

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -893,7 +893,8 @@ func copyStaticFiles(ctx context.Context, env environment.Env, workingDir string
 	ctx, span := tracing.StartSpan(ctx)
 	defer span.End()
 
-	locateBinariesOnce.Do(func() {
+	locateBinariesOnce, _ := locateBinariesOnceMap.LoadOrStore(workingDir, &sync.Once{})
```
bduffany, line 896: not sure I understand why the locateBinariesOnce had to be changed to a sync.Map -- are there tests that use multiple values of workingDir ? if not, my (possibly flawed) understanding was that each test should be run as its own process, so the sync.Once would be re-evaluated each time for the workingDir unique to each test

tempoz: No, `sync.Once` only runs once overall, not once per test. I initially was under the same misapprehension.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
tempoz: So `InternalError` is retryable? I changed this because locally [this test](https://github.com/buildbuddy-io/buildbuddy/blob/09449bd7ae68739d91b46188da851735d83e79ae/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go#L855-L857) wasn't passing, and was returning an `InternalError` that said a `ctx` deadline was exceeded, any idea how best to fix that?

tempoz: aha, lengthened a timeout in the test and it fixed everything.

bduffany (approved):

PR #2454, opened 2022-08-25
Scan GT when looking up ranges

tylerwilliams:
This fixes an issue where the wrong range was looked up.

image

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -757,7 +757,12 @@ func (sm *Replica) scan(db ReplicaReader, req *rfpb.ScanRequest) (*rfpb.ScanResp
 	case rfpb.ScanRequest_SEEKGE_SCAN_TYPE:
 		t = iter.SeekGE(req.GetLeft())
 	case rfpb.ScanRequest_SEEKGT_SCAN_TYPE:
-		t = iter.SeekGE(keys.Key(req.GetLeft()).Next())
+		t = iter.SeekGE(req.GetLeft())
+		// If the iter's current key is *equal* to left, go to the next
+		// key greater than this one.
+		if bytes.Compare(iter.Key(), req.GetLeft()) == 0 {
+			t = iter.Next()
+		}
 	default:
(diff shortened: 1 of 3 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -756,6 +756,8 @@ func (sm *Replica) scan(db ReplicaReader, req *rfpb.ScanRequest) (*rfpb.ScanResp
 		t = iter.SeekLT(req.GetLeft())
 	case rfpb.ScanRequest_SEEKGE_SCAN_TYPE:
 		t = iter.SeekGE(req.GetLeft())
+	case rfpb.ScanRequest_SEEKGT_SCAN_TYPE:
+		t = iter.SeekGE(keys.Key(req.GetLeft()).Next())
```
vadimberezniker, line 760: I think this is wrong. You only want to get the next key if the key the iterator points to is equal to the target key.

vadimberezniker (changes requested):

enterprise/server/raft/replica/replica.go
tylerwilliams: PTAL

enterprise/server/raft/replica/replica.go
```diff
@@ -756,6 +756,13 @@ func (sm *Replica) scan(db ReplicaReader, req *rfpb.ScanRequest) (*rfpb.ScanResp
 		t = iter.SeekLT(req.GetLeft())
 	case rfpb.ScanRequest_SEEKGE_SCAN_TYPE:
 		t = iter.SeekGE(req.GetLeft())
+	case rfpb.ScanRequest_SEEKGT_SCAN_TYPE:
+		t = iter.SeekGE(req.GetLeft())
+		// If the iter's current key is *equal* to left, go to the next
+		// key greater than this one.
+		if bytes.Compare(iter.Key(), req.GetLeft()) == 0 {
```
vadimberezniker, line 763: I would change this to `t && bytes.Compare` for the reader's sake even though this is not technically wrong.

PR #2456, opened 2022-08-26
Fix overlap check.

vadimberezniker:
For the right index, we find the first range that does *not* overlap the
range being tested. If the right index is 0 then it means there are no
ranges that overlap.

---

**Version bump**: None


tylerwilliams (approved):
Better by far :) Thanks for looking at this.

PR #2457, opened 2022-08-26
Fix store test a little bit

tylerwilliams:
- Always fetch the range descriptor from the store being fetched from.
- Construct a header that contains generation.

### DIFF
--- a/enterprise/server/raft/store/store_test.go
+++ b/enterprise/server/raft/store/store_test.go
@@ -403,17 +403,17 @@ func writeRecord(ctx context.Context, t *testing.T, stores []*TestingStore, grou
 	fk, err := filestore.New(true /*=isolateByGroupIDs*/).FileMetadataKey(fr)
 	require.Nil(t, err)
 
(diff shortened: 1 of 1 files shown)


enterprise/server/raft/store/store_test.go
```diff
@@ -403,17 +403,17 @@ func writeRecord(ctx context.Context, t *testing.T, stores []*TestingStore, grou
 	fk, err := filestore.New(true /*=isolateByGroupIDs*/).FileMetadataKey(fr)
 	require.Nil(t, err)
 
-	rd, err := stores[0].Sender.LookupRangeDescriptor(ctx, fk, true /*skipCache*/)
-	require.Nil(t, err, err)
-	rangeID := rd.GetRangeId()
-
 	for _, ts := range stores {
+		rd, err := ts.Sender.LookupRangeDescriptor(ctx, fk, true /*skipCache*/)
+		require.Nil(t, err, err)
```
vadimberezniker, line 408: There's also `require.NoError` if you don't mind using that instead.

vadimberezniker (approved):

PR #2461, opened 2022-08-26
add seperate flags for reading and writing from clickhouse

luluz66:
So that we can enable double write first; then enable reading from
clickhouse after the data is backfilled.

### DIFF
--- a/enterprise/server/invocation_stat_service/invocation_stat_service.go
+++ b/enterprise/server/invocation_stat_service/invocation_stat_service.go
@@ -63,7 +63,7 @@ func (i *InvocationStatService) getAggColumn(reqCtx *ctxpb.RequestContext, aggTy
 
 func (i *InvocationStatService) GetTrendBasicQuery(timezoneOffsetMinutes int32) string {
 	q := ""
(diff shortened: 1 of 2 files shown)


enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -58,7 +63,7 @@ func (i *InvocationStatService) getAggColumn(reqCtx *ctxpb.RequestContext, aggTy
 
 func (i *InvocationStatService) GetTrendBasicQuery(timezoneOffsetMinutes int32) string {
 	q := ""
-	if i.olapdbh == nil {
+	if i.olapdbh == nil || !*readFromOLAPDBEnabled {
```
tylerwilliams, line 66: can you pull this out to a function

func (i *ISS) OLAPEnabled() bool {
  return i.olapdbh != nil && *readFromOLAPDBEnabled
}

and then use this everywhere you have the || now?

luluz66: Done

tylerwilliams (approved):

PR #2462, opened 2022-08-27
formalized leaser

tylerwilliams:
Add a utility that allows wrapping a pebble DB and "leasing" handles to it.

### DIFF
--- a/enterprise/server/util/pebble/pebble.go
+++ b/enterprise/server/util/pebble/pebble.go
@@ -1,0 +1,153 @@
+package pebble
+
+import (
+	"io"
+	"flag"
+	"fmt"
+	"runtime"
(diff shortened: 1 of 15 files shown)


enterprise/server/util/pebble/pebble.go
```diff
@@ -0,0 +1,153 @@
+package pebble
```
vadimberezniker, line 1: can we call this pebbleutil and avoid needing to alias packages?

vadimberezniker (approved):

enterprise/server/util/pebble/pebble.go
tylerwilliams: Sure, done.

PR #2470, opened 2022-08-29
For public invocations, add (as opposed to replace) others_read permissions

maggie-lou:
Fixes buildbuddy-io/buildbuddy-internal#1589
   
When we create an invocation, we are setting the permissions based on the context (https://github.com/buildbuddy-io/buildbuddy/blob/master/server/backends/invocationdb/invocationdb.go#L91). These are
being overwritten for workflows for public repos. This may
incorrectly restrict the invocation, if we had previously set additional permissions on it

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -1253,11 +1253,13 @@ func (e *EventChannel) tableInvocationFromProto(p *inpb.Invocation, blobID strin
 	i.Attempt = p.Attempt
 	i.BazelExitCode = p.BazelExitCode
 
-	userGroupPerms, err := perms.GetFromContext(e.ctx, e.env)
+	userGroupPerms, err := perms.ForAuthenticatedGroup(e.ctx, e.env)
 	if err != nil {
-		return nil, err
+		// TODO(Maggie): Return the error here once we're confident this is stable
+		log.Warningf("Error fetching group perms for invocation %v", p.InvocationId)
+	} else {
+		i.Perms = userGroupPerms.Perms
 	}
-	i.Perms = userGroupPerms.Perms
 	if p.ReadPermission == inpb.InvocationPermission_PUBLIC {
 		i.Perms |= perms.OTHERS_READ
 	}
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -625,7 +625,7 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 		invocation.LastChunkId = e.logWriter.GetLastChunkId(ctx)
 	}
 
-	ti, err := tableInvocationFromProto(invocation, iid)
+	ti, err := e.tableInvocationFromProto(invocation, iid)
 	if err != nil {
 		return err
 	}
@@ -1024,7 +1024,7 @@ func (e *EventChannel) writeBuildMetadata(ctx context.Context, invocationID stri
 	if e.logWriter != nil {
 		invocationProto.LastChunkId = e.logWriter.GetLastChunkId(ctx)
 	}
-	ti, err = tableInvocationFromProto(invocationProto, ti.BlobID)
+	ti, err = e.tableInvocationFromProto(invocationProto, ti.BlobID)
 	if err != nil {
 		return err
 	}
@@ -1221,7 +1221,7 @@ func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*in
 	return invocation, nil
 }
 
-func tableInvocationFromProto(p *inpb.Invocation, blobID string) (*tables.Invocation, error) {
+func (e *EventChannel) tableInvocationFromProto(p *inpb.Invocation, blobID string) (*tables.Invocation, error) {
 	uuid, err := uuid.StringToBytes(p.InvocationId)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


maggie-lou:
@bduffany I see that workflows are created with group_read and group_write permissions by default (https://github.com/buildbuddy-io/buildbuddy/blob/697fc0dc160bd0bbff59e72cb802ce702af1f466/enterprise/server/workflow/service/service.go#L163). I'm assuming this gets translated onto the invocation object, but just cannot find the code that is doing that. Any chance you know where that is?

bduffany:
> @bduffany I see that workflows are created with group_read and group_write permissions by default (
> 
> https://github.com/buildbuddy-io/buildbuddy/blob/697fc0dc160bd0bbff59e72cb802ce702af1f466/enterprise/server/workflow/service/service.go#L163
> 
> ). I'm assuming this gets translated onto the invocation object, but just cannot find the code that is doing that. Any chance you know where that is?

@maggie-lou I think we only set the invocation table Perms once, in CreateInvocation (the invocation doesn't inherit perms from the workflow and it's purely based on whether the workflow is authenticated, which it should be for all workflows).

So because we aren't keeping track of these perms anywhere, when we call TableInvocationFromProto, the Perms only ever wind up being 0 or OTHERS_READ. This doesn't cause problems for non-public invocations because when calling UpdateInvocation, zero-valued fields are not actually applied as updates to the invocation row (because of GORM).

So I think in this TableInvocationFromProto we might need to do the same thing that we do in CreateInvocation, and check whether there is an authenticated group ID in the ctx. If so, we can apply GROUP_READ|GROUP_WRITE permissions. Then on top of that we can add on OTHERS_READ if the invocation is public.

server/util/perms/perms.go
```diff
@@ -290,3 +290,23 @@ func AuthenticatedGroupID(ctx context.Context, env environment.Env) (string, err
 func IsAnonymousUserError(err error) bool {
 	return status.IsUnauthenticatedError(err) || status.IsPermissionDeniedError(err) || status.IsUnimplementedError(err)
 }
+
+func GetFromContext(ctx context.Context, env environment.Env) (*UserGroupPerm, error) {
```
bduffany, line 294: The name `GetFromContext` feels a bit too general since there might be cases where it makes sense to set other perms that are derived from context, like USER_READ|USER_WRITE etc.

wdyt about renaming this to "ForAuthenticatedGroup", and adding a comment like

`ForAuthenticatedGroup returns GROUP_READ|GROUP_WRITE permissions for the authenticated group, or OTHERS_READ for anonymous users.`

server/util/capabilities/capabilities.go
```diff
@@ -57,3 +57,19 @@ func IsGranted(ctx context.Context, env environment.Env, cap akpb.ApiKey_Capabil
 	}
 	return user.HasCapability(cap), nil
 }
+
+func GetFromContext(ctx context.Context, env environment.Env) ([]akpb.ApiKey_Capability, error) {
```
bduffany, line 61: I think a better name might be `ForAuthenticatedUser` (so the callsite reads `capablities.ForAuthenticatedUser`) since it might be a bit unclear how we are figuring out the capabilities from the context.

server/util/capabilities/capabilities.go
```diff
@@ -57,3 +57,19 @@ func IsGranted(ctx context.Context, env environment.Env, cap akpb.ApiKey_Capabil
...
+	if auth := env.GetAuthenticator(); auth != nil {
+		u, err := auth.AuthenticatedUser(ctx)
+		if err == nil {
+			caps = u.GetCapabilities()
+		} else {
+			if perms.IsAnonymousUserError(err) && auth.AnonymousUsageEnabled() {
+				caps = DefaultAuthenticatedUserCapabilities
+			}
+		}
+	}
```
bduffany, line 72: nit: can use early returns here to reduce nesting, and also return an error in the case where there is a legitimate auth error. (In the invocationdb callsite, you can ignore the returned error, since that's what the previous code was doing)

```go
auth := env.GetAuthenticator()
if auth == nil {
  return nil, status.UnimplementedError("auth is not configured")
}
u, err := auth.AuthenticatedUser(ctx)
if err != nil {
  if perms.IsAnonymousUserError(err) && auth.AnonymousUsageEnabled() {
    return DefaultAuthenticatedUserCapabilities, nil
  }
  return nil, err
}
return u.GetCapabilities(), nil
```

server/util/perms/perms.go
```diff
@@ -290,3 +290,23 @@ func AuthenticatedGroupID(ctx context.Context, env environment.Env) (string, err
...
+		}
+	}
+
+	if permissions == nil && env.GetAuthenticator().AnonymousUsageEnabled() {
+		permissions = AnonymousUserPermissions()
+	} else if permissions == nil {
+		return nil, status.PermissionDeniedErrorf("Anonymous access disabled, permission denied.")
+	}
+
+	return permissions, nil
```
bduffany, line 311: (similar note as the Capabilities func re. reduced nesting, and returning an error if there's a legitimate auth error)

bduffany (commented):
mostly LGTM! just some style nits

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -1248,13 +1248,19 @@ func tableInvocationFromProto(p *inpb.Invocation, blobID string) (*tables.Invoca
...
-	}
 	i.LastChunkId = p.LastChunkId
 	i.RedactionFlags = redact.RedactionFlagStandardRedactions
 	i.Attempt = p.Attempt
 	i.BazelExitCode = p.BazelExitCode
+
+	userGroupPerms, err := perms.GetFromContext(e.ctx, e.env)
+	if err != nil {
+		return nil, err
+	}
```
bduffany, line 1259: nit: I think it'd be less risky to first log a warning here to verify that we aren't getting any unexpected errors here, and have a TODO to return the error once we're more confident.

maggie-lou:
* master
  * **PR #2470** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2470" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2470?utm_source=stack-comment).

server/util/perms/perms.go
maggie-lou: Ack - done

server/util/capabilities/capabilities.go
maggie-lou: Ack - done

server/util/capabilities/capabilities.go
maggie-lou: Ack - done

server/util/perms/perms.go
maggie-lou: Ack - done

server/build_event_protocol/build_event_handler/build_event_handler.go
maggie-lou: Ack - done

bduffany (approved):

PR #2474, opened 2022-08-30
Refactor pebble cache to use options struct instead of flags

maggie-lou:
buildbuddy-io/buildbuddy#2416

Because we want to migrate between 2 pebble caches, we need to make sure the logic uses the configurable 
data for the given cache. This is the pre-requisite for a larger change where we'll allow setting up multiple
pebble caches for a migration

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -95,11 +95,18 @@ const (
 // Options is a struct containing the pebble cache configuration options.
 // Once a cache is created, the options may not be changed.
 type Options struct {
-	RootDirectory       string
-	Partitions          []disk.Partition
-	PartitionMappings   []disk.PartitionMapping
-	MaxSizeBytes        int64
-	BlockCacheSizeBytes int64
+	RootDirectory     string
+	Partitions        []disk.Partition
+	PartitionMappings []disk.PartitionMapping
+
+	MaxSizeBytes           int64
+	BlockCacheSizeBytes    int64
+	MaxInlineFileSizeBytes int64
+
+	AtimeUpdateThreshold time.Duration
+	AtimeWriteBatchSize  int
+	AtimeBufferSize      int
+	MinEvictionAge       time.Duration
 }
 
 type sizeUpdate struct {
@@ -166,11 +173,16 @@ func Register(env environment.Env) error {
 		}
 	}
 	opts := &Options{
-		RootDirectory:       *rootDirectory,
-		Partitions:          *partitions,
-		PartitionMappings:   *partitionMappings,
-		BlockCacheSizeBytes: *blockCacheSizeBytes,
-		MaxSizeBytes:        cache_config.MaxSizeBytes(),
+		RootDirectory:          *rootDirectory,
+		Partitions:             *partitions,
+		PartitionMappings:      *partitionMappings,
+		BlockCacheSizeBytes:    *blockCacheSizeBytes,
+		MaxSizeBytes:           cache_config.MaxSizeBytes(),
+		MaxInlineFileSizeBytes: *maxInlineFileSizeBytes,
+		AtimeUpdateThreshold:   *atimeUpdateThreshold,
+		AtimeWriteBatchSize:    *atimeWriteBatchSize,
+		AtimeBufferSize:        *atimeBufferSize,
+		MinEvictionAge:         *minEvictionAge,
 	}
 	c, err := NewPebbleCache(env, opts)
 	if err != nil {
@@ -263,8 +275,8 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 	ensureDefaultPartitionExists(opts)
 
 	pebbleOptions := defaultPebbleOptions()
-	if *blockCacheSizeBytes > 0 {
-		c := pebble.NewCache(*blockCacheSizeBytes)
+	if opts.BlockCacheSizeBytes > 0 {
+		c := pebble.NewCache(opts.BlockCacheSizeBytes)
 		defer c.Unref()
 		pebbleOptions.Cache = c
 	}
@@ -303,7 +315,7 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 			if err := disk.EnsureDirectoryExists(blobDir); err != nil {
 				return err
 			}
-			pe, err := newPartitionEvictor(part, pc.fileStorer, blobDir, pc.leaser, pc.accesses)
+			pe, err := newPartitionEvictor(part, pc.fileStorer, blobDir, pc.leaser, pc.accesses, opts.AtimeBufferSize, opts.MinEvictionAge)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -95,11 +95,18 @@ const (
...
+	PartitionMappings []disk.PartitionMapping
+
+	MaxSizeBytes           int64
+	BlockCacheSizeBytes    int64
+	MaxInlineFileSizeBytes int64
+
+	AtimeUpdateThreshold time.Duration
+	AtimeWriteBatchSize  int
+	AtimeBufferSize      int
+	MinEvictionAge       time.Duration
```
tylerwilliams, line 109: I worry a little about partially filled options here -- wdyt?

Some options for fixing might be:
 - move the flag default values to consts, set anything in options to its proper default value if it's not set
 - leave it, just write a big scary comment
 - add some options validation code that barfs if it's a bad config

maggie-lou:
* master
  * **PR #2474** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2474" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2474?utm_source=stack-comment).

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Good call out - done

maggie-lou: Made options struct private and two constructors to make it explicit that there are two ways to create options - either everything gets set via flags, or non-nil fields are grabbed from the OptionsConstructor.

I added the OptionsConstructor bc it's valid for certain fields, like AtimeUpdateThreshold, to be 0, so we need pointers to differentiate whether a 0 value was explicitly set by the user, vs was just set as a default value

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -43,11 +43,11 @@ import (
...
 var (
-	rootDirectory          = flag.String("cache.pebble.root_directory", "", "The root directory to store the database in.")
-	blockCacheSizeBytes    = flag.Int64("cache.pebble.block_cache_size_bytes", 1000*megabyte, "How much ram to give the block cache")
-	maxInlineFileSizeBytes = flag.Int64("cache.pebble.max_inline_file_size_bytes", 1024, "Files smaller than this may be inlined directly into pebble")
-	partitions             = flagutil.New("cache.pebble.partitions", []disk.Partition{}, "")
-	partitionMappings      = flagutil.New("cache.pebble.partition_mappings", []disk.PartitionMapping{}, "")
+	rootDirectoryFlag          = flag.String("cache.pebble.root_directory", "", "The root directory to store the database in.")
+	blockCacheSizeBytesFlag    = flag.Int64("cache.pebble.block_cache_size_bytes", defaultBlockCacheSizeBytes, "How much ram to give the block cache")
+	maxInlineFileSizeBytesFlag = flag.Int64("cache.pebble.max_inline_file_size_bytes", defaultMaxInlineFileSizeBytes, "Files smaller than this may be inlined directly into pebble")
+	partitionsFlag             = flagutil.New("cache.pebble.partitionsFlag", []disk.Partition{}, "")
```
vadimberezniker, line 49: Was this flag name change accidental?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -90,16 +90,48 @@ const (
...
+
+	AtimeUpdateThreshold time.Duration
+	AtimeWriteBatchSize  int
+	AtimeBufferSize      int
+	MinEvictionAge       time.Duration
+}
+
+// OptionsConstructor contains fields required to initialize an options struct
+// If a field is nil, the corresponding field in the options struct will be set to the default value
+type OptionsConstructor struct {
```
vadimberezniker, line 122: Instead of renaming the public type, could we rename the internal type? Options/Opts is a common convention in Go for passing options to initialize a struct. Alternatively we could get rid of the internal options type and inline the fields into the cache struct.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -90,16 +90,48 @@ const (
...
+	RootDirectory     *string
```
vadimberezniker, line 123: Convention in Go is to use empty values (i.e. "" for string, 0 for integers) to indicate whether something is set. It's less common to use pointers.

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: The problem I was facing is that sometimes we explicitly want to set some of these values to 0. So I was having trouble detecting whether something was unset and should be set to the default, or was explicitly set to 0 and so we shouldn't overwrite it with the default value. Is there a more idiomatic way to handle that?

vadimberezniker: You can use pointers for the cases where 0 is a valid value, but I would recommend using plain values in all other cases.

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: I renamed the flags so that people don't accidentally reference the flags in the code. Before, we were using the flags to create an Options struct, but we were also just using the flags throughout the code, instead of referencing the fields from the Options struct

If we have two of the same cache type in a migration, we need to make sure we're using the options for the current cache

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Let me know if this is cleaner. Agree that it's confusing to have the internal and public options types, so inlined the options into the cache struct

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Updated. Do you find it confusing that certain fields are pointers but others are just the plain values?

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Ohhh 🤦‍♀️ I know what you're saying actually. Yes that was an accident. Will fix

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -57,10 +57,18 @@ var (
...
-	atimeBufferSize           = flag.Int("cache.pebble.atime_buffer_size", 100000, "Buffer up to this many atime updates in a channel before dropping atime updates")
-	minEvictionAge            = flag.Duration("cache.pebble.min_eviction_age", 6*time.Hour, "Don't evict anything unless it's been idle for at least this long")
+	atimeUpdateThresholdFlag  = flag.Duration("cache.pebble.atime_update_threshold", defaultAtimeUpdateThreshold, "Don't update atime if it was updated more recently than this")
+	atimeWriteBatchSizeFlag   = flag.Int("cache.pebble.atime_write_batch_size", defaultAtimeWriteBatchSize, "Buffer this many writes before writing atime data")
+	atimeBufferSizeFlag       = flag.Int("cache.pebble.atime_buffer_size", defaultAtimeBufferSize, "Buffer up to this many atime updates in a channel before dropping atime updates")
+	minEvictionAgeFlag        = flag.Duration("cache.pebble.min_eviction_age", defaultMinEvictionAge, "Don't evict anything unless it's been idle for at least this long")
+
+	// Default values for Options
+	defaultBlockCacheSizeBytes    = int64(1000 * megabyte)
```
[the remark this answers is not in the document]
maggie-lou: Some of them have to be vars so we can take their addresses for pointers, but moved the rest to const

PR #2480, opened 2022-08-30
Populate scorecard field in invocation webhook payload with action cache misses

bduffany:
After enabling detailed stats, we no longer populate the `score_card` field in invocation webhook payloads. This PR restores that functionality and also adds an integration test for invocation webhooks.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #2485, opened 2022-08-31
Support parsing objects from config yaml

maggie-lou:
For cache migrations, I thought it would be cleaner if:
1. We nested all migration cache config under the migration block.

i.e. 
Normally if you had a pebble cache, the config would look like:
cache
++ pebble

With a migration, it would look like:
cache
++migration
++++src
++++++pebble
++++dest
++++++pebble

Because we will now support migrating from two of the same cache type, nesting under the migration block lets us add another layer of `src` and `dest` keys to be more explicit about the direction of the migration. 

2. We parsed the cache config yaml as an object, as opposed to flags

Now that we have a new migration section of the config, it's kind of a pain to create a complete new set of flags for every cache type that's eligible for migration. 

i.e. 
Before, if we have the option rootDir for pebble cache, we'd have the flag cache.pebble.rootDir

Now we would need
- cache.pebble.RootDir for pebble caches not in a migration
- cache.migration.src.pebble.RootDir for source pebble caches in a migration
- cache.migration.dest.pebble.RootDir for dest pebble caches in a migration
- ... etc. for all cache options

It feels more readable and less complex to me to parse this new migration config as an object

Here's an example of how we would then use this new config object to initialize a migration_cache: https://github.com/buildbuddy-io/buildbuddy/pull/2486/files#diff-3c25189bbd67311458979d8b7dcacba95332e9afd3754e23966df0388287be9cL23

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -20,16 +20,10 @@ var (
 	// If you wish to extract a section of the config to the Config struct as opposed to parsing into flags:
 	// 1. Set the section name in configKeysToExtract so the flag parser will know to ignore it
 	// 2. Define the section in the Config struct using `yaml` tags
-	configKeysToExtract = []string{"migration"}
+	configKeysToExtract []string
 )
 
-type Config struct {
-	CacheBlock *CacheBlock `yaml:"cache"`
-}
-
-type CacheBlock struct {
-	MigrationConfig *MigrationConfig `yaml:"migration"`
-}
+type Config struct{}
 
 func init() {
 	// As this flag determines the YAML file we read the config from, it can't
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -3,19 +3,72 @@ package config
 import (
 	"flag"
 
+	"github.com/buildbuddy-io/buildbuddy/server/util/file"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+
 	flagyaml "github.com/buildbuddy-io/buildbuddy/server/util/flagutil/yaml"
 )
 
 const pathFlagName = "config_file"
 
-var configPath = flag.String(pathFlagName, "/config.yaml", "The path to a buildbuddy config file")
+var (
+	configPath = flag.String(pathFlagName, "/config.yaml", "The path to a buildbuddy config file")
+
+	// If you wish to extract a section of the config to the Config struct as opposed to parsing into flags:
+	// 1. Set the section name in configKeysToExtract so the flag parser will know to ignore it
+	// 2. Define the section in the Config struct using `yaml` tags
+	configKeysToExtract = []string{"migration"}
+)
+
+type Config struct {
+	CacheBlock *CacheBlock `yaml:"cache"`
+}
+
+type CacheBlock struct {
+	MigrationConfig *MigrationConfig `yaml:"migration"`
+}
 
 func init() {
 	// As this flag determines the YAML file we read the config from, it can't
 	// meaningfully be specified in the YAML config file.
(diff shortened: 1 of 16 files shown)


maggie-lou:
* refactor_cache_options
  * **PR #2485** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2485" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2485?utm_source=stack-comment).

server/config/config.go
```diff
@@ -3,19 +3,66 @@ package config
...
 const pathFlagName = "config_file"
 
-var configPath = flag.String(pathFlagName, "/config.yaml", "The path to a buildbuddy config file")
+var (
+	configPath = flag.String(pathFlagName, "/config.yaml", "The path to a buildbuddy config file")
+
+	// If you wish to extract a section of the config to the Config struct as opposed to parsing into flags:
+	// 1. Set the section name in configKeysToExtract so the flag parser will know to ignore it
+	// 2. Define the section in the Config struct using `yaml` tags
+	configKeysToExtract []string
```
maggie-lou, line 23: Example of how this would be used with migration_cache: https://github.com/buildbuddy-io/buildbuddy/pull/2486/files

server/util/flagutil/yaml/yaml.go
```diff
@@ -543,39 +559,40 @@ func onSIGHUP(fn func() error) {
...
+func RereadFlagsOnDisconnect(configFile string) {
 	onSIGHUP(func() error {
-		log.Infof("Re-reading buildbuddy config from '%s'", configFile)
-		fileBytes, err := os.ReadFile(configFile)
-		if err != nil {
-			return fmt.Errorf("Error reading config file: %s", err)
-		}
-		return PopulateFlagsFromData(fileBytes)
+		log.Infof("Re-reading buildbuddy config %s", configFile)
+		return PopulateFlagsFromFile(configFile)
```
tempoz, line 594: We can't just re-ingest a config file because:

1. The values from the previous read of the config file will still be set, so if you remove a definition in the YAML it will not be the flag default and will instead be the old value.

1. Slice flags work by appending new values encountered instead of replacing them, so even if you have the same config file, all slice definitions will append their values to the flag again.

If you want to do this, we need to make a `Clear` function for flags to allow us to clear them, then re-parse flags, then re-ingest the config file.

tempoz (commented):
Why don't we just define a new `JSONStructFlag` so we can maintain the YAML and the flags as one united entity? It will require a lot fewer code changes and be more straightforward, I think.

server/config/config.go
bduffany: @tempoz might also have some thoughts, but would it be possible to instead add support [here](https://cs.github.com/buildbuddy-io/buildbuddy/blob/82a217bf5f5c64788ac57a038b290aba79733c0e/server/util/flagutil/types/autoflags/autoflags.go?q=path%3Aautoflags+repo%3Abuildbuddy-io%2Fbuildbuddy#L48-L52) for parsing arbitrary struct types? (Currently I think we only support struct slice types.) That way you can write something like

```
var migrationConfig = flagutil.New("cache.migration", (*MigrationConfig)(nil), "Cache migration config")
```

And then using reflection maybe it's possible to get that type `MigrationConfig` from the nil ptr, and parse using that struct type?

bduffany: woops, I see she already made the same comment :)

maggie-lou: 👍 https://github.com/buildbuddy-io/buildbuddy/pull/2487?no-redirect=1

maggie-lou: As a note though, Zoey taught me that you can't have a nil pointer as the default value for a flag

"flag definitions return pointers to default-initialized data that may be set by flag.Parse (and now also the yaml population code), and nil pointers do not point to data that can be set, so the flag could never be set if it defaulted to nil"

PR #2486, opened 2022-08-31
Add register function for migration cache

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/server/backends/migration_cache/config.go
+++ b/server/backends/migration_cache/config.go
@@ -1,4 +1,4 @@
-package config
+package migration_cache
 
 import (
 	"time"
@@ -7,8 +7,8 @@ import (
 )
 
 type MigrationConfig struct {
-	Src  CacheConfig `yaml:"src"`
-	Dest CacheConfig `yaml:"dest"`
+	Src  *CacheConfig `yaml:"src"`
+	Dest *CacheConfig `yaml:"dest"`
 }
 
 type CacheConfig struct {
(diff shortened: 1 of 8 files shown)


maggie-lou:
* refactor_cache_options
  * **PR #2485** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2485" target="_blank">Graphite</a> 
    * **PR #2486** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2486" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2486?utm_source=stack-comment).

server/backends/migration_cache/migration_cache.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func validateCacheConfig(config CacheConfig) error {
+	numConfigs := 0
+	v := reflect.ValueOf(config)
+	for i := 0; i < v.NumField(); i++ {
+		if !v.Field(i).IsNil() {
+			numConfigs++
+		}
+	}
```
tylerwilliams, line 64: Kinda gnarly to use reflection here! Can you just write this code out?

server/backends/migration_cache/migration_cache.go
```diff
@@ -0,0 +1,184 @@
...
+	c.Start()
+	env.GetHealthChecker().RegisterShutdownFunction(func(ctx context.Context) error {
+		return c.Stop()
+	})
+	return c, nil
+}
+
+func (mc MigrationCache) WithIsolation(ctx context.Context, cacheType interfaces.CacheType, remoteInstanceName string) (interfaces.Cache, error) {
+	//TODO implement me
+	panic("implement me")
```
tylerwilliams, line 133: nit: I know this is placeholder code but typically we don't use panics in server code ever.

You could instead

```
return nil, status.UnimplementedError("not yet implemented")
```

tylerwilliams (changes requested):

server/backends/migration_cache/migration_cache.go
maggie-lou: Done

server/backends/migration_cache/migration_cache.go
maggie-lou: Yeah thought that would make it more generalizable, but agree that it's probably just over-engineering

server/backends/migration_cache/config.go
```diff
@@ -0,0 +1,37 @@
...
+
+type MigrationConfig struct {
+	Src  *CacheConfig `yaml:"src"`
+	Dest *CacheConfig `yaml:"dest"`
+}
+
+type CacheConfig struct {
+	DiskConfig   *DiskCacheConfig   `yaml:"disk"`
+	PebbleConfig *PebbleCacheConfig `yaml:"pebble"`
+}
```
tempoz, line 17: Idea: make this an interface that defines "CacheFromConfig() interfaces.Cache`, and then you don't have to validate the CacheConfig object.

tylerwilliams (approved):

server/backends/migration_cache/config.go
maggie-lou: Can you say more about how to do that in a way where we can still enforce the structure of what we're expecting of a DiskCacheConfig, for example?

maggie-lou: Going to merge this for now bc that will make some other development easier, but can come back and refactor this

tempoz: The idea is that you make `DiskCacheConfig` implement `CacheConfig`, and then whatever gets passed in as a `CacheConfig` is guaranteed to just be one concrete type of `CacheConfig`, so it doesn't need to be validated. Also if we want to add a new type later, it simply has to implement `CacheConfig` and there are no changes that need to be made elsewhere, either in validation or in `CacheConfig`'s (now non-existent) fields.

maggie-lou: I like the concept but does that work with the JSON struct reflection you helped me with last week? We'd need the yaml parser to figure out which concrete type the interface needs to be cast to right?

maggie-lou: I tried this, which is what I think you're suggesting: https://github.com/buildbuddy-io/buildbuddy/pull/2507/files

Getting errors like: 	panic: reflect.Set: value of type map[string]interface {} is not assignable to type migration_cache.CacheConfig

PR #2491, opened 2022-09-01
fix the if condition to check whether clickhouse is enabled

luluz66:


tylerwilliams (approved):
doh

enabled != disabled

tylerwilliams (changes requested):
Actually, can you switch these cases?

if OLAPenabled() {
 doOlapThing
 } else {
 doOldThing
}

luluz66:
> Actually, can you switch these cases?
> 
> if OLAPenabled() { doOlapThing } else { doOldThing }

Done.

PR #2496, opened 2022-09-06
Add compression support to GetFile API

bduffany:
The proto is copied from the remote execution proto, so that we can easily cast between the two types of protos. The field names are also the same, just for consistency (`acceptable_compressors` / `compressor`)

---

**Version bump**: Patch


tylerwilliams (approved):

bduffany:
I realized this is already supported by changing `/blobs/` to `/compressed-blobs/zstd/` in the URI field. Closing

PR #2498, opened 2022-09-06
Support dockerNetwork=standard

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -201,13 +201,14 @@ func (p *Provider) NewContainer(image string, options *PodmanOptions) container.
 }
 
 type PodmanOptions struct {
-	ForceRoot bool
-	User      string
-	Network   string
-	CapAdd    string
-	Devices   []container.DockerDeviceMapping
-	Volumes   []string
-	Runtime   string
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -281,8 +282,21 @@ func (c *podmanCommandContainer) getPodmanRunArgs(workDir string) []string {
 	} else if c.options.User != "" && userRegex.MatchString(c.options.User) {
 		args = append(args, "--user="+c.options.User)
 	}
-	if strings.ToLower(c.options.Network) == "off" {
-		args = append(args, "--network=none")
+	networkMode := c.options.DefaultNetworkMode
+	networkProp := strings.ToLower(c.options.Network)
```
tylerwilliams, line 286: can you switch directly on strings.ToLower(c.options.Network)?

tylerwilliams (approved):

vadimberezniker (approved):

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: Done

PR #2509, opened 2022-09-07
Task Sizing: Switch to TF serving for prediction

bduffany:
---

**Version bump**: Patch


enterprise/server/tasksize_model/tasksize_model.go
```diff
@@ -1,32 +1,309 @@
...
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	scpb "github.com/buildbuddy-io/buildbuddy/proto/scheduler"
+	tf_framework "github.com/buildbuddy-io/tensorflow-proto/tensorflow/core/framework"
+	tf "github.com/buildbuddy-io/tensorflow-proto/tensorflow_serving/apis"
+)
+
+var (
+	featuresConfigPath = flag.String("remote_execution.task_size_model.features_config_path", "", "Path pointing to features.json config file.")
+	servingAddress     = flag.String("remote_execution.task_size_model.serving_address", "", "Address pointing to TF server with saved Tensorflow models (cpu, mem).")
```
vadimberezniker, line 29: I might be missing some context. What was the motivation for serving the models from a separate server?

vadimberezniker (approved):

vadimberezniker (changes requested):
didn't mean to hit approve

enterprise/server/tasksize_model/tasksize_model.go
bduffany: Tyler and I decided that the build tooling changes were getting to be too complicated. On top of needing to generate the go-tensorflow repo (which might be high maintenance if we want to upgrade TF versions), we would've also had to create a separate build of the app for Linux which doesn't depend on TF, since we don't currently support dynamically linking `.so` libraries in the standalone binary (since everything needs to be in the embedded FS, but we can't link from that embedded FS at runtime).

enterprise/server/tasksize_model/tasksize_model.go
```diff
@@ -1,32 +1,309 @@
...
+	})
+	m.client = tf.NewPredictionServiceClient(conn)
+
+	// Run a test prediction to ensure the feature vector shape matches the
+	// model's expected input shape.
+	ctx := env.GetServerContext()
+	ctx, cancel := context.WithTimeout(ctx, initTimeout)
+	defer cancel()
+
+	x := m.featureVector(&repb.ExecutionTask{})
```
vadimberezniker, line 117: Is this check going to block app startup? would be better to allow the app to startup w/ the model disabled

vadimberezniker (changes requested):

enterprise/server/tasksize_model/tasksize_model.go
bduffany: I can update the PR that calls New() to log the returned error as log.Error instead of log.Fatal.

enterprise/server/tasksize_model/tasksize_model.go
```diff
@@ -1,32 +1,310 @@
...
+	if _, err := m.predict(ctx, memModelName, x); err != nil {
```
vadimberezniker, line 118: Can we do this check asynchronously to avoid blocking app startup on blocking RPCs?

vadimberezniker (approved):

enterprise/server/tasksize_model/tasksize_model.go
bduffany: Done.

PR #2512, opened 2022-09-08
Fix proto written when restoring digest from peer.

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):

tylerwilliams:
Nit: maybe you could modify the TestPostFactoSplit test to read the full file and verify it's the same (since I think this would be caught by it?) instead of only checking for the presence of the file metadata keys?

tylerwilliams (approved):

PR #2517, opened 2022-09-09
Document Prometheus metrics path and port

jonshea:
Added a line noting the path and port that Prometheus metrics are configured with. Also documented the `--monitoring_port` config flag.

**Version bump**: None


tylerwilliams (approved):
Thanks for the PR :)

PR #2519, opened 2022-09-10
Rough draft of adding a Commit method to write closers

tylerwilliams:


enterprise/config/buildbuddy.local.yaml
```diff
@@ -6,17 +6,17 @@ app:
   default_redis_target: "localhost:6379"
   enable_target_tracking: true
 database:
-  data_source: "sqlite3:///tmp/buildbuddy-enterprise.db"
+  data_source: "sqlite3:///tmp/buildbuddy-enterprise_${USER}.db"
```
tylerwilliams, line 9: lol i can revert this or move to another cl. too many people on lunchbox :P

enterprise/server/backends/distributed/distributed.go
```diff
@@ -995,6 +996,10 @@ func (c *Cache) Reader(ctx context.Context, d *repb.Digest, offset, limit int64)
...
 }
 
-func (c *Cache) Writer(ctx context.Context, d *repb.Digest) (io.WriteCloser, error) {
-	return c.multiWriter(ctx, d)
+func (c *Cache) Writer(ctx context.Context, d *repb.Digest) (interfaces.CommittedWriteCloser, error) {
+	mwc, err := c.multiWriter(ctx, d)
+	if err != nil {
+		return nil, err
+	}
+	return ioutil.AutoUpgradeCloser(mwc), nil
```
tylerwilliams, line 1004: note to self: should be able to remove this and just return an underlying cwc here

enterprise/server/composable_cache/composable_cache.go
```diff
@@ -242,9 +243,9 @@ func (c *ComposableCache) Writer(ctx context.Context, d *repb.Digest) (io.WriteC
 					}
 				},
 			}
-			return dw, nil
+			return ioutil.AutoUpgradeCloser(dw), nil
 		}
 	}
 
-	return innerWriter, nil
+	return ioutil.AutoUpgradeCloser(innerWriter), nil
```
tylerwilliams, line 250: remove these too

server/util/ioutil/ioutil.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if c.CloseFn != nil {
+		if err := c.CloseFn(); err != nil && firstErr == nil {
+			firstErr = err
+		}
+	}
+	return firstErr
+}
+
+func NewCustomCommitWriteCloser(w io.Writer) *CustomCommitWriteCloser {
```
tylerwilliams, line 96: remove prints; add comments about how to use with example

server/util/ioutil/ioutil.go
```diff
@@ -0,0 +1,127 @@
...
+		r.committed = true
+	}()
+	return r.WriteCloser.Close()
+}
+
+func (r *CloseCommitReplacer) Close() error {
+	return nil
+}
+
+func AutoUpgradeCloser(wc io.WriteCloser) interfaces.CommittedWriteCloser {
```
tylerwilliams, line 119: nit: add comments about why this is necessary and when it should be used

server/util/disk/disk.go
```diff
@@ -196,15 +196,20 @@ func (w *writeMover) Write(p []byte) (int, error) {
...
 	tmpName := w.File.Name()
 	if err := w.File.Close(); err != nil {
 		return err
 	}
 	return os.Rename(tmpName, w.finalPath)
 }
 
-func FileWriter(ctx context.Context, fullPath string) (io.WriteCloser, error) {
+func (w *writeMover) Close() error {
+	DeleteLocalFileIfExists(w.File.Name())
```
vadimberezniker, line 208: we should close the temp file handle here if it's still open

server/util/ioutil/ioutil.go
vadimberezniker, line 119: I don't think we need this utility. It looks like we should be able to replace all usages of `AutoUpgradeCloser` by updating the custom writers with Commit method.

vadimberezniker (changes requested):

enterprise/server/backends/gcs_cache/gcs_cache.go
```diff
@@ -473,10 +474,15 @@ func (wc *gcsDedupingWriteCloser) Write(in []byte) (int, error) {
 	return n, err
 }
 
-func (wc *gcsDedupingWriteCloser) Close() error {
+func (wc *gcsDedupingWriteCloser) Commit() error {
 	return swallowGCSAlreadyExistsError(wc.WriteCloser.Close())
```
maggie-lou, line 478: wc.WriteCloser.Commit? And wc.WriteCloser.Close should be added to line 481?

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -499,6 +499,24 @@ func (d *doubleWriter) Write(data []byte) (int, error) {
 	return srcN, srcErr
 }
 
+func (d *doubleWriter) Commit() error {
+	eg := &errgroup.Group{}
+	if d.dest != nil {
+		eg.Go(func() error {
+			dstErr := d.dest.Commit()
+			if dstErr != nil {
+				log.Warningf("Migration writer close err: %s", dstErr)
```
maggie-lou, line 508: nit: writer commit err

enterprise/server/backends/gcs_cache/gcs_cache.go
tylerwilliams: The underlying io.WriteCloser comes from the GCS storage library so it doesn't have a commit() method -- it finalizes a write when Close() is called.

To work around this, we call Close() on the gcs WriteCloser when Commit() is called, and when Close() is called, we cancel the context given to the gcs WriteCloser.

This means that if Close is called but Commit is not, the file will *not* be written, but resources will be cleaned up. If both are called, the file will be written and resources will be cleaned up.

enterprise/server/backends/migration_cache/migration_cache.go
tylerwilliams: Done.

siggisim (approved):

PR #2521, opened 2022-09-12
Remove runner pool memory restrictions for Firecracker

bduffany:
- Update the runner pool logic and `container.Stats()` API to allow returning 0 as a valid memory usage value, and to allow returning `nil` to represent "not implemented, use an estimate instead"
- Update all container implementations so that only Firecracker returns 0 memory from `Stats()`, and all the other impls which don't yet implement `Stats()` (or have it disabled) return `nil` instead.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -253,7 +253,7 @@ func main() {
 	}
 	executorID := executorUUID.String()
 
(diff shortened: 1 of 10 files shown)


tylerwilliams (approved):
does it make sense to add a test that a certain number of fc containers can fit?

PR #2527, opened 2022-09-13
Allow manually exporting Grafana dashboard JSON

bduffany:
Sometimes, the auto-sync from Grafana -> tools/metrics/grafana/dashboards/buildbuddy.json doesn't work. I haven't yet figured out why this happens, but until we have a proper fix, this gives an escape hatch to manually export saved changes by running the command `./tools/metrics/run.sh sync`

---

**Version bump**: Patch


bduffany:
I think the auto-sync should be fixed in https://github.com/buildbuddy-io/buildbuddy/pull/2533 (changing `sync` to `sync || true` so that we don't exit the sync loop if it fails initially)

PR #2546, opened 2022-09-16
Support wider range of auth providers

siggisim:
Improvements that make our auth library work better with a wider variety of auth providers.
- Don't assume auth token refresh interval on the front-end (different providers issue access tokens with a 5 minute lifespan). Instead set a cookie that the frontend can read when we set the login cookies.
- Don't assume all auth providers support `consent=prompt` (some don't) and put this behind a flag.
- If an auth provider gives us a new refresh token in the token refresh response, update the session so we use that one going forward (some auth providers will "roll" the refresh token, so old tokens become invalid once they're used once).
- Don't assume that refresh tokens contain an "id_token" (some providers don't set this). If they don't, preserve the original id token.

---

**Version bump**: Patch

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -59,7 +59,7 @@ export class AuthService {
   }
 
   getCookie(name: string) {
-    var match = document.cookie.match(new RegExp("(^| )" + name + "=([^;]+)"));
+    let match = document.cookie.match(new RegExp("(^| )" + name + "=([^;]+)"));
     if (match) return match[2];
   }
 
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -14,7 +14,7 @@ export { User };
 const SELECTED_GROUP_ID_LOCAL_STORAGE_KEY = "selected_group_id";
 const IMPERSONATING_GROUP_ID_SESSION_STORAGE_KEY = "impersonating_group_id";
 const AUTO_LOGIN_ATTEMPTED_STORAGE_KEY = "auto_login_attempted";
-const TOKEN_REFRESH_INTERVAL_MS = 30 * 60 * 1000; // 30 minutes
+const TOKEN_REFRESH_INTERVAL_SECONDS = 30 * 60; // 30 minutes
 
 export class AuthService {
   user: User = null;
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
ace

app/auth/auth_service.ts
```diff
@@ -50,11 +50,26 @@ export class AuthService {
...
+      })
+      .finally(() => {
+        setTimeout(() => {
+          this.startRefreshTimer();
+        }, 1000); // Wait a second before starting the refresh timer so we can grab the session duration.
       });
+  }
+
+  getCookie(name: string) {
+    var match = document.cookie.match(new RegExp("(^| )" + name + "=([^;]+)"));
```
bduffany, line 62: nit

```suggestion
    let match = document.cookie.match(new RegExp("(^| )" + name + "=([^;]+)"));
```

enterprise/server/auth/auth.go
```diff
@@ -46,6 +46,7 @@ var (
 	apiKeyGroupCacheTTL  = flag.Duration("auth.api_key_group_cache_ttl", 5*time.Minute, "TTL for API Key to Group caching. Set to '0' to disable cache.")
 	httpsOnlyCookies     = flag.Bool("auth.https_only_cookies", false, "If true, cookies will only be set over https connections.")
 	disableRefreshToken  = flag.Bool("auth.disable_refresh_token", false, "If true, the offline_access scope which requests refresh tokens will not be requested.")
+	forceApproval        = flag.Bool("auth.force_approval", false, "If true, when a user doesn't have a session (first time logging in, or manually logged out) force the auth provider to show the consent screen allowing the user to select an account if they have multiple. This isn't supported by all auth providers.")
```
bduffany, line 49: sanity check, should this default to true to keep the existing behavior?

app/auth/auth_service.ts
```diff
@@ -50,11 +50,26 @@ export class AuthService {
         } else {
           this.onUserRpcError(error);
         }
+      })
+      .finally(() => {
+        setTimeout(() => {
+          this.startRefreshTimer();
+        }, 1000); // Wait a second before starting the refresh timer so we can grab the session duration.
```
bduffany, line 57: not sure I understand why the 1s delay is needed, shouldn't the cookie be set on the response (once then() is done?)

bduffany (approved):

app/auth/auth_service.ts
siggisim: Done

app/auth/auth_service.ts
siggisim: In my testing about half of the time the cookie value was set by the time this ran, and the other half of the time it wasn't.

Not sure exactly how/when cookies are updated from an XHR. It's possible I just need to wait a tick instead of a whole second - but this is a longer running refresh loop that will happen in minutes to hours, so it doesn't matter much.

enterprise/server/auth/auth.go
siggisim: This slightly changes the default behavior, but it's a nice to have change and this default allows us to support any auth provider with the default config.

I'm going to send a separate config PR to update our Cloud yaml files to enable this.

PR #2547, opened 2022-09-19
Turn splits inside out

tylerwilliams:


enterprise/server/raft/replica/replica.go
```diff
@@ -743,194 +669,71 @@ func (sm *Replica) deleteStoredFiles(start, end []byte) error {
...
-		LowerBound: keys.Key(start),
-		UpperBound: keys.Key(end),
+func (sm *Replica) copyStoredFiles(req *rfpb.CopyStoredFilesRequest) (*rfpb.CopyStoredFilesResponse, error) {
+	appliesToThisReplica := sm.splitAppliesToThisReplica(req.GetSourceRange())
+	if !appliesToThisReplica {
+		sm.log.Debugf("Got copyStoredFiles cmd but it doesn't apply to this replica; skipping")
+		return &rfpb.CopyStoredFilesResponse{}, nil
+	}
+
+	localReplica, err := sm.store.GetReplica(req.GetTargetRangeId())
```
vadimberezniker, line 679: can we call this targetReplica or destReplica? both of the replicas involved are local

enterprise/server/raft/replica/replica.go
```diff
@@ -1695,6 +1513,52 @@ func (sm *Replica) ApplySnapshotFromReader(r io.Reader, db ReplicaWriter) error
...
+	readBuf := bufio.NewReader(r)
+	go func() {
+		defer close(ch)
+		for {
+			r, count, err := readDataFromReader(readBuf)
+			if err != nil {
+				if err == io.EOF {
+					break
+				}
+				ch <- Record{Error: err}
```
vadimberezniker, line 1533: looks like you are missing `continue` here and for the other error cases in this function

proto/raft.proto
```diff
@@ -184,6 +184,17 @@ message SplitReleaseRequest {
 }
 message SplitReleaseResponse {}
 
+// A request to hold or renew the split lock, blocking all writes. The split
```
vadimberezniker, line 187: fix comment

tylerwilliams: Done.

enterprise/server/raft/replica/replica.go
tylerwilliams: Added break -- any error in this function stops the show

enterprise/server/raft/replica/replica.go
tylerwilliams: Done

PR #2557, opened 2022-09-20
Upgrade podman to 4.2.1

vadimberezniker:
I don't think this addresses the hanging issue we saw, but it doesn't hurt to pick up the fixes in this release.

https://github.com/buildbuddy-io/buildbuddy-internal/issues/1661

---

**Version bump**: Patch


tylerwilliams (approved):
We should test this in dev before the release tomorrow

PR #2562, opened 2022-09-21
Add optional basic auth to monitoring port

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/server/util/monitoring/monitoring.go
+++ b/server/util/monitoring/monitoring.go
@@ -1,29 +1,44 @@
 package monitoring
 
 import (
+	"flag"
 	"net/http"
 	"net/http/pprof"
 
+	"github.com/buildbuddy-io/buildbuddy/server/util/basicauth"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/statusz"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+var (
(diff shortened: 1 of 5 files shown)


server/util/monitoring/monitoring.go
```diff
@@ -1,29 +1,44 @@
...
+var (
+	basicAuthUser = flag.String("monitoring.basic_auth.username", "", "Optional username for basic auth on the monitoring port.")
+	basicAuthPass = flag.String("monitoring.basic_auth.password", "", "Optional password for basic auth on the monitoring port.")
+)
+
 // Registers monitoring handlers on the provided mux. Note that using
 // StartMonitoringHandler on a monitoring-only port is preferred.
 func RegisterMonitoringHandlers(mux *http.ServeMux) {
+	handle := mux.Handle
+	if *basicAuthUser != "" || *basicAuthPass != "" {
```
tylerwilliams, line 23: Should this be &&?

vadimberezniker (approved):

server/util/monitoring/monitoring.go
bduffany: I don't think it's super consequential either way, but I went with `||` since its valid to have an empty username and only require a password.

tylerwilliams: SG. Could you add a testcase that does that so it's clear that should work?

(reason I commented is I saw you were using username to lookup the pw in a map, so assumed it shouldn't be "")

bduffany: Done

tylerwilliams (approved):
LGTM

PR #2564, opened 2022-09-21
Allow SSL to be used for monitoring endpoint on executor

bduffany:
When `--monitoring.ssl_port` >= 0, use the SSL config to serve the monitoring handler with HTTPS on that port.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -33,6 +33,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/byte_stream_server"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/content_addressable_storage_server"
 	"github.com/buildbuddy-io/buildbuddy/server/resources"
+	"github.com/buildbuddy-io/buildbuddy/server/ssl"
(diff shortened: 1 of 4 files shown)


enterprise/server/cmd/executor/executor.go
```diff
@@ -294,8 +296,18 @@ func main() {
 		repb.RegisterActionCacheServer(localServer, actionCacheServer)
 	}
 
+	// Setup SSL for monitoring endpoints (optional).
+	if err := ssl.Register(env); err != nil {
```
tylerwilliams, line 300: will this fail if no certs are configured? (It seems like enabling ssl should be fully optional)

I had a look in ssl.go and it seemed possible, but I haven't run it so not 100%.

bduffany: The SSL service becomes effectively a no-op if `ssl.ssl_enabled` is false, but I moved this inside the conditional below just to make it extra clear that SSL is for monitoring only.

tylerwilliams: oh i missed that flag. thanks for checking!

tylerwilliams (approved):

PR #2566, opened 2022-09-21
Don't allow remote syncproposes when the range is splitting

tylerwilliams:
Normal Read or Write RPCs to the replica will acquire a lease on the
DB which will fail during splitting. SyncPropose, however, proposes
a cmd to the raft statemachine, which (with some exceptions), cannot
apply during a split. To avoid SyncProposing anything into the raft
log during a range split, we check here if the replica is splitting
before doing the syncPropose.


tools/goreman/procfiles/Procfile.raft
```diff
@@ -1,7 +1,7 @@
-app1: bazel run enterprise/server -- --config_file=enterprise/config/buildbuddy.raft.yaml --telemetry_port=9901 --grpc_port=1970 --monitoring_port=9101 --internal_grpc_port=7701 --disable_telemetry --cache.raft.root_directory=/tmp/rnode1 --cache.raft.listen_addr=127.0.0.1:9201 --cache.raft.http_addr=127.0.0.1:9301 --cache.raft.grpc_addr=127.0.0.1:9401 --auto_migrate_db=false
+app1: bazel run enterprise/server -- --config_file=enterprise/config/buildbuddy.raft.yaml --telemetry_port=9901 --grpc_port=1970 --monitoring_port=9101 --internal_grpc_port=7701 --disable_telemetry --cache.raft.root_directory=/tmp/rnode1 --cache.raft.listen_addr=127.0.0.1:9201 --cache.raft.http_addr=127.0.0.1:9301 --cache.raft.grpc_addr=127.0.0.1:9401 --auto_migrate_db=false --cache.raft.test_mode=true
```
tylerwilliams, line 1: I keep adding this, figured I'd just check it in for now

vadimberezniker (approved):

PR #2568, opened 2022-09-21
Optional lame duck period for shutdown.

vadimberezniker:
This will allow clients (LBs) to continue sending requests until they mark the backends as unready.

---

**Version bump**: Patch

### DIFF
--- a/server/util/healthcheck/healthcheck.go
+++ b/server/util/healthcheck/healthcheck.go
@@ -21,7 +21,8 @@ import (
 )
 
 var (
-	maxShutdownDuration = flag.Duration("max_shutdown_duration", 25*time.Second, "Time to wait for shutdown")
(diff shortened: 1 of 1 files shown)


server/util/healthcheck/healthcheck.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 var (
-	maxShutdownDuration = flag.Duration("max_shutdown_duration", 25*time.Second, "Time to wait for shutdown")
+	maxShutdownDuration      = flag.Duration("max_shutdown_duration", 25*time.Second, "Time to wait for shutdown")
```
tylerwilliams, line 24: ah nice, i guess we did have a flag for this already :)

tylerwilliams (approved):

PR #2569, opened 2022-09-21
Add prometheus metric to track number of not found errors during a migration

maggie-lou:
---

**Version bump**: TODO


maggie-lou:
* master
  * **PR #2569** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2569" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2569?utm_source=stack-comment).

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -657,6 +660,10 @@ func (s *ContentAddressableStorageServer) GetTree(req *repb.GetTreeRequest, stre
 				}
 			} else {
 				log.Warningf("Not caching incomplete tree cache")
+				if _, migration := cache.(*migration_cache.MigrationCache); migration {
```
vadimberezniker, line 663: I don't think this is useful. We already see this warning for the existing cache so this is going to give you false positives.

server/metrics/metrics.go
```diff
@@ -132,6 +132,9 @@ const (
 
 	/// Whether the request was allowed by quota manager.
 	QuotaAllowed = "quota_allowed"
+
+	// The type of request a not found error was initiated from
+	NotFoundErrorType = "type"
```
vadimberezniker, line 137: We should call this `CacheRequestType` or something like that, no need to tie the label name to a specific metric.

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -196,6 +198,11 @@ func (mc *MigrationCache) Contains(ctx context.Context, d *repb.Digest) (bool, e
 		log.Warningf("Migration digest %v src contains %v, dest contains %v", d, srcContains, dstContains)
 	}
 
+	if doubleRead && srcContains && !dstContains {
+		metrics.MigrationNotFoundErrorCount.With(prometheus.Labels{metrics.NotFoundErrorType: "contains"}).Inc()
+		metrics.MigrationNotFoundErrorCount.With(prometheus.Labels{metrics.NotFoundErrorType: "all"}).Inc()
```
vadimberezniker, line 203: There's no need for an `all` value since the data can be aggregated in prometheus / grafana.

vadimberezniker (changes requested):

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
maggie-lou: I added this metric so we can make sure that the number of not found errors goes down over time, which we'd expect as more data gets copied. Was seeing a lot of this error when testing and assumed it was from data not being copied yet, so wanted to make sure there wasn't a permanent problem if this stays high. Seemed fine if the count includes false positives as long as it trends down - do you disagree?

server/metrics/metrics.go
maggie-lou: Done

enterprise/server/backends/migration_cache/migration_cache.go

vadimberezniker (approved):

PR #2570, opened 2022-09-22
Allow a lil docker tcp socket as a treat

tylerwilliams:
If folks want to run docker and connect to it over tcp, they will now be able to by setting

```
executor.docker_host=tcp://localhost:2375
```

in their config. Existing configs which set a file path should still have "unix://" prefixed to them and be unaffected.


bduffany (approved):
noice

PR #2571, opened 2022-09-22
Fix tools/enable_local_firecracker.sh to allow cgroup2 mounted on /sys/fs/cgroup/unified

bduffany:
In Ubuntu 20.04.1, cgroup2 is mounted on `/sys/fs/cgroup/unified`, not `/sys/fs/cgroup`. Update the script to read the cgroup2 path using `mount | grep cgroup2`. Also delete the old cgroup v1 stuff (cpuset), and fix a few shellcheck lints.

---

**Version bump**: Patch

### DIFF
--- a/tools/enable_local_firecracker.sh
+++ b/tools/enable_local_firecracker.sh
@@ -5,63 +5,62 @@ set -euo pipefail
 # user to run firecracker (via the jailer), without being root.
 
 # this script must be run as root, so check that first.
-if [ $(id -u) -ne 0 ]; then
+if [ "$(id -u)" -ne 0 ]; then
     echo "This script must be run as root (use sudo)."
     echo "sudo $0"
     exit 1
 fi
 
-if ! command -v jailer &> /dev/null
-then
+if ! command -v jailer &>/dev/null; then
     echo "jailer could not be found (install firecracker + jailer?)"
     exit 1
 fi
 
-if ! command -v ip &> /dev/null
-then
+if ! command -v ip &>/dev/null; then
     echo "ip could not be found (install iproute2?)"
     exit 1
 fi
 
-if ! command -v iptables &> /dev/null
-then
+if ! command -v iptables &>/dev/null; then
     echo "iptables could not be found"
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


tools/enable_local_firecracker.sh
```diff
@@ -5,63 +5,62 @@ set -euo pipefail
...
+chmod -R g+rw "$CGROUP2_PATH"/firecracker
 
-chown -R root:cgroups /sys/fs/cgroup/cgroup.subtree_control
-chmod -R g+rw /sys/fs/cgroup/cgroup.subtree_control
-chown -R root:cgroups /sys/fs/cgroup/cgroup.procs
-chmod -R g+rw /sys/fs/cgroup/cgroup.procs
+chown -R root:cgroups "$CGROUP2_PATH"/cgroup.subtree_control
+chmod -R g+rw "$CGROUP2_PATH"/cgroup.subtree_control
+chown -R root:cgroups "$CGROUP2_PATH"/cgroup.procs
+chmod -R g+rw "$CGROUP2_PATH"/cgroup.procs
```
tempoz, line 47: I know this is my fault, but this should probably be:
```suggestion
chown -R "$SUDO_USER":cgroups "$CGROUP2_PATH"/cgroup.subtree_control
chmod -R g+rw "$CGROUP2_PATH"/cgroup.subtree_control
chown -R "$SUDO_USER":cgroups "$CGROUP2_PATH"/cgroup.procs
chmod -R g+rw "$CGROUP2_PATH"/cgroup.procs
```
right?

tylerwilliams (approved):
bash 😍

tempoz (approved):

tools/enable_local_firecracker.sh
bduffany: firecracker_test seems to pass either way :shrug: 

I guess it doesn't matter since our user gets added to the cgroups group, and so we inherit permissions from there?

tempoz: Fair enough.

bduffany: Changed to SUDO_USER just for consistency.

PR #2572, opened 2022-09-22
Support dockerUser platform prop for firecracker

bduffany:
Can be specified as `USER[:GROUP]`, where both USER and GROUP can be either names or IDs (same as Docker).

To be consistent with Docker, we do not create the user automatically. Instead the user is expected to be provisioned as part of the Docker image.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1643

### DIFF
--- a/proto/vmexec.proto
+++ b/proto/vmexec.proto
@@ -64,7 +64,8 @@ message ExecRequest {
   // or not stdin is attached.
   bool open_stdin = 8;
 
-  // Optional user to run the command as.
+  // Optional user to run the command as. If unset, the command will run as
+  // root.
   string user = 9;
 }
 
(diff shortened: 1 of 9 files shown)


proto/vmexec.proto
```diff
@@ -63,6 +63,9 @@ message ExecRequest {
   // the command, because commands can behave differently depending on whether
   // or not stdin is attached.
   bool open_stdin = 8;
+
+  // Optional user to run the command as.
```
tylerwilliams, line 67: nit: if unset, command will run as root.

tylerwilliams (approved):

tylerwilliams (commented):
(looks like test might need updating to pass in the user now https://app.buildbuddy.io/invocation/99d77ea9-9522-4b01-8d33-5a8f04818641?target=%2F%2Fenterprise%2Fserver%2Fvmexec%3Avmexec_test)

proto/vmexec.proto
bduffany: Done

PR #2573, opened 2022-09-22
Add bazel "command" as an advanced filter

siggisim:
This column doesn't (yet) have an index, but this is under advanced filter - so we can worry about that if we get any negative performance feedback. 

Not sure if there's anything I need to do on the clickhouse side of things of whether this will "just work". cc @luluz66 

I can verify clickhouse compatibility when this goes to dev.

Screen Shot 2022-09-22 at 2 13 17 PM

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/invocation_stat_service/invocation_stat_service.go
+++ b/enterprise/server/invocation_stat_service/invocation_stat_service.go
@@ -123,6 +123,10 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
 		q.AddWhereClause("branch_name = ?", branchName)
 	}
 
+	if command := req.GetQuery().GetCommand(); command != "" {
+		q.AddWhereClause("command = ?", command)
+	}
+
 	if commitSHA := req.GetQuery().GetCommitSha(); commitSHA != "" {
 		q.AddWhereClause("commit_sha = ?", commitSHA)
 	}
@@ -264,6 +268,10 @@ func (i *InvocationStatService) GetInvocationStat(ctx context.Context, req *inpb
 
 	if branchName := req.GetQuery().GetBranchName(); branchName != "" {
 		q.AddWhereClause("branch_name = ?", branchName)
+	}
+
+	if command := req.GetQuery().GetCommand(); command != "" {
+		q.AddWhereClause("command = ?", command)
 	}
 
 	if commitSHA := req.GetQuery().GetCommitSha(); commitSHA != "" {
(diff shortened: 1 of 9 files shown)


siggisim:
Inspired by https://buildbuddy.slack.com/archives/CUY16GNK1/p1663873897444819

tylerwilliams (approved):
This LGTM but I'll let @luluz66 answer the clickhouse question

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -123,6 +123,10 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
 		q.AddWhereClause("branch_name = ?", branchName)
 	}
 
+	if command := req.GetQuery().GetCommand(); command != "" {
```
luluz66, line 126: "command" was not added as a column in clickhouse yet. 
https://cs.github.com/buildbuddy-io/buildbuddy/blob/6e1e97fcf1d1b435672ad43e4616ea63ff6fae29/server/util/clickhouse/clickhouse.go#L50

This means when we switch on clickhouse the query will be broken because the column doesn't exist. maybe only add the where clause when  i.isOLAPDBEnabled() is false.

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -266,6 +270,10 @@ func (i *InvocationStatService) GetInvocationStat(ctx context.Context, req *inpb
...
```
luluz66, line 273: same here.

luluz66 (commented):
"command" is not added as a clickhouse column. 
https://cs.github.com/buildbuddy-io/buildbuddy/blob/6e1e97fcf1d1b435672ad43e4616ea63ff6fae29/server/util/clickhouse/clickhouse.go#L50

We need to added it in the above struct and also in the `ToInvocationFromPrimaryDB`. function. After that, we need to ingest the column to dev.

luluz66 (changes requested):
forget to change to "request changes"

enterprise/server/invocation_stat_service/invocation_stat_service.go
siggisim: Added the column to clickhouse - definitely want this to work in either case.

Would there be some sort of backfill that I need to run? Or have we not done that in production yet?

enterprise/server/invocation_stat_service/invocation_stat_service.go
siggisim: Same as above

luluz66 (approved):

PR #2576, opened 2022-09-22
Fix firecracker workspace dir perms for non-root users

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
@@ -8,12 +8,14 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -504,7 +527,9 @@ func TestFirecrackerNonRoot(t *testing.T) {
 		t.Fatal(res.Error)
 	}
 	require.NoError(t, res.Error)
-	assert.Equal(t, "uid=65534(nobody) gid=65534(nobody)\n", string(res.Stdout))
+	require.Empty(t, string(res.Stderr))
+	require.Equal(t, 0, res.ExitCode)
+	require.Equal(t, "uid=65534(nobody) gid=65534(nobody)\n", string(res.Stdout))
```
tylerwilliams, line 532: probably fine but be aware https://stackoverflow.com/questions/34831861/can-i-assume-that-nobody-is-65534

tylerwilliams (approved):

PR #2578, opened 2022-09-23
Fix CLI RBE autoconfig

bduffany:
* Fix indentation (bazel fails to build w/ the current indentation)
* Import http_archive if not already imported

---

**Version bump**: Patch


siggisim (approved):
I think I want to remove this toolchain stuff altogether, but okay with this change for now!

PR #2586, opened 2022-09-24
Revert cgroup v1 change in tools/enable_local_firecracker

bduffany:
I removed these in the last PR, but turns out we do need these.

---

**Version bump**: Patch


tempoz (approved):
LGTM

PR #2587, opened 2022-09-26
Reset flags when reloading config

tempoz:
We have to reset the flags before we reload the config file, as some flags (currently just slice flags) can accumulate values when `Set` is called as opposed to replacing them. Using slice flags as an example, this would mean that if the flags were Set again without a reset, the flag would then be `[defaultValues..., originalConfigValues..., newConfigValues...]` (where `defaultValues` is normally empty).

Additionally, there was a bug where StringSliceFlags interpreted `""` as a new element that was the empty string, instead of as an empty slice with no new values. This has been fixed.

The code previously in `flagz` that was used to override flags has been moved to `common.go` so it can be shared between `common.ResetFlags` and `flagz`, and a bug where `flag.Value`s that were backed by `reflect.Value`s were not being correctly initialized has been addressed.

The side effect of a registering a signal handler has been removed from `PopulateFlagsFromFile` and is now explicitly done through `config.ReloadOnSIGHUP()`.

---

**Version bump**: Patch


tempoz:
* master
  * **PR #2587** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2587" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2587?utm_source=stack-comment).

server/util/signalutil/signalutil.go
```diff
@@ -0,0 +1,19 @@
...
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, s)
+	go func() {
+		for range c {
+			if err := fn(); err != nil {
+				errFn(err)
+			}
+		}
+	}()
+}
```
bduffany, line 19: I am a bit concerned about this func obscuring the fact that a goroutine is created and never cleaned up, which could lead to a goroutine leak if this is used improperly.

I think we should avoid introducing this util and instead use `signal.Notify() / go func() ...` directly.

tempoz: directly in `ReloadOnSIGHUP`? or directly in the `main` functions after calling `PopulateFlagsFromFile`? If the former, I suppose we could cleanup a previous goroutine if it exists and then start the goroutine, would that be good?

tempoz: Could also name it something like "MakeSignalListener" (likely terrible name, but that idea) to emphasize that when you make one, it sticks around and does something?

bduffany: I think it'd be fine to do it in ReloadOnSIGHUP and just add a clarifying comment to ReloadOnSIGHUP that this is only intended to be called in `main`.

tempoz: Cool; sounds great.

bduffany (approved):

PR #2591, opened 2022-09-27
Make main sidebar links cmd+clickable

bduffany:
---

**Version bump**: Patch


app/router/router.tsx
```diff
@@ -442,4 +397,38 @@ export class Path {
   static codePath = "/code/";
 }
 
+function getUnavailableMessage(href: string) {
+  let path = new URL(href, window.location.href).pathname;
+  if (!path.endsWith("/")) path += "/";
+
+  switch (path) {
```
siggisim, line 404: We can probably collapse this into a more generic message like:

This feature is not available in ${capabilities.name}.\n\nClick 'Upgrade to Enterprise' in the menu to enable user build history, organization build history, SSO, and more!

siggisim (approved):

app/router/router.tsx
bduffany: Done!

PR #2594, opened 2022-09-27
Convenience methods for manipulating args in the CLI

siggisim:
---

**Version bump**: Patch

### DIFF
--- a/cli/arg/BUILD
+++ b/cli/arg/BUILD
@@ -1,0 +1,8 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "arg",
+    srcs = ["arg.go"],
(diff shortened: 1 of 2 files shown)


bduffany (approved):
LGTM modulo the API change w/ accepting "foo" instead of "--foo=" for the desiredArg value

Also nit, omitting "arg" from the function names would be slightly more idiomatic, so the callsites are like `arg.HasArg => arg.Has`, `arg.GetArg => arg.Get` etc.

siggisim:
> LGTM modulo the API change w/ accepting "foo" instead of "--foo=" for the desiredArg value

Updated so these now take "foo" instead of "--foo=" so we can add support for more flag formats more easily later.

> Also nit, omitting "arg" from the function names would be slightly more idiomatic, so the callsites are like `arg.HasArg => arg.Has`, `arg.GetArg => arg.Get` etc.

Done

PR #2595, opened 2022-09-27
Remove arch-specific sidecar targets

bduffany:
We're running into some issues with `gosigar` with these arch-specific targets; not really sure why. Seems to build OK after this PR.

---

**Version bump**: Patch


tylerwilliams (approved):

bduffany:
Going to close this for now since removing the `gosigar` dep did the trick, and I don't fully understand whether the `gc_linkopts` were needed for some reason. I can reopen if we feel the simplicity we gain here is worth it.

PR #2611, opened 2022-09-28
Add method that parses .bazelrc files to a list of effective args

siggisim:
This is based on logic that currently lives here, but will be removed:
https://github.com/buildbuddy-io/buildbuddy/blob/master/cli/cmd/bb/bb.go#L61

Also makes logging non-verbose by default.

---

**Version bump**: None

### DIFF
--- a/cli/parser/parser.go
+++ b/cli/parser/parser.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 
 	"github.com/buildbuddy-io/buildbuddy/cli/arg"
-	
+
 	bblog "github.com/buildbuddy-io/buildbuddy/cli/logging"
 )
 
--- a/cli/parser/parser.go
+++ b/cli/parser/parser.go
@@ -4,10 +4,12 @@ import (
 	"bufio"
 	"io"
 	"os"
+	"os/user"
 	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/buildbuddy-io/buildbuddy/cli/arg"
 	bblog "github.com/buildbuddy-io/buildbuddy/cli/logging"
 )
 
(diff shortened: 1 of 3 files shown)


cli/parser/parser.go
```diff
@@ -116,3 +118,52 @@ func GetFlagValue(options []*BazelOption, phase, config, flagName, commandLineOv
...
+	if arg.Get(commandLineArgs, "workspace_rc") != "true" {
+		rcFiles = append(rcFiles, ".bazelrc")
+	}
+	if arg.Get(commandLineArgs, "home_rc") != "true" {
+		usr, err := user.Current()
+		if err == nil {
+			rcFiles = append(rcFiles, filepath.Join(usr.HomeDir, ".bazelrc"))
+		}
+	}
+	if b := arg.Get(commandLineArgs, "bazelrc"); b != "" {
```
bduffany, line 137: `--bazelrc` can be specified multiple times -- fix or add TODO?

bduffany (approved):

cli/parser/parser.go
```diff
@@ -4,10 +4,12 @@ import (
 	"bufio"
 	"io"
 	"os"
+	"os/user"
 	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/buildbuddy-io/buildbuddy/cli/arg"
 	bblog "github.com/buildbuddy-io/buildbuddy/cli/logging"
```
bduffany, line 13:
```suggestion
	"github.com/buildbuddy-io/buildbuddy/cli/arg"

	bblog "github.com/buildbuddy-io/buildbuddy/cli/logging"
```

cli/parser/parser.go
siggisim: This is an issue with the current logic as well, so I've just added a TODO for now.

cli/parser/parser.go
siggisim: Done

PR #2618, opened 2022-09-29
Clean up logging in CLI

siggisim:
Rename logging => log so we don't have to use bblog everywhere.

Differentiate between debug logs (that don't get printed unless verbose) and non-debug logs.

---

**Version bump**: None


bduffany (commented):
I feel like this might result in a lot of accidental auto-importing from `//server/util/log`, but let's try it out and see what happens :+1: the ergonomics are definitely better than having to manually add the `bblog` alias

bduffany (approved):

PR #2620, opened 2022-09-29
CLI: Add support for remote plugins

bduffany:
Support the "repo" parameter in the plugin spec.

Repos are cloned to the storage dir, under `~/.cache/buildbuddy/`, then under `plugins/HOST/OWNER/REPO`, then in a ref-specific subdir: either `latest` if no ref is specified or `ref/REF` if `@<branch,sha,etc>` is specified (the `ref/` is needed so that we can distinguish between "latest commit from the default branch" and "an actual branch named 'latest'")

Also fix a few bugs found when testing.

---

**Version bump**: Patch

### DIFF
--- a/cli/plugin/plugin.go
+++ b/cli/plugin/plugin.go
@@ -3,13 +3,18 @@ package plugin
 import (
 	"bytes"
 	"context"
+	"io"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 
+	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	"github.com/buildbuddy-io/buildbuddy/cli/storage"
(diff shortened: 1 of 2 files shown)


cli/plugin/plugin.go
```diff
@@ -109,19 +238,27 @@ func (p *Plugin) Path() string {
 //       for arg in "$@"; do
 //         echo "$arg"
 //       done
-//       if ! timeout 1 ping -c1 remote.buildbuddy.io; then
+//       # Note: redirect ping stdout to stderr, to avoid adding ping output to
+//       # bazel args.
```
bduffany, line 242: @siggisim I ran into this bug when writing a test plugin -- I imagine this'll be a pain point for plugin authors with the stdout-based API (having to remember to pipe unwanted output to stderr).

siggisim (approved):

PR #2623, opened 2022-09-29
Add metadata fields to cacheproxy

maggie-lou:
Fixes buildbuddy-io/buildbuddy-internal#1646


siggisim (approved):
Nice find!

siggisim:
I thin

> > Nice find!
> 
> Honestly surprising that it was even returning data for you sometimes... 🤷‍♀️

I think it happened when I got lucky and hit a machine that had the data locally, so we didn't have to go through this cache proxy code.

PR #2624, opened 2022-09-29
Use a fixed retry interval during splits.

vadimberezniker:
---

**Version bump**: None

### DIFF
--- a/enterprise/server/raft/sender/sender.go
+++ b/enterprise/server/raft/sender/sender.go
@@ -21,6 +21,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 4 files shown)


enterprise/server/raft/sender/sender.go
```diff
@@ -19,6 +20,10 @@ import (
 	rfspb "github.com/buildbuddy-io/buildbuddy/proto/raft_service"
 )
 
+const (
+	splitRetryInterval = 250 * time.Millisecond
```
tylerwilliams, line 24: nit: add a comment explaining why?

tylerwilliams (approved):

PR #2625, opened 2022-09-29
Add `login` command to CLI

siggisim:
---

**Version bump**: None

### DIFF
--- a/cli/login/login.go
+++ b/cli/login/login.go
@@ -55,6 +55,7 @@ func ConfigureAPIKey(args []string) []string {
 		return args
 	}
 
+	// TODO(siggisim): find a more graceful way of finding headers if we change the way we parse flags.
 	if !arg.Has(args, apiKeyHeader) {
 		return append(args, "--"+apiKeyHeader+"="+string(apiKeyBytes))
 	}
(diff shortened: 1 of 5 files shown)


cli/login/login.go
```diff
@@ -0,0 +1,101 @@
...
+	"path/filepath"
+
+	"github.com/buildbuddy-io/buildbuddy/cli/arg"
+	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	"github.com/buildbuddy-io/buildbuddy/cli/storage"
+)
+
+const (
+	apiKeyFileName = "apikey"
+	apiKeyHeader   = "remote_header=x-buildbuddy-api-key"
```
bduffany, line 15: nit: this might not work later if we make the arg parsing more robust - might save some headache later if we have a util func in the arg pkg like GetRemoteHeaders() that returns a map[string]string. Or a TODO so we don't forget

bduffany (approved):

cli/login/login.go
siggisim: Good catch, I've added a TODO. We might squeak by if we go the arg normalization / canonicalization route.

PR #2627, opened 2022-09-29
Be more flexible about choosing a split point; add tests

tylerwilliams:
This change makes the splitting more flexible: any splittable point in a range is valid. The algorithm searches through the entire range for a split point, and determines the best split point by how closely it approximates splitting the range into two even halves.

If a split point cannot be found for some reason, the range's keys are printed (so we can debug better).

We can tweak it if we want so it only ever splits in the middle third or something, but figured I'd start here. Open to feedback!


vadimberezniker (approved):
good change

PR #2628, opened 2022-09-30
Wait for migration-related channels that use cache resources (like copying) to close before shutting down the caches

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/pebble_cache"
@@ -42,6 +43,9 @@ type MigrationCache struct {
 	copyChan                    chan *copyData
 	copyChanFullWarningInterval time.Duration
 	numCopiesDropped            *int64
+
+	srcShutdownFn  func() error
+	destShutdownFn func() error
 }
 
 func Register(env environment.Env) error {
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #2628** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2628" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2628?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -75,7 +79,7 @@ func Register(env environment.Env) error {
 	return nil
 }
 
-func NewMigrationCache(migrationConfig *MigrationConfig, srcCache interfaces.Cache, destCache interfaces.Cache) *MigrationCache {
+func NewMigrationCache(migrationConfig *MigrationConfig, srcCache interfaces.Cache, destCache interfaces.Cache, srcShutdownFn func() error, destShutdownFn func() error) *MigrationCache {
```
vadimberezniker, line 82: it would be slightly cleaner to define an interface for caches that can be stopped instead of passing the shutdown function separately. You can add a Stop() function to disk_cache that doesn't do anything to satisfy the interface.
```
type StoppableCache interface {
  interfaces.Cache
  Stop() error
}
```

vadimberezniker (changes requested):

enterprise/server/backends/migration_cache/migration_cache.go

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -736,8 +734,40 @@ func (mc *MigrationCache) Start() error {
...
+	// Wait for migration-related channels that use cache resources (like copying) to close before shutting down
+	// the caches
 	if err := mc.eg.Wait(); err != nil {
 		return err
 	}
-	return nil
+
+	var wg sync.WaitGroup
+	var srcShutdownErr, dstShutdownErr error
+	if src, canStopSrc := mc.src.(interfaces.StoppableCache); canStopSrc {
```
maggie-lou, line 745: @vadimberezniker  Just to make sure I interpreted the docs correctly - when I cast mc.src here and call src.Stop(), it's calling Stop on mc.src right? It's not copying mc.src -> src and only calling Stop on the copy? 

Reading stuff like this : 
https://stackoverflow.com/questions/13511203/why-cant-i-assign-a-struct-to-an-interface

vadimberezniker: That's correct. You're effectively calling methods on the original mc.src

vadimberezniker (approved):

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Great - thanks for confirming

PR #2630, opened 2022-09-30
Smooth out CLI logging API

bduffany:
- Make the API match the underlying log API (`Print(...any)` instead of `Print(string)`)
- Add plain `Debug` and `Print` funcs

---

**Version bump**: Patch

### DIFF
--- a/cli/log/log.go
+++ b/cli/log/log.go
@@ -9,6 +9,10 @@ var (
 	verbose = flag.Bool("verbose", false, "If true, enable verbose buildbuddy logging.")
 )
 
+func init() {
+	log.SetFlags(0)
+}
+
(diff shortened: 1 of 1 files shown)


cli/log/log.go
```diff
@@ -9,8 +9,11 @@ var (
 	verbose = flag.Bool("verbose", false, "If true, enable verbose buildbuddy logging.")
 )
 
-func init() {
```
siggisim, line 12: Can we keep this SetFlags call?

siggisim (approved):
Lgtm with a nit

cli/log/log.go
bduffany: woops, reverted

PR #2631, opened 2022-09-30
CLI: Wire up plugins in `main`

bduffany:
Actually load plugins in `main()` and run their pre/post-bazel plugin hooks.

Tested with both pre- and post-bazel hooks, with both local and remote plugins.

(Note, I haven't made the change to the pre-bazel API yet so that it operates on a single file passed as $1; will do that in a follow-up PR)

---

**Version bump**: Patch

### DIFF
--- a/cli/bazelisk/bazelisk.go
+++ b/cli/bazelisk/bazelisk.go
@@ -2,6 +2,7 @@ package bazelisk
 
 import (
 	"context"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -14,7 +15,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 )
 
-func Run(args []string) {
+func Run(args []string, output io.Writer) (int, error) {
 	// If we were already invoked via bazelisk, then set the bazel version to
 	// the next version appearing in the .bazelversion file so that bazelisk
 	// doesn't just invoke us again (resulting in an infinite loop).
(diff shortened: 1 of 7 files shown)


cli/cmd/bb/bb.go
```diff
@@ -2,29 +2,56 @@ package main
...
 	"github.com/buildbuddy-io/buildbuddy/cli/remotebazel"
 	"github.com/buildbuddy-io/buildbuddy/cli/sidecar"
+	"github.com/buildbuddy-io/buildbuddy/cli/terminal"
 	"github.com/buildbuddy-io/buildbuddy/cli/tooltag"
 	"github.com/buildbuddy-io/buildbuddy/cli/version"
 )
 
 func main() {
 	flag.Parse()
+	exitCode, err := run()
```
siggisim, line 22: instead of pulling this out into a `run()` function, what do you think leaving the logic in main() and bringing back this die method?

https://github.com/buildbuddy-io/buildbuddy/commit/a83ffc7e271088a027e5c186b143144523cc27dd#diff-0ddf6d0f5a686b3782a8c089af99f308f250aeea9741808012cd92e9c4e98f5dL31

cli/cmd/bb/bb.go
```diff
@@ -2,29 +2,56 @@ package main
...
+		log.Fatal(err)
+	}
+	os.Exit(exitCode)
+}
+
+func run() (exitCode int, err error) {
+	// Load plugins
+	plugins, err := plugin.LoadAll()
+	if err != nil {
+		return 0, err
```
siggisim, line 33: nit: i think it doesn't matter right now (because we don't hit exit) - but i think we probably want a non-zero exit code in this case (and the errors below)

siggisim (approved):
looks good, just some nits!

cli/cmd/bb/bb.go
bduffany: I think the control flow is less error-prone if we have only one place in the program that can call `exit()`.

Usually we can look at a `defer` statement and say "this should get executed when the function returns," but if some function calls potentially call `exit` then we can no longer rely on that. So it's hard to tell whether a particular cleanup func will actually get called or not. (Even more basic than `defer`: we can usually look at a function call and say "the code after this call should execute once this function returns")

I ran into this when writing that `defer` in the block below -- it wasn't getting called and I was scratching my head.

cli/cmd/bb/bb.go
bduffany: made it `-1` instead.

(I didn't make a const for it, because I thought the made the code a bit unconventional -- it looks too much at a glance like we are returning something that is intended to be used when we actually want it to be discarded.)

cli/bazelisk/bazelisk.go
```diff
@@ -29,12 +30,53 @@ func Run(args []string) {
 	// Fetch releases, release candidates and Bazel-at-commits from GCS, forks from GitHub
 	repos := core.CreateRepositories(gcs, gcs, gitHub, gcs, gitHub, true)
 
+	// Temporarily redirect stdout/stderr so that we can capture the bazelisk
+	// output.
+	// Note, we might be able to use Bazel's "command.log" file instead, but
+	// would need to find a way to avoid a race condition if another Bazel
+	// command is run concurrently (since it will clear out command.log as
+	// soon as our bazelisk invocation below is complete and it is able to
+	// grap the workspace lock).
```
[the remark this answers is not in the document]
bduffany: Done

PR #2640, opened 2022-10-03
Enable test grid for `bazel coverage`

bduffany:
`bazel coverage` runs tests as well, so we can enable the test grid for it.

---

**Version bump**: Patch


tylerwilliams (approved):
LGTM

PR #2643, opened 2022-10-03
Refactor clickhouse table interface

luluz66:
1. TableClusterOptions() changed to a global function to reflect that
   this is not a per table config
2. Create getAllTables() function to be used in runMigrations()
3. Refactor tests to make it easier to add new tables


server/util/clickhouse/clickhouse.go
```diff
@@ -179,11 +188,16 @@ func (h *DBHandle) FlushInvocationStats(ctx context.Context, ti *tables.Invocati
...
 	log.Info("Auto-migrating clickhouse DB")
-	gdb = gdb.Set("gorm:table_options", (&Invocation{}).TableOptions())
-	if clusterOpts := (&Invocation{}).TableClusterOption(); clusterOpts != "" {
+	if clusterOpts := tableClusterOption(); clusterOpts != "" {
 		gdb = gdb.Set("gorm:table_cluster_options", clusterOpts)
 	}
-	return gdb.AutoMigrate(&Invocation{})
+	for _, t := range getAllTables() {
+		gdb = gdb.Set("gorm:table_options", t.TableOptions())
+		if err := gdb.AutoMigrate(&Invocation{}); err != nil {
```
bduffany, line 196: should this be `t` instead of `&Invocation{}`?

bduffany (approved):

server/util/clickhouse/clickhouse.go
luluz66: Done.

PR #2648, opened 2022-10-04
Use a temp file when writing through Set method.

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):

vadimberezniker:
Oops, disk.WriteFile already uses a temp file. Never mind...

PR #2654, opened 2022-10-04
Add test for composable cache read through.

vadimberezniker:
Disabled until we fix Writer and Reader.

---

**Version bump**: None


tylerwilliams (approved):
gonna enable before submitting?

vadimberezniker:
> gonna enable before submitting?

done

PR #2656, opened 2022-10-05
Use correct isolation when copying data during migration

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -44,6 +44,11 @@ type MigrationCache struct {
 	copyChan                    chan *copyData
 	copyChanFullWarningInterval time.Duration
 	numCopiesDropped            *int64
+
+	// TODO(Maggie): Clean up these fields when removing WithIsolation
+	// These fields describe the current isolation, as set by WithIsolation
+	remoteInstanceName string
+	cacheType          resource.CacheType
 }
 
 func Register(env environment.Env) error {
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #2656** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2656" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2656?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -714,7 +727,19 @@ func (mc *MigrationCache) logCopyChanFullInBackground() {
 func (mc *MigrationCache) copy(c *copyData) {
 	defer c.ctxCancel()
 
-	srcReader, err := mc.src.Reader(c.ctx, c.d, 0, 0)
+	cache, err := mc.WithIsolation(c.ctx, c.cacheType, c.remoteInstanceName)
```
vadimberezniker, line 730: You can add a (private) `withIsolation` method that returns a *MigrationCache instead of interfaces.Cache and call that instead to avoid needing to cast. `WithIsolation` can then also call the private `withIsolation` method.

vadimberezniker (approved):

enterprise/server/backends/migration_cache/migration_cache.go

PR #2663, opened 2022-10-05
Allow an invocation's pattern to be overridden by build metadata / workspace status

siggisim:
cc @brentleyjones 

---

**Version bump**: Patch

### DIFF
--- a/docs/guide-metadata.md
+++ b/docs/guide-metadata.md
@@ -174,7 +174,45 @@ build --workspace_status_command=$(pwd)/workspace_status.sh
 
 Then you'll need to add a `workspace_status.sh` file to the root of your workspace that prints `HOST yourhost`.
 
-### Custom Links
+## Pattern
+
+By default a build's pattern is determined by bazel command that is run.
+
+If the bazel command is:
+
+```
+bazel build //...
+```
+
+The pattern would be:
+
+```
+//...
+```
+
+You can override this using build metadata or workspace info.
+
+### Build metadata
+
+You can provide a custom pattern with Bazel's build_metadata flag with the key `PATTERN`. You can do this by adding the flag to your bazel invocations:
+
+```
+--build_metadata=PATTERN=yourpattern
+```
+
+### Workspace info
+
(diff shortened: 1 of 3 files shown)


docs/guide-metadata.md
```diff
@@ -174,7 +174,45 @@ build --workspace_status_command=$(pwd)/workspace_status.sh
...
+bazel build //...
+```
+
+The pattern would be:
+
+```
+//...
+```
+
+You can override this using build metadata or workspace info.
```
bduffany, line 193: Should we add an example here and maybe say something like:

This can be used for more advanced use cases, such as a tool that generates a `bazel build` command with a series of patterns, but wants to provide a more user-friendly version that is equivalent.

bduffany (approved):

docs/guide-metadata.md
siggisim: Added a note on it being useful for more advanced use-cases.

bduffany (approved):

server/build_event_protocol/event_parser/event_parser.go
```diff
@@ -315,6 +315,8 @@ func fillInvocationFromWorkspaceStatus(workspaceStatus *build_event_stream.Works
 			invocation.Host = item.Value
 		case "HOST":
 			invocation.Host = item.Value
+		case "PATTERN":
```
brentleyjones, line 318: Do we want the key to be something more specific, like `TARGET_PATTERN`? This matches the way that Bazel refers to them. And because this probably won't be the actual pattern, maybe `TARGET_PATTERN_OVERRIDE`?

siggisim: I'd rather match the BuildBuddy "pattern" naming that we use throughout than making it too Bazel specific: https://github.com/buildbuddy-io/buildbuddy/blob/bab9c19635c2d53c4e158c51a4e566a6691d34be/proto/api/v1/invocation.proto#L52

HOST and USER are also "overrides" and I'd rather not change those because it wouldn't be backwards compatible.

PR #2667, opened 2022-10-05
Show app version in the footer if it's available

siggisim:
This is helpful for debugging which app version a user is using.

---

**Version bump**: Patch


tylerwilliams (approved):
have always wanted this!

PR #2669, opened 2022-10-05
Don't sync pebble WAL so often.

vadimberezniker:
---

**Version bump**: None


tylerwilliams (approved):
TIL about pebble.Sync and pebble.NoSync

PR #2672, opened 2022-10-06
Bug fixes / improvements for FetchBlob

bduffany:
Bug fixes:
* If the remote URI returns an HTTP error status, return an error instead of reading the response body. This can cause empty responses to be returned.
* If we fail to fetch a particular URI, continue to the next URI instead of returning `Aborted`.
* Make sure we call `Close()` on the response body (documentation says the caller is responsible for calling `Close()`).

Improvements:
* Record the last fetch error and return it in the status message, in addition to logging. (We still return a gRPC code of NotFound so this should functionally be a NOP, but improve debuggability.)
* Make sure all logged error messages include the URI.
* Do a checksum on our side (since it's simple enough to do) so that we don't store useless blobs into the cache and so that the client doesn't have to download a bad payload only to fail the checksum on their side.

---

**Version bump**: Patch

### DIFF
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -132,43 +132,17 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
 
 	// Keep track of the last fetch error so that if we fail to fetch, we at
 	// least have something we can return to the client.
-	var fetchErr error
-	setFetchErr := func(err error) {
-		fetchErr = err
-		log.Warning(status.Message(err))
-	}
+	var lastFetchErr error
 
 	for _, uri := range req.GetUris() {
 		_, err := url.Parse(uri)
 		if err != nil {
 			return nil, status.InvalidArgumentErrorf("unparsable URI: %q", uri)
 		}
-		rsp, err := httpClient.Get(uri)
+		blobDigest, err := mirrorToCache(ctx, cache, httpClient, uri, expectedSHA256)
 		if err != nil {
-			setFetchErr(status.UnavailableErrorf("failed to fetch %q: HTTP GET failed: %s", uri, err))
-			continue
-		}
-		defer rsp.Body.Close()
-		if rsp.StatusCode < 200 || rsp.StatusCode >= 400 {
-			setFetchErr(status.UnavailableErrorf("failed to fetch %q: HTTP %s", uri, err))
-			continue
-		}
-		data, err := io.ReadAll(rsp.Body)
-		if err != nil {
-			setFetchErr(status.UnavailableErrorf("failed to read response body from %q: %s", uri, err))
-			continue
-		}
-		blobDigest, err := digest.Compute(bytes.NewReader(data))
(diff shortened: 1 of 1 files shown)


server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -158,7 +179,10 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
 	}
 
 	return &rapb.FetchBlobResponse{
-		Status: &statuspb.Status{Code: int32(gcodes.NotFound)},
+		Status: &statuspb.Status{
+			Code:    int32(gcodes.NotFound),
+			Message: status.Message(fetchErr),
+		},
```
tylerwilliams, line 185: can you just do Status: fetchErr.GRPCStatus()?

tylerwilliams: or just add a comment here explaining you intentionally always want to return NotFound for reason x

bduffany: Done

tylerwilliams (approved):

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -126,28 +129,46 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
 		}
 	}
 	httpClient := timeoutHTTPClient(ctx, req.GetTimeout())
+
+	// Keep track of the last fetch error so that if we fail to fetch, we at
+	// least have something we can return to the client.
+	var fetchErr error
+	setFetchErr := func(err error) {
```
vadimberezniker, line 136: This code would be slightly cleaner if you moved the fetching code for a single URL from inside the loop into a separate func.

bduffany: good call! done

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -126,28 +129,20 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
...
-		reader := bytes.NewReader(data)
-		blobDigest, err := digest.Compute(reader)
-		if err != nil {
-			log.Warningf("Error computing object digest: %s", err.Error())
-			continue
-		}
-		if err := cache.Set(ctx, blobDigest, data); err != nil {
-			log.Warningf("Error inserting object into cache: %s", err.Error())
+			lastFetchErr = err
+			log.Warning(status.Message(err))
```
vadimberezniker, line 145: Can you decorate this error message a bit so that it's clear where the error is coming from?
Something like log.Warningf("could not mirror URL to cache: %s", err)

In general I recommend ever having undecorated logs like log.Warning(err).

vadimberezniker (approved):

server/remote_asset/fetch_server/fetch_server.go
bduffany: Done

PR #2678, opened 2022-10-07
Fix symlink paths

bduffany:
Symlinks were being created relative to the executor's working directory rather than relative to the parent `Directory` containing the `SymlinkNode`.

---

**Version bump**: Patch


siggisim (approved):

tylerwilliams (approved):
this lgtm but would be great to add a test

tylerwilliams (approved):
Thanks!

PR #2687, opened 2022-10-11
CLI: Support live bazel output processing

bduffany:
This PR adds support for a new plugin hook `handle_bazel_output.sh` which allows plugins to intercept the output of Bazel (or the previous plugin in the pipeline) and optionally transform it.

We support multiple plugins defining this hook, in which case there will be a "pipeline" of plugins. Each plugin receives the output from the previous plugin.

Example: the `go_highlight` plugin can now directly highlight output, instead of copying output lines to the end of the build.

image

---

**Version bump**: Patch


cli/plugin/plugin.go
```diff
@@ -305,11 +306,93 @@ func (p *Plugin) PostBazel(bazelOutputPath string) error {
...
+// It returns the original reader if no output handler is configured for this
+// plugin.
+//
+// See cli/example_plugins/go_highlight/handle_bazel_output.sh for an example.
+func (p *Plugin) Pipe(r io.Reader) (io.Reader, error) {
+	path, err := p.Path()
+	if err != nil {
+		return nil, err
+	}
+	scriptPath := filepath.Join(path, "handle_bazel_output.sh")
```
bduffany, line 329: @siggisim I went with `handle_bazel_output.sh` instead of `on_output.sh` because I thought the latter made it sort of seem like the script would be called for each output line. But this new name might have the same confusion? :shrug: Open to suggestions

siggisim (approved):

PR #2717, opened 2022-10-13
Change the way IDs (invocation, request) are propagated to the logger.

vadimberezniker:
Instead of extracting a hardcoded set of IDs at log time, allow identifiers to be added to the context.

This will enable us to enrich log events with execution_id, raft replica id, etc.

---

**Version bump**: None


tylerwilliams (approved):
nice

PR #2727, opened 2022-10-14
Only log migration FindMissing discrepancies if we are double reading

maggie-lou:
* Fix bug where we'd log a FindMissing discrepancy even when we weren't double reading, so the destination cache array was always empty
* Add metrics to track how many double reads are successful to get a better sense of what % of reads are unsuccessful


maggie-lou:
* master
  * **PR #2727** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2727" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2727?utm_source=stack-comment).

vadimberezniker (approved):

PR #2729, opened 2022-10-14
Check in canary, useful for debugging

tylerwilliams:


server/util/canary/canary.go
```diff
@@ -0,0 +1,74 @@
...
+// StartWithLateFn starts a canary that will call lateFn every expectedDuration
+// period (if not cancelled) and once when cancelled.
+func StartWithLateFn(expectedDuration time.Duration, lateFn CanaryFunc, doneFn CanaryFunc) CancelFunc {
+	if !*enableCanaries {
+		return func() {}
+	}
+
+	done := make(chan struct{})
+	cancel := func() {
+		close(done)
```
vadimberezniker, line 28: can you guard against cancel being called multiple times?

server/util/canary/canary.go
```diff
@@ -0,0 +1,74 @@
+package canary
```
vadimberezniker, line 1: watchdog is another name to consider if you're looking for alternatives

vadimberezniker (approved):

server/util/canary/canary.go
tylerwilliams: Done.

server/util/canary/canary.go
tylerwilliams: Yeah I considered that, but I always think of a watchdog timer where you have to call reset on it or it'll restart your thing. This is sort of the opposite of that in a way?

PR #2730, opened 2022-10-14
Prevent disk cache race condition when cleaning up temp files

maggie-lou:
The disk cache tests have been really flaky for me due to a race condition where temp files are being deleted by the initialization function at the same time as when we're trying to [set data in unit tests](https://github.com/buildbuddy-io/buildbuddy/blob/master/server/util/disk/disk.go#L81)

### DIFF
--- a/server/backends/disk_cache/disk_cache.go
+++ b/server/backends/disk_cache/disk_cache.go
@@ -47,6 +47,10 @@ const (
 	// check the cache size.
 	janitorCheckPeriod = 100 * time.Millisecond
 
+	// staleFilePeriod is the timeframe after which an empty or temp file will be considered stale and eligible
+	// for automatic deletion
+	staleFilePeriod = 15 * time.Minute
+
 	DefaultPartitionID       = "default"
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #2730** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2730" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2730?utm_source=stack-comment).

vadimberezniker:
We should add a check that only deletes these stale files if they're older than a certain amount (say 15 minutes). 
I think the issue you are running into in tests can also happen under normal operation.

maggie-lou:
> We should add a check that only deletes these stale files if they're older than a certain amount (say 15 minutes). I think the issue you are running into in tests can also happen under normal operation.

Done

vadimberezniker (approved):
Thanks for finding this!

PR #2732, opened 2022-10-17
CLI: Support interactive rebuilding with `-w` or `--watch` flag.

bduffany:
Using [godemon](https://github.com/bduffany/godemon) as the file watcher for now since it embeds easily in Go. We might consider switching to [ibazel](https://github.com/bazelbuild/bazel-watcher) in the future, but we might want to either fork or contribute some upstream patches since it has some undesired behavior, e.g. https://github.com/bazelbuild/bazel-watcher/issues/305. We'd also need to figure out how to embed it in our app, since the Go API currently isn't visible externally.

---

**Version bump**: Patch

### DIFF
--- a/cli/watcher/watcher.go
+++ b/cli/watcher/watcher.go
@@ -1,0 +1,58 @@
+package watcher
+
+import (
+	"os"
+	"syscall"
+
+	"github.com/bduffany/godemon"
+	"github.com/buildbuddy-io/buildbuddy/cli/arg"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/google/shlex"
+)
+
+// Watch looks for the -w or --watch flag, and if present it reinvokes the
+// CLI as a subprocess on changes to source files.
+func Watch() (exitCode int, err error) {
+	args := os.Args
+
+	idx := -1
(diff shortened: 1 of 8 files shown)


cli/watcher/watcher.go
```diff
@@ -0,0 +1,58 @@
...
+	// Especially useful for development but can also be used to pull in newer
+	// godemon features.
+	if bin := os.Getenv("GODEMON_BINARY_PATH"); bin != "" {
+		if err := syscall.Exec(bin, argv, os.Environ()); err != nil {
+			return -1, err
+		}
+		panic("unreachable")
+	}
+
+	godemon.Main(argv)
```
siggisim, line 55: What paths does godemon watch for changes on?

siggisim (approved):

cli/watcher/watcher.go
bduffany: Any path under the bazel workspace that's not in the [default ignore list](https://github.com/bduffany/godemon/blob/master/default_ignore.go) or in your top-level `.gitignore` file.

Adding support for nested `.gitignore` is something I plan on adding soon.

PR #2737, opened 2022-10-18
Release config settings now use extant directories

tempoz:
There is no `/buildbuddy` directory in the release image, so we create a convenience symlink.

---

**Version bump**: Patch

### DIFF
--- a/server/cmd/buildbuddy/BUILD
+++ b/server/cmd/buildbuddy/BUILD
@@ -31,7 +31,7 @@ container_image(
     base = "@buildbuddy_go_image_base//image",
     symlinks = {
         "config.yaml": "app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy/config/buildbuddy.release.yaml",
-        "buildbuddy": "app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy",
+        "buildbuddy": "tmp",
     },
     tags = ["no-cache"],
     visibility = ["//visibility:public"],
--- a/server/cmd/buildbuddy/BUILD
+++ b/server/cmd/buildbuddy/BUILD
@@ -31,6 +31,7 @@ container_image(
     base = "@buildbuddy_go_image_base//image",
     symlinks = {
         "config.yaml": "app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy/config/buildbuddy.release.yaml",
+        "buildbuddy": "app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy",
     },
     tags = ["no-cache"],
     visibility = ["//visibility:public"],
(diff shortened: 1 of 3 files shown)


tempoz:
* master
  * **PR #2737** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2737" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2737?utm_source=stack-comment).

tempoz:
Related: buildbuddy-io/buildbuddy-internal#1751

config/buildbuddy.release.yaml
```diff
@@ -1,10 +1,10 @@
...
-  data_source: "sqlite3:///buildbuddy/buildbuddy.db"
+  data_source: "sqlite3:///app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy/buildbuddy.db"
 storage:
   ttl_seconds: 2592000 # 30 days
   disk:
-    root_directory: "/buildbuddy/blobs/"
+    root_directory: "/app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy/blobs/"
 cache:
   disk:
-    root_directory: "/buildbuddy/cache/"
```
bduffany, line 10: I think we `mkdir -p` on these directories, so they shouldn't need to exist in the image.

Is there some error we're seeing that you think this would fix?

bduffany: NVM, just read the linked issue.

I think maybe we should just `mkdir -p` on the sqlite parent dir.

tempoz: Swapped to using a symlink, as we discussed offline.

server/cmd/buildbuddy/BUILD
```diff
@@ -31,6 +31,7 @@ container_image(
     base = "@buildbuddy_go_image_base//image",
     symlinks = {
         "config.yaml": "app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy/config/buildbuddy.release.yaml",
+        "buildbuddy": "app/server/cmd/buildbuddy/buildbuddy.runfiles/buildbuddy",
```
bduffany, line 34: nit: Can we symlink `buildbuddy` to `/tmp` instead? Asking because runfiles are read only data deps (IIRC) and there might be a chance of clobbering something if we store arbitrary stuff there.

bduffany (approved):

server/cmd/buildbuddy/BUILD
bduffany: (I tested `/tmp` locally and it seems to work)

tempoz: Do we care about it working on restart of the container? `/tmp` destroys its contents on restart on most machines. Could we use `/home/nonroot` instead?

tempoz: nvm, discussed offline, using `/tmp`

PR #2739, opened 2022-10-18
Add metrics when writing to clickhouse

luluz66:


server/util/clickhouse/clickhouse.go
```diff
@@ -182,19 +184,30 @@ func (h *DBHandle) FlushInvocationStats(ctx context.Context, ti *tables.Invocati
...
 		}
 		if errors.Is(res.Error, syscall.ECONNRESET) || errors.Is(res.Error, syscall.ECONNREFUSED) {
 			// Retry since it's an transient error.
-			lastError = res.Error
 			log.Infof("attempt to write invocation (id: %q) to clickhouse failed: %s", res.Error, ti.InvocationID)
 			continue
 		} else {
-			return res.Error
+			break
 		}
```
tylerwilliams, line 197: I think you can simplify this like:

```
		if errors.Is(res.Error, syscall.ECONNRESET) || errors.Is(res.Error, syscall.ECONNREFUSED) {
			log.Infof("attempt to write invocation (id: %q) to clickhouse failed: %s", res.Error, ti.InvocationID)
			continue
		}
		break
}

```

luluz66: Done

tylerwilliams (approved):

PR #2741, opened 2022-10-18
CLI: POC for Go plugin w/ external deps

bduffany:
Rewrite `go_highlight` in Go to show how language-specific plugins can be implemented without any changes to the current API.

The approach taken is to use an init script which is invoked in `pre_bazel.sh`. The init script computes a hash of the plugin binary's deps, and if the plugin binary doesn't exist or the hash mismatches the hash from the last build, the plugin is rebuilt.

The init script adds about 20ms of overhead to each `bb` invocation, which can add up if there are lots of plugins. In the future, we may want to consider making `init.sh` a standard hook so that we can run all init scripts in parallel -- compared to `pre_bazel.sh` which are run synchronously.

---

**Version bump**: Patch


cli/example_plugins/go_highlight/.gitignore
```diff
@@ -0,0 +1,2 @@
+/go-highlight-bb-plugin
```
siggisim, line 1: should we include the hash in the filename and store this in the user temp directory under a plugins subdir?

cli/example_plugins/go_highlight/handle_bazel_output.sh
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-exec python3 ./handle_bazel_output.py "$@"
+exec ./go-highlight-bb-plugin "$@"
```
siggisim, line 2: i kinda liked the simplicity of the python one as an example for folks that want to make their own highlighters - is there a big performance improvement doing it in golang?

is there another plugin where having golang support would be more helpful?

cli/example_plugins/go_highlight/.gitignore
bduffany: Discussed offline -- the CLI can provide a PLUGIN_CACHE_DIR where plugins can save build artifacts like this, so that we don't need this .gitignore stuff.

So `init.sh` would output the binary and SHA to `$PLUGIN_CACHE_DIR/{go-highlight-plugin,plugin_deps.sum}` and then `handle_bazel_output.sh` can expect to find those there.

cli/example_plugins/go_highlight/handle_bazel_output.sh
bduffany: There's not a huge value to having this particular one written in Go tbh. This is mostly a POC that we don't have to submit; mainly wanted to gather your thoughts on the overall approach.

I was planning on having a more general "find-and-replace" plugin that would probably be best written in Go since it'll handle a bunch of regexes that run on each line.

Off-topic (can chat more about this offline), but that one would be interesting because it would need a way to pass config values to plugins, maybe something like:

```
- path: cli/example_plugins/regexp_replace
  with:
    - find: ^(.*.go:\d+:\d+:)(.*)
      replace: \x1b[33m$1\x1b[m$2
```

PR #2743, opened 2022-10-18
Improve FinalizeInvocation logging

luluz66:
1. Add CtxInfo and CtxWarning
2. Add invocation_id to the context and use log.Ctx*** for logging.
3. Add additional logging after we successfully enqueue stats recorder task.

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -281,36 +281,39 @@ func (r *statsRecorder) handleTask(ctx context.Context, task *recordStatsTask) {
 	// unnecessarily throttled.
 	time.Sleep(time.Until(task.createdAt.Add(*cacheStatsFinalizationDelay)))
 	ti := &tables.Invocation{InvocationID: task.invocationJWT.id, Attempt: task.invocationJWT.attempt}
+	ctx = log.EnrichContext(ctx, log.InvocationIDKey, task.invocationJWT.id)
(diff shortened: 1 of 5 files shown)


server/util/log/log.go
```diff
@@ -302,6 +303,15 @@ func Infof(format string, args ...interface{}) {
 	log.Info().Msgf(format, args...)
 }
 
+// CtxInfo logs to the INFO log.
+// Logs are enriched with information from the context
+// (e.g. invocation_id, request_id)
+func CtxInfo(ctx context.Context, message string) {
```
vadimberezniker, line 309: Can you add the other funcs (CtxError, CtxDebug, etc) for completeness?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -623,6 +626,7 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 	}
 
 	ctx, cancel := background.ExtendContextForFinalization(e.ctx, 10*time.Second)
+	ctx = log.EnrichContext(ctx, log.InvocationIDKey, iid)
```
vadimberezniker, line 629: Can you move this to OpenChannel, that way the invocation_id will be available through out the entire lifecycle.

vadimberezniker (changes requested):

server/util/log/log.go
luluz66: Done

server/build_event_protocol/build_event_handler/build_event_handler.go

vadimberezniker (approved):

PR #2746, opened 2022-10-18
Use chunked event logs in released FOSS image

tempoz:
This feature was previously not enabled in the released images, but is standard at this point.

---

**Version bump**: Patch

### DIFF
--- a/config/buildbuddy.release.yaml
+++ b/config/buildbuddy.release.yaml
@@ -5,6 +5,7 @@ storage:
   ttl_seconds: 2592000 # 30 days
   disk:
     root_directory: "/buildbuddy/blobs/"
+  enable_chunked_event_logs: true
 cache:
   disk:
     root_directory: "/buildbuddy/cache/"


tempoz:
* master
  * **PR #2737** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2737" target="_blank">Graphite</a> 
    * **PR #2746** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2746" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2746?utm_source=stack-comment).

tylerwilliams (approved):
LGTM

(can you double check with @brandon that the UI in the OSS version has everything needed to support this?)

siggisim:
Should we just flip the flag default to true instead?

tempoz:
> Should we just flip the flag default to true instead?

Maaaaaybe? My understanding is that best practice in go is to only use default false boolean flags, so that they are specified like options on the command line as opposed to assigning to them.

tylerwilliams:
+1 to flipping the flag instead

tempoz:
kk, I'll do that

PR #2754, opened 2022-10-19
smaller leases

tylerwilliams:
- Lease DB only for metadata changes
- Return unavailable so it can be retried
- Remove unused import

Note: I still see see split lease times over a second. I believe this is because scanning (and proto deserializing) so many files that have inline metadata content is slow.

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -1237,6 +1237,12 @@ func (sm *Replica) Writer(ctx context.Context, header *rfpb.Header, fileRecord *
 		}
 		defer db.Close()
 
+		// Re-validate the header to ensure the range did not split before
+		// we got to Commit().
+		if err := sm.validateRange(header); err != nil {
+			return err
+		}
+
 		batch := db.NewBatch()
 		now := time.Now()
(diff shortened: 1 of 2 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -1208,31 +1212,31 @@ func (sm *Replica) storedFileWriter(ctx context.Context, fileRecord *rfpb.FileRe
...
+		db, err := sm.leaser.DB()
+		if err != nil {
+			if sm.IsSplitting() {
+				return status.OutOfRangeErrorf("%s: id %d ", constants.RangeSplittingMsg, header.GetRangeId())
+			}
+			return err
+		}
+		defer db.Close()
+
 		batch := db.NewBatch()
```
vadimberezniker, line 1240: do we need to revalidate the header? it's possible the range descriptor has changed between the start and the end of the write (i.e. a split finished in the meantime)

vadimberezniker (approved):

enterprise/server/raft/replica/replica.go
tylerwilliams: Done.

vadimberezniker (approved):

PR #2756, opened 2022-10-19
Use CtxInfo in clickhouse logging

luluz66:
Also fix a logging bug.


server/util/clickhouse/clickhouse.go
```diff
@@ -187,7 +187,7 @@ func (h *DBHandle) FlushInvocationStats(ctx context.Context, ti *tables.Invocati
 		lastError = res.Error
 		if errors.Is(res.Error, syscall.ECONNRESET) || errors.Is(res.Error, syscall.ECONNREFUSED) || errors.Is(res.Error, context.DeadlineExceeded) {
 			// Retry since it's an transient error.
-			log.Infof("attempt to write invocation (id: %q) to clickhouse failed: %s", res.Error, ti.InvocationID)
+			log.CtxInfof(ctx, "attempt to write invocation to clickhouse failed: %s", ti.InvocationID)
```
vadimberezniker, line 190: i think arg is supposed to be `res.Error`

vadimberezniker (approved):

server/util/clickhouse/clickhouse.go
luluz66: Absolutely! Done.

PR #2761, opened 2022-10-20
Drain migration cache copy queue on shutdown

maggie-lou:
---

**Version bump**: TODO


maggie-lou:
* master
  * **PR #2761** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2761" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2761?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -761,6 +761,11 @@ func (mc *MigrationCache) copyDataInBackground() error {
 
 		select {
 		case <-mc.quitChan:
+			// Drain copy channel on shutdown
+			close(mc.copyChan)
```
vadimberezniker, line 765: I think there's a small window where we might try to write on a closed channel. 
We start shutting down and concurrently there are still some remaining requests that have not called into the cache yet (or have already entered into the cache code but may take some small amount of time before scheduling a copy, for example during FIndMissing handling). If we close the channel here, the app will panic when we try to enqueue another copy.

You could drain the queue by reading from the channel until there's nothing left in the queue. We might drop some last second copies, but I think it's ok if this is best effort.

vadimberezniker (changes requested):

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Ah right - fixed

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -761,6 +761,11 @@ func (mc *MigrationCache) copyDataInBackground() error {
 
 		select {
 		case <-mc.quitChan:
+			// Drain copy channel on shutdown
+			for len(mc.copyChan) > 0 {
```
vadimberezniker, line 765: can you leave a short comment why we're draining the channel in this manner. it's always good to leave a comment when doing something slightly unusual.

vadimberezniker (approved):

PR #2762, opened 2022-10-20
In migration cache, copy data on FindMissing

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -316,20 +316,17 @@ func (mc *MigrationCache) FindMissing(ctx context.Context, resources []*resource
 				log.Warningf("Migration FindMissing diff for digests %v: src %v, dest %v", resources, srcMissing, dstMissing)
 			}
 
+			instanceName := resources[0].GetInstanceName()
+			cacheType := resources[0].GetCacheType()
 			for _, d := range missingOnlyInDest {
-				d := d
-				eg.Go(func() error {
-					r := &resource.ResourceName{
-						Digest:       d,
-						InstanceName: mc.remoteInstanceName,
-						Compressor:   repb.Compressor_IDENTITY,
-						CacheType:    mc.cacheType,
-					}
-					mc.sendNonBlockingCopy(ctx, r, false)
-					return nil
-				})
+				r := &resource.ResourceName{
+					Digest:       d,
+					InstanceName: instanceName,
+					Compressor:   repb.Compressor_IDENTITY,
+					CacheType:    cacheType,
+				}
+				mc.sendNonBlockingCopy(ctx, r, false /*=onlyCopyMissing*/)
 			}
-			eg.Wait()
 		}
 	}
 
@@ -373,7 +370,7 @@ func (mc *MigrationCache) GetMulti(ctx context.Context, resources []*resource.Re
 	}
 
 	for _, r := range resources {
-		mc.sendNonBlockingCopy(ctx, r, true)
+		mc.sendNonBlockingCopy(ctx, r, true /*=onlyCopyMissing*/)
 	}
 
 	// Return data from source cache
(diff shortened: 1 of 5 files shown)


maggie-lou:
* master
  * **PR #2762** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2762" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2762?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -302,17 +302,33 @@ func (mc *MigrationCache) FindMissing(ctx context.Context, resources []*resource
...
-
-			if mc.logNotFoundErrors {
+			if mc.logNotFoundErrors || len(missingOnlyInSrc) != 0 {
 				log.Warningf("Migration FindMissing diff for digests %v: src %v, dest %v", resources, srcMissing, dstMissing)
 			}
+
+			for _, d := range missingOnlyInDest {
+				d := d
+				eg.Go(func() error {
+					mc.sendNonBlockingCopy(ctx, &resource.ResourceName{
```
vadimberezniker, line 322: What do you think about adding a parameter to sendNonBlockingCopy to skip the contains check? Then you wouldn't need to fire off goroutines for every digest since you already know they are not in the destination cache and the a plain channel enqueue is very fast.

vadimberezniker (changes requested):

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: I like that - added

maggie-lou: Added a couple additional debug logs too

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -302,17 +302,34 @@ func (mc *MigrationCache) FindMissing(ctx context.Context, resources []*resource
...
 		} else {
 			metrics.MigrationNotFoundErrorCount.With(prometheus.Labels{metrics.CacheRequestType: "findMissing"}).Inc()
-
-			if mc.logNotFoundErrors {
+			if mc.logNotFoundErrors || len(missingOnlyInSrc) != 0 {
 				log.Warningf("Migration FindMissing diff for digests %v: src %v, dest %v", resources, srcMissing, dstMissing)
 			}
+
+			for _, d := range missingOnlyInDest {
+				d := d
```
vadimberezniker, line 320: you can remove the async code now that sendNonBlockingCopy is a simple channel enqueue. this loop should be fast w/o the goroutines.

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -302,17 +302,34 @@ func (mc *MigrationCache) FindMissing(ctx context.Context, resources []*resource
...
+				eg.Go(func() error {
+					r := &resource.ResourceName{
+						Digest:       d,
+						InstanceName: mc.remoteInstanceName,
+						Compressor:   repb.Compressor_IDENTITY,
+						CacheType:    mc.cacheType,
+					}
+					mc.sendNonBlockingCopy(ctx, r, false)
```
vadimberezniker, line 328: it's a good practice to annotate arguments that are unclear at the call site using a comment with the parameter name, i.e. `false /*=onlyCopyMissing*/`

server/remote_cache/digest/digest.go
```diff
@@ -343,6 +343,40 @@ func ElementsMatch(s1 []*repb.Digest, s2 []*repb.Digest) bool {
...
+	for _, d := range s1 {
+		s1Set[d] = struct{}{}
+	}
+
+	s2Set := make(map[*repb.Digest]struct{}, len(s2))
+	for _, d := range s2 {
+		s2Set[d] = struct{}{}
+
+		_, inS1 := s1Set[d]
+		if !inS1 {
```
vadimberezniker, line 365: nit: in go you can inline the variable definition into the if clause if it's not needed outside the if statement (i.e. `if _, inS1 := s1Set[d]; !inS1 {`

vadimberezniker (approved):

server/remote_cache/digest/digest.go
maggie-lou: Done

enterprise/server/backends/migration_cache/migration_cache.go

PR #2787, opened 2022-10-25
Gossip per-node partition usage info.

vadimberezniker:
Moved usage gossip pub/sub to a separate package so that the data can be used indepndently by the driver & the evictor.

---

**Version bump**: None


enterprise/server/raft/replica/replica.go
```diff
@@ -166,16 +167,49 @@ func (sm *Replica) Usage() (*rfpb.ReplicaUsage, error) {
...
 		}
-		estimatedBytesUsed += sizeBytes
 	}
-	ru.EstimatedDiskBytesUsed = estimatedBytesUsed
-	return ru, nil
+
+	var partitionUsages []*rfpb.PartitionMetadata
+	for _, u := range usageByPartition {
+		partitionUsages = append(partitionUsages, u)
+	}
```
tylerwilliams, line 191: Not against doing it this way, but it feels like it might be inefficient to scan the whole range like this (lots of proto deserialization etc). I was thinking we could use this: https://pkg.go.dev/github.com/cockroachdb/pebble#DB.EstimateDiskUsage, and if we tracked the set of active partitions in a region we could just call that once for each region?

tylerwilliams (changes requested):

PR #2792, opened 2022-10-25
Expose secrets to trusted workflows

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -64,6 +64,7 @@ var (
 	workflowsCIRunnerBazelCommand = flag.String("remote_execution.workflows_ci_runner_bazel_command", "", "Bazel command to be used by the CI runner.")
 	workflowsLinuxComputeUnits    = flag.Int("remote_execution.workflows_linux_compute_units", 3, "Number of BuildBuddy compute units (BCU) to reserve for Linux workflow actions.")
 	workflowsMacComputeUnits      = flag.Int("remote_execution.workflows_mac_compute_units", 3, "Number of BuildBuddy compute units (BCU) to reserve for Mac workflow actions.")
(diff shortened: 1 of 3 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -64,6 +64,7 @@ var (
 	workflowsCIRunnerBazelCommand = flag.String("remote_execution.workflows_ci_runner_bazel_command", "", "Bazel command to be used by the CI runner.")
 	workflowsLinuxComputeUnits    = flag.Int("remote_execution.workflows_linux_compute_units", 3, "Number of BuildBuddy compute units (BCU) to reserve for Linux workflow actions.")
 	workflowsMacComputeUnits      = flag.Int("remote_execution.workflows_mac_compute_units", 3, "Number of BuildBuddy compute units (BCU) to reserve for Mac workflow actions.")
+	workflowsEnableSecrets        = flag.Bool("remote_execution.workflows_enable_secrets", false, "Whether to include secrets for trusted workflows.")
```
bduffany, line 67: @tylerwilliams FYI I put this behind a flag because the failing test made me realize that workflows would fail with "secrets were requested but secret service is missing" during the rollout. Maybe after tomorrow's release after all apps are rolled out we could push just this flag change.

tylerwilliams: This was why I had added "&& secretService != nil)" to the code when it was in the other location. Do you a prefer another flag (yaf) to doing that?

bduffany: ah I see how that works. Removed the flag and tweaked that check a bit -- wdyt?

bduffany: err, the logic I had in mind was wrong, need to tweak it a bit more.

tylerwilliams (approved):
u nasty

tylerwilliams (approved):
nice

PR #2793, opened 2022-10-25
Move attempt count increment into the claim script.

vadimberezniker:
This makes the operation atomic.

It was previously possible for the claim task to succeed but the increment to fail leading to a task that could no longer be claimed.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -130,6 +130,8 @@ var (
 			return 11
 		end
 
+		redis.call("hincrby", KEYS[1], "attemptCount", 1)
+
 		return redis.call("hset", KEYS[1], "claimed", "1") 
 		`)
 	// Claim field is removed only if it's present.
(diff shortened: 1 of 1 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -130,6 +130,8 @@ var (
 			return 11
 		end
 
+		redis.call("hincrby", KEYS[1], "attemptCount", 1)
```
bduffany, line 133: sanity check, will this do the right thing if this particular redis.call fails, or if this succeeds but the redis.call right afterwards fails?

bduffany (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
vadimberezniker: Practically speaking, these can only fail due to an internal Redis error. If that were to occur, redis.call would terminate the script and return the error to the app. The worst case that can happen is that we incorrectly increment the attempt count by one if the redis.call below fails, but the odds of that actually happening are virtually zero.

tylerwilliams:
such a good find :)

PR #2799, opened 2022-10-26
Move secrets RPCs back under admin only for now

tylerwilliams:


bduffany (approved):

bduffany:
fyi updated the UI as well in https://github.com/buildbuddy-io/buildbuddy/pull/2798

PR #2804, opened 2022-10-26
Add a Shutdown function for build event handler

luluz66:
Create a child context with Cancel function when we open a channel; and
put the cancel function in a map in build event handler. When the server
receives shutdown signal, we will call cancel functions.

In BuildEventServer, when the context from the channel is done, we will
return UnavailableError to signal bazel to retry.


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -149,6 +157,13 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
 	}
 }
 
+func (b *BuildEventHandler) Stop() {
+	for iid, cancelFn := range b.cancelFnsByInvID {
+		log.Infof("Cancel Event Channel for invocation %q", iid)
```
tylerwilliams, line 162: how about something like "Cancelling invocation %q because server received shutdown signal")

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -92,9 +92,10 @@ var (
 )
 
 type BuildEventHandler struct {
-	env           environment.Env
-	statsRecorder *statsRecorder
-	openChannels  *sync.WaitGroup
+	env              environment.Env
+	statsRecorder    *statsRecorder
+	openChannels     *sync.WaitGroup
+	cancelFnsByInvID map[string]func()
```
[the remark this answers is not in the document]
tylerwilliams: you can run the tests with --@io_bazel_rules_go//go/config:race and it will usually flag stuff like this (if there's a test that exercises it).

In this case I think you will need a mutex because you're writing the map from one goroutine (the RPC request goroutine) and you're reading it from another (the healthchecker shutdown goroutine)

tylerwilliams (changes requested):

server/build_event_protocol/build_event_handler/build_event_handler.go
vadimberezniker, line 98: you can use the context.CancelFunc alias here, it gives more clarity to this function

server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -104,20 +104,25 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 			defer channel.Close()
 		}
 
-		if err := channel.HandleEvent(in); err != nil {
-			if status.IsAlreadyExistsError(err) {
-				log.Warningf("AlreadyExistsError handling event; this means the invocation already exists and may not be retried: %s", err)
-				return err
+		select {
+		case <-channel.Context().Done():
+			return status.UnavailableError("context cancelled")
```
vadimberezniker, line 109: I think in this case the invocations won't be marked as disconnected because you are not using `disconnectWithErr` but you can't use `disconnectWithErr` as it would have the same issue that you are trying to fix.

I think you could use disconnectWithErr if you used the select around the stream recv:
```
var channelDone <-chan struct{}
select {
   case <-channelDone:
      return disconnectWithErr(...)
  case in:=<-streamMsg:
      // handle message
     // set channelDone when channel is created
}

this would allow you to cancel the stream as soon as the shutdown starts instead of only when new messages are received on the stream

**Review by vadimberezniker — CHANGES_REQUESTED** [review]

File: server/build_event_protocol/build_event_handler/build_event_handler.go [hunk 4]
```diff
@@ -105,26 +106,33 @@ func NewBuildEventHandler(env environment.Env) *BuildEventHandler {
...
-		openChannels:  openChannels,
-	}
+	return h
 }
 
 func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfaces.BuildEventChannel {
 	buildEventAccumulator := accumulator.NewBEValues(iid)
 	ctx = log.EnrichContext(ctx, log.InvocationIDKey, iid)
+	ctx, cancelFn := context.WithCancel(ctx)
+	b.cancelFnsByInvID[iid] = cancelFn
```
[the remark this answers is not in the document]
luluz66: Done.

server/build_event_protocol/build_event_handler/build_event_handler.go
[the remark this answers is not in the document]

server/build_event_protocol/build_event_handler/build_event_handler.go
luluz66: Done.

server/build_event_protocol/build_event_server/build_event_server.go
luluz66: Done.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -105,25 +106,38 @@ func NewBuildEventHandler(env environment.Env) *BuildEventHandler {
...
-		statsRecorder: statsRecorder,
-		openChannels:  openChannels,
-	}
+	return h
 }
 
 func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfaces.BuildEventChannel {
 	buildEventAccumulator := accumulator.NewBEValues(iid)
+	ctx = log.EnrichContext(ctx, log.InvocationIDKey, iid)
+	val, ok := b.cancelFnsByInvID.Load(iid)
```
bduffany, line 128: can the `sync.Map` be from `CancelFunc -> struct{}` ? It seems a little error prone to depend on the invocation ID being unique in this map, since we have invocation retries. Or if funcs can't be used as sync.Map keys then maybe use a UUID separate from the invocation ID.

luluz66: If sync.Map has the invocation id in the map, we will first call the stored cancel Function and then store the new cancel Function.

server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -89,38 +89,61 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
...
 		}
-		if err != nil {
-			log.CtxWarningf(ctx, "Error receiving build event stream %+v: %s", streamID, err)
+	}()
+
+	var channelDone <-chan struct{}
+	for {
+		select {
+		case <-channelDone:
+			return disconnectWithErr(status.UnavailableError("context cancelled due to server shut down"))
```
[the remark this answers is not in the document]
luluz66: Thank you for catching a bug. channelDone is when the channelDone is channel.Context().Done()

server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -89,38 +89,58 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
...
 
-		if err := channel.HandleEvent(in); err != nil {
-			if status.IsAlreadyExistsError(err) {
-				log.CtxWarningf(ctx, "AlreadyExistsError handling event; this means the invocation already exists and may not be retried: %s", err)
-				return err
+	var channelDone <-chan struct{}
+	for {
+		select {
+		case <-channelDone:
+			return disconnectWithErr(status.FromContextError(ctx))
```
bduffany, line 111:
```suggestion
			return disconnectWithErr(status.FromContextError(channel.Context()))
```

(I think?)

bduffany (approved):

server/build_event_protocol/build_event_server/build_event_server.go
luluz66: Done

tylerwilliams (approved):

PR #2806, opened 2022-10-26
Restrict visibility of (most of) //enterprise to //enterprise:__subpackages

iain-macdonald:
This PR removes the unnecessary @buildbuddy_internal//enterprise:__subpackages__ visibility and restricts the visibility of most packages in //enterprise to //enterprise:__subpackages. There are a few that are referenced outside of //enterprise, so I just made exceptions for those and we can clean 'em up later if necessary, but at least they're easier to find now.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/auth/BUILD
+++ b/enterprise/server/auth/BUILD
@@ -6,7 +6,6 @@ go_library(
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/auth",
     visibility = [
         "//enterprise:__subpackages__",
-        "@buildbuddy_internal//enterprise:__subpackages__",
     ],
     deps = [
         "//enterprise/server/selfauth",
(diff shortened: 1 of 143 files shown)


enterprise/server/auth/BUILD
```diff
@@ -6,7 +6,6 @@ go_library(
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/auth",
     visibility = [
         "//enterprise:__subpackages__",
```
tylerwilliams, line 8: There are a few changes in here where we remove `"@buildbuddy_internal//enterprise:__subpackages__",` but still leave the visibility = ["//enterprise:__subpackages__",] attribute set in a target. Can those also be moved to a `package(default_visibility = ["//enterprise:__subpackages__"])` rule instead?

tylerwilliams (changes requested):

enterprise/server/auth/BUILD
iain-macdonald: Yup, definitely, just missed those in my find-and-replace. Should be updated now.

tylerwilliams (approved):
LGTM

Thanks for cleaning this up :)

PR #2789, opened 2022-10-25
Fix logging in migration cache

maggie-lou:
* We shouldn't log destination not found errors if the src cache also had a not found error. This was making it seem like a lot more data was missing from the dest cache, when it was actually expected
* Also change a warning log -> debug log bc it was creating too much noise

PR #2811, opened 2022-10-27
Fix metrics bug and add metric for copy chan size

maggie-lou:
* The same bug I fixed [here](https://github.com/buildbuddy-io/buildbuddy/pull/2789) for logging was also making the metrics look worse than is true

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -512,9 +512,6 @@ func (mc *MigrationCache) Reader(ctx context.Context, r *resource.ResourceName, 
 	if doubleRead {
 		eg.Go(func() error {
 			destReader, dstErr = mc.dest.Reader(ctx, r, offset, limit)
-			if status.IsNotFoundError(dstErr) {
-				metrics.MigrationNotFoundErrorCount.With(prometheus.Labels{metrics.CacheRequestType: "reader"}).Inc()
-			}
 			if dstErr == nil {
 				metrics.MigrationNotFoundErrorCount.With(prometheus.Labels{metrics.CacheRequestType: "readerMatch"}).Inc()
 			}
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #2811** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2811" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2811?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -834,20 +836,26 @@ func (mc *MigrationCache) copyDataInBackground() error {
...
 }
 
-func (mc *MigrationCache) logCopyChanFullInBackground() {
-	ticker := time.NewTicker(mc.copyChanFullWarningInterval)
-	defer ticker.Stop()
+func (mc *MigrationCache) logCopyChanInfoInBackground() {
+	copyChanFullTicker := time.NewTicker(mc.copyChanFullWarningInterval)
+	defer copyChanFullTicker.Stop()
+
+	metricTicker := time.NewTicker(5 * time.Minute)
```
vadimberezniker, line 843: There's no harm in making this much finer, say every 5 seconds. 5 minutes might be enough time to burst and drain a lot of copies from the queue which would not really give you a good insight into the behavior of the queue.

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -834,20 +836,26 @@ func (mc *MigrationCache) copyDataInBackground() error {
 	}
 }
 
-func (mc *MigrationCache) logCopyChanFullInBackground() {
-	ticker := time.NewTicker(mc.copyChanFullWarningInterval)
-	defer ticker.Stop()
+func (mc *MigrationCache) logCopyChanInfoInBackground() {
```
vadimberezniker, line 839: what do you think about calling this `monitorCopyChanFullness`?

vadimberezniker (approved):

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Like that - fixed

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Ack

PR #2815, opened 2022-10-27
CLI: Wait for async sidecar uploads after bazel invocations on CI

bduffany:
* Add a `/Wait` RPC which waits for build_event_proxy and cache_proxy to flush any in-flight requests.
* On CI, call this RPC before the CLI exits.
* Improve debugging situation for the sidecar: write logs to a file (using the same hashing scheme as the socket file), and allow passing args to the sidecar like `BB_SIDECAR_ARGS='--app.log_level=debug'`. The file is truncated on each sidecar run, to avoid having to worry too much about the logs growing indefinitely.

---

**Version bump**: Patch


siggisim:
I'll let @tylerwilliams review this since it's a lot of golang - but any reason why you went with the /Wait rpc approach rather than just having the proxied build event stream / cache calls not return immediately?

bduffany:
> I'll let @tylerwilliams review this since it's a lot of golang - but any reason why you went with the /Wait rpc approach rather than just having the proxied build event stream / cache calls not return immediately?

Chatted with Tyler - I will close this for now and instead:
* Disable the BES proxy / cache proxy on CI until we implement the BES features like client-side redaction
* Send a separate PR with just the sidecar debugging fixes, since those are generally useful to have.

PR #2816, opened 2022-10-28
Fix build visibility in enterprise/server/cmd docs so //website builds.

iain-macdonald:
**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/executor/yaml_doc/BUILD
+++ b/enterprise/server/cmd/executor/yaml_doc/BUILD
@@ -34,4 +34,8 @@ genrule(
         cat $(location :generate_yaml) >> "$@" &&
         echo "\\`\\`\\`" >> "$@"
     """,
+    visibility = [
+        "//enterprise:__subpackages__",
+        "//website:__pkg__",
+    ],
 )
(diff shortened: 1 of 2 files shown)


enterprise/server/cmd/executor/yaml_doc/BUILD
```diff
@@ -34,4 +34,8 @@ genrule(
         cat $(location :generate_yaml) >> "$@" &&
         echo "\\`\\`\\`" >> "$@"
     """,
+    visibility = [
+        "//enterprise:__subpackages__",
```
maggie-lou, line 38: Do you need this line with line 3?

enterprise/server/cmd/server/yaml_doc/BUILD
```diff
@@ -34,4 +34,8 @@ genrule(
         cat $(location :generate_yaml) >> "$@" &&
         echo "\\`\\`\\`" >> "$@"
     """,
+    visibility = [
+        "//enterprise:__subpackages__",
```
maggie-lou, line 38: Here too

enterprise/server/cmd/executor/yaml_doc/BUILD
iain-macdonald: My reading of "Rule target visibility" here: https://bazel.build/concepts/visibility is that the visibility attribute overrides the default_visibility of the containing package. But in this case it's fine because nothing in //enterprise depends on this rule, so I'll go ahead and remove it.

maggie-lou (approved):

PR #2825, opened 2022-10-31
Show Linux RBE minutes on usage page

bduffany:
---

**Version bump**: Patch


enterprise/app/usage/usage.tsx
```diff
@@ -93,3 +95,7 @@ export default class UsageComponent extends React.Component<UsageProps, State> {
     );
   }
 }
+
+function formatMinutes(usec: Number): string {
```
siggisim, line 99: I think we charge by the second (but you can check in with @gtli7) - so it might be worth showing that

siggisim (approved):

enterprise/app/usage/usage.tsx
bduffany: Confirmed with George that we charge by minutes

PR #2826, opened 2022-10-31
Fix fetching remote assets greater than gRPC size limit

bduffany:
---

**Version bump**: Patch


server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -110,19 +110,37 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
...
-				SizeBytes: int64(-1),
+				Hash: fmt.Sprintf("%x", sha256),
+				// We don't know the digest size (yet), but we have to specify
+				// something here or the bytestream API will return an
+				// InvalidArgument error. The bytestream server doesn't actually
+				// check this value, it just needs something greater than 0.
+				//
+				// Note, the value we specify here is used as the max read
+				// buffer size, so want something relatively large here.
+				SizeBytes: readBufSize,
```
maggie-lou, line 123: I know this is the pre-existing behavior but won't a 12MiB max buffer size still be too limiting if we're trying to fetch large files?

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -57,8 +57,8 @@ func NewFetchServer(env environment.Env) (*FetchServer, error) {
 		return nil, status.FailedPreconditionError("A cache is required to enable the ByteStreamServer")
 	}
 	return &FetchServer{
-		env:   env,
-		cache: cache,
+		env:      env,
+		bsClient: env.GetByteStreamClient(),
```
maggie-lou, line 61: nit: You could just pull this off the env - feels a bit more in line with dependency injection pattern

maggie-lou (approved):

server/remote_asset/fetch_server/fetch_server.go
bduffany: Done

server/remote_asset/fetch_server/fetch_server.go
bduffany: This is just the buffer size for each chunk in the stream -- anything on the order of a few MB is probably an OK value, since the server can send multiple chunks at once without waiting for the client to ack each chunk.

(side note, I reduced this to 4 MB to be somewhat consistent with `cachetools.grpcMaxRecvSizeBytes` although it doesn't strictly depend on that value, I think)

PR #2837, opened 2022-11-01
Add a Shutdown function for build event handler

luluz66:
Create a child context with Cancel function when we open a channel; and
put the cancel function in a map in build event handler. When the server
receives shutdown signal, we will call cancel functions.

In BuildEventServer, when the context from the channel is done, we will
return UnavailableError to signal bazel to retry.


luluz66:
Note: This PR is exactly the same as https://github.com/buildbuddy-io/buildbuddy/pull/2804. Vadim had a change requested that blocks merging and he was on vacation.

bduffany (approved):

PR #2840, opened 2022-11-02
Merge changes from bazelbuild/remote-apis remote_execution.proto into proto/remote_execution.proto

iain-macdonald:
The diff is a bit of a mess because we haven't merged in a little while. You can see the diff against the bazelbuild version by running this command from a buildbuddy repo with this branch checked out:

`git clone https://github.com/bazelbuild/remote-apis /tmp/remote-apis && vimdiff proto/remote_execution.proto /tmp/remote-apis/build/bazel/remote/execution/v2/remote_execution.proto`

**Version bump**: Patch

**Related issues**: Unblocks https://github.com/buildbuddy-io/buildbuddy-internal/issues/1766


bduffany:
This might take some time to review since it looks like there are a few changes that are (maybe?) backwards-incompatible.

If you want to send a separate PR with just the changes for `output_paths` (and maybe the deprecation notes on the other fields), that would be an easy-LGTM and unblock the output_paths change :)

iain-macdonald:
Yeah, I noticed that. Sent you https://github.com/buildbuddy-io/buildbuddy/pull/2841 and I'll see what I can do with the rest of the proto separately...

PR #2842, opened 2022-11-02
Make workflow memory / CPU configurable

bduffany:
Example:

```
actions:
  - name: Test
    triggers: ...
    resource_requests:
      memory: 10GB   # = 10*(1024^3) bytes
      cpu: 4         # = 4 cores
    bazel_commands: ...
```

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1765

### DIFF
--- a/enterprise/server/remote_execution/platform/platform.go
+++ b/enterprise/server/remote_execution/platform/platform.go
@@ -15,6 +15,7 @@ import (
 	"google.golang.org/grpc/metadata"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	units "github.com/docker/go-units"
 )
 
 var (
@@ -98,8 +99,8 @@ const (
 	// requires a relatively large amount of free space compared to typical actions.
 	EstimatedFreeDiskPropertyName = "EstimatedFreeDiskBytes"
 
-	EstimatedMilliCPUPropertyName    = "EstimatedMilliCPU"
-	EstimatedMemoryBytesPropertyName = "EstimatedMemoryBytes"
+	EstimatedCPUPropertyName    = "EstimatedCPU"
+	EstimatedMemoryPropertyName = "EstimatedMemory"
 
 	BareContainerType        ContainerType = "none"
 	PodmanContainerType      ContainerType = "podman"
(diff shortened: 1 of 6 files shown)


siggisim (approved):
Could we use github.com/docker/go-units (which we already import) to make these a little more human friendly?

enterprise/server/workflow/service/service.go
```diff
@@ -720,6 +720,8 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 				{Name: platform.IncludeSecretsPropertyName, Value: includeSecretsPropertyValue},
 				{Name: platform.EstimatedComputeUnitsPropertyName, Value: fmt.Sprintf("%d", computeUnits)},
 				{Name: platform.EstimatedFreeDiskPropertyName, Value: "20000000000"}, // 20GB
+				{Name: platform.EstimatedMemoryBytesPropertyName, Value: fmt.Sprint(workflowAction.ResourceRequests.MemoryBytes)},
+				{Name: platform.EstimatedMilliCPUPropertyName, Value: fmt.Sprint(workflowAction.ResourceRequests.MilliCPU)},
```
tylerwilliams, line 724: nit: can you use strconv.Itoa instead of "fmt.Sprint" here?

bduffany: updated to accept human-readable units per Siggi's comment

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -445,6 +453,49 @@ func int64Prop(props map[string]string, name string, defaultValue int64) int64 {
...
+	f, err := strconv.ParseFloat(val, 64)
+	if err == nil {
+		return int64(f)
+	}
+	// Otherwise attempt to parse as an IEC number of bytes.
+	// Example: 4GB = 4*(1024^3) bytes
+	n, err := units.RAMInBytes(val)
+	if err != nil {
+		return defaultValue
+	}
```
tylerwilliams, line 472: I'm cool with this as-is, but we should have a way to indicate to users that their buildbuddy.yaml file is misconfigured, rather than always failing silently and using a default value (which could be really frustrating for a user trying to figure out why something is not having an effect)

tylerwilliams (approved):

enterprise/server/remote_execution/platform/platform.go
bduffany: I'll add a TODO here for that -- shouldn't be too hard but will require a bit of refactoring.

PR #2848, opened 2022-11-03
Increase max length of secret values

bduffany:
The current type of the secret `value` column is VARCHAR(255) which is too short. This PR changes the type to `text` which has a longer max length.

---

**Version bump**: Patch


siggisim (approved):

tylerwilliams (approved):

tylerwilliams:
Github limits these to 64K too so I think this brings us inline. (And for larger stuff they have a documented workaround)

PR #2851, opened 2022-11-03
Remote asset: Avoid buffering fetched contents in memory

bduffany:
* When downloading from CAS to compute the digest size, write to a counter instead of a buffer
* When fetching, use an intermediate temp file that will live on disk

---

**Version bump**: Patch


server/util/scratchspace/scratchspace.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"flag"
+	"os"
+	"path/filepath"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+)
+
+var (
+	tempDir_doNotUseDirectly = flag.String("storage.tempdir", filepath.Join(os.TempDir(), "buildbuddy-scratch"), "Staging directory for ephemeral files. If unspecified, defaults to 'TEMPDIR/buildbuddy' where TEMPDIR is the OS-specific temp dir.")
```
tylerwilliams: I don't think this needs to be named this way -- it's a package level private var you could just throw a comment on it?

tylerwilliams: I think it would be more straightforward to have your flag default to os.TempDir() and then your "TempDir()" func below can *always* append "buildbuddy-scratch" (which could could make a const?). Then it becomes impossible for someone to accidentally blow away the contents of some dir if they accidentally set --storage.tempdir to "/home/tylerw" or something.

server/util/scratchspace/scratchspace.go
```diff
@@ -0,0 +1,54 @@
...
+		return status.InvalidArgumentErrorf("failed to clear scratch directory: %s", err)
+	}
+	if err := os.MkdirAll(TempDir(), 0755); err != nil {
+		return status.InvalidArgumentErrorf("failed to create scratch directory: %s", err)
+	}
+	return nil
+}
+
+// TempDir returns the directory where scratch files are written.
+func TempDir() string {
```
tylerwilliams, line 29: nit: consider making this private if it's not used outside of this pcakage

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -196,22 +198,37 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
...
 	}
 	if expectedSHA256 != "" && blobDigest.Hash != expectedSHA256 {
 		return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, blobDigest.Hash, expectedSHA256)
 	}
 	cacheRN := digest.NewCASResourceName(blobDigest, remoteInstanceName)
-	if _, err := cachetools.UploadFromReader(ctx, bsClient, cacheRN, bytes.NewReader(data)); err != nil {
+	if _, err := f.Seek(0, 0); err != nil {
+		return nil, status.UnavailableErrorf("failed to seek temp file: %s", err)
+	}
+	if _, err := cachetools.UploadFromReader(ctx, bsClient, cacheRN, f); err != nil {
```
tylerwilliams, line 231: What you have is fine -- but if you wanted to close the file, upload it, then remove it, you could use cachetools.UploadFile instead.

tylerwilliams (changes requested):

server/util/scratchspace/scratchspace.go
bduffany: done

server/remote_asset/fetch_server/fetch_server.go
bduffany: hm, I'm not a fan that UploadFile() tries to add things to the filecache, even though the app isn't configured with one today :/

server/util/scratchspace/scratchspace.go
```diff
@@ -0,0 +1,60 @@
...
+	// NOTE: This flag shouldn't be used directly except in the tempDir() func.
+	// We intentionally write to a subdirectory under this dir to avoid blowing
+	// away existing files in the configured directory, in case of accidental
+	// misconfiguration.
+	tempDir_doNotUseDirectly = flag.String("storage.tempdir", os.TempDir(), "Root directory for temporary files. Defaults to the OS-specific temp dir.")
+)
+
+const (
+	// tempSubDir is a subdirectory under the temp dir where we actually write
+	// temporary files.
```
tylerwilliams, line 21: nit: if you only have one const you can just write

const tempSubDir = "buildbuddy-scratch"

Re comment can you also mention that we recursively rm everything in this dir when Init() is called?

// tempSubDir is a subdirectory under the temp dir where we actually write
// temporary files. It's contents are recursively removed when Init() is called.

server/remote_asset/fetch_server/fetch_server.go
bduffany: nvm, was looking at the wrong UploadFile(). Done -- looks much cleaner :)

server/util/scratchspace/scratchspace.go
bduffany: Added the comment but kept the const as a section so that future consts added will result in a cleaner diff

tylerwilliams (approved):

PR #2853, opened 2022-11-03
Add AC check to filestorer

maggie-lou:

PR #2859, opened 2022-11-04
Remove duplicate partition ID from file path in pebble

maggie-lou:
During the isolate by group ID migration, we should also remove the duplicate partition ID in the pebble cache file name

For partition default:
- previous name would be: PTdefault/default/groupid/ac/hash
- now it will be: PTdefault/groupid/ac/hash

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -180,6 +180,9 @@ type PebbleCache struct {
 	orphanedFilesDone chan struct{}
 
 	fileStorer filestore.Store
+
+	// TODO(Maggie): Clean this up after the isolateByGroupIDs migration
+	isolateByGroupIDs bool
 }
 
 // Register creates a new PebbleCache from the configured flags and sets it in
@@ -370,6 +373,7 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 		accesses:               make(chan *accessTimeUpdate, *opts.AtimeBufferSize),
 		evictors:               make([]*partitionEvictor, len(opts.Partitions)),
 		fileStorer:             filestore.New(filestore.Opts{IsolateByGroupIDs: opts.IsolateByGroupIDs}),
+		isolateByGroupIDs:      opts.IsolateByGroupIDs,
 	}
 
 	peMu := sync.Mutex{}
@@ -769,7 +773,11 @@ func (p *PebbleCache) makeFileRecord(ctx context.Context, r *resource.ResourceNa
 
 // partitionBlobDir returns a directory path under the root directory, for the specified partition, where blobs can be stored.
 func (p *PebbleCache) partitionBlobDir(partID string) string {
-	partDir := partitionDirectoryPrefix + partID
+	// In isolateByGroupID migration, remove duplicate partition ID from file name
+	partDir := ""
+	if !p.isolateByGroupIDs {
+		partDir = partitionDirectoryPrefix + partID
+	}
 	return filepath.Join(p.rootDirectory, "blobs", partDir)
 }
 
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


maggie-lou:
* master
  * **PR #2853** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2853" target="_blank">Graphite</a> 
    * **PR #2859** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2859" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2859?utm_source=stack-comment).

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -536,6 +540,10 @@ func (p *PebbleCache) deleteOrphanedFiles(quitChan chan struct{}) error {
 			default:
 			}
 
+			dupPartitionPattern := fmt.Sprintf("%s/%s", part.ID, part.ID)
```
vadimberezniker, line 543: Could we not use `p.isolateByGroupIDs` here? 
Should be fine to assume that a cache root should have only one type of directory structure.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -617,7 +626,8 @@ func (p *PebbleCache) scanForBrokenFiles(quitChan chan struct{}) error {
 			log.Errorf("Error unmarshaling metadata when scanning for broken files: %s", err)
 			continue
 		}
-		blobDir = p.partitionBlobDir(fileMetadata.GetFileRecord().GetIsolation().GetPartitionId())
+		pathHasDupPartitionID := !bytes.HasPrefix(fileMetadataKey, []byte("PT"))
+		blobDir = p.blobDir(pathHasDupPartitionID, fileMetadata.GetFileRecord().GetIsolation().GetPartitionId())
```
vadimberezniker, line 630: Same here, could we not use `p.isolateByGroupIDs`?

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: When manually testing I accidentally reused the same directory and set isolateByGroupIDs=true and scanForBrokenFiles deleted everything in the cache... So I added this as a precaution. This code should be temporary anyway and will be deleted after the migration is complete

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -536,6 +540,10 @@ func (p *PebbleCache) deleteOrphanedFiles(quitChan chan struct{}) error {
 			default:
 			}
 
+			dupPartitionPattern := fmt.Sprintf("%s/%s", part.ID, part.ID)
+			pathHasDupPartitionID, err := regexp.MatchString(dupPartitionPattern, path)
```
vadimberezniker, line 544: Do you need regexp here, would `strings.Contains` work just as well?

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Good call - don't know why I used regex

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Actually realized that we won't be able to clean up this code, because some on-prem customers may still be on the old naming even after the cloud migration. That being said, I still think the precaution is valuable in case someone manually deploys an old config on dev or something to do testing without realizing the repercussions of reusing the directory

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: AFAIK, nobody is using pebble on-prem yet.

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: AFAIK, nobody is using pebble on-prem yet.

maggie-lou: Oh great - then we'll be able to clean this up

PR #2860, opened 2022-11-04
Remote asset: Stream directly to cache if we know the digest ahead of time

bduffany:
If we get a `Content-Length` header and the client has sent up an expected SHA256, then we know the full digest and can pipe the HTTP response body directly to cache instead of downloading it to disk first.

---

**Version bump**: Patch

### DIFF
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -198,7 +198,24 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 	if rsp.StatusCode < 200 || rsp.StatusCode >= 400 {
 		return nil, status.UnavailableErrorf("failed to fetch %q: HTTP %s", uri, err)
 	}
-	// Download to a temp file to avoid running out of memory.
+
+	// If we know what the SHA256 should be and the content length is known,
+	// then we know the full digest, and can pipe directly from the HTTP
+	// response to cache.
+	if expectedSHA256 != "" && rsp.ContentLength >= 0 {
+		d := &repb.Digest{Hash: expectedSHA256, SizeBytes: rsp.ContentLength}
+		rn := digest.NewResourceName(d, remoteInstanceName)
+		if _, err := cachetools.UploadFromReader(ctx, bsClient, rn, rsp.Body); err != nil {
+			return nil, status.UnavailableErrorf("failed to upload %s to cache: %s", digest.String(d), err)
+		}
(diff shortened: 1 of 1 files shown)


server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -198,7 +198,24 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 	if rsp.StatusCode < 200 || rsp.StatusCode >= 400 {
 		return nil, status.UnavailableErrorf("failed to fetch %q: HTTP %s", uri, err)
 	}
-	// Download to a temp file to avoid running out of memory.
+
+	// If we know what the SHA256 should be and the content length is known,
+	// then we know the full digest, and can pipe directly from the HTTP
+	// response to cache.
+	if expectedSHA256 != "" && rsp.ContentLength >= 0 {
```
maggie-lou, line 205: We are no longer validating the digest before we write the data to the cache. Should that be a requirement?

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -198,7 +198,24 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
...
+		d := &repb.Digest{Hash: expectedSHA256, SizeBytes: rsp.ContentLength}
+		rn := digest.NewResourceName(d, remoteInstanceName)
```
maggie-lou, line 207: Should be NewCASResourceName

bduffany: Hm why is the CacheType needed in this case? The ByteStream API (used here) doesn't have a notion of CacheType -- in ResourceName.UploadString() we don't use the CacheType.

If we want to have CacheType defined everywhere for consistency reasons, should we get rid of NewCASResourceName() and have NewResourceName() default to CacheType.CAS (since AC is more of a special case?)

server/remote_asset/fetch_server/fetch_server.go
bduffany: It's not needed in this case since the ByteStreamServer will validate the digest before committing to the cache.

maggie-lou: Ack

server/remote_asset/fetch_server/fetch_server.go
maggie-lou: Ah you're right - my bad. Doesn't like the cachetype is needed here

PR #2861, opened 2022-11-04
Remote asset: Determine digest size using Metadata API

bduffany:
Speeds up fetches (and reduces load on the cache) by checking Metadata and renewing the cache entry with FindMissingBlobs instead of downloading the full asset using the ByteStream API.

---

**Version bump**: Patch

### DIFF
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -15,7 +15,6 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
-	"github.com/buildbuddy-io/buildbuddy/server/util/ioutil"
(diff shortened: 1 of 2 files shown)


server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -111,31 +121,43 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
...
+			blobDigest.SizeBytes = md.SizeBytes
+			// Even though we successfully fetched metadata, we need to renew
+			// the cache entry using FindMissingBlobs to ensure that it doesn't
+			// expire by the time the client requests it from cache.
+			cas := p.env.GetContentAddressableStorageClient()
+			req := &repb.FindMissingBlobsRequest{
+				InstanceName: req.GetInstanceName(),
+				BlobDigests:  []*repb.Digest{blobDigest},
+			}
+			res, err := cas.FindMissingBlobs(ctx, req)
```
maggie-lou, line 151: Basically the same but cache.Contains can also update LRU entries - unclear if either is clearer 🤷‍♀️

maggie-lou (approved):

PR #2863, opened 2022-11-04
CLI: Improve bazelrc parsing logic

bduffany:
* Expand --config args in place instead of appending them and then later trying to remove the args which were expanded from those configs
* Fix various bugs w/ bazelrc parsing logic, precedence etc.
* TODO in a future PR: add some build_metadata (maybe?) that tells BB what the original command line was before we expanded configs in place, so we can show a nicer "explicit command line" in the UI

---

**Version bump**: Patch

### DIFF
--- a/cli/arg/arg.go
+++ b/cli/arg/arg.go
@@ -9,23 +9,37 @@ import (
 
 // Returns true if the list of args contains desiredArg
 func Has(args []string, desiredArg string) bool {
-	arg, _ := find(args, desiredArg)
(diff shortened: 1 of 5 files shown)


siggisim (approved):

bduffany:
@siggisim FYI I wound up not trying to reuse Bazel's C++ parsing code because it turns out most of the heavy lifting is done in the Java code. There is some useful stuff in the C++ code but IMO not enough to justify trying to call it directly

PR #2867, opened 2022-11-07
local split

tylerwilliams:
This change moves the split triggering logic out of the driver and into the store + replica. Replicas periodically check their own usage and when it is over a certain size, they request a split from the store. This change simplifies things and also reduces the time it takes for a range to split when over size. It also reduces the load on the gossip network since the store no longer needs to "poll" so aggressively to catch size changes, instead they are driven by replicas increasing in size.

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -1226,7 +1226,6 @@ func (sm *Replica) Update(entries []dbsm.Entry) ([]dbsm.Entry, error) {
 			sm.log.Warningf("Error computing usage: %s", err)
 		} else {
 			if usage.GetEstimatedDiskBytesUsed() > *replicaSplitSizeBytes {
-				sm.log.Warningf("Requesting split")
 				sm.store.RequestSplit(sm.ClusterID)
 			}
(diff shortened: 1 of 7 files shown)


tylerwilliams:
Discussed offline, but I agree on symmetry re merges and if we implement them I will route some of this functionality through the driver.

enterprise/server/raft/replica/replica.go
```diff
@@ -1212,6 +1220,18 @@ func (sm *Replica) Update(entries []dbsm.Entry) ([]dbsm.Entry, error) {
 		sm.lastAppliedIndex = entry.Index
 	}
 
+	if sm.lastAppliedIndex-sm.lastUsageCheckIndex > entriesBetweenUsageChecks {
+		usage, err := sm.Usage()
+		if err != nil {
+			sm.log.Warningf("Error computing usage: %s", err)
+		} else {
+			if usage.GetEstimatedDiskBytesUsed() > *replicaSplitSizeBytes {
+				sm.log.Warningf("Requesting split")
```
vadimberezniker, line 1229: change to Info or remove

vadimberezniker (approved):

PR #2872, opened 2022-11-08
Write Executions to clickhouse

luluz66:
1. Create ExecutionCollector which collects executions in redis when execution is
complete. When an execution is complete, we look up the InvocationExecutions table, and append the execution to the lists in redis keyed by invocation ids.  
2. When an invocation is finalized, we write to clickhouse all the executions.

---

**Version bump**: TODO


server/util/clickhouse/clickhouse.go
```diff
@@ -10,6 +10,7 @@ import (
 	"syscall"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
+	iepb "github.com/buildbuddy-io/buildbuddy/proto/internal_execution"
```
bduffany, line 13: nit: move named imports to a separate section below

server/util/clickhouse/clickhouse.go
```diff
@@ -302,23 +303,63 @@ func ToInvocationFromPrimaryDB(ti *tables.Invocation) *Invocation {
 	}
 }
 
+func ToExecutionFromInternalProto(in *iepb.Execution, inv *tables.Invocation) *Execution {
```
bduffany, line 306: nit: can this be private?

also consider `buildExecution` since it builds it from both the internal proto and the DB invocation -- or split into separate funcs like `fillExecutionFromInternalProto`, `fillExecutionFromInvocation`.

server/util/clickhouse/clickhouse.go
```diff
@@ -302,23 +303,63 @@ func ToInvocationFromPrimaryDB(ti *tables.Invocation) *Invocation {
...
 	retrier := retry.DefaultWithContext(ctx)
 	var lastError error
 	for retrier.Next() {
-		res := h.DB(ctx).Create(inv)
+		res := h.DB(ctx).Create(value)
 		lastError = res.Error
 		if errors.Is(res.Error, syscall.ECONNRESET) || errors.Is(res.Error, syscall.ECONNREFUSED) || isTimeout(res.Error) {
 			// Retry since it's an transient error.
-			log.CtxInfof(ctx, "attempt (n=%d) to write invocation to clickhouse failed: %s", retrier.AttemptNumber(), res.Error)
+			log.CtxInfof(ctx, "attempt (n=%d) to clickhouse table %q failed: %s", retrier.AttemptNumber(), tableName, res.Error)
```
bduffany, line 362: nit:

```suggestion
			log.CtxWarningf(ctx, "Attempt (n=%d) to write to clickhouse table %q failed: %s", retrier.AttemptNumber(), tableName, res.Error)
```

bduffany (changes requested):
It would be great to have an automated test that mocks out the clickhouse DB and asserts that the executions were added. `bazel_rbe_test.go` has an example of using `runRemoteShellActionViaBazel` to run exactly one RBE action under a single invocation.

luluz66 (commented):
> It would be great to have an automated test that mocks out the OLAP DB interface and asserts that the executions were added. `bazel_rbe_test.go` has an example of using `runRemoteShellActionViaBazel` to run exactly one RBE action under a single invocation, using the RBE test env (we could mock out the OLAP DB in `rbetest.go`). That `runRemoteShellActionViaBazel` func would just need to be extended with the `bes_backend` flag so that it exercises the build event handler codepaths.

I am going to look into adding an integration test in a followup PR since this PR is already quite big.

enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
```diff
@@ -0,0 +1,73 @@
...
+
+func New(rdb redis.UniversalClient, rbuf *redisutil.CommandBuffer) *collector {
+	return &collector{
+		rdb:  rdb,
+		rbuf: rbuf,
+	}
+}
+
+func getKey(iid string) string {
+	return strings.Join([]string{redisExecutionKeyPrefix, iid}, ":")
```
vadimberezniker, line 44: We have generally been using forward slashes as delimiters in Redis keys.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -287,11 +287,50 @@ func (r *statsRecorder) flushInvocationStatsToOLAPDB(ctx context.Context, ij *in
...
 	// Temporary logging for debugging clickhouse missing data.
-	if err == nil {
-		log.CtxInfo(ctx, "successfully wrote invocation to clickhouse")
+	log.CtxInfo(ctx, "Successfully wrote invocation to clickhouse")
+
+	if r.env.GetExecutionCollector() == nil {
+		return nil
 	}
-	return err
+	const batchSize = 500_000
```
vadimberezniker, line 299: Do we need such a large batch size? Would for example 50k work just as well and reduce potential for large memory spikes?

proto/internal_execution.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
+
+package internal_execution;
```
vadimberezniker, line 3: This is fine, but you could also consider putting the new message in remote_execution.proto along with other BB specific-stuff. You could rename it slightly to make its meaning clear.

vadimberezniker (approved):

luluz66:
> 


> > It would be great to have an automated test that mocks out the OLAP DB interface and asserts that the executions were added. `bazel_rbe_test.go` has an example of using `runRemoteShellActionViaBazel` to run exactly one RBE action under a single invocation, using the RBE test env (we could mock out the OLAP DB in `rbetest.go`). That `runRemoteShellActionViaBazel` func would just need to be extended with the `bes_backend` flag so that it exercises the build event handler codepaths.
> 
> I am going to look into adding an integration test in a followup PR since this PR is already quite big.


See https://github.com/buildbuddy-io/buildbuddy/pull/2991

bduffany (approved):

PR #2875, opened 2022-11-08
Supercede `go_rules_dependencies` with our deps

tempoz:
`go_rules_dependencies` declares many repositories, which was causing our repository definitions with the same names to be silently ignored. To prevent this, we declare our deps first, and then apply gazelle patches as needed to the relevant libraries. Libraries that do not currently need to be patched now contain a commented-out block that would apply the patch as documentation in case the patch (or a modification of the patch) needs to be applied in the future.

---

**Version bump**: Patch

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -26,6 +26,13 @@ http_archive(
     ],
 )
 
+load(":deps.bzl", "install_buildbuddy_dependencies")
+
+# Install gazelle and go_rules dependencies after ours so that our go module versions take precedence.
+
+# gazelle:repository_macro deps.bzl%install_buildbuddy_dependencies
+install_buildbuddy_dependencies()
+
 load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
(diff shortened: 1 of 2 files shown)


tempoz:
* master
  * **PR #2875** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2875" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2875?utm_source=stack-comment).

bduffany (approved):
LGTM - for future reference, why is `com_github_gogo_protobuf` getting a new patch applied in this PR while some other deps are getting their existing patch removed?

tempoz:
Not all patches could be cleanly applied, and all but one was unnecessary: https://github.com/bazelbuild/bazel-gazelle/issues/561#issuecomment-506047843

Not sure why one of the gazelle patches was necessary, but it was 🤷‍♀️

PR #2876, opened 2022-11-08
Add option to log to GCP

tempoz:
Mac RBE currently logs locally, and no one ever looks at the logs, rendering them largely useless. This change allows us to configure the server logger to also forward log messages to GCP.

---

**Version bump**: Minor


tempoz:
This still needs to be tested on our macs and the process for configuring logging needs to be documented.

server/util/log/gcp/gcp.go
```diff
@@ -0,0 +1,47 @@
...
+	case zerolog.InfoLevel:
+		l.logger.StandardLogger(logging.Info).Print(string(p))
+	case zerolog.WarnLevel:
+		l.logger.StandardLogger(logging.Warning).Print(string(p))
+	case zerolog.ErrorLevel:
+		l.logger.StandardLogger(logging.Error).Print(string(p))
+	case zerolog.PanicLevel:
+		fallthrough
+	case zerolog.FatalLevel:
+		l.logger.StandardLogger(logging.Critical).Print(string(p))
```
tylerwilliams, line 41: can we use https://pkg.go.dev/cloud.google.com/go/logging#hdr-Synchronous_Logging here?

server/util/log/gcp/gcp.go
```diff
@@ -0,0 +1,47 @@
...
+	ctx    context.Context
+	logger *logging.Logger
+}
+
+func (l *logWriter) Write(p []byte) (int, error) {
+	l.logger.StandardLogger(logging.Default).Print(string(p))
+	return len(p), nil
+}
+
+func (l *logWriter) WriteLevel(level zerolog.Level, p []byte) (int, error) {
```
tylerwilliams, line 28: I haven't seen what these logs look like in the gcp viewer yet, but wondering if we manually need to add the same structured-logging stuff that zerolog is adding and/or the keys from the context (like in enrichEventFromContext)

tylerwilliams (commented):
This is pretty cool, excited to see it land :)

tylerwilliams (approved):
I think this LG -- do you have a link to some logs written from a mac that you could link in the CL desc?

Also added Vadim because he's touched some log stuff recently.

tempoz:
> I think this LG -- do you have a link to some logs written from a mac that you could link in the CL desc?

I'm still finalizing it, I'm not happy with how it looks right now and it's erasing levels from the normal logger for some reason.

server/util/log/gcp/gcp.go
tempoz: Fixed this: 
image

server/util/log/gcp/gcp.go
tempoz: Done.

server/util/log/gcp/gcp.go
bduffany: :heart_eyes:

tempoz:
This is now complete and awaits review.

server/util/log/gcp/gcp.go
```diff
@@ -0,0 +1,122 @@
...
+	"cloud.google.com/go/logging"
+	"github.com/rs/zerolog"
+
+	logpb "google.golang.org/genproto/googleapis/logging/v2"
+	structpb "google.golang.org/protobuf/types/known/structpb"
+)
+
+var (
+	ProjectID = flag.String("app.log_gcp_project_id", "", "The project ID to log to in GCP (if any).")
+	LogID     = flag.String("app.log_gcp_log_id", "", "The log ID to log to in GCP (if any).")
```
bduffany, line 19: since these apply to the executor as well as the app, wdyt about putting this under a separate top-level `log:` namespace? so

```
log:
  gcp:
    project_id: ...
    log_id: ...
```

I also think it'd be nice (at some point) to alias `log.level` --> `app.log_level`

bduffany (approved):

server/util/log/gcp/gcp.go
bduffany: ah I see we have a bunch of other `app.log_*` flags too, like `app.log_include_short_file_name`. Feel free to ignore :)

tempoz:
image
I think the logs look pretty good now.

server/util/log/gcp/gcp.go
tempoz: I think adding aliases for these is a good idea, but outside the scope of this PR.

server/util/log/gcp/gcp.go
```diff
@@ -0,0 +1,122 @@
...
+				}
+				jsonPayload.Fields["timestamp"] = structpb.NewStringValue(t)
+			}
+		}
+		if v, ok := m[zerolog.MessageFieldName]; ok {
+			if p, ok := v.(string); ok {
+				jsonPayload.Fields["message"] = structpb.NewStringValue(p)
+			}
+		}
+		if v, ok := m["executor_id"]; ok {
```
vadimberezniker, line 63: Let's avoid hardcoding these fields. Our logging code makes it easy to attach new fields.

You can look at the zerolog builtin logger for a reference on how they handle log fields: https://github.com/rs/zerolog/blob/3543e9d94bc5ed088dd2d9ad1d19c7ccd0fa65f5/console.go#L102

vadimberezniker (changes requested):

tempoz:
image
This is what the logs look like now.

server/util/log/gcp/gcp.go
```diff
@@ -0,0 +1,108 @@
...
+	entry.Payload = payload
+
+	// Populate the Entry fields, as this won't happen automatically. The map of
+	// json to Entry fields can be found here:
+	// https://cloud.google.com/logging/docs/structured-logging
+	fields := payload.GetFields()
+	if t, err := time.Parse(zerolog.TimeFieldFormat, fields[zerolog.TimestampFieldName].GetStringValue()); err == nil {
+		entry.Timestamp = t
+	}
+	if v, ok := fields["logging.googleapis.com/sourceLocation"]; ok {
```
vadimberezniker, line 84: Can you add a constant for this? We also refer to it in `log.go`.

server/util/log/log.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 
+	gcplogging "github.com/buildbuddy-io/buildbuddy/server/util/log/gcp"
```
vadimberezniker, line 21: Is this alias necessary? Could you refer to it as simply `gcp` or rename the package to `gcplog` if you think gcp is not descriptive?

server/util/log/gcp/gcp.go
vadimberezniker, line 84: Could you exclude fields from the Payload that are already present at the top level (such as this one, timestamp, severity, etc)?

vadimberezniker (approved):

PR #2887, opened 2022-11-09
Fix grafana dashboard.

vadimberezniker:
Unclear why but Raft & Golang panels are broken right now.

I re-saved the dashboard and it fixed itself... :shrug:

---

**Version bump**: TODO


tylerwilliams (approved):
wooo (thank you!)

PR #2910, opened 2022-11-11
CLI: Override explicit command line in the UI

bduffany:
We expand `.bazelrc` default arguments and `--config` flags, causing the explicit command line to be super long. This PR lets us show the original `bb` tool command line under "explicit command line" instead of the expanded Bazel command line.

---

**Version bump**: Patch

### DIFF
--- a/cli/metadata/metadata.go
+++ b/cli/metadata/metadata.go
@@ -1,0 +1,21 @@
+package metadata
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+func AppendBuildMetadata(args, originalArgs []string) ([]string, error) {
+	// Ensure arg0 is "bb" (by default, it is an abs path)
+	originalArgs = append([]string{"bb"}, originalArgs[1:]...)
+
+	originalArgsJSON, err := json.Marshal(originalArgs)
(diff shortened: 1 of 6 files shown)


cli/metadata/metadata.go
```diff
@@ -0,0 +1,21 @@
+package metadata
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+func AppendBuildMetadata(args, originalArgs []string) ([]string, error) {
+	// Ensure arg0 is "bb" (by default, it is an abs path)
+	originalArgs = append([]string{"bb"}, originalArgs[1:]...)
```
siggisim, line 10: Could hard-coding "bb" make it harder for folks that are using the .bazelversion approach?

siggisim (approved):

cli/metadata/metadata.go
bduffany: That's a good point, I'll try to see if there's a way to show "bazel" in that case.

bduffany: Done (showing "bazel" now if invoked via .bazelversion).

PR #2913, opened 2022-11-11
CLI: Include sidecar as a regular library instead of an embedded binary

bduffany:
To call the sidecar, we now run `bb [sidecar_args...]  --sidecar=1`. The `bb` subprocess then attempts to pop that `--sidecar=1` arg, and if present, it calls the sidecar main function instead of the usual CLI main function.

Because we're now building the sidecar directly into bb, this reduces the binary size from 47 MB to 30 MB (on Linux).

---

**Version bump**: Patch


siggisim (approved):
This is.... awesome!

PR #2916, opened 2022-11-14
Return update enabled capability based on API key permissions

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1801


siggisim (approved):

bduffany:
Hm, the test breakage seems legitimate, but it's unexpected. Seems like if the capability is unset, bazel has a fatal error: https://app.buildbuddy.io/invocation/04b4fe56-27ed-4f6b-8fb7-7c230c77aae4?target=%2F%2Fenterprise%2Fserver%2Ftest%2Fintegration%2Fremote_cache%3Aremote_cache_test

Looking into it.

bduffany:
I think the issue is that prior to bazel 6.0.0, setting `--remote_upload_local_results=true` would be considered an error if the cache write capabilities indicate that the user can't write to the AC, but 6.0.0 makes it a warning.

To make this work, I think we'd need to put this behind a version check for bazel >= 6.0.0.

bduffany:
Marking as draft for now and will revive once we can test against the official 6.0.0 release.

bduffany:
Tested with bazel 5.3.0, 6.0.0rc4, and 6.0.0 -- all behave as expected.

bduffany:
@tylerwilliams I updated this PR with client version checking -- PTAL

tylerwilliams (approved):

PR #2930, opened 2022-11-14
Add utilities for bazel client version parsing and comparison

bduffany:
Context: https://github.com/buildbuddy-io/buildbuddy/pull/2916#issuecomment-1314209533

This will also be useful to address https://github.com/buildbuddy-io/buildbuddy-internal/issues/1187

---

**Version bump**: Patch

### DIFF
--- a/server/util/bazel_request/bazel_request.go
+++ b/server/util/bazel_request/bazel_request.go
@@ -2,12 +2,20 @@ package bazel_request
 
 import (
 	"context"
+	"regexp"
+	"strconv"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/proto"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+)
+
+var (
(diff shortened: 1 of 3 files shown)


server/util/bazel_request/bazel_request.go
```diff
@@ -33,6 +41,89 @@ func GetInvocationID(ctx context.Context) string {
...
+			return nil
+		}
+	}
+	b.Suffix = m[bazelVersionPattern.SubexpIndex("suffix")]
+	return b
+}
+
+// IsAtLeast returns whether this bazel version is equal to or supercedes the
+// given version.
+func (v *Version) IsAtLeast(c *Version) bool {
```
tylerwilliams, line 92: I wonder if we should add a variant of this function that accepts a string version and returns (bool, err) instead, for ease of use / readability when we actually use this?

tylerwilliams (approved):
nice :)

server/util/bazel_request/bazel_request.go
bduffany: ^ going to submit as-is, and we can decide later whether we need a util method

tylerwilliams: SGTM

PR #2931, opened 2022-11-15
Support Command.output_paths (attempt #2)

iain-macdonald:
This is a rollforward (with fixes) of https://github.com/buildbuddy-io/buildbuddy/pull/2879 which was reverted in https://github.com/buildbuddy-io/buildbuddy/pull/2903 because some rules_kotlin rules expected the previous behavior of output_directories.

This PR does a few things:

Adds a more complex test for the various kinds of outputs (output_directories, output_files, and output_paths)
Supports output_paths instead of output_files and output_directories, when specified. Otherwise, falls back to output_files and output_directories. N.B. that the behavior of output_paths is slightly different than the behavior of output_files and output_directories: the executor will only create the parents of all of the specified output_paths, not the paths themselves.
Does not change the behavior when output_files and output_directories are used -- all specified output_directories will still be created and uploaded by the executor, and parents of output_files will be created and the files uploaded, just like before.
Version bump: Minor

Related issues: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1766

### DIFF
--- a/enterprise/server/test/integration/remote_execution/remote_execution_test.go
+++ b/enterprise/server/test/integration/remote_execution/remote_execution_test.go
@@ -787,15 +787,12 @@ func TestBasicActionIO(t *testing.T) {
 		Arguments: []string{
 			"sh", "-c", strings.Join([]string{
 				`set -e`,
-				// Create a file in the output directory.
-				// No need to create the output directory itself; executor is
-				// responsible for that.
+                // No need to create the output directory itself; executor is
+                // responsible for that because it's specified as an
+                // OutputDirectory.
 				`cp greeting.input out_dir/hello_world.output`,
 				`printf 'world' >> out_dir/hello_world.output`,
 				// Create a file in a child dir of the output directory.
-				// Need to create the child directory ourselves since it's not a declared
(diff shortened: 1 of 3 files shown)


enterprise/server/test/integration/remote_execution/remote_execution_test.go
```diff
@@ -878,19 +876,19 @@ func TestComplexActionIO(t *testing.T) {
...
+				opath="out_files_dir/$(echo "$path" | sed 's/.input/.output/')"
+				cat "$path" | tail -c +2 > "$opath"
 			done
 			`,
-			// Mirror the input tree to out_dir, skipping the first 2 bytes this time.
-			// We *do* need to create parent dirs since the executor is only
-			// responsible for creating the top-level out_dir.
+			// Mirror the input tree to out_dir, skipping the first 2 bytes this
+			// time. We *do* need to create parent dirs since the executor is
+			// only responsible for creating the top-level out_dir.
```
tylerwilliams, line 891: nit: can you make these / other comments 80 chars?

tylerwilliams (approved):

PR #2838, opened 2022-11-01
Support writing zstd-compressed data to pebble cache

maggie-lou:
If a client uploads zstd compressed bytes with the --experimental_remote_cache_compression flag, this PR enables pebble cache to save those compressed bytes directly to the cache. Previously in this situation, we would decompress the bytes before storing, and re-compress the bytes when read

It also adds support for cached vs. requested compression type being different. For example:
1. Client uploads compressed bytes for digest d
2. Compressed bytes are written to the cache
3. In a future build digest d is read but requested uncompressed
4. The cache contains compressed data for d. Even though the uncompressed version is requested, this is still considered a cache hit
5. The cache will uncompress the compressed data and return it

If a client uploads non-compressed bytes, this PR does not compress them before storing. This will come in a future PR

PR #2932, opened 2022-11-15
Compress all data above a certain size before writing to pebble cache

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -68,6 +68,7 @@ var (
 	forceCalculateMetadata    = flag.Bool("cache.pebble.force_calculate_metadata", false, "If set, partition size and counts will be calculated even if cached information is available.")
 	isolateByGroupIDsFlag     = flag.Bool("cache.pebble.isolate_by_group_ids", false, "If set, filepaths and filekeys for AC records will include groupIDs")
 	enableCompressionFlag     = flag.Bool("cache.pebble.enable_zstd_compression", false, "If set, zstd compressed files can be saved to the cache. Otherwise only decompressed bytes should be stored.")
+	compressWritesFlag        = flag.Bool("cache.pebble.compress_writes", false, "If set, data should be compressed before written to disk")
 
 	// Default values for Options
 	// (It is valid for these options to be 0, so we use ptrs to indicate whether they're set.
@@ -133,6 +134,7 @@ type Options struct {
 	PartitionMappings     []disk.PartitionMapping
 	IsolateByGroupIDs     bool
 	EnableZstdCompression bool
+	CompressWrites        bool
 
 	MaxSizeBytes           int64
 	BlockCacheSizeBytes    int64
@@ -191,6 +193,7 @@ type PebbleCache struct {
 	bufferPool *bytebufferpool.Pool
 
 	enableZstdCompression bool
+	compressWrites        bool
 
 	// TODO(Maggie): Clean this up after the isolateByGroupIDs migration
 	isolateByGroupIDs bool
@@ -230,6 +233,7 @@ func Register(env environment.Env) error {
 		PartitionMappings:      *partitionMappingsFlag,
 		IsolateByGroupIDs:      *isolateByGroupIDsFlag,
 		EnableZstdCompression:  *enableCompressionFlag,
+		CompressWrites:         *compressWritesFlag,
 		BlockCacheSizeBytes:    *blockCacheSizeBytesFlag,
 		MaxSizeBytes:           cache_config.MaxSizeBytes(),
 		MaxInlineFileSizeBytes: *maxInlineFileSizeBytesFlag,
@@ -386,6 +390,7 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 		evictors:               make([]*partitionEvictor, len(opts.Partitions)),
 		fileStorer:             filestore.New(filestore.Opts{IsolateByGroupIDs: opts.IsolateByGroupIDs}),
 		enableZstdCompression:  opts.EnableZstdCompression,
+		compressWrites:         opts.CompressWrites,
 		isolateByGroupIDs:      opts.IsolateByGroupIDs,
 		bufferPool:             bytebufferpool.New(CompressorBufSizeBytes),
 	}
@@ -1189,7 +1194,16 @@ type zstdCompressor struct {
 func (z *zstdCompressor) Write(p []byte) (int, error) {
 	z.compressBuf = compression.CompressZstd(z.compressBuf, p)
 	n, err := z.w.Write(z.compressBuf)
-	return n, err
+	z.bufferPool.Put(z.compressBuf)
+
+	metrics.CompressionRatio.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Set(float64(len(p)) / float64(n))
+	if n > len(p) {
+		metrics.BadCompressionRatioBufferSize.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Set(float64(len(p)))
+	}
+
+	// Return the full length of the buffer even though a different compressed buffer size may have been written,
+	// or clients will return a short write error
+	return len(p), err
 }
 
 func (z *zstdCompressor) Commit() error {
@@ -1201,9 +1215,7 @@ func (z *zstdCompressor) Metadata() *rfpb.StorageMetadata {
 }
 
 func (z *zstdCompressor) Close() error {
-	err := z.w.Close()
-	z.bufferPool.Put(z.compressBuf)
-	return err
+	return z.w.Close()
 }
 
 func (p *PebbleCache) Writer(ctx context.Context, r *resource.ResourceName) (interfaces.CommittedWriteCloser, error) {
@@ -1215,7 +1227,9 @@ func (p *PebbleCache) Writer(ctx context.Context, r *resource.ResourceName) (int
 
 	// If data is not already compressed, return a writer that will compress it before writing
 	// Only compress data over a given size for more optimal compression ratios
-	shouldCompress := r.GetCompressor() == repb.Compressor_IDENTITY && r.GetDigest().GetSizeBytes() >= p.maxInlineFileSizeBytes
+	shouldCompress := p.compressWrites && p.enableZstdCompression &&
+		r.GetCompressor() == repb.Compressor_IDENTITY &&
+		r.GetDigest().GetSizeBytes() >= p.maxInlineFileSizeBytes
 	if shouldCompress {
 		r = &resource.ResourceName{
 			Digest:       r.GetDigest(),
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1197,21 +1197,21 @@ type zstdCompressor struct {
 	bufferPool  *bytebufferpool.Pool
 }
 
-func (z *zstdCompressor) Write(p []byte) (int, error) {
-	z.compressBuf = compression.CompressZstd(z.compressBuf, p)
-	n, err := z.CustomCommitWriteCloser.Write(z.compressBuf)
+func (z *zstdCompressor) Write(decompressedBytes []byte) (int, error) {
+	z.compressBuf = compression.CompressZstd(z.compressBuf, decompressedBytes)
+	compressedBytesWritten, err := z.CustomCommitWriteCloser.Write(z.compressBuf)
 	if err != nil {
 		return 0, err
 	}
 
-	metrics.CompressionRatio.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Observe(float64(len(p)) / float64(n))
-	if n > len(p) {
-		metrics.BadCompressionBufferSize.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Observe(float64(len(p)))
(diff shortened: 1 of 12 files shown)


maggie-lou:
* master
    * **PR #2932** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2932" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2932?utm_source=stack-comment).

server/metrics/metrics.go
```diff
@@ -1439,6 +1439,24 @@ var (
...
+		Help:      "The size of buffers that increased in size when decompressed",
+	}, []string{
+		CompressionType,
+	})
+
+	CompressionRatio = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: bbNamespace,
+		Subsystem: "compressor",
+		Name:      "compression_ratio",
+		Help:      "The data compression ratio (decompressed bytes / compressed bytes)",
```
bduffany, line 1455: Since this is only measuring the data compression ratio for pebble and is measuring the ratio for each chunk, maybe "pebble" makes more sense as the subsystem? and also maybe clarify in the Help text that it's per written chunk

enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -981,21 +1071,84 @@ func TestCompression_ParallelRequests(t *testing.T) {
...
+				Compressor: repb.Compressor_IDENTITY,
 			}
 			reader, err := pc.Reader(ctx, compressedRN, 0, 0)
 			require.NoError(t, err)
 			defer reader.Close()
 			data, err := io.ReadAll(reader)
 			require.NoError(t, err)
-			decompressed, err := compression.DecompressZstd(nil, data)
-			require.Equal(t, blob, decompressed)
+			//decompressed, err := compression.DecompressZstd(nil, data)
```
bduffany, line 1081: nit: remove?

bduffany (commented):
Mostly LGTM - will let Vadim or Tyler approve since they're more familiar with the pebble internals

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1179,14 +1190,59 @@ func (dc *writeCloser) Write(p []byte) (int, error) {
...
+	*ioutil.CustomCommitWriteCloser
+	compressBuf []byte
+	bufferPool  *bytebufferpool.Pool
+}
+
+func (z *zstdCompressor) Write(p []byte) (int, error) {
+	z.compressBuf = compression.CompressZstd(z.compressBuf, p)
+	n, err := z.CustomCommitWriteCloser.Write(z.compressBuf)
+	if err != nil {
+		z.bufferPool.Put(z.compressBuf)
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1179,14 +1190,59 @@ func (dc *writeCloser) Write(p []byte) (int, error) {
...
+
+	// Return the size of the original buffer even though a different compressed buffer size may have been written,
+	// or clients will return a short write error
+	return len(p), nil
+}
+
+func (z *zstdCompressor) Close() error {
+	err := z.CustomCommitWriteCloser.Close()
+	z.bufferPool.Put(z.compressBuf)
+	return err
```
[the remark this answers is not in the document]
maggie-lou: Done

server/metrics/metrics.go
```diff
@@ -1439,6 +1439,24 @@ var (
...
+	}, []string{
+		CompressionType,
+	})
```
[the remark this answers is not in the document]
maggie-lou: Done

I kept the bad compression metric because that tracks the size of buffers with bad compression, not just the count

server/metrics/metrics.go
maggie-lou: Done

enterprise/server/backends/pebble_cache/pebble_cache_test.go
maggie-lou: Done

proto/raft.proto
```diff
@@ -65,7 +65,8 @@ message FileMetadata {
   FileRecord file_record = 1;
   StorageMetadata storage_metadata = 2;
 
-  // Size of the stored data (if data is compressed, this will be the compressed size).
+  // Size of the stored data (if data is compressed, this will be the compressed
+  // size).
   int64 size_bytes = 3;
```
tylerwilliams, line 70: maybe we should rename this to stored_size_bytes?

maggie-lou: Done

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1179,14 +1190,56 @@ func (dc *writeCloser) Write(p []byte) (int, error) {
...
+	z.compressBuf = compression.CompressZstd(z.compressBuf, decompressedBytes)
+	compressedBytesWritten, err := z.CustomCommitWriteCloser.Write(z.compressBuf)
+	if err != nil {
+		return 0, err
+	}
+
+	metrics.CompressionRatio.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Observe(float64(compressedBytesWritten) / float64(len(decompressedBytes)))
+	if compressedBytesWritten > len(decompressedBytes) {
+		metrics.BadCompressionBufferSize.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Observe(float64(len(decompressedBytes)))
+	}
```
tylerwilliams, line 1210: does it make sense to move this to the Close() method so this only happens once at the end of the write?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1179,14 +1190,56 @@ func (dc *writeCloser) Write(p []byte) (int, error) {
 	return n, nil
 }
 
-// TODO(Maggie): Compress all data above a given size before writing it
+// zstdCompressor compresses bytes before writing them to the nested writer
+type zstdCompressor struct {
```
tylerwilliams, line 1194: wdyt about having a New method for this struct that accepts a bufferPool and ioutil.CustomCommitWriteCloser and calls bufferPool.Get()? That way the bufferpool lifecycle is fully internal to this struct instead of being half the callers responsibility and half this code's responsibility?

tylerwilliams (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1179,14 +1190,56 @@ func (dc *writeCloser) Write(p []byte) (int, error) {
...
+func (z *zstdCompressor) Write(decompressedBytes []byte) (int, error) {
+	z.compressBuf = compression.CompressZstd(z.compressBuf, decompressedBytes)
+	compressedBytesWritten, err := z.CustomCommitWriteCloser.Write(z.compressBuf)
+	if err != nil {
+		return 0, err
+	}
+
+	metrics.CompressionRatio.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Observe(float64(compressedBytesWritten) / float64(len(decompressedBytes)))
+	if compressedBytesWritten > len(decompressedBytes) {
+		metrics.BadCompressionBufferSize.With(prometheus.Labels{metrics.CompressionType: "zstd"}).Observe(float64(len(decompressedBytes)))
```
bduffany, line 1209: A large "bad" buffer that is only bad by 1 byte, and a medium-sized buffer that is bad by 5000% can potentially both show up in the same bucket.

Would it be more useful to record `(compressedSize - decompressedSize)` here?

bduffany: (Or are you trying to measure whether it's true that it's mostly the smaller buffers that have bad compression? If that's the case, fine to leave as-is)

server/metrics/metrics.go
```diff
@@ -1532,3 +1571,16 @@ func exponentialBucketRange(min, max, factor float64) []float64 {
...
+// number of buckets.
+func bucketsWithInterval(min, max, interval float64) []float64 {
+	buckets := []float64{}
+	current := min
+	for current < max {
+		buckets = append(buckets, current)
+		current += interval
+	}
+	return buckets
+}
```
bduffany, line 1586: this seems equivalent to `prometheus.LinearBuckets` - https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#LinearBuckets

maggie-lou: Oh nice - replaced

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Yeah was trying to validate that smaller buffers have bad compression, so we can tune at what point we want to compress the data. I think the other compression ratio metric will help us track "badness", so we don't need to record (compressedSize - decompressedSize)

Also made the size at which we start compressing a flag, so it's more configurable if we notice that it's too small

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Done

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Changed. Now the metrics will be on the stream level instead of the chunk level, which I think will be more useful as a metric

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -54,20 +54,22 @@ var (
...
+	forceCompaction                 = flag.Bool("cache.pebble.force_compaction", false, "If set, compact the DB when it's created")
+	forceCalculateMetadata          = flag.Bool("cache.pebble.force_calculate_metadata", false, "If set, partition size and counts will be calculated even if cached information is available.")
+	isolateByGroupIDsFlag           = flag.Bool("cache.pebble.isolate_by_group_ids", false, "If set, filepaths and filekeys for AC records will include groupIDs")
+	enableCompressionFlag           = flag.Bool("cache.pebble.enable_zstd_compression", false, "If set, zstd compressed files can be saved to the cache. Otherwise only decompressed bytes should be stored.")
+	compressWritesFlag              = flag.Bool("cache.pebble.compress_writes", false, "If set, data should be compressed before written to disk")
+	minCompressionFileSizeBytesFlag = flag.Int64("cache.pebble.min_compression_file_size_bytes", DefaultMaxInlineFileSizeBytes, "If compression is on, files larger than this will be compressed before written to disk.")
```
tylerwilliams, line 72: So by setting this to DefaultMaxInlineFileSizeBytes, we're saying we'll compress everything not inlined, i.e. not written to pebble. Which seems OK because by default pebble is already using snappy compression: https://github.com/cockroachdb/pebble/blob/4a63cdb3a71eb3a58d1a8a9849b6314e7799c6a1/options.go#L402

I think we should add a flag to use zstd for pebble stuff too before migrating. It's about as fast as snappy and has a slightly better compression ratio

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -54,20 +54,22 @@ var (
...
+	atimeBufferSizeFlag             = flag.Int("cache.pebble.atime_buffer_size", DefaultAtimeBufferSize, "Buffer up to this many atime updates in a channel before dropping atime updates")
+	minEvictionAgeFlag              = flag.Duration("cache.pebble.min_eviction_age", DefaultMinEvictionAge, "Don't evict anything unless it's been idle for at least this long")
+	forceCompaction                 = flag.Bool("cache.pebble.force_compaction", false, "If set, compact the DB when it's created")
+	forceCalculateMetadata          = flag.Bool("cache.pebble.force_calculate_metadata", false, "If set, partition size and counts will be calculated even if cached information is available.")
+	isolateByGroupIDsFlag           = flag.Bool("cache.pebble.isolate_by_group_ids", false, "If set, filepaths and filekeys for AC records will include groupIDs")
+	enableCompressionFlag           = flag.Bool("cache.pebble.enable_zstd_compression", false, "If set, zstd compressed files can be saved to the cache. Otherwise only decompressed bytes should be stored.")
```
tylerwilliams, line 70: nit: can we name this var to match the flag name?

Also, might be worth considering how another var controlling the pebble cache compression type will be named (so it's clear which is which)

tylerwilliams (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Done. Renamed the other flag to try and make it more clear. Is that what you meant?

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: Okay I'll add the inlining zstd in a separate PR so this one isn't blocked while I test that

tylerwilliams (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
maggie-lou: @tylerwilliams looks like zstd is not as performant in pebble. Probably not worth it for minor compression ratio improvements: https://github.com/cockroachdb/pebble/issues/1582

"Writer throughput with zstd is still ~70% slower than with snappy."

PR #2936, opened 2022-11-15
CLI: Add a `print` command to view serialized bazel log files

bduffany:
Adds a command `bb print --grpc_log=PATH` which prints a sequence of JSON-serialized LogEntry protos.

This is effectively the same as `remote_client --grpc_log` (from https://github.com/bazelbuild/tools_remote) except the output is easier to parse with something like `jq`, and also the BuildBuddy extensions to REAPI V2 are rendered with their correct names, instead of tag numbers (such as ExecutedActionMetadata.usage_stats).

Will add website docs in a separate PR.

---

**Version bump**: Patch

### DIFF
--- a/cli/bazelisk/bazelisk.go
+++ b/cli/bazelisk/bazelisk.go
@@ -199,10 +199,12 @@ func setBazelVersion() error {
 	if err != nil && !os.IsNotExist(err) {
 		return err
 	}
-	// Bazelisk probably chose us because we were specified first in
-	// .bazelversion. Delete the first line, if it exists.
-	if len(parts) > 0 {
-		parts = parts[1:]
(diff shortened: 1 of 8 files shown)


siggisim (commented):
I like the concept, but the print-grpc-log naming and having to pass the path feels a little clunky.

I'm wondering if there's a way we could make it feel more ergonomic, either by simplifying the command name - maybe just print with a grpc log argument or adding an arg to an invocation that has a grpc log flag that will print it?

bduffany:
@siggisim  wdyt about `bb print --grpc_log <path>` ? then we could extend that with other log types, e.g. `--execution_log`

siggisim:
I like that

siggisim (approved):

PR #2946, opened 2022-11-16
Add AWS IAM support for DB connections.

vadimberezniker:
To support this functionality, users will need to use the new 'advanced' configuration stanzas. We need this to be able to generate the DSN using the short-lived tokens returned by AWS.

A typical use will look like this:

```
 advanced_data_source:
    driver: mysql
    use_aws_iam: true
    region: some-region
    endpoint: some-instance:3306
    username: some-user
    db_name: buildbuddy
```

Most users can continue to use the simpler 'data_source' option.

The other notable change is that we now control the creation of new DB connections so that we can generate a new DSN every time since the generated AWS IAM tokens are short-lived.

---

**Version bump**: Minor

### DIFF
--- a/enterprise/server/cmd/server/BUILD
+++ b/enterprise/server/cmd/server/BUILD
@@ -87,6 +87,9 @@ go_binary(
 container_image(
     name = "base_image",
     base = "@buildbuddy_go_image_base//image",
+    files = [
+        "@aws_rds_certs//file",
+    ],
     symlinks = {
         "config.yaml": "app/enterprise/server/cmd/server/buildbuddy.runfiles/buildbuddy/enterprise/config/buildbuddy.release.yaml",
         "buildbuddy": "tmp",
--- a/server/util/db/db.go
+++ b/server/util/db/db.go
@@ -2,7 +2,10 @@ package db
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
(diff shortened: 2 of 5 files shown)


server/util/db/db.go
```diff
@@ -2,16 +2,19 @@ package db
...
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"database/sql"
+	"database/sql/driver"
 	"errors"
 	"flag"
 	"fmt"
+	golog "log"
```
tylerwilliams, line 12: can you move this down below

enterprise/server/cmd/server/BUILD
```diff
@@ -87,6 +87,9 @@ go_binary(
 container_image(
     name = "base_image",
     base = "@buildbuddy_go_image_base//image",
+    files = [
+        "@aws_rds_certs//file",
+    ],
```
tylerwilliams, line 92: Can we move this to the //static target so it's also included in the binary bundle and will work for non-docker folks?

tylerwilliams (changes requested):
This looks really good overall, just a couple of small nits

server/util/db/db.go
vadimberezniker: Done.

tylerwilliams (approved):

PR #2949, opened 2022-11-17
Use `Timer` instead of `time.After` so we can drain channels and avoid a memory leak

tempoz:
Addresses buildbuddy-io/buildbuddy-internal#1806

---

**Version bump**: Patch

### DIFF
--- a/server/backends/chunkstore/chunkstore.go
+++ b/server/backends/chunkstore/chunkstore.go
@@ -399,8 +399,13 @@ type ChunkstoreWriter struct {
 }
 
 func (w *ChunkstoreWriter) readFromWriteResultChannel() (int, error) {
+	delay := time.NewTimer(w.writeTimeoutDuration)
 	select {
 	case result, open := <-w.writeResultChannel:
+		if !delay.Stop() {
+			// don't leak the timer
+			<-delay.C
+		}
 		if !open || result.Close {
 			close(w.writeChannel)
 			w.closed = true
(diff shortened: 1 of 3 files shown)


tempoz:
* master
  * **PR #2949** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2949" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2949?utm_source=stack-comment).

server/backends/chunkstore/chunkstore.go
```diff
@@ -399,8 +399,13 @@ type ChunkstoreWriter struct {
 }
 
 func (w *ChunkstoreWriter) readFromWriteResultChannel() (int, error) {
+	delay := time.NewTimer(w.writeTimeoutDuration)
 	select {
 	case result, open := <-w.writeResultChannel:
+		if !delay.Stop() {
+			// don't leak the timer
+			<-delay.C
+		}
```
bduffany, line 408: nit: it would be slightly more robust IMO to have this in a `defer` just after declaring `delay` - same in the diff below

tempoz: sgtm, I'll do that

tempoz: mmm, doesn't work because we already drained the channel in the `<-delay.C` case, so it deadlocks in the `defer`.

tempoz: solved it; we can read from the channel only if a value is ready in the defer.

tempoz: Fixed.

bduffany (approved):

PR #2950, opened 2022-11-17
Embed AWS RDS CA certs instead of packaging them in the docker image.

vadimberezniker:
---

**Version bump**: None

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -115,6 +115,12 @@ genrule(
     cmd_bash = "cp $(SRCS) $@",
 )
 
+# Certs that are distributed with the server binary.
+filegroup(
+    name = "embedded_certs",
+    srcs = [":rds-combined-ca-bundle.pem"],
+)
+
 # N.B. this is ignored by gazelle so must be updated by hand.
 # It must live at the repo root to be able to bundle other files using
 # "go:embed".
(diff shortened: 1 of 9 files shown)


BUILD
```diff
@@ -107,23 +107,30 @@ config_setting(
...
     srcs = ["bundle.go"],
-    embedsrcs = select({
-        ":fastbuild": [
-            "//:config_files",
-            "//static",
-        ],
+    embedsrcs = [
+        "//:config_files",
+        "//static",
+        ":rds-combined-ca-bundle.pem",
```
tylerwilliams, line 127: can we put this in //static? I'd like to avoid having every embedded file need to be separately listed here (if possible)

vadimberezniker: It feels a bit odd putting this under static. 
Static files are those that are meant to be served by the UI so it feels wrong to lump in the certs under there.

What's the concern about adding individual files into the bundle?

tylerwilliams: Would you be open to putting this in config_files if you feel that //static is a bad path? Or adding a filegroup like "//certs" instead?

Because we already do some switching on buildmode to decide what to include, my concern is I don't want people to have to think about build_modes x number_of_files. It's easier if you can just be like... oh yeah we need certs, configs, and static files.

vadimberezniker: I moved the cert into //:embedded_certs file group instead of depending on the file directly.

tylerwilliams (approved):

PR #2954, opened 2022-11-18
CLI: add a few basic integration tests

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/server/testutil/testbazel/testbazel.go
+++ b/server/testutil/testbazel/testbazel.go
@@ -5,20 +5,21 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/testutil/testfs"
 	"github.com/buildbuddy-io/buildbuddy/server/util/bazel"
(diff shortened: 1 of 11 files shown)


server/testutil/testbazel/testbazel.go
```diff
@@ -52,13 +55,6 @@ func shutdown(ctx context.Context, t *testing.T, workspaceDir string) {
 
 func MakeTempWorkspace(t *testing.T, contents map[string]string) string {
 	workspaceDir := testfs.MakeTempDir(t)
-	t.Cleanup(func() {
```
vadimberezniker, line 55: What's the reason to remove this?

vadimberezniker (approved):

server/testutil/testbazel/testbazel.go
bduffany: I found that it slows down the test unnecessarily. It's not really needed locally since bazel defaults to `max_idle_secs=15` when it's run under a bazel test (it checks whether `TEST_TMPDIR` is defined). It's also not needed remotely since we kill the whole container after each test action, which will kill bazel.

PR #2956, opened 2022-11-18
Use resource.ResourceName in distributed cache function

maggie-lou:
We need this change in order to pass a resource's compression type to distributed cache methods

### DIFF
--- a/enterprise/server/backends/distributed/BUILD
+++ b/enterprise/server/backends/distributed/BUILD
@@ -16,7 +16,6 @@ go_library(
         "//proto:resource_go_proto",
         "//server/environment",
         "//server/interfaces",
-        "//server/remote_cache/digest",
         "//server/resources",
         "//server/util/background",
         "//server/util/consistent_hash",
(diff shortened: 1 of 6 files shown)


maggie-lou:
* master
    * **PR #2956** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2956" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2956?utm_source=stack-comment).

vadimberezniker (changes requested):
I don't think this change is backwards compatible / rollout safe.
We need to keep reading the isolation field in RPCs if the new resources field is not set.
This code is also deployed on on-prem installations so we should expect that we cannot remove the backwards compatibility handling until we properly notify customers.

maggie-lou:
> I don't think this change is backwards compatible / rollout safe. We need to keep reading the isolation field in RPCs if the new resources field is not set. This code is also deployed on on-prem installations so we should expect that we cannot remove the backwards compatibility handling until we properly notify customers.

Ah yes duh - my bad. Added some fallback code for that case

vadimberezniker (approved):

PR #2959, opened 2022-11-19
Track # of file records & size in each replica.

vadimberezniker:
---

**Version bump**: None

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -383,7 +383,7 @@ func (sm *Replica) updatePartitionMetadata(wb *pebble.Batch, key, val []byte, fi
 
 	if op == fileRecordDelete {
 		pm.TotalCount--
-		pm.SizeBytes -= fileMetadata.GetSizeBytes()
+		pm.SizeBytes -= fileMetadata.GetStoredSizeBytes()
 	} else {
 		_, closer, err := wb.Get(key)
 		if err == nil {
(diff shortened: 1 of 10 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -232,6 +249,17 @@ func (sm *Replica) rangeCheckedSet(wb *pebble.Batch, key, val []byte) error {
 	if !keys.IsLocalKey(key) {
 		if sm.mappedRange != nil && sm.mappedRange.Contains(key) {
 			sm.rangeMu.RUnlock()
+
+			if isFileRecordKey(key) {
+				fileMetadata := &rfpb.FileMetadata{}
+				if err := proto.Unmarshal(val, fileMetadata); err != nil {
+					return err
+				}
```
tylerwilliams, line 257: unmarshalling every key written to the replica in order to count it feels pretty expensive. I see why this is convenient though since it lets us count replayed as well as directly written files.

Would it be possible to extract this information from the key instead and skip the unmarshalling?

vadimberezniker: Is the unmarshaling cost here worth worrying about it? 
I haven't measured it, but I imagine that cost is relatively low compared to the cost of the enclosing Raft write.

vadimberezniker: Updated to look at the key instead. I remembered that we need to look at the key for estimating DB size anyway so there's no point trying to avoid parsing parts of the key.

tylerwilliams (approved):

PR #2961, opened 2022-11-21
Speed up bazel integration tests

bduffany:
Each integration test that runs `bazel` currently spends ~3s extracting the bazel installation. This makes tests feel sluggish when run interactively, and discourages adding more of these tests since they feel heavyweight.

This PR removes the need for bazel to extract itself by making a pre-extracted bazel installation available in the runfiles tree, which tests can then use by passing `--install_base` as a Bazel startup option. The PR also updates all existing tests to reference this install base path.

Caveat: we have to fudge with the installation a bit, since the FS setup for both sandboxed and remote spawns result in a non-"pristine" installation that causes Bazel to crash in some scenarios. See the comments in `initInstallBase` in `testbazel.go`.

---

**Version bump**: Patch

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,1 +1,4 @@
+bazel-bin
+bazel-buildbuddy
+bazel-out
 node_modules
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,4 +1,1 @@
-bazel-bin
-bazel-buildbuddy
-bazel-out
 node_modules
--- a/enterprise/server/test/integration/ci_runner/ci_runner_test.go
+++ b/enterprise/server/test/integration/ci_runner/ci_runner_test.go
@@ -31,6 +31,7 @@ import (
 var (
 	workspaceContentsWithBazelVersionAction = map[string]string{
 		"WORKSPACE": `workspace(name = "test")`,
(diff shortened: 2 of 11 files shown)


.bazelignore
```diff
@@ -1 +1,4 @@
+bazel-bin
+bazel-buildbuddy
+bazel-out
```
bduffany, line 3: FYI, this fixes an issue that running `bazel test //...` tries to pick up targets under these dirs. I don't think it's related to this PR, since I've had to make these changes in the past, but just never got around to checking them in.

vadimberezniker:
Are these changes worth saving 3 seconds from each test? 
Could the overhead be mitigated by increasing test parallelism?

bduffany:
@vadimberezniker This change is more intended to make local development faster -- when authoring or debugging these tests I'll often run the same test several times in serial, and have wasted a lot of time just staring at the "Extracting bazel installation..." message.

I simplified the PR so that tests don't have to care about `spawn_strategy` or `install_base` anymore, and the changes are constrained just to `testbazel.BinaryPath`. If it causes any issues then the change can be more easily reverted or selectively disabled. wdyt?

vadimberezniker (approved):
Looks good.

PR #2962, opened 2022-11-21
Implement SupportsCompressor for distributed cache

maggie-lou:
---

**Version bump**: TODO


maggie-lou:
* master
  * **PR #2956** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2956" target="_blank">Graphite</a> 
      * **PR #2962** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2962" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2962?utm_source=stack-comment).

enterprise/server/backends/distributed/distributed.go
```diff
@@ -1001,7 +1008,16 @@ func (c *Cache) Writer(ctx context.Context, r *resource.ResourceName) (interface
 	return mwc, nil
 }
 
-// TODO(Maggie): Implement for distributed cache
-func (c *Cache) SupportsCompressor(compressor repb.Compressor_Value) bool {
-	return false
+// Compression should only be enabled for a resource if all peers support compression
+func (c *Cache) SupportsCompressor(ctx context.Context, compressor repb.Compressor_Value, r *resource.ResourceName) bool {
```
vadimberezniker, line 1012: What do you think about adding a flag to the distributed cache to indicate whether it should use the value returned by the underlying cache? That way we can roll this out by first enabling support in the underlying cache and then doing a second rollout to enable it in the distributed cache?

I'm not sure it's worth the cost of adding additional RPCs for every read/write that will be returning a fixed value after the rollout is done.

vadimberezniker (changes requested):

enterprise/server/backends/distributed/distributed.go
maggie-lou: Good point. It's definitely wasteful. Added a flag

To roll it out in dev, we will need to merge [this PR](https://github.com/buildbuddy-io/buildbuddy-internal/pull/1824) first, and then this current PR.

To roll it out in prod, we will need to:
1. Set distributed_cache.enable_local_compression_lookup=false
2. Enable compression in pebble cache
3. After #2 has been rolled out, set distributed_cache.enable_local_compression_lookup=true

vadimberezniker: Do we need the remote RPC at all? Would the following rollout ordering work?

1. Enable compression support in pebble cache
2. distributed cache continues to return false for compression support
3. enable enable_local_compression_lookup flag which will now return the value from the local cache which should be the same for all app instances

maggie-lou: That's true - that would be simpler. Made a new PR because that was easier than to reset a lot of the now unneeded remote changes I'd added to this PR: https://github.com/buildbuddy-io/buildbuddy/pull/3010

PR #2963, opened 2022-11-21
Make repo build with bazel 6

iain-macdonald:
I had to do a few things to make this work:
- Remove the managed_directories directive from our WORKSPACE file and set symlink_node_modules to False in the yarn_install directive.
- Migrate to platform definitions from https://github.com/bazelbuild/platforms.
- Upgrade to version 0.7 of rules_k8s (which switches to use the platform definitions above).
- Clear the go_version passed into k8s_go_deps, otherwise it specifies a version which conflicts with ours. I thought about changing our version to match theirs but that didn't work out of the box and I thought it'd be better to have ours be the only go version definition.

You can test this yourself by running (change version & checksum for different platforms):
`wget -O /tmp/bazel-6.0.0-pre.20221020.1-linux-x86_64 https://github.com/bazelbuild/bazel/releases/download/6.0.0-pre.20221020.1/bazel-6.0.0-pre.20221020.1-linux-x86_64 && echo "b8bf93fc9fd780c2e3e5fe7222de70edc019076a8f53c8f67ce0054d955a5ede  /tmp/bazel-6.0.0-pre.20221020.1-linux-x86_64" | sha256sum --check --status && chmod +x /tmp/bazel-6.0.0-pre.20221020.1-linux-x86_64 && git clone https://github.com/buildbuddy-io/buildbuddy /tmp/bb && cd /tmp/bb && git checkout bazel_6_its_one_bigger_isnt_it && /tmp/bazel-6.0.0-pre.20221020.1-linux-x86_64 build //...`

**Version bump**: Minor
**Related issues**: Partially fixes #1810

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -135,8 +135,8 @@ go_library(
 platform(
     name = "firecracker",
     constraint_values = [
-        "@bazel_tools//platforms:x86_64",
-        "@bazel_tools//platforms:linux",
+        "@platforms//cpu:x86_64",
+        "@platforms//os:linux",
     ],
     exec_properties = {
         "workload-isolation-type": "firecracker",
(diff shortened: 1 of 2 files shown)


bduffany:
Since managed_directories is going away I think it just means that we have to manually `yarn install` in order to create the node_modules folder so that VSCode (etc.) can provide syntax highlighting for NPM deps.

We'll have to live with it for now, but it might be possible to work around.

bduffany (approved):

PR #2964, opened 2022-11-21
Update buildfix.sh to use the gofmt tool included in rules_go

bduffany:
---

**Version bump**: Patch


bduffany:
@tylerwilliams I was able to create a run target like `bazel run :gofmt` using a starlark rule to access that GoSDK object, wdyt?

tylerwilliams (approved):
you fast :) thanks for doing this

PR #2966, opened 2022-11-22
Implement SupportsCompressor for migration cache

maggie-lou:
* Also fix a bug in FindMissing where we were hardcoding "Compressor: repb.Compressor_IDENTITY" when queuing the data to be copied. FindMissing is always called with resources with the identity compressor anyway, so this shouldn't matter. But it's technically incorrect and could cause a subtle bug later

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -147,6 +147,7 @@ func pebbleCacheFromConfig(env environment.Env, cfg *PebbleCacheConfig) (*pebble
 		AtimeWriteBatchSize:    cfg.AtimeWriteBatchSize,
 		AtimeBufferSize:        cfg.AtimeBufferSize,
 		MinEvictionAge:         cfg.MinEvictionAge,
+		EnableZstdCompression:  cfg.EnableZstdCompression,
 	}
 	c, err := pebble_cache.NewPebbleCache(env, opts)
 	if err != nil {
@@ -232,9 +233,15 @@ func (mc *MigrationCache) FindMissing(ctx context.Context, resources []*resource
 	// Some implementations of FindMissing sort the resources slice, which can cause a race condition if both
 	// the src and dest cache try to sort at the same time. Copy the slice to prevent this
 	var resourcesCopy []*resource.ResourceName
+	var hashToResource map[string]*resource.ResourceName
 	if doubleRead {
 		resourcesCopy = make([]*resource.ResourceName, len(resources))
 		copy(resourcesCopy, resources)
+
+		hashToResource = make(map[string]*resource.ResourceName)
+		for _, r := range resources {
+			hashToResource[r.GetDigest().GetHash()] = r
+		}
 	}
 
 	eg.Go(func() error {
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #2956** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2956" target="_blank">Graphite</a> 
      * **PR #2962** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2962" target="_blank">Graphite</a> 
        * **PR #2966** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2966" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/2966?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -232,9 +233,15 @@ func (mc *MigrationCache) FindMissing(ctx context.Context, resources []*resource
...
 	var resourcesCopy []*resource.ResourceName
+	var hashToResource map[string]*resource.ResourceName
 	if doubleRead {
 		resourcesCopy = make([]*resource.ResourceName, len(resources))
 		copy(resourcesCopy, resources)
+
+		hashToResource = make(map[string]*resource.ResourceName)
+		for _, r := range resources {
+			hashToResource[r.GetDigest().GetHash()] = r
+		}
```
bduffany, line 244: This change LGTM but seems unrelated -- just leaving this comment as a sanity check in case you meant to include it in another PR instead.

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -835,6 +835,9 @@ func (mc *MigrationCache) Stop() error {
 	return dstShutdownErr
 }
 
+// Compression should only be enabled during a migration if both the source and destination caches support the
+// compressor, in order to reduce complexity around trying to double read/write compressed bytes to one cache and
+// decompressed bytes to another
 func (mc *MigrationCache) SupportsCompressor(ctx context.Context, compressor repb.Compressor_Value, r *resource.ResourceName) bool {
-	return false
+	return mc.src.SupportsCompressor(ctx, compressor, r) && mc.dest.SupportsCompressor(ctx, compressor, r)
```
bduffany, line 842: This LGTM at a glance, assuming the underlying cache impls returns a correct value of `SupportsCompressor` during a rollout. There might be other concerns I'm not thinking of though, so it'd be good to get an LGTM from @vadimberezniker and/or @tylerwilliams as well

bduffany (approved):

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Thanks for the call out. I'll add the following note to the commit description too to reduce confusion

I changed this because I realized I was hardcoding "Compressor: repb.Compressor_IDENTITY" when queuing the data to be copied below. FindMissing is always called with resources with the identity compressor anyway, so this shouldn't matter. But it's technically incorrect and could cause a subtle bug later

enterprise/server/backends/migration_cache/migration_cache.go
maggie-lou: Sounds good

[Distributed cache will return SupportsCompressor=false ](https://github.com/buildbuddy-io/buildbuddy/pull/2962/files#diff-9410f9cf8e952fbcc75b77758e162afd98b0225ec51988fe9e2254a911b34efaR1012)unless all the peers return SupportsCompressor=true. So during a rollout, distributed cache should return SupportsCompressor=false and the migration cache should return SupportsCompressor=false. Once the rollout has proceeded far enough for distributed SupportsCompressor=true, then the migration cache should also return SupportsCompressor=true

maggie-lou: @vadimberezniker I think this code should be safe for the roll-out with the distributed roll out plan we discussed in that other PR but could you help sanity check me and brandon?

vadimberezniker (approved):

PR #2974, opened 2022-11-23
Wrap CLI arguments containing spaces in quotation marks when rendered on the invocation details card.

iain-macdonald:
**Version bump**: Patch
**Related issues**: Fixes #1756

### DIFF
--- a/app/invocation/invocation_details_card.tsx
+++ b/app/invocation/invocation_details_card.tsx
@@ -44,26 +44,28 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
   // and this will be returned:
   //   ["bazel", "build", "--output_filter='argument with spaces'"," "//..."],
   quote(pieces: string[]) {
-    return pieces.map((value) => {
-      if (value.includes("=") && value.includes(" ")) {
-        // shlex.quote everything after the first '=' so that arguments like:
-        // --flag="blah blah blah" are properly escaped.
-        let parts: string[] = value.split("=", 2);
-        return parts.map((val) => shlex.quote(val)).join("=");
-      }
-      return value;
-    })
-    .join(" ");
+    return pieces
+      .map((value) => {
+        if (value.includes("=") && value.includes(" ")) {
+          // shlex.quote everything after the first '=' so that arguments like:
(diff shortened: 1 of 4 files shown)


app/invocation/invocation_details_card.tsx
```diff
@@ -33,17 +34,40 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
...
+  // example, this command:
+  //   "bazel build --output_filter='argument with spaces' //..."
+  // is passed into this function as:
+  //   ["bazel", "build", "--output_filter=argument with spaces"," "//..."],
+  // and this will be returned:
+  //   ["bazel", "build", "--output_filter='argument with spaces'"," "//..."],
+  quote(pieces: string[]) {
+    return pieces
+      .map((value) => {
+        if (value.includes("=") && value.includes(" ")) {
```
bduffany, line 49: nit: remove ` && value.includes(" ")` so that other chars like `"` or `'` get quoted too (if the part after the `=` doesn't need quoting then shlex will leave it untouched)

bduffany (approved):

app/invocation/invocation_details_card.tsx
```diff
@@ -33,17 +34,40 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
...
+          // shlex.quote everything after the first '=' so that arguments like:
+          // --flag="blah blah blah" are properly escaped.
+          let parts: string[] = value.split("=", 2);
+          return parts.map((val) => shlex.quote(val)).join("=");
```
bduffany, line 53: was playing around with this since I didn't know `.split()` takes a second argument --

looks like this doesn't return the expected value in some cases

```
> console.log(quote(['--foo=ba r=baz']))
"--foo='ba r'"
```

app/invocation/invocation_details_card.tsx
iain-macdonald: Ah good call, done.

app/invocation/invocation_details_card.tsx
iain-macdonald: Whoops, fixed!

bduffany (approved):

PR #2982, opened 2022-11-26
Log claimTask errors

tylerwilliams:


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1144,10 +1144,13 @@ func (s *SchedulerServer) claimTask(ctx context.Context, taskID string, claimTim
 		// Success
```
bduffany, line 1144: do we also want to log on line 1134?

```go
	r, err := redisAcquireClaim.Run(ctx, s.rdb, []string{s.redisKeyForTask(taskID)}).Result()
	if err != nil {
		log.CtxErrorf(ctx, "claimTask error: redis script failed: %s", err)
		return err
	}
```

maybe 1139 too, logging the type info

```go
	c, ok := r.(int64)
	if !ok {
		log.CtxErrorf(ctx, "unexpected result from claim attempt: %v (type %T)", r, r)
		return status.FailedPreconditionErrorf("unexpected result from claim attempt: %v (type %T)", r, r)
	}
```

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1144,10 +1144,13 @@ func (s *SchedulerServer) claimTask(ctx context.Context, taskID string, claimTim
...
 		break
 	case 10:
+		log.CtxErrorf(ctx, "claimTask %q error: task does not exist", taskID)
```
bduffany, line 1147: Looking at executor logs, these are extremely common too -- maybe we can log these at Info level?

```suggestion
		log.CtxInfof(ctx, "claimTask %q error: task does not exist", taskID)
```

bduffany (approved):

PR #2983, opened 2022-11-28
CLI: Add an `analyze` command for dependency graph analysis

bduffany:
* Add a `bb analyze` subcommand which accepts a PATTERN argument (defaulting to `//...`) and then analyzes the dependency graph for that pattern, using various analysis types. By default all analysis types are enabled, but specific analyses can be selected using flags.
* Add a "cost" analysis which takes into account the number of changes to a target's source files over a given time period (defaults to 4 weeks), as well as the number of transitive deps of a target. Roughly speaking it is the minimum number of invalidated actions that are caused by changes to a target's source files. (I say "roughly" and "minimum" because it only counts commits and not edits, doesn't account for historical changes to the build graph structure, and only counts the current branch, e.g. the main branch)
* Add a "longest path" analysis which shows the longest dependency path through the graph.

The analysis is pretty basic at the moment, but we can extend it in various ways by augmenting the nodes in the graph with "weights" corresponding to things like cache upload/download metrics, build times, etc., possibly using the BB API.

Example output (stderr logging output is omitted)

```
$ bb analyze //enterprise/server --cost
RANK    CHANGED AFFECTS COST    TARGET
1       12      126     1512    //server/interfaces:interfaces
2       2       258     516     //proto:remote_execution_proto
3       3       134     402     //proto:raft_proto
4       1       218     218     //proto:cache_proto
5       1       210     210     //proto:usage_proto
6       2       100     200     //server/util/alert:alert
7       1       171     171     //server/util/log:log
8       2       85      170     //server/util/flagutil/types:types
9       2       68      136     //server/util/query_builder:query_builder
10      1       127     127     //server/tables:tables
11      1       112     112     //proto:remote_asset_proto
12      1       112     112     //server/metrics:metrics
13      1       109     109     //server/environment:environment
14      14      7       98      //server/backends/disk_cache:disk_cache
15      4       22      88      //server/util/db:db
16      2       39      78      //server/remote_cache/digest:digest
17      3       22      66      //app/flame_chart:flame_chart_typings
18      3       20      60      //app/invocation:invocation_typings
19      10      5       50      //enterprise/server/backends/pebble_cache:pebble_cache
20      2       25      50      //server/util/retry:retry
```

```
$ bb analyze --longest_path
LENGTH  TARGET
0       //website:serve
1       //website:website
2       //enterprise/server/cmd/server/yaml_doc:generate_mdx
3       //enterprise/server/cmd/server/yaml_doc:generate_yaml
4       //enterprise/server/cmd/server/yaml_doc:yaml_doc
5       //enterprise/server/cmd/server/yaml_doc:yaml_doc_lib
6       //enterprise/server/cmd/server:server_lib
7       //server/libmain:libmain
8       //:bundle
9       //app:sha
10      //app:app_bundle
11      //app:app
12      //app:app_typings
13      //app/root:root
14      //app/root:root_typings
15      //app/invocation:invocation
16      //app/invocation:invocation_typings
17      //app/target:target
18      //app/target:target_typings
19      //app/docs:docs
20      //app/docs:docs_typings
21      //app/components/button:button
22      //app/components/button:button_typings
23      //app/components/link:link
24      //app/components/link:link_typings
25      //app/router:router
26      //app/router:router_typings
27      //app/auth:user
28      //app/auth:user_typings
29      //app/service:service
30      //app/service:service_typings
31      //proto:buildbuddy_service_ts_proto
32      //proto:_buildbuddy_service_ts_proto_pbts
33      //proto:_buildbuddy_service_ts_proto_pbjs
34      //proto:_buildbuddy_service_ts_proto_protos
35      //proto:buildbuddy_service_proto
36      //proto:invocation_proto
37      //proto:cache_proto
38      //proto:resource_proto
39      //proto:remote_execution_proto
40      //proto:scheduler_proto
41      //proto:context_proto
42      //proto:user_id_proto
43      //proto:user_id.proto
```

---

**Version bump**: Patch

### DIFF
--- a/cli/analyze/analyze.go
+++ b/cli/analyze/analyze.go
@@ -1,0 +1,559 @@
+package analyze
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"os/exec"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/cli/arg"
+	"github.com/buildbuddy-io/buildbuddy/cli/bazelisk"
+	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	"github.com/buildbuddy-io/buildbuddy/cli/workspace"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/protobuf/proto"
+
+	bqpb "github.com/buildbuddy-io/buildbuddy/proto/bazel_query"
+)
+
+const (
+	fourWeeks   = 4 * 7 * 24 * time.Hour // 4 weeks
+	targetLimit = 20
+)
+
+var (
+	flags = flag.NewFlagSet("analyze", flag.ContinueOnError)
+
+	longestPathFlag = flags.Bool("longest_path", false, "Show the longest path in the build graph.")
+
+	costFlag  = flags.Bool("cost", false, "Analyze dependency costs for the target.")
+	sinceFlag = flags.Duration("since", fourWeeks, "How far back to look in git history to determine the number of edits.")
+
+	usage = `
+usage: bb ` + flags.Name() + ` [PATTERN]
+
+Analyzes the dependency graph for the given PATTERN, attempting to identify
+opportunities for restructuring your build in order to improve performance.
+
+The simplest usage is to run all analyses for all targets in the current repo,
+using the default //... as PATTERN argument:
+
+    bb ` + flags.Name() + `
+
+There are a few different analysis types available:
+
+    bb ` + flags.Name() + ` PATTERN --longest_path
+
+Shows the longest path in the dependency graph. Long dependency chains cannot
+be built in parallel, and slow down the whole build.
+
+    bb ` + flags.Name() + ` PATTERN --cost [--since=672h]
+
+Prints a cost metric for all transitive dependencies of a target in the current
+repository, based on the git change history of each dependency as well as the
+number of transitive dependencies.
+
+Targets with the highest cost are good candidates for splitting up into more
+granular targets. For example, it might make sense to split up a large,
+frequently changing utility library into smaller utilities, or split up a large
+piece of business logic into separate API and implementation targets.
+
+The cost of a target is computed roughly as the number of cache invalidations
+caused by changes to the target's source files. So if a target has 5 transitive
+dependents, and its sources were edited 10 times in the last month, its cost is
+5 * 10 = 50. The lookback duration defaults to 4 weeks, and can be modified
+using the --since flag.
+`
+)
+
+func HandleAnalyze(args []string) (int, error) {
+	cmd, idx := arg.GetCommandAndIndex(args)
+	if cmd != flags.Name() {
+		return -1, nil
+	}
+	if err := arg.ParseFlagSet(flags, args[idx+1:]); err != nil {
+		if err == flag.ErrHelp {
+			log.Print(usage)
+			return 1, nil
+		}
+		return -1, err
+	}
+
+	// TODO: Support more than one target
+	if len(flags.Args()) > 1 {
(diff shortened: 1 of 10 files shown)


cli/analyze/analyze.go
```diff
@@ -0,0 +1,559 @@
...
+		return nil, fmt.Errorf("%s", strings.TrimSpace(stderr.String()))
+	}
+	output := strings.TrimRight(stdout.String(), "\n")
+	if output == "" {
+		return nil, nil
+	}
+	return strings.Split(output, "\n"), nil
+}
+
+func anyKey[K comparable, V any](m map[K]V) *K {
```
tylerwilliams, line 521: Could you add comments to the functions here and below?

cli/analyze/analyze.go
```diff
@@ -0,0 +1,559 @@
...
+
+func getAllSourceFilePaths(dir string) (map[string]bool, error) {
+	lines, err := getOutputLines(dir, "git", "ls-files", "--exclude-standard")
+	if err != nil {
+		return nil, fmt.Errorf("git ls-files failed: %s", err)
+	}
+	return makeSet(lines), nil
+}
+
+func getOutputLines(dir, executable string, args ...string) ([]string, error) {
```
tylerwilliams, line 504: This is pretty darn close to https://pkg.go.dev/os/exec#Cmd.Output .

cli/analyze/analyze.go
```diff
@@ -0,0 +1,559 @@
...
+			s = append(s, m)
+		}
+	}
+	return &EdgeSet{Outgoing: outgoing, Incoming: incoming}
+}
+
+// TopologicalSort returns all node labels in the graph sorted so that each node
+// is sorted before any of its transitive deps.
+func (g *DependencyGraph) TopologicalSort() []string {
+	// https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm
```
tylerwilliams, line 478: Not to be confused with [Kuhn's algorithm](https://en.wikipedia.org/wiki/Kuhn%27s_algorithm).

tylerwilliams (changes requested):
this is pretty cool and a great MVP :) I'd love it if we could collect this data and visualize in the UI for more builds somehow

cli/analyze/analyze.go
bduffany: two differences -
- It properly splits output lines, since the most obvious implementation is `strings.Split(string(output), "\n")`, but that's incorrect since it results in the last line being blank (due to a trailing newline)
- The `err` returned by `cmd.Output()` doesn't include the command's stderr, which is important for understanding what went wrong. Instead the error message is something like "exit code 1". This func instead returns the command's stderr as the error, which is suitable for these git commands since they fail fast with succinct one-line error messages.

cli/analyze/analyze.go
bduffany: Done

cli/bazelisk/bazelisk.go
```diff
@@ -180,6 +181,30 @@ func makePipeWriter(w io.Writer) (pw *os.File, closeFunc func(), err error) {
 	return
 }
 
+// Redirects either os.Stdout or os.Stderr to the given writer. Calling the
+// returned close function stops redirection.
+func redirectStdio(w io.Writer, stdio **os.File) (close func(), err error) {
```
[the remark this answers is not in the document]
bduffany: This func needs to assign to `os.Stdout`, so it takes a pointer argument to `os.Stdout` so that it can do `*stdio = ...` to assign to it. But `os.Stdout` itself has type `*os.File`, so it needs to take a `**os.File`.

tylerwilliams (approved):

PR #2989, opened 2022-11-29
Make artifacts viewable in code editor

siggisim:
Screenshot of view button next to artifacts:
Screen Shot 2022-11-28 at 5 04 39 PM

Screenshot of single file code editor view:
Screen Shot 2022-11-28 at 5 04 44 PM

---

**Version bump**: Patch

### DIFF
--- a/app/root/root.css
+++ b/app/root/root.css
@@ -1138,7 +1138,6 @@ code .comment {
   justify-content: center;
   padding: 0 8px;
   color: #888;
-  margin-top: 4px;
   white-space: nowrap;
 }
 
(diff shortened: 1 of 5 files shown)


bduffany (approved):
nice!

app/root/root.css
```diff
@@ -1121,11 +1121,27 @@ code .comment {
...
   font-weight: 600;
   text-decoration: underline;
   cursor: pointer;
-  display: block;
+}
+
+.artifact-view {
+  font-weight: 600;
+  display: flex;
+  align-items: center;
```
bduffany, line 1137: (woops, my comment didn't make it through)

nit: maybe align to `baseline` so that the "View" text lines up vertically with the artifact link? (Might need to tweak the code icon manually to get it to look nice)

siggisim: Looks like the issue was the lack of a height constraint on the icon and some margins - fixed.

PR #1223, opened 2021-11-11
Add download and text-wrap buttons to build logs

siggisim:
---

**Version bump**: Patch

PR #2990, opened 2022-11-29
Support sorting invocations on the history page.

iain-macdonald:
This PR adds a new chip to the history page for sorting invocations. I added a handful of sort options including:
- Invocation start time
- Invocation end time
- Invocation duration
- Action Cache hit ratio
- Content Addressable Store hit ratio
- Cache download
- Cache upload
- Cache transfer

It supports sorting by any of these in ascending or descending order. It should be pretty easy to add new sort parameters in the future too. Here are some screenshots:

Screenshot 2022-11-30 at 8 33 07 AM
Screenshot 2022-11-30 at 8 33 29 AM
Screenshot 2022-11-30 at 8 33 40 AM

---

**Version bump**: Minor
**Related issues**: Fixes #1223


siggisim:
This is awesome! I'll let @bduffany do the review - but what do you think about putting the dropdown in between the filter and date dropdowns? That way filter and sort are next to one another (which feel the most closely related).

Also wondering if the default button shouldn't have any text (just the sort icon) until the default sort order is changed (maybe this is already the case!) so it doesn't take up as much space.

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -174,15 +174,36 @@ func (s *InvocationSearchService) QueryInvocations(ctx context.Context, req *inp
...
+		sort = defaultSortParams()
+	} else if req.Sort.SortField == inpb.InvocationSort_UNKNOWN_SORT_FIELD {
+		sort.SortField = defaultSortParams().SortField
 	}
 	switch sort.SortField {
 	case inpb.InvocationSort_CREATED_AT_USEC_SORT_FIELD:
 		q.SetOrderBy("created_at_usec", sort.Ascending)
 	case inpb.InvocationSort_UPDATED_AT_USEC_SORT_FIELD:
 		q.SetOrderBy("updated_at_usec", sort.Ascending)
+	case inpb.InvocationSort_DURATION_SORT_FIELD:
```
luluz66, line 188: I'm a bit concerned about the query performance, since these fields are not indexed. Have you profiled these queries or tested against our biggest customers?

enterprise/app/filter/filter.tsx
```diff
@@ -169,6 +197,22 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
 
+  private onSortByToggle(sortBy: string, selected: string) {
+    selected = sortBy;
+    router.setQuery({
+      ...Object.fromEntries(this.props.search.entries()),
+      [SORT_BY_NAME]: selected,
+    });
+  }
+
+  private onSortOrderToggle(sortOrder: string, selected: string) {
```
bduffany, line 208: nit: replace `Toggle` with `Change` in these 2 function names since it's changing the value of a radio selection rather than toggling a single checkbox element

app/router/router.tsx
```diff
@@ -19,6 +19,11 @@ export const COMMIT_PARAM_NAME = "commit";
 export const HOST_PARAM_NAME = "host";
 export const COMMAND_PARAM_NAME = "command";
 
+// Sort params for the global filter.
+
+export const SORT_BY_NAME = "sort-by";
+export const SORT_ORDER_NAME = "sort-order";
```
bduffany, line 25: (to match the consts above)

```suggestion
export const SORT_BY_PARAM_NAME = "sort-by";
export const SORT_ORDER_PARAM_NAME = "sort-order";
```

enterprise/app/filter/filter.tsx
```diff
@@ -385,13 +450,56 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
-                    <FilledButton onClick={this.handleApplyClicked.bind(this)}>Apply</FilledButton>
+                    <FilledButton onClick={this.handleFilterApplyClicked.bind(this)}>Apply</FilledButton>
                   </div>
                 </div>
               </form>
             )}
           </Popup>
         </div>
+        <div className="popup-wrapper">
+          <OutlinedButton className="sort-button icon-text-button" onClick={this.onOpenSortMenu.bind(this)}>
```
bduffany, line 461: nit: in the case where the button text is empty, it looks a bit too wide. It would be nice to have it look nice and square like the filter icon. When the text is empty, you can use the class `icon-button` instead of `icon-text-button` and remove the empty `<span>` element.

bduffany (changes requested):

enterprise/app/filter/filter.tsx
```diff
@@ -385,13 +450,56 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
+            {sortOrderValue == "asc" && <SortAsc className="icon" />}
+            {sortOrderValue == "desc" && <SortDesc className="icon" />}
+            <span>
+              {sortByValue == "start-time" && "Invocation Time (Start)"}
+              {sortByValue == "end-time" && "Invocation Time (End)"}
```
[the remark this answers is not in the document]
siggisim: Cool with anything here! "Start time" and "End time" sounds reasonable

enterprise/app/filter/filter.tsx
```diff
@@ -169,6 +195,22 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
+    });
+  }
+
+  private onSortOrderToggle(sortOrder: string, selected: string) {
+    selected = sortOrder;
+    router.setQuery({
+      ...Object.fromEntries(this.props.search.entries()),
+      [SORT_ORDER_NAME]: selected,
+    });
+  }
```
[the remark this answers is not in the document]
iain-macdonald: Good call, removed.

enterprise/app/filter/filter.tsx
```diff
@@ -204,6 +246,24 @@ export default class FilterComponent extends React.Component<FilterProps, State>
     });
   }
 
+  private renderSortByRadio(label: string, sortBy: string, selected: string) {
+    return (
+      <label onClick={this.onSortByToggle.bind(this, sortBy, selected)}>
+        <Radio checked={selected == sortBy} />
```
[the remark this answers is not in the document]
iain-macdonald: Whoops, thanks for pointing that out.

enterprise/app/history/history.tsx
```diff
@@ -53,6 +53,8 @@ interface Props {
   user?: User;
   search: URLSearchParams;
   hash: string;
+  sortBy?: string;
```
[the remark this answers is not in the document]
iain-macdonald: Done.

enterprise/app/filter/filter.tsx
```diff
@@ -385,13 +450,56 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
+              {sortByValue == "start-time" && "Invocation Time (Start)"}
+              {sortByValue == "end-time" && "Invocation Time (End)"}
+              {sortByValue == "duration" && "Invocation Duration"}
+              {sortByValue == "ac-hit-ratio" && "AC Hit Ratio"}
+              {sortByValue == "cas-hit-ratio" && "CAS Hit Ratio"}
+              {sortByValue == "cache-down" && "Cache Download"}
+              {sortByValue == "cache-up" && "Cache Upload"}
+              {sortByValue == "cache-xfer" && "Cache Transfer"}
+            </span>
+          </OutlinedButton>
```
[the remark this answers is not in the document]
iain-macdonald: Done.

enterprise/app/filter/filter.tsx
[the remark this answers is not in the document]
iain-macdonald: Yeah I prefer those as well, changed.

enterprise/app/filter/filter.tsx
iain-macdonald: Done.

enterprise/app/filter/filter.tsx
iain-macdonald: Good call, fixed.

enterprise/server/invocation_search_service/invocation_search_service.go
iain-macdonald: Thanks for pointing that out. In the common case (no sorting), the query to the database will be the same, so that shouldn't be affected by this PR. I suspect you're right that some of the other queries will be slower, but I don't think we should block this PR because this is a new feature and we can look into speeding it up if it proves to be a problem. Does that seem reasonable?

iain-macdonald (commented):
Thanks, Brandon & Lulu! I think I addressed all your comments, PTAL.

bduffany:
It looks like there are a few typecheck failures -- we removed type checking from the build and made them tests instead, since TS type checking is super slow. To run the type check tests locally you can do

```
bb test //enterprise/app/... //app/...
```

enterprise/app/filter/filter.tsx
```diff
@@ -251,6 +311,9 @@ export default class FilterComponent extends React.Component<FilterProps, State>
       };
     });
 
+    const sortByValue = this.props.search.get(SORT_BY_PARAM_NAME) || "";
+    const sortOrderValue = this.props.search.get(SORT_ORDER_PARAM_NAME) || "desc";
```
bduffany, line 315: TS will type these as string, so to get the type checking benefits you can change this to

```suggestion
    const sortByValue: SortBy | "" = (this.props.search.get(SORT_BY_PARAM_NAME) as SortBy) || "";
    const sortOrderValue: SortOrder = (this.props.search.get(SORT_ORDER_PARAM_NAME) as SortOrder) || "desc";
```

enterprise/app/history/history.tsx
```diff
@@ -97,6 +118,10 @@ export default class HistoryComponent extends React.Component<Props, State> {
         command: filterParams.command,
         groupId: this.props.user?.selectedGroup?.id,
       }),
+      sort: new invocation.InvocationSort({
+        sortField: this.getSortField(filterParams),
+        ascending: this.props.sortOrder === "asc" || filterParams.sortOrder === "asc" ? true : false,
```
bduffany, line 123:
```suggestion
        ascending: filterParams.sortOrder === "asc",
```

enterprise/app/filter/filter.tsx
iain-macdonald: Whoops, done. I went ahead and added "" as a valid SortBy value too so sortByValue can be just a SortBy here.

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -174,15 +174,36 @@ func (s *InvocationSearchService) QueryInvocations(ctx context.Context, req *inp
...
+	case inpb.InvocationSort_ACTION_CACHE_HIT_RATIO_SORT_FIELD:
+		// Treat 0/0 as 100% cache hit rate to avoid divide-by-zero weirdness.
+		q.SetOrderBy(`IFNULL(
+			action_cache_hits / (action_cache_hits + action_cache_misses), 1)`,
+			sort.Ascending)
+	case inpb.InvocationSort_CONTENT_ADDRESSABLE_STORE_CACHE_HIT_RATIO_SORT_FIELD:
+		// Treat 0/0 as 100% cache hit rate to avoid divide-by-zero weirdness.
+		q.SetOrderBy(`IFNULL(
+			cas_cache_hits / (cas_cache_hits + cas_cache_misses), 1)`,
+			sort.Ascending)
```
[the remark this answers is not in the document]
iain-macdonald: Whoops, forgot to comment on this...

I thought about this, and added the special divide-by-zero logic to deal with those items being ranked weird. I figured that for now this is probably good enough and if we hear feedback we can update this calculation to use a weight average or something more sophisticated.

iain-macdonald (commented):
Thanks, Brandon, PTAL!

bduffany (approved):
Awesome!

enterprise/app/filter/filter.tsx
```diff
@@ -385,13 +450,58 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
+            <span>
+              {sortByValue === "start-time" && "Start time"}
+              {sortByValue === "end-time" && "End time"}
+              {sortByValue === "duration" && "Duration"}
+              {sortByValue === "ac-hit-ratio" && "AC hit ratio"}
+              {sortByValue === "cas-hit-ratio" && "CAS hit ratio"}
+              {sortByValue === "cache-down" && "Cache download"}
+              {sortByValue === "cache-up" && "Cache upload"}
+              {sortByValue === "cache-xfer" && "Cache transfer"}
+            </span>
```
bduffany, line 475: (woops, forgot to add this in my previous review)

nit: the icon doesn't appear centered because of the empty `<span>` in the case where none of the labels is rendered.

iain-macdonald: Oh shoot, you left this comment in an earlier review and I accidentally ignored it. Fixed and updated screenshot in PR with the fix.

PR #2991, opened 2022-11-29
Integration tests for writing executions to olapdb

luluz66:
1. Add a testOLAPDBHandle which records invocation ids and execution ids that were flushed.
2. add testOLAPDBHandle and Execution Collector to buildbuddy server in rbetest env
3. set "app.enable_write_to_olap_db" to true in the test.
4. Add a ShutdownBuildbuddyServers() function in rbeEnv. We need to wait for the statsRecorder to pick up the task to write invocation and executions to olap db. We need to call this shutdown function to make sure that statsRecorder tasks run before we inspect the testOLAPDBHandle.
5. Adding assertions to current remote execution tests in bazel_rbe_test


enterprise/server/testutil/enterprise_testenv/enterprise_testenv.go
```diff
@@ -37,6 +38,10 @@ func GetCustomTestEnv(t *testing.T, opts *Options) *testenv.TestEnv {
 		cache := redis_cache.NewCache(redisClient)
 		env.SetCache(cache)
 	}
+
+	if flag.Lookup("app.enable_write_to_olap_db") != nil {
+		flags.Set(t, "app.enable_write_to_olap_db", opts.EnableWriteToOLAPDB)
```
bduffany, line 43: nit: you can also write `flags.Set(t, "app.enable_write_to_olap_db", true)` in `NewRBETestEnv(t)` to avoid introducing another field in `Options`

server/testutil/testolapdb/testolapdb.go
```diff
@@ -0,0 +1,62 @@
...
+
+	iepb "github.com/buildbuddy-io/buildbuddy/proto/internal_execution"
+)
+
+type TestOLAPDBHandle struct {
+	executionIDsByInvID sync.Map // map of invocationID => a slice of execution IDs
+	invIDs              sync.Map // map of invocationID => struct{}
+}
+
+func NewTestOLAPDBHandle() *TestOLAPDBHandle {
```
bduffany, line 20: nit: you could maybe call this `New()` so that the callsite reads like `testolapdb.New()`

`TestOLAPDBHandle` can also be named `Handle` so it reads like `testolapdb.Handle`

PR #2999, opened 2022-12-01
Use buffered IO for distributed cache backhaul

tylerwilliams:
I wanted to experiment with the idea of using bigger chunks for reads and writes between ddisk peers to reduce chattiness, and improve performance.


vadimberezniker (approved):
change look good but doesn't match PR description

tylerwilliams:
Updated description and PR quite a bit. PTAL :)

enterprise/server/util/cacheproxy/cacheproxy.go
```diff
@@ -319,8 +324,9 @@ func (c *CacheProxy) Read(req *dcpb.ReadRequest, stream dcpb.DistributedCache_Re
 	if resourceSize > 0 && resourceSize < bufSize {
 		bufSize = resourceSize
 	}
+	bufferedReader := bufio.NewReaderSize(reader, int(bufSize))
```
vadimberezniker, line 327: I think this is a no-op since CopyBuffer is already using a buffer of size bufSize.

vadimberezniker (changes requested):

enterprise/server/util/cacheproxy/cacheproxy.go
tylerwilliams: I think you're right. I'd prefer to use bufio.NewReader but I think it would require pooling those Readers so i'm going to just leave this part as-is for now.

vadimberezniker (approved):

PR #3001, opened 2022-12-01
Fetch ExecutionStat in GetTrendResponse

luluz66:
Also Modify ExecutionStat Proto


enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -195,13 +196,99 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
 				return nil, err
 			}
 		}
-		rsp.TrendStat = append(rsp.TrendStat, stat)
+		res = append(res, stat)
+	}
```
bduffany, line 200: do you need to check `rows.Err()` after the loop here?

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -195,13 +196,99 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
...
+
+	res := make([]*inpb.ExecutionStat, 0)
+
+	for rows.Next() {
+		stat := &inpb.ExecutionStat{}
+		if err := i.olapdbh.DB(ctx).ScanRows(rows, &stat); err != nil {
+			return nil, err
+		}
+		res = append(res, stat)
 	}
```
bduffany, line 253: check rows.Err() here too?

bduffany (approved):

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -195,13 +195,88 @@ func (i *InvocationStatService) GetTrend(ctx context.Context, req *inpb.GetTrend
...
+
+func (i *InvocationStatService) getExecutionTrendQuery(timezoneOffsetMinutes int32) string {
+	return fmt.Sprintf("SELECT %s as name,", i.olapdbh.DateFromUsecTimestamp("updated_at_usec", timezoneOffsetMinutes)) + `
+	quantilesExactExclusive(0.5, 0.75, 0.9, 0.95, 0.99)(IF(worker_start_timestamp_usec > queued_timestamp_usec, worker_start_timestamp_usec - queued_timestamp_usec, 0)) AS queue_duration_usec_quantiles
+	FROM Executions
+	`
+}
+
+func getQueryWithFlattenedArray(innerQuery string) string {
+	return `SELECT name,
```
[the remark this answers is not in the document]
luluz66: This was my first attempt. But I ended up getting error "[error] unsupported data type: &[]" form gorm. I think in order to let it parse to an array, i need to define a custom data type that implement Scanner and Valuer interface. 

The nested query performance is acceptable. For a big customer that I tested with 30 days data, it's like 0.04 seconds for the query. So I didn't go for the custom data type route for now.

enterprise/server/invocation_stat_service/invocation_stat_service.go
luluz66: Done

enterprise/server/invocation_stat_service/invocation_stat_service.go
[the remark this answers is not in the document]
luluz66: Done

enterprise/server/invocation_stat_service/invocation_stat_service.go
luluz66: Done.

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -23,6 +23,7 @@ import (
 
 var (
 	readFromOLAPDBEnabled = flag.Bool("app.enable_read_from_olap_db", false, "If enabled, complete invocations will be flushed to OLAP DB")
+	executionTrendEnabled = flag.Bool("app.enable_execution_trend", false, "If enabled, fill execution trend stat in GetTrendResponse")
```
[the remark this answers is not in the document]
luluz66: Done.

bduffany (approved):

PR #3002, opened 2022-12-01
Add a graph for remote execution percentile chart

luluz66:
Screenshot:
Screen Shot 2022-11-30 at 7 23 14 PM


siggisim:
This is awesome! What are your thoughts on separating RPCs for the the execution from invocation trends? 

It could be the same endpoint just with a boolean/enum in the request that determines if you're requesting invocation or execution trends data, and the UI could make 2 requests. 

That way if one of the requests is much slower than the other, the whole page isn't blocked on rendering until both are complete. 

We can wait to do this until we see in practice that one of the requests is much slower than the other (which may never happen).

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface Props {
+  title: string;
+  data: any[];
+  extractLabel: (datum: any) => string;
+  formatHoverLabel: (datum: any) => string;
+  extractP50: (datum: any) => number;
+  extractP75: (datum: any) => number;
+  extractP90: (datum: any) => number;
+  extractP95: (datum: any) => number;
+  extractP99: (datum: any) => number;
```
bduffany, line 15: can these `any` types be replaced by a stronger type?

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,80 @@
...
+const PercentilesChartTooltip = ({
+  active,
+  payload,
+  labelFormatter,
+  extractP50,
+  extractP75,
+  extractP90,
+  extractP95,
+  extractP99,
+}: any) => {
```
bduffany, line 27: replace `any` with a stronger type here too? (can write a separate non-exported interface like `interface PercentilesChartTooltipProps` to make it cleaner.)

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,80 @@
+import React from "react";
+
+import * as format from "../../../app/format/format";
+import { ResponsiveContainer, ComposedChart, CartesianGrid, XAxis, YAxis, Line, Legend, Tooltip } from "recharts";
+
+interface Props {
```
bduffany, line 6: nit: we aren't 100% consistent on this yet, but since the component is exported, the props type should be exported too as the props are part of the public API (and prefixed w/ component name to avoid name collisions)

```suggestion
export interface PercentilesChartProps {
```

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,80 @@
...
+  extractLabel: (datum: any) => string;
+  formatHoverLabel: (datum: any) => string;
+  extractP50: (datum: any) => number;
+  extractP75: (datum: any) => number;
+  extractP90: (datum: any) => number;
+  extractP95: (datum: any) => number;
+  extractP99: (datum: any) => number;
+}
+
+const PercentilesChartTooltip = ({
```
bduffany, line 18: nit: can you move this below `PercentilesChartComponent` since it's more of an implementation detail, and so that the chart props above are declared next to the chart component?

You can write it as `function PercentilesChartTooltip(...) {` instead of using an arrow function, to make the compiler happy.

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,80 @@
...
+    let data = payload[0].payload;
+    return (
+      <div className="trend-chart-hover">
+        <div className="trend-chart-hover-label">{labelFormatter(data)}</div>
+        <div className="trend-chart-hover-value">
+          <div>p50: {format.durationSec(extractP50(data) || 0)}</div>
+          <div>p75: {format.durationSec(extractP75(data) || 0)}</div>
+          <div>p90: {format.durationSec(extractP90(data) || 0)}</div>
+          <div>p95: {format.durationSec(extractP95(data) || 0)}</div>
+          <div>p99: {format.durationSec(extractP99(data) || 0)}</div>
```
bduffany, line 38: can you use `format.compactDurationSec` for these so that the y-axis tick labels don't wrap onto multiple lines?

enterprise/app/trends/percentile_chart.tsx
luluz66: Done

enterprise/app/trends/percentile_chart.tsx
luluz66: Done

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,121 @@
...
+        </ResponsiveContainer>
+      </div>
+    );
+  }
+}
+
+interface PercentilesChartTooltipProps {
+  active: boolean;
+  payload: any[];
+  labelFormaatter: (datum: string) => string;
```
bduffany, line 86:
```suggestion
  labelFormatter: (datum: string) => string;
```

not sure why the typecheck tests aren't failing :confused:

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,121 @@
...
+  active,
+  payload,
+  labelFormatter,
+  extractP50,
+  extractP75,
+  extractP90,
+  extractP95,
+  extractP99,
+}: PercentilesChartTooltipProps) {
+  if (active) {
```
bduffany, line 104: nit: would be slightly cleaner written as

```
if (!active) return null;

// (rest of function)
```

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -0,0 +1,121 @@
+import React from "react";
+
+import * as format from "../../../app/format/format";
+import { ResponsiveContainer, ComposedChart, CartesianGrid, XAxis, YAxis, Line, Legend, Tooltip } from "recharts";
+
+export interface PercentilesChartProps {
+  title: string;
+  data: any[];
```
bduffany, line 8:
```suggestion
  data: string[];
```

bduffany (approved):
Looks great!

enterprise/app/trends/percentile_chart.tsx
luluz66: Done

enterprise/app/trends/percentile_chart.tsx
luluz66: Done

enterprise/app/trends/percentile_chart.tsx
luluz66: Done

bduffany (approved):

luluz66:
> This is awesome! What are your thoughts on separating RPCs for the the execution from invocation trends?
> 
> It could be the same endpoint just with a boolean/enum in the request that determines if you're requesting invocation or execution trends data, and the UI could make 2 requests.
> 
> That way if one of the requests is much slower than the other, the whole page isn't blocked on rendering until both are complete.
> 
> We can wait to do this until we see in practice that one of the requests is much slower than the other (which may never happen).

We can wait to do this. my test from the prod data show the time to finish these two requests are similar.

PR #3011, opened 2022-12-01
Wait for replicas to catch up before adding them to the metarange

tylerwilliams:


enterprise/server/raft/store/store.go
```diff
@@ -1314,6 +1320,38 @@ func (s *Store) SplitCluster(ctx context.Context, req *rfpb.SplitClusterRequest)
...
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			break
+		}
+		if rd := s.lookupRange(clusterID); rd != nil {
+			if r, err := s.GetReplica(rd.GetRangeId()); err == nil {
+				if lastApplied, err := r.LastAppliedIndex(); err == nil {
```
vadimberezniker, line 1342: Should you return an error if either GetReplica or LastAppliedIndex fails? 
They should only error out in catastrophic cases, right?

vadimberezniker (approved):

PR #3012, opened 2022-12-01
Redact individual values in --build_metadata sub-flags instead of just redacting everything after the '='.

iain-macdonald:
I went back and forth a little bit in considering how to fix this bug. The issue is that when flags are specified as `--flag_name=KEY=@value` then the `KEY=@` matches the regular expression that tries to identify passwords (I think?) in flag values [here](https://github.com/buildbuddy-io/buildbuddy/blob/master/server/util/redact/redact.go#L31). Without the '@' it wouldn't match, which is why this bug isn't a bit more common.

The approach I took is to specifically look for the `--build_metadata` flag and when it's encountered treat the flag value as a list of comma-delimited, KEY=VALUE "sub-flags" and in this case just redact the values of these sub flags, leaving the keys intact. It should be easy to add additional flags to be treated this way in the future.

A couple of other approaches I considered included allowing specific patterns (like "PATTERN=@") bypass the redaction logic, which probably would have been fine, that's kind of what the defaultAllowedEnvVars is doing; or trying to add more complicated flag parsing logic that identifies when this happens without the list of specific flags, which seemed error-prone. I wanted to get a second opinion about whether the first approach (permitting specific patterns to pass redaction) would be preferable to what I have here.

---

**Version bump**: Patch 
**Related issues**: Fixes #1757

### DIFF
--- a/server/util/redact/redact.go
+++ b/server/util/redact/redact.go
@@ -40,7 +40,7 @@ var (
 	// parsed this way and have only the sub-values undergo redaction. This is
 	// the list of flag names that are permitted to be treated this way.
 	knownMultiFlags   = []string{"build_metadata"}
-	multiFlagKeyRegex = regexp.MustCompile(`[A-Z_]+=`)
+	multiFlagKeyRegex = regexp.MustCompile(`,[A-Z_]+=`)
 
 	knownGitRepoURLKeys = []string{
 		"REPO_URL", "GIT_URL", "TRAVIS_REPO_SLUG", "BUILDKITE_REPO",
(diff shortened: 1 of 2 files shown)


siggisim:
+ @bduffany who I think has the most context on redaction

server/util/redact/redact_test.go
```diff
@@ -109,6 +109,8 @@ func TestRedactMetadata_StructuredCommandLine(t *testing.T) {
 		{"remote_default_exec_properties", "container-registry-username=SECRET_USERNAME", "container-registry-username=<REDACTED>"},
 		{"remote_default_exec_properties", "container-registry-password=SECRET_PASSWORD", "container-registry-password=<REDACTED>"},
 		{"host_platform", "@buildbuddy_toolchain//:platform", "@buildbuddy_toolchain//:platform"},
+		{"build_metadata", "PATTERN=@//foo,NAME=@foo,PASSWORD=SECRET@bar,BAZ=", "PATTERN=@//foo,NAME=@foo,PASSWORD=bar,BAZ="},
```
bduffany, line 112: could you add a test for something like `{"build_metadata", "FOO=A=1,BAR=B=SECRET@foo", "FOO=A=1,BAR=B=@foo"}` to make sure it handles `=` inside the flag values themselves (assuming Bazel treats those as valid)

server/util/redact/redact.go
```diff
@@ -62,19 +74,70 @@ func stripURLSecrets(input string) string {
...
+}
+
+func isKnownMultiFlag(input string) bool {
+	for _, knownMultiFlag := range knownMultiFlags {
+		if strings.HasPrefix(input, knownMultiFlag) || strings.HasPrefix(input, "--"+knownMultiFlag) {
+			return true
+		}
+	}
+	return false
+}
```
bduffany, line 130: Would it be practical to treat non-multi flags as a special case of multi flags with a length of 1, so that everything uses the same multi-flag logic? Hoping we can avoid introducing separate special handling for single- and multi- valued flags.

bduffany (changes requested):

server/util/redact/redact.go
iain-macdonald: I thought about that. The drawback would be that if a single-value flag has an equals sign in it, then the portion before the equals sign wouldn't be subject to redaction, so I figured it'd be better to be cautious and only do this for a set of known flags. Maybe that's not an issue, though? WDYT?

bduffany: Ah fair point. I'm cool with leaving it as-is :+1:

server/util/redact/redact_test.go
iain-macdonald: Done. I had to change the flag parsing logic a little bit because it was matching the "A=" in the value.

PR #3014, opened 2022-12-02
Embed git version info via build flag

bduffany:
* Add a starlark flag `--//server/version:version_tag` that explicitly sets the git version tag variable in the `//server/version` go package.
  * If the flag is unset, it'll default to "unknown".
  * The commit SHA should still be printed on server startup, as long as the build was done with `--stamp`
* Remove the VERSION file and all references to it, and reference the embedded git version tag instead.
* Update release.py to compute the latest version by fetching tags and listing the most recent one, instead of reading from the VERSION file.
* Update release.py to remove the VERSION file commit step.
* Update release.py to set the new version flag when building docker images.
* Clean up unused functions in release.py, to further simplify. We can always revive these functions later if needed.
* Update GitHub workflows to set the version flag on release binaries.
* Clean up the use of `--define=version=...` in GH workflows since that define flag is only needed when pushing Docker images.

(See [buildbuddy-internal changes](https://github.com/buildbuddy-io/buildbuddy-internal/pull/1838/files?diff=split&w=1) as well)

---

**Version bump**: Patch


release.py
```diff
@@ -90,24 +79,40 @@ def create_and_push_tag(old_version, new_version, release_notes=''):
     push_tag_cmd = 'git push origin %s' % new_version
     run_or_die(push_tag_cmd)
 
-def update_docker_image(new_version, update_latest_tag):
+def push_image(target, version_tag, image_tag):
+    # Note:
+    # The flag "--define=version=..." sets the Docker image tag.
```
tylerwilliams, line 84: maybe it would be clearer if we renamed this "version" to "image_tag" in our BUILD files?

release.py
```diff
@@ -220,23 +143,17 @@ def main():
     bump_version = not args.skip_version_bump
     update_latest_tag = not args.skip_latest_tag
 
-    if args.allow_dirty or workspace_is_clean():
-        print("workspace is clean!")
+    if workspace_is_clean():
+        print("Workspace is clean!")
+    if args.allow_dirty:
```
tylerwilliams, line 148: should this be elif?

release.py
bduffany: I agree -- I'll send a follow-up for that because there's a lot of places that need to be updated.

release.py
bduffany: oops, fixed

tylerwilliams (approved):

PR #3016, opened 2022-12-02
Always use workspace version of gazelle for buildfix.sh

bduffany:
Instead of using the gazelle binary in `$PATH`, always invoke Gazelle using `bazel run //:gazelle`.

---

**Version bump**: Patch


tempoz (approved):
LGTM

PR #3020, opened 2022-12-05
Set pushed image tag using starlark flag

bduffany:
Migrate from setting the version with `--define=version=TAG` to `--//deployment:image_tag=TAG`.

Unfortunately the analysis cache is still discarded when changing starlark flags (see https://github.com/bazelbuild/bazel/issues/13591), but I figured since the docs say that "--define" is "legacy syntax" it's worth migrating to a starlark flag either way.

---

**Version bump**: Patch

### DIFF
--- a/deployment/BUILD
+++ b/deployment/BUILD
@@ -1,7 +1,9 @@
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load("@k8s_deploy//:defaults.bzl", "k8s_deploy")
 load("@io_bazel_rules_k8s//k8s:object.bzl", "k8s_object")
 load("@io_bazel_rules_k8s//k8s:objects.bzl", "k8s_objects")
 load("@io_bazel_rules_docker//container:push.bzl", "container_push")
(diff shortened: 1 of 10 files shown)


deployment/BUILD
```diff
@@ -21,6 +23,22 @@ k8s_objects(
...
 )
 
+# Set --//deployment:image_tag to set the tag when running container_push
+# targets. If unset, defaults to "latest".
+#
+# Note that this tag is shared across all targets, which is fine since only one
+# container_push target can run at once.
+string_flag(
+    name = "image_tag",
+    build_setting_default = "latest",
```
vadimberezniker, line 33: Should we set this to something like `dev`? The flag should generally be set and if something accidentally doesn't specify it, I don't think we should default to touching the latest tag.

vadimberezniker (approved):

deployment/BUILD
bduffany: Done (used `development` instead of `dev` to avoid potential confusion with the dev environment)

tylerwilliams (approved):

PR #3021, opened 2022-12-05
Update Linux execution duration usage to only track shared executor usage

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/execution_server/execution_server.go
+++ b/enterprise/server/remote_execution/execution_server/execution_server.go
@@ -919,11 +919,25 @@ func (s *ExecutionServer) updateUsage(ctx context.Context, cmd *repb.Command, ex
 		return err
 	}
 	counts := &tables.UsageCounts{}
+	// TODO: Incorporate remote-header overrides here.
 	plat := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+
+	executorGroupID, _, err := s.env.GetSchedulerService().GetGroupIDAndDefaultPoolForUser(ctx, plat.OS, plat.UseSelfHostedExecutors)
+	if err != nil {
+		return status.InternalErrorf("failed to determine executor group ID: %s", err)
+	}
+	sharedExecutorPoolGroupID, err := s.env.GetSchedulerService().GetSharedExecutorPoolGroupID(ctx)
+	if err != nil {
+		return status.InternalErrorf("failed to determine shared executor pool group ID: %s", err)
+	}
+
 	if plat.OS == platform.DarwinOperatingSystemName {
 		counts.MacExecutionDurationUsec += dur.Microseconds()
 	} else if plat.OS == platform.LinuxOperatingSystemName {
-		counts.LinuxExecutionDurationUsec += dur.Microseconds()
+		// Only increment linux execution duration if using shared executors.
+		if executorGroupID == sharedExecutorPoolGroupID {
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -919,11 +919,25 @@ func (s *ExecutionServer) updateUsage(ctx context.Context, cmd *repb.Command, ex
...
+	}
+
 	if plat.OS == platform.DarwinOperatingSystemName {
 		counts.MacExecutionDurationUsec += dur.Microseconds()
 	} else if plat.OS == platform.LinuxOperatingSystemName {
-		counts.LinuxExecutionDurationUsec += dur.Microseconds()
+		// Only increment linux execution duration if using shared executors.
+		if executorGroupID == sharedExecutorPoolGroupID {
+			counts.LinuxExecutionDurationUsec += dur.Microseconds()
+		}
```
tylerwilliams, line 940: could we move this check above the plat.OS switch so it's true regardless of OS? It seems brittle to encode the fact that mac-executors are never (currently) shared into this logic

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -919,11 +919,25 @@ func (s *ExecutionServer) updateUsage(ctx context.Context, cmd *repb.Command, ex
 		return err
 	}
 	counts := &tables.UsageCounts{}
+	// TODO: Incorporate remote-header overrides here.
 	plat := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+
+	executorGroupID, _, err := s.env.GetSchedulerService().GetGroupIDAndDefaultPoolForUser(ctx, plat.OS, plat.UseSelfHostedExecutors)
```
vadimberezniker, line 925: Maybe a slightly cleaner approach would be to have the scheduler service return a struct with this information as well as indicate whether we are using shared executors.

enterprise/server/remote_execution/execution_server/execution_server.go
bduffany: Done

enterprise/server/remote_execution/execution_server/execution_server.go
bduffany: Done, PTAL. Not super thrilled with the naming, lmk if anything comes to mind that seems more appropriate.

vadimberezniker: Only other alternative I can think of is `PoolMetadata`/`GetPoolMetadata` but not sure that's much better.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -887,33 +887,55 @@ func NewSchedulerServerWithOptions(env environment.Env, options *Options) (*Sche
...
 	}
 	if useSelfHosted {
-		return user.GetGroupID(), "", nil
+		return selfHostedPool, nil
 	}
-	return *sharedExecutorPoolGroupID, *defaultPoolName, nil
+	return sharedPool, nil
+}
+
+func (s *SchedulerServer) GetSharedExecutorPoolGroupID(ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
bduffany: woops, removed

PR #3022, opened 2022-12-05
Disable race detection for GH action builds

bduffany:
Looks like these are failing on GH actions too, not just BB workflows -- probably because the default execution image (ubuntu 16.04) has an outdated gcc version.

---

**Version bump**: Patch


tempoz (approved):
LGTM

PR #3026, opened 2022-12-05
only add --dns and --dns-network when network is not none or "container:id"

luluz66:


siggisim (approved):

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -296,6 +292,12 @@ func (c *podmanCommandContainer) getPodmanRunArgs(workDir string) []string {
 	if networkMode != "" {
 		args = append(args, "--network="+networkMode)
 	}
+	// "--dns" and "--dns=search" flags are invalid when --network is set to none
+	// or "containder:id"
+	if networkMode != "none" && strings.HasPrefix(networkMode, "container") {
```
tylerwilliams, line 297: should this be ||? (based on the comment)

tylerwilliams: nm, re-read it. long day :)

tylerwilliams (approved):

PR #3029, opened 2022-12-06
Use bazel6 rc4.

vadimberezniker:
---

**Version bump**: None


siggisim:
> LGTM if it builds

image

PR #3038, opened 2022-12-08
CLI: Increase cache proxy write queue size

bduffany:
Lots of remote writes are getting dropped due to the small queue capacity. Increasing the queue length so that the sidecar can better utilize upload bandwidth once the invocation is complete.

---

**Version bump**: Patch

### DIFF
--- a/cli/cache_proxy/cache_proxy.go
+++ b/cli/cache_proxy/cache_proxy.go
@@ -31,7 +31,7 @@ var (
 )
 
 const (
-	queueBufferSize = 100
+	queueBufferSize = 10_000
 )
 
 // CacheProxy implements a local GRPC cache that proxies a remote GRPC cache.


siggisim (approved):

cli/cache_proxy/cache_proxy.go
```diff
@@ -31,7 +31,7 @@ var (
 )
 
 const (
-	queueBufferSize = 100
+	queueBufferSize = 1000
```
tylerwilliams, line 34: if these are just digests, you could probably do 10K or more without worrying about memory

tylerwilliams (approved):

cli/cache_proxy/cache_proxy.go
bduffany: Bumped to 10K.

As a side note, since the digests can have different sizes, I have mild concerns about creating a large backlog and having a background process that seems to be unexpectedly hogging upload speed, e.g. when trying to do a video call shortly after a `bb` build. We might want to make this based on the total size, rather than the number of uploads.

PR #3041, opened 2022-12-08
Clear login cookie if it's invalid

bduffany:
If the session cookie is invalid (points to a row in the DB that does not exist), this can cause the auth logic to get into a bad state where going through the login oauth flow will not grant a refresh token, because the login flow assumes that if the session cookie is set at all (valid or not) then we don't need to re-request a refresh token.

This can happen if the DB gets cleared, e.g. during local development (which uses a DB under /tmp/). I'm not sure if there are other situations where it can happen.

---

**Version bump**: Patch


siggisim (approved):

tylerwilliams (approved):
nice find

PR #3043, opened 2022-12-08
Delete old network namespaces in enable_local_firecracker.sh so people can use it to clean up after old Firecracker runs that didn't exit cleanly.

iain-macdonald:
**Version bump**: None

### DIFF
--- a/tools/enable_local_firecracker.sh
+++ b/tools/enable_local_firecracker.sh
@@ -79,5 +79,9 @@ if [ "$FOUND_ENTRY" != "$IP_ENTRY" ]; then
     echo "$IP_ENTRY" | EDITOR='tee -a' visudo >/dev/null
 fi
 
+# Clean up vestigial network namespaces, these can be left around if firecracker
+# doesn't exit cleanly and interfere with subsequent VM startup.
+ip --all netns delete
+
 echo "All done! You should be ready to run the executor as your user now."
 echo "You will need to run this program again if you restart or update the jailer binary"


tools/enable_local_firecracker.sh
```diff
@@ -79,5 +79,9 @@ if [ "$FOUND_ENTRY" != "$IP_ENTRY" ]; then
     echo "$IP_ENTRY" | EDITOR='tee -a' visudo >/dev/null
 fi
 
+# Clean up vestigial network namespaces, these can be left around if firecracker
+# doesn't exit cleanly and interfere with subsequent VM startup.
+ip --all netns delete
```
tylerwilliams, line 84: How would you feel about running this when the executor binary started up instead?

We sort of already assume that only one executor is running per-machine in prod, so this would be safe, and then it would also happen more regularly (since people almost never use this script)

server/util/networking/networking.go
```diff
@@ -57,6 +57,12 @@ func namespace(netNamespace string, args ...string) []string {
 	return append([]string{"ip", "netns", "exec", netNamespace}, args...)
 }
 
+// Deletes all of the net namespaces. These can be left behind if the executor
+// doesn't exit gracefully.
+func DeleteNetNamespaces(ctx context.Context) error {
+	return runCommand(ctx, "ip", "--all", "netns", "delete")
```
bduffany, line 63: I'm not sure that the executor will be the only thing that creates network namespaces on the host system, especially for local development and on-prem.

wdyt about creating all net namespaces under a common prefix like `bb-executor` and then deleting only the ones matching the prefix?

tools/enable_local_firecracker.sh
iain-macdonald: Sounds good, done.

server/util/networking/networking.go
iain-macdonald: Good idea, changed and added a flag to support multiple executors on a machine like we discussed.

server/util/networking/networking.go
```diff
@@ -54,14 +57,22 @@ func runCommand(ctx context.Context, args ...string) error {
...
-	return append([]string{"ip", "netns", "exec", netNamespace}, args...)
+	return append([]string{"ip", "netns", "exec", netNamespacePrefix + netNamespace}, args...)
+}
+
+// Deletes all of the executor net namespaces. These can be left behind if the
+// executor doesn't exit gracefully. Unfortunately, "ip netns delete" doesn't
+// support patterns, so this runs a short shell for-loop.
+func DeleteNetNamespaces(ctx context.Context) error {
+	return runCommand(ctx, "sh", "-c",
+		"for ns in $(ip netns list | grep ^bb-executor | cut -d \" \" -f 1); do ip netns delete $ns; done")
```
bduffany, line 68: what does the `| cut` part do? for me the output is the same with or without it (wonder if we have different versions of `ip`)

```
$ ip netns list | grep ^bb-executor
bb-executor-1
bb-executor-2
$ ip netns list | grep ^bb-executor  | cut -d " " -f 1
bb-executor-1
bb-executor-2
```

bduffany (approved):

PR #3044, opened 2022-12-09
Do not rely on replica usage in driver

tylerwilliams:


enterprise/server/raft/driver/driver.go
```diff
@@ -108,41 +95,45 @@ func variance(samples []float64, mean float64) float64 {
 }
 
 type clusterMap struct {
-	mu              *sync.RWMutex
-	nodeDescriptors map[string]*rfpb.NodeDescriptor
-	nodeReplicas    map[string]replicaSet
-	observations    map[replicaStruct]observation
+	mu *sync.RWMutex
+
+	nodeReplicas map[string]replicaSet
```
vadimberezniker, line 100: seems unused

vadimberezniker (approved):

enterprise/server/raft/driver/driver.go
tylerwilliams: whoops forgot to push latest

PR #3045, opened 2022-12-09
Add an approximate LRU implementation.

vadimberezniker:
This is largely based on the code in the Pebble cache with some minor differences.

I plan to switch over the Pebble code to this implementation after a bit more testing.

I'm also removing the older 'approximatelru' implementation that keeps state on all the keys. It's not being used and is unlikely to be used any time soon.

---

**Version bump**: None


tylerwilliams (approved):
This looks really nice

PR #3048, opened 2022-12-09
Add debug logging if pebble metadata size does not match size in cached digest

maggie-lou:
This logging will have some false positive rate, because on dev compression is on and in those cases we'd expect the sizes to be different.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -898,7 +898,7 @@ func (p *PebbleCache) Metadata(ctx context.Context, r *resource.ResourceName) (*
 	}
 
 	if md.GetStoredSizeBytes() != md.GetFileRecord().GetDigest().GetSizeBytes() {
-		log.Debugf("Pebble metadata size mismatch: %v", md)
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3048** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3048" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3048?utm_source=stack-comment).

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -897,6 +897,10 @@ func (p *PebbleCache) Metadata(ctx context.Context, r *resource.ResourceName) (*
 		return nil, err
 	}
 
+	if md.GetStoredSizeBytes() != md.GetFileRecord().GetDigest().GetSizeBytes() {
+		log.Debugf("Pebble metadata size mismatch: %v", md)
```
vadimberezniker, line 901: We don't enable Debug level logs in prod so let's make this an Info level log.

BTW, try to use the log.Ctx* functions if there's a context already available. It makes it possible for us to enrich our logs with more metadata.

vadimberezniker (approved):

PR #3049, opened 2022-12-10
Add debug logging if pebble metadata size does not match size in cached digest - Move to lookups and writes

maggie-lou:
Realized we'd get more hits on this log if we moved it to metadata lookups and writes, because the cache Metadata function doesn't get as much traffic

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1359,12 +1359,6 @@ func (p *PebbleCache) Writer(ctx context.Context, r *resource.ResourceName) (int
 			metrics.DiskCacheAddedFileSizeBytes.Observe(float64(bytesWritten))
 		}
 
-		fr := md.GetFileRecord()
-		if fr.GetCompressor() == repb.Compressor_IDENTITY &&
-			md.GetStoredSizeBytes() != fr.GetDigest().GetSizeBytes() {
-			log.CtxInfof(ctx, "Pebble write metadata size mismatch: %v", md)
-		}
-
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3049** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3049" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3049?utm_source=stack-comment).

vadimberezniker:
Should we do this check only for uncompressed data? 
This would cause a lot of noise in dev.

maggie-lou:
Actually realized the most useful will be to log when we're writing the corrupted data to the cache. Inspecting the data on reads likely won't give us any helpful information: https://github.com/buildbuddy-io/buildbuddy/pull/3054

vadimberezniker:
> Actually realized the most useful will be to log when we're writing the corrupted data to the cache. Inspecting the data on reads likely won't give us any helpful information: #3054

It would be useful to know if there's already data in the cache that's corrupted. It might not get written again if it doesn't get evicted.

maggie-lou:
Can't we assume there is corrupted data in the cache given the errors from last week? If you filter the customer's failed builds from the day of the deploy, you can see there are lots of different digests that were triggering the errors

vadimberezniker:
> Can't we assume there is corrupted data in the cache given the errors from last week? If you filter the customer's failed builds from the day of the deploy, you can see there are lots of different digests that were triggering the errors

So far we've only seen evidence of this in the "remote asset" fetcher feature which they've since disabled.

maggie-lou:
> > Can't we assume there is corrupted data in the cache given the errors from last week? If you filter the customer's failed builds from the day of the deploy, you can see there are lots of different digests that were triggering the errors
> 
> So far we've only seen evidence of this in the "remote asset" fetcher feature which they've since disabled.

Okay - just reopened this PR

vadimberezniker (approved):

PR #3050, opened 2022-12-12
Fix executor startup

bduffany:
- Use `sudo -A` to make sure the executor never gets stuck prompting for a sudo password.
- Instead of using `sudo sh -c '<script that invokes ip command>'` for ip netns cleanup, run `sudo ip` commands directly and process the intermediate results with Go, since `enable_local_firecracker.sh` only sets up passwordless sudo for the `ip` command, not the `sh` command.
- If the networking cleanup fails, log a debug message instead of `exit(1)`, since we don't want this to be a fatal err on Mac or if someone hasn't run `enable_local_firecracker.sh` on Linux.

---

**Version bump**: Patch

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -63,9 +63,10 @@ func namespace(netNamespace string, args ...string) []string {
 }
 
 // Deletes all of the executor net namespaces. These can be left behind if the
-// executor doesn't exit gracefully. Unfortunately, "ip netns delete" doesn't
-// support patterns, so this runs a short shell for-loop.
+// executor doesn't exit gracefully.
 func DeleteNetNamespaces(ctx context.Context) error {
+	// "ip netns delete" doesn't support patterns, so we list all
+	// namespaces then delete the ones that match the bb executor pattern.
 	b, err := sudoCommand(ctx, "ip", "netns", "list")
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


server/util/networking/networking.go
```diff
@@ -64,8 +66,23 @@ func namespace(netNamespace string, args ...string) []string {
...
+	}
+	output := strings.TrimSpace(string(b))
+	if len(output) == 0 {
+		return nil
+	}
+	for _, ns := range strings.Split(output, "\n") {
+		if !strings.HasPrefix(ns, netNamespacePrefix) {
+			continue
+		}
+		if _, err := sudoCommand(ctx, "ip", "netns", "delete", ns); err != nil {
```
iain-macdonald, line 81: Sometimes the output of "ip netns list" can contain spaces, e.g.:

`[07:29:49] iain@lunchbox:~ > ip netns ls
bb-executor-213
bb-executor-21
bb-executor-2
3fe4313e-eb76-4b6d-9d61-53caf12b87e6 (id: 344)
2ab15e85-d1c3-47bc-ad40-74e2941157a4 (id: 332)`

Can you split ns on " " and just delete the first term from the output?

iain-macdonald (changes requested):

server/util/networking/networking.go
iain-macdonald: Argh, the formatting got all messed up:

[07:29:49] iain@lunchbox:~ > ip netns ls
bb-executor-213
bb-executor-21
bb-executor-2
3fe4313e-eb76-4b6d-9d61-53caf12b87e6 (id: 344)
2ab15e85-d1c3-47bc-ad40-74e2941157a4 (id: 332)

iain-macdonald (approved):

server/util/networking/networking.go
```diff
@@ -74,15 +74,24 @@ func DeleteNetNamespaces(ctx context.Context) error {
 	if len(output) == 0 {
 		return nil
 	}
+	var lastErr error
 	for _, ns := range strings.Split(output, "\n") {
+		// Sometimes the output contains spaces, like
+		//     bb-executor-1
+		//     bb-executor-2 3fe4313e-eb76-4b6d-9d61-53caf12b87e6 (id: 344) 2ab15e85-d1c3-47bc-ad40-74e2941157a4 (id: 332)
```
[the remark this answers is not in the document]
bduffany: Done

PR #3052, opened 2022-12-12
Implement Raft cache eviction.

vadimberezniker:
`usageTracker` in the store collects usage data from local replicas and remote nodes. Local usage data is pushed by replicas. Remote usage data is gossiped and processed by the usage manager.

Each store does its own eviction for locally owned replicas. The amount of data evicted is proportional to the percent of data owned by the store.

---

**Version bump**: None


enterprise/server/raft/replica/replica.go
```diff
@@ -63,7 +64,7 @@ var (
 type IStore interface {
 	AddRange(rd *rfpb.RangeDescriptor, r *Replica)
 	RemoveRange(rd *rfpb.RangeDescriptor, r *Replica)
-	NotifyUsage(ru *rfpb.ReplicaUsage)
+	NotifyUsage(ru *rfpb.ReplicaUsage, pu []*rfpb.PartitionMetadata)
```
tylerwilliams, line 67: could we move []*rfpb.PartitionMetadata inside of ReplicaUsage and leave this API as is?

enterprise/server/raft/store/store.go
```diff
@@ -61,13 +64,402 @@ const (
...
+				up.SizeBytes += ru.GetSizeBytes()
+				up.TotalCount += ru.GetTotalCount()
+			}
+		}
+		nu.PartitionUsage = append(nu.PartitionUsage, up)
+	}
+	return nu
+}
+
+func (ut *usageTracker) broadcastLoop() {
```
tylerwilliams, line 434: this looping and gossiping gets really noisy with large numbers of nodes. Could we drive broadcasts based on a change in usage data greater than a certain size?

vadimberezniker: The looping is cheap, but I think reducing the frequency of gossiping is a fair point. I can follow that up in a separate PR, it would be good to get this change merged.

enterprise/server/raft/keys/keys.go
```diff
@@ -31,3 +37,7 @@ func IsLocalKey(key Key) bool {
 	}
 	return key[0] == '\x01'
 }
+
+func Range(key []byte) ([]byte, []byte) {
```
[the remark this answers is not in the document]
vadimberezniker: Done

enterprise/server/raft/replica/replica.go
tylerwilliams: Still in favor of moving this into ReplicaUsage -- because I think we can delete NodeUsage. I'm happy to send a CL now to do so. (StoreUsage is the new NodeUsage, but this data does not need to be in StoreUsage)

enterprise/server/raft/store/store.go
tylerwilliams: SGTM, but I do think this is relatively important to do soon.

enterprise/server/raft/replica/replica.go
vadimberezniker: Moved.

enterprise/server/raft/store/store.go
vadimberezniker: It'll be in the next PR.

tylerwilliams (approved):

PR #3055, opened 2022-12-12
Fix the netns provided to Firecracker, it wasn't updated to include the new 'bb-executor' prefix

iain-macdonald:
**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -586,7 +586,7 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
 
 	var netNS string
 	if c.constants.EnableNetworking {
-		netNS = "/var/run/netns/" + c.id
+		netNS = "/var/run/netns/bb-executor-" + c.id
 	}
 
 	var stdout io.Writer = c.vmLog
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -586,7 +586,7 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
 
 	var netNS string
 	if c.constants.EnableNetworking {
-		netNS = "/var/run/netns/" + c.id
+		netNS = "/var/run/netns/bb-executor-" + c.id
```
bduffany, line 589: wdyt about putting this in a func like `networking.NamespacePath(c.id)` so that the prefixing is transparent here?

iain-macdonald: I thought you might ask for that :-P  Done

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -586,7 +586,7 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context, workspaceDirOve
 
 	var netNS string
 	if c.constants.EnableNetworking {
-		netNS = "/var/run/netns/" + c.id
+		netNS = "/var/run/netns/" + networking.NamespacePath(c.id)
```
bduffany, line 589: nit (sorry, I should've been more clear) -- I was thinking `NamespacePath` would return `"/var/run/netns/bb-prefix-" + c.id` -- but this looks good too if you want to just rename it to `Namespace()` or `NamespaceName()`?

bduffany (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
iain-macdonald: Ohhh, that's a good idea -- added it and renamed this one to "NetNamespace".

bduffany (approved):

PR #3058, opened 2022-12-12
Add --executor.firecracker_kill_host_on_failure flag that kills the host executor process if there's a problem with Firecracker.

iain-macdonald:
**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -59,7 +59,7 @@ import (
 
 var firecrackerMountWorkspaceFile = flag.Bool("executor.firecracker_mount_workspace_file", false, "Enables mounting workspace filesystem to improve performance of copying action outputs.")
 var firecrackerCgroupVersion = flag.String("executor.firecracker_cgroup_version", "1", "Specifies the cgroup version for firecracker to use.")
-var firecrackerKillHostOnFailure = flag.Bool("executor.firecracker_kill_host_on_failure", false, "Kills the host executor process if any command orchestrating or running Firecracker fails. Useful for capturing failures preemptively. WARNING: using this option MAY leave the host machine in an unhealthy state on Firecracker failure; some post-hoc cleanup may be necessary.")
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -59,6 +59,7 @@ import (
 
 var firecrackerMountWorkspaceFile = flag.Bool("executor.firecracker_mount_workspace_file", false, "Enables mounting workspace filesystem to improve performance of copying action outputs.")
 var firecrackerCgroupVersion = flag.String("executor.firecracker_cgroup_version", "1", "Specifies the cgroup version for firecracker to use.")
+var firecrackerKillHostOnFailure = flag.Bool("executor.firecracker_kill_host_on_failure", false, "Kills the host executor process if any command orchestrating or running Firecracker fails. Useful for capturing failures preemptively. WARNING: using this option MAY leave the host machine in an unhealthy state on Firecracker failure; some post-hoc cleanup may be necessary.")
```
tylerwilliams, line 62: how about calling this executor.die_on_firecracker_failure instead? (Since the host machine is not actually killed)

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -747,6 +748,10 @@ func (c *FirecrackerContainer) hotSwapWorkspace(ctx context.Context, execClient
 
 func nonCmdExit(err error) *interfaces.CommandResult {
 	log.Errorf("nonCmdExit returning error: %s", err)
+	if *dieOnFirecrackerFailure {
+		log.Errorf("--executor.die_on_firecracker_failure=true executor is dying now")
+		os.Exit(1)
```
tylerwilliams, line 753: you could instead call 

```log.Fatalf("dying on firecracker error: %s", err)```

iain-macdonald: Done

tylerwilliams (approved):

PR #3063, opened 2022-12-13
CLI: Fix missing scorecard data

bduffany:
* Add gRPC metadata to queued ByteStream Write requests to ensure that request metadata is forwarded to the remote.
* Add an integration test that runs `bb` against BuildBuddy and makes sure we can read the scorecard.
* Fix an issue that piping terminal output to `/dev/null` would crash the CLI. This was discovered through the integration test, since the default stdout/stderr behavior for `exec.Command` is to pipe to /dev/null.
* Update test bazel version from 5.3.0 to 5.3.2 since it is needed to fix an issue that `unix://` backend targets aren't handled properly.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1861


server/testutil/testbazel/test_rules.bzl
```diff
@@ -0,0 +1,18 @@
+def _run_shell_impl(ctx):
```
tylerwilliams, line 1: maybe we save this for a later day? I'm a little shy about customizing bazel too hard because things will break / change / require upkeep as bazel versions change.

tylerwilliams (changes requested):
just one thing, otherwise lg

server/testutil/testbazel/test_rules.bzl
bduffany: Sounds good - changed back to sh_binary

tylerwilliams (approved):

PR #3064, opened 2022-12-13
Don't create ip routes that already exist.

iain-macdonald:
This isn't really an issue as we handle the failure fine, but when running with `--executor.die_on_firecracker_failure` it makes the executor die when it would otherwise be fine.

**Version bump**: Patch

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -210,6 +210,24 @@ func DeleteRuleIfSecondaryNetworkEnabled(ctx context.Context, vmIdx int) error {
 	return runCommand(ctx, "ip", "rule", "del", "from", getCloneIP(vmIdx))
 }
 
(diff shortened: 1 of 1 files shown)


server/util/networking/networking.go
```diff
@@ -210,6 +210,24 @@ func DeleteRuleIfSecondaryNetworkEnabled(ctx context.Context, vmIdx int) error {
...
+	if err != nil {
+		return false, err
+	}
+	output := strings.TrimSpace(string(b))
+	if len(output) == 0 {
+		return false, nil
+	}
+	providedRoute := source + " via " + gateway
+	for _, route := range strings.Split(output, "\n") {
+		if strings.HasPrefix(route, providedRoute) {
```
bduffany, line 224: nit: can you add a comment with an example output line, so that it's more clear why we need `HasPrefix` vs `==`?

bduffany (approved):

PR #3065, opened 2022-12-13
CLI: Rewrite bytestream:// URIs in BES to remote cache target

bduffany:
File URIs in BEP contained a string like `bytestream://///tmp/sidecar-12345.sock/PATH`, and so they could not be fetched via the BB UI (which doesn't know about the sidecar socket). This caused the timing profile to be missing.

This PR fixes that by rewriting all File URIs so that they reference the remote cache target, instead of the sidecar socket.

Will add an integration test in a follow-up PR.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1860


siggisim:
Could this be done with this flag instead? https://bazel.build/reference/command-line-reference#flag--remote_bytestream_uri_prefix

bduffany:
> Could this be done with this flag instead? https://bazel.build/reference/command-line-reference#flag--remote_bytestream_uri_prefix

Yes! that's way simpler. Closing this PR and opening a new one

PR #3066, opened 2022-12-13
Flip executor.firecracker_cgroup_version to 2 everywhere.

iain-macdonald:
**Version bump**: Patch


tylerwilliams (approved):
LGTM if it looked good on dev

PR #3073, opened 2022-12-13
Put --wait in all iptables commands in networking.go

iain-macdonald:
When running the executor with firecracker locally, concurrent `iptables` commands can conflict with one another, returning errors like:
`2022/12/13 14:30:43.906 FTL dying on firecracker error: rpc error: code = Internal desc = Error running "/usr/bin/sudo -A ip netns exec bb-executor-f34cfb00-e3ec-43f3-89ef-6ed99154d4e6 iptables -t nat -A POSTROUTING -o veth0mmV9K -s 192.168.241.2 -j SNAT --to 192.168.0.99" Another app is currently holding the xtables lock. Perhaps you want to use the -w option?
: exit status 4`

This doesn't happen (or at least not as frequently) with --jobs=1 which makes me more confident in my diagnosis. I could have fixed this by protecting these calls with a mutex, but this seems simpler for now.

**Version bump**: Patch

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -189,7 +189,7 @@ func attachAddressToVeth(ctx context.Context, netNamespace, ipAddr, vethName str
 }
 
 func removeForwardAcceptRule(ctx context.Context, vethName, defaultDevice string) error {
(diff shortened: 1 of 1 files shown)


vadimberezniker (approved):
Interesting that you're running into this!

PR #3087, opened 2022-12-15
Write Executions to clickhouse even after invocation is complete

luluz66:
1. In build_event_handler, when an invocation is complete, write the invocation
into redis.
2. In excution_server, when an execution is complete, check invocation
   in redis. If there is an entry in redis, directly write the execution
   to clickhouse; otherwise, add the execution to redis.


enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
```diff
@@ -3,17 +3,22 @@ package redis_execution_collector
...
 	"google.golang.org/protobuf/proto"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	sipb "github.com/buildbuddy-io/buildbuddy/proto/stored_invocation"
 )
 
 const (
-	redisExecutionKeyPrefix = "exec"
+	redisExecutionKeyPrefix  = "exec"
+	redisInvocationKeyPrefix = "inv"
```
vadimberezniker, line 19: `inv` is not an obvious abbreviation for invocation so let's write out `invocation`

proto/stored_invocation.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package stored_invocation;
```
vadimberezniker, line 3: do we need a separate file? we already have a lot of custom messages in `invocation.proto`, could we add add this there?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -50,9 +50,11 @@ import (
 	inpb "github.com/buildbuddy-io/buildbuddy/proto/invocation"
 	pgpb "github.com/buildbuddy-io/buildbuddy/proto/pagination"
 	pepb "github.com/buildbuddy-io/buildbuddy/proto/publish_build_event"
+	sipb "github.com/buildbuddy-io/buildbuddy/proto/stored_invocation"
 	uidpb "github.com/buildbuddy-io/buildbuddy/proto/user_id"
 	api_common "github.com/buildbuddy-io/buildbuddy/server/api/common"
 	api_config "github.com/buildbuddy-io/buildbuddy/server/api/config"
+	olapdb_config "github.com/buildbuddy-io/buildbuddy/server/olapdb/config"
```
vadimberezniker, line 57: instead of aliasing, why don't we call the package `olapdbconfig`? 
i'm not a big fan of the proliferation of `config` packages.

vadimberezniker (changes requested):

proto/stored_invocation.proto
luluz66: invocation.proto is used by the client. My main consideration here is to not expose the stored invocation proto to the client.

enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
luluz66: Done.

enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
```diff
@@ -3,17 +3,22 @@ package redis_execution_collector
...
+
+	invExpiration = 24 * time.Hour
```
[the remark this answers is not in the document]
luluz66: Done.

server/build_event_protocol/build_event_handler/build_event_handler.go

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -313,9 +313,26 @@ func (s *ExecutionServer) recordExecution(ctx context.Context, executionID strin
 
 	for _, link := range links {
 		executionProto := execution.TableExecToProto(&executionPrimaryDB, link)
-		if err := s.env.GetExecutionCollector().Append(ctx, link.InvocationID, executionProto); err != nil {
-			log.CtxErrorf(ctx, "failed to append execution %q to invocation %q", executionID, link.InvocationID)
+		inv, err := s.env.GetExecutionCollector().GetInvocation(ctx, link.InvocationID)
+		if err != nil {
+			log.CtxErrorf(ctx, "failed to get invocation %q from ExecutionCollector", link.InvocationID)
```
vadimberezniker, line 318: log the error in the message

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -313,9 +313,26 @@ func (s *ExecutionServer) recordExecution(ctx context.Context, executionID strin
...
+			continue
+		}
+		if inv == nil {
+			// The invocation hasn't finished yet. Add the execution to ExecutionCollector, and flush it once
+			// the invocation is complete
+			if err := s.env.GetExecutionCollector().Append(ctx, link.InvocationID, executionProto); err != nil {
+				log.CtxErrorf(ctx, "failed to append execution %q to invocation %q", executionID, link.InvocationID)
```
vadimberezniker, line 325: log the error in the message

proto/stored_invocation.proto
vadimberezniker: You can leave it here if you feel strongly about it, but I don't see why it matters that the client happens to use the messages in invocation.proto. Adding the message there does not affect the client in any way.

vadimberezniker (approved):

PR #3089, opened 2022-12-15
Remove unhelpful compression metric

maggie-lou:
In prod, the majority of blobs with a bad compression ratio (increased in size when compressed) were between 10,000B and 100,000B. This metric was added to see if there was a clear size cutoff above which compression was helpful in the majority of cases so we could use it as our minimum compression size. That doesn't seem to hold true, so the metric isn't really useful.


maggie-lou:
* master
  * **PR #3089** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3089" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3089?utm_source=stack-comment).

bduffany:
For each of these buckets (10K, 100K, etc.) do we know the fraction of incompressible blobs? For example, can we say anything like "X% of blobs that are 10K or smaller are incompressible?"

bduffany (approved):

PR #3091, opened 2022-12-16
Add some prom metrics for cacheload errors

tylerwilliams:
This will let us run this thing with --keep_going  and then check errors in grafana or set prom alerts for them.


vadimberezniker (commented):
we need to run an http server to expose the metrics for scraping

tylerwilliams:
right you are

vadimberezniker (approved):

PR #3092, opened 2022-12-16
Support detecting cgroup version from system

iain-macdonald:
This PR:
- Detects the cgroup version from the system (by running `stat -fc %T /sys/fs/cgroup/` and searching for `cgroup2fs` or `tmpfs`) when `--executor.firecracker_cgroup_version` is an empty string. Otherwise, the value from `--executor.firecracker_cgroup_version` is used.
- Switches the default value of `--executor.firecracker_cgroup_version` to be an empty string, enabled the behavior above.

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -190,8 +190,7 @@ func getCgroupVersion() (string, error) {
 	}
 	b, err := exec.Command("stat", "-fc", "%T", "/sys/fs/cgroup/").Output()
 	if err != nil {
-		log.Debugf("Error determining system cgroup version: %s", err.Error())
-		return "", err
+		return "", status.UnavailableErrorf("Error determining system cgroup version: %v", err)
 	}
 	v := strings.TrimSpace(string(b))
 	if v == "cgroup2fs" {
@@ -199,7 +198,6 @@ func getCgroupVersion() (string, error) {
 	} else if v == "tmpfs" {
 		return "1", nil
 	} else {
-		log.Debugf("Error determining system cgroup version. System-reported version: %s", v)
 		return "", status.InternalErrorf("No cgroup version found (system reported %s)", v)
 	}
 }
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -176,6 +176,30 @@ func openFile(ctx context.Context, env environment.Env, fileName string) (io.Rea
 	return env.GetFileResolver().Open(fileName)
 }
 
+// Returns the cgroup version available on this machine, which is returned from
+//
+//	stat -fc %T /sys/fs/cgroup/
+//
+// and should be either 'cgroup2fs' (version 2) or 'tmpfs' (version 1)
+//
+// More info here: https://kubernetes.io/docs/concepts/architecture/cgroups/
+func getCgroupVersion() string {
+	b, err := exec.Command("stat", "-fc", "%T", "/sys/fs/cgroup/").Output()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -176,6 +176,30 @@ func openFile(ctx context.Context, env environment.Env, fileName string) (io.Rea
...
+//	stat -fc %T /sys/fs/cgroup/
+//
+// and should be either 'cgroup2fs' (version 2) or 'tmpfs' (version 1)
+//
+// More info here: https://kubernetes.io/docs/concepts/architecture/cgroups/
+func getCgroupVersion() string {
+	b, err := exec.Command("stat", "-fc", "%T", "/sys/fs/cgroup/").Output()
+	if err != nil {
+		log.Debugf("Error determining system cgroup version. Falling back to --executor.firecracker_cgroup_version (which is \"%s\"). Error: %s", *firecrackerCgroupVersion, err.Error())
+		return *firecrackerCgroupVersion
```
vadimberezniker, line 190: Instead of falling back to this flag, I think it would be better to change the semantics of the flag so that an empty value (new default) means auto detect and any other value is used explicitly. If the flag value is "" and we cannot auto-detect then we return error.

It would also be nice if we could detect this earlier rather than when a container is being created. In the podman implementation we have an abstraction where we create a provider on startup and the provider in turns creates new container instances. This abstraction gives us a place to do startup configuration like this and be able to return error conditions.

vadimberezniker (changes requested):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
iain-macdonald: Yeah, that all sounds good. I added a TODO for the second part because it's a little bit more involved, LMK what you think.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -176,6 +176,34 @@ func openFile(ctx context.Context, env environment.Env, fileName string) (io.Rea
...
+//
+// More info here: https://kubernetes.io/docs/concepts/architecture/cgroups/
+// TODO(iain): preemptively get this in a provider a la Provider in podman.go.
+func getCgroupVersion() (string, error) {
+	if *firecrackerCgroupVersion != "" {
+		return *firecrackerCgroupVersion, nil
+	}
+	b, err := exec.Command("stat", "-fc", "%T", "/sys/fs/cgroup/").Output()
+	if err != nil {
+		log.Debugf("Error determining system cgroup version: %s", err.Error())
```
vadimberezniker, line 193: the caller is likely already going to report/log the error so instead of a separate log statement it would be better to enrich the returned error message with more information, something like

`return "", status.UnavailableErrorf("error determining...: %s", err)`

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -176,6 +176,34 @@ func openFile(ctx context.Context, env environment.Env, fileName string) (io.Rea
...
+		return "", err
+	}
+	v := strings.TrimSpace(string(b))
+	if v == "cgroup2fs" {
+		return "2", nil
+	} else if v == "tmpfs" {
+		return "1", nil
+	} else {
+		log.Debugf("Error determining system cgroup version. System-reported version: %s", v)
```
vadimberezniker, line 202: this log statement is redundant since the caller is going to report the error and it has the same information as this log

vadimberezniker (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
iain-macdonald: Done

PR #3101, opened 2022-12-16
Support directly downloading zipped test outputs.

jdhollen:
- add an RPC to fetch the zip file manifest
- add a URL param to fetch a single file from an archive
- split rendering for test outputs to be per-run
- render zip manifests below the file that contains them.


jdhollen:
Alright, this is a little big, happy to split it up or do whatever else.  Also, figured I'd solicit your opinions on test coverage before I went super deep, since some of these files currently don't have coverage.

jdhollen:
Siggi, bouncing this to you because GIthub **"suggested"** you but more so that you can tell me who to send it to than anything.

siggisim:
> Siggi, bouncing this to you because GIthub **"suggested"** you but more so that you can tell me who to send it to than anything.

Conceptually this all looks reasonable to me, I'll let @bduffany take a closer look - particularly at the golang side of things

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1036,13 +1051,38 @@ func (s *BuildBuddyServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if len(archiveReference) > 0 {
+		b, err := base64.StdEncoding.DecodeString(archiveReference)
+		if err != nil {
+			log.Warningf("Error downloading file: %s", err.Error())
+			http.Error(w, "File not found", http.StatusBadRequest)
+			return
+		}
+		entry := &arpb.ManifestEntry{}
+		if err := proto.Unmarshal(b, entry); err != nil {
+			log.Warningf("Error downloading file: %s", err.Error())
```
bduffany, line 1067: nit:

```suggestion
			log.Warningf("Failed to unmarshal ManifestEntry: %s", err.Error())
```

bduffany (approved):

bduffany (commented):
(woops, this didn't make it through in my review)

It would be great to have an integration test case in `buildbuddy_server_test.go`

proto/archive.proto
```diff
@@ -0,0 +1,54 @@
+syntax = "proto3";
+
+import "proto/context.proto";
+
+package archive;
```
tylerwilliams, line 5: wanna call this zip or are there other archive formats you think would make sense here?

proto/buildbuddy_service.proto
```diff
@@ -40,6 +41,10 @@ service BuildBuddyService {
   rpc GetInvocationOwner(invocation.GetInvocationOwnerRequest)
       returns (invocation.GetInvocationOwnerResponse);
 
+  // Archive API
+  rpc GetArchiveManifest(archive.GetArchiveManifestRequest)
```
tylerwilliams, line 45: ubernit: this CL looks p clean but in the future it'd be nice to send the proto defs first then the service impl after

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -51,6 +53,7 @@ import (
 	requestcontext "github.com/buildbuddy-io/buildbuddy/server/util/request_context"
 	gcodes "google.golang.org/grpc/codes"
 	gstatus "google.golang.org/grpc/status"
+	proto "google.golang.org/protobuf/proto"
```
tylerwilliams, line 56: I don't think you need the name alias here, you can just remove it and move this import line up to the stanza before the aliased imports

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1036,13 +1051,38 @@ func (s *BuildBuddyServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	// TODO(siggisim): Figure out why this JWT is overriding authority auth and remove.
+	ctx := context.WithValue(r.Context(), "x-buildbuddy-jwt", nil)
+
+	var archiveReference = params.Get("z")
+	if len(archiveReference) > 0 {
+		b, err := base64.StdEncoding.DecodeString(archiveReference)
+		if err != nil {
+			log.Warningf("Error downloading file: %s", err.Error())
```
tylerwilliams, line 1061: we are often lazy and omit the ".Error()" because it will work fine without that

tylerwilliams: (it's inconsistent, and this is older code so probably wasn't as common then)

server/bytestream/BUILD
```diff
@@ -2,10 +2,14 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
 go_library(
     name = "bytestream",
-    srcs = ["bytestream.go"],
+    srcs = [
+        "bytestream.go",
+        "zip.go",
```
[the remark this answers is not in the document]
tylerwilliams: +1 to this suggestion

server/bytestream/bytestream.go
```diff
@@ -14,11 +15,140 @@ import (
...
+	footer, err := io.ReadAll(reader)
+	if err != nil {
+		return nil, err
+	}
+
+	// Find and parse the End of Central Directory header or fail.
+	eocd, err := readDirectoryEnd(footer, r.GetDigest().GetSizeBytes())
+	if err != nil {
+		return nil, err
+	}
```
tylerwilliams: instead of using a pipe and a goroutine, you can use bytes.Buffer (which basically just implements io.Reader and io.Writer interfaces on a []byte) and stream your bytes into the buffer then read from it.

Something like:
```
var buf bytes.Buffer
err = StreamBytestreamFileChunk(ctx, env, url, offset, r.GetDigest().GetSizeBytes()-offset, func(data []byte) {
  buf.Write(data)
})
if err != nil {
  return nil, err
}
eocd, err := readDirectoryEnd(buf.Bytes(), r.GetDigest().GetSizeBytes())
```

server/bytestream/BUILD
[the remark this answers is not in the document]
jdhollen: Sure, branched this out and cleaned it up a bit so that there isn't as much bleeding over between the two packages.  Feels like this could still be a bit better, open to suggestions.

server/bytestream/bytestream.go
```diff
@@ -14,11 +15,140 @@ import (
...
+
+	cdStart := eocd.directoryOffset - offset
+	cdEnd := cdStart + eocd.directorySize
+
+	out := &arpb.ArchiveManifest{}
+	entries, err := readDirectoryHeader(footer[cdStart:cdEnd], eocd)
+	out.Entry = entries
+	return out, nil
```
tylerwilliams, line 62: you could rewrite this (and check this error) like:

```
entries, err := readDirectoryHeader(footer[cdStart:cdEnd], eocd)
if err != nil {
  return nil, err
}
return &arpb.ArchiveManifest{Entry: entries}, nil
```

server/bytestream/bytestream.go
```diff
@@ -14,11 +15,140 @@ import (
...
+		err := StreamBytestreamFileChunk(ctx, env, url, entry.GetHeaderOffset(), fileHeaderLen, func(data []byte) {
+			writer.Write(data)
+		})
+		writer.CloseWithError(err)
+	}()
+
+	header := make([]byte, fileHeaderLen)
+	if _, err := io.ReadFull(reader, header); err != nil {
+		return -1, err
+	}
```
tylerwilliams, line 76: same thing with the bytes.Buffer should work fine here

server/bytestream/zip.go
```diff
@@ -0,0 +1,190 @@
...
+	directoryEndLen          = 22
+	directoryHeaderLen       = 46
+	fileHeaderLen            = 30
+)
+
+var (
+	ErrFormat    = errors.New("zip: not a valid zip file")
+	ErrAlgorithm = errors.New("zip: unsupported compression algorithm")
+	ErrChecksum  = errors.New("zip: checksum error")
+	ErrZip64     = errors.New("zip: zip64 not supported")
```
tylerwilliams: we don't usually use errors.New because they don't go over the wire very well. You can grab an appropriate error type from server/util/status and then it will be a proper gRPC error if it ever goes across the wire (and we'll be able to parse it etc).

Also not 100% sure you want these to be exported from this package but maybe?

server/bytestream/zip.go
```diff
@@ -0,0 +1,190 @@
...
+		return arpb.ManifestEntry_COMPRESSION_TYPE_NONE
+	case 8:
+		return arpb.ManifestEntry_COMPRESSION_TYPE_FLATE
+	default:
+		return arpb.ManifestEntry_COMPRESSION_TYPE_UNKNOWN
+	}
+}
+
+func readDirectoryEnd(input []byte, trueSize int64) (dir *directoryEnd, err error) {
+
```
tylerwilliams: nit: rm newline

server/bytestream/zip.go
```diff
@@ -0,0 +1,190 @@
...
+	b := readBuf(input[10:]) // skip signature, disk fields
+	d := &directoryEnd{
+		directoryRecords: int64(b.uint16()),
+		directorySize:    int64(b.uint32()),
+		directoryOffset:  int64(b.uint32()),
+		commentLen:       b.uint16(),
+	}
+	l := int(d.commentLen)
+	if l > len(b) {
+		return nil, errors.New("zip: invalid comment length")
```
tylerwilliams: nit: same as above re error type

tylerwilliams (changes requested):
ok whew LMK which parts of zip.go I should review? Also if it makes sense to wrap those zip.go errors in a different place that's fine with me, just don't want to return errors.New stuff to an RPC

proto/archive.proto
jdhollen: Alright, both of you commented on this now, so I updated it all.

proto/buildbuddy_service.proto
jdhollen: Sorry about that, got a little excited with all of this, will split things up reasonably next time.

server/buildbuddy_server/buildbuddy_server.go
jdhollen: Fixed

server/buildbuddy_server/buildbuddy_server.go
jdhollen: Fixed this and the one below, too.

server/bytestream/bytestream.go
jdhollen: Thanks for this, it's a lot less ugly for the cases where we know the (small) amount of data we need.

I updated StreamByteStreamFile to take a Writer, so that meant updating the little api_server call to provide a writer, I'm not sure if I've quite done it in the "idiomatic" way, so it'd be good if you could take a peek at that.

server/bytestream/bytestream.go
jdhollen: Gah, sorry, dunno how i missed the check here.

server/bytestream/zip.go
jdhollen: Yeah, this were public in the original package, but I made them private now and updated to hopefully reasonable error types.

I have a few other IO-related errors (theoretical-but-would-be-weird error outputs from writes to Writers, etc.) that are new and look like they might leak out.  Should I be wrapping those up as well?

enterprise/server/api/api_server.go
```diff
@@ -319,6 +323,13 @@ func (s *APIServer) GetLog(ctx context.Context, req *apipb.GetLogRequest) (*apip
 	}, nil
 }
 
+func (gfs *GetFileServer) Write(data []byte) (int, error) {
```
tylerwilliams, line 326: ubernit: you could call the struct getFileWriter (it doesn't need to be exported) and you can just define it and the Write() method on it right above GetFile since that's the only place it's likely to be used.

proto/zip.proto
```diff
@@ -0,0 +1,54 @@
+syntax = "proto3";
+
+import "proto/context.proto";
+
+package zip;
+
+message ZipManifestEntry {
```
tylerwilliams, line 7: haha since this is in package zip you can lose the Zip prefix in the "object" types (sorry brandon and I conspired to lead you here into zip.ZipVille)

server/bytestream/bytestream.go
```diff
@@ -12,13 +13,101 @@ import (
...
+	}()
+
+	// Validate that the dynamic portion of the file header makes sense.
+	extras := make([]byte, dynamicHeaderBytes)
+	if _, err := io.ReadFull(reader, extras); err != nil {
+		return err
+	}
+	if err := ziputil.ValidateLocalFileNameAndExtras(extras, entry); err != nil {
+		return err
+	}
```
tylerwilliams, line 96: if either of these return an error do you want to call reader.CloseWithError() so the goroutine you started dies?

server/bytestream/bytestream.go
```diff
@@ -33,23 +122,23 @@ func StreamBytestreamFile(ctx context.Context, env environment.Env, url *url.URL
...
 	if err != nil {
-		err = streamFromUrl(ctx, url, false, callback)
+		err = streamFromUrl(ctx, url, false, offset, limit, writer)
 	}
 
 	return err
 }
 
-func streamFromUrl(ctx context.Context, url *url.URL, grpcs bool, callback func([]byte)) error {
+func streamFromUrl(ctx context.Context, url *url.URL, grpcs bool, offset int64, limit int64, writer io.Writer) error {
```
tylerwilliams, line 141: the writer change is nice :)

tylerwilliams (changes requested):
few minor nits, otherwise lg

server/bytestream/bytestream.go
tylerwilliams: I missed the defer reader.Close above -- I think that should be sufficient

server/bytestream/bytestream.go
jdhollen: $$$

server/bytestream/bytestream.go
jdhollen: i have `defer reader.Close()` up where the pipe is opened--once the reader is closed, the writer will get back an `ErrPipeClosed` for subsequent writes, so this should all be good to go already.

proto/zip.proto
jdhollen: updated--i left Zip in the service name and by extension in the request/response types just to match our rpc naming scheme.

enterprise/server/api/api_server.go
jdhollen: thanks, still getting used to conventions.

tylerwilliams (approved):

PR #3106, opened 2022-12-19
Performance: Store claims in context to avoid parsing

bduffany:
For API requests, we create a `Claims` based on the authenticated group and then mint a JWT from that, and store only the JWT in the context. Later, when we retrieve the claims (in `authenticatedUser`), we have to parse the JWT to reconstitute the original claims.

This PR stores both the claims and the JWT in the context so that we can just retrieve the original claims instead of having to reconstitute it from the JWT. This should reduce app CPU usage by about 9% based on current profiling data.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -70,6 +70,10 @@ const (
 	// to/from the outgoing/incoming request contexts.
 	contextTokenStringKey = "x-buildbuddy-jwt"
 
+	// The key the Claims are stored under in the context.
+	// If unset, the JWT can be used to reconstitute the claims.
+	contextClaimsKey = "auth.claims"
+
 	// The key that the basicAuth object is stored under in the
 	// context.
 	contextBasicAuthKey = "basicauth.user"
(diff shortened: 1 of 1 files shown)


enterprise/server/auth/auth.go
```diff
@@ -705,6 +709,7 @@ func authContextFromClaims(ctx context.Context, claims *Claims, err error) conte
 		return AuthContextWithError(ctx, err)
 	}
 	ctx = context.WithValue(ctx, contextTokenStringKey, tokenString)
+	ctx = context.WithValue(ctx, contextClaimsKey, claims)
```
vadimberezniker, line 712: It would be nice if we could store a single value (struct) in the context that contains both the jwt and the claims information. That would prevent them from getting out of sync as the auth code evolves in the future. It would be a more invasive change so no need to do it in this PR, but I think it would be good to follow up on it or file an issue for the future.

vadimberezniker (approved):

enterprise/server/auth/auth.go
tylerwilliams: +1

tylerwilliams (approved):

tylerwilliams:
Also, nice find :)

PR #3107, opened 2022-12-19
Update smash tool to facilitate A/B tests

bduffany:
- Allow running a fixed number of requests, instead of running for a fixed duration.
- Allow setting arbitrary remote_headers, not just API key.

These changes allow comparing two different runs of the smash tool, both running a fixed number of requests, one that sets a remote header (as a "flag" that enables a particular optimization) and the other which doesn't, and then comparing them by looking at which one completed the requests in a shorter duration.

Example (assuming the code enables an optimization based on the header `enable-cool-optimization`):

```
bazel build -c opt enterprise/server/cmd/smash
hyperfine \
    --warmup 1 \
    --runs 10 \
    --parameter-list enabled '0,1' \
    'bazel-bin/enterprise/server/cmd/smash/smash_/smash \
        --remote_header=enable-cool-optimization={enabled} \
        --test_total_requests=1000000'
# hyperfine will report whether the optimization made things faster or slower.
```

---

**Version bump**: Patch


bduffany:
> What do you think about moving these changes to tools/cacheload.go instead? I'd like to deprecate and remove smash which is sort of inflexible (doesn't support streaming RPCs well for instance)

SGTM, will update this PR.

bduffany:
> > What do you think about moving these changes to tools/cacheload.go instead? I'd like to deprecate and remove smash which is sort of inflexible (doesn't support streaming RPCs well for instance)
> 
> SGTM, will update this PR.

Nevermind, it actually looks somewhat difficult since cacheload is doing both reading and writing, and the reads depend on the writes. Will revisit this the next time I need to do an A/B test

PR #3114, opened 2022-12-20
Performance: Optimize `bazel_request.GetInvocationID`

bduffany:
This should save ~1-3% on app CPU based on current profiling data, and also save ~0.9KB worth of allocations per request.

```
BenchmarkGetInvocationID-8              10000000               964.0 ns/op          1184 B/op         14 allocs/op
BenchmarkFastGetInvocationID-8          10000000               128.4 ns/op           320 B/op          3 allocs/op
```

---

**Version bump**: Patch

### DIFF
--- a/server/util/bazel_request/bazel_request.go
+++ b/server/util/bazel_request/bazel_request.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/encoding/protowire"
 	"google.golang.org/protobuf/proto"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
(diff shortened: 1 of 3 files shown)


server/util/bazel_request/bazel_request.go
```diff
@@ -20,25 +21,55 @@ var (
...
+	}
+	rmd := &repb.RequestMetadata{}
+	if err := proto.Unmarshal(b, rmd); err != nil {
+		return nil
 	}
-	return nil
+	return rmd
 }
 
 func GetInvocationID(ctx context.Context) string {
```
tylerwilliams, line 44: What if we pre-extracted the invocation ID in one of our interceptors and set it as a value on the context -- then this method could check for the presence of that value on the context before extracting grpc MD and unmarshaling?

Seems like a simpler solution than this proto stuff (all though I am into this lol)

bduffany: yep, exactly

tylerwilliams: +1 to your suggestion re internal RPCs

PR #3115, opened 2022-12-20
Gossip node partition usages when size delta is bigger than threshold.

vadimberezniker:
If node usage has not reached the threshold, force an update to be sent after 15 minutes to make the usage information reasonably up to date across the cache nodes.

Other changes:
 - Proactively refresh replica usages every 15 seconds instead of relying on periodic updates from replicas. This makes the usage information more accurate and there's no performance hit as replica Usage() is now a cheap operation.
 - When evicting data locally, speculatively update usage for remote nodes assuming they are evicting at a similar rate to avoid evicting more data from nodes that are getting more write requests.

---

**Version bump**: TODO

### DIFF
--- a/enterprise/server/raft/store/store.go
+++ b/enterprise/server/raft/store/store.go
@@ -70,17 +70,17 @@ const (
 	// considered to be full and eviction will kick in.
 	evictionCutoffThreshold = .90
 
-	// How often nodes wil check whether to gossip usage data if it is
+	// How often stores wil check whether to gossip usage data if it is
 	// sufficiently different from the last broadcast.
-	nodePartitionUsageCheckInterval = 15 * time.Second
+	storePartitionUsageCheckInterval = 15 * time.Second
 
-	// How often nodes can go without broadcasting usage information.
-	// Usage data will be gosipped after this time if no updated were triggered
+	// How often stores can go without broadcasting usage information.
+	// Usage data will be gossiped after this time if no updated were triggered
 	// based on data changes.
-	nodePartitionUsageMaxAge = 15 * time.Minute
+	storePartitionUsageMaxAge = 15 * time.Minute
 
-	// How old node partition usage data can be before we consider it invalid.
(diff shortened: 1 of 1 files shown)


enterprise/server/raft/store/store.go
```diff
@@ -69,16 +70,23 @@ const (
...
-	// How often nodes wil gossip about their partition usage.
-	nodePartitionUsageGossipInterval = 15 * time.Second
+	// How often nodes wil check whether to gossip usage data if it is
+	// sufficiently different from the last broadcast.
+	nodePartitionUsageCheckInterval = 15 * time.Second
+
+	// How often nodes can go without broadcasting usage information.
+	// Usage data will be gosipped after this time if no updated were triggered
+	// based on data changes.
+	nodePartitionUsageMaxAge = 15 * time.Minute
```
tylerwilliams, line 80: Can we s/nodes/stores/g in the comment and variable name? (just to better distinguish this from the old node usage stuff and clarify that it comes from the store)

enterprise/server/raft/store/store.go
```diff
@@ -291,19 +317,41 @@ func (ut *usageTracker) Statusz(ctx context.Context) string {
...
-		percentFull := float64(globalSizeBytes) / float64(p.MaxSizeBytes)
+		percentFull := (float64(globalSizeBytes) / float64(p.MaxSizeBytes)) * 100
 
 		buf += fmt.Sprintf("\t\tCapacity: %s / %s (%2.2f%% full)\n", units.BytesSize(float64(globalSizeBytes)), units.BytesSize(float64(p.MaxSizeBytes)), percentFull)
 		buf += "\t\tLocal Ranges:\n"
+
 		u.mu.Lock()
-		for rid, pu := range u.replicas {
+		// Show ranges in a consistent order so that they don't jump around when
+		// refreshing the statusz page.
```
tylerwilliams, line 327: yay

tylerwilliams (approved):

PR #3118, opened 2022-12-20
Clean up a bunch of 'x might be undefined' warnings, etc., in trends component.

jdhollen:
Added a little function to return a default proto when getting values from the map (values will still be zero).


enterprise/app/trends/trends.tsx
```diff
@@ -177,6 +178,14 @@ export default class TrendsComponent extends React.Component<Props, State> {
     return moment(date).format("MMM D");
   }
 
+  getStat(date: string): invocation.TrendStat {
+    return this.state.dateToStatMap.get(date) || new invocation.TrendStat();
+  }
```
bduffany, line 183: If you switch back to `ITrendStat` then you could write

```
  getStat(date: string): invocation.TrendStat {
    return this.state.dateToStatMap.get(date) || {};
  }
```

and I think the type checker should recognize `{}` as being assignable to `ITrendStat`, not 100% sure though. Same in the func below

bduffany (commented):
nice - just dropping https://github.com/buildbuddy-io/buildbuddy-internal/issues/936 to get GitHub to link this PR to that issue, since I think this PR makes some progress towards it.

jdhollen:
I have no clue what happened here, but my eslint config decided it doesn't care about this anymore.  rather than trying to enforce my personal eslint preferences on everyone right now, i'll save changes like this for when we actually have agreed on a common config for it.

PR #3121, opened 2022-12-20
Clean up a bunch of 'x might be undefined' warnings, etc., in trends component.

jdhollen:
Added a little function to return an empty object when getting values from the map (values will still be zero).


jdhollen:
Alright, sorry @bduffany, as soon as I closed https://github.com/buildbuddy-io/buildbuddy/pull/3118, i went to compile again and eslint got mad all over again.  I did address this comments you made over there, this is using the interfaces now.

As you said before, just dropping in a link to https://github.com/buildbuddy-io/buildbuddy-internal/issues/936

jdhollen:
Side note, if you totally hate this, we could also just cast with Number(...) and then we'd get the same NaN behavior for everything as we already have?

bduffany (approved):

PR #3122, opened 2022-12-20
remove CommandBuffer in execution collector

luluz66:
The CommandBuffer flushes every 250 ms by default. This can cause an
issue where execution server appends the execution is reids, but when build event
handler tries to fetch the executions from redis, the execution was still in the
buffer and not in redis.

Also, added some logging to help debugging.

### DIFF
--- a/enterprise/server/backends/redis_execution_collector/BUILD
+++ b/enterprise/server/backends/redis_execution_collector/BUILD
@@ -6,7 +6,6 @@ go_library(
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/redis_execution_collector",
     visibility = ["//visibility:public"],
     deps = [
-        "//enterprise/server/util/redisutil",
         "//proto:remote_execution_go_proto",
         "//proto:stored_invocation_go_proto",
         "//server/environment",
--- a/enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
+++ b/enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


bduffany:
The build event handler has a 500ms delay after the invocation is finalized and before it fetches from Redis, to account for the 250ms buffer flush interval. Is the delay not enough, or are there executions being added to the buffer after the invocation is finalized? (Or something else entirely?)

luluz66:
> The build event handler has a 500ms delay after the invocation is finalized and before it fetches from Redis, to account for the 250ms buffer flush interval. Is the problem that the 500ms delay is not enough, or is it that there are executions being added to the buffer after the invocation is finalized? (Or something else entirely?)

(1) Build event handler adds invocation to redis ([code](https://cs.github.com/buildbuddy-io/buildbuddy/blob/7c693450bbe6683c651f454a0b7d7ad0c6ff297a/server/build_event_protocol/build_event_handler/build_event_handler.go?q=build_event_handler.go#L319)) 
(2) Immediately after this, Build event handler tries to fetch executions with the invocation id, and flushed them to clickhouse. 

Execution Server checks whether the invocation is in redis ([code](https://cs.github.com/buildbuddy-io/buildbuddy/blob/354ce6e1d5fd31d7b9f42e8d0c004c127d59d6c2/enterprise/server/remote_execution/execution_server/execution_server.go?q=execution_server.go#L316)) When the invocation was not in redis, it will try to add the execution into redis. If the invocation check happens right before (1) and (2) happens before the buffer is flushed, the execution will not write to clickhouse.

bduffany (approved):
(Discussed offline; writing directly to redis is fine since the expected QPS should be relatively small, compared to the cache QPS which is what we needed the buffering for)

PR #3124, opened 2022-12-20
Add more migration logging to investigate eof error

maggie-lou:
---

**Version bump**: TODO

PR #3125, opened 2022-12-20
diffread

tylerwilliams:
- Log resource name and bytes read for migration read errors


tylerwilliams:
obviated by the superior https://github.com/buildbuddy-io/buildbuddy/pull/3124

PR #3128, opened 2022-12-21
Don't log on EOF errors when reading chunks in migration cache

maggie-lou:
Don't log on EOF errors when reading chunks, because the readers from different caches could be reading at different rates. Only log on Close() if the total number of bytes read differs

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -425,7 +425,7 @@ func (d *doubleReader) Read(p []byte) (n int, err error) {
 	// Don't log on EOF errors when reading chunks, because the readers from different caches
 	// could be reading at different rates. Only log on Close() if the total number of bytes read differs
 	if dstErr != nil && dstErr != srcErr && dstErr != io.EOF {
-		log.Warningf("Migration %v read err, src err: %s, dest err: %s", d.r, srcErr, dstErr)
+		log.Warningf("Migration %v read err, src err: %v, dest err: %s", d.r, srcErr, dstErr)
 	}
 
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3128** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3128" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3128?utm_source=stack-comment).

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -422,8 +422,10 @@ func (d *doubleReader) Read(p []byte) (n int, err error) {
 	d.bytesReadSrc += srcN
 
 	eg.Wait()
-	if dstErr != nil && dstErr != srcErr {
-		log.Warningf("Migration %v read err, src err: %s, dest err: %s, src read %d, dest read %d", d.r, srcErr, dstErr, d.bytesReadSrc, d.bytesReadDest)
+	// Don't log on EOF errors when reading chunks, because the readers from different caches
+	// could be reading at different rates. Only log on Close() if the total number of bytes read differs
+	if dstErr != nil && dstErr != srcErr && dstErr != io.EOF {
+		log.Warningf("Migration %v read err, src err: %s, dest err: %s", d.r, srcErr, dstErr)
```
vadimberezniker, line 428: If the error could be nil, you can use %v instead of %s to print the nil value properly.

vadimberezniker (approved):

enterprise/server/backends/migration_cache/migration_cache.go

PR #3132, opened 2022-12-21
CAS tool: Don't append trailing newline when stdout is not a terminal

bduffany:
---

**Version bump**: Patch


vadimberezniker (approved):
thanks!

PR #3136, opened 2022-12-21
Fix seeking in compressed blobs.

vadimberezniker:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -2009,7 +2009,7 @@ func (p *PebbleCache) readerForCompressionType(ctx context.Context, resource *re
 
 	if requestedCompression == cachedCompression {
 		if requestedCompression != repb.Compressor_IDENTITY && (uncompressedOffset != 0 || uncompressedLimit != 0) {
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1997,8 +1979,38 @@ func (r *compressionReader) Close() error {
...
+		if status.IsNotFoundError(err) || os.IsNotExist(err) {
+			p.handleMetadataMismatch(ctx, err, fileMetadataKey, fileMetadata)
+		}
+		return nil, err
+	}
+	sendAtimeUpdate(p.accesses, fileMetadataKey, fileMetadata, p.atimeUpdateThreshold, p.atimeBufferSize)
+
 	if requestedCompression == cachedCompression {
+		if requestedCompression != repb.Compressor_IDENTITY && (uncompressedOffset != 0 || uncompressedLimit != 0) {
+			return nil, status.FailedPreconditionError("non-zero offset or limit requested for pass through")
```
tylerwilliams, line 2012: how about "passthrough-compression does not support offset/limit"?

tylerwilliams (approved):

PR #3145, opened 2022-12-22
Support --prober_name in bazelrbe

iain-macdonald:
When set, this option creates and builds the probed target under '//name' instead of '//' which should make it easier to idenfy prober invocations. When the option is unset, the default behavior is the same as today.

**Version bump**: None
**Related issues**: Unblocks https://github.com/buildbuddy-io/buildbuddy-internal/issues/1905

### DIFF
--- a/tools/probers/bazelrbe/bazelrbe.go
+++ b/tools/probers/bazelrbe/bazelrbe.go
@@ -20,7 +20,7 @@ import (
 var (
 	bazelBinary = flag.String("bazel_binary", "", "Path to bazel binary")
 	bazelArgs   = flag.String("bazel_args", "", "Space separated list of args to pass to Bazel")
-	proberName  = flag.String("prober_name", "", "Short, human-readable name of this prober")
+	proberName  = flag.String("prober_name", "", "Short, human-readable name of this prober. This name must be a valid bazel package name (only '.', '@', '-', '_' and alphanumeric characters allowed).")
 
 	numTargets         = flag.Int("num_targets", 10, "Number targets to generate")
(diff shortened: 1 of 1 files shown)


tools/probers/bazelrbe/bazelrbe.go
```diff
@@ -20,6 +20,7 @@ import (
 var (
 	bazelBinary = flag.String("bazel_binary", "", "Path to bazel binary")
 	bazelArgs   = flag.String("bazel_args", "", "Space separated list of args to pass to Bazel")
+	proberName  = flag.String("prober_name", "", "Short, human-readable name of this prober")
```
vadimberezniker, line 23: also worth noting that it should be a valid bazel package name

tools/probers/bazelrbe/bazelrbe.go
```diff
@@ -63,6 +64,14 @@ func createWorkspace(dir string, numTargets, numInputsPerTarget, inputSizeBytes
 		return err
 	}
 
+	if *proberName != "" {
+		dir = dir + "/" + *proberName
+		err := os.Mkdir(dir, 0777)
```
vadimberezniker, line 69: doesn't matter much here but 0755 is typically a better default for directories

vadimberezniker (approved):

tools/probers/bazelrbe/bazelrbe.go
iain-macdonald: Done

tools/probers/bazelrbe/bazelrbe.go
iain-macdonald: Done

bduffany (approved):

PR #3147, opened 2022-12-22
Performance: Further reduce JWT parsing overhead

bduffany:
JWT parsing overhead is still around 6-7% in prod, even after https://github.com/buildbuddy-io/buildbuddy/pull/3106. This should reduce that overhead by a factor of ~100.

```
BenchmarkParseJWT_Current-8       146936             16517 ns/op            4440 B/op         61 allocs/op
BenchmarkParseJWT_Cache-8       11399373               105.3 ns/op            16 B/op          1 allocs/op
```

---

**Version bump**: Patch


tylerwilliams (commented):
I think the LRU may modify last accessed time on Get so you probably want a regular mutex instead of the RWMutex. Otherwise LGTM

tylerwilliams (approved):

PR #3151, opened 2022-12-22
Performance: Optimize API key redaction

bduffany:
(Credit to @vadimberezniker for suggesting this optimization)

Instead of Marshaling/Unmarshaling each build event to redact API keys, redact by traversing the object structure with reflection and redacting any strings encountered. This should save about 1% of total CPU usage in prod.

For a stream of 1800 build events (`bb build //...` in the buildbuddy repo), saves 600K allocations totaling 44MB, and has 20X lower CPU overhead.

```
BenchmarkRedact_MarshalUnmarshal-8            16          69822699 ns/op        44800414 B/op     603938 allocs/op
BenchmarkRedact_Reflect-8                    357           3166731 ns/op             784 B/op         31 allocs/op
```

---

**Version bump**: Patch

### DIFF
--- a/server/util/redact/BUILD
+++ b/server/util/redact/BUILD
@@ -26,5 +26,6 @@ go_test(
         "//server/testutil/testenv",
         "@com_github_stretchr_testify//assert",
         "@com_github_stretchr_testify//require",
+        "@org_golang_google_protobuf//encoding/protojson",
     ],
 )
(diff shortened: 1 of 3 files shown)


bduffany:
Sending out just to gauge reactions to the code -- I can clean up the benchmarks / flags etc. if the overall approach looks good.

tylerwilliams (approved):

PR #3158, opened 2023-01-02
Add direct group navigation (admin-only)

bduffany:
This allows impersonating a group directly (by group ID or slug), rather than needing an invocation ID (which is often inconvenient).

https://user-images.githubusercontent.com/2414826/210277974-d76a5494-e792-49d6-900b-bce15c143603.mp4

---

**Version bump**: Patch


siggisim (approved):
This is awesome. Could we support group_id too?

bduffany:
> This is awesome. Could we support group_id too?

Done - also renamed "Find org" to "Go to org".

PR #3163, opened 2023-01-03
Add error if we try to pass offset or limit to cache reader for compressed data

maggie-lou:
We added a check in the byte stream server to not use compression if we're using an offset or limit, but we should add an error in the cache itself so if we accidentally do this in the future it will be more clear what the problem is


maggie-lou:
* master
  * **PR #3163** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3163" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3163?utm_source=stack-comment).

bduffany:
Do you mind adding the tests from https://github.com/buildbuddy-io/buildbuddy/pull/3133/files into this PR, except in the offset case just assert that an error is returned?

maggie-lou:
Nevermind - I'd missed this [PR](https://github.com/buildbuddy-io/buildbuddy/pull/3136/files) where I see this was already added

PR #3168, opened 2023-01-04
Fix compression bug in BatchReadBlobs

maggie-lou:
This fixes a bug that caused invalid digest errors when we tried to turn compression on for executors. There was a period of time where we'd added the code to support compression to pebble, but we hadn't enabled compression for callers to the cache. Because we didn't have this condition, we were reading compressed data from the cache, but later in the cas_server code there's some additional logic that checks whether the cache supports compression. Because one spot thought compression was on and the other didn't, that caused the errors.


maggie-lou:
* master
  * **PR #3168** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3168" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3168?utm_source=stack-comment).

tylerwilliams (approved):
LGTM

Any idea why this wasn't being hit when we ran remote-exec locally, but *was* showing up in dev?

maggie-lou:
> LGTM
> 
> Any idea why this wasn't being hit when we ran remote-exec locally, but _was_ showing up in dev?

It needed a specific confluence of factors - code where pebble supports compression is merged, config is using pebble cache, pebble compression support turned off (enable_zstd_compression:false). When I used those locally was able to recreate the error

PR #3169, opened 2023-01-04
Performance: Remove unused syscall in disk.FileReader

bduffany:
When `length > 0`, we make a `stat` syscall and discard the result. This PR removes the unnecessary `stat`.

---

**Version bump**: Patch

### DIFF
--- a/server/util/disk/disk.go
+++ b/server/util/disk/disk.go
@@ -205,12 +205,12 @@ func FileReader(ctx context.Context, fullPath string, offset, length int64) (io.
 	if err != nil {
 		return nil, err
 	}
+	if length > 0 {
+		return &readCloser{io.NewSectionReader(f, offset, length), f, ctx}, nil
+	}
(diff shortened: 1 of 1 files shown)


server/util/disk/disk.go
```diff
@@ -205,13 +205,13 @@ func FileReader(ctx context.Context, fullPath string, offset, length int64) (io.
 	if err != nil {
 		return nil, err
 	}
+	if length > 0 {
+		return &readCloser{io.NewSectionReader(f, offset, length), f, ctx}, nil
+	}
 	info, err := f.Stat()
 	if err != nil {
 		return nil, err
 	}
```
bduffany, line 214: Note the behavior is changed in that we no longer return an error if `length > 0` and `Stat()` would fail. But I think that `Stat` should almost never fail because the preceding call to `Open` would have failed.

tylerwilliams (approved):
nice

PR #3171, opened 2023-01-04
Allow downloading single file in zipped test outputs, behind a flag. (Revert 3153 revert 3101 split shard artifacts)

jdhollen:
This addresses the error from #3153 where, for very large zip outputs with lots of files, we were panicking due to a negative index in the read files.

This also wraps everything in a flag (used from the client).


server/static/static.go
```diff
@@ -39,6 +39,7 @@ var (
 	expandedSuggestionsEnabled = flag.Bool("app.expanded_suggestions_enabled", false, "If set, enable more build suggestions in the UI.")
 	enableWorkflows            = flag.Bool("remote_execution.enable_workflows", false, "Whether to enable BuildBuddy workflows.")
 	enableExecutorKeyCreation  = flag.Bool("remote_execution.enable_executor_key_creation", false, "If enabled, UI will allow executor keys to be created.")
+	testOutputManifestsEnabled = flag.Bool("app.test_output_manifests_enabled", false, "If set, the target page will render the contents of test output zips.")
```
tylerwilliams, line 42: you could default this to true if you want

tylerwilliams (approved):

server/static/static.go
jdhollen: sure yeah done

PR #3173, opened 2023-01-04
Parity between disk cache and pebble cache metrics.

vadimberezniker:
---

**Version bump**: Minor

### DIFF
--- a/server/backends/disk_cache/disk_cache.go
+++ b/server/backends/disk_cache/disk_cache.go
@@ -31,6 +31,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/statusz"
 	"github.com/docker/go-units"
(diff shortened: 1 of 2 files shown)


server/backends/disk_cache/disk_cache.go
```diff
@@ -50,6 +51,9 @@ const (
 	// check the cache size.
 	janitorCheckPeriod = 100 * time.Millisecond
 
+	// refreshMetricsPeriod is how often we update polled metrics.
+	refreshMetricsPeriod = 5 * time.Second
```
tylerwilliams, line 55: wdyt about a minute instead?

tylerwilliams (approved):

server/backends/disk_cache/disk_cache.go
vadimberezniker: All of the metrics are cheap so I don't see the harm of exporting it more frequently.

PR #3174, opened 2023-01-04
Add a duration filter to the history page.

iain-macdonald:
This PR adds a new, double-ended slider (from [this React component](https://github.com/zillow/react-slider)) to the advanced filters on the history page letting users filter invocations by duration. The slider is semi-logarithmic, with steps at [1,2,3,4,5,10,15,20,25,30,40,50] of seconds and minutes, and then one-hour increments above that up to 12 hours and one final option of 1 day. This works by mapping linear slider values to these semi-logarithmic values in the filtering logic. I thought that 1 day seemed like a good stopping point as there are only 3 invocations longer than that in the database (out of 55 million, full list of 100 longest [here](https://docs.google.com/document/d/1TbIdVWe_1oH2SAX83B5Dj4damO5PreTyJFUh3XvnunM/edit)), but I'm open to alternatives. Previously the maximum value was infinity, but that's more awkward to support when mapping these to values stored as seconds in the code, so I switched it to 1 day.

Here's how it looks (I did my best! Open to suggestions / fixes.):
Jan-05-2023 17-21-02

**Version bump**: Minor


iain-macdonald:
Hold off on reviewing this, I'm going to see if using this component looks a little better: https://mui.com/material-ui/react-slider/#range-slider

app/components/slider/slider.css
```diff
@@ -0,0 +1,38 @@
...
+  margin-left: 20px;
+}
+
+.slider-thumb-0 .slider-thumb-value {
+  margin-left: -16px;
+}
+
+.slider-thumb-1 .slider-thumb-value {
+  margin-right: -16px;
+}
```
bduffany, line 38: To avoid negative margins here, maybe you could make the `.slider-thumb` parent (which contains the circle and the value) a column-directed flexbox with items centered vertically, like

```css
.slider-thumb {
  display: flex;
  flex-direction: column;
  align-items: center;
}
```

If you still want to nudge the slider-thumb-value text a bit, you could apply a non-negative padding-right rather than a negative margin-left.

enterprise/app/filter/filter.tsx
```diff
@@ -369,6 +394,11 @@ export default class FilterComponent extends React.Component<FilterProps, State>
                 <Wrench /> {commandValue}
               </span>
             )}
+            {(minimumDurationValue || maximumDurationValue) && (
+              <span className="advanced-badge">
+                <Timer /> {minimumDurationValue} - {maximumDurationValue}
```
bduffany, line 399: nit: the timer icon is not very legible since it's very small (12px wide). Could you increase it to 16px?

image

I think the relevant style is `.advanced-badge svg` which is setting it to 12px in `filter.css`.

app/components/slider/slider.css
```diff
@@ -0,0 +1,38 @@
...
+  height: 2px;
+  background-color: gray;
+}
+
+.slider-thumb {
+  width: 32px;
+}
+
+.slider-thumb-circle {
+  margin-top: -5px;
```
bduffany, line 15: nit, negative margins are generally a last resort since they make the layout harder to understand (especially since devtools doesn't render them visually)

You might be able to achieve the same effect by applying padding to a parent element, or something like that.

Also, the 5px value is somewhat confusing -- if this is calculated from something, could you leave a comment? I usually write something like `/* = 4px (half of height) + 1px offset */`

app/components/slider/slider.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import ReactSlider from "react-slider";
+
+export type SliderProps = Omit<JSX.IntrinsicElements["input"], "type">;
+
+export function Slider({ className, ...rest }: SliderProps) {
+  return (
+    <ReactSlider className="horizontal-slider" thumbClassName="slider-thumb" trackClassName="slider-track" {...rest} />
```
bduffany, line 8: nit: forward the `className` prop through, like this:

```suggestion
    <ReactSlider className={`horizontal-slider ${className || ""}`} thumbClassName="slider-thumb" trackClassName="slider-track" {...rest} />
```

bduffany (commented):
Looking great so far!

A couple of notes from playing around with it --

* The bounding box of the slider handles is a bit too small and off-center, making it hard to grab, e.g. if I click slightly above the handle, it doesn't detect a click:

image

Could you center the bounding box for those and ideally make it 48px by 48px?

*It might be nice to have the middle bar of the slider be a slightly different color, so that it's more clear that it's the "selected" region. Maybe just make the non-selected regions dimmer, like material gray 300: https://materialui.co/colors/

* The slider overall feels a bit small. Maybe increase the vertical height from 2px to 4px and give a `border-radius: 2px;` on the slider bar container to make it nice and rounded. Something like

```css
.slider-track {
  height: 4px;
}
.slider-track-0 {
  border-top-left-radius: 2px;
  border-bottom-left-radius: 2px;
}
.slider-track-2 {
    border-top-right-radius: 2px;
    border-bottom-right-radius: 2px;
}
```

bduffany:
> Hold off on reviewing this, I'm going to see if using this component looks a little better: https://mui.com/material-ui/react-slider/#range-slider

Oops, I saw this a little too late :sweat_smile: 

I am a little weary of pulling in any MUI deps; they are a little heavyweight and don't fit our general aesthetic. I think there's value in using MUI rather than having our own custom component lib, but we'd probably want to migrate all of our other components at the same time.

iain-macdonald:
Ahh, sounds good re: the material slider. I just thought it looked nicer, but hopefully with your input we can get the one here to look just as good.

bduffany:
> Ahh, sounds good re: the material slider. I just thought it looked nicer, but hopefully with your input we can get the one here to look just as good.

sg, let me know if you would like to talk through any of it

app/components/slider/slider.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import ReactSlider from "react-slider";
+
+export type SliderProps = Omit<JSX.IntrinsicElements["input"], "type">;
```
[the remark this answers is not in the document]
iain-macdonald: Done.

enterprise/app/filter/filter_util.tsx
```diff
@@ -138,3 +144,106 @@ export function toStatusParam(statuses: Iterable<invocation.OverallStatus>): str
...
+  }
+  get(key: number): string {
+    return this.forwardMap[key];
+  }
+  revGet(key: string): number {
+    return this.reverseMap[key];
+  }
+}
+
+export class DurationSlider {
```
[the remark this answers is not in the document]
iain-macdonald: Done by removing this class in favor of just a few constants, LMK what you think.

enterprise/app/filter/filter_util.tsx
```diff
@@ -138,3 +144,106 @@ export function toStatusParam(statuses: Iterable<invocation.OverallStatus>): str
...
+      [31, "6h"],
+      [32, "7h"],
+      [33, "8h"],
+      [34, "9h"],
+      [35, "10h"],
+      [36, "11h"],
+      [37, "12h"],
+      [38, "∞"],
+    ])
+  );
```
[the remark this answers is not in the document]
iain-macdonald: Done.

proto/invocation.proto
```diff
@@ -230,6 +230,12 @@ message InvocationQuery {
 
   // The bazel command that was used. Ex: "build", "test", "run"
   string command = 11;
+
+  // The minimum invocation duration in milliseconds.
+  int64 minimum_duration_millis = 12;
+
+  // The maximum invocation duration in milliseconds.
+  int64 maximum_duration_millis = 13;
```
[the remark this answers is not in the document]
iain-macdonald: Done (but I didn't add support for the nanos field of Duration for now).

app/components/slider/slider.css
iain-macdonald: Done, but there's still one negative margin to get the circle onto the slider, let me know if you can see a way to get rid of that one.

enterprise/app/filter/filter.tsx
iain-macdonald: Do you want me to make them all bigger? It's the same size as the other ones right now.

app/components/slider/slider.tsx
iain-macdonald: Removed this because it's not used right now.

iain-macdonald:
I think I addressed all of your comments, PTAL

app/components/slider/slider.tsx
bduffany: If someone passes in `className` as a prop (e.g. because they want to add margins or padding, which is a pretty normal use case) then it will override the `horizontal-slider` className prop and probably break the styling, so I think it's worth forwarding through the className prop like we do for the other components under `app/components`

enterprise/app/filter/filter_util.tsx
```diff
@@ -138,3 +144,48 @@ export function toStatusParam(statuses: Iterable<invocation.OverallStatus>): str
     .sort((a, b) => statusFromString(a) - statusFromString(b))
     .join(" ");
 }
+
+export const DURATION_SLIDER_VALUES = [
```
bduffany, line 148:
```suggestion
/** Duration slider values, in seconds. **/
export const DURATION_SLIDER_VALUES = [
```

enterprise/app/filter/filter_util.tsx
```diff
@@ -138,3 +144,48 @@ export function toStatusParam(statuses: Iterable<invocation.OverallStatus>): str
...
+  240,
+  300,
+  600,
+  900,
+  1200,
+  1500,
+  1800,
+  2400,
+  3000,
+  3600,
```
bduffany, line 174: (uber nit, just to make these easier to read / tweak in the future)

```suggestion
  2 * 60,
  3 * 60,
  4 * 60,
  5 * 60,
  10 * 60,
  15 * 60,
  20 * 60,
  25 * 60,
  30 * 60,
  40 * 60,
  50 * 60,
  60 * 60,
```

enterprise/app/filter/filter_util.tsx
```diff
@@ -138,3 +144,48 @@ export function toStatusParam(statuses: Iterable<invocation.OverallStatus>): str
...
+  14400,
+  18000,
+  21600,
+  25200,
+  28800,
+  32400,
+  36000,
+  39600,
+  43200,
+  86400,
```
bduffany, line 186:
```suggestion
  2 * 60 * 60,
  3 * 60 * 60,
  4 * 60 * 60,
  5 * 60 * 60,
  6 * 60 * 60,
  7 * 60 * 60,
  8 * 60 * 60,
  9 * 60 * 60,
  10 * 60 * 60,
  11 * 60 * 60,
  12 * 60 * 60,
  24 * 60 * 60,
```

enterprise/app/history/history.tsx
```diff
@@ -117,6 +118,12 @@ export default class HistoryComponent extends React.Component<Props, State> {
         commitSha: this.props.commit || filterParams.commit,
         command: filterParams.command,
         groupId: this.props.user?.selectedGroup?.id,
+        minimumDuration: {
+          seconds: new Long(Number(filterParams.minimumDuration)),
+        },
+        maximumDuration: {
+          seconds: new Long(Number(filterParams.maximumDuration)),
+        },
```
bduffany, line 126: nit, you can omit `Long()` here -- our client side RPCs all use regular old numbers currently which is nice because `Long` is sort of a protobufjs-ism and we might want to move away from it at some point.

enterprise/app/filter/filter.tsx
bduffany: Actually, I think we could switch to the `Clock` icon which we already use for invocation durations:

image

The clock hands are larger compared to the `Timer` icon so it should be legible at 12px.

app/components/slider/slider.css
bduffany: Looks like you got rid of it, nice :+1:

app/components/slider/slider.tsx
iain-macdonald: Done.

enterprise/app/filter/filter_util.tsx
iain-macdonald: Done.

enterprise/app/filter/filter_util.tsx
iain-macdonald: Done.

enterprise/app/history/history.tsx
iain-macdonald: Doesn't work :-(

https://app.buildbuddy.io/invocation/fa15669d-d691-426a-a1e6-4494882aac01

PR #3176, opened 2023-01-05
Add some small tests for zip manifest parsing.

jdhollen:
This includes a test that would have caught my earlier mistake.  Split up functions a bit so that there is a testable blob of code without needing to mock out grpc


tylerwilliams (approved):
LGTM

sad that 42.zip is not in this CL

PR #3178, opened 2023-01-05
Clarify executor task queue resource logging

bduffany:
With the current logging, it's unclear whether the information logged represents the queue state before the task is scheduled or after the task is scheduled.

This PR updates the logging so that it happens whenever the queue's resource capacity changes, and adds a prefix that indicates whether the change was due to a task being scheduled or being completed. It also adds additional useful info and improves the number formatting to make it more readable.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
+++ b/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
@@ -22,6 +22,8 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/text/language"
+	"golang.org/x/text/message"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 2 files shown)


enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -463,3 +475,13 @@ func (q *PriorityTaskScheduler) GetQueuedTaskReservations() []*scpb.EnqueueTaskR
 	defer q.mu.Unlock()
 	return q.q.GetAll()
 }
+
+// Formats a large number as groups of 3 digits separated by underscores.
```
vadimberezniker, line 479: Elsewhere we use the following which gives comma separated numbers:
```
	p := message.NewPrinter(language.English)
    return p.Sprint(number)`
```

vadimberezniker (approved):
Very nice, thanks!

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
bduffany: ah cool, TIL. Done!

PR #3179, opened 2023-01-05
Sort the 'longest build' trend bar's landing page by build duration.

jdhollen:
i don't even necessarily know that i want to keep this chart around instead of having 50/75/90 percentile lines, but for now this seems better than what we've got.

the interfaces for mucking with the URL definitely need some love, but figured i'd do this first.


bduffany (approved):
nice!

PR #3181, opened 2023-01-05
Fallback to invocation.duration_usec

luluz66:
When buildToollogs event is missing, we currently display 0s. This PR
fallbacks to invocation.duration_usec.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1780

### DIFF
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -426,12 +426,7 @@ export default class InvocationModel {
   }
 
   getEndTimeDate(): Date {
-    let durationMillis = 0;
-    if (!this.finished && this.started) {
-      durationMillis = Math.max(0, new Date().getTime() - this.getStartTimeDate().getTime());
-    } else {
-      durationMillis = Number(this.toolLogMap.get("elapsed time") || 0) * 1_000;
-    }
+    let durationMillis = +this.getDurationMicros() * 1000;
     return new Date(this.getStartTimeDate().getTime() + durationMillis);
   }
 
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_model.tsx
```diff
@@ -426,12 +426,7 @@ export default class InvocationModel {
   }
 
   getEndTimeDate(): Date {
-    let durationMillis = 0;
-    if (!this.finished && this.started) {
-      durationMillis = Math.max(0, new Date().getTime() - this.getStartTimeDate().getTime());
-    } else {
-      durationMillis = Number(this.toolLogMap.get("elapsed time") || 0) * 1_000;
-    }
+    let durationMillis = +this.getDurationMicros() * 1000;
```
bduffany, line 429: getDurationMicros() already returns number so the `+` is redundant:

```suggestion
    let durationMillis = this.getDurationMicros() * 1000;
```

app/invocation/invocation_model.tsx
```diff
@@ -443,21 +438,24 @@ export default class InvocationModel {
...
+    if (this.toolLogMap.has("elapsed time")) {
+      return +this.toolLogMap.get("elapsed time") * 1000000;
     }
+    return +this.invocations.find(() => true)?.durationUsec;
+  }
 
-    return `${this.toolLogMap.get("elapsed time")} seconds`;
+  getDurationSeconds() {
+    let durationMicros = this.getDurationMicros();
+    return `${+durationMicros / 1000000} seconds`;
```
bduffany, line 453:
```suggestion
    return `${durationMicros / 1000000} seconds`;
```

bduffany (approved):

app/invocation/invocation_model.tsx
luluz66: Done.

app/invocation/invocation_model.tsx
luluz66: Done.

PR #3182, opened 2023-01-05
Take compression into account for cache upload/download byte metrics

maggie-lou:
Before, we always read/wrote decompressed bytes from the cache, so we could use the digest size as the number of bytes uploaded/downloaded from the cache. Now that the cache can read/write decompressed bytes, we need to modify the metric to use the actual number of bytes transferred.

### DIFF
--- a/server/util/compression/compression.go
+++ b/server/util/compression/compression.go
@@ -117,6 +117,16 @@ func (d *zstdDecompressor) Close() error {
 	return lastErr
 }
 
+type ZstdCompressor struct {
+	*io.PipeReader
+	// numDecompressedBytes tracks the number of decompressed bytes that were passed into the compressor
+	numDecompressedBytes int
+}
+
+func (c ZstdCompressor) NumDecompressedBytes() int {
+	return c.numDecompressedBytes
+}
+
 // NewZstdCompressingReader returns a reader that reads chunks from the given
 // reader into the read buffer, and makes the zstd-compressed chunks available
 // on the output reader. Each chunk read into the read buffer is immediately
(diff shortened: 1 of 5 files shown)


maggie-lou:
* master
  * **PR #3182** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3182" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3182?utm_source=stack-comment).

siggisim:
Does this affect the Usages table?

siggisim:
If so - we should probably track uncompressed vs compressed bytes separately, as it'll affect billing.

maggie-lou:
> Does this affect the Usages table?

No - it will only change the grafana metrics remote_cache.download_size_bytes and remote_cache.upload_size_bytes

server/util/compression/compression.go
```diff
@@ -138,6 +148,9 @@ func (d *zstdDecompressor) Close() error {
 // at least a few hundred bytes.
 func NewZstdCompressingReader(reader io.Reader, readBuf []byte, compressBuf []byte) (io.ReadCloser, error) {
```
bduffany, line 149: would something like this work ?

```
counter := ioutil.Counter{}
compressed := compression.NewZstdCompressingReader(io.TeeReader(counter, reader), ...)
doSomethingWith(counter.Count())
```

tylerwilliams (approved):
Thanks for considering the alternative

server/util/compression/compression.go
maggie-lou: Nice! Done. @bduffany could you look over the final version when you have a chance?

PR #3183, opened 2023-01-05
Recover and alert on both gRPC and HTTP panics

tylerwilliams:
- Recover and alert on gRPC panics
- Recover and alert on HTTP panics


vadimberezniker (approved):

server/rpc/interceptors/interceptors.go
```diff
@@ -257,6 +260,39 @@ func quotaStreamServerInterceptor(env environment.Env) grpc.StreamServerIntercep
...
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (rsp interface{}, err error) {
+		defer func() {
+			if panicErr := recover(); panicErr != nil {
+				rsp = nil
+				err = status.InternalError("A panic occurred")
+				alertOnPanic()
+			}
+		}()
+		rsp, err = handler(ctx, req)
+		return
```
jdhollen, line 279: Is this the, uh, idiomatic way to write these statements (instead of just doing `return handler(ctx, req)` (and `return handler(srv, stream)` below)?  Sincerely just a question, just learning what I should be doing to fit in.

jdhollen (approved):

PR #3184, opened 2023-01-05
add a test for zip file downloads and drop a bit of validation.

jdhollen:
As commented in the code, this validation isn't actually performed in the golang zip utils--it's something i wrote myself, and it turns out trying to perform it actually makes your life harder because it's sometimes not present.

### DIFF
--- a/server/bytestream/bytestream_test.go
+++ b/server/bytestream/bytestream_test.go
@@ -4,14 +4,13 @@ import (
 	"bytes"
 	"context"
 	"io"
-	"log"
 	"net/url"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/bazelbuild/rules_go/go/tools/bazel"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
-	"github.com/stretchr/testify/assert"
 
 	zipb "github.com/buildbuddy-io/buildbuddy/proto/zip"
 )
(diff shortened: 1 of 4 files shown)


server/bytestream/bytestream_test.go
```diff
@@ -62,3 +68,35 @@ func TestManifest_TooManyFilesZip(t *testing.T) {
 	assert.Nil(t, manifest)
 	assert.Contains(t, err.Error(), "code = Unimplemented")
 }
+
+func createBytestreamer(b []byte) Bytestreamer {
+	return func(ctx context.Context, env environment.Env, url *url.URL, offset int64, limit int64, writer io.Writer) error {
+		log.Printf("%d %d", offset, limit-offset)
```
tylerwilliams, line 74: you wanna keep this?

server/bytestream/bytestream_test.go
```diff
@@ -62,3 +68,35 @@ func TestManifest_TooManyFilesZip(t *testing.T) {
...
+		return nil
+	}
+}
+
+func validateZipContents(t *testing.T, entry *zipb.ManifestEntry, expectedContent string, streamer Bytestreamer) {
+	var buf bytes.Buffer
+	streamSingleFileFromBytestreamZipInternal(nil, nil, nil, entry, &buf, streamer)
+	out := make([]byte, entry.GetUncompressedSize())
+	_, err := io.ReadFull(&buf, out)
+	assert.Nil(t, err)
```
tylerwilliams, line 85: there is also a "require" package instead of "assert" if you need to fail the test as soon as a precondition is false

server/bytestream/bytestream_test.go
jdhollen: Ah crud, thought i got all of these.

server/bytestream/bytestream_test.go
jdhollen: I simplified these and dropped the dep on the asserts package so that i could feel cool / like a Gopher or whatever.

tylerwilliams (approved):
LGTM this works too

PR #3192, opened 2023-01-07
Fix types in //enterprise/app/filter to comply with TypeScript strict type checking.

iain-macdonald:
Starting with these because I've done the most damage in here already. I'm curious what you think of what I did to the advancedFilterStateFromUrl function and if there's a better way to solve that problem -- I thought of one, to make State a class with updating functionality, but that's a bit more involved.

**Version bump**: Minor
**Related issues**: Unblocks https://github.com/buildbuddy-io/buildbuddy-internal/issues/936

### DIFF
--- a/enterprise/app/filter/filter.tsx
+++ b/enterprise/app/filter/filter.tsx
@@ -167,31 +167,6 @@ export default class FilterComponent extends React.Component<FilterProps, State>
     };
   }
 
-  advancedFilterStateFromUrl(search: URLSearchParams) {
-    return {
-      isAdvancedFilterOpen: Boolean(
-        search.get(USER_PARAM_NAME) ||
-          search.get(REPO_PARAM_NAME) ||
-          search.get(BRANCH_PARAM_NAME) ||
-          search.get(COMMIT_PARAM_NAME) ||
-          search.get(HOST_PARAM_NAME) ||
-          search.get(COMMAND_PARAM_NAME) ||
-          search.get(MINIMUM_DURATION_PARAM_NAME) ||
-          search.get(MAXIMUM_DURATION_PARAM_NAME)
-      ),
-      user: search.get(USER_PARAM_NAME),
-      repo: search.get(REPO_PARAM_NAME),
-      branch: search.get(BRANCH_PARAM_NAME),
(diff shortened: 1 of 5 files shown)


enterprise/app/filter/filter_util.tsx
```diff
@@ -37,7 +37,7 @@ export type SortBy =
   | "cache-down"
   | "cache-up"
   | "cache-xfer";
-export type SortOrder = "asc" | "desc";
+export type SortOrder = "" | "asc" | "desc";
```
bduffany, line 40: I think it would be cleaner to leave `SortOrder` as-is and then change the call-site to `SortOrder | ""` or `SortOrder | null` to make it more clear that it's an optional sort order, rather than mixing "optional" and "present" values into the same type.

enterprise/app/filter/filter.tsx
```diff
@@ -123,16 +126,44 @@ export default class FilterComponent extends React.Component<FilterProps, State>
...
+      sortBy: search.get(SORT_BY_PARAM_NAME) || "",
+      sortOrder: search.get(SORT_ORDER_PARAM_NAME) || "",
+    };
+  }
+
+  updateFilterState(prevState: State, search: URLSearchParams) {
+    return {
+      isDatePickerOpen: prevState.isDatePickerOpen,
+      isFilterMenuOpen: prevState.isFilterMenuOpen,
+      isSortMenuOpen: prevState.isSortMenuOpen,
```
bduffany, line 146: when updating state, the keys which aren't being updated can be omitted:

```suggestion
```

enterprise/app/filter/filter_util.tsx
iain-macdonald: Changed these and other filter parameters to be optional and set to undefined when not set, per offline discussion.

enterprise/app/filter/filter.tsx
iain-macdonald: Oh weird, I thought these were required to be set if they weren't optional in the State declaration, but it looks like it works fine without setting them, so removed.

bduffany (approved):
nice!

PR #3199, opened 2023-01-10
Store invocation links in redis

luluz66:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1934

Also, rename some methods in ExecutionCollector interfaces to make is
clearer as it starts to collect invocation, invocation links in addition to executions.


proto/stored_invocation.proto
```diff
@@ -17,3 +17,10 @@ message StoredInvocation {
   int64 invocation_status = 10;
   bool success = 11;
 }
+
+// Proto representation of the InvocationExecution. Only used in backends.
+message StoredInvocationLink {
+  string invocation_id = 1;
+  string execution_id = 2;
+  int32 type = 3;
```
vadimberezniker, line 25: Can you make this an enum? It would be good if we could share the enum between this code and the mysql code.

vadimberezniker (approved):
I notic we don't set a TTL for the execution or the execution link keys. It would be good to do that in case the normal cleanup is not working properly for some reason. Don't need to do this in this PR though.

enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
```diff
@@ -47,6 +48,10 @@ func getInvocationKey(iid string) string {
 	return strings.Join([]string{redisInvocationKeyPrefix, iid}, "/")
 }
 
+func getInvocationLinkKey(execution_id string) string {
```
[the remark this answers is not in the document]
luluz66: oops. Done

proto/stored_invocation.proto
luluz66: Done

luluz66:
> I notic we don't set a TTL for the execution or the execution link keys. It would be good to do that in case the normal cleanup is not working properly for some reason. Don't need to do this in this PR though.

I will address it in a follow-up PR.

PR #3200, opened 2023-01-10
Add a 'strict' option to ts_library that runs TypeScript type checking in strict mode.

iain-macdonald:
I know that you both said you don't think we should do this but I took a closer look and:
1. There are a bunch of errors, like 100+ across 40+ files based on my sample and while most of them are straightforward, there are some that are kind of involved. So, it'll probably take a while to write, test, and deploy fixes for the entire repo.
2. This makes it easier to tackle bit by bit because at least you know which package you're working on.
3. This doesn't seem that bad to me?

Curious to hear what both of you think, though.

**Version bump**: None
**Related issues**: Unblocks https://github.com/buildbuddy-io/buildbuddy-internal/issues/936

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -71,6 +71,12 @@ ts_config(
     src = ":tsconfig.json",
 )
 
+ts_config(
+    name = "tsconfig_strict",
+    src = ":tsconfig_strict.json",
+    deps = [":tsconfig"],
+)
+
 filegroup(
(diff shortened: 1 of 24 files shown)


jdhollen (approved):
I'd like for Brandon to OK the change too, but provided this isn't doing anything gnarly to build times, I'm fine with it.

PR #3207, opened 2023-01-11
Allow clicking on percentile charts

jdhollen:
I'll be using this on some charts for build times.

There isn't a concrete click handler for lines that provides the "closest" point, so I had to improvise a bit: this just uses the last date / x-axis value that we showed a tooltip for.  Since the percentile chart *always* shows a tooltip on hover, this seems alright to me--the user's click action will match the date that recharts thought they were hovering on.

### DIFF
--- a/enterprise/app/trends/percentile_chart.tsx
+++ b/enterprise/app/trends/percentile_chart.tsx
@@ -1,7 +1,17 @@
 import React from "react";
 
 import * as format from "../../../app/format/format";
-import { ResponsiveContainer, ComposedChart, CartesianGrid, XAxis, YAxis, Line, Legend, Tooltip } from "recharts";
+import {
+  ResponsiveContainer,
+  ComposedChart,
+  CartesianGrid,
+  XAxis,
+  YAxis,
+  Line,
+  Legend,
+  Tooltip,
+  CurveProps,
+} from "recharts";
 
 export interface PercentilesChartProps {
   title: string;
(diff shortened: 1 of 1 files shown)


brentleyjones:
Does this still work for mobile (thinking about how hover works differently there)?

enterprise/app/trends/percentile_chart.tsx
```diff
@@ -100,9 +115,11 @@ function PercentilesChartTooltip({
   extractP90,
   extractP95,
   extractP99,
+  triggerCallback,
 }: PercentilesChartTooltipProps) {
   if (!active) return null;
   let data = payload[0].payload;
+  triggerCallback(data);
```
bduffany, line 122: nit: IIRC the React docs strongly recommend against having `render()` do anything that can mutate state. Would it work to put this in a class component and maybe put this in `componentDidUpdate`?

jdhollen:
@brentleyjones this *should* at least click out to the right thing on mobile because a hover event will fire prior to the click, but obviously it busts up "tap to see the tooltip" when the user does something that the browser counts as a click.

we currently have the same problem with bar charts on mobile: tapping on a bar with a click handler will immediately take you through.

we could do something gnarly where we validate that the tooltip is (a) showing and (b) matches the current value that we think the user just clicked on (i.e., two taps to actually click through on mobile).  this is a bit of a pain to test with a local instance, though.  how bout i get this out to dev and then go from there?  this code is unused / will be flag-guarded anyway.

enterprise/app/trends/percentile_chart.tsx
jdhollen: Sure yeah done

PR #3211, opened 2023-01-11
Remove isolate by group ids.

tylerwilliams:
- Make isolate_by_group_ids the default
- Buildfix


enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -1569,121 +1344,6 @@ func TestLRU_IsolateByGroupIDs(t *testing.T) {
 	require.LessOrEqual(t, len(evictionsByQuartile[2]), len(evictionsByQuartile[3]))
 }
 
-func TestMigrationFromDiskV1(t *testing.T) {
```
tylerwilliams, line 1572: I removed these tests because they assume isolate_by_group_ids is *false*. I'm going to remove this functionality in a follow up CL because now migration cache is the way to migrate anyway.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -576,7 +569,10 @@ func (p *PebbleCache) deleteOrphanedFiles(quitChan chan struct{}) error {
 
 			dupPartitionPattern := fmt.Sprintf("%s/%s", part.ID, part.ID)
 			pathHasDupPartitionID := strings.Contains(path, dupPartitionPattern)
```
maggie-lou, line 571: We can remove this - the migrated caches shouldn't have the duplicated partitions

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -661,7 +657,10 @@ func (p *PebbleCache) scanForBrokenFiles(quitChan chan struct{}) error {
 			continue
 		}
 		pathHasDupPartitionID := !bytes.HasPrefix(fileMetadataKey, []byte("PT"))
-		blobDir = p.blobDir(pathHasDupPartitionID, fileMetadata.GetFileRecord().GetIsolation().GetPartitionId())
+		blobDir = p.blobDir(fileMetadata.GetFileRecord().GetIsolation().GetPartitionId())
+		if pathHasDupPartitionID {
```
maggie-lou, line 661: Can remove here too

maggie-lou:
You can delete the field from MigrationCache::Config too https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/backends/migration_cache/config.go#L47

vadimberezniker (approved):

PR #3212, opened 2023-01-12
WIP Add webdriver test for dev

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/enterprise/server/test/webdriver/invocation/invocation_test.go
+++ b/enterprise/server/test/webdriver/invocation/invocation_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestAuthenticatedInvocation_LogUploadEnabled(t *testing.T) {
+func TestAuthenticatedInvocation_Local_LogUploadEnabled(t *testing.T) {
 	wt := webtester.New(t)
 	app := buildbuddy_enterprise.Run(t)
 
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #3212** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3212" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3212?utm_source=stack-comment).

enterprise/server/test/webdriver/invocation/invocation_test.go
```diff
@@ -78,3 +78,73 @@ func TestAuthenticatedInvocation_LogUploadEnabled(t *testing.T) {
 	// default BB org, and make sure we get PermissionDenied instead of the
 	// login page
 }
+
+func TestAuthenticatedInvocation_Dev_LogUploadEnabled(t *testing.T) {
```
bduffany, line 82: I know this is WIP so maybe you're already thinking of this -- but I was thinking that rather than having separate `_Dev` suffixes etc., we can abstract out any env-specific config/logic into an interface like `WebdriverTarget` or something (where the `Local` target maybe has an `Init()` implementation that actually spins up the server) and then enable those things explicitly via test flags / env vars

PR #3220, opened 2023-01-12
Remove restrictions on double compressing inlined data in pebble

maggie-lou:
Our estimated cache size is larger than the actual disk usage numbers, meaning we aren't utilizing the full availability of our allocated disk space. A hypothesis is that the actual size on our disks is smaller because Pebble compresses the data that we inline, so it's smaller than we expect. Before we had added checks to not double compress data. To more accurately get the size that's actually written, we want to compress all data, because then it will approximate the compressed size that Pebble is writing to disk, even if this means that inlined data will be compressed again by Pebble.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -298,10 +298,6 @@ func validateOpts(opts *Options) error {
 		}
 	}
 
-	if opts.MinBytesAutoZstdCompression < opts.MaxInlineFileSizeBytes {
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3220** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3220" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3220?utm_source=stack-comment).

tylerwilliams (approved):

PR #3226, opened 2023-01-13
Add PebbleKey with versions

tylerwilliams:
Principles of Migration:
 - you must always be able to read and write multiple versions of the data
 	- this ensures the DB is "rollback-safe"
	- first ship a code change that speaks the new version, then, once you're sure it won't be rolled back, flip the flag to change current-version
 - switch to the new version then once everything is migrated, bump db version so multi-version reads are no longer necessary

A **PebbleKey** is version agnostic -- it can be read or written at any version
	- this is enforced by unit tests that read and write keys at all versions from min-to-max and ensure they are compatible

A **Migrator** is responsible for reading keys *and values* at one version and writing them at another
	- this allows us to upgrade values if we desire

When you write a new key to the DB, it's written at currentVersion (flag controlled)
When you read a key from the DB, it's looked for between (dbVersion, latestVersion)

- currently, everything in the DB is of undefined version
- our first migration will do nothing but store the version in the key
- we'll then migrate everything to version 2 which reorders keys

For simplicity, we'll only ever do one migration at a time.

### DIFF
--- a/enterprise/server/raft/filestore/filestore.go
+++ b/enterprise/server/raft/filestore/filestore.go
@@ -61,17 +61,23 @@ type PebbleKeyVersion int
 const (
 	// UndefinedKeyVersion is the version of all keys in the database
 	// that have not yet been versioned.
-	UndefinedKeyVersion PebbleKeyVersion = 0
+	UndefinedKeyVersion PebbleKeyVersion = iota
 
 	// Version1 is the first key version that includes a version in the
 	// key path, to disambiguate reading old keys.
-	Version1 = 1
+	Version1
 
 	// Version2 is the same as Version1, plus a change that moves the
 	// remote instance name hash to the end of the key rather than the
 	// beginning. This allows for even sampling across the keyspace,
 	// regardless of remote instance name.
-	Version2 = 2
+	Version2
+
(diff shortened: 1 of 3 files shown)


enterprise/server/raft/filestore/filestore.go
```diff
@@ -55,10 +56,147 @@ func fileRecordSegments(r *rfpb.FileRecord) (partID string, groupID string, isol
...
+		}
+		partDir := "/v1/" + PartitionDirectoryPrefix + pmk.partID
+		filePath = filepath.Join(partDir, filePath)
+		return []byte(filePath), nil
+	case Version2:
+		filePath := filepath.Join(pmk.hash, pmk.isolation, pmk.remoteInstanceHash)
+		if pmk.isolation == "ac" {
+			filePath = filepath.Join(pmk.groupID, filePath)
+		}
+		partDir := "/v3/" + PartitionDirectoryPrefix + pmk.partID
```
vadimberezniker, line 119: v2?

tylerwilliams: lol yeah, test caught this and I fixed it :) good 👁️

vadimberezniker (approved):

PR #3230, opened 2023-01-13
set expiration for invocation links and executions in collector

luluz66:


enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
```diff
@@ -81,7 +84,14 @@ func (c *collector) AddInvocationLink(ctx context.Context, link *sipb.StoredInvo
 	if err != nil {
 		return err
 	}
-	return c.rdb.SAdd(ctx, getInvocationLinkKey(link.GetExecutionId()), string(b)).Err()
+	key := getInvocationLinkKey(link.GetExecutionId())
+	if err = c.rdb.SAdd(ctx, key, string(b)).Err(); err != nil {
+		return err
+	}
+	if err = c.rdb.Expire(ctx, key, invocationExpiration).Err(); err != nil {
```
vadimberezniker, line 91: You can use c.rdb.TxPipeline() to send these commands together & have them be atomic

vadimberezniker (approved):

enterprise/server/backends/redis_execution_collector/redis_execution_collector.go
luluz66: Done.

PR #3237, opened 2023-01-17
Flip strict=True for //enterprise/app/root

iain-macdonald:
Version bump: Patch
Related issues: Unblocks https://github.com/buildbuddy-io/buildbuddy-internal/issues/936

### DIFF
--- a/app/root/root.tsx
+++ b/app/root/root.tsx
@@ -16,7 +16,7 @@ import UserPreferences from "../preferences/preferences";
 declare var window: any;
 
 interface State {
-  user: User;
+  user?: User;
   hash: string;
   path: string;
   search: URLSearchParams;
@@ -27,7 +27,7 @@ capabilities.register("BuildBuddy Community Edition", false, [Path.invocationPat
 
 export default class RootComponent extends React.Component {
   state: State = {
-    user: null,
+    user: undefined,
     hash: window.location.hash,
     path: window.location.pathname,
     search: new URLSearchParams(window.location.search),
@@ -80,7 +80,7 @@ export default class RootComponent extends React.Component {
                 hash={this.state.hash}
                 search={this.state.search}
                 preferences={this.state.preferences}
-                user={null}
+                user={undefined}
               />
             )}
             {compareInvocationIds && (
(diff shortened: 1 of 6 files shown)


app/terminal/terminal.tsx
```diff
@@ -97,7 +97,7 @@ export default class TerminalComponent extends React.Component<TerminalProps, St
     window.removeEventListener("keydown", this.windowKeyDownListener);
   }
 
-  componentDidUpdate(_prevProps: TerminalProps, prevState: State, snapshot: Snapshot | null) {
+  componentDidUpdate(_prevProps: TerminalProps, prevState: State, snapshot: Snapshot) {
```
bduffany, line 100: I think maybe this needs to be `snapshot?: Snapshot` based on the type signature here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/522a694de4e20d1ab69598affc0a97d2ee8d2556/types/react/v16/index.d.ts#L681

I remember hitting an issue that snapshot was either null or undefined which is why there's the `if (snapshot && ...)` check below

app/root/root.tsx
```diff
@@ -16,7 +16,7 @@ import UserPreferences from "../preferences/preferences";
 declare var window: any;
 
 interface State {
-  user: User;
+  user?: User;
```
bduffany, line 19: Seems like the tests are failing -- I wonder if we have some logic somewhere that is checking specifically for `user === null`. Might be less risky to change this to `user: User | null`

bduffany (approved):

app/terminal/terminal.tsx
iain-macdonald: Ahh yeah good catch.

app/root/root.tsx
iain-macdonald: Per discussion it's this code in invocation.tsx: https://github.com/buildbuddy-io/buildbuddy/blob/e753ffa527b2fb96b4095d4ee6acb29c32dfbc0f/app/invocation/invocation.tsx#L152

But I flipped it back to "| null" and added a todo to change it to optional in the future.

PR #3238, opened 2023-01-17
Add metrics for bytes sent to/from server and for ac_server

maggie-lou:
In prod, the number of bytes downloaded/uploaded from the cache don't seem to have gone down after turning on compression in the executors. Add additional metrics to look into this more and a server label to help break down where our cache requests are originating from. Also start emitting cache metrics from the action cache server.

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -139,6 +139,9 @@ const (
 	// Describes the type of cache request
 	CacheRequestType = "type"
 
+	// Describes the name of the server that handles a client request, such as "byte_stream_server" or "cas_server"
+	ServerName = "server_name"
+
 	// Describes the type of compression
 	CompressionType = "compression"
 
@@ -311,6 +314,7 @@ var (
 		Help:      "Number of bytes downloaded from the remote cache in each download. Use the **`_sum`** suffix to get the total downloaded bytes and the **`_count`** suffix to get the number of downloaded files.",
 	}, []string{
 		CacheTypeLabel,
+		ServerName,
 	})
 
 	/// #### Examples
@@ -348,6 +352,7 @@ var (
 		Help:      "Number of bytes uploaded to the remote cache in each upload. Use the **`_sum`** suffix to get the total uploaded bytes and the **`_count`** suffix to get the number of uploaded files.",
 	}, []string{
 		CacheTypeLabel,
+		ServerName,
 	})
 
 	/// #### Examples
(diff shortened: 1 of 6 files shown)


maggie-lou:
* master
  * **PR #3238** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3238" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3238?utm_source=stack-comment).

server/remote_cache/hit_tracker/hit_tracker.go
```diff
@@ -331,11 +331,34 @@ func cacheEventTypeLabel(c counterType) string {
...
+			metrics.ServerName:     serverLabel,
+		}).Observe(bytesTransferredCache)
+		metrics.ServerUploadSizeBytes.With(prometheus.Labels{
+			metrics.CacheTypeLabel: cacheTypeLabel,
+			metrics.ServerName:     serverLabel,
+		}).Observe(bytesTransferredClient)
+		metrics.DigestUploadSizeBytes.With(prometheus.Labels{
+			metrics.CacheTypeLabel: cacheTypeLabel,
+			metrics.ServerName:     serverLabel,
+		}).Observe(digestSizeBytes)
```
vadimberezniker, line 347: missing return

server/metrics/metrics.go
```diff
@@ -1538,6 +1543,50 @@ var (
...
+		Subsystem: "server",
+		Name:      "download_size_bytes",
+		Buckets:   prometheus.ExponentialBuckets(1, 10, 9),
+		Help:      "Number of bytes downloaded from the server in each download. Use the **`_sum`** suffix to get the total downloaded bytes and the **`_count`** suffix to get the number of downloaded files.",
+	}, []string{
+		CacheTypeLabel,
+		ServerName,
+	})
+
+	DigestUploadSizeBytes = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
vadimberezniker, line 1568: Is the digest upload/download size interesting? Seems like we care about the actual bytes transferred.

vadimberezniker (approved):

server/remote_cache/hit_tracker/hit_tracker.go
maggie-lou: This function doesn't return anything (before it returned the name of the metric, now it actually emits the metrics)

server/metrics/metrics.go
maggie-lou: Not normally but I want to sanity check that compression is working. I will remove these digest metrics afterwards

PR #3239, opened 2023-01-17
Executor: Warm up new Ubuntu 20.04 images

bduffany:
- Add the Ubuntu 20.04 RBE image and Ubuntu 20.04 CI image to the set of images to be warmed up.
- Move the workflow image warmup behind a flag, so that we don't have to warm up as many Firecracker images during scale up of the "normal" (non-workflow) executor pool. (I plan to enable this new flag only for the workflow executors)

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/auth"
-	"github.com/buildbuddy-io/buildbuddy/enterprise/server/hostedrunner"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/commandutil"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/container"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/containers/bare"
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -836,23 +848,46 @@ func (p *pool) Warmup(ctx context.Context) {
...
 		log.Warningf("Error warming up containers: %s", err)
 	}
 }
 
+func (p *pool) warmupConfigs() []WarmupConfig {
+	var out []WarmupConfig
+	for _, isolation := range platform.GetExecutorProperties().SupportedIsolationTypes {
+		// Warm up the default execution image for all isolation types, as well
+		// as the new Ubuntu 20.04 image.
+		out = append(out, WarmupConfig{
```
vadimberezniker, line 867: Have you looked at what the performance is like warming up 2 images? Curious how much the parallelism helps here.

vadimberezniker (approved):

tylerwilliams (approved):

enterprise/server/remote_execution/runner/runner.go
bduffany: It seems like the parallelism is helping. Here are some numbers for various scenarios I tried (times are in seconds):

```
# This PR
97.0604944562
# This PR + serial image pull
127.32699358059999

# Current behavior @HEAD
74.78818619555555
# Current behavior @HEAD, + disable CI image warmup
65.98134982509998
```

This is measuring the time it takes to warmup all images -- podman pulls complete more quickly than firecracker pulls so the extra 23s added by this PR affects firecracker more than it does podman.

PR #3240, opened 2023-01-17
Take existing blob size into account when updating delta.

vadimberezniker:
Enabling compression might cause some stored sizes to be updated.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1200,9 +1200,9 @@ func (p *PebbleCache) Writer(ctx context.Context, r *resource.ResourceName) (int
 		if err == nil {
 			partitionID := fileRecord.GetIsolation().GetPartitionId()
 			if sizeDelta != 0 {
-				p.sendSizeUpdate(partitionID, fileMetadataKey, bytesWritten)
+				p.sendSizeUpdate(partitionID, fileMetadataKey, sizeDelta)
 			}
-			metrics.DiskCacheAddedFileSizeBytes.With(prometheus.Labels{metrics.CacheNameLabel: p.name}).Observe(float64(sizeDelta))
+			metrics.DiskCacheAddedFileSizeBytes.With(prometheus.Labels{metrics.CacheNameLabel: p.name}).Observe(float64(bytesWritten))
 		}
 
 		return err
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1185,22 +1185,24 @@ func (p *PebbleCache) Writer(ctx context.Context, r *resource.ResourceName) (int
...
 			tracing.AddStringAttributeToCurrentSpan(ctx, "pebble.duplicate_write", "true")
+			sizeDelta = bytesWritten - existingMD.GetStoredSizeBytes()
 		}
 
 		err = db.Set(fileMetadataKey, protoBytes, &pebble.WriteOptions{Sync: false})
 		if err == nil {
 			partitionID := fileRecord.GetIsolation().GetPartitionId()
-			if !alreadyExists {
+			if sizeDelta != 0 {
 				p.sendSizeUpdate(partitionID, fileMetadataKey, bytesWritten)
```
maggie-lou, line 1203: Should this be sizeDelta

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1185,22 +1185,24 @@ func (p *PebbleCache) Writer(ctx context.Context, r *resource.ResourceName) (int
...
 			}
-			metrics.DiskCacheAddedFileSizeBytes.With(prometheus.Labels{metrics.CacheNameLabel: p.name}).Observe(float64(bytesWritten))
+			metrics.DiskCacheAddedFileSizeBytes.With(prometheus.Labels{metrics.CacheNameLabel: p.name}).Observe(float64(sizeDelta))
```
maggie-lou, line 1205: Is it intended that this metric could have a negative value? Not sure whether this metric would be more useful if we used sizeDelta to show that we are overwriting files to be smaller, or if we used bytesWritten if we wanted to know the actual sizes of files on the disk

vadimberezniker: This is supposed to be bytesWritten. I changed the wrong line :facepalm:

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: Yeah :facepalm:

maggie-lou (approved):

PR #3246, opened 2023-01-18
turn on flag --force_message for jspb so that we always use concrete message types.

jdhollen:


bduffany (approved):
Awesome, thanks for doing this :)

iain-macdonald (approved):

PR #3251, opened 2023-01-18
Remove flag to register self-auth handlers

maggie-lou:
We want our webdriver tests to self-auth because google oauth has bot-prevention measures. Remove this check so that the self-auth handlers are always registered. Self-auth will still only be automatically added as an oauth provider if the flag is on. Only organization slugs that have explicitly added self-auth as an auth provider in the config will be able to use it.

### DIFF
--- a/enterprise/server/selfauth/selfauth.go
+++ b/enterprise/server/selfauth/selfauth.go
@@ -95,9 +95,6 @@ type tokenJSON struct {
 }
 
 func Register(env environment.Env) error {
-	if !Enabled() {
-		return nil
-	}
 	oauth, err := NewSelfAuth()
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3251** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3251" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3251?utm_source=stack-comment).

tempoz (commented):
_Deleted by user_

tempoz (approved):
LGTM

PR #3255, opened 2023-01-19
enable strict ts for //apps/enterprise/trends

jdhollen:


enterprise/app/trends/cache_chart.tsx
```diff
@@ -6,29 +6,33 @@ import { invocation } from "../../../proto/invocation_ts_proto";
...
 
-const CacheChartTooltip = ({ active, payload, labelFormatter, extractHits, extractMisses, extractWrites }: any) => {
+const CacheChartTooltip = ({
+  active,
+  payload,
+  labelFormatter,
+  extractHits,
+  secondaryBarName,
+  extractSecondary,
+}: any) => {
```
bduffany, line 23: nit, do you mind defining a `CacheChartTooltipProps` while you're here?

jdhollen: Sorry for sitting on this, here are all the `TooltipProps` you could ever want!

enterprise/app/trends/trends_chart.tsx
```diff
@@ -38,34 +38,37 @@ interface Props {
...
+interface TrendsChartTooltipProps extends TooltipProps<any, any> {
+  labelFormatter: (datum: any) => string;
+  valueFormatter?: (datum: number) => string;
+  secondaryValueFormatter?: (datum: number) => string;
+  extractSecondaryValue?: (datum: any) => number;
+}
+
+class TrendsChartTooltip extends React.Component<TrendsChartTooltipProps> {
+  render() {
+    const p = this.props;
```
bduffany, line 50: uber nit: might want to write out `this.props.<blah>` to be consistent with the code style for the other components -- or alternatively make it a function component again and replace `any` with TrendsChartTooltipProps (then you could reference params directly instead of props)

jdhollen: dunno what i was doing yesterday, somehow convinced myself that i couldn't make strict mode happy with a function component because of something about the way the tooltip worked, but obviously not the case.  swapped this back because i really didn't like have 100-character lines that were repeating `this.props` over and over.

bduffany (approved):

PR #3256, opened 2023-01-19
Don't return an error if an AWS blob is NotFound in the BlobExists method

siggisim:
This should fix the issue with chunked logs on AWS.

---

**Version bump**: Patch


tylerwilliams (approved):

vadimberezniker (approved):
:clap:

PR #3257, opened 2023-01-19
Add statusz option to serve just the version string

bduffany:
This lets us avoid using a hacky regex to extract the version from the /statusz response.

e.g. for a local server:

```
$ curl http://localhost:9090/statusz?version
unknown
```

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1964


tylerwilliams:
(discussed offline, I'd prefer to just have a prom metric for this which should be easily curlable)

tylerwilliams:
whoops, didn't mean to close this but I think you were about to anyway :)

PR #3260, opened 2023-01-19
Add webdriver tests for dev and prod

maggie-lou:
---

**Version bump**: TODO


maggie-lou:
* master
  * **PR #3260** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3260" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3260?utm_source=stack-comment).

enterprise/server/test/webdriver/invocation/invocation_test.go
```diff
@@ -2,18 +2,90 @@ package invocation_test
...
 	"github.com/buildbuddy-io/buildbuddy/server/testutil/testbazel"
 	"github.com/buildbuddy-io/buildbuddy/server/testutil/webtester"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestAuthenticatedInvocation_LogUploadEnabled(t *testing.T) {
+var (
+	targetStr   = flag.String("webdriver_target", "local", "Target that should be tested by the webdriver.")
+	devSSOSlug  = flag.String("webdriver_dev_slug", "", "Dev SSO slug to login.")
```
maggie-lou, line 18: The plan is to save these SSO slugs as secrets in the github action that will do the release. Then after we deploy dev for example, we can run this webdriver test on the dev target with the dev SSO secret

app/invocation/cache_requests_card.tsx
```diff
@@ -302,7 +302,7 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
           {/* Filtering controls */}
           <div className="separator" />
           <label>Show</label>
-          <Select value={this.getFilterIndex()} onChange={this.onChangeFilter.bind(this)}>
+          <Select id="filter-cache-requests" value={this.getFilterIndex()} onChange={this.onChangeFilter.bind(this)}>
```
bduffany, line 305: nit: we usually use `className` instead of `id`

enterprise/server/test/webdriver/invocation/invocation_test.go
bduffany, line 18: Instead of `local`, `dev`, `prod`, wdyt about having `local` and `remote`, where remote has to be explicitly configured with `--remote_app_endpoint` and `--remote_sso_slug`? I think it might be cleaner since this binary would only have to know about one k8s env at a time, instead of having the potential to accidentally communicate with multiple envs

enterprise/server/test/webdriver/invocation/invocation_test.go
```diff
@@ -2,18 +2,90 @@ package invocation_test
...
+	prodSSOSlug = flag.String("webdriver_prod_slug", "", "Prod SSO slug to login.")
+)
+
+type webdriverTarget interface {
```
bduffany, line 22: Can we move these structs and flags to a shared package, maybe `server/testutil/testtarget` or `server/testutil/webtest_target` ? (doesn't need to be under enterprise/, since we might want to have some library functions which depend on the Target, but can be used with OSS webdriver tests). That way when we add more webdriver tests we don't need need to duplicate these

app/invocation/cache_requests_card.tsx
siggisim: Trying webdriver tests to classNames can be brittle (i.e. you could accidentally break a webdriver test by changing the style of an element).

A few options (I'm open to any of theses):
- We could have a special classname prefix that we know will be selected by webdriver, so we shouldn't change it
- We could use another attribute that's only used for webdriver ids like `debug-id="foo"`
- We could make IDs be our 
- We could roll with class names, and just be careful to search the webdriver tests before changing them

Here's a post on selenium locator best practices: https://www.screenster.io/selenium-locators-best-practices-7-helpful-tips-for-ui-testers/

bduffany: I like the `debug-id` idea; we might have to call it `data-debug-id` to make React/TypeScript happy.

server/testutil/webtester/webtester.go
```diff
@@ -168,14 +182,22 @@ func HasClass(el *Element, class string) bool {
 // BuildBuddy-specific functionality
 // ===
 
-// Login uses the Web app to log into BuildBuddy as the default self-auth user.
-// It expects that no user is currently logged in, and that self-auth is
-// enabled.
-func Login(wt *WebTester, appBaseURL string) {
+// LoginSelfAuth uses the Web app to log into BuildBuddy as the default self-auth user.
+// It expects that no user is currently logged in, and that self-auth is enabled on the server.
+func LoginSelfAuth(wt *WebTester, appBaseURL string) {
```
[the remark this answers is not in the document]
maggie-lou: Done. Makes sense to have the local webdriver act as similarly to the remote ones as possible too

server/testutil/webtester/webtester.go
```diff
@@ -87,6 +87,14 @@ func (wt *WebTester) FindAll(cssSelector string) []*Element {
 	return out
 }
 
+// FindName returns the element matching the given CSS name. Exactly one
+// element must be matched, otherwise the test fails.
+func (wt *WebTester) FindName(name string) *Element {
```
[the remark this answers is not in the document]
maggie-lou: Got rid of this

enterprise/server/test/webdriver/invocation/invocation_test.go
```diff
@@ -2,18 +2,90 @@ package invocation_test
...
+		break
+	case "dev":
+		target = &devTarget{}
+		break
+	case "prod":
+		target = &prodTarget{}
+		break
+	default:
+		require.FailNowf(t, "invalid target", "%s is an invalid target for the webdriver invocation tests.", targetStr)
+	}
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/server/test/webdriver/invocation/invocation_test.go
maggie-lou: Moved it under /enterprise because OSS doesn't support auth. If we end up needing these drivers in the OSS tests, we can refactor this later to move the interface to OSS, but keep drivers with auth in /enterprise

enterprise/server/test/webdriver/invocation/invocation_test.go
maggie-lou: Done

app/invocation/cache_requests_card.tsx
maggie-lou: Good idea. Modified to use `debug-id`

siggisim (approved):

bduffany (approved):

PR #3261, opened 2023-01-19
Add metric for binary version

bduffany:
---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1967


tylerwilliams (approved):
Nice! We should add a graph of the raw data of this guage and then we can see rollouts better (and we'll be more likely to catch things like the executors falling behind...)

PR #3264, opened 2023-01-20
Fix action results not showing in UI for EU invocations

bduffany:
This PR fixes a long-standing bug that action results are not showing up in the UI for invocations that use `--remote_executor=europe.remote.buildbuddy.io`, even though the results are present in AC (as they can be fetched using `tools/cas`).

The root cause turned out to be that we were calling `DialTargetWithOptions` with the URL `actioncache://europe.remote.buildbuddy.io`, and `DialTargetWithOptions` parses that URL to the target+options `("europe.remote.buildbuddy.io", grpc.WithInsecure())`. The request fails with the error "connection closed before server preface received" because the `europe.remote.buildbuddy.io` target requires TLS.

This bug does not reproduce in the US because we first attempt to download the file from `localhost` before we attempt to download using the gRPC target (`remote.buildbuddy.io` for US), and the localhost target does not require TLS, so the first request succeeds and we don't continue on to the "fallback" path of requesting via the `remote.buildbuddy.io` target.

The bug also doesn't reproduce for bytestream URLs in EU, because `DialTargetWithOptions` has special handling for the `"bytestream"` URL scheme, but not the `"actioncache"` scheme. So our first request to localhost fails because the file is not hosted in the US, but then the request to EU succeeds only because there's special handling for `bytestream://`.

An alternative fix would have been to change the `DialTargetWithOptions` scheme handling to `(u.Scheme == "bytestream" || u.Scheme == "actioncache")`, but I figured this is error prone and complex. For example, there are a few call sites which use that `DialTargetWithOptions` func, setting `grpcsBytestream` to either `true` or `false`, when in fact both call sites do not need to be setting that param and could instead just call `DialTarget`. So the bytestream:// support is adding complexity to the API of `DialTarget` which is used in a lot of places, but there's really only one caller which needs bytestream:// support (the HTTP handler).

Instead of having `DialTargetWithOptions` be aware of bytestream:// and actioncache:// URLs, we can instead just pass in a `grpcs://` target directly. This also allows us to remove `DialTargetWithOptions` in favor of just `DialTarget`.

I checked all the call sites of both `DialTargetWithOptions` and `DialTarget` and verified that the `/file/download/` handler is the only call site which attempts to call that function with a `bytestream://` URL, so this change should be safe.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/distributed/distributed_test.go
+++ b/enterprise/server/backends/distributed/distributed_test.go
@@ -57,7 +57,7 @@ func newMemoryCache(t *testing.T, maxSizeBytes int64) interfaces.Cache {
 }
 
 func waitForReady(t *testing.T, addr string) {
-	conn, err := grpc_client.DialTargetWithOptions("grpc://"+addr, false, grpc.WithBlock(), grpc.WithTimeout(maxWaitForReadyDuration))
+	conn, err := grpc_client.DialTarget("grpc://"+addr, grpc.WithBlock(), grpc.WithTimeout(maxWaitForReadyDuration))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/server/bytestream/bytestream.go
+++ b/server/bytestream/bytestream.go
@@ -150,6 +150,14 @@ func StreamBytestreamFileChunk(ctx context.Context, env environment.Env, url *ur
 	return err
 }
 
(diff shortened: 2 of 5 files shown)


tylerwilliams (approved):
I really love this, thanks for cleaning this up and checking the callsites!

PR #3270, opened 2023-01-23
invocation: avoid crash when render query profiles

sluongng:
When a profile is generated from `bazel query` or `bazel cquery`, it would contains no executionData and thus, causes a crash in rendering "Timing" tab on the Invocation page.

Fix it by skipping the Breakdown section when the data is empty.

---

**Version bump**: TODO

### DIFF
--- a/app/invocation/invocation_breakdown_card.tsx
+++ b/app/invocation/invocation_breakdown_card.tsx
@@ -71,8 +71,9 @@ export default class InvocationBreakdownCardComponent extends React.Component<Pr
           <div className="title">Timing Breakdown</div>
           <div className="details">
             <div className="cache-sections">
-              {renderBreakdown(phaseData, "Phase breakdown", "Breakdown of build phases")}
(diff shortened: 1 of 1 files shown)


bduffany (approved):
Nice!

bduffany (changes requested):
Actually nvm, I think `phaseData` should never be empty. (It is created as a static list with 4 elements.) The null pointer exception is probably coming from somewhere else?

sluongng:
> Actually nvm, I think `phaseData` and `executionData` should never be empty. (They are created as static, non-empty lists.) The `reduce` exception is probably coming from somewhere else?

There is a `filter()` that comes right after `sort()` that is easy to miss. Essentially we are only displaying non-zero items in the list. If everything is zero then the list should be empty after `filter()` was called.

bduffany (approved):
nvm yet again, I see we are calling `.filter()` on phaseData :hot_face:

PR #3271, opened 2023-01-23
Workflows: Fix bazel shutdown deadlock

bduffany:
Occasionally, workflows timeout with the error "FATAL: Attempted to kill stale server process (pid=XXX) using SIGKILL, but it did not die in a timely fashion."

This error can happen if the Bazel server process becomes stale (reaching `--max_idle_secs`) just after the workflow starts, and it becomes a zombie. Because the vmexec server holds reapMutex while executing tasks, the stale Bazel server cannot be reaped.

This diagram illustrates the deadlock:

image

The fix in this PR is to remove `reapMutex` and instead protect child processes from having their `wait()` status "stolen" by spawning them under an intermediate `vmexec` process:

image

---

**Version bump**: Patch

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1178


enterprise/server/cmd/goinit/main.go
```diff
@@ -229,7 +235,12 @@ func main() {
 	die(mkdirp("/mnt/dev", 0755))
 	die(mount("/dev", "/mnt/dev", "", syscall.MS_MOVE, ""))
 
+	// TODO(bduffany): Spawn vmvfs via the init binary like we do for vmexec, to
+	// save scratch disk space. I tried it, but for some reason the fuse library
+	// panics saying that "/dev/null" doesn't exist:
+	// https://github.com/hanwen/go-fuse/blob/915cf5413cdef5370ae3f953f8eb4cd9ac176d5c/splice/splice.go#L57
 	die(copyFile("/vmvfs", "/mnt/vmvfs", 0555))
+	die(copyFile("/init", "/mnt/init", 0555))
```
tylerwilliams, line 243: Where is this used?

enterprise/server/cmd/goinit/main.go
```diff
@@ -46,6 +45,8 @@ var (
 	setDefaultRoute         = flag.Bool("set_default_route", false, "If true, will set the default eth0 route to 192.168.246.1")
 	initDockerd             = flag.Bool("init_dockerd", false, "If true, init dockerd before accepting exec requests. Requires docker to be installed.")
 	gRPCMaxRecvMsgSizeBytes = flag.Int("grpc_max_recv_msg_size_bytes", 50000000, "Configures the max GRPC receive message size [bytes]")
+
+	isVMExec = flag.Bool("vmexec", false, "Whether to run as the vmexec server.")
```
tylerwilliams, line 49: is this saving a bunch of space? Feels cleaner to call vmexec as a separate binary but maybe I'm missing why you did it like this.

bduffany: It increases the size of the initial RAM disk by 25 MB (it's currently 49 MB). I also figured it's slightly easier to manage flags, and share the common logging setup etc.

I agree that having it as a separate binary is probably a simpler mental model, though. I can make a separate `main` package for it, and then if we find the initrd size becomes a problem, we can have the executor write it directly to the scratch disk instead of having goinit copy it from from initrd to the scratch disk. wdyt?

enterprise/server/cmd/goinit/main.go
bduffany: This is so that we can re-exec the init process with `--vmexec` after we've chroot'ed.

tylerwilliams (approved):

iain-macdonald (approved):
Nice find!

PR #3272, opened 2023-01-23
Firecracker: Monitor peak FS usage

bduffany:
Monitor peak disk usage (for scratch and workspace disks separately) throughout VM action execution and report in ExecutedActionMetadata.

This will let us (and users) confirm whether certain errors are due to the VM running out of disk space.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/goinit/main.go
+++ b/enterprise/server/cmd/goinit/main.go
@@ -312,6 +312,7 @@ func main() {
 	}
 	die(os.WriteFile("/etc/hosts", []byte(strings.Join(hosts, "\n")), 0755))
 	die(os.WriteFile("/etc/resolv.conf", []byte("nameserver 8.8.8.8"), 0755))
+	die(syscall.Symlink("/proc/mounts", "/etc/mtab"))
 
 	if *setDefaultRoute {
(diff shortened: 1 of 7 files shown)


enterprise/server/cmd/goinit/main.go
```diff
@@ -312,6 +312,7 @@ func main() {
 	}
 	die(os.WriteFile("/etc/hosts", []byte(strings.Join(hosts, "\n")), 0755))
 	die(os.WriteFile("/etc/resolv.conf", []byte("nameserver 8.8.8.8"), 0755))
+	die(syscall.Symlink("../proc/self/mounts", "/etc/mtab"))
```
bduffany, line 315: Note, this was needed because `gosigar` was returning an error that `/etc/mtab` doesn't exist. On my system, this file is just a symlink to `../proc/self/mounts` so creating that symlink seemed to do the trick.

bduffany: Yep, done. Also it looks like this can be `/proc/mounts` instead of `/proc/self/mounts`.

tylerwilliams (approved):

PR #3274, opened 2023-01-23
github action: ensure git user name and email are in quotes

sluongng:
Username could contain spaces in the middle and without quotes, git-config would not be able to parse it and throw error with exit code 129.

Fix by adding quotes between the Github Action expressions.

---

**Version bump**: TODO


sluongng:
Example past fail run: https://github.com/buildbuddy-io/buildbuddy/actions/runs/3988392237/jobs/6839483148

bduffany (approved):

PR #3275, opened 2023-01-23
Fix sidebar being hidden on small devices by css bug

siggisim:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1975

---

**Version bump**: None


sluongng (approved):
image

Can confirm that applying this patch fixed the issue.

PR #3280, opened 2023-01-24
[WIP] User-owned API keys

bduffany:
TODOs:
- [ ] Add tests
- [ ] When disabling API keys at the org level, disable current user keys

---

**Version bump**: Patch


enterprise/server/backends/userdb/userdb.go
```diff
@@ -233,30 +241,144 @@ func (d *UserDB) CreateAPIKey(ctx context.Context, groupID string, label string,
...
+	if err := perms.AuthorizeGroupAccess(ctx, d.env, groupID); err != nil {
+		return nil, err
+	}
+	q := query_builder.NewQuery(`SELECT * FROM APIKeys`)
+	q.AddWhereClause(`user_id = ?`, u.GetUserID())
+	q.AddWhereClause(`group_id = ?`, groupID)
+	q.SetOrderBy("label", true /*ascending*/)
+	queryStr, args := q.Build()
+	var keys []*tables.APIKey
+	err = d.h.DB(ctx).Transaction(func(tx *db.DB) error {
```
vadimberezniker, line 264: Is a transaction needed here? There are no writes involved.

enterprise/server/backends/userdb/userdb.go
```diff
@@ -638,6 +793,13 @@ func (d *UserDB) UpdateGroupUsers(ctx context.Context, groupID string, updates [
 					groupID).Error; err != nil {
 					return err
 				}
+				if err := tx.Exec(`
+						DELETE FROM APIKeys
+						WHERE user_id = ? AND group_id = ?`,
+					update.GetUserId().GetId(),
```
vadimberezniker, line 799: Do group API keys have a NULL or empty user_id value? A malicious user could pass an empty user ID here.

vadimberezniker (changes requested):

enterprise/server/backends/userdb/userdb.go
bduffany: Nope - removed it

enterprise/server/backends/userdb/userdb.go
bduffany: Moved this check into https://github.com/buildbuddy-io/buildbuddy/pull/3380

vadimberezniker (approved):

tylerwilliams (approved):

PR #3282, opened 2023-01-24
Stream logs instead of loading entire logs into memory

tempoz:
Logs were being loaded into memory on both the client and the server, which would be a problem
for very large logs. Instead, we now stream the logs from the blobstore chunks on the server
side and download the stream to disk on the client side.

---

**Version bump**: Minor

### DIFF
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -31,9 +31,10 @@ import InvocationActionCardComponent from "./invocation_action_card";
 import TargetsComponent from "./invocation_targets";
 import { BuildBuddyError } from "../util/errors";
 import UserPreferences from "../preferences/preferences";
-import { eventlog } from "../../proto/eventlog_ts_proto";
+import Long from "long";
 import capabilities from "../capabilities/capabilities";
 import CacheRequestsCardComponent from "./cache_requests_card";
+import rpc_service from "../service/rpc_service";
 
 interface State {
   loading: boolean;
@@ -41,6 +42,7 @@ interface State {
   error: BuildBuddyError | null;
 
   model: InvocationModel;
+  attempt: Long;
 }
 
 interface Props {
@@ -62,6 +64,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
     error: null,
 
     model: new InvocationModel(),
+    attempt: new Long(0),
   };
 
   private timeoutRef: number;
@@ -109,6 +112,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
           inProgress: showInProgressScreen,
           model: InvocationModel.modelFromInvocations(response.invocation as invocation.Invocation[]),
           loading: false,
+          attempt: response.invocation[0].attempt,
         });
         document.title = `${this.state.model.getUser(
           true
@@ -209,16 +213,9 @@ export default class InvocationComponent extends React.Component<Props, State> {
     });
     const isBazelInvocation = this.state.model.isBazelInvocation();
     const fetchBuildLogs = () => {
-      return rpcService.service
-        .getEventLogChunk(
-          new eventlog.GetEventLogChunkRequest({
-            invocationId: this.props.invocationId,
-            minLines: maxInt32, // int32 max value: max number of lines we can request.
-          })
-        )
-        .then((response: eventlog.GetEventLogChunkResponse) => {
-          return new TextDecoder().decode(response.buffer || new Uint8Array());
-        });
+      rpc_service.downloadBytestreamFile("//buildlog", "bytestream://", this.props.invocationId, {
+        attempt: this.state.attempt.toString(),
+      });
     };
 
     const suggestions = getSuggestions({
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -31,7 +31,6 @@ import InvocationActionCardComponent from "./invocation_action_card";
 import TargetsComponent from "./invocation_targets";
 import { BuildBuddyError } from "../util/errors";
 import UserPreferences from "../preferences/preferences";
-import Long from "long";
 import capabilities from "../capabilities/capabilities";
 import CacheRequestsCardComponent from "./cache_requests_card";
 import rpc_service from "../service/rpc_service";
@@ -42,7 +41,6 @@ interface State {
   error: BuildBuddyError | null;
 
   model: InvocationModel;
-  attempt: Long;
 }
 
 interface Props {
@@ -55,7 +53,6 @@ interface Props {
 
 const largePageSize = 100;
 const smallPageSize = 10;
-const maxInt32 = 2147483647;
 
 export default class InvocationComponent extends React.Component<Props, State> {
   state: State = {
@@ -64,7 +61,6 @@ export default class InvocationComponent extends React.Component<Props, State> {
     error: null,
 
     model: new InvocationModel(),
-    attempt: new Long(0),
   };
 
   private timeoutRef: number;
@@ -112,7 +108,6 @@ export default class InvocationComponent extends React.Component<Props, State> {
           inProgress: showInProgressScreen,
           model: InvocationModel.modelFromInvocations(response.invocation as invocation.Invocation[]),
           loading: false,
-          attempt: response.invocation[0].attempt,
         });
         document.title = `${this.state.model.getUser(
           true
@@ -214,7 +209,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
     const isBazelInvocation = this.state.model.isBazelInvocation();
     const fetchBuildLogs = () => {
       rpc_service.downloadBytestreamFile("//buildlog", "bytestream://", this.props.invocationId, {
-        attempt: this.state.attempt.toString(),
+        attempt: Number(this.state.model.invocations[0]?.attempt ?? 0).toString()
       });
     };
 
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -208,9 +208,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
     });
     const isBazelInvocation = this.state.model.isBazelInvocation();
     const fetchBuildLogs = () => {
-      rpc_service.downloadBytestreamFile("//buildlog", "bytestream://", this.props.invocationId, {
-        attempt: Number(this.state.model.invocations[0]?.attempt ?? 0).toString()
-      });
+      rpc_service.downloadLog(this.props.invocationId, Number(this.state.model.invocations[0]?.attempt ?? 0));
     };
 
     const suggestions = getSuggestions({
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -44,6 +44,10 @@ class RpcService {
     return false;
   }
 
+  getDownloadUrl(params: Record<string, string>): string {
+    return `/file/download?${new URLSearchParams(params)}`;
+  }
+
   getBytestreamUrl(
     bytestreamURL: string,
     invocationId: string,
@@ -63,7 +67,16 @@ class RpcService {
         params[k] = additionalParams[k];
       }
     }
-    return `/file/download?${new URLSearchParams(params)}`;
+    return this.getDownloadUrl(params);
+  }
+
+  downloadLog(invocationId: string, attempt: number) {
+    const params: Record<string, string> = {
+      invocation_id: invocationId,
+      attempt: attempt.toString(),
+      artifact: "buildlog",
+    };
+    window.open(this.getDownloadUrl(params));
   }
 
   downloadBytestreamFile(
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -44,7 +44,12 @@ class RpcService {
     return false;
   }
 
-  getBytestreamUrl(bytestreamURL: string, invocationId: string, { filename = "", zip = "" } = {}): string {
+  getBytestreamUrl(
+    bytestreamURL: string,
+    invocationId: string,
+    { filename = "", zip = "" } = {},
+    additionalParams?: Record<string, string>
+  ): string {
     const encodedRequestContext = uint8ArrayToBase64(context.RequestContext.encode(this.requestContext).finish());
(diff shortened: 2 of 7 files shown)


tempoz:
* master
  * **PR #3282** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3282" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3282?utm_source=stack-comment).

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1058,6 +1064,38 @@ func (s *BuildBuddyServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			c := chunkstore.New(
+				s.env.GetBlobstore(),
+				&chunkstore.ChunkstoreOptions{},
+			)
+			attempt := uint64(0)
+			if n, err := strconv.ParseUint(params.Get("attempt"), 10, 64); err == nil {
+				attempt = n
+			}
+			// Stream the file back to our client
+			w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=buildlog_for_invocation_%s", params.Get("invocation_id")))
```
tylerwilliams, line 1081: how bout "filename=invocation-%s.log"?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1058,6 +1064,38 @@ func (s *BuildBuddyServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			w.Header().Set("Content-Type", "application/octet-stream")
+			io.Copy(
```
tylerwilliams, line 1083: io.Copy returns an error -- should we check it and either return or log if it's set?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1058,6 +1064,38 @@ func (s *BuildBuddyServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// TODO(siggisim): Figure out why this JWT is overriding authority auth and remove.
 	ctx := context.WithValue(r.Context(), "x-buildbuddy-jwt", nil)
 
+	if lookup.Filename[0:2] == "//" {
```
tylerwilliams, line 1067: this is slightly panic inducing

image

Could we use strings.HasPrefix / strings.TrimPrefix (below) instead?

app/invocation/invocation.tsx
```diff
@@ -31,16 +31,18 @@ import InvocationActionCardComponent from "./invocation_action_card";
...
 import CacheRequestsCardComponent from "./cache_requests_card";
+import rpc_service from "../service/rpc_service";
 
 interface State {
   loading: boolean;
   inProgress: boolean;
   error: BuildBuddyError | null;
 
   model: InvocationModel;
+  attempt: Long;
```
bduffany, line 45: nit: is it possible to avoid using `Long` here and use `number` instead? I'd like to avoid Long unless we really need it, since it's protobufjs-specific and eventually we might want to migrate away from it since it has a bunch of issues.

app/invocation/invocation.tsx
```diff
@@ -62,6 +64,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
     error: null,
 
     model: new InvocationModel(),
+    attempt: new Long(0),
```
bduffany, line 67: Since `model` already encapsulates the Invocation proto, could we instead add a getter to `InvocationModel` like `getAttempt()` which returns `response.invocation[0]?.attempt || 1`

app/invocation/invocation.tsx
```diff
@@ -209,16 +213,9 @@ export default class InvocationComponent extends React.Component<Props, State> {
...
-        .getEventLogChunk(
-          new eventlog.GetEventLogChunkRequest({
-            invocationId: this.props.invocationId,
-            minLines: maxInt32, // int32 max value: max number of lines we can request.
-          })
-        )
-        .then((response: eventlog.GetEventLogChunkResponse) => {
-          return new TextDecoder().decode(response.buffer || new Uint8Array());
-        });
+      rpc_service.downloadBytestreamFile("//buildlog", "bytestream://", this.props.invocationId, {
```
bduffany, line 216: Since the bytestream API is only used to fetch things from cache (CAS/AC), and build logs are stored in Blobstore (not cache), I think it'd be clearer to add a separate function to RPCService called `downloadBlob` which fetches from Blobstore .

It could still use the same `/file/download` endpoint, but just add a separate URL param like `?blobtype=buildlog` or something like that, which tells the file download handler to fetch the invocation build log from blobstore.

app/invocation/invocation.tsx
```diff
@@ -209,16 +213,9 @@ export default class InvocationComponent extends React.Component<Props, State> {
...
+        attempt: this.state.attempt.toString(),
```
bduffany, line 217: For my own understanding - is the `attempt` needed so that the downloaded logs match the current invocation attempt that is visible in the UI?

server/buildbuddy_server/buildbuddy_server.go
tempoz: haha, absolutely. Great idea.

server/buildbuddy_server/buildbuddy_server.go
tempoz: That's better.

app/terminal/terminal.tsx
```diff
@@ -296,25 +296,8 @@ export default class TerminalComponent extends React.Component<TerminalProps, St
...
-        .fullLogsFetcher()
-        .then(serveLog)
-        .catch((e) => errorService.handleError(e))
-        .finally(() => {
-          this.setState({ isLoadingFullLog: false });
-        });
-      return;
-    }
-    serveLog(this.props.value);
+    event.preventDefault();
```
bduffany, line 299: Regular `<button>` elements don't have a default action when clicked, so the `event.preventDefault()` here is essentially a nop.

(An example where `preventDefault()` would be needed is if you had an `<a href="...">` and you wanted to prevent the browser's default behavior of navigating to the URL, and instead handle the link click with custom application logic.)

server/buildbuddy_server/buildbuddy_server.go
tempoz: Yeah, sounds good; I spent a long time trying to use `http.ServeContent` (which returns nothing) here instead, but it was just serving nothing (still don't really understand why). So when I switched to `io.Copy`, I forgot to start doing error checking :|

app/terminal/terminal.tsx
bduffany, line 299: I think we should keep the default download behavior if `this.props.fullLogsFetcher` is undefined, otherwise this will throw an NPE. This is useful e.g. for downloading test logs (which uses the same Terminal component)

server/backends/chunkstore/chunkstore.go
```diff
@@ -251,6 +251,10 @@ func (r *chunkstoreReader) shiftToFront(p []byte, bytesRead *int) {
 }
 
 func (r *chunkstoreReader) Read(p []byte) (int, error) {
+	return r.read(p)
+}
+
+func (r *chunkstoreReader) read(p []byte) (int, error) {
```
bduffany, line 257: nit: seems like this diff is no longer needed?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -990,6 +992,10 @@ type bsLookup struct {
 }
 
 func getBestFilename(filename, blobname string) string {
+	if strings.HasPrefix(filename, "//") && !strings.HasPrefix(filename, "///") {
+		// This is not a normal file path, let the implementation handle it
+		return filename
+	}
```
bduffany, line 998: This is somewhat redundant w/ my earlier comment re. RPC service, but I think we shouldn't introduce "special" bytestream URIs which actually map to blobstore resources, since blobstore is not accessible via the bytestream API. A separate `blobtype=buildlog` URL param would be cleaner, IMO; that way we are keeping bytestream and blobstore separate.

bduffany (changes requested):

app/invocation/invocation.tsx
tempoz: sgtm

app/invocation/invocation.tsx
tempoz: yup, sounds good.

app/terminal/terminal.tsx
tempoz: Ooooooh, thanks for explaining.

tempoz: sgtm
```suggestion
```

server/backends/chunkstore/chunkstore.go
tempoz: oops, yeah.

```suggestion
```

server/buildbuddy_server/buildbuddy_server.go
tempoz: works for me.

app/invocation/invocation.tsx
tempoz: sure, fine with me.

server/buildbuddy_server/buildbuddy_server.go
tempoz:
```suggestion
```

app/service/rpc_service.ts
```diff
@@ -44,7 +44,16 @@ class RpcService {
     return false;
   }
 
-  getBytestreamUrl(bytestreamURL: string, invocationId: string, { filename = "", zip = "" } = {}): string {
+  getDownloadUrl(params: Record<string, string>): string {
+    return `/file/download?${new URLSearchParams(params)}`;
+  }
+
+  getBytestreamUrl(
```
bduffany, line 51: Can the `additionalParams` be removed from this func and below?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1038,26 +1040,91 @@ func (s *BuildBuddyServer) getAnyAPIKeyForInvocation(ctx context.Context, invoca
...
+
+	if params.Get("artifact") != "" {
+		s.serveArtifact(ctx, w, params)
+		return
+	}
+	if params.Get("bytestream_url") != "" {
+		// bytestream request
+		s.serveBytestream(ctx, w, params)
+		return
+	}
```
bduffany, line 1060:
```suggestion
	}
	http.Error(w, `One of "artifact" or "bytestream_url" query param is required`, http.StatusBadRequest)
```

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1038,26 +1040,91 @@ func (s *BuildBuddyServer) getAnyAPIKeyForInvocation(ctx context.Context, invoca
...
+func (s *BuildBuddyServer) serveBytestream(ctx context.Context, w http.ResponseWriter, params url.Values) {
 	lookup, err := parseByteStreamURL(params.Get("bytestream_url"), params.Get("filename"))
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
 	if lookup.URL.User == nil {
-		apiKey, _ := s.getAnyAPIKeyForInvocation(r.Context(), params.Get("invocation_id"))
+		apiKey, _ := s.getAnyAPIKeyForInvocation(ctx, params.Get("invocation_id"))
```
bduffany, line 1122: Same question here: since the JWT is stripped from the ctx on line 1050 it seems like `getAnyAPIKeyForInvocation` is now being passed an anonymous context?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1038,26 +1040,91 @@ func (s *BuildBuddyServer) getAnyAPIKeyForInvocation(ctx context.Context, invoca
...
+// serveArtifact handles requests that specify particular build artifacts
+func (s *BuildBuddyServer) serveArtifact(ctx context.Context, w http.ResponseWriter, params url.Values) {
+	switch params.Get("artifact") {
+	case "buildlog":
+		iid := params.Get("invocation_id")
+		if iid == "" {
+			log.Warningf("Build log requested with empty invocation id.")
+			http.Error(w, "File not found", http.StatusNotFound)
+		}
+		if _, err := s.env.GetInvocationDB().LookupInvocation(ctx, iid); err != nil {
```
bduffany, line 1072: Does this work for authenticated invocations? The JWT is being nil'd out on line 1050, so at a glance it seems like the ctx here would effectively be an anonymous user context.

server/buildbuddy_server/buildbuddy_server.go
tempoz: So this was working, but I think only because the `auth.claims` Value was still filled out. Good catch, fixed.

server/buildbuddy_server/buildbuddy_server.go
tempoz: Same as above.

app/service/rpc_service.ts
tempoz: Yup, done.

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1038,25 +1040,91 @@ func (s *BuildBuddyServer) getAnyAPIKeyForInvocation(ctx context.Context, invoca
...
+}
+
+// serveArtifact handles requests that specify particular build artifacts
+func (s *BuildBuddyServer) serveArtifact(ctx context.Context, w http.ResponseWriter, params url.Values) {
+	switch params.Get("artifact") {
+	case "buildlog":
+		iid := params.Get("invocation_id")
+		if iid == "" {
+			log.Warningf("Build log requested with empty invocation id.")
+			http.Error(w, "File not found", http.StatusNotFound)
```
bduffany, line 1068:
```suggestion
			http.Error(w, "File not found", http.StatusNotFound)
			return
```

bduffany (approved):
nice! just one nit

PR #3285, opened 2023-01-24
Add `Banner` component

bduffany:
We have a bunch of these components which have an icon on the left-hand side, rounded corners, and light background color:

image

image

image

This next one doesn't have an icon, but an info-level note would make sense here:

image

Extracting a `Banner` component so we can reuse these more easily and make them more consistent.

---

**Version bump**: Patch


siggisim (approved):
Sweet! We'll need a component library soon :p

PR #3286, opened 2023-01-24
Add clickhouse schema for TestTargetStatuses

luluz66:
1. add function to find existing projections from CREATE TABLE statement
2. add test for functions to extract projections
3. add projection in runMigrations if not exist

https://github.com/buildbuddy-io/buildbuddy-internal/issues/1821


server/util/clickhouse/clickhouse.go
```diff
@@ -259,6 +255,112 @@ func (e *Execution) AdditionalFields() []string {
 	}
 }
 
+// TestTargetStatus represents the status of a target, the target info and invocation details
+type TestTargetStatus struct {
```
bduffany, line 259: out of curiosity, why `TestTargetStatus` instead of `TargetStatus` like the mysql table name?

server/util/clickhouse/clickhouse.go
```diff
@@ -259,6 +255,112 @@ func (e *Execution) AdditionalFields() []string {
...
+}
+
+const (
+	beforeCreateBody int = iota
+	inCreateBody
+	inProjection
+	afterCreateBody
+)
+
+// adapted from https://github.com/go-gorm/clickhouse/blob/master/migrator.go
```
bduffany, line 329: nit: add a comment with the context for why we couldn't use GORM's migration code directly? (I guess because it doesn't support migrating PROJECTION ?)

server/util/clickhouse/clickhouse.go
luluz66: This table only includes test targets. I assume that we will have different schemas for build targets.

server/util/clickhouse/clickhouse.go
```diff
@@ -259,6 +255,112 @@ func (e *Execution) AdditionalFields() []string {
 	}
 }
 
+// TestTargetStatus represents the status of a target, the target info and invocation details
+type TestTargetStatus struct {
+	GroupID       string
```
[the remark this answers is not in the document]
luluz66: Done

server/util/clickhouse/clickhouse.go
luluz66: Done.

bduffany (approved):
Exciting :)

luluz66:
I rebased this change on top of clickhouse-refactor instead of master to prevent merge error when I need to tweak the TestTargetStatus schema. 

A few schema changes:
- removed `role` from the sort key because currently when we write TestTargetStatus, the role is always "CI" so we don't need to have this condition in the select query. (The reason we have this condition in MySQL is b/c we want to prune the Invocations table before the Joins)
- added `command` to TestTargetStatus. Currently, command can be "test" or "coverage" in TestTargetStatus table. This allow us to add filters on command in the future.
- Rename `created_at_usec` to `invocation_start_time_usec`. `created_at_usec` is determined right before gorm inserts the Invocation. We can avoid a invocation read if we use the start time usec of the invocation instead.
- Use `invocation_uuid` as the last sort key instead of `created_at_usec`. Because we didn't store `start_time_usec` for invocation, when we backfill TestTargetStatus we have to use invocation `created_at_usec`. As a result, clickhouse won't be able to de-duplicate the entires during rollout because `created_at_usec` is slightly different from the `start_time_usec`.  I have tested this change won't affect the select query.

PR #3287, opened 2023-01-25
Add option to periodically remove old AC entries.

vadimberezniker:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -39,6 +39,8 @@ import (
 	"github.com/elastic/gosigar"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/text/language"
+	"golang.org/x/text/message"
 	"google.golang.org/protobuf/proto"
 
 	rfpb "github.com/buildbuddy-io/buildbuddy/proto/raft"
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -607,6 +650,11 @@ func (p *PebbleCache) scanForBrokenFiles(quitChan chan struct{}) error {
 			continue
 		}
 
+		if time.Since(lastUpdate) > 15*time.Minute {
```
tylerwilliams, line 653: could we make this 1/minute?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -615,21 +663,43 @@ func (p *PebbleCache) scanForBrokenFiles(quitChan chan struct{}) error {
...
+				}
 			}
 		}
 
-		if fileMetadata.GetFileRecord().GetCompressor() == repb.Compressor_IDENTITY {
+		if opts.deleteACEntriesOlderThan != 0 {
+			age := time.Since(time.UnixMicro(fileMetadata.GetLastAccessUsec()))
+			if age > opts.deleteACEntriesOlderThan {
+				if err := db.Delete(fileMetadataKey, pebble.NoSync); err != nil {
+					log.Warningf("Could not delete old AC key %q: %s", string(fileMetadataKey), err)
```
tylerwilliams, line 683: will you hit this if the file was deleted above? (L670ish)

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -615,21 +663,43 @@ func (p *PebbleCache) scanForBrokenFiles(quitChan chan struct{}) error {
...
+					removedEntry = true
+				}
 			}
 		}
 
-		if fileMetadata.GetFileRecord().GetCompressor() == repb.Compressor_IDENTITY {
+		if opts.deleteACEntriesOlderThan != 0 {
+			age := time.Since(time.UnixMicro(fileMetadata.GetLastAccessUsec()))
+			if age > opts.deleteACEntriesOlderThan {
+				if err := db.Delete(fileMetadataKey, pebble.NoSync); err != nil {
```
tylerwilliams, line 682: i'm prolly just missing where it happens -- but how does this code check that it's deleting an AC entry?

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: Done.

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: Err oops.

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: Added an additional check to avoid the unnecessary delete (but Delete would not fail if the key was not present)

tylerwilliams (approved):

PR #3291, opened 2023-01-25
Dedupe firecracker disk image conversion

bduffany:
This is mostly experimental -- if we don't see a lot of successful de-dupes logged in practice, we can revert this code.

---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/1906

### DIFF
--- a/enterprise/server/util/container/container.go
+++ b/enterprise/server/util/container/container.go
@@ -10,23 +10,32 @@ import (
 	"os/user"
 	"path/filepath"
 	"sort"
+	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/container"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/containers/docker"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/ext4"
 	"github.com/buildbuddy-io/buildbuddy/server/util/disk"
+	"github.com/buildbuddy-io/buildbuddy/server/util/hash"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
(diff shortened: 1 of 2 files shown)


enterprise/server/util/container/container.go
```diff
@@ -10,23 +10,32 @@ import (
...
+	"golang.org/x/sync/singleflight"
 
 	dockerclient "github.com/docker/docker/client"
 )
 
 const (
 	diskImageFileName = "containerfs.ext4"
+
+	// Minimum timeout used for background Firecracker disk image conversion.
+	imageConversionTimeout = 15 * time.Minute
```
iain-macdonald, line 31: 15 minutes seems long? Any idea what the p99 is for this right now?

server/util/hash/hash.go
```diff
@@ -7,10 +7,24 @@ import (
...
 func String(input string) string {
 	return fmt.Sprintf("%x", sha256.Sum256([]byte(input)))
 }
 
+// Strings returns a hash of the concatenated string hashes.
+//
+// Note: This is not the same as the hash of the concatenated strings. The
+// returned hash is not necessarily useful as a content hash, but may be useful
+// as a deduping key.
+func Strings(input ...string) string {
```
iain-macdonald, line 20: This will generate hashes with length proportional to the size of the input array, which users of this library might find confusing (and, for some use cases, not performant). Is there a reason to not just concatenate the strings and return the SHA256 of that for now?

iain-macdonald (commented):
Nice!

iain-macdonald (approved):

vadimberezniker (approved):

PR #3293, opened 2023-01-26
Move some API key auth logic into UserDB transactions

bduffany:
- Move `authorizeAPIKeyWrite` from buildbuddy_server into DB transactions in userDB, so that we atomically check perms and then perform the mutate.
- Add a few more tests to userDB
- Update userDB tests to use a more realistic authenticator (based on userDB + authDB instead of a static mapping), so that we're exercising the real auth logic as the DB is mutated throughout the test, rather than needing to manually update the TestAuthenticator's fixed user map after each DB write.

---

**Version bump**: Patch


enterprise/server/backends/userdb/userdb_test.go
```diff
@@ -118,14 +172,8 @@ func TestCreateUser_Cloud_JoinsOnlyDomainGroup(t *testing.T) {
...
-	slug := "org1-url-identifier"
-	orgGroupID, err := udb.InsertOrUpdateGroup(ctx, &tables.Group{
-		URLIdentifier: &slug,
-		OwnedDomain:   "org1.io",
-	})
-	require.NoError(t, err)
-
-	err = udb.InsertUser(ctx, &tables.User{
+	// Create the user US1; this should also create GR1.
+	err := udb.InsertUser(ctx, &tables.User{
```
tempoz, line 176: Can we use your `createUser` function here?

enterprise/server/backends/userdb/userdb_test.go
```diff
@@ -387,32 +448,93 @@ func TestUpdateGroupUsers_Role(t *testing.T) {
...
+
+	createUser(t, ctx, env, "US1", "org1.io")
+	ctx1 := authUserCtx(ctx, env, t, "US1")
+	gr1 := getSelfOwnedGroup(t, ctx1, env)
+
+	k1 := getSingleAPIKey(t, ctx1, env, gr1.Group.GroupID)
+	k1.Label = "US1-Updated-Label"
+	err := udb.UpdateAPIKey(ctx1, k1)
+
+	require.NoError(t, err, "US1 should be able to update their own API key")
```
tempoz, line 529: Should this test to see if a user can delete their own API key? If not, this section should probably just be removed, since it is already tested above in the relatedly-named function.

tempoz (approved):
LGTM over all, just one nit and one suspected copy-paste-o.

tylerwilliams (approved):

enterprise/server/backends/userdb/userdb_test.go
bduffany: Thanks for calling this out! Turns out this PR actually introduced a deadlock in `DeleteUser()` caused by trying to start a new transaction within the existing transaction. Updated this test and fixed the bug.

enterprise/server/backends/userdb/userdb_test.go
bduffany: Yep, also replaced the other `InsertUser` calls throughout this file.

PR #3294, opened 2023-01-26
Clean up `auth.ClaimsFromSubID`

bduffany:
- Remove unused `a` param
- Reorder `ctx, env` params to match the convention we use elsewhere

---

**Version bump**: Patch


tempoz (approved):
LGTM

PR #3295, opened 2023-01-26
Roll-forward #3272: Firecracker: Monitor peak FS usage

bduffany:
Fixes the "file exists" error caused by `/etc/mtab` already existing in some images.

The fix is in this commit: https://github.com/buildbuddy-io/buildbuddy/pull/3295/commits/5bee4595eb35e6250e91740427311c25542bb06b

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/goinit/main.go
+++ b/enterprise/server/cmd/goinit/main.go
@@ -323,7 +323,13 @@ func main() {
 	}
 	die(os.WriteFile("/etc/hosts", []byte(strings.Join(hosts, "\n")), 0755))
 	die(os.WriteFile("/etc/resolv.conf", []byte("nameserver 8.8.8.8"), 0755))
-	die(syscall.Symlink("/proc/mounts", "/etc/mtab"))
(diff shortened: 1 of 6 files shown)


enterprise/server/cmd/goinit/main.go
```diff
@@ -323,6 +323,16 @@ func main() {
...
+	if _, err := os.Stat("/etc/mtab"); err != nil {
+		if os.IsNotExist(err) {
+			die(syscall.Symlink("/proc/mounts", "/etc/mtab"))
+			log.Infof("Wrote /etc/mtab")
+		} else {
+			die(err)
+		}
+	} else {
+		log.Infof("Found existing /etc/mtab")
+	}
```
tylerwilliams, line 335: do we need the log.Infofs here? I'm not even sure where these logs show up -- the console?

bduffany: They show up when running firecracker in debug mode. Got rid of them though since they aren't super useful

tylerwilliams (approved):

PR #3298, opened 2023-01-26
Add a perKeyMutex implementation

tylerwilliams:


server/util/lockmap/lockmap.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return &pkm
+}
+
+func (p *perKeyMutex) gc() {
+	for {
+		p.mutexes.Range(func(key, value any) bool {
+			rcm := value.(*refCountedMutex)
+			if rcm.Val() == 0 {
+				p.bigLock.Lock()
```
bduffany, line 85: if Val() == 0 here but then gets incremented just before bigLock.Lock(), could that cause a race?

server/util/lockmap/lockmap.go
```diff
@@ -0,0 +1,133 @@
...
+// RLock()ed and Lock()ed by different threads, they will share the same
+// RWMutex. The lockmap will cull unused locks periodically to save memory.
+type Locker interface {
+	Lock(key string) func()
+	RLock(key string) func()
+}
+
+type perKeyMutex struct {
+	mutexes sync.Map
+	bigLock *sync.RWMutex
```
vadimberezniker, line 67: nit: these can be value types, there's nothing gained by making them pointers and having the extraneous initialization in New

server/util/lockmap/lockmap.go
```diff
@@ -0,0 +1,133 @@
...
+	defer p.bigLock.Unlock()
+
+	p.once.Do(func() {
+		go p.gc()
+	})
+
+	value, _ := p.mutexes.LoadOrStore(key, newRefCountedMutex())
+	rcm := value.(*refCountedMutex)
+
+	rcm.Lock()
```
vadimberezniker, line 106: I think it would be better if you implemented Lock/Unlock on refCountedMutex and manage the refcounting there instead of requiring additional calls here, but doesn't make a big difference.

server/util/lockmap/lockmap.go
```diff
@@ -0,0 +1,133 @@
...
+		time.Sleep(100 * time.Millisecond)
+	}
+}
+
+func (p *perKeyMutex) Lock(key string) func() {
+	p.bigLock.Lock()
+	defer p.bigLock.Unlock()
+
+	p.once.Do(func() {
+		go p.gc()
```
vadimberezniker, line 100: Any reason not to start the gc as part of New? Seems like it would be fairly harmless even if the map is not actually used.

server/util/lockmap/lockmap.go
vadimberezniker: Right, I don't think this is safe. I don't think can avoid locking the map for gc. You should benchmark it, but I don't think it's going to be prohibitive the lock the map for cleanup.

tylerwilliams: Whoops, moved this out (but still per-key).

vadimberezniker (approved):

PR #3300, opened 2023-01-26
Move more user/group management logic into UserDB transactions

bduffany:
Continuing to move more authorization logic into userdb and strengthen userdb_test in preparation for user-level API keys.

The goal is eventually to have buildbuddy_server just delegate directly to userdb, which is much better tested and runs its business logic inside transactions. This also feels like better separation of concerns, since currently the business logic (auth checks, precondition checks etc.) is sort of split between buildbuddy_server and userdb without much rhyme or reason.

- Instead of the `CreateGroup` RPC calling InsertOrUpdateGroup and then AddUserToGroup, do both of these in a single transaction in userDB, in a new `CreateGroup` method.
- Move the auth check for `UpdateGroup` RPC into `InsertOrUpdateGroup` RPC.
- Add explicit auth checks to `AddUserToGroup` instead of depending on role filter

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/userdb/userdb_test.go
+++ b/enterprise/server/backends/userdb/userdb_test.go
@@ -73,7 +73,7 @@ func getSelfOwnedGroup(t *testing.T, ctx context.Context, env environment.Env) *
 	return nil
 }
 
-func ptrTo[T any](val T) *T {
+func stringPointer(val string) *string {
 	return &val
 }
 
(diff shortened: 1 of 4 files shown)


enterprise/server/backends/userdb/userdb_test.go
```diff
@@ -73,6 +73,10 @@ func getSelfOwnedGroup(t *testing.T, ctx context.Context, env environment.Env) *
 	return nil
 }
 
+func ptrTo[T any](val T) *T {
```
tylerwilliams, line 76: maybe just stringPointer for clarity?

tylerwilliams (commented):
one nit but otherwise LGTM -- do you want to submit post-release so we give this more soak time in dev?

enterprise/server/backends/userdb/userdb_test.go
bduffany: Done

bduffany:
> one nit but otherwise LGTM -- do you want to submit post-release so we give this more soak time in dev?

sgtm!

tylerwilliams (approved):

PR #3309, opened 2023-01-28
Add a LockID() string method to PebbleKey and use it

tylerwilliams:
This changes the pebble lock-key from a raw fileMetadataKey to a string containing:

AC: cache type, remote instance name, digest
CAS: cache type, digest

### DIFF
--- a/enterprise/server/raft/filestore/filestore.go
+++ b/enterprise/server/raft/filestore/filestore.go
@@ -98,6 +98,13 @@ func (pmk *PebbleKey) String() string {
 	return string(fmk)
 }
 
+func (pmk *PebbleKey) LockID() string {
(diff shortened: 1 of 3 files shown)


enterprise/server/raft/filestore/filestore.go
```diff
@@ -98,6 +98,13 @@ func (pmk *PebbleKey) String() string {
 	return string(fmk)
 }
 
+func (pmk *PebbleKey) LockID() string {
+	if pmk.isolation == "ac" {
+		return filepath.Join(pmk.isolation, pmk.remoteInstanceHash, pmk.hash)
```
vadimberezniker, line 103: Do you need group ID in here as well?

vadimberezniker (approved):

PR #3310, opened 2023-01-28
Add version read/write functions to manage db version

tylerwilliams:
Adds a version metadata proto and utility functions to read, write and update it safely in the db.


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -401,6 +406,105 @@ func olderThanThreshold(t time.Time, threshold time.Duration) bool {
...
+	return versionMetadata, nil
+}
+
+// currentDatabaseVersion returns the currently stored filestore.PebbleKeyVersion.
+// It is safe to call this function in a loop -- the underlying metadata will
+// only be fetched a max of once per second.
+func (p *PebbleCache) currentDatabaseVersion() (filestore.PebbleKeyVersion, error) {
+	unlockFn := p.locker.RLock(string(p.databaseVersionKey()))
+
+	if time.Since(p.lastDBVersionUpdate) > time.Second {
```
vadimberezniker, line 450: Is it not sufficient to read the value once on startup and then rely on the current value is in p.lastDBVersion? Under what circumstances would the database version change under us in the DB?

vadimberezniker (approved):

PR #3314, opened 2023-01-30
Implement UI for user-level API keys

bduffany:
Setup changes:
* Add "(personal key)" label to the dropdown
* Don't show the API key value in the dropdown (since it was getting too long)

---

image

---

Org settings:
* Add checkbox to enable (code is commented out for now, until we're ready to launch).
* Unchecking the checkbox shows a warning banner indicating that the change will deactivate any existing keys.

---

image

---

Settings:
* Add tabs
* Render personal keys page

---

image

---

image (1)

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/org/org_form.tsx
+++ b/enterprise/app/org/org_form.tsx
@@ -190,6 +190,24 @@ export default abstract class OrgForm<T extends GroupRequest> extends React.Comp
             <span>Default to self-hosted executors</span>
           </label>
         )}
+        {capabilities.config.userOwnedKeysEnabled && (
+          <label className="form-row input-label">
+            <input
+              autoComplete="off"
+              onFocus={this.onFocus.bind(this)}
+              onChange={this.onChange.bind(this)}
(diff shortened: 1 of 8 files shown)


enterprise/app/org/org_form.tsx
```diff
@@ -190,6 +190,23 @@ export default abstract class OrgForm<T extends GroupRequest> extends React.Comp
             <span>Default to self-hosted executors</span>
           </label>
         )}
+        {/* <label className="form-row input-label">
```
siggisim, line 193: Instead of commenting, what do you think about putting this all behind a flag so it can be enabled / disabled more easily?

siggisim (approved):

enterprise/app/org/org_form.tsx
bduffany: Done! `--app.user_owned_keys_enabled`

siggisim (approved):

PR #3316, opened 2023-01-30
Fix hash (thanks github) DO NOT SUBMIT

tylerwilliams:


WORKSPACE
```diff
@@ -155,7 +155,7 @@ _go_image_repos()
 
 http_archive(
     name = "rules_foreign_cc",
-    sha256 = "e14a159c452a68a97a7c59fa458033cc91edb8224516295b047a95555140af5f",
+    sha256 = "2c65080689198deca8757930f5cfb7999ba4f4bb62ce5f568b3288e199bf8dcc",
```
brentleyjones, line 158: This will probably get reverted back shortly, but we might as well fix our stuff for a bit?

tylerwilliams: Yeah I assumed it would, just wanted to see if it was possible to patch around this. Not even sure we can yet...

brentleyjones: True, it might be different based on different CDN nodes, or it might just be random.

PR #3318, opened 2023-01-30
Add script to query app liveness for canary

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/enterprise/tools/check_metrics/check_metrics.go
+++ b/enterprise/tools/check_metrics/check_metrics.go
@@ -12,6 +12,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
 	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil/yaml"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/prometheus/client_golang/api"
 	"github.com/prometheus/common/model"
 	"golang.org/x/sync/errgroup"
@@ -71,7 +72,7 @@ func main() {
 				case <-monitoringTimer:
 					return nil
 				case <-monitoringTicker.C:
-					healthy, err := metricHealthy(promAPI, metric.Name, metric.CanaryQuery, metric.BaselineQuery, metric.HealthThreshold)
+					healthy, err := metricHealthy(promAPI, metric.Name, metric.Query, metric.HealthThreshold)
 					if err != nil {
 						log.Warningf("Error querying metric %s: %s", metric.Name, err)
 					}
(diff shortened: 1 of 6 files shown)


maggie-lou:
* master
  * **PR #3318** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3318" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3318?utm_source=stack-comment).

tylerwilliams:
mentioned on slack but I think we should find a more standard way to specify the timeouts -- i sent you some options but lmk if you want to chat about this

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	prometheusClient, err := api.NewClient(api.Config{
+		Address: config.PrometheusAddress,
+	})
+	if err != nil {
+		fmt.Printf("Error creating prometheus client: %v\nDid you run the prometheus port-forward command?", err)
+		os.Exit(1)
+	}
+	promAPI := v1.NewAPI(prometheusClient)
+
+	eg := &errgroup.Group{}
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+					// Poll the metrics to make sure canary looks healthy
+					sustainedUnhealthy, err := canarySustainedUnhealthy(
+						promAPI, config.UnhealthyMonitoringTimeframeSeconds, metric.Name, metric.CanaryQuery, metric.BaselineQuery, metric.CanaryHealthThreshold)
+					if err != nil {
+						log.Fatalf("Error querying canary health: %s\nDid you run the prometheus port-forward command?", err)
+					}
+					if sustainedUnhealthy {
+						log.Warningf("Metrics unhealthy for %s. Exiting script.", metric.Name)
+						os.Exit(metricsUnhealthyExitCode)
+					}
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+				default:
+					if scriptDurationComplete {
+						return nil
+					}
+				}
+			}
+		})
+	}
+
+	eg.Wait()
```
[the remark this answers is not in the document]
maggie-lou: Done

maggie-lou:
> mentioned on slack but I think we should find a more standard way to specify the timeouts -- i sent you some options but lmk if you want to chat about this

Refactored to use util.Retry -  let me know if that addresses your concerns

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	err = log.Configure()
+	if err != nil {
+		log.Fatalf("Could not configure logger: %s", err)
+	}
+	prometheusClient, err := api.NewClient(api.Config{
+		Address: config.PrometheusAddress,
+	})
+	if err != nil {
+		log.Fatalf("Error creating prometheus client: %v\nDid you run the prometheus port-forward command?", err)
```
tylerwilliams, line 54: nit: you can format errors with "%s", doesn't need to be %v

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	promAPI := v1.NewAPI(prometheusClient)
+
+	eg, gctx := errgroup.WithContext(context.Background())
+	for _, metric := range config.PrometheusMetrics {
```
tylerwilliams, line 59: nice; this is much simpler!

tylerwilliams (changes requested):
LGTM with a few small tweaks

enterprise/tools/check_metrics/check_metrics.go
maggie-lou: Done

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	// If all metrics look healthy for the duration of the monitoring timeframe, continue with rollout
+	os.Exit(metricsHealthyExitCode)
+}
+
+// Returns whether the given metric is healthy for the canary
+// If the canary's success rate is lower than that of the other apps by the input threshold, this will return false
+// Otherwise will return true
+func canaryHealthy(v1api v1.API, metricName string, canaryQuery string, baselineQuery string, canaryHealthThreshold float64) (bool, error) {
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return false, err
+	}
+	baselineResultVector := baselineResult.(model.Vector)
+	baselineMetricValue := model.SampleValue(0)
+	if len(baselineResultVector) > 0 {
+		baselineMetricValue = baselineResultVector[0].Value
+	}
+
+	// If the metric differs from the baseline by the set threshold, the metric is considered unhealthy
```
bduffany, line 129: since we're doing Abs() here could we get into a weird state where the canary is actually more healthy than the other apps and we're reporting it as unhealthy?

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,134 @@
...
+func metricHealthy(v1api v1.API, metricName string, query string, baselineQuery string, healthThreshold float64) (bool, error) {
+	ctx := context.Background()
+	log.Debugf("Querying metric %s", metricName)
+
+	result, _, err := v1api.Query(ctx, query, time.Now())
+	if err != nil {
+		return false, err
+	}
+	resultVector := result.(model.Vector)
+	metricValue := model.SampleValue(0)
```
bduffany, line 114: sanity check, does the 0'th value correspond to the latest value?

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,134 @@
...
+						log.Warningf("Error querying metric %s: %s", metric.Name, err)
+					}
+
+					if healthy {
+						unhealthyCount = 0
+					} else {
+						unhealthyCount++
+					}
+
+					if unhealthyCount >= metric.MaxUnhealthyCount {
```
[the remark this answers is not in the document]
maggie-lou: Personally representing it as a count is more intuitive to me than a percentage. I don't necessarily tie the two values together. For example, if the liveness check is down for 5 minutes, that is a clear signal to me we need a rollback because 5 minutes is a long time for an app to be down, not because that's 50% of the total duration I want to monitor the metrics. 

If it sounds good to you we can leave it as is for now and refactor if people are finding it confusing to configure?

enterprise/tools/check_metrics/check_metrics.go
maggie-lou: The API only returns one value for the current time. If there's data, the vector has size 1. Otherwise it's empty. This line `model.SampleValue(0)` is in case no data is returned, which happens for some of our counter metrics

enterprise/tools/check_metrics/check_metrics.go
maggie-lou: Yeah... I thought about that too. I guess that could happen if we fix an error, so the canary has fewer GRPC errors than the other apps for example

Tried to clarify this, but let me know if I just made it more confusing

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil/yaml"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/prometheus/client_golang/api"
+	"github.com/prometheus/common/model"
+	"golang.org/x/sync/errgroup"
+
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/tools/check_metrics/check_metrics.go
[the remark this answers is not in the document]
maggie-lou: Actually would it be more clear if it was MaxUnhealthyTimeframeSeconds instead of MaxUnhealthyCount?

enterprise/tools/check_metrics/check_metrics.go
maggie-lou: Refactored the way to specify health thresholds. Hopefully it's a lot more clearer now

tylerwilliams (approved):

PR #3320, opened 2023-01-31
Create stumps for new heatmap and drilldown RPCs

jdhollen:


jdhollen:
I am open to / still thinking of splitting out a completely separate RPC for executions--I'd still be able to share a bunch of code under the hood, but it would theoretically make it easier to divert logic in the appropriate places (selecting breakdowns, filters, etc.).  Any thoughts?  Could easily be changed later, as well.

proto/invocation.proto
```diff
@@ -532,3 +541,75 @@ message GetTrendResponse {
...
+  int64 max = 3;
+}
+
+// Fetches a heatmap for the requested metric--a heatmap is basically a set of
+// columns, where each column represents a histogram for that metric in a window
+// of time.
+message GetStatHeatmapRequest {
+  context.RequestContext request_context = 1;
+  TrendQuery query = 2;
+  Metric metric = 3;
```
luluz66, line 564: MetricType to be clearer. My first reaction without looking up is that it's a message that include different metrics.

proto/invocation.proto
```diff
@@ -532,3 +541,75 @@ message GetTrendResponse {
...
+  int64 num_metric_buckets = 4;
+}
+
+message HeatmapColumn {
+  string timestamp = 1;
```
luluz66, line 569: why not an int64?

proto/invocation.proto
jdhollen: Sure yeah, done.

proto/invocation.proto
jdhollen: I was playing along with other trends RPCs, and using dates makes it easier to write the query, but being able to bucket into windows smaller than days would be nice and it seems like we have everything in place to just use an int64 here, so sure.

proto/invocation.proto
```diff
@@ -532,3 +541,75 @@ message GetTrendResponse {
...
+}
+
+message GetStatDrilldownResponse {
+  context.ResponseContext response_context = 1;
+  // Each chart shows a drilldown based on a different aggregation type.
+  repeated DrilldownChart chart = 2;
+  // Total fields that can be used to render percentages, if that's desirable.
+  int64 total_in_base = 3;
+  int64 total_in_selection = 4;
+}
```
tylerwilliams, line 615: It doesn't look like these new protos need much from invocation -- wdyt about making a "stat{s}.proto" and putting all this new heatmap jazz in there?

jdhollen: The dependency graph here gets a little fun: I'm depending on the `TrendQuery` proto which seems worth sharing, but it's reasonable to argue that that should move over to `stats` as well and probably wouldn't be too gnarly to refactor with this change.

I'm inclined to just pull over everything that I'm using over to here at first, and then refactor the rest separately?

Some, like `AggType` don't even have an obvious home--I might pull that out with `StatFilter` into yet another proto package called `filter` or something like that?

jdhollen: I guess I forgot to say that if I'm going to make `stats.proto` then I think it makes sense to eventually pull all of the trends requests over to there too, just don't want to make this too too gnarly.

luluz66 (approved):

proto/invocation.proto
jdhollen: Here's what I came up with--moved trends protos into `stats.proto`, moved stuff that still needs to be depended on by `invocation.proto` into `stat_filter.proto`, updated a boatload of references (which should all be harmless since the field numbers and types are unchanged).  What do you all think?

luluz66 (approved):

tylerwilliams (approved):

PR #3322, opened 2023-01-31
Use a local propose for range leases.

vadimberezniker:
A successful local propose guarantees that the change, and any prior writes, have been applied to the local replica.

---

**Version bump**: None


tylerwilliams (approved):
Sweet

PR #3331, opened 2023-02-01
Fix webdriver test

maggie-lou:
For prod and dev targets, the builds may have cached data from previous runs

### DIFF
--- a/enterprise/server/test/webdriver/invocation/invocation_test.go
+++ b/enterprise/server/test/webdriver/invocation/invocation_test.go
@@ -56,7 +56,6 @@ func TestAuthenticatedInvocation_CacheEnabled(t *testing.T) {
 	cacheRequestsCard := wt.FindByDebugID("cache-results-table").Text()
 	assert.Contains(t, cacheRequestsCard, "Write")
 	assert.NotContains(t, cacheRequestsCard, "Hit")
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3331** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3331" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3331?utm_source=stack-comment).

siggisim (approved):
I'll approve if this is blocking, but I'll let @bduffany decide if we want to test this caching stuff in a different way (or maybe we should do something like `--noremote_accept_cached` on the first build in dev/prod.

bduffany:
`--noremote_accept_cached` seems reasonable; we could also change the test to use a unique `--remote_instance_name` (e.g. generate it as a UUID at the start of the test) and then inspect only AC results in the scorecard.

maggie-lou:
> I'll approve if this is blocking, but I'll let @bduffany decide if we want to test this caching stuff in a different way (or maybe we should do something like `--noremote_accept_cached` on the first build in dev/prod.

Good idea - added

siggisim (approved):

bduffany (approved):

PR #3337, opened 2023-02-02
CLI: improve login command

bduffany:
* Use the new `/settings/cli-login` route (which redirects to the user-level keys page if enabled, else org API keys) instead of `/settings/org/api-keys`
* Automatically open the URL in the browser
* Store the config in `.git/config`, under `buildbuddy.api-key`, so that the API key is repo-specific.

---

**Version bump**: Patch

### DIFF
--- a/cli/cmd/bb/bb.go
+++ b/cli/cmd/bb/bb.go
@@ -74,8 +74,10 @@ func run() (exitCode int, err error) {
 	if err != nil || exitCode >= 0 {
 		return exitCode, err
 	}
-	// TODO: Convert this to (exitCode, err) convention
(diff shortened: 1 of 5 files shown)


bduffany:
@siggisim I was on the fence about creating a `.buildbuddy` dir and adding it to `.gitignore`. For remote-bazel, we use a config section in `.git/config` to store the preferred git remote, e.g.

```
[buildbuddy]
    remote-bazel-remote-name = fork
```

do you think it makes sense to do the same thing here? e.g.

```
[buildbuddy]
    api-key = ...
```

siggisim (approved):

PR #3338, opened 2023-02-02
CLI: De-flake integration test

bduffany:
Failed consistently with `--runs_per_test=100` before; now it passes consistently.

---

**Version bump**: Patch


siggisim (approved):
Should we file an issue for this so we don't forget to fix?

bduffany:
> Should we file an issue for this so we don't forget to fix?

I don't think it's worth fixing unless someone wants to repurpose this code for something else. (Seems unlikely, but in case it does happen, I just wanted to leave that note as a warning)

PR #3339, opened 2023-02-02
Make TestLRU actually work

tylerwilliams:


enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -1219,6 +1220,7 @@ func TestLRU(t *testing.T) {
 
 	require.LessOrEqual(t, len(evictionsByQuartile[1]), len(evictionsByQuartile[2]))
 	require.LessOrEqual(t, len(evictionsByQuartile[2]), len(evictionsByQuartile[3]))
+	require.Greater(t, len(evictionsByQuartile[3]), 0)
 }
 
 func TestLRU_IsolateByGroupIDs(t *testing.T) {
```
maggie-lou, line 1226: We can just delete this one now that isolate by group ids is always true

maggie-lou (approved):

enterprise/server/backends/pebble_cache/pebble_cache_test.go
tylerwilliams: Done.

PR #3340, opened 2023-02-02
Randomly sample by constructing a key with the specified digest

tylerwilliams:
This changes the randomsample() function so it generates sample keys by constructing a valid PebbleKey and then serializing keyBytes at the current database version. This will allow random sampling to work correctly across database versions.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -165,8 +165,7 @@ type PebbleCache struct {
 	atimeBufferSize      int
 	minEvictionAge       time.Duration
 
-	lastDBVersionUpdate time.Time
-	lastDBVersion       filestore.PebbleKeyVersion
+	lastDBVersion filestore.PebbleKeyVersion
 
 	env    environment.Env
 	db     *pebble.DB
@@ -378,7 +377,7 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 			if err := disk.EnsureDirectoryExists(blobDir); err != nil {
 				return err
 			}
-			pe, err := newPartitionEvictor(part, pc.fileStorer, blobDir, pc.leaser, pc.locker, pc.accesses, *opts.AtimeBufferSize, *opts.MinEvictionAge, opts.Name)
+			pe, err := newPartitionEvictor(part, pc.fileStorer, blobDir, pc.leaser, pc.locker, pc, pc.accesses, *opts.AtimeBufferSize, *opts.MinEvictionAge, opts.Name)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1574,6 +1579,7 @@ func newPartitionEvictor(part disk.Partition, fileStorer filestore.Store, blobDi
 	pe.sizeBytes = sizeBytes
 	pe.casCount = casCount
 	pe.acCount = acCount
+	pe.lru.UpdateSizeBytes(sizeBytes)
```
tylerwilliams, line 1582: this is a tiny fix -- previously we were not updating the lru size until a write happened -- now we also update it when reading the cached size from the db (or if it was force recomputed).

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -165,8 +165,7 @@ type PebbleCache struct {
 	atimeBufferSize      int
 	minEvictionAge       time.Duration
 
-	lastDBVersionUpdate time.Time
-	lastDBVersion       filestore.PebbleKeyVersion
+	lastDBVersion filestore.PebbleKeyVersion
```
tylerwilliams, line 168: The lastDBVersionUpdate was a vestige of something removed in a previous CL and unused, so I removed it.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -378,7 +377,7 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 			if err := disk.EnsureDirectoryExists(blobDir); err != nil {
 				return err
 			}
-			pe, err := newPartitionEvictor(part, pc.fileStorer, blobDir, pc.leaser, pc.locker, pc.accesses, *opts.AtimeBufferSize, *opts.MinEvictionAge, opts.Name)
+			pe, err := newPartitionEvictor(part, pc.fileStorer, blobDir, pc.leaser, pc.locker, pc, pc.accesses, *opts.AtimeBufferSize, *opts.MinEvictionAge, opts.Name)
```
maggie-lou, line 380: nit: Don't feel strongly about this but would it be cleaner to just pass in the pebble cache as an argument, as opposed to passing it in as a versionGetter interface + pass in a bunch of its other fields individually?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1771,16 +1777,30 @@ func (e *partitionEvictor) Statusz(ctx context.Context) string {
...
 var digestChars = []byte("abcdef1234567890")
 
-func (e *partitionEvictor) randomKey(n int) []byte {
-	partID := e.partitionKeyPrefix()
-	cacheType := "/cas/"
-	buf := bytes.NewBuffer(make([]byte, 0, len(partID)+len(cacheType)+n))
-	buf.Write([]byte(partID))
-	buf.Write([]byte(cacheType))
-	for i := 0; i < n; i++ {
+func (e *partitionEvictor) randomKey(digestLength int) ([]byte, error) {
```
maggie-lou, line 1780: nit: Should this be a constant

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: Yeah this bothered me too -- but pebble_cache holding evictors and evictors needing a reference to pebble cache seemed kinda gross. I'll prolly fix this in a followup.

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: It's this way to allow us to support multiple digests in the future.

PR #3343, opened 2023-02-02
Allow visually debugging webdriver tests locally

bduffany:
When running bazel tests with `--config=webdriver-debug`, tests will now spawn the GUI for chrome (instead of running in `--headless` mode), so you can see what the test is doing.

If the test fails, the browser window will stay open for 3s so you can see what state the app is in when the failure happened. This duration can be extended further so you can go in and check devtools etc., by setting `--test_arg=-webdriver_end_of_test_delay=1h` (this is super rudimentary for now; there is probably a better way to do this).

Also when running under this debug config, pass `--verbose` to chromedriver, so that if chrome crashes on startup, the root cause will be logged (by default, it logs a super opaque error just saying that chrome has crashed).

Tangentially related: this PR also updates the Go repo version for `rules_webtesting` to match the `http_archive` version in /WORKSPACE (the Go repo was behind the WORKSPACE repo due to https://github.com/bazelbuild/rules_webtesting/issues/452)

https://user-images.githubusercontent.com/2414826/216427609-9259ac88-e183-4eb5-a74a-fab2dbe84be3.mp4

---

**Version bump**: Patch

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -128,6 +128,15 @@ build --enable_runfiles
 test --test_tag_filters=-docker
 build --build_tag_filters=-docker
 
+# Run Webdriver tests with --config=webdriver-debug to debug webdriver tests locally.
+# See server/testutil/webtester/webtester.go for more details.
+test:webdriver-debug --test_arg=-webdriver_debug
+# Forward X server display for local webdriver tests.
+test:webdriver-debug --test_env=DISPLAY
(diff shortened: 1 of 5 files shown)


maggie-lou (approved):
Nice! I kept manually adding support for this while developing the webdriver tests - will be much nicer to have a config

PR #3351, opened 2023-02-03
Make docs site more visually appealing

siggisim:
Mostly css changes. DocSide bar is "[swizzled](https://docusaurus.io/docs/swizzling)" and only slightly modified to remove a class.

---

**Version bump**: None

PR #3353, opened 2023-02-03
Add thin-scrollbar class to docs sidebar

bduffany:
context: https://github.com/buildbuddy-io/buildbuddy/pull/3351#discussion_r1095853234

---

**Version bump**: Patch


siggisim (commented):
Lol this is the only thing I removed when swizzling - because it made the scrollbar on mac always visible (instead of hiding when not interacted).

Is there some way we could add this to just non-mac?

bduffany:
> Is there some way we could add this to just non-mac?

done

siggisim (approved):

PR #3354, opened 2023-02-03
Refactor user-owned keys API with user-specific endpoints

bduffany:
* Add separate RPCs (and stub impls) for user-level API keys. The Request/Response protos are exactly the same as the corresponding group-level RPCs.
* Remove `user_id` field that was added previously, since the user API keys API will always use the authenticated user ID.
* Add 2 new UserDB methods (and stub impls): GetUserAPIKeys and CreateUserAPIKey. These operations are scoped to the authenticated user ID.
* Keep Update/Delete DB methods the same, since those operate in API key ID scope, rather than user/group scope. We use generic `perms`/`role`-based auth for those, so the logic will be exactly the same for user-owned and group-owned keys.
* Add docs to all API keys methods of UserDB.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -227,6 +227,14 @@ func (d *UserDB) CreateAPIKey(ctx context.Context, groupID string, label string,
 	return createAPIKey(d.h.DB(ctx), groupID, newAPIKeyToken(), label, caps, visibleToDevelopers)
 }
 
+func (d *UserDB) GetUserAPIKeys(ctx context.Context, groupID string) ([]*tables.APIKey, error) {
+	return nil, status.UnimplementedError("Not implemented")
+}
+
+func (d *UserDB) CreateUserAPIKey(ctx context.Context, groupID, label string) (*tables.APIKey, error) {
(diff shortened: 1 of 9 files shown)


enterprise/server/backends/userdb/userdb.go
```diff
@@ -164,15 +164,15 @@ func (d *UserDB) getAPIKey(tx *db.DB, apiKeyID string) (*tables.APIKey, error) {
 	return key, nil
 }
 
-func (d *UserDB) GetAPIKeys(ctx context.Context, groupID string, checkVisibility bool) ([]*tables.APIKey, error) {
+func (d *UserDB) GetAPIKeys(ctx context.Context, groupID string, authorizeRole bool) ([]*tables.APIKey, error) {
```
tylerwilliams, line 167: can you add a comment here about when authorizeRole should be set / why?

bduffany: @tylerwilliams Split out https://github.com/buildbuddy-io/buildbuddy/pull/3357/files, PTAL

bduffany: Rebased this PR, should be good to go

tylerwilliams (approved):

PR #3363, opened 2023-02-03
Allows cli plugins to see and modify exec args

tempoz:
In the same way that plugins could see and modify bazel args, they can now see and modify
args that are being passed to the executable being run by bazel (arguments after the `--`,
if present).

Additionally, adds the plugin `config-per-target` which leverages this new feature.

---

**Version bump**:  Minor


tempoz:
* master
  * **PR #3363** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3363" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3363?utm_source=stack-comment).

cli/plugins/config-per-target/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Install this plugin with:
+
+```
+bb install buildbuddy-io/plugins:config-per-target
+```
+
+## Configuration
+
+Create a `config.json` file in the plugin directory. The JSON should specify a
```
bduffany, line 19: The plugin directory will be under `~/.cache` when installed like `bb install buildbuddy-io/plugins:config-per-target` on Linux. When the user clears their cache (e.g. to reclaim disk space) we don't want to blow away this config.

What do you think about putting this in `$BUILD_WORKSPACE_DIRECTORY/.config-per-target.json` ? That way it's also clear that the config only applies to the current Bazel workspace. (The downside is that it requires modifying .gitignore)

We could also put it in `$USER_CONFIG_DIR/config-per-target/config.json` but it should probably be namespaced by repo somehow, e.g. 

```
{
  "/path/to/workspace1": {
    "//target1": ["..."],
  }
}
```

cli/arg/arg.go
```diff
@@ -98,8 +98,9 @@ func GetCommandAndIndex(args []string) (string, int) {
 	return "", -1
 }
 
-// Returns any "passthrough" arugments in args (those to the right of the first " -- ", if any)
-func GetPassthroughArgs(args []string) []string {
+// Returns any arugments to be passed to the executable generated by bazel in
```
bduffany, line 101:
```suggestion
// Returns any arguments to be passed to the executable generated by bazel in
```

bduffany (commented):
Nice :)

Could you update `cli_test.go`? I would probably update `TestBazelBuildWithLocalPlugin` and change `build` to `run`, and throw in a BUILD file with an `sh_binary` target in there that prints out its args, and you could have the existing pre_bazel plugin append an arg that we then expect to get printed out.

buildbuddy.yaml
```diff
@@ -62,5 +62,6 @@ plugins:
   - path: cli/plugins/go-deps
   - path: cli/plugins/open-invocation
   - path: cli/plugins/notify
+  - path: cli/plugins/config-per-target
```
siggisim, line 65: Sorry for the drive-by, but the target level `args` parameter just be used for this? For example:
https://github.com/buildbuddy-io/buildbuddy/blob/master/server/cmd/buildbuddy/BUILD#L9

I'm not sure it's worth complicating the plugin API for this, i.e. making plugins take two arguments parameters instead of one.

tempoz: The functionality was supposed to be a natural extension of the current plugin `pre-bazel` functionality; I, at least, was (very) surprised that the plugins could not modify the "exec args", basically, and I wanted to be able to write a plugin that can. I would like to be able to create configurations on a per-target basis that do not need to be written to BUILD files and thus do not need to be included in the git repository. I was using `tools/bazel` to do this, but it was comparatively brittle.

cli/plugins/config-per-target/README.md
tempoz: Works for me! Also, this is the kind of file I just put in `.git/info/exclude`, honestly. I actually already have :P

docs/cli-plugins.md
```diff
@@ -117,15 +117,17 @@ path/to/plugin/
...
 The `pre_bazel.sh` script will be called before Bazel is run.
 
-It is called with a single argument, which is the path to a file
+It is called with two arguments, which are the paths to files
 containing all arguments that will be passed to Bazel (including the
 arguments specified in your `.bazelrc` and expanded via any `--config=`
-flags). **Each line in this file contains a single argument.**
+flags) and all the arguments that will be passed to the executable
+generated by bazel (the arguments after the `--`), if any. **Each line
+in this file contains a single argument.**
```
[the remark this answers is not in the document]
tempoz: Absolutely. Definitely happy to rephrase for clarity.

tempoz:
This now uses (and documents) the EXEC_ARGS_FILE environment variable.

siggisim (approved):
This looks great

cli/test/integration/cli/cli_test.go
```diff
@@ -77,6 +77,59 @@ func TestBazelBuildWithLocalPlugin(t *testing.T) {
...
+	})
+	testfs.MakeExecutable(t, ws, "echo.sh")
+
+	testfs.WriteAllFileContents(t, ws, map[string]string{
+		"plugins/test/pre_bazel.sh": `
+			echo 'Hello from pre_bazel.sh!'
+			if grep '\--build_metadata=FOO=bar' "$1" >/dev/null ; then
+				echo "--build_metadata FOO=bar was canonicalized as expected!"
+			fi
+			echo "" >> $EXEC_ARGS_FILE
```
bduffany, line 95: uber nit: could this line be deleted?

docs/cli-plugins.md
```diff
@@ -340,6 +340,36 @@ config dir, if needed, using something like
...
 different versions of your plugin.
 
+#### $EXEC_ARGS_FILE
+
+This is the path of a file that contains the args that would be passed to an
+executable built by bazel as a result of a `bazel run` command. Specifically,
+these are any positional arguments remaining after canonicalization of any
+options that take arguments into "--option=value" options, excepting the `run`
+subcommand itself and the build target. These are generally the arguments
+following a `--` in the argument list passed to bazel, if any.
```
bduffany, line 350: nit: I think it's a bit confusing to mention canonicalization here, since it doesn't change the args after `--`. wdyt about removing that sentence?

```suggestion
executable built by bazel as a result of a `bazel run` command. These are the arguments
following a `--` in the argument list passed to bazel, if any.
```

docs/cli-plugins.md
```diff
@@ -340,6 +340,36 @@ config dir, if needed, using something like
...
+
+This environment variable will only be set for the `pre_bazel` script in the
```
bduffany, line 352:
```suggestion
This environment variable will only be set for the `pre_bazel.sh` script in the
```

docs/cli-plugins.md
```diff
@@ -340,6 +340,36 @@ config dir, if needed, using something like
...
+plugin. Similar to the way in which the bazel args file works, any changes
+made to this file by a plugin will be read in and reflected in the args
+eventually passed to bazel.
```
bduffany, line 355:
```suggestion
plugin. Plugins can change this file to change the arguments
passed to Bazel.
```

bduffany (approved):
Nice :) just a few nits

docs/cli-plugins.md
tempoz: So, technically, all `--` does is mark the following arguments as positional args (as opposed to options), and what bazel actually does is pass any positional args following the `run` subcommand and a single target (the first two positional arguments) which do not have a corresponding option to the resulting executable. People normally use `--` because some of the args they want to pass will begin with `-` or `--` (because they want the resulting executable to interpret them as options), and without the `--` bazel would try to interpret them as options instead of parsing them as positional args and forwarding them to the resulting executable.

It is technically valid, for instance, to call `bazel run -- //server --help`, and it will print the help text for the server. On testing and code inspection, though, it seems the buildbuddy cli is actually just partitioning on the `--` to separate these two sets of arguments (for the purposes of the files with lists of arguments to be edited), which is not really correct.

I have added a test to demonstrate this behavior, and a `TODO` about fixing it.

tempoz: Note that bazel interprets the next argument after an option that takes a positional argument as a value for that option, even if that argument is `--`, so there is no need to worry about positional arguments *after* a `--` potentially being values for a bazel option.

bduffany: Ah, TIL. SG to fix this later :)

bduffany (approved):

PR #3375, opened 2023-02-06
Enable strict mode for //enterprise/app/api_keys

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/api_keys/api_keys.tsx
+++ b/enterprise/app/api_keys/api_keys.tsx
@@ -20,7 +20,7 @@ import { api_key } from "../../../proto/api_key_ts_proto";
 
 export interface ApiKeysComponentProps {
   /** The authenticated user. */
-  user?: User;
+  user: User;
 
   /** Whether to show only user-owned keys. */
   userOwnedOnly?: boolean;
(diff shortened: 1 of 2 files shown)


iain-macdonald (approved):

enterprise/app/api_keys/api_keys.tsx
```diff
@@ -315,13 +321,13 @@ export default class ApiKeysComponent extends React.Component<ApiKeysComponentPr
...
     onRequestClose: () => any;
-    onSubmit: () => any;
+    onSubmit: (e: React.FormEvent) => any;
     onChange: (name: string, value: any) => any;
     ref: React.RefObject<HTMLFormElement>;
     formState: FormState<T>;
   }) {
     return (
-      <Modal isOpen={isOpen} onRequestClose={onRequestClose} shouldFocusAfterRender={false}>
+      <Modal isOpen={Boolean(isOpen)} onRequestClose={onRequestClose} shouldFocusAfterRender={false}>
```
jdhollen, line 330: I guess i'm surprised you need the cast here? Doesn't the destructuring carry the type info along with it?

jdhollen (approved):

enterprise/app/api_keys/api_keys.tsx
bduffany: Good catch, I made this change before making all the fields required.

PR #3378, opened 2023-02-07
Add RPCs for heatmaps and drilldowns and support filters and duration for all stats queries.

jdhollen:
I don't know that I'd quite call myself fully happy with this (the queries are kind of weird, to be honest), but this is fully functional.

If you feel strongly about it, I'd split up the change to do one RPC at a time.

I still have other changes to come to support stat filters in SearchInvocations (and fetching invocations via execution filters in SearchInvocations).


proto/stat_filter.proto
```diff
@@ -7,6 +7,7 @@ enum MetricType {
   DURATION_USEC_METRIC = 1;
   CAS_CACHE_MISSES_METRIC = 2;
   EXECUTION_QUEUE_TIME_USEC_METRIC = 3;
+  UPDATED_AT_USEC_METRIC = 4;
```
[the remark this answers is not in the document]
jdhollen: Ah, yikes, you know, it hadn't even occurred to me that these were actually separated "updated at" fields and i had naively assumed that this was always the invocation field.

I could either split out the metric to "ExecutionMetric" and "InvocationMetric" (and, if we want, even split to separate RPCs as well..) or I could prefix each of this in the same way as I did EXECUTION_QUEUE_TIME--what do you all think?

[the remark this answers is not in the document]
luluz66: ExecutionMetricType and InvocationMetricType make sense. I don't think we need to split to separate RPCs yet.

server/util/filter/filter.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+func GenerateFilterStringAndArgs(f *stat_filter.StatFilter, paramPrefix string) (string, []interface{}, error) {
+	metric, err := MetricToDbField(f.GetMetric(), paramPrefix)
+	if err != nil {
+		return "", nil, err
+	}
+
+	return fmt.Sprintf("(%s BETWEEN ? AND ?)", metric), []interface{}{f.GetMin(), f.GetMax() - 1}, nil
```
luluz66, line 31: is it possible that either getMin or getMax is not set? and instead of `Between..And` we want `>` and `<`

jdhollen: I didn't do this at first because I didn't need it, but it's so easy, why not.  I also changed the semantics very slightly here (I'm not subtracting 1 from max anymore) because that makes way more sense--I'll change the request to what I actually want client-side instead of hard-coding weird trash server-side.

proto/stat_filter.proto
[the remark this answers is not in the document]
jdhollen: This was easier than I thought it'd be, done.

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -135,79 +140,93 @@ func flattenTrendsQuery(innerQuery string) string {
...
 	}
 
-	q.AddWhereClause(`group_id = ?`, groupID)
-	q.SetGroupBy("name")
+	if tq.GetMinimumDuration().GetSeconds() != 0 {
+		q.AddWhereClause(`duration_usec >= ?`, tq.GetMinimumDuration().GetSeconds()*1e6)
+	}
+	if tq.GetMaximumDuration().GetSeconds() != 0 {
+		q.AddWhereClause(`duration_usec <= ?`, tq.GetMaximumDuration().GetSeconds()*1e6)
+	}
```
tylerwilliams, line 208: We've screwed up usec/nsec/sec before. You could do:

tq.GetMaximumDuration().AsDuration().Microseconds()

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -374,6 +400,264 @@ func (i *InvocationStatService) GetInvocationStatBaseQuery(aggColumn string) str
...
+	if lowTime != nil && !lowTime.IsValid() {
+		return nil, "", status.InvalidArgumentError(fmt.Sprintf("Invalid low timestamp: %v", lowTime))
+	}
+	if highTime != nil && (!highTime.IsValid() || highTime.AsTime().Unix() < lowTime.AsTime().Unix()) {
+		return nil, "", status.InvalidArgumentError(fmt.Sprintf("Invalid high timestamp: %v", highTime))
+	}
+
+	offsetSec := int64(timezoneOffsetMinutes) * 60
+
+	startSec := time.Now().Unix() - ((7 * 24 * time.Hour).Microseconds() / 1e6)
```
tylerwilliams, line 481: you could make this a const like:
```
  const oneWeek = 7 * 24 * time.Hour
```

then have startSec := time.Now().Unix() - oneWeek.Seconds()

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -374,6 +400,264 @@ func (i *InvocationStatService) GetInvocationStatBaseQuery(aggColumn string) str
...
+	}
+	mStr := make([]string, metricBucketCount)
+	for i := range mStr {
+		mStr[i] = fmt.Sprint(metricBuckets[i])
+	}
+	metricArrayStr := "array(" + strings.Join(mStr, ",") + ")"
+	return metricBuckets, metricArrayStr, nil
+}
+
+func getTimestampBuckets(q *stpb.TrendQuery, timezoneOffsetMinutes int32) ([]int64, string, error) {
```
tylerwilliams, line 469: if you pass timezoneOffsetMinutes as a time.Duration instead of an int32 you can simplify some of your code below and just grab .Seconds() or .Add it without needing to multiply / convert it

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -374,6 +400,264 @@ func (i *InvocationStatService) GetInvocationStatBaseQuery(aggColumn string) str
...
+	for currentString != endString {
+		timestampBuckets = append(timestampBuckets, currentDateBracket.UnixMicro())
+		currentDateBracket = currentDateBracket.AddDate(0, 0, 1)
+		currentString = currentDateBracket.Format("2006-01-02")
+	}
+
+	dStr := make([]string, len(timestampBuckets)-1)
+	for i, d := range timestampBuckets[:len(timestampBuckets)-1] {
+		dStr[i] = fmt.Sprintf("%d", d)
+	}
```
tylerwilliams, line 511: big nit but you could just define a var for `numDateBuckets = len(timestampBuckets)-1` then use that instead of len(timestampBuckets)-1 in multiple places so folks don't have to squint at the range indexing as hard

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -485,6 +777,211 @@ func (i *InvocationStatService) GetInvocationStat(ctx context.Context, req *inpb
...
+	for _, s := range drilldownFields {
+		subQStr, subQArgs := i.getDrilldownSubquery(ctx, drilldownFields, req, whereString, whereArgs, drilldownStr, drilldownArgs, s)
+		queries = append(queries, subQStr)
+		args = append(args, subQArgs...)
+	}
+
+	return fmt.Sprintf("SELECT * FROM (%s) ORDER BY totals_first", strings.Join(queries, " UNION ALL ")), args, nil
+}
+
+func addOutputChartEntry(m *map[inpb.AggType]*stpb.DrilldownChart, dm *map[inpb.AggType]float64, aggType inpb.AggType, label *string, inverse int64, selection int64, totalInBase int64, totalInSelection int64) {
```
tylerwilliams, line 868: maps types are references so you don't need to pass them with a pointer in go (which should simplify a lot of the code below)

tylerwilliams (changes requested):
small nits / go-isms but otherwise lgtm

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -374,6 +400,264 @@ func (i *InvocationStatService) GetInvocationStatBaseQuery(aggColumn string) str
...
+	// limit on the upper bound of requested ranges. Each bucket is [min, max).
+	width := valueRange.High + 1 - valueRange.Low
+	var step int64
+
+	bucketCount := int64(20)
+	if width <= bucketCount {
+		step = 1
+	} else {
+		step = width / bucketCount
+		if width%bucketCount != 0 {
```
luluz66, line 440: The format is a bit wierd here. should there be space around '%'

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -374,6 +400,264 @@ func (i *InvocationStatService) GetInvocationStatBaseQuery(aggColumn string) str
...
+func getTableForMetric(metric *sfpb.Metric) string {
+	switch metric.MetricType.(type) {
+	case *sfpb.Metric_Execution:
+		return "Executions"
+	default:
+		return "Invocations"
+	}
+}
+
+func (i *InvocationStatService) getMetricRange(ctx context.Context, table string, metric string, whereClauseStr string, whereClauseArgs []interface{}) (int64, int64, int64, error) {
```
luluz66, line 412: nit: I feel if we define a go struct and return the struct here it would be more readable?

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -374,6 +400,264 @@ func (i *InvocationStatService) GetInvocationStatBaseQuery(aggColumn string) str
...
+			SELECT arrayElement(%s, number + 1) AS bucket FROM numbers(%d)) AS b
+		ORDER BY timestamp, bucket`,
+		timestampArrayStr, len(timestampBuckets)-1, metricArrayStr, len(metricBuckets)-1)
+
+	return &HeatmapQueryInputs{
+		PlaceholderBucketQuery: pbq,
+		TimestampBuckets:       timestampBuckets,
+		TimestampArrayStr:      timestampArrayStr,
+		MetricBuckets:          metricBuckets,
+		MetricArrayStr:         metricArrayStr}, nil
```
luluz66, line 551: nit: should "}, nil" go to the next line?

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -374,6 +400,264 @@ func (i *InvocationStatService) GetInvocationStatBaseQuery(aggColumn string) str
...
+
+type QueryAndBuckets = struct {
+	Query            string
+	QueryArgs        []interface{}
+	TimestampBuckets []int64
+	MetricBuckets    []int64
+}
+
+/** This function returns (nil, nil) if it finds that there is no data to fetch. */
+func (i *InvocationStatService) getHeatmapQueryAndBuckets(ctx context.Context, req *stpb.GetStatHeatmapRequest, timezoneOffsetMinutes int32) (*QueryAndBuckets, error) {
```
luluz66, line 571: // getHeatmapQueryAndBuckets returns...

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Sure yeah yes done

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: non-actionable comment: it really kinda bugs me that this outputs with float64 (micros outputs int64).  I guess I'm supposed to take for granted that the underlying type is an int64 and these conversion methods are only converting to float at the last possible moment (so I'll never get close to the int-precision limit in seconds?)

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Done

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Sure, why not.

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Thanks, I went wild with these before I realized what the deal was (strings too, you should have seen it), and got most things but missed this.

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: This is what my autoformatter does when I have them just out in the open, I think to convey order of operations or something?  I agree it's ugly.  When I put this in parens, it doesn't "fix" it anymore, and I think it's easier on the eyes anyway, so let's go with that.

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Sure, done

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: This is an autoformatter quirk of mine, I did something different now, I have basically no opinion on this :P

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Made this comment not stink as much

jdhollen:
@luluz66 @tylerwilliams any more comments on this? there's one more gnarly-ish frontend change sitting behind this one and then i should be able to do the rest as quick / small things.

tylerwilliams (approved):

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -135,79 +140,93 @@ func flattenTrendsQuery(innerQuery string) string {
...
+	if start := tq.GetUpdatedAfter(); start.IsValid() {
 		q.AddWhereClause("updated_at_usec >= ?", start.AsTime().UnixMicro())
 	} else {
 		// If no start time specified, respect the lookback window field if set,
 		// or default to 7 days.
 		// TODO(bduffany): Delete this once clients no longer need it.
 		lookbackWindowDays := 7 * 24 * time.Hour
-		if w := req.GetLookbackWindowDays(); w != 0 {
-			if w < 1 || w > 365 {
+		if lookback != 0 {
```
luluz66, line 184: nit: is the unit of lookback day? and lookbackWindowDays is actually hours? can we rename the two names to clarify the units?

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -485,6 +795,211 @@ func (i *InvocationStatService) GetInvocationStat(ctx context.Context, req *inpb
...
+
+	rsp := &stpb.GetStatDrilldownResponse{}
+
+	rsp.Chart = make([]*stpb.DrilldownChart, 0)
+	m := make(map[inpb.AggType]*stpb.DrilldownChart)
+	dm := make(map[inpb.AggType]float64)
+	type queryOut struct {
+		GormUser       *string
+		GormHost       *string
+		GormRepoUrl    *string
```
luluz66, line 949: nit: URL for go conventions. see https://github.com/golang/go/wiki/CodeReviewComments#initialisms

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -485,6 +795,211 @@ func (i *InvocationStatService) GetInvocationStat(ctx context.Context, req *inpb
...
+		GormBranchName *string
+		GormCommitSha  *string
```
luluz66, line 951: nit: SHA for the same reason

luluz66 (approved):
lgtm overall. have some nits on variable or field names

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: thanks, i fixed up the units on these (the "days" field is a duration anyway)

PR #3379, opened 2023-02-07
Auth support for user-level keys

bduffany:
* Fill `user_id` field in authenticated `APIKeyGroup`, so that we can tell internally whether a key is user-owned if desired.
* Fill the `user_owned_keys_enabled` field in the authenticated user's group roles.
* Disable authentication with user-owned keys when the setting is disabled via org settings.

Note, this is tested in userdb_test in https://github.com/buildbuddy-io/buildbuddy/pull/3280. The existing functionality for group-level keys is also tested in existing integration tests which use enterprise_testauth (remote_execution_test, workflow service test, etc.)

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -21,9 +21,14 @@ func NewAuthDB(h interfaces.DBHandle) *AuthDB {
 }
 
 type apiKeyGroup struct {
+	UserID                 string
 	GroupID                string
 	Capabilities           int32
 	UseGroupOwnedExecutors bool
+}
+
+func (g *apiKeyGroup) GetUserID() string {
+	return g.UserID
 }
 
 func (g *apiKeyGroup) GetGroupID() string {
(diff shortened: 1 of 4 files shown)


vadimberezniker (approved):

tylerwilliams (approved):

enterprise/server/backends/authdb/authdb.go
```diff
@@ -180,16 +189,32 @@ func (d *AuthDB) LookupUserFromSubID(ctx context.Context, subID string) (*tables
...
 			g.use_group_owned_executors
 		FROM ` + "`Groups`" + ` AS g,
 		APIKeys AS ak
 	`)
 	qb.AddWhereClause(`ak.group_id = g.group_id`)
+	if allowUserOwnedKeys {
+		// Note: the org can disable user-owned keys at any time, and the
+		// predicate here ensures that existing keys are effectively deactivated
+		// (but not deleted).
+		qb.AddWhereClause(`(
```
vadimberezniker, line 207: Should we change `AddWhereClause` to automatically add parenthesis around the clause? Expecting users to remember to do it seems like asking for trouble.

vadimberezniker (approved):

enterprise/server/backends/authdb/authdb.go
bduffany: Seems like a good idea to me. I can send a separate PR for that

tylerwilliams: +1

tylerwilliams (approved):

enterprise/server/backends/authdb/authdb.go
bduffany: Sent https://github.com/buildbuddy-io/buildbuddy/pull/3444/files

PR #3383, opened 2023-02-08
create a heatmap for exploring invocation and execution stats.

jdhollen:
this is just the component, still lots more to come, but this is a nice standalone piece.

note that this does not add any new node deps; i've avoided adding duplicate deps and am instead depending on our existing d3 and resizer versions, which are good enough for me.


enterprise/app/trends/heatmap.tsx
```diff
@@ -0,0 +1,422 @@
...
+  top: 1,
+  right: 1,
+  bottom: 39,
+  left: 89,
+};
+
+// This is a magic number that states there will only be one axis label for
+// every N pixels of rendered axis length (currently 100).
+const TICK_LABEL_SPACING_MAGIC_NUMBER = 100;
+const heatmapColorScale = [
```
bduffany, line 47: nit: is there a way to generate this color scale automatically? (if there is a tool you can link to here, that might be helpful, otherwise no worries)

enterprise/app/trends/heatmap.tsx
```diff
@@ -0,0 +1,422 @@
...
+  pendingClick?: [TooltipData, TooltipData];
+  selectedData?: [TooltipData, TooltipData];
+
+  constructor(props: HeatmapProps) {
+    super(props);
+    this.state = {};
+    this.xScaleBand = scaleBand();
+    this.yScaleBand = scaleBand();
+    this.svgRef = React.createRef();
+  }
```
bduffany, line 90: nit: might be slightly cleaner to initialize these in the top level of the component -- then the constructor can be removed

enterprise/app/trends/heatmap.tsx
```diff
@@ -0,0 +1,422 @@
...
+  width: number;
+  height: number;
+  valueFormatter: (value: number) => string;
+  metricBucketName: string;
+  metricBucketFormatter: (value: number) => string;
+  selectionCallback?: (s?: HeatmapSelection) => void;
+}
+
+interface State {
+  selectionToRender?: [TooltipData, TooltipData];
```
bduffany, line 20: at first glance it's unclear why this is a pair of `TooltipData` rather than just one, mind adding a comment?

enterprise/app/trends/heatmap.tsx
```diff
@@ -0,0 +1,422 @@
...
+  timestamp: number;
+  timestampBucketIndex: number;
+  metric: number;
+  metricBucketIndex: number;
+  value: number;
+};
+
+export type HeatmapSelection = {
+  dateRangeMicros: number[];
+  bucketRange: number[];
```
bduffany, line 33: nit: if these are pairs, maybe type as `[startInclusive: number, endExclusive: number]` (assuming the end range is exclusive, haven't grokked that part yet :sweat_smile: )

enterprise/app/trends/heatmap.tsx
```diff
@@ -0,0 +1,422 @@
...
+    this.props.heatmapData.column.map((column) => {
+      column.value.map((value) => {
+        min = Math.min(+value, min);
+        max = Math.max(+value, max);
+      });
+    });
+
+    const selection = this.maybeRenderSelection();
+    const interpolator = (v: number) =>
+      heatmapColorScale[Math.floor((heatmapColorScale.length * 0.99 * (v - min)) / (max - min))];
```
bduffany, line 368: nit: the `* 0.99` looks a little weird, is there a cleaner way to do this calculation? (`length - 1`  maybe?)

enterprise/app/trends/heatmap.tsx
```diff
@@ -0,0 +1,422 @@
...
+    this.yScaleBand = scaleBand<number>().range([0, height]).domain(yDomain);
+
+    let min = 0;
+    let max = 0;
+    this.props.heatmapData.column.map((column) => {
+      column.value.map((value) => {
+        min = Math.min(+value, min);
+        max = Math.max(+value, max);
+      });
+    });
```
bduffany, line 364: uber nit:

```suggestion
    this.props.heatmapData.column.forEach((column) => {
      column.value.forEach((value) => {
        min = Math.min(+value, min);
        max = Math.max(+value, max);
      });
    });
```

app/components/tooltip/tooltip.tsx
```diff
@@ -2,9 +2,14 @@ import React from "react";
...
 
+export type MouseCoords = {
+  readonly clientX: number;
+  readonly clientY: number;
+};
+
 type TooltipProps = JSX.IntrinsicElements["div"] & {
   /** Renders the tooltip content if the tooltip is hovered. */
-  renderContent: () => React.ReactNode;
+  renderContent: (c?: MouseCoords) => React.ReactNode;
```
bduffany, line 12: Can this be `c: MouseCoords` so that users of the component don't need to handle undefined? (maybe if MouseCoords are unknown we can set to the last known value, or don't render initially until mouse coords are known?)

enterprise/app/trends/heatmap.tsx
jdhollen: I wrote out what I did in words; if we wanted to check it in as a tool i'd be happy to do that, but it was just some dumb little for loop that i wrote.

enterprise/app/trends/heatmap.tsx
jdhollen: Thanks, I had some meatier stuff in here before and forgot to clean this up.

enterprise/app/trends/heatmap.tsx
jdhollen: Sure yes, and I changed the name as well because tooltipdata stinks and is dated.

enterprise/app/trends/heatmap.tsx
jdhollen: Yes, correct, and yes, good idea.

enterprise/app/trends/heatmap.tsx
jdhollen: I was trying to avoid getting screwed up by floating point division with my floor() call.  thinking about this a bit harder, i *think* i'm being overly paranoid because i'm not going to be using huge values and v === max should cleanly pop out as integer 1.

enterprise/app/trends/heatmap.tsx
jdhollen: bah, thank you

app/components/tooltip/tooltip.tsx
jdhollen: Already more or less works as you described it: `renderContent` is not invoked unless the tooltip is set to `visible`, and the MouseCoords should be correct in that case.

enterprise/app/trends/heatmap.tsx
```diff
@@ -0,0 +1,434 @@
...
+      metric: +metric,
+      metricBucketIndex: stepY,
+      value: +column.value[stepY],
+    };
+  }
+
+  renderTooltip(c?: MouseCoords) {
+    if (!c || this.pendingClick) {
+      return null;
+    }
```
bduffany, line 136: nit, IIUC this would work now?

```suggestion
  renderTooltip(c: MouseCoords) {
    if (this.pendingClick) {
      return null;
    }
```

bduffany (approved):

enterprise/app/trends/heatmap.tsx
jdhollen: Oops, yup, fixed.

PR #3384, opened 2023-02-08
Update RBE setup docs

bduffany:
---

**Version bump**: Patch 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2023

### DIFF
--- a/docs/rbe-setup.md
+++ b/docs/rbe-setup.md
@@ -20,7 +20,7 @@ This points Bazel at BuildBuddy Cloud as a remote executor. A simple repo that h
 
 ## Configuring your workspace
 
-There are several options for configuring your platforms and toolchains, the most fully features of which being [bazel-toolchains](https://releases.bazel.build/bazel-toolchains.html). It comes with an `rbe_autoconfig` rule that works nicely with BuildBuddy.
+There are several options for configuring your platforms and toolchains, the most fully featured of which being [bazel-toolchains](https://releases.bazel.build/bazel-toolchains.html). It comes with an `rbe_autoconfig` rule that works nicely with BuildBuddy.
 
 Unfortunately, bazel-toolchains has a dependency on Docker and can take quite some time to start up in a clean workspace, so we provide a simple and easy-to-use [BuildBuddy toolchain](https://github.com/buildbuddy-io/toolchain) that enables you to get up and running quickly, and works for most use cases.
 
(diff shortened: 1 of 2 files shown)


docs/rbe-setup.md
```diff
@@ -114,10 +150,10 @@ build:remote --platforms=@buildbuddy_toolchain//:platform
...
 build:remote --crosstool_top=@buildbuddy_toolchain//:toolchain
 build:remote --extra_toolchains=@buildbuddy_toolchain//:cc_toolchain
-build:remote --javabase=@buildbuddy_toolchain//:javabase_jdk8
-build:remote --host_javabase=@buildbuddy_toolchain//:javabase_jdk8
-build:remote --java_toolchain=@buildbuddy_toolchain//:toolchain_jdk8
-build:remote --host_java_toolchain=@buildbuddy_toolchain//:toolchain_jdk8
+build:remote --javabase=@buildbuddy_toolchain//:javabase
+build:remote --host_javabase=@buildbuddy_toolchain//:javabase
+build:remote --java_toolchain=@buildbuddy_toolchain//:java_toolchain
+build:remote --host_java_toolchain=@buildbuddy_toolchain//:java_toolchain
```
sluongng, line 156: In bazel 6, I think most of these flags are marked as deprecated.

I think these are recommended instead

```
--java_runtime_version=
--tool_java_runtime_version=
```

Not too sure about `--java_language_version` 🤔

sluongng: From https://bazel.build/docs/user-manual#java-language-version

```
--java_language_version=11
--tool_java_language_version=11
--java_runtime_version=remotejdk_11
--tool_java_runtime_version=remotejdk_11
```

bduffany: If I'm reading that right, `remotejdk_11` uses a hermetic Java. I am wondering if we should instead recommend `11` instead of `remotejdk_11`, since the RBE image already has Java 11 installed. But maybe it'd be better to recommend remotejdk anyway, so people don't have to depend on the Java version in our execution image?

sluongng: > You can extend the values by registering custom JVM using either local_java_repository or remote_java_repostory repository rules.

I think sticking with `remotejdk_11` for the simplicity of it. And we could add a line of doc pointing to Bazel's official recommendation if people want to bring their own JDK. WDYT?

bduffany:
I don't have the bandwidth to resume work on this right now - closing

PR #3386, opened 2023-02-08
Allow workflows to attach invocation artifacts

bduffany:
Adds support for bazel commands like `test //... --experimental_remote_grpc_log=$ARTIFACTS_DIRECTORY/grpc.log`.

The CI runner provisions `$ARTIFACTS_DIRECTORY` as `${workspace_root}/artifacts/command-${i}/` for each command index `i` in `bazel_commands[]`.

After each bazel command is complete, the CI runner kicks off background uploads of all artifacts under that directory, and publishes a `NamedSetOfFilesEvent` with the id `command-${i}` and each file named according to their relative path under that directory (`grpc.log` in this example).

It logs results for all uploads at the end of the workflow, so users can see how much was uploaded and how long each upload took.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/1943


enterprise/server/bes_artifact_uploader/bes_artifact_uploader.go
```diff
@@ -0,0 +1,104 @@
...
+	bep *build_event_publisher.Publisher
+	// Bytestream server host name and port.
+	bytestreamHost string
+	// Remote instance name to be used for uploaded artifacts.
+	instanceName string
+	// Channel where upload results are published (non-blocking).
+	resultsChan chan *Result
+}
+
+func New(ctx context.Context, bsClient bspb.ByteStreamClient, bep *build_event_publisher.Publisher, bytestreamHost, instanceName string) *Uploader {
```
tylerwilliams, line 44: Rather than injecting bsClient, bytestreamHost (and instanceName?) could you use flags in this file for those things and construct that stuff in New() (and return an err also)

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -187,10 +210,61 @@ type workspace struct {
...
+		return nil, status.InvalidArgumentErrorf("failed to parse cache target %q: %s", cacheTarget, err)
+	}
+	conn, err := grpc_client.DialTarget(cacheTarget)
+	if err != nil {
+		return nil, err
+	}
+	// Note: we don't defer conn.Close() here since it's OK to keep it alive
+	// until the CI runner exits.
+	bsClient := bspb.NewByteStreamClient(conn)
+	ul := bes_artifact_uploader.New(ctx, bsClient, bep, u.Host, instanceName)
```
tylerwilliams, line 225: this code could all move to uploader.New

enterprise/server/bes_artifact_uploader/bes_artifact_uploader.go
```diff
@@ -0,0 +1,104 @@
+package bes_artifact_uploader
```
tylerwilliams, line 1: could this be `package bes`?

Then your code would read bes.Uploader (and would not stutter)

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -187,10 +210,61 @@ type workspace struct {
...
+	return ul, nil
+}
+
+func artifactsRootPath(ws *workspace) string {
+	return filepath.Join(ws.rootDir, artifactsDirName)
+}
+
+func artifactsPathForCommand(ws *workspace, bazelCommandIndex int) string {
+	return filepath.Join(artifactsRootPath(ws), fmt.Sprintf("command_%d", bazelCommandIndex))
```
tylerwilliams, line 234: would command-%d look nicer?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -216,7 +290,15 @@ func newBuildEventReporter(ctx context.Context, besBackend string, apiKey string
 		return nil, status.UnavailableErrorf("failed to initialize build event publisher: %s", err)
 	}
 	bep.Start(ctx)
-	return &buildEventReporter{apiKey: apiKey, bep: bep, log: newInvocationLog(), invocationID: iid, isWorkflow: isWorkflow}, nil
+
+	log := newInvocationLog()
+
+	ul, err := newArtifactUploader(ctx, bep, cacheBackend, instanceName)
+	if err != nil {
+		log.Printf("WARNING: Failed to initialize BES artifact uploader; artifacts written to $%s will not be uploaded: %s", artifactsDirEnvVarName, err)
```
tylerwilliams, line 298: should this just be an error?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -803,12 +885,45 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
...
+		allUploadsChan := make(chan []*bes_artifact_uploader.Result)
+		// Buffer upload results
+		go func() {
+			defer close(allUploadsChan)
+			var uploads []*bes_artifact_uploader.Result
+			for u := range ar.reporter.uploader.Results() {
+				uploads = append(uploads, u)
+			}
+			allUploadsChan <- uploads
+		}()
```
tylerwilliams, line 898: Is this channel necessary?

enterprise/server/cmd/ci_runner/main.go
bduffany: Yeah, that simplifies things. Done!

enterprise/server/cmd/ci_runner/main.go
bduffany: I like that better - done

enterprise/server/cmd/ci_runner/main.go
bduffany: Done - I was on the fence about having the uploader do a Dial operation in the constructor but I think it doesn't really matter for now. Done

enterprise/server/cmd/ci_runner/main.go
bduffany: Got rid of it in favor of having the Uploader be responsible for buffering all the upload result metadata.

enterprise/server/bes_artifact_uploader/bes_artifact_uploader.go
bduffany: Done

enterprise/server/bes_artifact_uploader/bes_artifact_uploader.go
bduffany: Renamed to `bes_artifacts` to avoid making the scope of this package seem too broad; wdyt?

bduffany (commented):
I restructured the uploader a bit to upload files as one NamedSetOfFiles per directory, instead of one NamedSetOfFiles event per file. This makes the frontend logic a bit cleaner and also cleans up the ci_runner a bit since now the uploader is responsible for walking the dirs. PTAL :)

sluongng:
I am a big fan of this change.

This should enable our customers (and us too) to self-troubleshoot their Workflow runs a lot easier.

enterprise/server/bes_artifacts/bes_artifacts.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// uploadFile starts a background file upload.
+func (u *Uploader) uploadFile(setID, path, name string) chan *Result {
+	ch := make(chan *Result, 1)
+	u.eg.Go(func() (err error) {
+		start := time.Now()
+		result := &Result{NamedSetID: setID, Name: name, Path: path}
+		defer func() {
+			result.Err = err
```
vadimberezniker, line 134: Since there are only 2 returns paths, what do you think about explicitly publishing a result for each of the two cases rather than using this defer which is slightly more "magical"?

vadimberezniker (approved):

enterprise/server/bes_artifacts/bes_artifacts.go
bduffany: Good point - cleaned this up. Also while looking at this I realized there was a data race on `u.results` - fixed that by adding a lock.

enterprise/server/bes_artifacts/bes_artifacts.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"golang.org/x/sync/errgroup"
+
+	bespb "github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
+)
+
+type Result struct {
```
tylerwilliams, line 24: nit: want to add a comment for this?

enterprise/server/bes_artifacts/bes_artifacts.go
```diff
@@ -0,0 +1,159 @@
...
+	bytestreamURIPrefix := "bytestream://" + u.Host
+
+	conn, err := grpc_client.DialTarget(cacheTarget)
+	if err != nil {
+		return nil, err
+	}
+	go func() {
+		<-ctx.Done()
+		conn.Close()
+	}()
```
tylerwilliams, line 69: wdyt about doing this in Wait()?

That way the lifetime of this object is more explicit and likely to be shorter. It also avoids a leak if someone calls it with a background context or something silly.

bduffany: Done

enterprise/server/bes_artifacts/bes_artifacts.go
bduffany: Done

tylerwilliams (approved):

PR #3394, opened 2023-02-10
Add retries for buildbuddy tests in dev qa script for flaky tests

maggie-lou:
---

**Version bump**: TODO

### DIFF
--- a/tools/dev_qa.sh
+++ b/tools/dev_qa.sh
@@ -64,14 +64,16 @@ buildbuddy(name = "buildbuddy_toolchain", container_image = UBUNTU20_04_IMAGE)' 
     >> WORKSPACE
     fi
 
-    bazel_cmd_output=0
-    for attempt in $(seq 1 "$num_retries"); do
+    for attempt in $(seq 0 "$num_retries"); do
+      bazel_cmd_output=0
       bazel clean
-      bazel_cmd_output=$("$@")
+      "$@" || bazel_cmd_output=$?
 
       if [ "$bazel_cmd_output" -eq 0 ]; then
             break
       fi
+
+      echo "Retrying buildbuddy tests"
     done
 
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3394** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3394" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3394?utm_source=stack-comment).

tools/dev_qa.sh
```diff
@@ -79,13 +94,15 @@ tensorflow_iid=$(invocation_id)
 run_test \
     https://github.com/buildbuddy-io/buildbuddy \
     buildbuddy \
+    3 \
```
bduffany, line 97: Since this is an "optional" param you could alternatively set it as a variable scoped to the run_test command, like

`NUM_RETRIES=3 run_test ...`

Then in the func

```
: "${NUM_RETRIES:=0}"
```

bduffany (commented):
Is this flaky because of a certain test, or are we failing to fetch some deps or something like that? (If it's the former, bazel lets you set the retry count per-test; see `--flaky_test_attempts` in our .bazelrc for an example)

tools/dev_qa.sh
maggie-lou: > Is this flaky because of a certain test, or are we failing to fetch some deps or something like that? (If it's the former, bazel lets you set the retry count per-test; see `--flaky_test_attempts` in our .bazelrc for an example)

Thanks that's simpler. Flaky because of certain tests. Made a new PR https://github.com/buildbuddy-io/buildbuddy/pull/3398/files

PR #3396, opened 2023-02-12
docs: document workflow disk space config

sluongng:
Add doc for the config that was introduced in commit b071ad2dd3d79b6f8fbd56200e2f074b61a912e7.

---

**Version bump**: None


bduffany (approved):
Thanks!

PR #3397, opened 2023-02-13
Workflows: add resource requests to recycling key

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -197,6 +197,8 @@ type commandRunner struct {
 
 	// task is the current task assigned to the runner.
 	task *repb.ExecutionTask
+	// taskSize is the size of the last task assigned to the runner.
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -1377,6 +1388,12 @@ func (p *pool) setLimits() {
 		p.maxRunnerCount, p.maxRunnerMemoryUsageBytes, p.maxRunnerDiskUsageBytes)
 }
 
+func taskSizesEqual(a, b *scpb.TaskSize) bool {
```
siggisim, line 1391: Should we allow matching as long as the current requests are smaller than the limits (rather then just equal?)

Also, is there a way to compare against the actual limits rather than just what the previous task looked like? That way we could enable this for non-workflows too?

bduffany: Resolved offline to address this later.

PR #3405, opened 2023-02-13
Add metrics for clickhouse SQL query

luluz66:
Use a similar approach to our main database to set up monitoring for
query duration.


luluz66:
> > Use a similar approach to our main database to set up monitoring for query duration. Didn't add error count or query count since we can get these metrics from clickhouse operator dashboard.
> 
> I see an error count in the PR. Was the PR updated? If you're adding the error count, you might as well add the total query count to make it easy to query using the same labels.

I updated the PR description and add the total query count.

vadimberezniker (approved):

PR #3406, opened 2023-02-14
Add ability to fetch invocations with execution filters.

jdhollen:


enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -94,63 +99,58 @@ func addPermissionsCheckToQuery(u interfaces.UserInfo, q *query_builder.Query) {
...
-	if command := req.GetQuery().GetCommand(); command != "" {
+	if command := iq.GetCommand(); command != "" {
 		q.AddWhereClause("i.command = ?", command)
 	}
-	if sha := req.GetQuery().GetCommitSha(); sha != "" {
+	if sha := iq.GetCommitSha(); sha != "" {
 		q.AddWhereClause("i.commit_sha = ?", sha)
 	}
-	if group_id := req.GetQuery().GetGroupId(); group_id != "" {
+	if group_id := iq.GetGroupId(); group_id != "" {
```
luluz66, line 141: if we are using clickhouse, i think group_id is required for searching invocation or executions. Without group_id, it will result in full scan.

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -170,6 +170,93 @@ func (s *InvocationSearchService) QueryInvocations(ctx context.Context, req *inp
 	if statusQuery != "" {
 		q.AddWhereClause(fmt.Sprintf("(%s)", statusQuery), statusArgs...)
 	}
+}
+
+func (s *InvocationSearchService) searchInvocationsInExecutions(ctx context.Context, req *inpb.SearchInvocationRequest) (*query_builder.Query, error) {
+	if s.oh == nil {
+		return nil, status.UnimplementedError("Execution filters require an OLAP DB.")
+	}
+	q := query_builder.NewQuery(`SELECT DISTINCT invocation_uuid FROM Executions as i`)
```
luluz66, line 179: i think use "i" as an alias of Executions is confusing. can we use "e" instead?

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -170,6 +170,93 @@ func (s *InvocationSearchService) QueryInvocations(ctx context.Context, req *inp
...
+	addCommonWhereClauses(q, req.GetQuery())
+	for _, f := range req.GetQuery().GetFilter() {
+		if f.GetMetric().Execution == nil {
+			continue
+		}
+		str, args, err := filter.GenerateFilterStringAndArgs(f, "i.")
```
luluz66, line 185: again - I think we need to make sure that "group_id" and "updated_at_usec" is added as part of the where clause, otherwise the performance is going to be bad.

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -170,6 +170,93 @@ func (s *InvocationSearchService) QueryInvocations(ctx context.Context, req *inp
...
+		return nil, status.UnimplementedError("Sorting and pagination is not supported for execution searches.")
+	}
+	q.SetOrderBy("i.updated_at_usec", false)
+
+	// More stuff...
+	rows, err := s.oh.DB(ctx).Raw(qString, qArgs...).Rows()
+	if err != nil {
+		return nil, err
+	}
+	type o struct {
```
luluz66, line 208: can we have a more descriptive name here? like output? or Row?

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -170,6 +170,93 @@ func (s *InvocationSearchService) QueryInvocations(ctx context.Context, req *inp
...
+		u, err := uuid.Parse(out.InvocationUuid)
+		if err != nil {
+			return nil, err
+		}
+
+		ids = append(ids, u.String())
+	}
+
+	invocationQuery := query_builder.NewQuery("SELECT * FROM Invocations AS i")
+	invocationQuery.AddWhereClause("i.invocation_id IN ?", ids)
```
luluz66, line 226: I think invocation_id is not in clickhouse Invocations table. but only invocation_uuid.

Also when querying clickhouse Invocation table, it will be faster if provide group_id, updated_at_usec range.

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -140,7 +140,7 @@ func flattenTrendsQuery(innerQuery string) string {
 	FROM (` + innerQuery + ")"
 }
 
-func addWhereClauses(q *query_builder.Query, tq *stpb.TrendQuery, reqCtx *ctxpb.RequestContext, lookbackWindowDays int32) error {
+func addWhereClauses(q *query_builder.Query, tq *stpb.TrendQuery, reqCtx *ctxpb.RequestContext, lookbackWindowDays int32, isExecutionQuery bool) error {
```
luluz66, line 143: I find the boolean here not super readable. Wondering if there are other ways to construct the code? for example, have two different methods for invocation and executions; extract common code in a separate method and etc.

jdhollen:
You know what, this one is just a little too weird.  I am gonna swap this out for fetching executions directly, make the cards link to /invocation-id#executions, and do away with all of the weird join logic i was trying to introduce in this change.

PR #3411, opened 2023-02-14
Update GetGroupUsers to use group_id from request context

bduffany:
Instead of relying on the group ID field, use `request_context.group_id`, which is authenticated in auth.go (`ClaimsFromSubID`) and exposed via `UserInfo.GetGroupID()`.

---

**Version bump**: Patch


enterprise/server/backends/userdb/userdb.go
```diff
@@ -554,17 +554,24 @@ func (d *UserDB) RequestToJoinGroup(ctx context.Context, userID string, groupID
...
+	u, err := perms.AuthenticatedUser(ctx, d.env)
+	if err != nil {
 		return nil, err
 	}
+	if err := authutil.AuthorizeGroupRole(u, u.GetGroupID(), role.Admin); err != nil {
+		return nil, err
+	}
+	if len(statuses) == 0 {
+		return nil, status.InvalidArgumentError("statuses list must not be empty")
+	}
```
tylerwilliams, line 567: It's nice for readability (and performance) to check preconditions before other stuff -- can we move this up before L558?

bduffany: Done

vadimberezniker (approved):

bduffany:
FYI I reverted the frontend/proto schema change since some apps will still expect to read the group ID from the group_id field during the rollout.

bduffany:
Closing this out for now and will re-open when I have time to revisit (and resolve merge conflicts)

PR #3412, opened 2023-02-14
strictify //enterprise/app/workflows/...

jdhollen:


enterprise/app/workflows/create_workflow.tsx
```diff
@@ -34,8 +34,13 @@ export default class CreateWorkflowComponent extends React.Component<CreateWorkf
     request: new workflow.CreateWorkflowRequest({ gitRepo: new workflow.CreateWorkflowRequest.GitRepo() }),
   };
 
-  private onChange(object: Record<string, any>, key: string, e: React.ChangeEvent<HTMLInputElement>) {
-    object[key] = e.target.value;
+  private handleRepoUrlChange(e: React.ChangeEvent<HTMLInputElement>) {
+    this.state.request.gitRepo!.repoUrl = e.target.value;
+    this.setState({ request: new workflow.CreateWorkflowRequest({ ...this.state.request }) });
```
jdhollen, line 39: note, the logic is very slightly different here--the old code modified the instance of `this.state.request` from the time of last render, whereas this change now modifies `this.state.request` directly.  i don't think this should matter, but am flagging it in case i'm missing some funky react quirk.

bduffany: Yeah, seems OK since you're setting the state to a new object.

bduffany (approved):
nice!

PR #3416, opened 2023-02-15
Add more validation to InsertUser and add tests

bduffany:
Check that userID and subID are both non-empty (these should be set by `Authenticator.FillUser`), and add tests.

---

**Version bump**: Patch 

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2053

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -689,15 +689,6 @@ func (d *UserDB) getDefaultGroupConfig() *tables.Group {
 }
 
 func (d *UserDB) createUser(ctx context.Context, tx *db.DB, u *tables.User) error {
-	groupIDs := make([]string, 0)
-	for _, group := range u.Groups {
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/userdb/userdb.go
```diff
@@ -698,6 +698,12 @@ func (d *UserDB) createUser(ctx context.Context, tx *db.DB, u *tables.User) erro
...
 
+	if u.UserID == "" {
+		return status.FailedPreconditionError("UserID is required")
+	}
+	if u.SubID == "" {
+		return status.FailedPreconditionError("SubID is required")
+	}
 	if u.Email == "" {
 		return status.FailedPreconditionErrorf("Auth token does not contain an email address")
 	}
```
tylerwilliams, line 709: Could you move move these checks up before line 692 so they are the first thing that happens in this method?

tylerwilliams (approved):

PR #3419, opened 2023-02-15
Handle `bb [build|query|test]` without specified targets

siggisim:
If no targets are specified, we'll build/test/query all targets.

---

**Version bump**: None


cli/arg/arg.go
```diff
@@ -98,6 +98,23 @@ func GetCommandAndIndex(args []string) (string, int) {
 	return "", -1
 }
 
+// Returns a list of bazel targets specified in the given set of arguments, if any.
+func GetTargets(args []string) []string {
```
bduffany, line 102:
```suggestion
// Returns a list of bazel targets specified in the given set of arguments, if any.
// This function assumes that the arguments are canonicalized.
func GetTargets(args []string) []string {
```

cli/cmd/bb/bb.go
```diff
@@ -146,6 +146,13 @@ func run() (exitCode int, err error) {
 		}
 	}
 
+	// If the command is build, test, or query without a specified target - apply to all targets.
+	// TODO(siggisim): show a picker of executable targets is run is specified without a target.
```
bduffany, line 150:
```suggestion
	// TODO(siggisim): show a picker of executable targets if run is specified without a target.
```

bduffany (approved):
nice!

PR #3421, opened 2023-02-15
Add script that prints out gorm auto migration commands without executing them

maggie-lou:
TODO: Do the same for clickhouse 

These scripts can be used if you want to know if a git branch will result in any new gorm auto-migrations (to prevent automatic release of that branch, for example)

The go script adds a callback to Gorm to write every SQL command executed during an auto-migration to a file.

The bash script copies the current schema of a database to a temporary test database. It then checks out a new git branch, runs auto-migration on the test database, and logs whether there are any new auto-migration commands


maggie-lou:
* master
  * **PR #3421** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3421" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3421?utm_source=stack-comment).

tools/gorm/print_auto_migration.go
```diff
@@ -0,0 +1,63 @@
...
+	bundle "github.com/buildbuddy-io/buildbuddy"
+)
+
+var (
+	dataSource = flag.String("data_source", "sqlite3:///tmp/buildbuddy.db", "The SQL database to connect to, specified as a connection string.")
+	outputPath = flag.String("output_path", "migration.txt", "The output file to write the gorm migration logs to.")
+)
+
+// This script prints the SQL commands that would be executed when running db.RunMigrations
+// It does not execute any of the commands
```
bduffany, line 22: Is it possible that some commands will be conditional on previous commands executing? I wonder if it would make sense to just let the commands execute.

If you are worried about modifying dev, you could maybe run a fresh mysql DB initialized to the current schema in dev (e.g. using Docker or maybe actions-setup-mysql to init a fresh DB, and running `SHOW TABLES` + `SHOW CREATE TABLE` statements against dev to recreate the current schema).

Or (simpler but less accurate) you could run the migration once at the current prod commit SHA against the empty mysql DB. 

Or (even simpler but more risky, maybe?) run against the actual DB but using a readonly connection, and just see if it fails due to trying to execute an ALTER TABLE statement but not being allowed to because of the readonly perms

tools/gorm/compare_migrations.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+# It assumes that the input data source has the db schema from the master branch
+# It generates the gorm auto-migration SQL statements on the master branch, then compares these statements to the
+# SQL statements on the input branch and prints any diff
+# It does not execute any migrations on the db
+
+git_branch=$1
+data_source=$2
+
+git checkout master
```
bduffany, line 15: will this be configurable? (the master branch is unstable so I guess this will probably not reflect the actual migration that'll happen)

tools/gorm/print_auto_migration.go
maggie-lou: Modified to run the auto-migration on a db copy

tylerwilliams (changes requested):
I think we'd be better off with this as a flag on the buildbuddy binary so we ensure that there's never a version discrepancy between the server (which would actually do the schema updates) and the tool (checking if there are schema updates).

maggie-lou:
> I think we'd be better off with this as a flag on the buildbuddy binary so we ensure that there's never a version discrepancy between the server (which would actually do the schema updates) and the tool (checking if there are schema updates).

Done

server/util/db/db.go
```diff
@@ -650,17 +652,52 @@ func GetConfiguredDatabase(env environment.Env) (interfaces.DBHandle, error) {
...
-		log.Infof("Database migration completed. Exiting due to --auto_migrate_db_and_exit.")
-		os.Exit(0)
-	}
-	if *autoMigrateDB {
+
 		if err := runMigrations(driverName, primaryDB); err != nil {
 			return nil, err
 		}
+
+		if *autoMigrateDBOutputFile != "" {
```
tylerwilliams, line 672: WDYT about just printing the changes to stdout below some marker line?

Like:

"Schema changes follow:"
"CREATE TABLAH"
"INSERT TABLAH"
"ALTER TABLAH"

or

"No schema changes necessary."

Then this code becomes simpler (yay) and your calling script can just check for the presence of "No schema changes necessary." and bail if it doesn't see that. Or alternatively, later, it could include the schema changes in a message to us saying please fix or whatever.

tools/gorm/compare_migrations.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
```
tylerwilliams, line 1: do we need this anymore?

tylerwilliams (changes requested):

server/util/db/db.go
maggie-lou: I was thinking it would be nice if the bash script could easily parse the schema changes, so it could do something like send a slack message with the schema changes, and people could use a slackbot to still continue the auto-deploy depending on the changes

We could still do this if we just printed the changes to stdout, but then the bash script would have to do some additional (likely gross) regex parsing to extract the schema changes. So slight preference for the output file because I don't think it adds too much complexity, but don't feel strongly if you do

tylerwilliams: Ack -- I understand what you're saying here. One thing that's nice about printing this stuff to stdout is that then it's easier to compose this command in bash and you don't need to worry about figuring out a file path. It's more unix-y if that makes sense. For instance, you could do:

```
SCHEMA_DIFF = $(/path/to/buildbuddy --print_schema_diff_and_exit)
if [[ $SCHEMA_DIFF ]]; then
  echo "no changes yay"
else
  slack_send $SCHEMA_DIFF
fi
```

I suggested printing lines before the schema change in my previous comment because I thought we might be printing stuff to stdout and needed a way to distinguish, but I checked and logging etc is all going to stderr, so you can just print the schema to sdout and then exit. So the following would dump the schema to just a file without log changes:

```
/path/to/buildbuddy --print_schema_diff_and_exit > schema_changes.diff 2>/dev/null
```

The other thing is that it's *nice* to have all the shit code be in bash -- it's sort of what bash is meant for :)

server/util/db/db.go
```diff
@@ -73,8 +74,9 @@ var (
 	logQueries             = flag.Bool("database.log_queries", false, "If true, log all queries")
 	slowQueryThreshold     = flag.Duration("database.slow_query_threshold", 500*time.Millisecond, "Queries longer than this duration will be logged with a 'Slow SQL' warning.")
 
-	autoMigrateDB        = flag.Bool("auto_migrate_db", true, "If true, attempt to automigrate the db when connecting")
-	autoMigrateDBAndExit = flag.Bool("auto_migrate_db_and_exit", false, "If true, attempt to automigrate the db when connecting, then exit the program.")
+	autoMigrateDB           = flag.Bool("auto_migrate_db", true, "If true, attempt to automigrate the db when connecting")
+	autoMigrateDBAndExit    = flag.Bool("auto_migrate_db_and_exit", false, "If true, attempt to automigrate the db when connecting, then exit the program.")
+	autoMigrateDBOutputFile = flag.String("database.auto_migrate_db_output_file", "", "If set, print auto-migration SQL statements to this file.")
```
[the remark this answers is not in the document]
maggie-lou: Done

server/util/db/db.go
maggie-lou: Good point about logs going to stderr - that resolves my earlier concerns! Changed to print to stdout

tools/gorm/compare_migrations.sh
maggie-lou: Removed

server/util/db/db.go
```diff
@@ -650,17 +651,41 @@ func GetConfiguredDatabase(env environment.Env) (interfaces.DBHandle, error) {
...
+
 		if err := runMigrations(driverName, primaryDB); err != nil {
 			return nil, err
 		}
+
+		if *printSchemaChangesAndExit {
+			// Print schema changes to stdout
+			// Logs go to stderr, so this output will be easy to isolate and parse
+			if len(sqlStrings) > 0 {
+				fmt.Println("Auto-migration schema changes:")
```
tylerwilliams, line 675: you could just remove this (and below) if you want the stdout to *only* be schema changes or nothing if there are none?

tylerwilliams (approved):
one nit

PR #3422, opened 2023-02-15
Safer / better bytestream file streaming

tylerwilliams:


server/bytestream/bytestream.go
```diff
@@ -135,19 +136,19 @@ func StreamBytestreamFileChunk(ctx context.Context, env environment.Env, url *ur
 		}
 		localURL.Host = "localhost:" + grpcPort
 		err = streamFromUrl(ctx, localURL, false, offset, limit, writer)
-	}
-
-	// If that fails, try to connect over grpcs
-	if err != nil || env.GetCache() == nil {
+	} else {
```
tylerwilliams, line 139: @siggisim 

(I'm not married to the control flow change here, but the code before was a little funny)

The big change here is that if we have a cache configured, we'll always hit that rather than parsing the bytestream URL. If we think this will break existing use cases I can revert this piece.

tylerwilliams (commented):
reverted control flow changes; ptal

siggisim (approved):

PR #3423, opened 2023-02-15
Add unit tests for authdb

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -80,7 +80,7 @@ func (d *AuthDB) GetAPIKeyGroupFromAPIKey(ctx context.Context, apiKey string) (i
 	})
 	if err != nil {
 		if db.IsRecordNotFound(err) {
-			return nil, status.UnauthenticatedErrorf("Invalid API key %q", redactInvalidAPIKey(apiKey))
+			return nil, status.UnauthenticatedErrorf("Invalid API key %q", redactInvalidToken(apiKey))
 		}
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


enterprise/server/backends/authdb/authdb_test.go
```diff
@@ -0,0 +1,235 @@
...
+	require.Truef(
+		t, status.IsUnauthenticatedError(err),
+		"expected Unauthenticated error; got: %v", err)
+}
+
+func requireRecordNotFoundError(t *testing.T, err error) {
+	require.Truef(
+		t, db.IsRecordNotFound(err),
+		"expected RecordNotFound error; got: %v", err)
+}
```
tylerwilliams, line 235: These tests would be clearer without these methods -- they sort of obfuscate what's going on and don't really add a lot of additional value.

enterprise/server/backends/authdb/authdb.go
```diff
@@ -113,16 +113,15 @@ func (d *AuthDB) GetAPIKeyGroupFromBasicAuth(ctx context.Context, login, pass st
 		qb.AddWhereClause(`g.write_token = ?`, pass)
 		q, args := qb.Build()
 		existingRow := tx.Raw(q, args...)
-		return existingRow.Scan(akg).Error
+		return existingRow.Take(akg).Error
 	})
 	if err != nil {
 		if db.IsRecordNotFound(err) {
-			return nil, status.UnauthenticatedErrorf("User/Group specified by %s:%s not found", login, pass)
+			return nil, status.UnauthenticatedErrorf("User/Group specified by %s:%s not found", login, redactInvalidToken(pass))
```
vadimberezniker, line 120: Let's remove the password entirely from the message. Even with redaction, it can give away enough information.

enterprise/server/backends/authdb/authdb_test.go
bduffany: Removed

vadimberezniker (approved):

enterprise/server/backends/authdb/authdb.go
bduffany: Done

tylerwilliams (approved):

PR #3427, opened 2023-02-16
Add explicit check for empty group ID

bduffany:
Users should never have a group with an empty GroupID in their `GetGroupMemberships()` list, so `authutil.AuthorizeGroupRole` should always return "You do not have access to the requested organization" today if an empty GroupID is passed in.

This PR just makes the empty group ID case a bit more explicit, to give an even stronger guarantee that we always return an error if the group ID is empty.

---

**Version bump**: Patch


tylerwilliams (approved):
Could also use InvalidArgumentError here if that's more appropriate; your call tho!

bduffany:
> Could also use InvalidArgumentError here if that's more appropriate; your call tho!

I feel mildly better about keeping the existing error code since I'm not sure if anything depends on it :sweat_smile:

vadimberezniker (approved):

PR #3428, opened 2023-02-16
Read keys across multiple versions

tylerwilliams:


tylerwilliams:
Yeah I have a followup CL (also a revert) that changes eviction to work based on bytes rather than versioned keys, lemme send you that one too.

vadimberezniker:
> Yeah I have a followup CL (also a revert) that changes eviction to work based on bytes rather than versioned keys, lemme send you that one too.

Ok, can you address the one in backgroundRepairIteration?

vadimberezniker (approved):

tylerwilliams:
Yup, will do

PR #3430, opened 2023-02-16
Change UpdateGroupUsers to use group_id from request context

bduffany:
Depends on https://github.com/buildbuddy-io/buildbuddy/pull/3411/files

---

**Version bump**: Patch


bduffany:
Closing to remove from PR list - can re-open later if we deem it relevant.

PR #3434, opened 2023-02-17
use bb instead of bazel in post_bazel.py

luluz66:
Without this change, someone who installed bb but not bazel
encounters error when use bb to fix dependencies.

### DIFF
--- a/cli/plugins/go-deps/post_bazel.py
+++ b/cli/plugins/go-deps/post_bazel.py
@@ -51,7 +51,7 @@ def main():
 
     with tempfile.NamedTemporaryFile() as run_script:
         print(
-            "\x1b[90m> bazel run //:gazelle -- "
+            "\x1b[90m> bb run //:gazelle -- "
             + "".join(packages)
             + "\x1b[m  🛠️  fixing...",
             end=""
@@ -59,7 +59,7 @@ def main():
         sys.stdout.flush()
         p = subprocess.run(
             [
-                "bazel",
+                "bb",
                 "run",
                 "--script_path=" + run_script.name,
                 "//:gazelle",
@@ -73,7 +73,7 @@ def main():
         if p.returncode != 0:
             erase_current_line()
             print(
-                "\x1b[32m> bazel run //:gazelle -- "
+                "\x1b[32m> bb run //:gazelle -- "
                 + "".join(packages)
                 + "\x1b[m  ❌ fix failed"
             )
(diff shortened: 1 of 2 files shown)


bduffany (changes requested):
This is a bit tricky because the user might be running bb via `bazelisk`, and not actually have `bb` in their `$PATH`. The way it works is that users can write `buildbuddy-io/5.0.3` in `.bazelversion`, and bazelisk will download "bazel" (which is actually the bb CLI release hosted [here](https://github.com/buildbuddy-io/bazel/releases/tag/5.0.3))

I think one thing we could do is create a temporary folder while the CLI is running, like `/tmp/bb-tools-${RANDOM}`, which contains symlinks `bb`, `bazelisk`, and `bazel`, which all point to `os.Args[0]` (the bb CLI). Then we'd invoke plugins with that PATH taking priority, like `PATH=/tmp/bb-tools-${RANDOM}:$PATH`. Maybe @siggisim has thoughts about this too

siggisim:
> This is a bit tricky because the user might be running bb via `bazelisk`, and not actually have `bb` in their `$PATH`. The way it works is that users can write `buildbuddy-io/5.0.3` in `.bazelversion`, and bazelisk will download "bazel" (which is actually the bb CLI release hosted [here](https://github.com/buildbuddy-io/bazel/releases/tag/5.0.3))
> 
> I think one thing we could do is create a temporary folder while the CLI is running, like `/tmp/bb-tools-$RANDOM`, which contains symlinks `bb`, `bazelisk`, and `bazel`, which all point to `os.Args[0]` (the bb CLI). Then we'd invoke plugins with that PATH taking priority, like `PATH=/tmp/bb-tools-$RANDOM:$PATH`. Maybe @siggisim has thoughts about this too

Could we pass the path to the current binary (i.e. os.Arg[0]) as an environment variable to plugins and have them use that?

cli/plugins/go-deps/post_bazel.py
```diff
@@ -49,17 +50,22 @@ def main():
         print("$BUILD_WORKSPACE_DIRECTORY is not set; exiting.", file=sys.stderr)
         return 1
 
+    bb_binary_path = os.environ.get("BB_EXECUTABLE", "")
```
bduffany, line 53: nit: `s/bb_binary_path/bb_executable/g`

bduffany (approved):

PR #3438, opened 2023-02-17
Add a util package for length-delimited proto streams

bduffany:
* Add a package `protodelim` for consuming length-delimited proto streams. The name is inspired by the discussion [here](https://github.com/golang/protobuf/issues/1382) and also follows a similar naming convention to proto packages that deal with proto encodings (`protojson`, `protowire`, etc.)
* Use the package in `cli/printlog` and `server/util/protofile`. Will also be used in https://github.com/buildbuddy-io/buildbuddy/pull/3386
This should also reduce the amount of allocations in `ReadProto` (in `protofile.go`) since the new util reuses the same buffer across all unmarshal operations.

---

**Version bump**: Patch

### DIFF
--- a/cli/printlog/BUILD
+++ b/cli/printlog/BUILD
@@ -9,6 +9,7 @@ go_library(
         "//cli/arg",
         "//cli/log",
         "//proto:remote_execution_log_go_proto",
+        "//server/util/protodelim",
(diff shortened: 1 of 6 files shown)


bduffany:
Closing in favor of `google.golang.org/protobuf/encoding/protodelim`

PR #3440, opened 2023-02-17
Read keys at multiple versions

tylerwilliams:
Once more with feeling!

Also fixed a small race condition that jim drew my attention to:

```
		unlockFn := p.locker.RLock(key.LockID())
        	found, err := p.iterHasKey(iter, key)
                if err != nil {
                        return nil, err
                }
                if !found {
                        missing = append(missing, r.GetDigest())
                }
		unlockFn()
```

in the error case above, RLock is never unlocked


jdhollen (approved):
I will admit I only looked at the diff with what you rolled back, but that looks correct to me.

PR #3444, opened 2023-02-17
Make QueryBuilder auto-parenthesize WHERE clauses

bduffany:
This way, if people write code like `.AddWhereClause("a.b = ? OR c.d = ?")` and then later `.AddWhereClause("x.y = ?")`, the query will be correctly generated as `(a.b = ? OR c.d = ?) AND (x.y = ?)`

Also add a `normalize` func to help normalize whitespace (for test assertions only), since I was going crazy trying to check the differences in all the generated queries.

---

**Version bump**: Patch

### DIFF
--- a/server/util/query_builder/query_builder_test.go
+++ b/server/util/query_builder/query_builder_test.go
@@ -1,12 +1,45 @@
 package query_builder_test
 
 import (
+	"regexp"
 	"strings"
 	"testing"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/query_builder"
 	"github.com/stretchr/testify/assert"
 )
+
+// normalize adjusts whitespace so that if two queries are syntactically
+// equivalent then they should have the same normalized representation. This
+// makes test assertions easier.
+func normalize(t *testing.T, query string) string {
+	// The current logic in this func blindly strips whitespace, which might
+	// change the meaning of queries containing strings. So for now, just fail
+	// the test if we see any string delimiters.
+	if strings.Contains(query, "\"") || strings.Contains(query, "'") || strings.Contains(query, "`") {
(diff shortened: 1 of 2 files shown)


server/util/query_builder/query_builder_test.go
```diff
@@ -1,13 +1,44 @@
...
 	"github.com/stretchr/testify/assert"
 )
 
+// normalize applies a few simple normalization to make query assertions easier.
+func normalize(query string) string {
+	// The current logic in this func assumes there are no strings in order for
+	// the meaning to be equivalent, since it strips whitespace without
+	// considering whether the spaces are inside strings or not.
+	if strings.Contains(query, "\"") || strings.Contains(query, "'") || strings.Contains(query, "`") {
+		panic("normalizeSpace cannot yet handle string or backtick literals")
```
luluz66, line 18: instead of panic. can we do t.Error? or return an error. and use require.NoError at the caller site?

server/util/query_builder/query_builder_test.go
```diff
@@ -1,13 +1,44 @@
...
+	"regexp"
 	"strings"
 	"testing"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/query_builder"
 	"github.com/stretchr/testify/assert"
 )
 
+// normalize applies a few simple normalization to make query assertions easier.
+func normalize(query string) string {
```
luluz66, line 13: Thank you! This is great! make the test more readable

server/util/query_builder/query_builder_test.go
bduffany: Done

luluz66 (approved):

vadimberezniker (approved):

PR #3448, opened 2023-02-18
Fix up commit & branch links

siggisim:
Currently the links on commit history just take you to a github search for the commit sha - which gives you pretty lame results.

With this change - we'll link directly to the commit or branch if it's a github repo. Otherwise we won't show the button.

We can later expand this logic to support other providers.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/history/history.tsx
+++ b/enterprise/app/history/history.tsx
@@ -17,7 +17,7 @@ import HistoryInvocationStatCardComponent from "./history_invocation_stat_card";
 import { ProtoFilterParams, getProtoFilterParams } from "../filter/filter_util";
 import Spinner from "../../../app/components/spinner/spinner";
 import Long from "long";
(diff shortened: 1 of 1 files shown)


enterprise/app/history/history.tsx
```diff
@@ -415,15 +429,15 @@ export default class HistoryComponent extends React.Component<Props, State> {
                   <span>
                     <span className="history-title">{this.props.username}'s builds</span>
                     <a className="history-button" href={`/trends/?user=${this.props.username}`}>
-                      <BarChart2 /> Trends
+                      <BarChart2 /> View Trends
```
bduffany, line 432: nit: do we want to make these `Sentence case` like we do for other labels?

bduffany (approved):
(LGTM modulo nit)

enterprise/app/history/history.tsx
siggisim: I went back and forth on this one, but ok - went with sentence case

PR #3449, opened 2023-02-18
Run `bb fix` on repo with gazelle typescript enabled

siggisim:
How I generated this PR:
- Added these two directives to the root BUILD file to pick up our custom ts_library rule and avoid running on the node modules directory:
```
# gazelle:map_kind ts_project ts_library //rules/typescript:index.bzl
# gazelle:exclude **/node_modules/**
```
- Removed any glob srcs (gazelle doesn't like these https://github.com/bazelbuild/bazel-gazelle/issues/1036)
- Ran `bb fix` with our typescript gazelle extension enabled
- Resolved 1 cyclical dependency in `app/invocation/child_invocation_card.tsx`
- Preserved strict = True labels where globbed targets were split up
- Added # keep for two strange imports  
  - `"@npm//path-browserify"` which is needed but not imported by libsodium
  - `"@npm//@types/dagre-d3"` a misnamed types package which should be `"@npm//@types/dagre-d3-react"`

Excludes website/** for now because those files import css files, which our typescript plugin doesn't yet support.

TL;DR of the changes:
- Splits globbed ts_library into one target per ts file (similar to go gazelle) - without globs targets are more granular and will have to be rebuilt less frequently
- Removes some unused dependencies

---

**Version bump**: Patch


bduffany (commented):
Amazing - this is going to save so much time :)

bduffany (approved):

PR #3451, opened 2023-02-21
Allow live-migration of key versions

tylerwilliams:


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -367,6 +391,40 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
...
+		if pc.activeDatabaseVersion() >= filestore.Version2 {
+			// Migrate keys from 1->2.
+			pc.migrators = append(pc.migrators, &v1ToV2Migrator{})
+		}
+	}
+
+	// Check that there is a migrator enabled to update us to (or past) the
+	// activeKeyVersion (flag configured). Warn if not.
+	if len(pc.migrators) > 0 && pc.activeDatabaseVersion() > pc.migrators[len(pc.migrators)-1].ToVersion() {
+		log.Warningf("Cache versions will never converge!")
```
vadimberezniker, line 426: Any reason not to make this a fatal error?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -566,6 +629,110 @@ func (p *PebbleCache) updateAtime(key filestore.PebbleKey) error {
...
+
+		select {
+		case <-quitChan:
+			return nil
+		default:
+		}
+		_ = limiter.Wait(p.env.GetServerContext())
+
+		if time.Since(lastStatusUpdate) > 10*time.Second {
+			log.Infof("Pebble Cache: data migration progress: saw %d keys, migrated %d to version: %d in %s", keysSeen, keysMigrated, maxVersion, time.Since(migrationStart))
```
vadimberezniker, line 672: Can you add iter.Key() to the log message? Since the keys are sorted in Pebble, it gives you some idea of how far along the scan is.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -566,6 +629,110 @@ func (p *PebbleCache) updateAtime(key filestore.PebbleKey) error {
...
+			if version < minVersion {
+				minVersion = version
+			}
+
+		}
+		if version == oldVersion {
+			continue
+		}
+
+		keyBytes, err := key.Bytes(version)
```
vadimberezniker, line 713: I think you need to lock the key here. Someone could do a read-then-modify operation on the key you are migrating and unintentionally restore it.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -566,6 +629,110 @@ func (p *PebbleCache) updateAtime(key filestore.PebbleKey) error {
...
+			if version != migrator.FromVersion() {
+				return status.FailedPreconditionErrorf("Migrator %+v cannot migrate key from version %d", migrator, version)
+			}
+
+			valBytes = migrator.Migrate(valBytes)
+			version = migrator.ToVersion()
+			keysMigrated += 1
+
+			if version > maxVersion {
+				maxVersion = version
```
vadimberezniker, line 702: This logic seems off. It looks like you can end up with the wrong versions at the end of the migration. Let's say you set write version to v2 and have two migrators, v0->v1 and v1->v2. 
After the loop, you will have minVersion=v1 and maxVersion=v2 where's what you want is minVersion=v2 since there should be no v1 keys at that point.

vadimberezniker (changes requested):

tylerwilliams:
PTAL

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -368,6 +392,43 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
...
+			pc.migrators = append(pc.migrators, &v1ToV2Migrator{})
+		}
+	}
+
+	// Check that there is a migrator enabled to update us to (or past) the
+	// activeKeyVersion (flag configured). Warn if not.
+	if len(pc.migrators) > 0 {
+		lastMigratorVersion := pc.migrators[len(pc.migrators)-1].ToVersion()
+		if pc.activeDatabaseVersion() > lastMigratorVersion {
+			log.Fatalf("Cache versions will never converge! Active key version %d > last migrator version: %d", pc.activeDatabaseVersion(), lastMigratorVersion)
```
vadimberezniker, line 429: NBD, but this func already has an error return type so you could return an error here and let the caller call log.Fatal

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2322,6 +2503,9 @@ func (p *PebbleCache) Start() error {
 	p.eg.Go(func() error {
 		return p.backgroundRepair(p.quitChan)
 	})
+	p.eg.Go(func() error {
+		return p.migrateData(p.quitChan)
```
vadimberezniker, line 2507: let's add a check here for err != nil + alert.UnexpectedEvent so we know if a migration did not finish successfully

vadimberezniker (approved):

PR #3453, opened 2023-02-21
Explicitly reject invalid keys

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/test/integration/remote_cache/remote_cache_test.go
+++ b/enterprise/server/test/integration/remote_cache/remote_cache_test.go
@@ -296,6 +296,50 @@ func TestBuild_RemoteCache_ScoreCard(t *testing.T) {
 	}
 }
 
+func TestBuild_RemoteCache_RejectsInvalidAPIKeys(t *testing.T) {
+	app := buildbuddy_enterprise.Run(t, "--auth.enable_anonymous_usage=true")
+	ctx := context.Background()
+	ws := testbazel.MakeTempWorkspace(t, workspaceContents)
+
+	{
+		// explicit empty API key; should fail
+		testbazel.Clean(ctx, t, ws)
+		iid := newUUID(t)
+		buildFlags := []string{
+			"//:hello.txt", "--invocation_id=" + iid,
+			"--remote_header=x-buildbuddy-api-key=",
+			"--remote_upload_local_results"}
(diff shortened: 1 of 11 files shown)


enterprise/server/test/integration/remote_cache/remote_cache_test.go
```diff
@@ -296,6 +296,37 @@ func TestBuild_RemoteCache_ScoreCard(t *testing.T) {
...
+	ctx := context.Background()
+	ws := testbazel.MakeTempWorkspace(t, workspaceContents)
+
+	{
+		// Build 1: explicit empty API key
+		testbazel.Clean(ctx, t, ws)
+		iid := newUUID(t)
+		buildFlags := []string{
+			"//:hello.txt", "--invocation_id=" + iid,
+			"--remote_header=x-buildbuddy-api-key=",
```
siggisim, line 310: Should we add a test to make sure that non-explicitly set empty keys (i.e. omitted keys) don't fail?

bduffany: Good call, this case is tested above, but not intentionally. Done

tylerwilliams (approved):

server/util/prefix/prefix.go
```diff
@@ -20,9 +21,14 @@ func addPrefix(prefix, key string) string {
...
-		if u, err := auth.AuthenticatedUser(ctx); err == nil {
-			if u.GetGroupID() == "" {
-				return "", status.PermissionDeniedErrorf("Attempting to write to cache as a user with no group.")
+		u, err := auth.AuthenticatedUser(ctx)
+		if err != nil {
+			if authutil.IsInvalidCredentialsError(err) {
+				return "", err
+			}
+		} else {
+			if u.GetGroupID() == "" && !env.GetAuthenticator().AnonymousUsageEnabled() {
```
vadimberezniker, line 30: If we get here, doesn't it mean the user is not anonymous? In which case it makes sense to keep the old check, otherwise we will could call addPrefix with an empty group ID.

vadimberezniker (changes requested):

server/util/prefix/prefix.go
bduffany: Reverted back to the old logic in this case.

PR #3455, opened 2023-02-21
refactor: Move IsAnonymousUserError to authutil

bduffany:
Context: The `perms` package was originally intended to just be utils for dealing with `perms` bits, and I added a few unrelated funcs there a while back before I understood its original purpose and before `authutil` existed. This func was one of them :)

There are a few other misplaced funcs in here like `perms.AuthenticatedUser` -- will find better homes for those in future PRs.

---

**Version bump**: Patch

### DIFF
--- a/server/util/authutil/authutil.go
+++ b/server/util/authutil/authutil.go
@@ -30,3 +30,13 @@ func AuthorizeGroupRole(u interfaces.UserInfo, groupID string, allowedRoles role
 	}
 	return nil
 }
+
+// IsAnonymousUserError can be used to check whether an error returned by
+// functions which return the authenticated user (such as AuthenticatedUser or
+// AuthenticateSelectedGroupID) is due to an anonymous user accessing the
+// service. This is useful for allowing anonymous users to proceed, in cases
+// where anonymous usage is explicitly enabled in the app config, and we support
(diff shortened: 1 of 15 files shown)


vadimberezniker (approved):

server/util/authutil/authutil.go
```diff
@@ -30,3 +30,13 @@ func AuthorizeGroupRole(u interfaces.UserInfo, groupID string, allowedRoles role
...
 	return nil
 }
+
+// IsAnonymousUserError can be used to check whether an error returned by
+// functions which return the authenticated user (such as AuthenticatedUser or
+// AuthenticateSelectedGroupID) is due to an anonymous user accessing the
+// service. This is useful for allowing anonymous users to proceed, in cases
+// where anonymous usage is explicitly enabled in the app config, and we support
+// anonymous usage for the part of the service where this is used.
+func IsAnonymousUserError(err error) bool {
```
bduffany, line 40: happy to also rename this to `IsMissingCredentialsError` as part of this PR if people think it's clearer.

bduffany: Will follow up in a separate PR with this, if we want to do it.

PR #3456, opened 2023-02-21
Remove anonymous user fallback

bduffany:
Replaces https://github.com/buildbuddy-io/buildbuddy/pull/3453

---

**Version bump**: Patch


server/util/prefix/prefix.go
```diff
@@ -19,26 +20,24 @@ func addPrefix(prefix, key string) string {
 }
 
 func userPrefixCacheKey(ctx context.Context, env environment.Env, key string) (string, error) {
-	if auth := env.GetAuthenticator(); auth != nil {
```
bduffany, line 22: note: the nil check isn't needed because we use `nullauth` instead of having a `nil` authenticator. If this were ever `nil`, then line 31 would have panicked in the current implementation.

tylerwilliams: Wanna put this in a comment?

tylerwilliams (approved):

server/util/authutil/authutil.go
```diff
@@ -31,12 +44,41 @@ func AuthorizeGroupRole(u interfaces.UserInfo, groupID string, allowedRoles role
...
+		if err := detail.UnmarshalTo(info); err != nil {
+			// not an ErrorInfo detail; ignore.
+			continue
+		}
+		if info.GetReason() == missingCredentialsErrorReason {
+			return true
+		}
+	}
+	return false
 }
```
bduffany, line 84: note: I changed this implementation to check for an explicit error status created by `AnonymousUserError`, so that we can have stronger guarantees about anonymous user errors vs. other types of errors that we happen to be setting to "Unauthenticated." I felt this is important since now we are leaning more heavily on AnonymousUserError in the cache prefix code, but let me know if I am overthinking it and if it seems fine to just use the old check instead.

server/util/prefix/prefix.go
bduffany: Done

PR #3458, opened 2023-02-22
Hackathon: open telemetry

sluongng:
Provide an exporter that would export BuildBuddy invocations as otel
events.

---

**Version bump**: None


sluongng:
After discussing with the team, we would want to explore implementing this as a webhooks background worker in the future https://github.com/buildbuddy-io/buildbuddy/blob/ede1e1d07122a3bdd22a61b31dd7c6a8d8bd458d/server/build_event_protocol/webhooks/webhooks.go

PR #3461, opened 2023-02-24
Use go.work

sluongng:
When a user sync the generated go code from Bazel's output into
//proto/..., 'go mod tidy -e' command output would change as Go toolings
start to detect //proto sub-directories as a go packages.

This would move some of the indirect dependencies to direct
dependencies, namely 'github.com/golang/protobuf'. This made the linter
result to be inconsistent between local development and CI.

Go toolings does not provide a way to do a `ignore dir`. The only method
is to create a go.mod file under that dir.

  touch proto/go.mod

This approach enable us to run ./tools/fix_go_deps.sh correctly, however
it would tell gopls, the go language server to also ignore //proto which
mess with out editor work.

For these reasons, let's make use of Go Workspace feature.

  go work init
  go work use -r .
  go work sync

This let gopls know that we are working in a multi-modules environment
and be aware of multiple go.mod files. We could also ensure that we are
using consistent version between different go.mod files using the
'go work sync' command.

---

**Version bump**: None


sluongng:
It seems like the inclusion of `edit_instance_template` require a bump in `go_googleapis`.
let me address this in a different PR and rotate back to this later

PR #3462, opened 2023-02-24
sluongng/upgrade googleapis

sluongng:
GoogleAPIs has been notiriously difficult to upgrade.
The reason being rules_go manage it's own version of GoogleAPIs package and Gazelle is
configured to map all GoogleAPI compilation to the version pinned under rules_go.

This means that unless rules_go were to upgrade their GoogleAPIs version, everybody else
is stuck at the same version as theirs.

We resolve this by doing a couple things:

1. Override rules_go's `go_googleapis` archive with our own version. Generating from
   a relatively new commit.

2. Upgrade our go.mod and go.sum to latest commit by using `go get -u <pkg_name>@main`
   where `main` is the default branch of the package's repository.

3. Override Gazelle resolve directives to use the go_repository version during compilation where possible.

4. Switch dependencies to //proto to consistently link all of our Go binary using the right version.

At the end, we manage to build our repository using a newer, self-managed version of go_googleapis.

To demonstrate, remove the restriction from building `tools/edit_instance_template`
that was put there earlier due to difficulties in upgrading `go_googleapis`.

I expect this to help unblocking us from using latest dependencies and features in the go ecosystem.

---

**Version bump**: TODO


sluongng:
Should help unblock https://github.com/buildbuddy-io/buildbuddy/pull/3462 and https://github.com/buildbuddy-io/buildbuddy/pull/3458. But overall low priority.

tylerwilliams:
lol this is amazing, but also pretty hairy. could we get rules_go to update their version of google APIs?

sluongng:
> could we get rules_go to update their version of google APIs?

They tried it, and it broke the latest release so there was a revert and patch release after https://github.com/bazelbuild/rules_go/pull/3432

sluongng:
I could instead look into bazelbuild/rules_go#1986 where the plan is to replace go_googleapis with com_google_googleapis.

But will tackle that slowly as it's probably gonna require some change coordination between gazelle and rules_go. 🤔

sluongng:
let's close this as we wait for a proper fix in rules_go / gazelle.

See https://bazelbuild.slack.com/archives/CDBP88Z0D/p1677518730474199 and https://github.com/wyattanderson/googleapis-reproducer for more information.

PR #3463, opened 2023-02-24
vmexec: mark test as flaky

sluongng:
Discovered in #3462 that this is flaky

---

**Version bump**: None


sluongng:
could not reproduce 🤔 

Not sure if the error margin is low enough that it does not matter. Let's close for now.

Issue #3464, opened 2023-02-24
Janitor fails to delete blobs because directory is not empty


JaredNeil:
```WRN Error deleting blob (03516230-c019-4f85-8283-1059f6512a4a): remove /buildbuddy/03516230-c019-4f85-8283-1059f6512a4a: directory not empty```

I think this has something to do with how `Chunkstore` and `DiskBlobStore` interact, but I think it comes down to 
`DiskBlobStore.DeleteBlob` calls `disk.DeleteFile`, which calls `os.Remove`, which always fails because the blob is actually a directory, and not empty.

tylerwilliams:
Hi @JaredNeil, sorry for the delay on this, but the fix is now submitted and will go out in next week's release :)

PR #3465, opened 2023-02-24
Delete blob dirs too

tylerwilliams:
Delete the blob directory if it's a dir.

Fixes: https://github.com/buildbuddy-io/buildbuddy/issues/3464


server/backends/blobstore/blobstore.go
```diff
@@ -265,7 +265,16 @@ func (d *DiskBlobStore) DeleteBlob(ctx context.Context, blobName string) error {
 	}
 	start := time.Now()
 	ctx, spn := tracing.StartSpan(ctx)
-	err = disk.DeleteFile(ctx, fullPath)
+
+	fileInfo, err := os.Stat(blobName)
+	if err != nil {
```
bduffany, line 270: should we move the `IsNotExist` check up into this `err != nil` block ? since before we would ignore IsNotExist errs

tylerwilliams: Done.

siggisim (approved):

PR #3466, opened 2023-02-24
Better interface for snaploader

tylerwilliams:
Make an interface; implement it. This will let me swap out snaploader with a different implementation.


bduffany (approved):
LGTM, assuming firecracker_test still passes

PR #3470, opened 2023-02-27
Have executor request re-enqueue as part of closing the lease.

vadimberezniker:
Previously it was possible for a task to get stuck in limbo if the executor succesfully released the lease but was not able to contact the scheduler to re-enqueue the task.

---

**Version bump**: Patch

### DIFF
--- a/proto/scheduler.proto
+++ b/proto/scheduler.proto
@@ -1,5 +1,6 @@
 syntax = "proto3";
 
+import "google/rpc/status.proto";
 import "google/protobuf/timestamp.proto";
 import "proto/acl.proto";
 import "proto/context.proto";
@@ -24,6 +25,7 @@ message LeaseTaskRequest {
   // Mutually exclusive with `release`.
   bool finalize = 2;
 
+  // DEPRECATED
   // Indicates that the lease should be released without finalizing (deleting)
   // the task.
   // Mutually exclusive with `finalize`.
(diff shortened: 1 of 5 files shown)


proto/scheduler.proto
```diff
@@ -24,13 +24,21 @@ message LeaseTaskRequest {
...
 
   // ID of the executor making the request.
   string executor_id = 4;
+
+  // Indicates that the leased task could not be run to completition and should
+  // be re-enqueued to be retried.
+  bool re_enqueue = 5;
+  // Optional description of why the task needs to be re-enqueued (may be
+  // visible to end user).
+  string re_enqueue_reason = 6;
```
tylerwilliams, line 41: Could we use a proto.Status here so we get the code too?

proto/scheduler.proto
```diff
@@ -24,13 +24,21 @@ message LeaseTaskRequest {
   // Mutually exclusive with `release`.
   bool finalize = 2;
 
+  // DEPRECATED
```
tylerwilliams, line 27: In a followup can we remove and mark this field as reserved?

tylerwilliams (approved):

bduffany (approved):

proto/scheduler.proto
vadimberezniker: Done.

proto/scheduler.proto
vadimberezniker: We can't do this just yet because of user owned runners. I will go check if people are upgrading their executors.

PR #3474, opened 2023-02-28
Add script that prints clickhouse auto migration diff

maggie-lou:
---

**Version bump**: TODO


maggie-lou:
* master
  * **PR #3421** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3421" target="_blank">Graphite</a> 
    * **PR #3474** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3474" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3474?utm_source=stack-comment).

server/tables/tables.go
```diff
@@ -996,6 +996,34 @@ func isStrictModeEnabled(db *gorm.DB) (bool, error) {
 	return isStrictModeEnabled, nil
 }
 
+// RegisterLogSqlCallback adds a callback to the db that appends all SQL executed to a slice of strings
+func RegisterLogSqlCallback(db *gorm.DB, s *[]string, excludeQueries bool) error {
```
luluz66, line 1000: nit: SQL instead of Sql

server/tables/tables.go
```diff
@@ -996,6 +996,34 @@ func isStrictModeEnabled(db *gorm.DB) (bool, error) {
...
+	if err := db.Callback().Raw().Register("save_sql", func(db *gorm.DB) {
+		executedSql := db.Statement.SQL.String()
+		if excludeQueries && strings.HasPrefix(executedSql, "SELECT") {
```
luluz66, line 1003: same here

maggie-lou:
There's no built-in tool to copy the schema of a clickhouse table, which is why I created a separate script for clickhouse. In order to copy the schema of the dev db, this script runs the gorm auto-migration, and assumes that will match the schema in dev. It's not as guaranteed it will match the schema exactly though

If people think there should be one script / we should be more accurate in matching the exact dev schema, I can write a script that parses clickhouse `SHOW CREATE TABLE` output so we can reuse the other print_differences script

server/tables/tables.go
```diff
@@ -996,6 +996,30 @@ func isStrictModeEnabled(db *gorm.DB) (bool, error) {
 	return isStrictModeEnabled, nil
 }
 
+// RegisterLogSQLCallback adds a callback to the db that appends all SQL executed to a slice of strings
+func RegisterLogSQLCallback(db *gorm.DB, s *[]string, excludeQueries bool) error {
```
luluz66, line 1000: can we move these two functions to some other libraries? for example, gormutil.go?

it feels weird to me that clickhouse.go is depended on tables.go (which is mysql table definition)

tools/gorm/print_schema_changes.sh
```diff
@@ -10,9 +10,32 @@
 
 git_branch=$1
 db_conn_string=$2
+# TODO: Remove when buildbuddy-internal#2096 is fixed
```
maggie-lou, line 13: I simplified the script in https://github.com/buildbuddy-io/buildbuddy/pull/3508, so rebased this clickhouse change off those simplications.

maggie-lou:
I realized I could simplify this script by fixing something in the gorm clickhouse driver. It's taking some time to get that change approved, so I added a local buildpatch in this PR. @luluz66 could you please take another look at this?

My PR to gorm-clickhouse with more context on the problem: https://github.com/go-gorm/clickhouse/pull/91

As a reminder, currently clickhouse gorm autoMigrate always runs column migrations, even if there aren't any schema changes (the reason for that is describe in the PR above). In addition to being unnecessary, it also makes this schema change script less useful, because it appears as if we're always running a migration, and makes it harder to detect when we're making a meaningful schema change. Applying this patch will fix both of those problems

server/util/gormutil/gormutil.go
```diff
@@ -22,3 +27,28 @@ func InstrumentMetrics(gdb *gorm.DB, beforeKey string, beforeFn func(*gorm.DB),
...
+	}
+	return nil
+}
+
+func PrintMigrationSchemaChanges(sqlStrings []string) {
+	// Print schema changes to stdout
+	// Logs go to stderr, so this output will be easy to isolate and parse
+	if len(sqlStrings) > 0 {
+		for _, sqlStr := range sqlStrings {
+			fmt.Printf("%s\n", sqlStr)
```
luluz66, line 51: nit: fmt.Println()

luluz66 (approved):

maggie-lou:
Fixes buildbuddy-io/buildbuddy-internal#2096

PR #3480, opened 2023-03-06
bazel: fix //enterprise package visibility

sluongng:
We organize `//enterprise/...` code to have a different license compare to
code in `//... -//enterprise/...`. Yet, there are packages outside of
`enterprise` dir that depends on the code inside `enterprise` directory.

This PR is the first step in correcting that problem. Ensure that all
code within `//enterprise/...` has a strict visibility set to
`//enterprise:__subpackages__`. Exceptions are strictly stated so that
we could query for them in the future and apply appropriate refactoring.

Before this PR
```
> bazel query 'visible(//..., //enterprise/...)' | wc -l
29
```

After this PR
```
> bazel query 'visible(//..., //enterprise/...)' | wc -l
4
```

---

**Version bump**: None

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,4 +1,4 @@
-load("@bazel_gazelle//:def.bzl", "gazelle")
+load("@bazel_gazelle//:def.bzl", "DEFAULT_LANGUAGES", "gazelle", "gazelle_binary")
 load("@io_bazel_rules_go//go:def.bzl", "go_library", "nogo")
 load("@npm//@bazel/typescript:index.bzl", "ts_config")
 load("//rules/go:index.bzl", "go_sdk_tool")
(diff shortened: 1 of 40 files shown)


tylerwilliams (commented):
Thanks for cleaning this up! We've done this several times and it always regresses over time as we add new stuff.

@siggisim mentioned there was a gazelle directive for visibility -- could we use that in the BUILD file a the root of the enterprise/ subdir to ensure that all new code created there will be visibile to enterprise:__subpackages__ only?

sluongng:
My latest commit added Gazelle's new language plugin `bazel/visibility` to help manage the `package(default_visbility = [])`. It does have a downside which I reported in https://github.com/bazelbuild/bazel-gazelle/issues/1467, but it should not be a blocker for us.

I wonder if it's worth reinforcing the package's default_visibility with `gazelle:go_visibility`.
Since this PR is already getting quite big, I might add it in a future PR instead.

tylerwilliams (approved):

sluongng:
Rebased onto latest master to make sure nothing fail before merging.

PR #3483, opened 2023-03-06
Add remote execution tests to Mac workflow

bduffany:
Also fix broken tests.

---

**Version bump**: Patch


vadimberezniker (approved):

sluongng (approved):
Interesting.

I had similar a fix locally to commandutil package but I think yours is much better.

I think there are a few more problems in `enterprise/server/remote_execution/` but we could address them separately.

I sent https://github.com/buildbuddy-io/buildbuddy/pull/3486 for reference.

bduffany:
> I think there are a few more problems in `enterprise/server/remote_execution/` but we could address them separately.

I think these are flaky -- I will try to de-flake before submitting this PR, or just mark as flaky.

PR #3486, opened 2023-03-06
sluongng/fix macos tests

sluongng:
- Fix running most tests on darwin os.
- Enable `test //...` on CI with smaller caveats

---

**Version bump**: None

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -24,8 +24,6 @@ actions:
         test --config=mac-workflows
              --
              //...
-             -//deployment:all
-             -//tools/smarter_device_manager:push_image
              -//enterprise/server/test/integration/remote_execution/bazel_rbe:all
   - name: Benchmark
     container_image: ubuntu-20.04
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -26,6 +26,7 @@ actions:
              //...
              -//deployment:all
              -//tools/smarter_device_manager:push_image
+             -//enterprise/server/test/integration/remote_execution/bazel_rbe:all
   - name: Benchmark
     container_image: ubuntu-20.04
     triggers:
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -22,7 +22,6 @@ actions:
     bazel_commands:
       - |
         test --config=mac-workflows
-             --test_tag_filters=-performance,-webdriver,-docker
              --
              //...
              -//deployment:all
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -20,9 +20,13 @@ actions:
         branches:
           - "master"
     bazel_commands:
-      # TODO: Fix `test //...` on Mac and do that instead
-      - build //enterprise/server/cmd/executor //enterprise/server //server //cli/cmd/bb --config=mac-workflows
-      - test //enterprise/server/remote_execution/... //cli/... --config=mac-workflows --test_tag_filters=-performance,-webdriver,-docker
+      - |
+        test --config=mac-workflows
+             --test_tag_filters=-performance,-webdriver,-docker
+             --
+             //...
+             -//deployment:all
+             -//tools/smarter_device_manager:push_image
   - name: Benchmark
     container_image: ubuntu-20.04
     triggers:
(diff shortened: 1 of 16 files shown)


enterprise/server/test/integration/ci_runner/ci_runner_test.go
```diff
@@ -230,7 +230,7 @@ func TestCIRunner_Push_WorkspaceWithCustomConfig_RunsAndUploadsResultsToBES(t *t
 	app := buildbuddy.Run(t)
 	runnerFlags = append(runnerFlags, app.BESBazelFlags()...)
 
-	result := invokeRunner(t, runnerFlags, []string{}, wsPath)
+	result := invokeRunner(t, runnerFlags, []string{"USER=root", "HOME=/root"}, wsPath)
```
bduffany, line 233: I think these should not be needed due to ensureHomeDir() and ensureUser() in `ci_runner/main.go` - what happens if these are omitted?

buildbuddy.yaml
```diff
@@ -20,9 +20,12 @@ actions:
...
     bazel_commands:
-      # TODO: Fix `test //...` on Mac and do that instead
-      - build //enterprise/server/cmd/executor //enterprise/server //server //cli/cmd/bb --config=mac-workflows
-      - test //enterprise/server/remote_execution/... //cli/... --config=mac-workflows --test_tag_filters=-performance,-webdriver,-docker
+      - |
+        test --config=mac-workflows
+             --
+             //...
+             -//deployment:all
+             -//tools/smarter_device_manager:push_image
```
bduffany, line 28: does it work if we mark this `push_image` target with `tags = ["manual"]`?

enterprise/server/test/integration/remote_execution/bazel_rbe/BUILD
```diff
@@ -5,6 +5,7 @@ go_test(
     size = "small",
     srcs = ["bazel_rbe_test.go"],
     shard_count = 15,
+    target_compatible_with = ["@platforms//os:linux"],
```
bduffany, line 8: Ditto - we probably don't want to disable this

enterprise/server/test/integration/ci_runner/ci_runner_test.go
sluongng: I tried really hard to reproduce this from an old commit I found in my reflog... but could not.

Will revert!

enterprise/server/test/integration/remote_execution/BUILD
```diff
@@ -9,6 +9,7 @@ go_test(
         "//enterprise/server/test/integration/remote_execution/command:testcommand",
     ],
     shard_count = 32,
+    target_compatible_with = ["@platforms//os:linux"],
```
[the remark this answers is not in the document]
sluongng: When I tried to run these tests locally, I found that the actions were requesting `linux + amd64` executor. Tried digging around but could not find what is causing that.

> No registered executors in pool "" with os "linux" with arch "amd64"

Let me check if I could narrow down the error

enterprise/server/test/integration/remote_execution/bazel_rbe/BUILD
sluongng: ack

enterprise/server/util/vfs_server/vfs_server_test.go
```diff
@@ -67,6 +67,10 @@ func newServer(t *testing.T) (*vfs_server.Server, string) {
 }
 
 func TestGetLayout(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		// TODO(sluongng): the directory size and default symlink permission is different on MacOS
+		t.Skipf("Test assumed running on linux")
+	}
```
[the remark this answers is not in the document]
sluongng: ack

buildbuddy.yaml
sluongng: yup, that did the magic

enterprise/server/test/integration/remote_execution/BUILD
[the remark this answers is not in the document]
sluongng: Managed to narrow down the error to some certain set of tests asking for linux executors.

tylerwilliams (approved):
LGTM

PR #3491, opened 2023-03-07
Fix extract_bazel_installation rule

bduffany:
When running this rule under workflows locally, Bazel sometimes complains that the USER env var is not defined. And then when setting USER=nobody, it complains that it can't create some dir like `/nobody/.cache`. Fix by setting `--output_user_root` explicitly so Bazel doesn't try to create it in `~/.cache`.

---

**Version bump**: Patch


sluongng (approved):
Yesterday I also encountered a similar issue, not with Bazel but Go stdlib was expecting USER/HOME to be set. They were not set in MacOS process-wrapper sandbox thus causing tests to fail. Seems like a common assumption.

This fix LGTM

PR #3492, opened 2023-03-07
Fix CI runner duplicate stderr logging

bduffany:
`buildEventReporter` already copies all writes to `os.Stderr` since it wraps an `invocationLog` which sets up an stderr `MultiWriter` [here](https://cs.github.com/buildbuddy-io/buildbuddy/blob/90a109254b9a39a175b0712b17cb4b7ee626b231/enterprise/server/cmd/ci_runner/main.go#L666). So by setting up `ws.log` as `io.MultiWriter(os.Stderr, buildEventReporter)`, we are effectively duplicating the writes to stderr, resulting in confusing output. This PR fixes that.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -467,7 +467,7 @@ func run() error {
 
 	// Write setup logs to the current task's stderr (to make debugging easier),
 	// and also to the invocation.
-	ws.log = io.MultiWriter(os.Stderr, buildEventReporter)
+	ws.log = buildEventReporter
 	ws.hostname, ws.username = getHostAndUserName()
 
 	// Change the current working directory to respect WORKDIR_OVERRIDE, if set.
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -467,7 +467,7 @@ func run() error {
 
 	// Write setup logs to the current task's stderr (to make debugging easier),
 	// and also to the invocation.
-	ws.log = io.MultiWriter(os.Stderr, buildEventReporter)
+	ws.log = buildEventReporter
```
vadimberezniker, line 470: Comment is slightly out of date. Either remove it or mention that build event reporter will handle all the logging, both to the invocation and to stderr.

vadimberezniker (approved):

enterprise/server/cmd/ci_runner/main.go
bduffany: thanks - updated

PR #3493, opened 2023-03-07
Firecracker: improve `--debug_stream_command_outputs` behavior

bduffany:
In https://github.com/buildbuddy-io/buildbuddy/pull/1611, `--debug_stream_command_outputs` was repurposed to start the VM in debug mode. Debug mode streams the VM logs to the terminal, but does not actually stream any output from the commands executed (I'm not sure if this was always the case or if it broke at some point).

This PR makes 2 changes:

* Now that we have stdout/stderr streaming (in `vmexec_client.go`), `--debug_stream_command_outputs` can actually stream the command stderr/stdout (which is the original intent of the flag).
* Instead of enabling debug mode whenever `--debug_stream_command_outputs` is enabled, this PR adds a separate flag `--executor.firecracker_debug_mode` to explicitly enable debug mode. The main reason for this change is that debug mode hijacks the TTY so that Ctrl+C is handled by the VM instead of the executor, and you have to quit the whole terminal emulator to kill the executor, which is super annoying.

---

**Version bump**: Patch


iain-macdonald (approved):

vadimberezniker (approved):
Nice!

PR #3499, opened 2023-03-07
[POC] Support arbitrary workflow steps

bduffany:
Adds support for workflow steps which can contain arbitrary bash scripts as well as bazel commands. See `buildbuddy.yaml` for the config syntax.

This is just an experiment, but if we feel this is worth adding, I can polish it up and add tests.

Some TODOs:
* What to do if the user writes `run: bazel test //...` (tries to run bazel as a bash script step) instead of `bazel: test //...` (runs bazel with our supported bazelrc / output base / metadata args etc.)? Do we want to just make these fail for now (i.e. link `bazel` in PATH to a script that fails and prints an error), or should we try to write a bazel wrapper in PATH which somehow links the bazel invocations to the parent workflow?
* Display the script commands in the UI a little better (don't link to an invocation 404 page)

image

---

**Version bump**: Patch

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -8,8 +8,10 @@ actions:
       pull_request:
         branches:
           - "master"
-    bazel_commands:
-      - test //... --config=workflows --test_tag_filters=-performance,-webdriver,-docker
+    steps:
+      - run: echo 'Hello world!'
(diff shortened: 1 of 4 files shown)


tylerwilliams:
Discussed offline but I'd like to wait on this until we have a better story around handling arbitrary shell commands in workflows.

PR #3501, opened 2023-03-07
Add metric for executor registrations.

vadimberezniker:
This will allow us to keep track of what versions are out in the wild.

---

**Version bump**: None

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -662,6 +662,15 @@ var (
 		Arch,
 	})
 
+	RemoteExecutionNumExecutorRegistrations = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: bbNamespace,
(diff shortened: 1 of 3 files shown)


server/metrics/metrics.go
```diff
@@ -662,6 +662,15 @@ var (
 		Arch,
 	})
 
+	RemoteExecutionNumExecutorRegistrations = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: bbNamespace,
+		Subsystem: "remote_execution",
+		Name:      "num_executor_registrations",
```
bduffany, line 668: nit: `executor_registration_count` to be consistent with other metric naming

bduffany (approved):

tylerwilliams (approved):
Thanks for adding this!

PR #3502, opened 2023-03-07
Add proto fields for filtering by build pattern (exact match).

jdhollen:


proto/invocation.proto
```diff
@@ -241,6 +241,9 @@ message InvocationQuery {
 
   // Stat filters (duration_usec, cas_cache_misses, etc.)
   repeated stat_filter.StatFilter filter = 14;
+
+  // The pattern for the targets built. Ex: "//..."
```
tylerwilliams, line 245: Since patterns are a little ambiguous do you want to comment that this is an exact match?

tylerwilliams (approved):

proto/invocation.proto
jdhollen: Sure yeah, done.

PR #3517, opened 2023-03-09
Fix edge cases with org join flow

bduffany:
:point_right: **Recommended:** [review with whitespace diff hidden](https://github.com/buildbuddy-io/buildbuddy/pull/3517/files?diff=split&w=1). (The UserDB diff is much cleaner this way) 

* Instead of `buildbuddy_server` performing multiple separate transactions to look up the group and decide whether to auto-join by domain association or submit a request to join the org, consolidate all of this logic into a single DB transaction in `RequestToJoinOrg`, and remove the `AddUserToGroup` func in UserDB.
* Fix edge case where `addUserToGroup` would not auto-promote to admin if the group was empty and there were existing membership requests which had not been approved.
* Fix edge case where we return PermissionDenied if users try to join an org after the org has claimed domain ownership and also after the user was created (this was because of the admin auth check in AddUserToGroup - this PR removes that func entirely).

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -727,7 +727,10 @@ func (d *UserDB) RequestToJoinGroup(ctx context.Context, groupID string) (grpb.G
 			return err
 		}
 		if existing != nil {
-			return status.AlreadyExistsError("You've already requested to join this organization.")
+			if existing.MembershipStatus == int32(grpb.GroupMembershipStatus_REQUESTED) {
+				return status.AlreadyExistsError("You've already requested to join this organization.")
+			}
+			return status.AlreadyExistsError("You're already in this organization.")
 		}
 		return tx.Create(&tables.UserGroup{
 			UserUserID:       userID,
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -649,6 +649,9 @@ func (d *UserDB) AddUserToGroup(ctx context.Context, userID string, groupID stri
 }
 
 func (d *UserDB) addUserToGroup(tx *db.DB, userID, groupID string) error {
+	// Count the number of users in the group.
+	// If there are no existing users, then user should join with admin role,
+	// otherwise they should join with default role.
 	row := &struct{ Count int64 }{}
 	err := tx.Raw(`
 		SELECT COUNT(*) AS count FROM UserGroups
@@ -658,7 +661,6 @@ func (d *UserDB) addUserToGroup(tx *db.DB, userID, groupID string) error {
 		return err
 	}
 	r := role.Default
-	// If no existing users in the group, promote to admin automatically.
 	if row.Count == 0 {
 		r = role.Admin
 	}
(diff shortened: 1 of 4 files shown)


enterprise/server/backends/userdb/userdb.go
```diff
@@ -639,15 +643,26 @@ func (d *UserDB) DeleteGroupGitHubToken(ctx context.Context, groupID string) err
...
+		return d.addUserToGroup(tx, userID, groupID)
 	})
 }
 
-func (d *UserDB) addUserToGroup(ctx context.Context, tx *db.DB, userID, groupID string) error {
+func (d *UserDB) addUserToGroup(tx *db.DB, userID, groupID string) error {
+	row := &struct{ Count int64 }{}
+	err := tx.Raw(`
+		SELECT COUNT(*) AS count FROM UserGroups
+		WHERE group_group_id = ? AND membership_status = ?
```
tylerwilliams, line 655: should this query also look at userID?

enterprise/server/backends/userdb/userdb.go
```diff
@@ -656,60 +671,73 @@ func (d *UserDB) addUserToGroup(ctx context.Context, tx *db.DB, userID, groupID
...
+		}
+
+		// Check if there's an existing request and return AlreadyExists if so.
+		existing, err := getUserGroup(tx, userID, groupID)
+		if err != nil {
+			return err
+		}
+		if existing != nil {
 			return status.AlreadyExistsError("You've already requested to join this organization.")
 		}
```
tylerwilliams, line 729: do we want to distinguish between already requested and already a member?

enterprise/server/backends/userdb/userdb.go
```diff
@@ -656,60 +671,73 @@ func (d *UserDB) addUserToGroup(ctx context.Context, tx *db.DB, userID, groupID
...
-		return status.InvalidArgumentError("User ID is required.")
+		return 0, status.InvalidArgumentError("User ID is required.")
 	}
 	if groupID == "" {
-		return status.InvalidArgumentError("Group ID is required.")
+		return 0, status.InvalidArgumentError("Group ID is required.")
 	}
-	return d.h.Transaction(ctx, func(tx *db.DB) error {
-		existing, err := getUserGroup(tx, userID, groupID)
+	var newStatus grpb.GroupMembershipStatus
```
tylerwilliams, line 704: might be clearer to call this membershipStatus

enterprise/server/backends/userdb/userdb.go
```diff
@@ -1162,3 +1190,11 @@ func hasAdminOnlyCapabilities(capabilities []akpb.ApiKey_Capability) bool {
 	}
 	return false
 }
+
+func getEmailDomain(email string) string {
+	chunks := strings.Split(email, "@")
+	if len(chunks) == 0 {
+		return ""
+	}
+	return chunks[len(chunks)-1]
```
tylerwilliams, line 1199: No strong preference here, but another way to do this if you expect two parts is to use `strings.SplitN`

```
parts := strings.SplitN(email, "@", 2)
if len(parts) != 2 {
  return ""
}
return parts[1]
```

tylerwilliams (changes requested):

enterprise/server/backends/userdb/userdb.go
bduffany: Woops, missed this comment. We do distinguish these 2 cases - the "already a member" case is just moved into the `addUserToGroup` func

bduffany: Actually nvm - that addUserToGroup is only called in the owned-domain case. Will fix and add a test

bduffany: Done - PTAL

tylerwilliams (approved):

PR #3519, opened 2023-03-09
Allow filtering by exact-match of the build target pattern.

jdhollen:
(behind a flag)


bduffany (approved):
nice! was just wishing I had this the other day, to filter by workflow name.

PR #3523, opened 2023-03-09
cli/picker: narrow query scope

sluongng:
Only query for local executable targets vs global ones.
Also ignore any errors/failures when running such query.

---

**Version bump**: Patch


siggisim (approved):
Thanks!

PR #3525, opened 2023-03-09
remove extra day in heatmap when timestamp is exactly midnight.

jdhollen:


tylerwilliams (approved):

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -501,7 +501,7 @@ func getTimestampBuckets(q *stpb.TrendQuery, timezoneOffset time.Duration) ([]in
 
 	endSec := time.Now().Unix()
 	if highTime != nil {
-		endSec = highTime.GetSeconds()
+		endSec = highTime.GetSeconds() - 1
```
tylerwilliams, line 504: could add a lil comment as a treat

iain-macdonald (approved):

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: always

PR #3528, opened 2023-03-09
Support self-hosted workflows

siggisim:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -641,7 +641,7 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 	os := strings.ToLower(workflowAction.OS)
 	computeUnits := *workflowsMacComputeUnits
 	// Use the CI runner image if the OS supports containerized actions.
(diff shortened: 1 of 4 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -641,7 +641,7 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 	os := strings.ToLower(workflowAction.OS)
 	computeUnits := *workflowsMacComputeUnits
 	// Use the CI runner image if the OS supports containerized actions.
-	if os == "" || os == platform.LinuxOperatingSystemName {
+	if !workflowAction.SelfHosted && (os == "" || os == platform.LinuxOperatingSystemName) {
```
bduffany, line 644: I think maybe this check should move to the `if *enableFirecracker {` check below? (we still want to apply the configured linux compute units if the OS is linux)

siggisim: Good call, done.

bduffany (approved):
woo, nice!

PR #3529, opened 2023-03-09
Fix recycling failures on second recycling attempt

bduffany:
Fixes "Snapshot configuration already fetched with different digest"

---

**Version bump**: Patch


tylerwilliams (approved):

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -161,8 +161,8 @@ func (l *FileCacheLoader) GetConfigurationData(snapshotDigest *repb.Digest) ([]b
 // UnpackSnapshot unpacks all of the files in a snapshot to the specified output
 // directory.
 func (l *FileCacheLoader) UnpackSnapshot(snapshotDigest *repb.Digest, outputDirectory string) error {
-	if l.snapshotDigest != snapshotDigest {
-		return status.InvalidArgumentError("Snapshot configration already fetched with different digest")
+	if l.snapshotDigest != nil && l.snapshotDigest != snapshotDigest {
```
bduffany, line 164: The only substantial change in this PR is the `!= nil` check here.

PR #3533, opened 2023-03-10
Remove digest.NewGenericResourceName; use digest.NewResourceName everywhere.

tylerwilliams:
- Remove digest.NewGenericResourceName; use digest.NewResourceName everywhere
- Add explicit error checking to ResourceName.{UploadString,DownloadString} methods: it's an error to try to upload or download AC objects using the bytestream API, and an indication that the wrong type of resource is being used.


siggisim (approved):
Sorry I thought I hit approve on this yesterday

PR #3534, opened 2023-03-10
Add an affordance to zoom in on the selected section of the heatmap

jdhollen:
This needs a better UI (should probably be floating right next to the selection) but this is at least a good starting point for  having this very-necessary feature available.

### DIFF
--- a/enterprise/app/trends/drilldown_page.tsx
+++ b/enterprise/app/trends/drilldown_page.tsx
@@ -573,12 +573,14 @@ export default class DrilldownPageComponent extends React.Component<Props, State
                                     )}
                                   />
                                   <Bar
+                                    cursor="pointer"
                                     dataKey={(entry: stats.DrilldownEntry) =>
                                       +entry.baseValue / +(this.state.drilldownData?.totalInBase || 1)
                                     }
                                     fill="#8884d8"
                                   />
                                   <Bar
+                                    cursor="pointer"
                                     dataKey={(entry: stats.DrilldownEntry) =>
                                       +entry.selectionValue / +(this.state.drilldownData?.totalInSelection || 1)
                                     }
--- a/enterprise/app/trends/drilldown_page.tsx
+++ b/enterprise/app/trends/drilldown_page.tsx
@@ -12,7 +12,7 @@ import FilledButton, { OutlinedButton } from "../../../app/components/button/but
 import { invocation } from "../../../proto/invocation_ts_proto";
 import { stat_filter } from "../../../proto/stat_filter_ts_proto";
 import { stats } from "../../../proto/stats_ts_proto";
-import { google as google_timestamp } from "../../proto/timestamp_ts_proto";
+import { google as google_timestamp } from "../../../proto/timestamp_ts_proto";
 import { usecToTimestamp } from "../../../app/util/proto";
 import { getProtoFilterParams, isExecutionMetric } from "../filter/filter_util";
 import { HeatmapComponent, HeatmapSelection } from "./heatmap";
@@ -244,7 +244,7 @@ export default class DrilldownPageComponent extends React.Component<Props, State
       pageToken: "",
       count: 25,
     });
-    this.addZoomFiltersToQuery(request.query);
+    this.addZoomFiltersToQuery(request.query!);
 
     rpcService.service
       .searchInvocation(request)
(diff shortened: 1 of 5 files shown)


enterprise/app/trends/drilldown_page.tsx
```diff
@@ -307,6 +321,45 @@ export default class DrilldownPageComponent extends React.Component<Props, State
...
+
+  addZoomFiltersToQuery(
+    // A little bit of structural typing for TrendQuery and InvocationQuery
+    query: {
+      updatedBefore?: google_timestamp.protobuf.Timestamp | null;
+      updatedAfter?: google_timestamp.protobuf.Timestamp | null;
+      filter?: stat_filter.StatFilter[];
+    }
+  ) {
+    console.log(this.currentZoomFilters);
```
bduffany, line 344:
```suggestion
```

enterprise/app/trends/drilldown_page.tsx
```diff
@@ -431,6 +484,58 @@ export default class DrilldownPageComponent extends React.Component<Props, State
...
+              Date: {startDate} - {endDate}
+            </div>
+            <div className="drilldown-page-zoom-filter-attr">
+              Value: {startValue} - {endValue}
+            </div>
+          </div>
+        </OutlinedButton>
+      </div>
+    ) : (
+      <div className="drilldown-page-zoom-summary buttonless">
```
bduffany, line 517: nit: the "buttonless" is mildly confusing because the div contains a FilledButton :sweat_smile:

enterprise/app/trends/heatmap.tsx
```diff
@@ -226,6 +227,13 @@ class HeatmapComponentInternal extends React.Component<HeatmapProps, State> {
     }
   }
 
+  maybeFireZoomCallback() {
+    const s = this.computeHeatmapSelection();
```
bduffany, line 231: nit: the single-letter naming convention is more of a go-ism

enterprise/app/trends/trends.css
```diff
@@ -77,6 +77,45 @@
...
 
+.drilldown-page-zoom-summary {
+  display: inline-block;
+  height: 40px;
+  margin: 0 0 0 36px;
+  vertical-align: bottom;
+}
+
+.drilldown-page-zoom-summary.buttonless {
+  margin: 0 0 0 61px;
```
bduffany, line 88: nit: the 61px looks slightly weird because it's not snapped to a multiple of 4px, maybe add a clarifying comment (or just snap to 4px if it looks ok)

enterprise/app/trends/drilldown_page.tsx
```diff
@@ -307,6 +321,45 @@ export default class DrilldownPageComponent extends React.Component<Props, State
...
+
+  handleClearZoom() {
+    this.currentHeatmapSelection = undefined;
+    this.currentZoomFilters = undefined;
+    this.fetch();
+  }
+
+  addZoomFiltersToQuery(
+    // A little bit of structural typing for TrendQuery and InvocationQuery
+    query: {
```
bduffany, line 338: nit: you could also extract an explicit type def to make the func param definition slightly less syntax-soupy

enterprise/app/trends/drilldown_page.tsx
```diff
@@ -270,7 +280,9 @@ export default class DrilldownPageComponent extends React.Component<Props, State
       maximumDuration: filterParams.maximumDuration,
       status: filterParams.status,
     });
+    this.addZoomFiltersToQuery(heatmapRequest.query);
 
+    console.log(heatmapRequest);
```
bduffany, line 285:
```suggestion
```

bduffany (approved):
LGTM, just some nits

enterprise/app/trends/drilldown_page.tsx
jdhollen: Bah, thanks.

enterprise/app/trends/drilldown_page.tsx
jdhollen: ahahaha sorry, i restructured this like five times.  renamed it to 'zoomed', which is the actual state instead of a comment about the style.

enterprise/app/trends/heatmap.tsx
jdhollen: fair enough, replaced this and the one above it, too.

enterprise/app/trends/trends.css
jdhollen: Added a comment--this is accounting for a border, so it's either this or a transparent border, so it'd need a comment no matter what :P

enterprise/app/trends/drilldown_page.tsx
jdhollen: Done

PR #3536, opened 2023-03-10
Switch on executor context-based shutdown by default

bduffany:
This has been enabled in prod for a while now.

---

**Version bump**: Patch


iain-macdonald (approved):
Wanna remove the places where this is set to true in our prod configs too?
https://github.com/search?q=repo%3Abuildbuddy-io%2Fbuildbuddy-internal%20context_based_shutdown_enabled&type=code

bduffany:
> Wanna remove the places where this is set to true in our prod configs too? https://github.com/search?q=repo%3Abuildbuddy-io%2Fbuildbuddy-internal%20context_based_shutdown_enabled&type=code

yep, will send another PR for that

PR #3539, opened 2023-03-10
decrease remote execution test shards to decrease data transfer in ou…

jdhollen:


enterprise/server/test/integration/remote_execution/BUILD
```diff
@@ -8,7 +8,7 @@ go_test(
     data = [
         "//enterprise/server/test/integration/remote_execution/command:testcommand",
     ],
-    shard_count = 32,
+    shard_count = 4,
```
bduffany, line 11: just so people don't unintentionally undo this :)

```suggestion
    # Shard count is intentionally somewhat small here to avoid
    # excessive cache transfer for each test shard action, since this test
    # pulls in a lot of dependencies.
    shard_count = 4,
```

bduffany (approved):

PR #3540, opened 2023-03-10
In release.py, don't create a new release tag if there already exists a draft release

maggie-lou:
This will make the script more idempotent if the auto-release workflow reruns it, and will help minimize user error if you forget to add the skip_version_bump flag.

### DIFF
--- a/release.py
+++ b/release.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import argparse
+from github import Github
 import re
 import subprocess
 import sys
@@ -44,6 +45,11 @@ def workspace_is_clean():
                          shell=True, stdout=subprocess.PIPE,
                          stderr=subprocess.STDOUT)
     return len(p.stdout.readlines()) == 0
+
+def is_draft_release(version_tag):
+    g = Github()
+    release = g.get_repo("buildbuddy-io/buildbuddy").get_release(version_tag)
+    return release.draft
 
 def bump_patch_version(version):
     parts = version.split(".")
@@ -145,9 +151,6 @@ def main():
     parser.add_argument('--skip_latest_tag', default=False, action='store_true')
     args = parser.parse_args()
 
-    bump_version = not args.skip_version_bump
-    update_latest_tag = not args.skip_latest_tag
-
     if workspace_is_clean():
         print("Workspace is clean!")
     elif args.allow_dirty:
@@ -157,6 +160,7 @@ def main():
             'Please run this in a clean workspace!')
 
     old_version = get_latest_remote_version()
+    bump_version = not(args.skip_version_bump or is_draft_release(old_version))
     new_version = old_version
     if bump_version:
         new_version = bump_patch_version(old_version)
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #3540** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3540" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3540?utm_source=stack-comment).

release.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import argparse
+from github import Github
```
bduffany, line 3: This `github` package likely isn't installed on everyone's dev machines - it'd be slightly less of a hassle to curl the releases API using the built-in requests lib

release.py
```diff
@@ -149,6 +155,7 @@ def main():
     parser.add_argument('--allow_dirty', default=False, action='store_true')
     parser.add_argument('--skip_version_bump', default=False, action='store_true')
     parser.add_argument('--skip_latest_tag', default=False, action='store_true')
+    parser.add_argument('--github_token')
```
bduffany, line 158: nit: use an env var for this instead, e.g. `os.environ.get('GITHUB_TOKEN')`

release.py
```diff
@@ -46,10 +46,16 @@ def workspace_is_clean():
                          stderr=subprocess.STDOUT)
     return len(p.stdout.readlines()) == 0
 
-def is_draft_release(version_tag):
-    g = Github()
-    release = g.get_repo("buildbuddy-io/buildbuddy").get_release(version_tag)
-    return release.draft
+def is_draft_release(github_token, version_tag):
+    query_url = f"https://api.github.com/repos/buildbuddy-io/buildbuddy/releases/tags/{version_tag}"
```
bduffany, line 50: hm I'm not sure this `/tags/{tag}` endpoint works until the release is published - over in over in `buildbuddy-io/buildbuddy-ci-playground` I created a draft release v0.0.0 and then a published release v0.0.1, and if I curl this URL for v0.0.0 I get a 404 response even when authenticated, but the v0.0.1 release returns just fine.

Instead I think we could curl the latest page of `/releases` then scan for a release with the tag_name field == version_tag, and draft == True

release.py
```diff
@@ -46,10 +46,16 @@ def workspace_is_clean():
...
+    headers = {'Authorization': f'token {github_token}'}
+    r = requests.get(query_url, headers=headers)
+    if r.status_code == 401:
+        die("Invalid github credentials. Did you pass a valid github token?")
+    elif r.status_code == 404:
+        die(f"No release found for version tag {version_tag}.")
```
bduffany, line 56: I think we can return false in this case? (no release found means no draft release was found either)

release.py
maggie-lou: Done

release.py
maggie-lou: Ah good catch - updated

release.py
maggie-lou: Refactored so that if no release is found, we also skip the version bump. I thought this could be helpful if we create a tag with release.py, but the workflow that creates the release fails. We may want to rerun release.py, but wouldn't expect it to bump the version again

iain-macdonald (approved):

PR #3541, opened 2023-03-11
In release.py, don't push to gcr if image with that tag already exists

maggie-lou:
---

**Version bump**: TODO


maggie-lou:
* master
  * **PR #3540** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3540" target="_blank">Graphite</a> 
    * **PR #3541** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3541" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3541?utm_source=stack-comment).

release.py
```diff
@@ -117,20 +123,34 @@ def update_docker_images(version_tag, update_latest_tag):
...
-    push_image('//enterprise/deployment:release_executor_enterprise', image_tag='enterprise-'+version_tag)
+    if not gcr_tag_exists("flame-public/buildbuddy-executor-enterprise", enterprise_tag):
+        push_image('//enterprise/deployment:release_executor_enterprise', image_tag=enterprise_tag)
+        pushed_enterprise_exec_image = True
 
     # update "latest" tags
     if update_latest_tag:
         # OSS app
-        push_image('//deployment:release_onprem', image_tag='latest')
+        if pushed_oss_image:
```
bduffany, line 146: if this script fails on e.g.  line 136, then when we retry this script, `pushed_oss_image` will be false since the `gcr_tag_exists` check will return false, and we won't update the latest tag on line 147, but should.

one thing that could work is to use the docker API for each of these 3 images to resolve the image digests for both the vX.Y.Z tag and the 'latest' tag, and if the digests are mismatched, then pull the vX.Y.Z tag and re-tag+push as 'latest' (pull+push would be much faster than re-building, and guarantees that the rebuild doesn't result in a new digest if the bazel build isn't 100% reproducible from the preceding failed run, which should be rare but can happen)

bduffany: Looks like we can also do it without pushing/pulling: https://stackoverflow.com/a/72908550

release.py
```diff
@@ -88,6 +88,11 @@ def confirm_new_version(version):
         version = get_version_override()
     return version
 
+def gcr_tag_exists(project, tag):
+    query_url = f"https://gcr.io/v2/{project}/tags/list"
```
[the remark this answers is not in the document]
maggie-lou: Nice - fixed.

release.py
maggie-lou: Fixed

release.py
```diff
@@ -106,35 +113,48 @@ def create_and_push_tag(old_version, new_version, release_notes=''):
...
+        if latest_image is None:
+            die(f"Could not fetch image with latest tag from project {project}.")
+
+        should_update_latest_tag = version_image["config"]["digest"] != latest_image["config"]["digest"]
+        if should_update_latest_tag:
+            add_tag_cmd = f"y | gcloud container images add-tag gcr.io/{project}:{version_tag} gcr.io/{project}:latest"
+            run_or_die(add_tag_cmd)
+
+def push_image_with_bazel(bazel_target, image_tag):
+    print(f"Pushed docker image target {bazel_target} tag {image_tag}")
```
iain-macdonald, line 136: s/Pushed/Pushing  ?

release.py
```diff
@@ -106,35 +113,48 @@ def create_and_push_tag(old_version, new_version, release_notes=''):
     push_tag_cmd = 'git push origin %s' % new_version
     run_or_die(push_tag_cmd)
 
-def push_image(target, image_tag):
+def push_image_for_project(project, version_tag, bazel_target, skip_update_latest_tag):
+    version_image = get_image(project, version_tag)
+    if version_image is None:
+        push_image_with_bazel(bazel_target, version_tag)
+
+    if not skip_update_latest_tag:
```
iain-macdonald, line 121: nit: I would use a guard clause here:

```
if skip_update_latest_tag:
    return
...
```

iain-macdonald (approved):

bduffany (approved):

PR #3544, opened 2023-03-13
Remove digest.Parse function and usages

tylerwilliams:


siggisim (approved):

enterprise/server/test/integration/ci_runner/ci_runner_test.go
```diff
@@ -690,7 +690,7 @@ func TestHostedBazel_ApplyingAndDiscardingPatches(t *testing.T) {
 			"--target_repo_url=file://" + targetRepoPath,
 			"--target_branch=master",
 			"--cache_backend=" + app.GRPCAddress(),
-			"--patch_digest=" + fmt.Sprintf("%s/%d", patchDigest.GetHash(), patchDigest.GetSizeBytes()),
+			"--patch_digest=" + fmt.Sprintf("blobs/%s/%d", patchDigest.GetHash(), patchDigest.GetSizeBytes()),
```
bduffany, line 693: if we add the instance name here would that let us remove the backwards-compatibility code?

tools/cas/cas.go
```diff
@@ -66,16 +66,25 @@ func main() {
 	if *blobDigest == "" {
 		log.Fatalf("Missing --digest")
 	}
-	d, err := digest.Parse(*blobDigest)
+	var ind *digest.ResourceName
+	var err error
+
+	if *blobType == "ActionResult" {
+		ind, err = digest.ParseActionCacheResourceName(*blobDigest)
```
bduffany, line 73: IIUC this means the `-digest` flag requires a `blobs/ac` prefix now, or `blobs/` if you're fetching from CAS, which feels a bit more cumbersome

Would be nicer imo to build the resource name string based on the existing digest / instance name flags (using /ac/ only if blobtype is ActionResult), so that we could remove the backwards compatibility code

enterprise/server/test/integration/ci_runner/ci_runner_test.go
tylerwilliams: The test does not set a remote_instance_name, so no changes are required here. I did change hosted_runner.go though which sets these -- now it always passes a full correct download string URI, and I removed the fallback code.

bduffany, line 693: nit: we can rename this to `--patch_uri` since we probably don't need to worry about compatibility during the rollout

bduffany (approved):
LGTM, just one nit

tools/cas/cas.go
tylerwilliams: Fixed this.

PR #3546, opened 2023-03-13
Disable ci_runner_test in Mac workflows for now

bduffany:
It times out a lot - will try to debug this when I have more time.

Also strip newlines in the YAML so that the bazel command renders nicely in the UI.

---

**Version bump**: Patch


sluongng (approved):
Lgtm. Though we should consider fixing the UI rendering in the future

PR #3548, opened 2023-03-13
Add a proto for runner recycling key

bduffany:
Add a proto to replace the [`query` struct in `runner.go`](https://github.com/buildbuddy-io/buildbuddy/blob/bc27c5bc208c2fd6998632a3bbae390d4353c3e9/enterprise/server/remote_execution/runner/runner.go#L1146).

This will let us serialize and hash the runner pool key for easier debugging of recycling issues (e.g. "not reusing runner with key X for task with key Y"), and also greatly simplify the pool match implementation and make it less error-prone since we can replace all the manual equality checking with a proto equality comparison.

At some point, it might be interesting to also expose the runner pool info via a monitoring endpoint or maybe even the UI (executors page), as well as store computed runner match properties in the ExecutedActionMetadata. Having this info in proto format will be convenient for all of those purposes as well.

---

**Version bump**: Patch


tylerwilliams:
I know we have some small mods to remote_execution.proto, but I think we should try to minimize them so it's easier to update that proto in the future. Could we move these protos to their own file or find an existing one they would fit with?

bduffany:
> I know we have some small mods to remote_execution.proto, but I think we should try to minimize them so it's easier to update that proto in the future. Could we move these protos to their own file or find an existing one they would fit with?

Ah, we have `runner.proto` which is referencing our hosted bazel stuff but I think it's closely related enough that it could fit there. wdyt?

bduffany:
Closing in favor of https://github.com/buildbuddy-io/buildbuddy/pull/3560/files -- since there's only 3 fields now, I don't feel as strongly about having a proto.

PR #3558, opened 2023-03-15
cli/fix: use tree sitter query

sluongng:
Experimental switching to use Tree Sitter Query to collect Typescript metadata.

This comes with a performance tradeoff:

- Before this PR
  ```
  > hyperfine 'bb fix'
  Benchmark 1: bb fix
    Time (mean ± σ):      1.712 s ±  0.007 s    [User: 1.732 s, System: 0.091 s]
    Range (min … max):    1.703 s …  1.724 s    10 runs
  ```

- After this PR
  ```
  > hyperfine 'bb fix'
  Benchmark 1: bb fix
    Time (mean ± σ):      2.432 s ±  0.015 s    [User: 2.435 s, System: 0.100 s]
    Range (min … max):    2.419 s …  2.471 s    10 runs
  ```

So it's `1.732 -> 2.435 (+40.5%)` slower.

But we gain the ability to precisely finding structure within the Typescript AST.

This PR is split into 3 commits for easier reviewing:
```
> PAGER= git log -3 --oneline --reverse
# Apply 'bb fix' to latest master
5991b4f14 apply bb fix

# Apply new code changes to switch to TreeSitter Query 
77b711b02 cli/fix: switch to use tree-sitter

# Apply new 'bb fix' with new metadata extracted using TreeSitter Query
adde3f78b apply bb fix
```

---

**version bump**: patch


sluongng:
I cook this up during my downtime prep-ing for flights / during flights / post-flight-but-no-5G.

Even though it's slower than regex approach, now we have the ability to generate more intricate rules (i.e. the future rules_js / rules_ts from Aspect folks).

This could also be applied to other languages in the tree-sitter ecosystem, including https://github.com/amaanq/tree-sitter-starlark

siggisim:
My initial version used tree-sitter queries like this, but I found that it was much slower than I liked. I want `bb fix` to be fast so it can be run regularly (i.e. on every save). 

Our repo only has 157 .ts/.tsx files, so I'm a bit worried on how this will scale to larger repos. 

I'd almost rather go back to the regex approach than slow things down considerably. There might be some other solution like only running this on changed files that would make this good enough though.

sluongng:
Yeah I think changed files might be an interesting direction to take the perf.
Detecting file changes is tricky though, https://lobste.rs/s/mjppkn/navigating_complexities_file_watching is a good thread on the topic. I know that we are using github.com/bduffany/godemon which in turn uses github.com/fsnotify/fsnotify. I will do a bit more research and come back to this in the future.

PR #3564, opened 2023-03-15
Support streaming images to executors using the soci-snapshotter

iain-macdonald:
This PR:
- Adds a dependency on a forked version of the soci-snapshotter repository which adds support for podman: https://github.com/iain-macdonald/soci-snapshotter
- Bundles the soci-store binary built from that repository into the executor
- Removes all of the stargz stuff from podman.go and replaces it with soci equivalents (where applicable) so that images can be streamed using the soci snapshotter, when --executor.podman.image_streaming.enabled is set to true

Note that for now images can only be streamed if there's a soci index stored alongside the image in the container registry. If one isn't found, it falls back to regular old podman behavior. I'm planning to look into adding support for all images after getting this working.

I had this working in dev the other day, but I'll verify it again before submitting.

**Version bump**: Minor


vadimberezniker (approved):

enterprise/server/cmd/executor/BUILD
```diff
@@ -96,7 +96,11 @@ container_layer(
         ":firecracker",
         ":jailer",
         "@com_github_containerd_stargz_snapshotter-v0.11.4-linux-amd64//:stargz-store",
```
bduffany, line 98: nit: did you want to delete this one?

bduffany (approved):

enterprise/server/cmd/executor/BUILD
iain-macdonald: Whoops, yup, done

PR #3566, opened 2023-03-15
Fetch firecracker and stargz-store binaries with bazel

bduffany:
This shaves about 38MB off of our repo size which should speed up `git clone` times when run with `--filter=blob:none`.

In a follow-up PR I plan on doing this for the redis binaries that we have checked into the repo (11MB). Ideally we can also do this for `enterprise/vmsupport/vmlinux` (17MB)

---

**Version bump**: Patch


iain-macdonald (approved):
Nice!

vadimberezniker (approved):

tylerwilliams (approved):
🪶

bduffany:
This broke dev - looks like the executor can't find jailer in PATH. It looks like the symlink was not properly created - digging into it

PR #3578, opened 2023-03-16
Send timezone with request context, use it in the heat map.

jdhollen:


enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -546,8 +555,8 @@ type HeatmapQueryInputs = struct {
 	MetricArrayStr         string
 }
 
-func (i *InvocationStatService) generateQueryInputs(ctx context.Context, table string, metric string, q *stpb.TrendQuery, whereClauseStr string, whereClauseArgs []interface{}, timezoneOffsetMinutes int32) (*HeatmapQueryInputs, error) {
-	timestampBuckets, timestampArrayStr, err := getTimestampBuckets(q, time.Duration(timezoneOffsetMinutes)*time.Minute)
+func (i *InvocationStatService) generateQueryInputs(ctx context.Context, table string, metric string, q *stpb.TrendQuery, whereClauseStr string, whereClauseArgs []interface{}, requestContext *ctxpb.RequestContext) (*HeatmapQueryInputs, error) {
```
tylerwilliams, line 558: not really related to this CL but this is a gnarly mf

tylerwilliams (commented):
tiny nit on go side but otherwise LGTM

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -493,39 +495,46 @@ func getTimestampBuckets(q *stpb.TrendQuery, timezoneOffset time.Duration) ([]in
...
-		endSec = highTime.GetSeconds() - 1
+		endSec = highTime.GetSeconds()
+	}
+
+	var loc *time.Location
+	timezoneOffset := time.Duration(requestContext.GetTimezoneOffsetMinutes()) * time.Minute
+	timezone := requestContext.GetTimezone()
+	// Find the user's timezone. time.LoadLocation defaults the empty string to
+	// UTC, but we want to fall back to an offset, so we need to do that first.
+	if !(*useTimezoneInHeatmapQueries) || timezone == "" {
```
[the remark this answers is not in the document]
jdhollen: you are right of course, i hated the way it looked so i tried shoving some parens in there and was surprised that gofmt didnt take them out, but i will now, to be more Goth™

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -493,39 +495,46 @@ func getTimestampBuckets(q *stpb.TrendQuery, timezoneOffset time.Duration) ([]in
...
+		loc = time.FixedZone("Fixed Offset", -int(timezoneOffset.Seconds()))
+	} else {
+		loadedLoc, err := time.LoadLocation(timezone)
+		if err != nil {
+			loc = time.FixedZone("Fixed Offset", -int(timezoneOffset.Seconds()))
+		} else {
+			loc = loadedLoc
+		}
 	}
```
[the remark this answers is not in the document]
jdhollen: Nah yeah this is way better, I hated this block

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Sigh, yeah.  It's mostly structured this way for reuse, but I will take another pass at it someday.  It's not good when you already are thinking "who wrote this" a month after you write something.

app/service/rpc_service.ts
```diff
@@ -24,14 +24,20 @@ export type BuildBuddyServiceRpcName = RpcMethodNames<buildbuddy.service.BuildBu
 class RpcService {
   service: ExtendedBuildBuddyService;
   events: Subject<string>;
-  requestContext = new context.RequestContext({
-    timezoneOffsetMinutes: new Date().getTimezoneOffset(),
```
bduffany, line 28: nit: we don't need to support IE11/Opera Mini etc. if you want to call `Intl.DateTimeFormat().resolvedOptions().timeZone` here without any conditionals etc https://caniuse.com/?search=resolvedOptions

bduffany (approved):

app/service/rpc_service.ts
jdhollen: Fiiine

PR #3581, opened 2023-03-16
Add RPC for searching executions in clickhouse.

jdhollen:
---

**Version bump**: TODO


enterprise/server/execution_search_service/execution_search_service.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, status.InvalidArgumentError("Failed to find user's group when searching executions.")
+	}
+	if err := perms.AuthorizeGroupAccessForStats(ctx, s.env, u.GetGroupID()); err != nil {
+		return nil, err
+	}
+
+	q := query_builder.NewQuery(`SELECT * FROM Executions`)
+
+	// Always filter to the currently selected (and authorized) group.
+	q.AddWhereClause("group_id = ?", u.GetGroupID())
```
vadimberezniker, line 130: The currently selected group is going to be in the RequestContext object of the request message.

enterprise/server/execution_search_service/execution_search_service.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, err
+	}
+	if u.GetGroupID() == "" {
+		return nil, status.InvalidArgumentError("Failed to find user's group when searching executions.")
+	}
+	if err := perms.AuthorizeGroupAccessForStats(ctx, s.env, u.GetGroupID()); err != nil {
+		return nil, err
+	}
+
+	q := query_builder.NewQuery(`SELECT * FROM Executions`)
```
vadimberezniker, line 127: We have code in perms.go (AddPermissionsCheckToQuery) to automatically craft a permission filter clause. You should add that on top of whatever other filtering you are doing.

vadimberezniker (changes requested):

enterprise/server/execution_search_service/execution_search_service.go
jdhollen: There aren't any perms in the clickhouse table for executions, so what I've done here is added a filter to restrict to the authz'd group and then am also checking perms as part of the lookup in mysql--we don't return any results if we don't find even one of the executions in the permissions-guarded mysql query.  I started using AddPermissionsCheckToQuery there instead of what i'm using now (and i'm still including the group filter in both cases).

enterprise/server/execution_search_service/execution_search_service.go
jdhollen: That doesn't include impersonation, right? This seemed like the right way to get a handle of the "acting" group id and matches what we do in setting permissions checks.

vadimberezniker: Yeah you're right.

vadimberezniker (approved):

PR #3585, opened 2023-03-17
Add backend implementation for bot suggestions

siggisim:
Only enabled if an OpenAI api key is provided in your config.yaml

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/suggestion/suggestion.go
+++ b/enterprise/server/suggestion/suggestion.go
@@ -1,0 +1,151 @@
+package suggestion
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"flag"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/eventlog"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+
+	elpb "github.com/buildbuddy-io/buildbuddy/proto/eventlog"
+	supb "github.com/buildbuddy-io/buildbuddy/proto/suggestion"
+)
+
+var apiKey = flag.String("openai.api_key", "", "OpenAI API key")
+var model = flag.String("openai.model", "gpt-3.5-turbo", "OpenAI model name to use, if not set 'gpt-3.5-turbo' is used.")
+
+const (
+	minLines = 1000
+	maxChars = 8000
+)
+
+type suggestionService struct {
+	env environment.Env
+}
+
+func Register(env environment.Env) error {
+	if *apiKey != "" {
+		env.SetSuggestionService(New(env))
+	}
+	return nil
+}
+
+func New(env environment.Env) *suggestionService {
+	return &suggestionService{
+		env: env,
+	}
+}
+
+func (s *suggestionService) GetSuggestion(ctx context.Context, req *supb.GetSuggestionRequest) (*supb.GetSuggestionResponse, error) {
+	if req.GetInvocationId() == "" {
+		return nil, status.InvalidArgumentErrorf("GetSuggestionRequest must contain a valid invocation_id")
+	}
+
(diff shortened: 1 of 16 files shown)


sluongng (approved):
image

This is so so so cool.
I am HYPED!

enterprise/server/suggestion/suggestion.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/buildbuddy-io/buildbuddy/server/eventlog"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+
+	elpb "github.com/buildbuddy-io/buildbuddy/proto/eventlog"
+	supb "github.com/buildbuddy-io/buildbuddy/proto/suggestion"
+)
+
+var apiKey = flag.String("openai.api_key", "", "OpenAI API key")
+var model = flag.String("openai.model", "gpt-3.5-turbo", "OpenAI model name to use, if not set 'gpt-3.5-turbo' is used.")
```
tylerwilliams, line 22: nit: you don't need the default in the help text, the flags lib will show it in the --help output like this:

```
  -telemetry_endpoint string
    	The telemetry endpoint to use (default "grpcs://t.buildbuddy.io:443")
  -telemetry_interval duration
    	How often telemetry data will be reported (default 24h0m0s)
```

enterprise/server/suggestion/suggestion.go
```diff
@@ -0,0 +1,151 @@
...
+
+const (
+	minLines = 1000
+	maxChars = 8000
```
tylerwilliams, line 26: nit: could use a comment here about what these are and how to set them

enterprise/server/suggestion/suggestion.go
```diff
@@ -0,0 +1,151 @@
...
+	res := &supb.GetSuggestionResponse{
+		Suggestion: []string{completionResponse.Choices[0].Message.Content},
+	}
+
+	return res, nil
+}
+
+// TODO(siggisim): Pull this into its own backend if we want to use this in other places.
+func getCompletions(data *completionRequest) (*completionResponse, error) {
+	url := "https://api.openai.com/v1/chat/completions"
```
tylerwilliams, line 96: nit: can be const

enterprise/server/suggestion/suggestion.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	defer postResp.Body.Close()
+
+	body, err := io.ReadAll(postResp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if postResp.StatusCode != http.StatusOK {
+		log.Printf("%+v %+v", postResp.StatusCode, string(body))
```
tylerwilliams, line 123: Maybe log.Debugf?

enterprise/server/suggestion/suggestion.go
siggisim: Addressed in: https://github.com/buildbuddy-io/buildbuddy/pull/3583/commits/2f05336a0f6f4d071e240592e93e459888a62324

enterprise/server/suggestion/suggestion.go
siggisim: Addressed in: https://github.com/buildbuddy-io/buildbuddy/pull/3583/commits/2f05336a0f6f4d071e240592e93e459888a62324

enterprise/server/suggestion/suggestion.go
siggisim: Addressed in: https://github.com/buildbuddy-io/buildbuddy/pull/3583/commits/2f05336a0f6f4d071e240592e93e459888a62324

enterprise/server/suggestion/suggestion.go
siggisim: That's awesome, addressed in: https://github.com/buildbuddy-io/buildbuddy/pull/3583/commits/2f05336a0f6f4d071e240592e93e459888a62324

PR #3586, opened 2023-03-17
Fetch example executions when the user selects some executions in the heatmap.

jdhollen:


siggisim (approved):

enterprise/app/trends/drilldown_page.tsx
```diff
@@ -219,16 +226,54 @@ export default class DrilldownPageComponent extends React.Component<Props, State
...
+
+    rpcService.service
+      .searchExecution(request)
+      .then((response) => {
+        console.log(response);
+        this.setState({
+          eventData: { executions: response.execution },
+        });
+      })
+      .catch(() => this.setState({ eventsFailed: true, eventData: undefined }))
```
bduffany, line 267: instead of `eventsFailed` you might want to store the error message and display it to the user so that we can diagnose. Probably the easiest thing is to use `errorService.handleError` to show the error banner instead of having custom error handling code

bduffany (approved):

enterprise/app/trends/drilldown_page.tsx
jdhollen: I do like having *some* permanent / visible error state in the specific section that failed, but I don't really want to dump the whole error into it--so let's do both.

PR #3587, opened 2023-03-17
Frontend changes for bot suggestions

siggisim:
---

**Version bump**: Patch

### DIFF
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -640,8 +640,8 @@
   height: 16px;
 }
 
-.invocation-ask-button {
-  background-color: #2196f3 !important;
+.filled-button.invocation-ask-button {
+  background-color: #2196f3;
 }
 
 .invocation-share-dialog .row {
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -632,11 +632,16 @@
   flex-grow: 10;
 }
 
-.invocation-share-button svg {
+.invocation-share-button svg,
+.invocation-ask-button svg {
   margin-right: 12px;
   margin-left: -4px;
   width: 16px;
   height: 16px;
+}
+
+.invocation-ask-button {
+  background-color: #2196f3 !important;
 }
 
 .invocation-share-dialog .row {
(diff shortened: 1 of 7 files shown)


app/invocation/invocation_model.tsx
```diff
@@ -22,6 +25,8 @@ export default class InvocationModel {
   invocations: invocation.Invocation[] = [];
   cacheStats: cache.CacheStats[] = [];
   scoreCard: cache.IScoreCard;
+  botSuggestions: string[] = [];
+  onChange: Subject<string> = new Subject();
```
bduffany, line 29: nit: if you omit `<string>` would that let you do `onChange.next()` instead of passing an empty string?

app/invocation/invocation_buttons.tsx
```diff
@@ -25,6 +32,12 @@ export default class InvocationButtons extends React.Component<InvocationButtons
     return repoUrl.startsWith("https://github.com/");
   }
 
+  private async onAskClicked() {
+    this.setState({ askLoading: true });
+    await this.props.model.fetchSuggestions();
+    this.setState({ askLoading: false });
+  }
```
bduffany, line 39: nit: can use `finally` so the `askLoading: false` line runs unconditionally

```suggestion
  private onAskClicked() {
    this.setState({ askLoading: true });
    return this.props.model.fetchSuggestions().finally(() => this.setState({ askLoading: false }));
  }
```

app/invocation/invocation.css
```diff
@@ -632,13 +632,18 @@
...
+.invocation-share-button svg,
+.invocation-ask-button svg {
   margin-right: 12px;
   margin-left: -4px;
   width: 16px;
   height: 16px;
 }
 
+.invocation-ask-button {
+  background-color: #2196f3 !important;
```
bduffany, line 644: nit: would this work instead of `!important` ?

```suggestion
.filled-button.invocation-ask-button {
  background-color: #2196f3;
```

bduffany (approved):

app/invocation/invocation_model.tsx
siggisim: Had to do <void> but that works!

app/invocation/invocation_buttons.tsx
siggisim: Done

app/invocation/invocation.css
siggisim: Ah yeah, this was me being lazy during hack-week. Fixed!

PR #3590, opened 2023-03-17
Small cleanup to snaploader

bduffany:
* Remove `instanceName` from constructor (instead, we'll want to change `snapshotDigest *repb.Digest` to `snapshotResourceName *rspb.ResourceName` and pass around the instance name that way)
* Add `ctx` to all interface methods, since we'll need this for a remote cache-based `Loader` implementation
* Don't store `ctx` in the loader struct

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -79,16 +79,14 @@ func enumerateFiles(snapOpts *LoadSnapshotOptions) []string {
 }
 
 type Loader interface {
-	CacheSnapshot(snapOpts *LoadSnapshotOptions) (*repb.Digest, error)
-	UnpackSnapshot(snapshotDigest *repb.Digest, outputDirectory string) error
-	DeleteSnapshot(snapshotDigest *repb.Digest) error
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -79,29 +79,25 @@ func enumerateFiles(snapOpts *LoadSnapshotOptions) []string {
...
-	instanceName     string // TODO: remove (this is unused)
 	workingDirectory string
 	snapshotDigest   *repb.Digest
 	manifest         *manifestData
 }
 
 // New returns a new snapshot.Loader that can be used to download the specified
 // snapshot into the target chroot.
-func New(ctx context.Context, env environment.Env, workingDirectory, instanceName string) (Loader, error) {
+func New(ctx context.Context, env environment.Env, workingDirectory string) (Loader, error) {
```
vadimberezniker, line 97: Do we need ctx at all here?

vadimberezniker (approved):

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: Nope, removed

PR #3594, opened 2023-03-18
Add a download command to the CLI

tylerwilliams:
usage: bb download {digest}/{size}

Downloads the blob identified by digest from the CAS and outputs its contents.

If the blob type is specified, the downloaded bytes are interpreted as a proto
of the specified type and text-marshalled before being output. Otherwise the raw
uninterpreted bytes will be output.

Example of downloading a blob and piping it to another command:
  $ bb download 686f25dbadc55ac13aca8580efead40b84d2a812278900b95ce15014dfd94702/80 | sha256sum

Example of displaying an Action with a remote instance name to stdout:
  $ bb download /foo/blobs/b8907c47a5c0954a440c673832f6e219e33311f577a59ea7c9de4686900140ac/234 --type=Action

Example of dumping a Command to a file:
  $ bb download 0e83f9edaff969afa4d16de9f8f5c1de2778148a982e1c19c744fc11a4f96811/1321 --type=Command --output_file=cmd.pb.txt

### DIFF
--- a/cli/cmd/bb/BUILD
+++ b/cli/cmd/bb/BUILD
@@ -13,6 +13,7 @@ go_library(
         "//cli/arg",
         "//cli/bazelisk",
         "//cli/cmd/sidecar",
+        "//cli/download",
         "//cli/fix",
         "//cli/help",
         "//cli/log",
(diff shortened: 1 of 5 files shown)


siggisim (approved):
Nice!!

PR #3596, opened 2023-03-20
Add CLI upload functionality as well

tylerwilliams:
usage: bb upload filename

Uploads the file specified by filename to the CAS and outputs the digest.

If an input file is specified, that file will be uploaded. To upload from
stdin, set the --stdin flag.

Example of uploading a blob from stdin:
  $ echo "buildbuddy" | bb upload --stdin

Example of uploading a file with a remote instance name:
  $ echo -n "buildbuddy" > input_file.txt
  $ bb upload input_file.txt --remote_instance_name=foo

### DIFF
--- a/cli/upload/upload.go
+++ b/cli/upload/upload.go
@@ -43,7 +43,7 @@ Example of uploading a blob from stdin:
 
 Example of uploading a file with a remote instance name:
   $ echo -n "buildbuddy" > input_file.txt
-  $ bb upload input_file.txt
+  $ bb upload input_file.txt --remote_instance_name=foo
 `
 )
 
--- a/cli/upload/upload.go
+++ b/cli/upload/upload.go
@@ -35,11 +35,11 @@ usage: bb ` + flags.Name() + ` filename
 
 Uploads the file specified by filename to the CAS and outputs the digest.
 
-If an input file is specified, that file will be uploaded. Otherwise input
-is read from stdin and uploaded.
+If an input file is specified, that file will be uploaded. To upload from
+stdin, set the --stdin flag.
 
-Example of upload a blob:
-  $ echo "buildbuddy" | bb upload
+Example of uploading a blob from stdin:
+  $ echo "buildbuddy" | bb upload --stdin
 
 Example of uploading a file with a remote instance name:
   $ echo -n "buildbuddy" > input_file.txt
--- a/cli/upload/upload.go
+++ b/cli/upload/upload.go
@@ -1,0 +1,149 @@
+package upload
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/buildbuddy-io/buildbuddy/cli/arg"
+	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	"github.com/buildbuddy-io/buildbuddy/cli/storage"
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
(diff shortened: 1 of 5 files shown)


cli/upload/upload.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+var (
+	flags = flag.NewFlagSet("upload", flag.ContinueOnError)
+
+	target             = flags.String("target", "grpcs://remote.buildbuddy.io", "Cache gRPC target")
+	remoteInstanceName = flags.String("remote_instance_name", "", "Remote instance name")
+	compress           = flags.Bool("compress", true, "If true, enable compression of uploads to remote caches")
+	stdin              = flags.Bool("stdin", false, "If true, read from stdin")
+	digestFunction     = flags.String("digest_function", "SHA256", "If set, use this digest function for uploads.")
```
sluongng, line 31: Can we specify what the supported digest functions here?

cli/upload/upload.go
```diff
@@ -0,0 +1,149 @@
...
+
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	rspb "github.com/buildbuddy-io/buildbuddy/proto/resource"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
+)
+
+var (
+	flags = flag.NewFlagSet("upload", flag.ContinueOnError)
+
+	target             = flags.String("target", "grpcs://remote.buildbuddy.io", "Cache gRPC target")
```
sluongng, line 27: nit: this naming is slightly confusing with `bazel build targets` (such as `//my/target:a`)

Would something such as `remote_instance_url` or `remote_instance_host` work better?

cli/upload/upload.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return err
+	}
+	ind := digest.NewResourceName(d, *remoteInstanceName, rspb.CacheType_CAS)
+	if *compress {
+		ind.SetCompressor(repb.Compressor_ZSTD)
+	}
+	if _, err := f.Seek(0, io.SeekStart); err != nil {
+		return err
+	}
```
sluongng, line 111:
```suggestion
	if _, err := f.Seek(0, io.SeekStart); err != nil {
		return err
	}
	ind := digest.NewResourceName(d, *remoteInstanceName, rspb.CacheType_CAS)
	if *compress {
		ind.SetCompressor(repb.Compressor_ZSTD)
	}
```

nit: couple the seek reset with the file read.

sluongng (commented):
Question: Should we retire `tools/cas` now that these functionality have made it into the `bb` CLI? 🤔

cli/upload/upload.go
tylerwilliams: This matches [download.](https://cs.github.com/buildbuddy-io/buildbuddy/blob/f1f111595aef724bf871f9cf8c8526992153aaf4/cli/download/download.go?q=cli%2Fdownload%2Fdownload.go#L29)

I think most people won't change this so it's probably not a big deal, but if we did want to change it I'd prefer something short and possibly matching bazel, like "remote_cache".

cli/upload/upload.go
tylerwilliams: End users should not set this to anything besides SHA256 for now, so I don't think we should document these yet.

cli/upload/upload.go
tylerwilliams: This is coupled with the second read, which happens in cachetools.UploadFromReader.

Typically a seek would not be necessary because a file is usually only read once. In this case we read it twice. I think its clearer to have the Seek right before the second read, that way a reader does not need to remember / go further back in the code to see the file was read and seeked previously. It's kind of a stand in for a file.Open call if that makes sense.

tylerwilliams:
We can delete tools/cas soon, but probably makes sense to find homes for all that functionality first.

siggisim (approved):

PR #3601, opened 2023-03-20
Fix filecache entries not being overwritable

bduffany:
`Filecache.Add` does not overwrite entries at a given key. This is usually not an issue,  since contents should normally match the keys (which are SHA256 digests of the content), and so any overwrites would be a NOP.

But `snaploader.FileCacheLoader` depends on being able to write entries where the key doesn't necessarily reflect the content digest, similar to AC.

This was broken and was causing problems in some unit tests that I'm adding in one of my branches, though I'm not sure yet whether this caused any problems in practice.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/filecache/BUILD
+++ b/enterprise/server/remote_execution/filecache/BUILD
@@ -32,8 +32,11 @@ go_test(
     deps = [
         ":filecache",
         "//proto:remote_execution_go_proto",
+        "//server/metrics",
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):

bduffany:
Note: This depends on https://github.com/buildbuddy-io/buildbuddy/pull/3605 because we don't want the newly added call to `Remove()` to affect the last_evicted_age metric.

PR #3605, opened 2023-03-21
Allow distinguishing manual removals from evictions in LRU OnEvict callback

bduffany:
We don't want manual removals to count towards `last_eviction_age` metrics.

---

**Version bump**: Patch

### DIFF
--- a/server/util/lru/lru.go
+++ b/server/util/lru/lru.go
@@ -12,13 +12,24 @@ import (
 	xxhash "github.com/cespare/xxhash/v2"
 )
 
+// EvictionReason describes the reason for an entry being evicted from LRU.
+type EvictionReason int
+
+const (
+	// SizeEviction means this was the least recently used entry which had to
+	// be evicted to make room for newer entries.
+	SizeEviction EvictionReason = iota
+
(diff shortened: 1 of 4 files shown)


server/util/lru/lru.go
```diff
@@ -12,8 +12,13 @@ import (
...
 
-// EvictedCallback is used to get a callback when a cache Entry is evicted
-type EvictedCallback func(value interface{})
+// EvictedCallback is a function invoked whenever an LRU entry is evicted or
+// removed.
+//
+// It is passed the value which was removed or evicted, as well as a boolean
+// that is set to true if the value was removed or false if it was evicted. Note
+// that RemoveOldest is counted as an eviction, not a removal.
+type EvictedCallback func(value interface{}, removed bool)
```
tylerwilliams, line 21: this change really adds a lot of complexity and confusion for what it buys.

If you're not sure it's necessary, we could just leave things as is.
If you do think it's necessary, using a new value for removalType might be clearer than a "removed" bool. Something like:

```
type EvictionType int
const (
  SizeEviction EvictionType = iota
  ManualRemoval
)

type EvictedCallback func(value interface{}, type EvictionType)
```

PR #3608, opened 2023-03-21
dev_qa: allow selecting specific repos for testing

bduffany:
* Rewrite in Python (not Go because we still want to be able to run this easily on GH actions)
* Allow selecting specific repos like `--repos=buildbuddy,bazel-gazelle` etc.

---

**Version bump**: Patch


maggie-lou (approved):
Thanks for doing this. Will be a lot easier to read and edit a python script

PR #3614, opened 2023-03-22
protobuf: patch link opts on macos

sluongng:
After recent protobuf upgrade, some team members have experienced the
mysterious error

```
Executing genrule @com_google_protobuf//src/google/protobuf:gen_wkt_cc_sources
...
dyld[97646]: symbol not found in flat namespace '_CFRelease'
```

This seems to be caused by
`com_google_protobuf/src/google/protobuf/compiler/protoc_nowkt`
was linked with missing symbols on MacOS.  Overriding the default link
opts to add `-framework CoreFoundation` fixes the issue.

---

**Version bump**: Patch

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -135,6 +135,8 @@ http_archive(
 
 http_archive(
     name = "com_google_protobuf",
+    patch_args = ["-p1"],
+    patches = ["//buildpatches:protobuf.patch"],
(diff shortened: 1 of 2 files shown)


sluongng:
also sent the fix to https://github.com/protocolbuffers/protobuf/pull/12320 and left some comments in https://github.com/protocolbuffers/protobuf/issues/12173

PR #3619, opened 2023-03-22
More snaploader refactoring/cleanup

bduffany:
Taking a few small steps towards supporting remote snapshot sharing and supporting `AllowSnapshotStart`, which will let us start from a snapshot even if the executor is restarted and the runner pool is empty.

* Use an `Action` as the key for snapshots, with `Platform` as part of the key rather than `ConfigurationHash()`, as a step towards representing the manifest as an `ActionResult`.
  * The eventual goal is to store the manifests in AC (which only allows ActionResult values), which will also let us use `dirtools` to download the "input" files in the snapshot.
  * Using the AC also naturally lets us partition by group ID and instance name, as well as use Platform as a way to match new actions with existing snapshots, even if there is no runner in the runner pool corresponding to the snapshot.
  * To prevent the snapshots from getting too stale, we can use the current day (or hour, etc.) as part of the Action `salt`, e.g. "2023-01-02"

* Explicitly represent the runner ID as part of the snapshot filecache key, to avoid potential bugs with runners trying to use the same snapshot concurrently.
  * Previously, the per-runner cache isolation was implicit in the manifest's absolute file path, which is unique per runner. If we want to migrate the manifest to be an ActionResult, we won't want to deal with absolute paths.

* Require a group ID for snapshot access, and explicitly represent the authenticated group ID as part of the snapshot's filecache key.
  * Previously, the group-level isolation was implicit in the fact that runners have unique working directories, and that 

* Simplify snapshot saving/loading so that the firecracker container doesn't need to keep state about the paused snapshot digest.

---

**Version bump**: Patch


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -340,25 +339,16 @@ type FirecrackerContainer struct {
...
-func (c *FirecrackerContainer) ConfigurationHash() *repb.Digest {
-	params := []string{
-		fmt.Sprintf("cpus=%d", c.constants.NumCPUs),
-		fmt.Sprintf("mb=%d", c.constants.MemSizeMB),
-		fmt.Sprintf("scratch=%d", c.constants.ScratchDiskSizeMB),
-		fmt.Sprintf("net=%t", c.constants.EnableNetworking),
-		fmt.Sprintf("dockerd=%t", c.constants.InitDockerd),
-		fmt.Sprintf("debug=%t", c.constants.DebugMode),
-		fmt.Sprintf("container=%s", c.containerImage),
-	}
```
tylerwilliams, line 354: these were effectively invariants -- because they cannot be changed between vm runs from a snapshot. Are you sure you want to remove these rather than incorporating them into the key?

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -583,39 +593,37 @@ func (c *FirecrackerContainer) SaveSnapshot(ctx context.Context, instanceName st
...
 	}
 	log.CtxDebugf(ctx, "VMM ResumeVM took %s", time.Since(resumeStart))
 
-	return snapshotDigest, nil
+	return nil
 }
 
 // LoadSnapshot loads a VM snapshot from the given snapshot digest and resumes
 // the VM. If workspaceDirOverride is set, it will also hot-swap the workspace
 // drive; otherwise, the workspace will be loaded as-is from the snapshot.
```
tylerwilliams, line 619: nit: comment still mentions "given snapshot digest"

tylerwilliams:
I know I'm a little early to the party here ( but it'd be nice to break this up (maybe along the lines of your bullet-pointed features in the CL desc?). I don't think any of it is too bad, but taken together it's kind of a lot to review!

enterprise/server/remote_execution/containers/firecracker/firecracker.go
tylerwilliams: Yeah I could imagine introducing a default that we don't allow specifying via a platform prop that is also an invariant.

I like your suggestion of hashing the constants struct and including it in the action salt.

bduffany:
Will close this out since I'm just using it as a reference while flushing smaller PRs.

PR #3620, opened 2023-03-22
Verify downloaded file hash

tylerwilliams:
Verify the hash of blobs downloaded with cachetools.GetBlobChunk.

### DIFF
--- a/server/remote_cache/cachetools/cachetools.go
+++ b/server/remote_cache/cachetools/cachetools.go
@@ -49,7 +49,7 @@ type nopCloser struct {
 
 func (nopCloser) Close() error { return nil }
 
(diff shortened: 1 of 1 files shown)


server/remote_cache/cachetools/cachetools.go
```diff
@@ -101,6 +106,10 @@ func GetBlobChunk(ctx context.Context, bsClient bspb.ByteStreamClient, r *digest
 		}
 		written += int64(n)
 	}
+	computedDigest := fmt.Sprintf("%x", checksum.Sum(nil))
```
bduffany, line 109: should the actual checksum go in `GetBlob` since the chunk digest won't necessarily match the full blob digest? (or only do the checksum if Offset=0 and Limit=0)

tylerwilliams: Ah yeah that's a good point.

tylerwilliams: Done.

PR #3623, opened 2023-03-22
Factor out OAuth handler from GitHub client

bduffany:
A GitHub App is also an OAuth app, so it will need to reuse this same OAuth code.

---

**Version bump**: Patch


tylerwilliams (approved):
nice

PR #3625, opened 2023-03-22
Add tables for GitHub app

bduffany:
Might still need to tweak uniqueness constraints etc. (e.g. can a repo be assigned to multiple groups?) But I think this is a good starting point.

---

**Version bump**: Patch

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -489,8 +489,65 @@ func (ts *TargetStatus) TableName() string {
 	return "TargetStatuses"
 }
 
+// GitHubAppInstallation represents a BuildBuddy GitHub App installation linked
+// to an organization.
+//
+// We'll listen to app uninstallation webhook events to proactively remove these
+// from the DB, but this is not 100% reliable, so GitHub is ultimately the
+// source of truth for whether an installation is valid or not. Typically, the
+// app can optimistically try to create a token for the installation ID, and
+// GitHub will return an error due to the installation no longer existing or no
+// longer being enabled for a particular repo.
+type GitHubAppInstallation struct {
+	Model
+
+	// UserID is the user that registered this installation to BuildBuddy.
+	UserID string `gorm:"not null"`
+
+	// GroupID is the group linked to the GitHub app installation.
+	GroupID string `gorm:"primaryKey"`
+	Perms   int    `gorm:"not null"`
+
+	// InstallationID is the GitHub app installation ID.
+	InstallationID int64 `gorm:"not null"`
+
+	// Owner is the GitHub login of the installation (either a user or
+	// organization name). This can be computed from the installation ID, but we
+	// store it here so that we can run queries to associate repos with
+	// installations.
+	Owner string `gorm:"primaryKey"`
+}
+
+func (gh *GitHubAppInstallation) TableName() string {
+	return "GitHubAppInstallations"
+}
+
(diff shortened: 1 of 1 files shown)


tylerwilliams:
With our new auto-release flow schema changes must be applied out-of-band. The auto-release job won't apply them (intentionally), and cannot proceed until they are applied.

Just mentioning this because you'll have to apply this multiple times if you want to tweak it later. (So if it's not too much work to figure out uniqueness constraints now, it might be worth trying to do that?)

server/tables/tables.go
```diff
@@ -489,8 +489,61 @@ func (ts *TargetStatus) TableName() string {
...
+	// GroupID is the group linked to the GitHub app installation.
+	GroupID string `gorm:"primaryKey"`
+	Perms   int
+	// InstallationID is the GitHub app installation ID.
+	InstallationID int64 `gorm:"not null"`
+	// Owner is the GitHub login of the installation (either a user or
+	// organization name). This can be computed from the installation ID, but we
+	// store it here so that we can run queries to associate repos with
+	// installations.
+	Owner string `gorm:"primaryKey"`
```
tylerwilliams, line 515:
```suggestion
	// UserID is the user that registered this installation to BuildBuddy.
	UserID string `gorm:"not null;default:''"`

	// GroupID is the group linked to the GitHub app installation.
	GroupID string `gorm:"primaryKey"`
	Perms   int

	// InstallationID is the GitHub app installation ID.
	InstallationID int64 `gorm:"not null"`

	// Owner is the GitHub login of the installation (either a user or
	// organization name). This can be computed from the installation ID, but we
	// store it here so that we can run queries to associate repos with
	// installations.
	Owner string `gorm:"primaryKey"`
```

server/tables/tables.go
```diff
@@ -489,8 +489,61 @@ func (ts *TargetStatus) TableName() string {
...
+
+	// RepoURL is the normalized repo URL.
+	RepoURL string `gorm:"primaryKey;unique"`
+	UserID  string
+	GroupID string `gorm:"primaryKey"`
+	Perms   int
+	// InstanceNameSuffix is the remote instance name suffix to apply to any
+	// BuildBuddy-initiated invocations within this repo, such as workflows or
+	// remote Bazel.
+	InstanceNameSuffix string `gorm:"default:''"`
```
tylerwilliams, line 534:
```suggestion
	// RepoURL is the normalized repo URL.
	RepoURL string `gorm:"primaryKey;unique"`
	UserID  string
	GroupID string `gorm:"primaryKey"`
	Perms   int

	// InstanceNameSuffix is the remote instance name suffix to apply to any
	// BuildBuddy-initiated invocations within this repo, such as workflows or
	// remote Bazel.
	InstanceNameSuffix string `gorm:"default:''"`
```

tylerwilliams (approved):
minor nit on the formatting but LGTM

bduffany:
> With our new auto-release flow schema changes must be applied out-of-band. The auto-release job won't apply them (intentionally), and cannot proceed until they are applied.
> 
> Just mentioning this because you'll have to apply this multiple times if you want to tweak it later. (So if it's not too much work to figure out uniqueness constraints now, it might be worth trying to do that?)

ACK, after taking another glance, I'm fine with the uniqueness constraints in their current state. We can later make repo URL non-unique if we find there's a use case.

PR #3628, opened 2023-03-22
Support user-level GitHub unlink

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -361,6 +361,20 @@ func (d *UserDB) CreateUserAPIKey(ctx context.Context, groupID, label string, ca
 	return createdKey, nil
 }
 
+func (d *UserDB) DeleteUserGitHubToken(ctx context.Context) error {
+	u, err := perms.AuthenticatedUser(ctx, d.env)
+	if err != nil {
+		return err
+	}
+	if u.GetUserID() == "" {
+		return status.FailedPreconditionError("user ID must not be empty")
+	}
+	return d.h.DB(ctx).Exec(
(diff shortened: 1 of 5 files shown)


siggisim (approved):

proto/github.proto
```diff
@@ -8,6 +8,9 @@ package github;
 // associated with the group selected in the request context.
 message UnlinkGitHubAccountRequest {
   context.RequestContext request_context = 1;
+
+  // Whether to unlink the user-level account.
+  bool user = 2;
```
tylerwilliams, line 13: nit: maybe unlink_user_account is a clearer name for this?

bduffany: Done

enterprise/server/backends/userdb/userdb.go
```diff
@@ -361,6 +361,20 @@ func (d *UserDB) CreateUserAPIKey(ctx context.Context, groupID, label string, ca
...
+func (d *UserDB) DeleteUserGitHubToken(ctx context.Context) error {
+	u, err := perms.AuthenticatedUser(ctx, d.env)
+	if err != nil {
+		return err
+	}
+	if u.GetUserID() == "" {
+		return status.FailedPreconditionError("user ID must not be empty")
+	}
+	return d.h.DB(ctx).Exec(
+		`UPDATE Users SET github_token = "" WHERE user_id = ?`,
```
vadimberezniker, line 373: To support postgres, we are going to use more standard sql quoting which uses single quotes for string values and double quotes for identifiers. Let's start using single quotes for strings for any new queries we add.

bduffany: Done - and will keep this in mind

PR #3630, opened 2023-03-23
blobstore: support path_prefix for storing blobs

### LINKED ISSUE #3595: Allow to set prefix for S3 storage
We use the same s3 bucket for buildbuddy and [bazel-remote](https://github.com/buchgr/bazel-remote). The latter allows to specify prefix via config, the former doesn't. It makes it hard to navigate bucket since the root is polluted and also to specify different lifecycle policies for cache and buildbuddy reports.

Basically, we want this kind of structure:
```
cache/
   ac/
   cas.v/
buildbuddy/
   24971223-462c-4a5d-85e7-3c6f8c50f0a6/
   24989c17-ab53-4744-ace6-d76ddda9d720/
   ...
```

sluongng:
This is a follow up from #3597 where were added a feature to allow users
to specify a path prefix in their S3 to store cache objects. This
enables users to configure BuildBuddy with a single bucket.

Apply the same change to blobstore so user could have this configuration
for build events as well.

Closes #3595

---

**Version bump**: Patch

### DIFF
--- a/server/backends/blobstore/blobstore_test.go
+++ b/server/backends/blobstore/blobstore_test.go
@@ -15,13 +15,13 @@ func TestDiskBlobStore(t *testing.T) {
 		blob             []byte
 	}{
 		{
-			"without_prefix",
+			"WithoutPrefix",
 			"",
 			"test_blob",
 			[]byte("test"),
 		},
 		{
-			"with_prefix",
+			"WithPrefix",
 			"my_prefix",
 			"test_blob",
 			[]byte("test"),


tylerwilliams (approved):
LGTM

server/backends/blobstore/blobstore.go
```diff
@@ -505,7 +511,7 @@ func (a *AwsS3BlobStore) createBucketIfNotExists(ctx context.Context, bucketName
 
 func (a *AwsS3BlobStore) ReadBlob(ctx context.Context, blobName string) ([]byte, error) {
 	start := time.Now()
-	b, err := a.download(ctx, blobName)
+	b, err := a.download(ctx, blobPath(blobName))
```
bduffany, line 514: nit: in the download func below, rename `blobName` to `path`. Or (probably cleaner): revert the change on this line, and instead change line 525 to `Key: aws.String(blobPath(blobName))`

server/backends/blobstore/blobstore.go
```diff
@@ -537,15 +543,15 @@ func (a *AwsS3BlobStore) WriteBlob(ctx context.Context, blobName string, data []
 		return 0, err
 	}
 	start := time.Now()
-	n, err := a.upload(ctx, blobName, compressedData)
+	n, err := a.upload(ctx, blobPath(blobName), compressedData)
```
bduffany, line 546: `upload()` is already calling `blobPath()`, so this line should be reverted to avoid double-prefixing:

```suggestion
	n, err := a.upload(ctx, blobName, compressedData)
```

server/backends/blobstore/blobstore.go
```diff
@@ -558,7 +564,7 @@ func (a *AwsS3BlobStore) upload(ctx context.Context, blobName string, compressed
 
 func (a *AwsS3BlobStore) DeleteBlob(ctx context.Context, blobName string) error {
 	start := time.Now()
-	err := a.delete(ctx, blobName)
+	err := a.delete(ctx, blobPath(blobName))
```
bduffany, line 567: nit: call `blobPath` in `delete()` instead

server/backends/blobstore/blobstore.go
```diff
@@ -224,7 +226,7 @@ func (d *DiskBlobStore) blobPath(blobName string) (string, error) {
 	if strings.Contains(blobName, "..") {
 		return "", fmt.Errorf("blobName (%s) must not contain ../", blobName)
 	}
-	return filepath.Join(d.rootDir, blobName), nil
+	return filepath.Join(*pathPrefix, d.rootDir, blobName), nil
```
bduffany, line 229: The `pathPrefix` should go after `rootDir` here; probably would be clearer to use `blobPath()` to be consistent with how we construct the paths for the other blobstores.

```suggestion
	return filepath.Join(d.rootDir, blobPath(blobName)), nil
```

bduffany (changes requested):

server/backends/blobstore/blobstore.go
sluongng: Agree! Changing line 525 made it looks a lot cleaner

server/backends/blobstore/blobstore.go
sluongng: fixed

server/backends/blobstore/blobstore.go
sluongng: nice catch

sluongng:
I pushed a newer version according to the review feedback.

There was a small bug caught in `DiskBlobStore.DeleteBlob()` where we were running `os.Stat(blobName)` instead of `os.Stat(fullPath)`. I added a basic test to prevent future regressions.

@bduffany PTAL 🙇

server/backends/blobstore/blobstore_test.go
```diff
@@ -0,0 +1,61 @@
...
+		blob             []byte
+	}{
+		{
+			"without_prefix",
+			"",
+			"test_blob",
+			[]byte("test"),
+		},
+		{
+			"with_prefix",
```
bduffany, line 24: uber nit: since the `name` becomes part of the test name, it looks slightly better IMO to have it in CapitalCase

TestDiskBlobStore/WithoutPrefix: PASS
TestDiskBlobStore/WithPrefix: PASS

```suggestion
			"WithoutPrefix",
			"",
			"test_blob",
			[]byte("test"),
		},
		{
			"WithPrefix",
```

bduffany (approved):

bduffany (approved):

PR #3632, opened 2023-03-23
Set flaky_test_attempts for Mac workflow

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -62,8 +62,7 @@ build:untrusted-ci --repository_cache=/home/runner/repo-cache/untrusted/
 build:untrusted-ci --disk_cache=
 build:untrusted-ci --flaky_test_attempts=2
 
-# Configuration used for BuildBuddy workflows
-build:workflows --config=remote-shared
+# Configuration used for all BuildBuddy workflows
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -77,18 +76,18 @@ build:workflows --flaky_test_attempts=//enterprise/server/test/webdriver/invocat
...
+# Configuration used for Linux workflows
+build:linux-workflows --config=remote-shared
+build:linux-workflows --config=workflows
+build:linux-workflows --config=buildbuddy_remote_executor
+
+# Configuration used for Mac workflows
 # TODO(bduffany): Enable RBE for Mac workflows, and reconcile this with other configs
 build:mac-workflows --config=cache
-build:mac-workflows --remote_instance_name=buildbuddy-io/buildbuddy/workflows
-build:mac-workflows --experimental_remote_cache_compression
```
tylerwilliams, line 87: did this get dropped? Trying to find where it gets pulled in as part of workflows but not sure it's there (or maybe i'm missing it)

bduffany: experimental_remote_cache_compression is already set in `--config=cache` so it should be safe to remove

PR #3634, opened 2023-03-23
Use a deterministic cache key for firecracker snapshots

bduffany:
Instead of making the snapshot digest optional (`opts.ForceSnapshotDigest`) and/or stateful (`c.pausedSnapshotDigest`), always use a non-nil digest that is deterministically computed from the VM configuration hash (and runner ID, for now[^1]).

This allows us to simplify the logic around saving/loading snapshots, while also resolving one of the blockers for snapshot sharing, since snapshot sharing will require computing a deterministic snapshot digest based solely on the VM configuration required by the task, and then checking the cache (local and/or remote) to see whether a matching snapshot is available.

[^1]: In the future, we'll remove runner ID to enable snapshot sharing - but we are blocked on having a way to ensure that runners can't concurrently access the same snapshot

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -2,7 +2,6 @@ package snaploader
 
 import (
 	"context"
-	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -22,6 +21,22 @@ const (
...
+// NewKey returns the cache key that can be used to look up the snapshot
+// manifest.
+// TODO(bduffany): once runners are capable of sharing snapshots, remove
+// runnerID from the digest hash.
+// TODO(bduffany): use ResourceName instead of digest and incorporate
+// remote_instance_name. The instance name is currently implicit in the runnerID
+// since we only match tasks to runners if their instance name matches the
+// runner's instance name. But once we have snapshot sharing, runnerID will no
+// longer be part of the snapshot key.
+func NewKey(configurationHash, runnerID string) *repb.Digest {
```
bduffany, line 33: FYI, I am calling this "Key" instead of "Digest", because in a later PR I'm planning to add a `Key` struct which contains `ResourceName` and `RunnerID` (`RunnerID` will be used to prevent concurrent access by multiple runners, and is only relevant to the local filecache, not the remote cache).

vadimberezniker (approved):

PR #3637, opened 2023-03-23
A few more small heatmap bug fixes

jdhollen:
- Use last-hovered selection if the user drags their mouse outside of the heatmap area--this seems less surprising than just dropping the selection attempt (Maggie and Brandon both had feedback to this effect).
- Clear selection when the user clicks on a drilldown bar (otherwise, the user keeps seeing a list of invocations and bar charts even though the new heatmap doesn't show a selection).
- Clear the active selection when zoom is cleared.
- Scroll to top when a bar is selected for filtering.
---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/trends/drilldown_page.tsx
+++ b/enterprise/app/trends/drilldown_page.tsx
@@ -506,6 +506,7 @@ export default class DrilldownPageComponent extends React.Component<Props, State
   handleClearZoom() {
     router.setQuery({
       ...Object.fromEntries(this.props.search.entries()),
+      [DD_SELECTED_AREA_URL_PARAM]: "",
       [DD_ZOOM_URL_PARAM]: "",
     });
   }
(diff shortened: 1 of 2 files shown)


enterprise/app/trends/drilldown_page.tsx
```diff
@@ -506,6 +506,7 @@ export default class DrilldownPageComponent extends React.Component<Props, State
   handleClearZoom() {
     router.setQuery({
       ...Object.fromEntries(this.props.search.entries()),
+      [DD_SELECTED_AREA_URL_PARAM]: "",
```
maggie-lou, line 509: This field doesn't seem to be used anywhere. Maybe this PR is missing the changes for "Clear the active selection when zoom is cleared."? Don't see changes for that anywhere

maggie-lou:
Another piece of non-blocking minor design feedback: 

When you click on a drilldown bar, the page reloads and a filter is added. Because I had scrolled down on the page and couldn't see the filter section, when I clicked a bar, the page reloaded and it wasn't immediately clear what had happened. Maybe on page reload we should scroll up to the top of the page so the filter section is always visible?

enterprise/app/trends/drilldown_page.tsx
jdhollen: This is already read at L466: https://github.com/buildbuddy-io/buildbuddy/blob/a116cb72b76a3c3b8e22f92c7209841bd8bbd345/enterprise/app/trends/drilldown_page.tsx#L466

This change is just fixing a case where this field wasn't getting reset

jdhollen:
> Another piece of non-blocking minor design feedback:
> 
> When you click on a drilldown bar, the page reloads and a filter is added. Because I had scrolled down on the page and couldn't see the filter section, when I clicked a bar, the page reloaded and it wasn't immediately clear what had happened. Maybe on page reload we should scroll up to the top of the page so the filter section is always visible?

Sure, I added this, I hadn't noticed this because my screen is kinda huge.

enterprise/app/trends/drilldown_page.tsx
maggie-lou: Ah my bad - didn't have master pulled in my local branch

maggie-lou (approved):

PR #3640, opened 2023-03-24
sluongng/executor terraform

sluongng:
- Rename terraform.tf to provider.tf
- Add terraform for executor deployment

---

**Version bump**: Patch


sluongng:
There seems to be 2 way to push this forward:

1. To create a new default executor image where we don't execute `docker-credential-gcr configure-docker` and update Docker's `config.json` file.

2. Modifying podman/docker/firecracker isolation type to support an explicit `anonymous_image_download` option where no authentication is provided. In such cases, the username `anonymous` and empty-string password could be used as various container registries widely accept it.

Initial attempt with (2) shows that it's not simple as we are using various CLIs instead of using the libraries.

I will attempt (1) next.

sluongng:
Side-stepping the runner's image pull authentication problem by using a public Docker Hub image (`ubuntu`) for now.

sluongng:
Will submit the `podman` executor config as an example config in the helm chart.
The terraform is pretty trivial to set up once the helm chart is ready.

PR #3641, opened 2023-03-24
Check `bb fix` diff in checkstyle

bduffany:
This should keep TypeScript BUILD deps in a good state.

---

**Version bump**: Patch


.github/workflows/checkstyle.yaml
```diff
@@ -63,10 +63,10 @@ jobs:
           cat go-deps-diff.txt
 
       - name: gazelle
-        # Keep Gazelle version in sync with WORKSPACE
```
tempoz, line 66: We should also remove the comment in [`WORKSPACE`](https://github.com/buildbuddy-io/buildbuddy/blob/master/WORKSPACE#L35) if we're removing this one.

tempoz (approved):
One nit, other than that looks good!

PR #3644, opened 2023-03-24
Implement UI for GitHub App OAuth flow

bduffany:
* Org-level link now shows a warning if the GitHub App is enabled and an org-level token is linked, and encourages a user-level link instead.
* New user-level link page points to the GitHub App OAuth flow, and shows a profile picture.

---

image

---

image

---

**Version bump**: Patch


app/components/banner/banner.css
```diff
@@ -6,6 +6,14 @@
   align-items: start;
 }
 
+.banner .banner-content > p:first-child {
```
jdhollen, line 9: these margin overrides (basically content styling) feel like they'd be better placed in settings.css since the input child element for the banner is just an arbitrary div

enterprise/app/settings/settings.css
```diff
@@ -144,12 +154,24 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
   padding: 8px 16px;
   position: relative;
   overflow: hidden;
   transition: box-shadow 100ms ease-out;
 }
 
+.settings .settings-tab > * {
```
jdhollen, line 166: Maybe better to just enclose the tab in the click listener or check up the tree?

enterprise/app/settings/user_github_link.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { github } from "../../../proto/github_ts_proto";
+
+export interface Props {
+  user: User;
+}
+
+export interface State {
+  deleteModalVisible?: boolean;
+  isDeleting?: boolean;
+  isRefreshingUser?: boolean;
```
jdhollen, line 32: since this is internal state (i.e., the caller doesn't need to specify any of these), i think it's better to just have them be non-optional and init to false in the constructor, but i'm admittedly not up on what the cool kids do in react.

enterprise/app/settings/user_github_link.tsx
```diff
@@ -0,0 +1,188 @@
...
+        .getAuthenticated()
+        .then((response) => {
+          this.setState({
+            account: {
+              name: response.data.name || "",
+              login: response.data.login,
+              avatarUrl: response.data.avatar_url,
+            },
+          });
+        })
```
jdhollen, line 80: missing an error catch here?

jdhollen (approved):
Just some minor stuff

enterprise/app/settings/settings.css
bduffany: Ah yeah this was some residual hackathon laziness :laughing: 

Fixed by using a `<Link>` component for `SettingsTab` and getting rid of the manual `onClick` handler

enterprise/app/settings/user_github_link.tsx
bduffany: Agreed - dealing with undefined is annoying. Done!

enterprise/app/settings/user_github_link.tsx
bduffany: woops, done

PR #3648, opened 2023-03-24
Exclude more flaky tests from Mac workflows

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -25,6 +25,8 @@ actions:
         test --config=mac-workflows
         --
         //...
+        -//enterprise/server/raft/store:all
+        -//enterprise/server/remote_execution/commandutil:all
         -//enterprise/server/test/integration/ci_runner:all
         -//enterprise/server/test/integration/workflow:all
         -//enterprise/server/test/integration/remote_execution/...


buildbuddy.yaml
```diff
@@ -25,6 +25,8 @@ actions:
         test --config=mac-workflows
         --
         //...
+        -//enterprise/server/raft/store:all
```
maggie-lou, line 28: Why are we just excluding the flaky tests for mac?

bduffany: These tests were previously excluded and recently re-enabled; I am just re-disabling them. We can probably increase the timeouts or re-enable them at some point, but I figured it's not worth thinking about right now.

bduffany: Actually, I didn't realize I hadn't submitted https://github.com/buildbuddy-io/buildbuddy/pull/3632 . Let's merge that one first and see if it helps

bduffany: It didn't help - reopening.

maggie-lou: It still says it's closed for me so I can't approve it

PR #3651, opened 2023-03-26
blobstore: refactoring into separate files

sluongng:
This is a no-op refactoring for blobstore code into separate files base
on which storage backend is used. This should help us (and self-hosted
customers) to navigate relevant code a lot easier as often we would
only care about 1 out of 4 available storage backend.

---

**Version bump**: Patch


bduffany (commented):
I am in favor of splitting up into separate files, but historically we have used a convention of having one `.go` file per bazel package (with a few exceptions like platform-specific variants, e.g. `firecracker_darwin.go`).

The reasoning is that if you reference a symbol without a package prefix (like `Bar` instead of `foo_package.Bar`), then you can expect to find that symbol in the same file you're currently in.

In this case, putting these into subdirs like `server/backends/blobstore/aws/aws.go` etc., would also lead to a healthier build graph since these things have BUILD deps that are specific to each cloud provider.

sluongng:
@bduffany I attempted at refactoring it into smaller packages.

However, the original code was quite messy in how the helper funcs and flags are used, leading to the need for `metric` and `util` packages included. This causes a relatively noisy diff, but I believe the final result is a lot cleaner than before.

server/backends/blobstore/metric/metric.go
```diff
@@ -0,0 +1,66 @@
+package metric
```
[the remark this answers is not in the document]
sluongng: Let's move them under util to make it simpler.

server/backends/blobstore/gcp/gcp.go
```diff
@@ -0,0 +1,136 @@
+package gcp
```
bduffany, line 1: ugh, sorry I missed this in the last review round. I think maybe this should be "gcs" to be consistent with the blobstore name. Otherwise LGTM

bduffany (approved):

server/backends/blobstore/gcp/gcp.go
bduffany: auto-merge strikes again :P I can send a PR to take care of it

bduffany: oh that is a good point, I hadn't thought of that. I just named it based on the blobstore name, like `NewAWSBlobstore` etc

PR #3654, opened 2023-03-27
testfs: avoid collisions in temp dirs/files

sluongng:
In rare scenarios where the test execution is run locally on a weaker
sandbox environment (darwin), a single random digit temp file/dir could
cause collisions and led to 1 test deleting another test's files/dirs.

This causes flakiness in our tests when executing in Darwin, leading to
timeout in some scenarios. Example:

https://app.buildbuddy.io/invocation/cea0ac33-beca-4793-9588-87528336f277?target=%2F%2Fserver%2Fbackends%2Fdisk_cache%3Adisk_cache_test

---

**Version bump**: None


sluongng:
wrong approach, let's close this to investigate further

PR #3655, opened 2023-03-27
sluongng/redirect https port

### LINKED ISSUE #3653: ssl.upgrade_insecure uses wrong port
The `ssl.upgrade_insecure: true` setting issues a redirect to `https://` but uses the `http` port.

E.g.

```
curl -i  http://example.com:1980/   
HTTP/1.1 301 Moved Permanently
Content-Type: text/html; charset=utf-8
Location: https://example.com:1980/
Date: Sun, 26 Mar 2023 20:36:07 GMT
Content-Length: 62

<a href="https://example.com:1980/">Moved Permanently</a>.
```

`ssl.enable_ssl` is `true` and `--ssl_port` is specified as `1981`.

`app.build_buddy_url` is specified as `https://example.com:1981`.

sluongng:
When we redirect http traffic to https, make sure that we use the configured ssl server's address and port
instead of the original request's host.

Refactor RedirectIfNotForwardedHTTPS to do the following:

1. Added a basic unit test to verify current behavior.

2. Remove env parameter, which is no longer needed since a637a0cad803d6f6159496c77f21496875e15eef.

3. Pass sslServer.Addr from libmain down to the interceptor and use it as the redirected host.

---

**Version bump**: Patch 

Fixes #3653

### DIFF
--- a/server/http/interceptors/interceptors_test.go
+++ b/server/http/interceptors/interceptors_test.go
@@ -75,7 +75,7 @@ func TestRedirectIfNotForwardedHTTPS(t *testing.T) {
 			}
 
 			rr := httptest.NewRecorder()
-			RedirectIfNotForwardedHTTPS(nil, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			RedirectIfNotForwardedHTTPS(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				fmt.Fprint(w, r.Header)
 			})).ServeHTTP(rr, req)
 
--- a/server/http/interceptors/interceptors_test.go
+++ b/server/http/interceptors/interceptors_test.go
@@ -1,0 +1,89 @@
+package interceptors
+
+import (
+	"fmt"
+	"net/http"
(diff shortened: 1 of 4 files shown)


server/http/interceptors/interceptors_test.go
```diff
@@ -0,0 +1,107 @@
...
+			http.StatusOK,
+			http.Header{},
+			"example.com",
+			func(req *http.Request) {
+				req.Header.Del("X-Forwarded-Proto")
+				req.Header.Set("User-Agent", "GoogleHC/1.0")
+			},
+		},
+		{
+			"custom ssl address",
```
vadimberezniker, line 74: Use named fields so it's more obvious what each of these values mean w/o having to look back at the struct definition.

vadimberezniker (approved):

siggisim:
Wrote some info on why I don't think we should do this here: https://github.com/buildbuddy-io/buildbuddy/issues/3653#issuecomment-1485981369

sluongng:
Oh i see. Let's close this then.

PR #3656, opened 2023-03-27
Ignore directory_output in TargetComplete events

bduffany:
These artifacts aren't very useful since according to the BES proto, they will never have a URI set, and also the current logic is preventing `output_group` files from being visible when `directory_output` is present.

---

**Version bump**: Patch


bduffany:
Closing because the files are actually visible in `important_output`, not `output_group`, and we aren't reading `important_output` currently (and the BES proto recommends not to depend on it). Need to figure out an alternate approach to make these files visible.

sluongng (approved):
LGTM

PR #3660, opened 2023-03-27
Silence gorm logging during print schema script or output will interfere

maggie-lou:
If there is a slow query that takes > 200ms, GORM will print that to stdout by default. The schema diff script assumes that we are only logging to stdout if there is a schema change, so we want to silence any other output while it's running.


maggie-lou:
* master
  * **PR #3660** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3660" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3660?utm_source=stack-comment).

iain-macdonald (approved):

PR #3662, opened 2023-03-27
Add protos for GH app installations and git repositories APIs

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/proto/github.proto
+++ b/proto/github.proto
@@ -115,11 +115,16 @@ message GetReposRequest {
 
   // The GitHub username or organization name to search within. The
   // authenticated user must have access to a GitHub App installation associated
-  // with the repo
+  // with the repo.
+  // Required.
   string owner = 2;
 
-  // The search query (optional). If no query is specified, list all
-  // repositories.
+  // The plain text search query attempting to match all or part of the repo
+  // name. This is interpreted according to GitHub's repository search syntax:
+  // https://docs.github.com/en/rest/search?apiVersion=2022-11-28#search-repositories
+  //
+  // Note that an additional filter is applied to the query to restrict the
+  // search to repos owned by the specified owner.
   string query = 3;
 }
 
--- a/proto/github.proto
+++ b/proto/github.proto
@@ -21,3 +21,155 @@ message UnlinkGitHubAccountResponse {
   // the GitHub UI.
   repeated string warning = 2;
 }
+
+// GitHub App installations API.
+//
+// Users can use GitHub to install and uninstall the BuildBuddy GitHub App,
+// without using the BuildBuddy UI as an entrypoint. After installing the app on
+// GitHub, the user is redirected by GitHub back to the app, which uses these
+// APIs to "link" the new installation to the currently selected organization.
+//
+// A linked installation allows any build within the org to get installation
+// access tokens. For example, when running an authenticated build with ROLE=CI,
+// the commit status reporter will attempt to authorize via an app installation
+// token associated with the repository, by parsing the repo URL owner (username
+// or org name) and looking for a BuildBuddy-linked installation matching that
+// owner.
+//
+// An app installation can only be linked to one BuildBuddy org. These APIs can
+// be used to re-assign the org that an installation is linked to, or to unlink
+// an installation from an org (note that unlinking is not the same as
+// uninstalling).
+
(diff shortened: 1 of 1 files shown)


proto/github.proto
```diff
@@ -21,3 +21,170 @@ message UnlinkGitHubAccountResponse {
...
+    // The repo URL.
+    // Example: "https://github.com/buildbuddy-io/buildbuddy"
+    string repo_url = 1;
+
+    // Whether the repository is linked to the authenticated organization.
+    bool linked = 2;
+  }
+
+  // The repos matching the request criteria.
+  repeated Repo repos = 2;
```
jdhollen, line 157: we have a few random things that violate this, but generally speaking the proto convention is to use singular nouns even for repeated fields.   pretty sure this convention arises from how lists are accessed in some languages.

https://github.com/search?q=repo%3Abuildbuddy-io%2Fbuildbuddy+%22repeated+%22&type=code&p=2

proto/github.proto
```diff
@@ -21,3 +21,170 @@ message UnlinkGitHubAccountResponse {
...
+  string repo_url = 2;
+}
+
+message LinkRepoResponse {
+  context.ResponseContext response_context = 1;
+}
+
+// A request to unlink a repo from the authenticated org.
+message UnlinkRepoRequest {
+  context.RequestContext request_context = 1;
```
jdhollen, line 185: should this have a repo url?

proto/github.proto
```diff
@@ -21,3 +21,170 @@ message UnlinkGitHubAccountResponse {
...
+//
+// Unlinking an installation does not automatically unlink all repositories
+// associated with the installation. However, members of the organization will
+// no longer be able to get repository access tokens (unless the installation is
+// re-linked).
+message UnlinkAppInstallationRequest {
+  context.RequestContext request_context = 1;
+
+  // The installation ID to unlink.
+  int64 installation_id = 2;
```
jdhollen, line 123: if `UpdateAppInstallationRequest` needs `group_id` i assume this needs one, too? Or does `Update` not actually need it?

jdhollen (changes requested):

proto/github.proto
bduffany: So re. `Update` having a group ID, my original thinking was that in the settings page, we can allow the user to see all the GitHub installations they have access to, as well as which organizations they are linked to (if any). And if the installation is linked to a different org, the user can use this RPC to re-assign the the installation to a different org.

But it probably won't be super common that users want to re-assign an installation to a different org, so I've removed the group_id field from `UpdateAppInstallationRequest`. Instead of re-assigning, they can always Unlink then re-link (as a different group).

re. `Unlink` not needing a group_id, we don't need a group ID because installations can only be assigned to one group, so the group_id comes from the GitHubAppInstallation row in the DB.

proto/github.proto
bduffany: Fixed

proto/github.proto
bduffany: woops, yeah

proto/github.proto
```diff
@@ -21,3 +21,155 @@ message UnlinkGitHubAccountResponse {
...
+  context.RequestContext request_context = 1;
+
+  // The GitHub username or organization name to search within. The
+  // authenticated user must have access to a GitHub App installation associated
+  // with the repo
+  string owner = 2;
+
+  // The search query (optional). If no query is specified, list all
+  // repositories.
+  string query = 3;
```
vadimberezniker, line 123: What's the format of this query?

vadimberezniker (approved):

proto/github.proto
bduffany: lol, it looks like I didn't actually fix it in my original commit. I will just keep it as-is then :sweat_smile:

proto/github.proto
bduffany: Added more detail

proto/github.proto
jdhollen: Getting rid of updated makes sense to me.  One remaining wrinkle is that this means an "unlink" operation will blow away the linkage even if someone else has made a change to the linkage in the meantime.  This probably doesn't matter, but the group ID could be useful here just to confirm that you're operating on the state that you meant to operate on.

PR #3671, opened 2023-03-28
rbeperf: move to enterprise/tools

sluongng:
RBE is an enterprise-only feature of BuildBuddy. Our bench test tool
rbeperf has dependencies on //enterprise source code and therefore,
could not be built from FOSS version of the repo.

Because our internal repo reference this tool, we could not move this
under //enterprise in 1 PR. Instead, let's do the following:

1. Create an identical copy of the tool under //enterprise.

2. Switch references in internal to the new location.

3. Remove the old location and related visibility settings.

This PR should complete (1).

---

**Version bump**: None


sluongng:
Yup, that's fine as well. Let me combine the changes and submit another PR

PR #3673, opened 2023-03-28
Fix remote ActionResult fetches in the invocation action page

bduffany:
Fixes a regression in the change here: https://github.com/buildbuddy-io/buildbuddy/pull/3299/files#diff-01b2f827d97e58166368c89e6233cb1f240a2b9964f79c1a676a5e6786122b30R177-R184

That change caused us to stop passing the gRPC target in the bytestream URL, but the target is required in some cases (e.g. for europe.remote.buildbuddy.io)

---

**Version bump**: Patch


sluongng (approved):
Previously `address` would always be empty. This fixes that.

LGTM

PR #3677, opened 2023-03-28
Support streaming some but not all images

iain-macdonald:
This PR gets rid of the `executor.podman.image_streaming.enabled` flag in favor of a new `executor.podman.streamable_images` flag that defined a comma-separated list of images that can be streamed. This is intended to be used so we can stream some images but rely on the old pull-and-run for images that don't have a soci index stored in the container registry.

**Version bump**: Minor

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -22,6 +22,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
 	"github.com/buildbuddy-io/buildbuddy/server/util/background"
 	"github.com/buildbuddy-io/buildbuddy/server/util/disk"
+	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/networking"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -39,10 +39,11 @@ var (
...
-	imageStreamingEnabled = flag.Bool("executor.podman.image_streaming.enabled", false, "Whether container image streaming is enabled by default")
-	pullTimeout           = flag.Duration("executor.podman.pull_timeout", 10*time.Minute, "Timeout for image pulls.")
+	memUsagePathTemplate = flag.String("executor.podman.memory_usage_path_template", "/sys/fs/cgroup/memory/libpod_parent/libpod-{{.ContainerID}}/memory.usage_in_bytes", "Go template specifying a path pointing to a container's current memory usage, in bytes. Templated with `ContainerID`.")
+	cpuUsagePathTemplate = flag.String("executor.podman.cpu_usage_path_template", "/sys/fs/cgroup/cpuacct/libpod_parent/libpod-{{.ContainerID}}/cpuacct.usage", "Go template specifying a path pointing to a container's total CPU usage, in CPU nanoseconds. Templated with `ContainerID`.")
+	// TODO(iain): switch to boolean flag once image streaming supports all container images.
+	streamableImages = flag.String("executor.podman.streamable_images", "", "Comma-separated list of images that can be streamed by podman. Image names must not contain commas!")
```
vadimberezniker, line 45: Our flag utils support slice flags. For example, take a look at `remote_execution.sharded_redis.shards`. 
Makes the values easier to write in the configs.

vadimberezniker (changes requested):

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Oh nice, switched to use that instead.

vadimberezniker (approved):

PR #3678, opened 2023-03-28
add a created_at_usec index on Executions

luluz66:
https://github.com/buildbuddy-io/buildbuddy-internal/issues/2172

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -930,7 +930,7 @@ func dropIndexIfExists(m gorm.Migrator, table, indexName string) {
 
 // Manual migration called after auto-migration.
 func PostAutoMigrate(db *gorm.DB) error {
-	indexes := map[string]string{
+	invocationIndices := map[string]string{
 		"invocations_trends_query_index":      "(`group_id`, `updated_at_usec`)",
 		"invocations_trends_query_role_index": "(`group_id`, `role`, `updated_at_usec`)",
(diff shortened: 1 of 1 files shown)


server/tables/tables.go
```diff
@@ -952,32 +952,37 @@ func PostAutoMigrate(db *gorm.DB) error {
...
-			if err := db.Migrator().DropIndex("Invocations", "invocations_test_grid_query_index"); err != nil {
-				log.Errorf("Error dropping deprecated index: %s", err)
+	indicesByTable := map[string]map[string]string{
+		"Invocations": invocationIndices,
+		"Executions":  executionIndices,
+	}
+
+	m := db.Migrator()
+	for tableName, indexes := range indicesByTable {
+		if m.HasTable("Invocations") {
```
vadimberezniker, line 970: This should be `tableName` instead of `Invocations`.

Do you know why we have this check to begin with? I can't think of a circumstance under which these tables would not exist.

vadimberezniker (approved):

server/tables/tables.go
vadimberezniker: Looks like Tyler added it in https://github.com/buildbuddy-io/buildbuddy/commit/f35cf8c0e411f1cab7371180b411ec054928ce8b but I'm not sure why :shrug:

luluz66: Done.

PR #3680, opened 2023-03-29
Add crypter service that supports basic {enc,dec}ryption.

vadimberezniker:
The encryption method is described in this document: https://docs.google.com/document/d/1rpvqs6FX_ibVO11IuOsfDiiZy3C5uv-DCu4RZqi4XYQ/edit

---

**Version bump**: None

### DIFF
--- a/enterprise/server/crypter_service/crypter_service.go
+++ b/enterprise/server/crypter_service/crypter_service.go
@@ -1,0 +1,291 @@
+package crypter_service
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/binary"
+	"fmt"
+	"io"
+
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
+	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"golang.org/x/crypto/hkdf"
+)
+
+const (
+	encryptedDataHeaderSignature = "BB"
+	encryptedDataHeaderVersion   = 1
+	gcmStandardNonceSizeBytes    = 12
+	gcmTagSize                   = 16
+	plainTextChunkSize           = 1024 * 1024 // 1 MiB
+	// 4 byte chunk counter, 24 bytes group ID
+	encryptedChunkAuthHeaderSize = 28
+	encryptedChunkOverhead       = gcmStandardNonceSizeBytes + gcmTagSize
+)
+
+// TODO(vadim): pool buffers to reduce allocations
+// TODO(vadim): include file digest in authentication header
+type Crypter struct {
+	auth interfaces.Authenticator
+	dbh  interfaces.DBHandle
+	kms  interfaces.KMS
+}
+
+func New(env environment.Env) *Crypter {
+	return &Crypter{
+		kms:  env.GetKMS(),
+		auth: env.GetAuthenticator(),
+		dbh:  env.GetDBHandle(),
+	}
+}
+
+type Encryptor struct {
+	key           *tables.EncryptionKeyVersion
+	ciph          cipher.AEAD
+	groupIDHeader []byte
+	w             interfaces.CommittedWriteCloser
+	wroteHeader   bool
+	chunkCounter  uint32
+	nonceBuf      []byte
+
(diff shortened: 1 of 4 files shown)


enterprise/server/crypter_service/crypter_service.go
```diff
@@ -0,0 +1,291 @@
...
+				return 0, err
+			}
+		}
+	}
+
+	return len(p), nil
+}
+
+func (e *Encryptor) Commit() error {
+	if err := e.flushBlock(); err != nil {
```
maggie-lou, line 136: Should we move flushing the final block to Write()? It's kind of confusing that Write() will not flush the last block if the bufCap is not a multiple of len(p)

i.e. If we hit the case where `readLen > len(p)-readIdx`, we won't enter the if statement `if e.bufIdx == e.bufCap` and flush the last block in Write

I could see it being confusing that all the blocks are flushed in Write() except the last one. i.e. If you're trying to debug a Write call and can't figure out why the last block keeps getting omitted

enterprise/server/crypter_service/crypter_service.go
```diff
@@ -0,0 +1,291 @@
...
+func (d *Decryptor) Read(p []byte) (n int, err error) {
+	if !d.headerValidated {
+		fileHeader := make([]byte, 3)
+		if _, err := d.r.Read(fileHeader); err != nil {
+			return 0, err
+		}
+		if string(fileHeader[0:2]) != encryptedDataHeaderSignature {
+			return 0, status.InternalErrorf("invalid file signature %d %d", fileHeader[0], fileHeader[1])
+		}
+		if fileHeader[2] != encryptedDataHeaderVersion {
```
maggie-lou, line 170: Could the version ever go into 2 digits?

enterprise/server/crypter_service/crypter_service.go
```diff
@@ -0,0 +1,291 @@
...
+	// buffer is larger to allow encryption to be done in place.
+	bufCap int
+}
+
+func fixedWidthGroupID(groupID string) []byte {
+	return []byte(fmt.Sprintf("%24s", groupID))
+}
+
+func makeChunkAuthHeader(chunkIndex uint32, groupIDHeader []byte) ([]byte, error) {
+	if 4+len(groupIDHeader) != encryptedChunkAuthHeaderSize {
```
maggie-lou, line 70: nit: Make 4 a constant

server/tables/tables.go
```diff
@@ -670,6 +670,34 @@ func (*QuotaGroup) TableName() string {
 	return "QuotaGroups"
 }
 
+type EncryptionKey struct {
+	Model
+	KeyID   string `gorm:"primaryKey"`
```
bduffany, line 675: wdyt about calling it EncryptionKeyID to be consistent with the `${TableName}ID` pk-naming pattern? e.g. `APIKeyID`

server/tables/tables.go
```diff
@@ -670,6 +670,34 @@ func (*QuotaGroup) TableName() string {
...
+	Version int    `gorm:"primaryKey"`
+
+	// BuildBuddy portion of the composite key encrypted using the master key.
+	MasterEncryptedKey []byte
+	// URI of the customer key in the external Key Management Service.
+	CustomerKeyURI string
+	// Customer portion of the composite key encrypted using the above customer
+	// key.
+	CustomerEncryptedKey []byte
```
bduffany, line 694: Should we use "Group" here instead of "Customer" to be consistent with the rest of the DB tables? or is this standard naming?

enterprise/server/crypter_service/crypter_service.go
```diff
@@ -0,0 +1,291 @@
...
+			return 0, status.InternalErrorf("invalid file version %d", fileHeader[2])
+		}
+		d.headerValidated = true
+	}
+
+	// No plaintext available, need to decrypt another chunk.
+	if d.bufIdx >= d.bufLen {
+		n, err := io.ReadFull(d.r, d.buf)
+		// We expect a short read on the last chunk.
+		if err != nil && err != io.ErrUnexpectedEOF {
```
bduffany, line 180: Instead of the short-read error check, it might be slightly clearer to use a regular `Read` and then allow `io.EOF` only if the value of `n` plus the number of bytes we've read so far matches the full stream length. That way it doesn't look like we're ignoring an unexpected error

bduffany (commented):
just a first pass, as I'm still wrapping my head around the encryption scheme

bduffany (approved):

maggie-lou (approved):

PR #3682, opened 2023-03-29
foss: move testredis and redisutil to foss

sluongng:
This PR moves `testredis` and `redisutil` packages from `//enterprise` to `//server`.

1. Our FOSS version does depend on Redis.
   Some FOSS code paths and leverage `testredis` for testing support.
   `testredis` in turn depends on `redisutil`.

2. These are none-critical packages and does not contain any "enterprise-only" features such as distributed cache or RBE.

Moving these packages under FOSS should allow us to maintain FOSS code a lot easier.

---

**Version bump**: None


sluongng:
Alternatives:
- Only move `testredis` package to foss and only move part of `redisutil`
- Move dsingleflight_test under `enterprise`

I think both are subjectively worse than the current approach.

sluongng:
This is very RFC-ish so no hurry 🤔

bduffany (changes requested):
The OSS version of the app should not depend on redis (and it is nice to have Bazel help us enforce that). So IMO we should move `server/util/dsingleflight` to enterprise/ instead of moving redis-related things out of enterprise. The 'd' in dsingleflight stands for 'distributed' and generally speaking the OSS version doesn't support a distributed setup. It's also only used by enterprise/server/image_convert, so the visibility shouldn't be an issue.

sluongng:
Yup, moving dsingleflight to //enterprise is the superior approach 👍 

Updated this PR description and commit to do that. PTAL

enterprise/server/util/dsingleflight/BUILD
```diff
@@ -3,7 +3,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 go_library(
     name = "dsingleflight",
     srcs = ["dsingleflight.go"],
-    importpath = "github.com/buildbuddy-io/buildbuddy/server/util/dsingleflight",
+    importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/util/dsingleflight",
     visibility = ["//visibility:public"],
```
bduffany, line 7: woops, I didn't catch this before but we should probably delete this line.
```suggestion
```

PR #3685, opened 2023-03-29
Render partial action result timelines (local runs only have worker and execution timestamps)

siggisim:
This allows us to render action result timelines for local actions!

---

**Version bump**: Patch


bduffany (approved):
Nice!

PR #3688, opened 2023-03-30
Fix spin loop in pebble atime update handling.

vadimberezniker:
Credit to Jim for flagging this. If the atime update channel was empty, we would spin loop on the default case and do nothing.

This changes the select to depend on `quitChan` directly. We used to need this indirection when we batched atime updates, but it's no longer necessary.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -757,15 +757,16 @@ func (p *PebbleCache) processAccessTimeUpdates(quitChan chan struct{}) error {
 			}
 		case <-quitChan:
 			// Drain any updates in the queue before exiting.
-			select {
-			case u := <-p.accesses:
-				if err := p.updateAtime(u.key); err != nil {
-					log.Warningf("Error updating atime: %s", err)
+			for {
+				select {
+				case u := <-p.accesses:
+					if err := p.updateAtime(u.key); err != nil {
+						log.Warningf("Error updating atime: %s", err)
+					}
+				default:
+					return nil
 				}
-			default:
-				break
 			}
-			return nil
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -749,29 +749,20 @@ func (p *PebbleCache) migrateData(quitChan chan struct{}) error {
...
-		default:
-			exitMu.Lock()
-			done := exiting
-			exitMu.Unlock()
-
-			if done {
-				return nil
+		case <-quitChan:
+			// Drain any updates in the queue before exiting.
+			for u := range p.accesses {
```
bduffany, line 760: does `p.accesses` ever get closed?

jdhollen (approved):
this looks good to me but i won't pretend i'm the King of Go yet or anything--I'd prefer that Brandon takes a peek, too.

enterprise/server/backends/pebble_cache/pebble_cache.go
bduffany: after a quick read through it looks like `p.accesses` doesn't get closed so I think this will hang until the server is forcibly shut down

vadimberezniker: it doesn't bleh

vadimberezniker: fixed

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -749,29 +749,23 @@ func (p *PebbleCache) migrateData(quitChan chan struct{}) error {
...
+		case <-quitChan:
+			// Drain any updates in the queue before exiting.
+			select {
+			case u := <-p.accesses:
+				if err := p.updateAtime(u.key); err != nil {
+					log.Warningf("Error updating atime: %s", err)
+				}
+			default:
+				break
 			}
```
bduffany, line 767: I think you'd need a for-loop and double-break here :P might be easier to write it in a func

bduffany (approved):

PR #3692, opened 2023-03-30
Add execution_metadata to ScoreCard

luluz66:
Execution metadata from ActionCache contains the timestamps for the original execution in the cache. This data can help us calculate how much time it saved per action hit.

### DIFF
--- a/proto/cache.proto
+++ b/proto/cache.proto
@@ -206,6 +206,9 @@ message ScoreCard {
 
     // The type of cache relevant to this result.
     resource.CacheType cache_type = 14;
+
+    // The details of the execution that originally produced this result.
+    build.bazel.remote.execution.v2.ExecutedActionMetadata execution_metadata = 15;
   }
 
   // In the interest of saving space, we only show cache misses.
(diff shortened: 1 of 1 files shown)


siggisim (approved):

proto/cache.proto
```diff
@@ -206,6 +206,10 @@ message ScoreCard {
 
     // The type of cache relevant to this result.
     resource.CacheType cache_type = 14;
+
+    // The details of the execution that originally produced this result.
+    build.bazel.remote.execution.v2.ExecutedActionMetadata execution_metadata =
```
bduffany, line 211: This data might increase the scorecard entry size by a lot. When storing, will we populate only a subset of the proto?

bduffany (approved):

proto/cache.proto
luluz66: I need the execution started timestamp and the execution completed timestamp. I can store the two timestamps instead to decrease the size.

bduffany (approved):

PR #3694, opened 2023-03-31
Implement GitHub app install+link flow

bduffany:
Implements the GitHub app installation flow:

image

The group link dialog looks like this:

image

---

**Version bump**: Patch


app/auth/auth_service.ts
```diff
@@ -183,19 +183,10 @@ export class AuthService {
...
-      await this.refreshUser();
-    } else {
-      const userCopy = new User(this.user);
-      userCopy.selectedGroup = selectedGroup;
-      this.emitUser(userCopy);
-    }
+    // Refresh the user to re-fetch the user properties associated with their
+    // selected group, such as the allowed RPCs list.
+    rpcService.requestContext.groupId = groupId;
+    await this.refreshUser();
```
siggisim, line 189: Do we not need to update the selectedGroup anymore?

siggisim (approved):

app/auth/auth_service.ts
bduffany: It gets updated by setting the group ID in request context then calling refreshUser().

The "else" branch that I deleted here had a bug that it wasn't re-fetching the allowedRPCs list for the user, which could lead to buggy UI states. So I just eliminated that path and now we always re-fetch the user.

bduffany: FWIW the only code which would have been hitting the "else" branch is the "Join org" flow, in the case where you're already a member of the org. The new logic still works in that case (there's just a tiny delay now when you click "View builds" since the user has to be re-fetched).

PR #3695, opened 2023-03-31
integration/remote_execution: fix darwin tests

sluongng:
Fix a set of previously skipped tests because build actions were missing
platform properties to select the executor created and registered
locally.

---

**Version bump**: None


bduffany (approved):
awesome, thanks :)

PR #3697, opened 2023-03-31
Implement GitHub app installation management

bduffany:
- Shows all installations that your linked GitHub account has access to, as well as all installations linked to your org (even if your linked GitHub account doesn't have access to the installation)
  - If not yet linked to BB, show a "Link" button that calls LinkGitHubAppInstallation
  - If linked to BB (any org, even ones you aren't a member of), show an unlink button
    - If not linked to the currently selected org, show a warning

In a separate PR, I might add convenience links for managing each individual installation (via GitHub), but for now that can be done via the "Setup" button then selecting the installation (it's just an extra click)

https://user-images.githubusercontent.com/2414826/229212245-386b24e6-261d-4e99-bf47-89ad632c1272.mp4

---

**Version bump**: Patch

### DIFF
--- a/app/root/root.css
+++ b/app/root/root.css
@@ -1994,11 +1994,14 @@ code .comment {
 
 .loading {
   width: 100%;
-  height: 100%;
   display: flex;
   align-items: center;
   justify-content: center;
   flex-grow: 1;
+}
+
+.loading:not(.loading-slim) {
+  height: 100%;
 }
 
 .loading:before {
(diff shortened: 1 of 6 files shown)


siggisim:
Also, what are your thoughts on hiding the old link UI and warning message unless you already have an old account linked? 

That way it can only be used for unlinking (and not for linking new accounts).

bduffany:
> what are your thoughts on only showing repos that are linked to you current org instead? 

I implemented this because I thought people might get confused about why an installation isn't showing up in the org even though they have linked it to BB. But it would simplify things a lot to only show repos linked to the current org, so I've removed this for now and we can always change it later if people think it's confusing.

>  Also, what are your thoughts on hiding the old link UI and warning message unless you already have an old account linked?

I had this same thought earlier today - will do that in a follow-up PR

PR #3700, opened 2023-04-01
cli/version: simplified w. embedsrcs

sluongng:
I was trying to figure out how we are installing bb CLI in Github Action lint check
when I stumbled upon this code.

Added a test for it as well.

---

**Version bump**: Patch


cli/version/BUILD
```diff
@@ -1,36 +1,32 @@
...
-
-var cliVersionFlag = "%s"
-""",
-)
-
-go_library(
-    name = "version_lib",
-    srcs = [":version_flag.go"],
-    importpath = "github.com/buildbuddy-io/buildbuddy/cli/version",
+    template = "%s",
```
bduffany, line 13: nit: you can omit `template` if the flag is the only content being written

```suggestion
```

bduffany (approved):
nice

PR #3703, opened 2023-04-03
store ExecutionActionMetadata in ScoreCard

luluz66:


server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -159,12 +159,13 @@ func (s *ActionCacheServer) GetActionResult(ctx context.Context, req *repb.GetAc
 		ht.TrackMiss(d)
 		return nil, status.NotFoundErrorf("ActionResult (%s) not found: %s", d, err)
 	}
-	defer downloadTracker.CloseWithBytesTransferred(int64(len(blob)), int64(len(blob)), repb.Compressor_IDENTITY, "ac_server")
```
bduffany, line 162: I think it'd be simpler/cleaner to set the executedAction metadata as a public field (or public method like `SetExecutedActionMetadata`), e.g. `downloadTracker.ExecutedActionMetadata = rsp.GetExecutionMetadata()` - that way we don't need the two separate calls to `Close...()` (one deferred and one non-deferred)

luluz66: Agreed and done!

bduffany (approved):
nice :)

PR #3704, opened 2023-04-03
Give soci-store a second to start up before invoking it via filesystem

iain-macdonald:
**Version bump**: Patch


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -121,6 +129,22 @@ additionallayerstores=["/var/lib/soci-store/store:ref"]
 	}, nil
 }
 
+// Wait (sleep) for path to exist for up to timeMax in timeInc increments.
+func waitForFile(path string, timeInc, timeMax time.Duration) error {
```
bduffany, line 133: this func is available in `server/util/disk`:

`disk.WaitUntilExists(env.GetServerContext(), "/var/lib/soci-store/store", disk.WaitOpts{PollInterval: 100*time.Millisecond, Timeout: 1*time.Second})` here

iain-macdonald: Ah, should have seen that. Thanks!

bduffany (approved):

PR #3705, opened 2023-04-03
Incorporate encryption support into pebble cache.

vadimberezniker:
We store a `bool` property on the group to indicate whether a customer has encryption enabled. This property is looked up and cached as part of the normal auth check so that we can avoid doing any additional database lookups to determine if auth is enabled.

Cache partitions have an additional property to indicate whether encrypted content can be stored on that partition. We allow partitions (e.g. the default one) to share CAS data across customers which is incompatible with encryption. We will require dedicated customer partitions before encryption can be enabled.

---

**Version bump**: None

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1159,6 +1159,34 @@ func (p *PebbleCache) lookupGroupAndPartitionID(ctx context.Context, remoteInsta
 	return groupID, DefaultPartitionID, nil
 }
 
+func (p *PebbleCache) encryptionEnabled(ctx context.Context, partitionID string) (bool, error) {
+	auth := p.env.GetAuthenticator()
+	if auth == nil {
+		return false, nil
+	}
+	u, err := auth.AuthenticatedUser(ctx)
+	if err != nil {
+		return false, nil
+	}
+	if !u.GetCacheEncryptionEnabled() {
+		return false, nil
+	}
+	if p.env.GetCrypter() == nil {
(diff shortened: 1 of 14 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1159,6 +1159,34 @@ func (p *PebbleCache) lookupGroupAndPartitionID(ctx context.Context, remoteInsta
...
+	}
+	u, err := auth.AuthenticatedUser(ctx)
+	if err != nil {
+		return false, nil
+	}
+	if !u.GetCacheEncryptionEnabled() {
+		return false, nil
+	}
+	if p.env.GetCrypter() == nil {
+		return false, status.FailedPreconditionError("encryption requested by crypter not available")
```
maggie-lou, line 1175: nit: "but"

enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/kms"
```
maggie-lou, line 7: nit: Move dependencies

server/testutil/testdigest/testdigest.go
```diff
@@ -1,6 +1,7 @@
 package testdigest
 
 import (
+	"github.com/buildbuddy-io/buildbuddy/proto/resource"
```
maggie-lou, line 4: nit: Move import

bduffany (approved):

PR #3708, opened 2023-04-04
TypeScript: remove baseUrl to force relative imports

bduffany:
---

**Version bump**: Patch

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,7 +12,6 @@
     "skipLibCheck": true,
     "composite": true,
     "target": "ES2016",
-    "baseUrl": ".",
     "lib": ["es2017", "es2019", "dom", "dom.iterable"],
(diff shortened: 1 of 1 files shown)


siggisim (approved):

tsconfig.json
```diff
@@ -24,6 +23,8 @@
       "bazel-out/darwin_arm64-opt/bin",
       "bazel-out/darwin_arm64-fastbuild/bin"
     ]
+    // Note: we intentionally don't set "baseUrl": "." in order to force import
+    // module specifiers to be relative.
```
bduffany, line 27: Added this little note because I think I might've been the one who introduced `"baseUrl"` in the first place, thinking it was harmless - don't want others to repeat the same mistake :sweat_smile:

PR #3717, opened 2023-04-05
Add Writer interface to blobstore

tempoz:
Add interface to blobstore to get an `io.Writer` so we can stream to blobstore.

---

**Version bump**: Minor

### DIFF
--- a/server/backends/blobstore/aws/aws.go
+++ b/server/backends/blobstore/aws/aws.go
@@ -284,10 +284,15 @@ func (a *AwsS3BlobStore) Writer(ctx context.Context, blobName string) (interface
 	zw := util.NewCompressWriter(pw)
 	cwc := ioutil.NewCustomCommitWriteCloser(zw)
 	cwc.CommitFn = func(int64) error {
-		var err error
 		if compresserCloseErr := zw.Close(); compresserCloseErr != nil {
-			err = compresserCloseErr
+			cancel() // Don't try to finish the commit op if Close() failed.
+			if pipeCloseErr := pw.Close(); pipeCloseErr != nil {
+				log.Errorf("Error closing the pipe for %s: %s", blobName, pipeCloseErr)
+			}
+			<-errch // Canceling the context makes any error here meaningless.
+			return compresserCloseErr
 		}
+		var err error
 		if writerCloseErr := pw.Close(); writerCloseErr != nil {
 			err = writerCloseErr
 		}
--- a/server/backends/blobstore/aws/aws.go
+++ b/server/backends/blobstore/aws/aws.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"flag"
+	"io"
 	"strings"
 	"time"
 
@@ -253,4 +254,24 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
 	}
 
 	return true, nil
+}
+
+func (a *AwsS3BlobStore) Writer(ctx context.Context, blobName string) (io.WriteCloser, error) {
+	// Open a pipe.
+	pr, pw := io.Pipe()
+
+	errch := make(chan error)
+
+	// Upload from pr in a separate Go routine.
+	go func() {
+		_, err := a.uploader.Upload(&s3manager.UploadInput{
+			Bucket: a.bucket,
+			Key:    aws.String(blobName),
+			Body:   pr,
+		})
+		errch <- err
+		close(errch)
+	}()
+
+	return util.NewCompressedBlobStoreWriter(ctx, pw, &util.AsyncCloser{Closer: pw, ErrorChannel: errch}, awsS3Label), nil
 }
--- a/server/backends/blobstore/aws/aws.go
+++ b/server/backends/blobstore/aws/aws.go
@@ -289,17 +289,14 @@ func (a *AwsS3BlobStore) Writer(ctx context.Context, blobName string) (interface
 			if pipeCloseErr := pw.Close(); pipeCloseErr != nil {
 				log.Errorf("Error closing the pipe for %s: %s", blobName, pipeCloseErr)
 			}
-			<-errch // Canceling the context makes any error here meaningless.
+			// Canceling the context makes any error in errch meaningless; don't
+			// bother to read it.
 			return compresserCloseErr
 		}
-		var err error
 		if writerCloseErr := pw.Close(); writerCloseErr != nil {
-			err = writerCloseErr
+			return writerCloseErr
 		}
-		if writeErr := <-errch; writeErr != nil {
-			err = writeErr
-		}
-		return err
+		return <-errch
 	}
 	cwc.CloseFn = func() error {
 		cancel()
--- a/server/backends/blobstore/aws/aws.go
+++ b/server/backends/blobstore/aws/aws.go
@@ -19,6 +19,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/backends/blobstore/util"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/ioutil"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
@@ -280,14 +281,24 @@ func (a *AwsS3BlobStore) Writer(ctx context.Context, blobName string) (interface
 		close(errch)
 	}()
 
-	committer := &util.CustomCommitter{CommitOp: func() error {
-		return (&util.AsyncCloser{Closer: pw, ErrorChannel: errch}).Close()
-	}}
-
-	closer := &util.CustomCloser{CloseOp: func() error {
+	zw := util.NewCompressWriter(pw)
+	cwc := ioutil.NewCustomCommitWriteCloser(zw)
+	cwc.CommitFn = func(int64) error {
+		var err error
+		if compresserCloseErr := zw.Close(); compresserCloseErr != nil {
+			err = compresserCloseErr
+		}
+		if writerCloseErr := pw.Close(); writerCloseErr != nil {
+			err = writerCloseErr
+		}
+		if writeErr := <-errch; writeErr != nil {
+			err = writeErr
+		}
+		return err
+	}
+	cwc.CloseFn = func() error {
 		cancel()
 		return nil
-	}}
-
-	return util.NewCompressedBlobStoreWriter(ctx, committer, pw, closer, awsS3Label), nil
+	}
+	return cwc, nil
 }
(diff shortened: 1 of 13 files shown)


server/backends/blobstore/util/util.go
```diff
@@ -121,3 +123,54 @@ func RecordDeleteMetrics(typeLabel string, startTime time.Time, err error) {
...
+func NewCompressedBlobStoreWriter(ctx context.Context, w io.Writer, c io.Closer, label string) io.WriteCloser {
+	return &compressedWriter{ctx: ctx, gzipWriter: gzip.NewWriter(w), closer: c, label: label, n: 0}
+}
+
+func (d *compressedWriter) Write(p []byte) (int, error) {
+	start := time.Now()
+	_, spn := tracing.StartSpan(d.ctx)
+	n, err := d.gzipWriter.Write(p)
+	spn.End()
+	RecordWriteMetrics(d.label, start, d.n, err)
```
bduffany, line 144: we might want to omit these Write metrics for now -- the metrics could get skewed in the case where the gzipWriter is doing internal buffering/async flushing (and the metrics would show that the Write() completes in nanoseconds, even the bytes have just been moved from one in-memory buffer to another).

Another option is to do it in `Close()` but that is also not ideal because it depends on client upload latency+throughput (while the existing usages of this metric only depend on the server -> blobstore latency+throughput).

server/testutil/mockstore/mockstore.go
```diff
@@ -73,9 +75,32 @@ func (m *Mockstore) WriteBlob(_ context.Context, blobName string, data []byte) (
...
+	buf      bytes.Buffer
+	m        *Mockstore
+	blobName string
+}
+
+func (w *WriteCloser) Write(p []byte) (int, error) {
+	w.m.mu.Lock()
+	defer w.m.mu.Unlock()
+	n, err := w.buf.Write(p)
+	w.m.BlobMap[w.blobName] = w.buf.Bytes()
```
bduffany, line 100: nit: if this is updated to be a CommitWriteCloser then this line would probably fit better in Commit()

server/backends/blobstore/aws/aws.go
```diff
@@ -254,3 +255,23 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
...
 }
+
+func (a *AwsS3BlobStore) Writer(ctx context.Context, blobName string) (io.WriteCloser, error) {
+	// Open a pipe.
+	pr, pw := io.Pipe()
+
+	errch := make(chan error)
+
+	// Upload from pr in a separate Go routine.
+	go func() {
```
bduffany, line 266: In these async goroutine funcs, we should begin the goroutine with `defer pr.Close()` as the first line, so that if the async blobstore upload fails, any subsequent writes to the returned CompressedBlobstoreWriter would fail (due to the pipe being closed), rather than getting stuck.

Example: https://go.dev/play/p/2TQ4JP8y10w

bduffany (changes requested):

server/backends/blobstore/aws/aws.go
```diff
@@ -254,3 +255,23 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
...
+		_, err := a.uploader.Upload(&s3manager.UploadInput{
+			Bucket: a.bucket,
+			Key:    aws.String(blobName),
```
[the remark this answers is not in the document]
sluongng: I sent out https://github.com/buildbuddy-io/buildbuddy/pull/3736

@tempoz I intend to merge it only after your series lands. Let me know if you prefer to have it merged first.

server/testutil/mockstore/mockstore.go
tempoz: Done!

server/backends/blobstore/util/util.go
```diff
@@ -121,3 +123,54 @@ func RecordDeleteMetrics(typeLabel string, startTime time.Time, err error) {
 		metrics.BlobstoreTypeLabel: typeLabel,
 	}).Observe(float64(duration.Microseconds()))
 }
+
+type compressedWriter struct {
+	ctx        context.Context
+	gzipWriter *gzip.Writer
```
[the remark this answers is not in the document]
bduffany: Ah gotcha, I forgot we were already using gzip to store things in blobstore.

server/backends/blobstore/util/util.go
```diff
@@ -121,3 +127,82 @@ func RecordDeleteMetrics(typeLabel string, startTime time.Time, err error) {
...
+		ctx: ctx, committer: committer,
+		zw:     newCompressWriter(writer),
+		closer: closer,
+		label:  label,
+		n:      0,
+	}
+}
+
+func (c *compressedWriter) Commit() error {
+	c.zw.Close()
```
bduffany, line 157: if `c.zw.Close()` fails, should we assume something went wrong and abort - i.e. return the error instead of calling `committer.Commit()` ?

server/backends/blobstore/aws/aws.go
```diff
@@ -254,3 +256,38 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
...
+				Key:    aws.String(util.BlobPath(blobName)),
+				Body:   pr,
+			},
+		)
+		errch <- err
+		close(errch)
+	}()
+
+	committer := &util.CustomCommitter{CommitOp: func() error {
+		return (&util.AsyncCloser{Closer: pw, ErrorChannel: errch}).Close()
```
bduffany, line 284: Since the `AsyncCloser` is only being used in these `committer` funcs now, and the `Close()` logic is relatively short (*), wdyt about inlining the logic here and removing the AsyncCloser class?

(*) the `Close()` logic could also be a little shorter:

```
	if err := pw.Close(); err != nil {
    // This next if {} block can be removed
    // since errch is buffered
		if err := <-a.ErrorChannel; err != nil {
			return err
		}
		return err
	}
	return <-errch
```

server/backends/blobstore/aws/aws.go
```diff
@@ -254,3 +256,38 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
...
+				Bucket: a.bucket,
+				Key:    aws.String(util.BlobPath(blobName)),
+				Body:   pr,
+			},
+		)
+		errch <- err
+		close(errch)
+	}()
+
+	committer := &util.CustomCommitter{CommitOp: func() error {
```
vadimberezniker, line 283: This functionality is already available via `ioutil.NewCustomCommitWriteCloser`. 
You could have the compressed writer be responsible only for the compression and then wrap it here with additional commit/close funcs. 
```	
    cwc := ioutil.NewCustomCommitWriteCloser(util.NewCompressedBlobStoreWriter(ctx, pw, awsS3Label))
	cwc.CommitFn = func(i int64) error {
		// ...
	}
	cwc.CloseFn = func() error {
		// ...
	}
```

server/backends/blobstore/util/util.go
```diff
@@ -121,3 +127,82 @@ func RecordDeleteMetrics(typeLabel string, startTime time.Time, err error) {
...
+	return c.committer.Commit()
+}
+
+func (c *compressedWriter) Write(p []byte) (int, error) {
+	//Unnecessary without the RecordWriteMetrics call
+	// start := time.Now()
+	_, spn := tracing.StartSpan(c.ctx)
```
vadimberezniker, line 164: This is too fine grained for a span. Normally a span represents a slightly higher level operation (i.e. "write blob") and individual writes might be represented as span events.

server/backends/blobstore/util/util.go
tempoz: Good point, done.

server/backends/blobstore/aws/aws.go
```diff
@@ -254,3 +257,53 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
...
+			cancel() // Don't try to finish the commit op if Close() failed.
+			if pipeCloseErr := pw.Close(); pipeCloseErr != nil {
+				log.Errorf("Error closing the pipe for %s: %s", blobName, pipeCloseErr)
+			}
+			<-errch // Canceling the context makes any error here meaningless.
+			return compresserCloseErr
+		}
+		var err error
+		if writerCloseErr := pw.Close(); writerCloseErr != nil {
+			err = writerCloseErr
```
vadimberezniker, line 297: You can return the error directly instead of storing it. Since errch is buffered, there's no harm if you return before reading from the channel.

server/backends/blobstore/aws/aws.go
```diff
@@ -254,3 +257,53 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
...
+		}
+		if writeErr := <-errch; writeErr != nil {
+			err = writeErr
```
vadimberezniker, line 300: Same here.

server/backends/blobstore/gcs/gcs.go
```diff
@@ -134,3 +136,24 @@ func (g *GCSBlobStore) BlobExists(ctx context.Context, blobName string) (bool, e
...
+func (g *GCSBlobStore) Writer(ctx context.Context, blobName string) (interfaces.CommittedWriteCloser, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	bw := g.bucketHandle.Object(util.BlobPath(blobName)).NewWriter(ctx)
+
+	zw := util.NewCompressWriter(bw)
+	cwc := ioutil.NewCustomCommitWriteCloser(zw)
+	cwc.CommitFn = func(int64) error {
+		if compresserCloseErr := zw.Close(); compresserCloseErr != nil {
+			cancel()       // Don't try to finish the commit op if Close() failed.
+			_ = bw.Close() // Canceling the context makes any error here meaningless.
```
vadimberezniker, line 149: It's not 100% clear from the GCS client docs, but cancelling the context automatically closes the writer so it is likely safer not to try to call `Close` as well since there might be a race between the context cancellation and the close that could lead to the write succeeding.

server/backends/blobstore/util/util.go
vadimberezniker: If you do want to bring it back in the future, the way to do it is create a new span in `Writer` and then call `span.AddEvent` for each individual write. This allows you to add fine grained timing information w/o creating new spans.

vadimberezniker (approved):

server/backends/blobstore/aws/aws.go
```diff
@@ -254,3 +257,53 @@ func (a *AwsS3BlobStore) BlobExists(ctx context.Context, blobName string) (bool,
...
+
+	zw := util.NewCompressWriter(pw)
+	cwc := ioutil.NewCustomCommitWriteCloser(zw)
+	cwc.CommitFn = func(int64) error {
+		if compresserCloseErr := zw.Close(); compresserCloseErr != nil {
+			cancel() // Don't try to finish the commit op if Close() failed.
+			if pipeCloseErr := pw.Close(); pipeCloseErr != nil {
+				log.Errorf("Error closing the pipe for %s: %s", blobName, pipeCloseErr)
+			}
+			<-errch // Canceling the context makes any error here meaningless.
```
bduffany, line 292: nit: this channel receive can be deleted since the channel is buffered. it'll just collect the write error once line 280 is hit, then get garbage collected.

bduffany (approved):

server/backends/blobstore/aws/aws.go
tempoz: Done.

server/backends/blobstore/gcs/gcs.go
tempoz: Thanks, didn't realize; fixed!

server/backends/blobstore/aws/aws.go
tempoz: Thanks, didn't know; done.

PR #3722, opened 2023-04-05
Upgrade gorm to support type aliases.

vadimberezniker:
This will allow us to use `bool` as a schema type instead of using the mysql-specific tinyint.

---

**Version bump**: Patch


tempoz (approved):
Super excited for this!

luluz66 (approved):

PR #3723, opened 2023-04-05
Implement GitHub app repo linking

bduffany:
---

**Version bump**: Patch


enterprise/app/workflows/github_app_import.tsx
```diff
@@ -0,0 +1,303 @@
...
+    document.title = "Link GitHub repo | BuildBuddy";
+    this.fetch();
+  }
+
+  componentDidUpdate(prevProps: GitHubAppImportProps) {
+    if (this.props.user !== prevProps.user) this.fetch();
+  }
+
+  private octokit() {
+    return new Octokit({ auth: this.props.user.githubToken });
```
siggisim, line 81: What do we use Octokit for here?

siggisim (approved):

enterprise/app/workflows/github_app_import.tsx
bduffany: It's to fetch installations accessible to the user - I wasn't sure whether to show the installations accessible to the user, or the installations linked to the org, since the user might not have access to some of those installations. But thinking about it more, it's probably better to show only the installations linked to the org and just let the UI show an error state if the user doesn't have access to the installation. I can update the PR to use that approach instead

bduffany: Done (removed octokit in favor of fetching org-linked app installations)

PR #3724, opened 2023-04-05
Add SociArtifactStore proto

iain-macdonald:
This adds a new RPC service that has one request: Read which returns pointers to all of the SOCI artifacts required to stream the requested image. These pointers are pointers to data in the CAS, which the client will read from if it needs the data.

**Version bump**:  Minor

### DIFF
--- a/proto/soci.proto
+++ b/proto/soci.proto
@@ -5,42 +5,41 @@ import "proto/remote_execution.proto";
 package soci;
 
 enum Type {
-    UNKNOWN_TYPE = 0;
+  UNKNOWN_TYPE = 0;
 
-    // The SOCI Index is a JSON file containing indexing information for some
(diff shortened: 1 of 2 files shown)


proto/soci.proto
```diff
@@ -0,0 +1,40 @@
...
+message ReadRequest {
+  string image_name = 1;
+}
+
+message ReadResponse {
+  repeated Artifact = 1;
+}
+
+service SociArtifactStore {
+  rpc Read(ReadRequest) returns (ReadResponse);
```
vadimberezniker, line 39: since this API is no longer returning any data directly, you might consider renaming this to something like `GetArtifacts`

vadimberezniker (approved):

proto/soci.proto
iain-macdonald: Good call, updated

PR #3725, opened 2023-04-05
Don't fail the release assets script if there are assets missing

maggie-lou:
Failing should be reserved if something actually breaks

### DIFF
--- a/tools/check_release_assets_uploaded.py
+++ b/tools/check_release_assets_uploaded.py
@@ -80,7 +80,6 @@ def main():
 
     if len(expected_assets) > 0:
         print(f"Missing assets {expected_assets}", file=sys.stderr)
-        exit(1)
 
 if __name__ == "__main__":
     main()


maggie-lou:
* master
  * **PR #3725** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3725?utm_source=stack-comment).

iain-macdonald (changes requested):
What's the issue if it fails? Are there cases where it's okay for the assets to be missing?

maggie-lou:
> > What's the issue if it fails? Are there cases where it's okay for the assets to be missing?
> 
> +1, if the intention is to ignore the error and maybe send the stderr to slack, maybe it's clearer to explicitly handle the failing exit code in bash?

This week in the release, the script failed because the version tag interpolation failed. However I had script || true in the workflow, so it wasn't registered as a failure. We need a way to detect a difference between the script failing because of something unexpected vs the script detecting missing assets where we want to capture stderr and send to slack

bduffany (approved):

PR #3726, opened 2023-04-06
Implement SociArtifactStore RPC Service

iain-macdonald:
This PR implements the RPC Service added in https://github.com/buildbuddy-io/buildbuddy/pull/3724 by adding a new RPC endpoint that serves pointers to soci artifacts, generating them on demand when requested.

**Version bump**: Minor


enterprise/server/soci/soci.go
```diff
@@ -0,0 +1,316 @@
...
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/commandutil"
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
```
vadimberezniker, line 12: nit: move the aliased imports below into a separate group

enterprise/server/soci/soci.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return err
+	}
+	env.SetSociArtifactStoreServer(server)
+	return nil
+}
+
+func NewSociArtifactStore(env environment.Env) (error, *SociArtifactStore) {
+	if env.GetCache() == nil {
+		return status.InternalError("soci artifact server requires a cache"), nil
```
vadimberezniker, line 39: nit: FailedPreconditionError (same below)

enterprise/server/soci/soci.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+type ImageInfo struct {
+	Digest string `json:"Digest"`
+}
+
+// The image id is the digest of the actual image and changes as image tags are
+// altered. Grab it by running `skopeo inspect`.
+func imageId(ctx context.Context, imageName string) (string, error) {
+	cmd := []string{"skopeo", "inspect", "docker://" + imageName}
```
vadimberezniker, line 85: Do we need `skopeo` in the mix? Container Registry APIs are relatively simple. Take a look at `registry.go` which has code for getting image information.

enterprise/server/soci/soci.go
```diff
@@ -0,0 +1,316 @@
...
+	log.Infof("Indexing image %s took %s", imageName, time.Since(start))
+	if res.Error != nil {
+		return nil, nil, res.Error
+	}
+	return parseSociOutput(string(res.Stdout))
+}
+
+// Parses the command output described above into a digest of the soci index
+// and a list of digests of the ztocs and returns those.
+func parseSociOutput(output string) (*repb.Digest, []*repb.Digest, error) {
```
vadimberezniker, line 212: Having to parse the soci output is a bit unfortunate. The "create command" looks [relatively simple](https://github.com/awslabs/soci-snapshotter/blob/main/cmd/soci/commands/create.go). Have you considered using the soci APIs (same ones the create command is using) to this indexing?

enterprise/server/soci/soci.go
```diff
@@ -0,0 +1,316 @@
+package soci
```
vadimberezniker, line 1: nit: can we call this `socistore` or something like that? this package is implementing a specific API with a very specific surface and i don't think the current name conveys that.

enterprise/server/soci/soci.go
iain-macdonald: Done

enterprise/server/soci/soci.go
iain-macdonald: Done

enterprise/server/soci/soci.go
iain-macdonald: Nope, looks like `registry.go` does the trick too, thanks for the recommendation.

enterprise/server/soci/soci.go
iain-macdonald: Changed to `sociartifactstore` -- I didn't want to use `socistore` because that's the name of the binary that implements the fs that podman reads from. Open to other suggestions though.

enterprise/server/soci/soci.go
iain-macdonald: Yeah good call, changed to read from the artifact db here as discussed and added a TODO to get rid of this entirely and just call the `create()` code directly.

iain-macdonald:
Thanks Vadim, PTAL

buildpatches/gazelle.patch
```diff
@@ -4,7 +4,7 @@ index 2ff57db..7d0c6be 100644
 +++ b/cmd/gazelle/gazelle.go
```
vadimberezniker, line 4: needs to be reverted?

cli/fix/langs.go
```diff
@@ -18,7 +18,7 @@ package main
 import (
 	"github.com/bazelbuild/bazel-gazelle/language"
 	"github.com/bazelbuild/bazel-gazelle/language/bazel/visibility"
-	"github.com/bazelbuild/bazel-gazelle/language/go"
+	golang "github.com/bazelbuild/bazel-gazelle/language/go"
```
vadimberezniker, line 21: nit: move alias import to a separate import section below

enterprise/server/registry/registry.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/prefix"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
vadimberezniker, line 25: why?

vadimberezniker (approved):

buildpatches/gazelle.patch
iain-macdonald: Whoops, done

cli/fix/langs.go
iain-macdonald: Reverted, this was from messing around with the dependency upgrading stuff

PR #3735, opened 2023-04-07
If dockerUser is set to UID:GID, don't create user/group.

vadimberezniker:
This mimics the behavior of docker.

---

**Version bump**: Patch

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -201,6 +201,27 @@ func NewPodmanCommandContainer(env environment.Env, imageCacheAuth *container.Im
 	}
 }
 
+func addUserArgs(args []string, options *PodmanOptions) []string {
+	if options.ForceRoot {
+		args = append(args, "--user=0:0")
+	} else if options.User != "" && userRegex.MatchString(options.User) {
+		args = append(args, "--user="+options.User)
+		parts := strings.Split(options.User, ":")
+		// If the user is set to UID:GUID, tell podman not to add entries
+		// to /etc/passwd and /etc/group which mimics the docker behavior.
+		if len(parts) == 2 {
+			_, err := strconv.Atoi(parts[0])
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -201,6 +201,27 @@ func NewPodmanCommandContainer(env environment.Env, imageCacheAuth *container.Im
...
 
+func addUserArgs(args []string, options *PodmanOptions) []string {
+	if options.ForceRoot {
+		args = append(args, "--user=0:0")
+	} else if options.User != "" && userRegex.MatchString(options.User) {
+		args = append(args, "--user="+options.User)
+		parts := strings.Split(options.User, ":")
+		// If the user is set to UID:GUID, tell podman not to add entries
+		// to /etc/passwd and /etc/group which mimics the docker behavior.
+		if len(parts) == 2 {
```
bduffany, line 212: Should we also handle the case where only the user ID (`--user=509`) or only the group ID (`--user=:509`) is set? (Probably not as common but would be nice to be compatible with docker in that case too)

e.g.

```
$ podman run --user=509 --rm -it ubuntu:20.04 sh -c 'id; cat /etc/group | grep 509'
uid=509(509) gid=0(root) groups=0(root)
509:x:509:509

$ docker run --user=509 --rm -it ubuntu:20.04 sh -c 'id; cat /etc/group | grep 509'
uid=509 gid=0(root) groups=0(root)
```

bduffany (approved):

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: (If it complicates the code too much I'm fine with just submitting this for now)

vadimberezniker: I don't think it's important right now. I think it's unlikely someone is going to run into this give how uncommon this usage is.

PR #3736, opened 2023-04-10
blobstore/util: wrap all blobstore impl with prefix

sluongng:
Introduce PrefixBlobstore to wrap arround all existing blobstore
implementations to provide prefix support out-of-the-box.  This should
help us avoid the overhead of having to call util.BlobPath explicitly in
each blobstore implementation.

until.BlobPath is kept exported so that we could mock it during testing.

This may conflict with #3717 and require a merge conflict resolve if
this is merged after.
In which case, author could re-use this comby command for refactoring.

```bash
comby 'util.BlobPath(:[1])' ':[1]' server/backends/blobstore/**/*.go -exclude server/backends/blobstore/util/
```

---

**Version bump**: Patch

### DIFF
--- a/server/backends/blobstore/util/util.go
+++ b/server/backends/blobstore/util/util.go
@@ -3,21 +3,21 @@ package util
 import (
 	"bytes"
 	"compress/gzip"
+	"context"
 	"flag"
 	"fmt"
 	"io"
 	"path/filepath"
 	"time"
 
+	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 
 	gstatus "google.golang.org/grpc/status"
 )
 
-var (
-	pathPrefix = flag.String("storage.path_prefix", "", "The prefix directory to store all blobs in")
-)
+var pathPrefix = flag.String("storage.path_prefix", "", "The prefix directory to store all blobs in")
 
 func NewCompressWriter(w io.Writer) io.WriteCloser {
 	return gzip.NewWriter(w)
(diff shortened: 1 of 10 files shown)


server/backends/blobstore/util/util.go
```diff
@@ -67,6 +69,30 @@ var BlobPath = func(blobName string) string {
 	return filepath.Join(*pathPrefix, blobName)
 }
 
+type prefixBlobstore struct {
+	interfaces.Blobstore
+}
+
+func NewPrefixBlobstore(b interfaces.Blobstore) *prefixBlobstore {
```
bduffany, line 76: nit: it might be worth having a method `WithPrefix(b)` that just returns `b` if prefix is empty, else returns `NewPrefixBlobstore(b)`. That way we avoid the `prefixBlobstore` overhead for the common case of not setting a prefix.

bduffany (approved):
this LGTM if you want to pull in Zoey's `Writer()` method and merge :+1:

sluongng:
I made some updates to this PR as I rebased on top of https://github.com/buildbuddy-io/buildbuddy/pull/3717

Added test cases for the `Writer()` method.

server/backends/blobstore/util/util.go
```diff
@@ -71,8 +71,47 @@ func Compress(in []byte) ([]byte, error) {
...
+// blobstore and prefixes all blob names with the given prefix.
+// Intent to support testing only, you probably want to use
+// NewDefaultPrefixBlobstore instead.
+func NewPrefixBlobstore(b interfaces.Blobstore, prefix string) *prefixBlobstore {
+	return &prefixBlobstore{b, prefix}
+}
+
+func (p *prefixBlobstore) blobPath(blobName string) string {
+	return filepath.Join(p.prefix, blobName)
+}
```
bduffany, line 95: since this func exists I think we can delete `util.BlobPath` now? Do you want to do that in this PR?

bduffany (approved):

server/backends/blobstore/util/util.go
sluongng: I did remove util.BlobPath in this PR 🤔 , do you mean removing `prefixBlobstore.blobPath()` in favor of using `filepath.Join()` directly?

bduffany: woops, ignore me - I missed the diff :sweat_smile:

PR #3740, opened 2023-04-10
Add support for rendering `coverage` runs

siggisim:
For `bazel coverage` runs, we'll now show this file level breakdown on the test target page:
Screenshot 2023-04-10 at 3 19 20 PM

If using github, and build metadata is set - clicking on these files will show you a test coverage overlay:
Screenshot 2023-04-10 at 3 19 34 PM

---

**Version bump**: Patch

### DIFF
--- a/enterprise/app/code/code.tsx
+++ b/enterprise/app/code/code.tsx
@@ -23,6 +23,7 @@ import Dialog, {
   DialogTitle,
 } from "../../../app/components/dialog/dialog";
 import Modal from "../../../app/components/modal/modal";
+import { parseLcov } from "../../../app/util/lcov";
 
 interface Props {
   user: User;
@@ -116,7 +117,7 @@ export default class CodeComponent extends React.Component<Props, State> {
   }
 
   isSingleFile() {
-    return Boolean(this.props.search.get("bytestream_url"));
+    return Boolean(this.props.search.get("bytestream_url")) || Boolean(this.props.search.get("lcov"));
   }
 
   componentDidMount() {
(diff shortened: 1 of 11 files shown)


enterprise/app/code/code.tsx
```diff
@@ -135,14 +136,68 @@ export default class CodeComponent extends React.Component<Props, State> {
...
 
+    let lcovURL = this.props.search.get("lcov");
+    let commit = this.props.search.get("commit");
+    if (this.isSingleFile() && lcovURL) {
+      this.octokit.rest.repos
+        .getContent({
+          owner: this.currentOwner(),
+          repo: this.currentRepo(),
+          path: this.currentPath(),
+          ref: commit || this.state.commitSHA || "master",
```
jdhollen, line 154: Am I misreading or does this get messed up if the user has added / removed lines from the file?

jdhollen: Sorry, I meant to say, if the user has added / removed lines *and not pushed to github yet*

siggisim: That's right, if it's not a CI run - but a local run and the changes haven't been published to Github, the source file might be out of date. I'm not sure there's much we can do in that scenario unless Bazel sends us the source file (which it likely does in the remote execution case, so that's an optimization we could make later).

jdhollen: I don't really care about people who are hacking the URL up, but you could check that the revision actually exists on the coverage card and not link out to the code view if the file wasn't found?

siggisim: I don't really want to do any requests to Github from outside of the code UI. 

You cool with seeing if this becomes a common issue, and if it is we can come up with something fancier?

siggisim: Slash, will you let me get this into tomorrow's release and I promise to make improvements in the future :)

jdhollen (approved):

PR #3742, opened 2023-04-11
gazelle: fix missing file warning

sluongng:
When a file referenced in embedsrcs is a generated file, Gazelle will
try it best to look up neighbor rules within the same package for "out"
and "outs" attributes that declare such file.

However, our write_flag_to_file uses "name" attribute to specify the
file name, thus trip up Gazelle at runtime.

Provide an optional "out" attribute in write_flag_to_file to avoid
Gazelle warning.

---

**Version bump**: Patch

### DIFF
--- a/rules/flags/index.bzl
+++ b/rules/flags/index.bzl
@@ -1,7 +1,8 @@
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 
 def _write_flag_to_file_impl(ctx):
-    out = ctx.actions.declare_file(ctx.attr.name)
+    out_file = ctx.attr.out if ctx.attr.out else ctx.attr.name
+    out = ctx.actions.declare_file(out_file)
     ctx.actions.write(out, ctx.attr.template % ctx.attr.flag[BuildSettingInfo].value)
     return DefaultInfo(files = depset([out]))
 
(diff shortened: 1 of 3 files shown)


sluongng:
Reference Gazelle code https://github.com/bazelbuild/bazel-gazelle/blob/36b2084aa034ec0313f16d5e205f8973c964a1ae/walk/walk.go#L300

rules/flags/index.bzl
```diff
@@ -1,7 +1,8 @@
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 
 def _write_flag_to_file_impl(ctx):
-    out = ctx.actions.declare_file(ctx.attr.name)
+    out_file = ctx.attr.out if ctx.attr.out else ctx.attr.name
+    out = ctx.actions.declare_file(out_file)
```
bduffany, line 5: wdyt about making `out` a required arg rather than defaulting to `name`, so that we don't hit this again in the future?

bduffany (approved):

rules/flags/index.bzl
sluongng: I would need to refactor other places that use this rule to follow that convention.

Let me give it a try to see if it's worth it. I am also willing to change Gazelle code if the convention does not fit our use case :P.

sluongng:
@bduffany PTAL. Idk if I like this better. Right now the usage of this rule is also relatively small for it to matter too much.

bduffany (approved):

PR #3750, opened 2023-04-12
proto/bes: upgrade to latest BES change

sluongng:
This upgrade our BES proto to the latest BES from Bazel's master branch.
The latest change to BES proto in bazel.git, as of this change, is

```
commit d31dd094c56270d892040ce65586af9b76a99744
Author: Googler <wilwell@google.com>
Date:   Thu Dec 22 04:52:07 2022 -0800

    Support multiplex workers in WorkerMetricsCollector.
```

This PR was generated by extracting all the changes from bazel.git with
```
$ git diff c674101d004eccd795a06e19674e317b04b1b36a..HEAD \
      -- \
      src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto \
      > bes.patch
```

The change c674101d004eccd795a06e19674e317b04b1b36a was already applied in our version of bes.proto
so I assumed that's when we last updated/copied the file.

And then apply it into buildbuddy.git with

```
$ git apply -v ../../bazelbuild/bazel/bes.patch -C2 -p10
```

Note that `-C` was used to fuzzily apply some hunks without exact match.

Some style changes was manually filtered out when commiting this PR

```
--- a/proto/build_event_stream.proto
+++ b/proto/build_event_stream.proto
@@ -444,7 +444,8 @@ message WorkspaceStatus {
 // build.
 message BuildMetadata {
   // Custom metadata for the build.
-  map<string, string> metadata = 1;
+  map<string, string> metadata = 1
+      ;
 }

 // Payload of an event reporting details of a given configuration.
@@ -452,7 +453,8 @@ message Configuration {
   string mnemonic = 1;
   string platform_name = 2;
   string cpu = 3;
-  map<string, string> make_variable = 4;
+  map<string, string> make_variable = 4
+      ;
   // Whether this configuration is used for building tools.
   bool is_tool = 5;
 }
```

---

**Version bump**: Patch

### DIFF
--- a/proto/build_event_stream.proto
+++ b/proto/build_event_stream.proto
@@ -453,6 +453,8 @@ message Configuration {
   string platform_name = 2;
   string cpu = 3;
   map<string, string> make_variable = 4;
+  // Whether this configuration is used for building tools.
+  bool is_tool = 5;
 }
 
 // Payload of the event indicating the expansion of a target pattern.
@@ -522,6 +524,13 @@ message File {
     // The contents of the file, if they are guaranteed to be short.
     bytes contents = 3;
   }
+
+  // Digest of the file, using the build tool's configured digest algorithm,
+  // hex-encoded.
+  string digest = 5;
+
+  // Length of the file in bytes.
+  int64 length = 6;
 }
 
 message Tree {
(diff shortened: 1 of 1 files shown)


sluongng:
Tangent: should we consider using the original bes proto and maintain a `buildbuddy.patch` file separately? The 2 then could be recombined using a genrule calling `patch(1)`.

It would help us track the internal changes against the upstream file easier. 🤔

bduffany:
> Tangent: should we consider using the original bes proto and maintain a `buildbuddy.patch` file separately? The 2 then could be recombined using a genrule calling `patch(1)`.

I like having the BES proto checked into the repo since it's one of the most frequently referenced protos. Maybe we could have a script that runs the genrule and copies the output contents to the source tree

luluz66 (approved):

PR #3751, opened 2023-04-12
Disable more flaky tests on Mac CI

bduffany:
We need to dedicate some time to fixing these tests, but for now let's disable them since they almost always show up red and might hide other unexpected errors that are mac-specific.

---

**Version bump**: Patch


sluongng (approved):
It's really a shame that my ci_runner fix did not work 😮‍💨

PR #3754, opened 2023-04-12
Fix SELECT FOR UPDATE syntax in installation unlink query

bduffany:
Also add FOR UPDATE to a few related transactions that do SELECT followed by DELETE.

---

**Version bump**: Patch


vadimberezniker (approved):

tempoz (approved):
LGTM

PR #3755, opened 2023-04-12
Add keyboard shortcuts

iain-macdonald:
This PR adds several keyboard shortcuts in the enterprise app, specifically:

- ? and Esc for showing and hiding a little help menu.
- j/k for navigating the invocation history page.
- Enter and u for navigating to and from invocations from the history page.
- Shift-C for copying the invocation page link.
- And several g-Key shortcuts for navigating to different pages (all builds, trends, settings, etc.)

This is all protected behind a user preference that's enabled in the settings page.

**Version bump**: Minor


enterprise/app/settings/settings.tsx
```diff
@@ -196,6 +196,15 @@ export default class SettingsComponent extends React.Component<SettingsProps> {
                     onClick={() => this.props.preferences.toggleLightTerminal()}>
                     Switch to {this.props.preferences.lightTerminalEnabled ? "dark" : "light"} log viewer theme
                   </FilledButton>
+                  <div className="settings-option-title">Keyboard Shortcuts</div>
+                  <div className="settings-option-description">
+                    Enables keyboard shortcuts -- hit '?' for help when enabled.
```
bduffany, line 201: nit:

```suggestion
                    Enables keyboard shortcuts. Hit '?' for help when enabled.
```

enterprise/app/root/root.tsx
```diff
@@ -182,12 +290,13 @@ export default class EnterpriseRootComponent extends React.Component {
                 path={this.state.path}
                 hash={this.state.hash}
                 user={this.state.user}
-                search={this.state.search}></SidebarComponent>
+                search={this.state.search}
+                coolMode={this.state.coolMode}></SidebarComponent>
```
bduffany, line 294: it looks like this prop isn't used?

app/invocation/invocation_share_button.tsx
```diff
@@ -106,11 +130,19 @@ export default class InvocationShareButtonComponent extends React.Component<
...
-          {/* TODO: Use an icon that signifies the current permissions */}
-          <Share2 className="icon white" />
-          Share
-        </FilledButton>
+        {this.state.showingCopied && (
+          <OutlinedButton className="invocation-share-button">
+            <ClipboardCheck className="icon grey" />
+            <div class="invocation-share-button-text">Copied</div>
+          </OutlinedButton>
+        )}
```
bduffany, line 138: We have a few different places in the UI now that have custom "Copied" states like this.

I think we could simplify the code and unify the UX by calling `alert_service.success("Copied invocation link to clipboard")` which will show a banner for a second or two. We could even put the alert in `app/util/clipboard.ts > copyToClipboard()` and give it a second argument like `notification` which defaults to a generic "Copied to clipboard!" message. Could also add a special alert type like `clipboard` to give it a nicer icon / background, if you feel like it

app/invocation/invocation_share_button.tsx
```diff
@@ -42,12 +45,33 @@ export default class InvocationShareButtonComponent extends React.Component<
 
   private inputRef = React.createRef<HTMLInputElement>();
 
+  componentWillMount() {
+    let handle = shortcuts.register(KeyCombo.shift_c, () => {
+      this.copyShareUrl();
+    });
+    this.state.keyboardShortcutHandle = handle;
```
bduffany, line 52: Setting things on state like `this.state.<...> = <...>` without calling `setState()` is generally  an anti-pattern in React.

Another option is to change the `state` initializer at the top of the class:

```
state = {
  keyboardShortcutHandle: shortcuts.register(KeyCombo.shift_c, () => this.copyShareUrl()),
  ...this.getInitialState(),
}
```

enterprise/app/settings/settings.tsx
iain-macdonald: Done.

app/invocation/invocation_share_button.tsx
iain-macdonald: Ahh, didn't see that. Changed to use it and backed out the outlined button changes.

app/invocation/invocation_share_button.tsx
iain-macdonald: Whoops, thought I cleaned all these up.

app/shortcuts/shortcuts.tsx
```diff
@@ -0,0 +1,154 @@
...
+  // same shortcut is registered multiple times, so if the calling component
+  // will unmount and remount, shortcuts should be deregistered and
+  // reregistered.
+  public registerSequence(keyCombo: KeyCombo[], action: () => void): string {
+    if (this.shortcuts == null) {
+      this.shortcuts = new Map<string, Shortcut>();
+      document.addEventListener(
+        "keydown",
+        function (event: KeyboardEvent) {
+          if (window.localStorage.getItem(keyboardShortcutsKey) !== keyboardShortcutsValue) {
```
bduffany, line 125: nit: not a huge deal but I wonder if this might cause some subtle jank while typing since it's called on every keystroke -  I know localStorage used to be somewhat slow but maybe it has been optimized at this point. wdyt about referencing `preferences.keyboardShortcutsEnabled` here instead just to be safe?

enterprise/app/root/root.tsx
```diff
@@ -164,6 +182,96 @@ export default class EnterpriseRootComponent extends React.Component {
             </OutlinedButton>
           </div>
         )}
+        {this.state.keyboardShortcutHelpShowing && (
```
bduffany, line 185: sorry I should've caught this in the last round - do you mind putting this in a separate package like `enterprise/app/shortcuts`? Then you could put it at the bottom along with the other modals like `<GroupSearchComponent>`, `<AlertComponent>` etc.

bduffany (approved):

app/shortcuts/shortcuts.tsx
iain-macdonald: Sounds good, done

enterprise/app/root/root.tsx
iain-macdonald: Sure thing, done

PR #3760, opened 2023-04-12
strictify enterprise/app/root for real

jdhollen:
---

**Version bump**: Patch


iain-macdonald (approved):
so strict

jdhollen:
OK, I feel good about this now--the only code I see that is checking for an 'undefined' user is in `InvocationComponent`, but the FOSS root component statically passes `null` to `InvocationComponent`, and the previous code in the enterprise root always passed either `null` or a real user, so swapping over to undefined and removing the undefined check in `InvocationComponent` should be safe.

iain-macdonald (approved):

PR #3763, opened 2023-04-12
Add platform and credentials to GetSociArtifacts

iain-macdonald:
Because it has to resolve the image name into a reference just like GetOptimizedImage, so the request may as well look the same.

**Version bump**: Minor

### DIFF
--- a/proto/soci.proto
+++ b/proto/soci.proto
@@ -1,5 +1,6 @@
 syntax = "proto3";
 
+import "proto/registry.proto";
 import "proto/remote_execution.proto";
 
 package soci;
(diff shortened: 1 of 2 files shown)


proto/soci.proto
```diff
@@ -28,7 +29,14 @@ message Artifact {
 }
 
 message GetArtifactsRequest {
-  string image_name = 1;
+  string image = 1;
+
+  // If an image has multiple variants, this selector is used
+  // to select a single variant.
+  registry.Platform platform = 2;
```
vadimberezniker, line 36: feel free to copy over these two messages into this file. that way we can delete `registry.proto` when you're done.

vadimberezniker (approved):

proto/soci.proto
iain-macdonald: Because the code using these is going to be interacting with the stuff in registry.go which uses the registry.proto versions, it's actually easier to have a shared definition, so I'll just leave them in there for now and we can move them in the future if/when it makes sense.

PR #3765, opened 2023-04-13
Simplify PR template

bduffany:
- Remove semver stuff since we aren't consuming that info anywhere.
- Shorten reminders at the top (I can add these back if people feel they are helpful, but I'm guessing people read this once then ignore it. We can add these to a separate CONTRIBUTING.md file and link to it, if we think this info is valuable?)

---

**Version bump**: Patch


siggisim (approved):
Can we remove that `---`? That's my least favorite part

bduffany:
> Can we remove that `---`? That's my least favorite part

Done

PR #3766, opened 2023-04-13
Add a caching layer for encryption/decryption keys.

vadimberezniker:
Allow keys to be cached in memory for up to 10 minutes. Start trying to refresh keys after 5 minutes, if the key has been used recently. Remove key from cache if we are unable to refresh it before expiration or if it was not used recently.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/crypter_service/crypter_service.go
+++ b/enterprise/server/crypter_service/crypter_service.go
@@ -6,21 +6,33 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/retry"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/jonboulle/clockwork"
+	"go.uber.org/atomic"
 	"golang.org/x/crypto/chacha20poly1305"
 	"golang.org/x/crypto/hkdf"
+	"golang.org/x/sync/singleflight"
 	"gorm.io/gorm"
 
 	rfpb "github.com/buildbuddy-io/buildbuddy/proto/raft"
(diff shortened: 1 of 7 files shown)


enterprise/server/crypter_service/crypter_service.go
```diff
@@ -6,53 +6,387 @@ import (
...
+			})
+
+			c.mu.Lock()
+			c.lastRefreshRun = c.clock.Now()
+			c.mu.Unlock()
+		}
+	}()
+}
+
+func (c *keyCache) testGetLastRefreshRun() time.Time {
```
maggie-lou, line 172: Did you mean to have these test functions?

maggie-lou (approved):

enterprise/server/crypter_service/crypter_service.go
```diff
@@ -6,53 +6,387 @@ import (
...
+	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/retry"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/jonboulle/clockwork"
```
bduffany, line 23: nit: I recall at one point I pulled in a lib like this and after some discussion w/ Tyler I added `timeutil.Clock` and `testclock` utils. I don't have a super strong opinion but we might want to use those just for consistency, to avoid introducing this direct dep

enterprise/server/crypter_service/crypter_service.go
maggie-lou: Ah yup 👍

bduffany (approved):

enterprise/server/crypter_service/crypter_service.go
vadimberezniker: I kind of wrecked my brain trying to think of a less invasive way to test it, but ended up with this slightly unsatisfactory approach.

enterprise/server/crypter_service/crypter_service.go
vadimberezniker: I'm relying on some additional functionality from the clockwork library, notably `time.After` which is slightly more complicated. 

I understand his reasoning for not wanting to add a dep for a simple use like `time.Now` but now that we have more complex needs I think it's reasonable to pull in the `clockwork` dep instead of implementing it ourselves. 

To make things more consistent, I can go ahead and replace the use of timeutil.Clock with clockwork (in a separate PR). How do you feel about that?

PR #3769, opened 2023-04-14
Add protos for summary in Trends page.

luluz66:


proto/stats.proto
```diff
@@ -121,6 +121,22 @@ message TrendStat {
   int64 branch_count = 20;
 }
 
+message Summary {
+  // The total number of builds from a specific period.
+  int64 num_builds = 1;
+
```
jdhollen, line 127: what happened to field #2 :)

luluz66: Done.

proto/stats.proto
```diff
@@ -143,6 +159,11 @@ message GetTrendResponse {
   repeated ExecutionStat execution_stat = 3;
 
   bool has_invocation_stat_percentiles = 4;
+
+  // A summary of metrics from the period specified in the query.
+  Summary summary_current = 5;
+  // A summary of metrics from the previous period.
```
jdhollen, line 165: maybe just add a little bit here about this period being equal in length to the one the user requested.

jdhollen (approved):

proto/stats.proto
luluz66: Done.

PR #3774, opened 2023-04-14
Add public API to execute a workflow

maggie-lou:
Customers have requested a public API to execute workflows, for example if they want to schedule them with a cron ([Example thread](https://buildbuddy.slack.com/archives/CUY16GNK1/p1681490853254989))

Personally excited for this so we don't have to use the janky local webhook script to test workflows :) 

**Related Issues:** https://github.com/buildbuddy-io/buildbuddy-internal/issues/2080 and https://github.com/buildbuddy-io/buildbuddy-internal/issues/2079 (if combined with GitHub cron scheduling)


maggie-lou:
* master
  * **PR #3774** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3774" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3774?utm_source=stack-comment).

proto/api/v1/workflow.proto
```diff
@@ -0,0 +1,49 @@
...
+
+import "google/rpc/status.proto";
+
+message ExecuteWorkflowRequest {
+  // URL of the repo the workflow is running for
+  // Ex. "https://github.com/some-user/acme"
+  string repo_url = 1;
+  // Branch the workflow should run on
+  // Ex. "cool-feature"
+  string branch = 2;
```
sluongng, line 13: 1. I would suggest renaming this to a `ref` as it should behave the same for git branch as well as git tag.

2. I wonder if we should use commit instead of ref. Reason being that there could be a small window between the time a Workflow was requested and the actual repo being fetch where user could force push and update a branch. In such case, it might be a lot less confusing to have separate builds. Such branch could be rapidly updated by user, or is rapidly updated main branch in a monorepo with thousands of user.

proto/api/v1/workflow.proto
```diff
@@ -0,0 +1,49 @@
...
+  // OPTIONAL FIELDS
+
+  // Name of the workflow action to execute. Corresponds to an action in
+  // buildbuddy.yaml, or an action name in the default workflow config if
+  // buildbuddy.yaml is missing from the repo.
+  //
+  // If this is not set, the API will execute all actions in the workflow
+  //
+  // Ex. "Test all targets"
+  string action_name = 3;
```
sluongng, line 24: I noticed that the response could contain multiple actions, should we make this a list of action instead of a single action?

proto/api/v1/workflow.proto
sluongng: Technically `ref` is a pointer to `commit_sha`, so `commit_sha` alone should suffice.

However, our Web UI could really use the additional data from `ref` to help group the invocations under the same branch/tag together. Therefore, I prefer option (b). 

I don't see any scenarios where you would _NOT_ want to supply a branch/tag name to BuildBuddy.
In fact, I wonder if we should be adding an additional K/V field to allow additional Build Metadata for each Workflow invocation 🤔

maggie-lou: Talked offline. For now, will just have a `ref` field and not a `commit_sha` field and translate the ref to a commit sha in the backend

We're expecting that for the majority of use cases of this manual API, people will want the top commit. This will save them the tedium of having to grab it themselves. We can easily add an optional commit field if it's requested

proto/api/v1/workflow.proto
maggie-lou: Added in this PR: https://github.com/buildbuddy-io/buildbuddy/pull/3805

server/backends/github/github.go
```diff
@@ -441,6 +441,60 @@ func (c *GithubClient) CreateStatus(ctx context.Context, ownerRepo string, commi
...
+	}
+
+	var b githubBranch
+	if err = json.Unmarshal(bytes, &b); err != nil {
+		return "", err
+	}
+
+	return b.Commit.Sha, nil
+}
+
```
bduffany, line 497: One downside of making this API request is that it consumes API quota. I am starting to worry a little more about API quota since Brentley already hit a quota limit with rules_xcodeproj.

If the commit_sha is not needed except by the CI runner action, maybe the CI runner could just default to HEAD if the `--commit_sha` flag is missing?

enterprise/server/workflow/service/service.go
```diff
@@ -601,25 +555,106 @@ func (ws *workflowService) getActions(ctx context.Context, wf *tables.Workflow,
...
+func (ws *workflowService) useCleanWorkflow(ctx context.Context, wf *tables.Workflow) error {
+	isLegacyWorkflow := !isRepositoryWorkflowID(wf.WorkflowID)
+
+	suffix, err := random.RandomString(10)
+	if err != nil {
+		return err
+	}
+	wf.InstanceNameSuffix = suffix
+
+	if isLegacyWorkflow {
```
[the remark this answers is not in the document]
maggie-lou: Ah good catch

server/backends/github/github.go
maggie-lou: Removed the call. Here's the commit to default to HEAD in the CI runner: https://github.com/buildbuddy-io/buildbuddy/pull/3826

maggie-lou: Checked that this [reference](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/workflow/service/service.go#L978) is okay if the sha is not set

This [reference](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/workflow/service/service.go#L1158) will be a nop if the sha is not set, because we need a commit to update the status for

sluongng: alternatively, we could run `git ls-remote` without any authentication.

```
> git ls-remote https://github.com/git/git.git next
bd7f14d9353b7ce9acf82b32f80f585373a5ebe3        refs/heads/next
```

Not sure if Github does rate limit on unauthenticated git operations.

enterprise/server/workflow/service/service.go
```diff
@@ -610,25 +561,104 @@ func (ws *workflowService) getActions(ctx context.Context, wf *tables.Workflow,
...
+		FROM Workflows
+		WHERE group_id = ?
+		AND repo_url = ?
+	`, groupID, repoURL).Take(wf).Error
+	if err != nil {
+		if db.IsRecordNotFound(err) {
+			return nil, status.NotFoundErrorf("repo %q not found", repoURL)
+		}
+		return nil, status.InternalErrorf("failed to look up repo %q: %s", repoURL, err)
+	}
```
bduffany, line 616: I think we should remove the legacy workflow lookup, and we can tell people that if they want to use the public API then GitHub App authentication is required.

I eventually want to phase out the Workflows table and this is sort of a small step towards that.

enterprise/server/workflow/service/service.go
```diff
@@ -610,25 +561,104 @@ func (ws *workflowService) getActions(ctx context.Context, wf *tables.Workflow,
...
 	}
 	gitRepository := &tables.GitRepository{}
-	err = ws.env.GetDBHandle().DB(ctx).Raw(`
+	err := ws.env.GetDBHandle().DB(ctx).Raw(`
 		SELECT *
 		FROM GitRepositories
 		WHERE group_id = ?
 		AND repo_url = ?
-	`, groupID, repoURL.String()).Take(gitRepository).Error
+	`, groupID, repoURL).Take(gitRepository).Error
```
bduffany, line 661: not sure if GORM knows to automatically call `.String()` here, but it feels a bit less "magical" to be explicit about it either way:

```suggestion
	`, groupID, repoURL.String()).Take(gitRepository).Error
```

server/backends/github/github.go
```diff
@@ -441,6 +441,60 @@ func (c *GithubClient) CreateStatus(ctx context.Context, ownerRepo string, commi
...
+		return "", status.UnknownErrorf("HTTP %s: %q", res.Status, string(bytes))
+	}
+
+	var b githubBranch
+	if err = json.Unmarshal(bytes, &b); err != nil {
+		return "", err
+	}
+
+	return b.Commit.Sha, nil
+}
```
bduffany, line 496: can this file be reverted since this is now unused?

bduffany (approved):

enterprise/server/api/api_server.go
```diff
@@ -492,6 +492,11 @@ func (s *APIServer) ExecuteWorkflow(ctx context.Context, req *apipb.ExecuteWorkf
 
 	wf, err := wfs.GetWorkflowByGroupAndRepo(ctx, user.GetGroupID(), req.GetRepoUrl())
 	if err != nil {
+		if status.IsNotFoundError(err) {
+			return nil, status.NotFoundErrorf("Workflow for repo %s not found. Note that Github App authentication is required"+
+				" to use this API. If you are using the legacy Workflow product, you will need"+
+				" to migrate to a Github App first.", req.GetRepoUrl())
```
bduffany, line 498: The workflows page encourages people to migrate, so we can link people there as an actionable step, and also avoid having to update this messaging later

```suggestion
			return nil, status.NotFoundErrorf("Workflow for repo %q not found. Check the Workflows page in the BuildBuddy app and ensure this repo is properly linked.", req.GetRepoUrl())
```

bduffany (approved):
nice!

PR #3775, opened 2023-04-15
Add cache API to check whether encryption is supported for the user.

vadimberezniker:
This is currently only true for pebble cache if the user is mapped to a partition that has encryption support enabled.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -900,3 +900,7 @@ func (mc *MigrationCache) Stop() error {
 func (mc *MigrationCache) SupportsCompressor(compressor repb.Compressor_Value) bool {
 	return mc.src.SupportsCompressor(compressor) && mc.dest.SupportsCompressor(compressor)
 }
+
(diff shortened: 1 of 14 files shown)


enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -900,3 +900,7 @@ func (mc *MigrationCache) Stop() error {
 func (mc *MigrationCache) SupportsCompressor(compressor repb.Compressor_Value) bool {
 	return mc.src.SupportsCompressor(compressor) && mc.dest.SupportsCompressor(compressor)
 }
+
+func (mc *MigrationCache) SupportsEncryption(ctx context.Context) bool {
+	return mc.src.SupportsEncryption(ctx) && mc.dest.SupportsEncryption(ctx)
```
bduffany, line 905: should we maybe add a precondition check in the migration cache so that we can't accidentally migrate from a cache that supports encryption, to a cache that doesn't?

bduffany (approved):

bduffany (approved):

PR #3776, opened 2023-04-16
Refactor ExecuteWorkflow to execute all actions if action name is not set in the request

maggie-lou:


maggie-lou:
* master
  * **PR #3776** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3776" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3776?utm_source=stack-comment).

proto/workflow.proto
```diff
@@ -169,8 +172,11 @@ message ExecuteWorkflowResponse {
   // The response context.
   context.ResponseContext response_context = 1;
 
+  // TODO(Maggie): Deprecate this field and clean up references
   // The invocation ID of the workflow execution.
   string invocation_id = 2;
+
+  map<string, string> action_name_to_invocation_id = 3;
```
bduffany, line 179: Can we use something like `repeated WorkflowActionInvocation invocations = 3;` and then have `message WorkflowActionInvocation` with invocation_id and action_name fields? That way if we need to attach more metadata to each started action, we don't need a breaking API change. (e.g. maybe we'll want to add a `status` field which contains a gRPC status for any actions that failed to start?)

enterprise/server/workflow/service/service.go
```diff
@@ -492,7 +485,59 @@ func (ws *workflowService) ExecuteWorkflow(ctx context.Context, req *wfpb.Execut
...
+	actions, err := ws.getActions(ctx, wf, wd, req.GetActionName())
+	if err != nil {
+		return nil, err
+	}
+
+	eg, egCtx := errgroup.WithContext(ctx)
+	actionToInvocationID := make(map[string]string, len(actions))
+	for _, action := range actions {
+		action := action
+		eg.Go(func() error {
```
bduffany, line 497: it might make sense to use a regular `WaitGroup` for this so that we can allow some actions to be started and some to fail (e.g. if they try to execute with `os: darwin` and no mac executors are registered). Then we can report a google.grpc.Status for each started action, with the error code

maggie-lou: Done. Added a status_code to the Action struct

enterprise/server/workflow/service/service.go
```diff
@@ -505,28 +550,27 @@ func (ws *workflowService) ExecuteWorkflow(ctx context.Context, req *wfpb.Execut
...
-	// The workflow execution is trusted since we're authenticated as a member of
-	// the BuildBuddy org that owns the workflow.
-	isTrusted := true
-	executionID, err := ws.executeWorkflow(ctx, apiKey, wf, wd, isTrusted, action, invocationID, extraCIRunnerArgs)
-	if err != nil {
-		return nil, err
-	}
-	if err := ws.waitForWorkflowInvocationCreated(ctx, executionID, invocationID); err != nil {
-		return nil, err
+	if !foundAction {
```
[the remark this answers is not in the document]
maggie-lou: Think I fixed what you were getting at here

proto/workflow.proto
maggie-lou: Done

proto/workflow.proto
```diff
@@ -169,8 +172,25 @@ message ExecuteWorkflowResponse {
...
+  // Corresponds to an action in buildbuddy.yaml, or an action name in the
+  // default workflow config if buildbuddy.yaml is missing from the repo. Ex.
+  // "Test all targets"
+  string action_name = 1;
+
+  // The invocation ID of the execution of the action.
+  string invocation_id = 2;
+
+  // The status code for executing the action.
+  string grpc_status_code = 3;
```
bduffany, line 193: you can use `google.rpc.Status` here (from `google/rpc/status.proto`) so that the client can see the full error message in addition to the status code, which will help with debugging. There's a handy util `gstatus.Convert(err).Proto()` to get a status proto from a go error (also returns the correct `nil` status when `err == nil`)

proto/workflow.proto
```diff
@@ -169,8 +172,25 @@ message ExecuteWorkflowResponse {
...
 
+  // TODO(Maggie): Deprecate this field and clean up references
   // The invocation ID of the workflow execution.
   string invocation_id = 2;
+
+  // A list of the actions executed by the API.
+  repeated WorkflowAction actions_started = 3;
+}
+
+message WorkflowAction {
```
bduffany, line 183: I feel like `WorkflowAction` is a little ambiguous - it could be referring to the action config, a completed action, or something else. Maybe nesting it under `ExecuteWorkflowResponse` and calling it `ActionStatus` would be less ambiguous?

Sorry if this sounds nit-picky, just want to make sure we avoid confusion down the road and have to introduce a breaking change in the generated client code.

maggie-lou: Np! Thanks as always for the detailed reviews 🙏

proto/workflow.proto
maggie-lou: Ah cool!

PR #3721, opened 2023-04-05
Delete timing profile from cache after copying to blobstore

tempoz:
Don't continue to persist the timing profile in the cache after it has been stored in blobstore,
as it would be redundant and use up cache that could be devoted to something else.

---

**Version bump**: Minor

PR #3720, opened 2023-04-05
Add fallback for `file/download` requests

tempoz:
Add the `with_fallback` param to the `file/download` endpoint so we can gracefully fall back
to an alternative location for an artifact (timing profile, test logs) if it isn't found in
the first location. Also change logging around a bit so that fallbacks don't result in
noisy logs.

---

**Version bump**: Minor

PR #3719, opened 2023-04-05
Serve timing profile artifact from blobstore

tempoz:
Add the ability to retrieve an invocation's timing profile artifact from blobstore.

---

**Version bump**: Minor

PR #3718, opened 2023-04-05
Store timing profile in blobstore

tempoz:
Store timing profile in blobstore so we can persist it after it is no longer in the cache.

---

**Version bump**: Minor

PR #3780, opened 2023-04-17
Persist test logs in blobstore

tempoz:
Persist test logs in the blobstore so that they are still visible after they would have fallen
out of the cache.

---

**Version bump**: Minor

### DIFF
--- a/server/build_event_protocol/accumulator/accumulator.go
+++ b/server/build_event_protocol/accumulator/accumulator.go
@@ -75,14 +75,14 @@ type Accumulator interface {
 // memory for the life of the stream, so it should not save every single event
 // in full (that data lives in blobstore).
 type BEValues struct {
-	valuesMap               map[string]string
-	sawWorkspaceStatusEvent bool
-	sawBuildMetadataEvent   bool
-	sawFinishedEvent        bool
-	buildStartTime          time.Time
-	profileURI              *url.URL
-	profileName             string
-	hasTestActionOutputs    bool
+	valuesMap                      map[string]string
+	sawWorkspaceStatusEvent        bool
+	sawBuildMetadataEvent          bool
+	sawFinishedEvent               bool
+	buildStartTime                 time.Time
+	bytestreamProfileURI           *url.URL
+	profileName                    string
+	hasBytestreamTestActionOutputs bool
 
 	// TODO(bduffany): Migrate all parser functionality directly into the
 	// accumulator. The parser is a separate entity only for historical reasons.
@@ -124,21 +124,18 @@ func (v *BEValues) AddEvent(event *build_event_stream.BuildEvent) {
 			if uri := toolLog.GetUri(); uri != "" {
 				if url, err := url.Parse(uri); err != nil {
 					log.Warningf("Error parsing uri from BuildToolLogs: %s", uri)
-				} else {
-					v.profileURI = url
-					v.profileName = toolLog.Name
+				} else if url.Scheme == "bytestream" {
+					v.bytestreamProfileURI = url
 				}
 				break
 			}
 		}
 	case *build_event_stream.BuildEvent_TestResult:
 		for _, f := range p.TestResult.TestActionOutput {
-			if f.GetName() == "test.log" || f.GetName() == "test.xml" || f.GetName() == "test.lcov" {
-				if u, err := url.Parse(f.GetUri()); err != nil {
-					log.Warningf("Error parsing uri from TestResult: %s", f.GetUri())
-				} else if u.Scheme == "bytestream" {
-					v.hasTestActionOutputs = true
-				}
+			if u, err := url.Parse(f.GetUri()); err != nil {
+				log.Warningf("Error parsing uri from TestResult: %s", f.GetUri())
+			} else if u.Scheme == "bytestream" {
+				v.hasBytestreamTestActionOutputs = true
 			}
 		}
 	}
@@ -221,16 +218,12 @@ func (v *BEValues) BuildFinished() bool {
 	return v.sawFinishedEvent
 }
 
-func (v *BEValues) ProfileURI() *url.URL {
-	return v.profileURI
+func (v *BEValues) BytestreamProfileURI() *url.URL {
+	return v.bytestreamProfileURI
 }
 
-func (v *BEValues) ProfileName() string {
-	return v.profileName
-}
-
-func (v *BEValues) HasTestActionOutputs() bool {
-	return v.hasTestActionOutputs
+func (v *BEValues) HasBytestreamTestActionOutputs() bool {
+	return v.hasBytestreamTestActionOutputs
 }
 
 func (v *BEValues) getStringValue(fieldName string) string {
--- a/server/build_event_protocol/accumulator/accumulator.go
+++ b/server/build_event_protocol/accumulator/accumulator.go
@@ -82,6 +82,8 @@ type BEValues struct {
 	buildStartTime          time.Time
 	profileURI              *url.URL
 	profileName             string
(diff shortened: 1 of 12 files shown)


tempoz:
* master
  * **PR #3717** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3717" target="_blank">Graphite</a> 
    * **PR #3718** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3718" target="_blank">Graphite</a> 
      * **PR #3719** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3719" target="_blank">Graphite</a> 
        * **PR #3720** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3720" target="_blank">Graphite</a> 
          * **PR #3721** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3721" target="_blank">Graphite</a> 
            * **PR #3780** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3780" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3780?utm_source=stack-comment).

server/build_event_protocol/accumulator/accumulator.go
```diff
@@ -127,6 +129,17 @@ func (v *BEValues) AddEvent(event *build_event_stream.BuildEvent) {
...
+		for _, f := range p.TestResult.TestActionOutput {
+			if f.GetName() == "test.log" {
+				if url, err := url.Parse(f.GetUri()); err != nil {
+					log.Warningf("Error parsing uri from TestResult: %s", f.GetUri())
+				} else {
+					v.testLogURI = url
+					v.testLogName = f.GetName()
+				}
+			}
+		}
```
bduffany, line 142: If there are multiple test results, then I think we'll want the ability to serve multiple log files.

We don't want to add multiple log files to the accumulator though since those are variable in length and in general we want the accumulator to be a flat struct that doesn't depend on the size of the invocation, since it sits in memory while the invocation is in progress.

Instead, we could have the test log endpoint read the invocation and then scan the BES events for the desired test log.

bduffany (changes requested):

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -777,6 +777,7 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 		invocation,
 		map[string]*url.URL{
 			"timing_profile/" + e.beValues.ProfileName(): e.beValues.ProfileURI(),
+			"test_log/" + e.beValues.TestLogName():       e.beValues.TestLogURI(),
```
bduffany, line 780: we'll probably want to scan the BES stream here to get a list of test logs. In `webhookNotifier` (which also runs in the background, similarly to statsRecorder), we are already looking up the full invocation BES event stream. Maybe we could somehow reuse that work? ideally in a way where we aren't keeping the full BES stream buffered in memory for too long.

server/build_event_protocol/accumulator/accumulator.go
tempoz: Done!

server/build_event_protocol/build_event_handler/build_event_handler.go
tempoz: This seemed very complicated to do; I think it is work worth doing, but may be outside the scope of this PR. We probably want to make some sort of BuildEventStreamPostProcess pipeline that we can feed arbitrary functions into and then it can load the BES (ideally in small chunks via streaming) and map all the functions over the events?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -408,35 +416,40 @@ func (r *statsRecorder) handleTask(ctx context.Context, task *recordStatsTask) {
...
-			log.CtxErrorf(ctx, "Failed to commit to blobstore for path %s to persist cache artifact at %s: %s", path, uri.String(), err)
-			w.Close()
-			continue
-		}
-		if err := w.Close(); err != nil {
-			log.CtxErrorf(ctx, "Failed to close blobstore writer for path %s to persist cache artifact at %s: %s", path, uri.String(), err)
+		switch uri.Scheme {
+		case "":
+			switch uri.Path {
+			case "//test_data/":
```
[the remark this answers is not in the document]

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -464,6 +477,82 @@ func (r *statsRecorder) Stop() {
...
+			if resultURI.Scheme != "bytestream" {
+				continue
+			}
+			if _, ok := testArtifactsToPersist[f.Name]; !ok {
+				continue
+			}
+			target := strings.TrimPrefix(e.BuildEvent.Id.GetTestResult().Label, "//")
+			if i := strings.LastIndex(target, ":"); i != -1 {
+				target = target[:i] + "/" + target[i:]
+			}
```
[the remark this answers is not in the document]
siggisim: I think doing this path manipulation is error prone, and there are all kinds of edge cases that you'd have to account for i.e. test with multiple runs, tests with multiple shards, retries, etc.

You wouldn't have to do a fallback for every request if you used cache paths - you could just take a boolean flag on whether or not to fallback to blobstore, and we could set that in the UI on places where there might be a fallback (i.e. test logs and timing profiles for now).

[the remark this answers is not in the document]
bduffany: Not sure if this is what you meant @siggisim but instead of using the path at all, we could use the SHA256 of the test artifact as the blobstore path. e.g. if the file name is bytestream://remote.buildbuddy.io/blob/abc123/456 then we could store it in blobstore under `{invocation_id}/artifacts/sha256/abc123` -- would that work?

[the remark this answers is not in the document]
tempoz: switched to cache-style paths.

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -93,6 +94,13 @@ var (
 			"local cache stats to the backing storage, before finalizing stats in the DB.")
 )
 
+var cacheArtifactsBucket = path.Join("artifacts", "cache")
```
bduffany, line 97: nit: s/Bucket/BlobstorePath/ (since bucket has a different meaning GCS)

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1187,16 +1188,10 @@ func (s *BuildBuddyServer) serveUsingParams(w http.ResponseWriter, r *http.Reque
...
-		// a fallback is specified.
-		if fallback, parseErr := url.Parse(params.Get("with_fallback")); parseErr != nil {
-			http.Error(w, parseErr.Error(), http.StatusBadRequest)
-		} else if fallback.RawQuery == "" {
-			http.Error(w, fmt.Sprintf("Fallback without query is invalid: %v", fallback), http.StatusBadRequest)
-		} else {
-			q := fallback.Query()
-			q.Del("with_fallback") // Only fall back once.
-			s.serveUsingParams(w, r, q)
+		// Attempt to fall back gracefully if the specified file is not found
```
bduffany, line 1191: I think we should only do this fallback in the "bytestream_url" case above, since it looks like in the "artifact" case it's just a retry and not a fallback, if I'm reading it right?

bduffany (approved):

PR #3781, opened 2023-04-17
Add schema for tagging builds.

jdhollen:
I'm going with a plain comma-separated list in SQL-land for now, and using an array in Clickhouse.  Adding a multi-value JSON index in MySQL requires a `COPY` update on the parent table, which is a non-starter for `Invocations`.  I don't see a way to get a performant query in MySQL for this use case using a Join table, and the complexity that it would add is prohibitive (in this author's opinion).  **So**, the current plan is to do tag queries (when we need them) via Clickhouse, but keep storing the tags in MySQL so that the data at least exists for on-prem deployments that don't have an OLAP db.

**Related issues**: N/A

### DIFF
--- a/server/util/clickhouse/schema/schema.go
+++ b/server/util/clickhouse/schema/schema.go
@@ -120,7 +120,6 @@ func (i *Invocation) ExcludedFields() []string {
 		"RedactionFlags",
 		"CreatedWithCapabilities",
 		"Perms",
-		"Tags", // TODO(jdhollen): Remove when adding write support.
 	}
 }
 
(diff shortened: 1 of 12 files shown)


server/util/clickhouse/schema/schema.go
```diff
@@ -108,6 +109,7 @@ type Invocation struct {
 	TotalUncachedActionExecUsec       int64
 	DownloadThroughputBytesPerSecond  int64
 	UploadThroughputBytesPerSecond    int64
+	Tags                              pq.StringArray `gorm:"type:Array(String);"`
```
luluz66, line 112: isn't pq a Postgres driver?

jdhollen: I somehow convinced myself that this was a universal driver, I'm going to hack on making one for Clickhouse for a bit because there doesn't seem to be one.

jdhollen: This is going to take me a bit, so I'm gonna shelve this for a day or two and get the customer value UI stuff out the door first.

PR #3785, opened 2023-04-18
Add API for enabling and disabling encryption.

vadimberezniker:
**Related issues**: N/A


enterprise/server/crypter_service/crypter_service.go
```diff
@@ -622,3 +623,159 @@ func (c *Crypter) testGetLastCacheRefreshRun() time.Time {
...
+		return status.InternalErrorf("could not generate key: %s", err)
+	}
+	keyID, err := tables.PrimaryKeyForTable("EncryptionKeys")
+	if err != nil {
+		return status.InternalErrorf("could not generate key id: %s", err)
+	}
+
+	encMasterKeyPart, err := masterKeyClient.Encrypt(masterKeyPart, nil)
+	if err != nil {
+		return status.InternalErrorf("could not encrypt master portion of composite ley: %s", err)
```
maggie-lou, line 666: nit: Typo

maggie-lou (approved):

proto/encryption.proto
```diff
@@ -0,0 +1,28 @@
...
+
+import "proto/context.proto";
+
+message SetEncryptionConfigRequest {
+  context.RequestContext request_context = 1;
+
+  bool enabled = 2;
+
+  string key_uri = 3;
+}
```
bduffany, line 13: will the client have to set all fields in the request, and we will directly upsert these to the DB? Might be worth mentioning as a comment that all fields are required, e.g. if the client wants to enable encryption then they also need to set key_uri

enterprise/server/crypter_service/crypter_service.go
```diff
@@ -622,3 +623,159 @@ func (c *Crypter) testGetLastCacheRefreshRun() time.Time {
...
+		return &enpb.SetEncryptionConfigResponse{}, nil
+	}
+
+	if req.Enabled {
+		if err := c.enableEncryption(ctx, req.KeyUri); err != nil {
+			return nil, err
+		}
+		return &enpb.SetEncryptionConfigResponse{}, nil
+	} else {
+		if err := c.disableEncryption(ctx); err != nil {
```
bduffany, line 760: should we also clear out the key URI here? or maybe pass it through from the req

Otherwise it seems there's not a way to clear out / change the key URI without enabling encryption

enterprise/server/crypter_service/crypter_service.go
vadimberezniker: done

enterprise/server/crypter_service/crypter_service.go
vadimberezniker: Disabling encryption deletes the keys from the database which deletes the key URI as well. 
Once encryption is enabled, it will not be possible to change the key URI without first disabling encryption.

bduffany (approved):

PR #3792, opened 2023-04-18
Add protos for runner pool state

bduffany:
Add protos to make the runner pool state serializable so that we can save it to a file and then reconstruct the runner pool state from the file after an executor restart.

Supporting only firecracker for now, but in theory we could also support resuming bare/podman/docker persistent workers after a restart.

**Related issues**: N/A

### DIFF
--- a/proto/runner.proto
+++ b/proto/runner.proto
@@ -1,6 +1,7 @@
 syntax = "proto3";
 
 import "proto/context.proto";
+import "proto/remote_execution.proto";
 
 package runner;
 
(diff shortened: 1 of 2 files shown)


maggie-lou (approved):

proto/runner.proto
```diff
@@ -87,3 +88,60 @@ message RunResponse {
...
+  RunnerKey runner_key = 1;
+
+  // State specific to the isolation type in use.
+  ContainerState container_state = 2;
+
+  // Short debug ID used to identify the runner.
+  string debug_id = 3;
+
+  // Counter that is incremented each time the runner is assigned a new task.
+  int64 task_number = 4;
```
maggie-lou, line 129: nit: Rename to `assigned_task_count`? `number` sounds a bit like an ID

bduffany: much better - done

proto/runner.proto
```diff
@@ -87,3 +88,60 @@ message RunResponse {
...
+  // Runner key.
+  RunnerKey runner_key = 1;
+
+  // State specific to the isolation type in use.
+  ContainerState container_state = 2;
+
+  // Short debug ID used to identify the runner.
+  string debug_id = 3;
+
+  // Counter that is incremented each time the runner is assigned a new task.
```
vadimberezniker, line 128: What's the need for this field?

vadimberezniker (approved):

proto/runner.proto
bduffany: It provides some signal about the health of the runner pool when looking at logs. If this value is very low on average, it means we're hitting the pool limits too frequently or failing to recycle too frequently.

PR #3797, opened 2023-04-18
Allow restoring firecracker runner state after executor restarts

bduffany:
- Write a file `${buildRoot}/_runner_pool_state.bin` on executor shutdown containing the state of all currently `paused` runners.
- When starting the executor, consume this file (read then delete), and reconstitute runners from the encoded `RunnerState` proto.

**Related issues**: N/A


enterprise/server/remote_execution/container/container.go
```diff
@@ -201,6 +202,14 @@ type CommandContainer interface {
 	Stats(ctx context.Context) (*repb.UsageStats, error)
 }
 
+// Stater can optionally be implemented by a CommandContainer to have
+// its state persisted across executor reboots.
+type Stater interface {
```
vadimberezniker, line 207: I would add a `State` method to the container interface and have non-supporting implementations return UNIMPLEMENTED error code, but I'm also okay with this approach if you want to stick with it.

vadimberezniker (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -354,14 +356,7 @@ func NewContainer(ctx context.Context, env environment.Env, imageCacheAuth *cont
...
 	c := &FirecrackerContainer{
-		vmConfig: &fcpb.VMConfiguration{
-			NumCpus:           opts.NumCPUs,
-			MemSizeMb:         opts.MemSizeMB,
-			ScratchDiskSizeMb: opts.ScratchDiskSizeMB,
-			EnableNetworking:  opts.EnableNetworking,
-			InitDockerd:       opts.InitDockerd,
-			DebugMode:         opts.DebugMode,
-		},
+		vmConfig:           opts.VMConfiguration,
```
maggie-lou, line 359: Is it correct that this will be nil if opts.SavedState is set?

To prevent confusion where opts.VMConfiguration can be optionally set, should we just always populate it from opts.SavedState?

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -1245,6 +1367,34 @@ func (p *pool) Shutdown(ctx context.Context) error {
 		if len(runnersToRemove) > 0 {
 			log.Infof("Runner pool: removing %s", runnerSlice(runnersToRemove))
 		}
+
+		for _, r := range pausedRunners {
```
maggie-lou, line 1371: Because we're only saving state of paused runners, does that mean if all runners are executing tasks during a restart, it will take longer to boot up the runner pool?

bduffany:
I am going to close this PR for now because I realized that it will complicate things to have both this PR in place as well as the readonly snapshot / snapshot sharing implementation.

bduffany:
Resurrecting this PR after some discussion with Tyler. We can revert this PR once we have forkable VMs implemented, as this PR will be redundant. But in the meantime, this PR should buy us an increased runner hit rate after executor rollouts.

vadimberezniker (approved):

enterprise/server/remote_execution/container/container.go
bduffany: Done

PR #3799, opened 2023-04-19
Add RemoteCacheOption to Invocation MySQL and ClickHouse

luluz66:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2236

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -159,6 +159,17 @@ type Invocation struct {
 	Success                          bool   `gorm:"type:tinyint(1)"`
 	Attempt                          uint64 `gorm:"not null;default:0"`
 	BazelExitCode                    string
+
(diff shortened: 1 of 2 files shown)


jdhollen (commented):
Somebody else would know this better than I would -- do we want to care about the space occupied by each bool field here?  for build configuration info, would we rather use a bitmap?

bduffany:
> Somebody else would know this better than I would -- do we want to care about the space occupied by each bool field here? for build configuration info, would we rather use a bitmap?

I think until now, we haven't cared. It might be worth doing a migration at some point to collapse all of the bool fields. But I imagine we would do that once we've optimized the comparably huge fields like IID, pattern, commit_sha, repo_url etc.

bduffany (approved):

PR #3800, opened 2023-04-19
executor/dirtools: ensure UploadTree set result symlinks

sluongng:
Currently we only support OutputDirs and OutputFiles. Therefore actions
which produce only symlinks would fail when executed remotely with
BuildBuddy.

See https://github.com/bazelbuild/rules_go/pull/3539 as an example.

Implement support for OutputFileSymlinks and OutputDirectorySymlinks.

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2234


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -307,6 +339,20 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
 		return uploadableFile.FileNode(), nil
 	}
 
+	uploadSymlinkFn := func(parentDir string, target string, info os.FileInfo) (*repb.SymlinkNode, error) {
+		uploadableSymlink, err := newSymlinksToUpload(instanceName, digestFunction, parentDir, info, target)
+		if err != nil {
+			return nil, err
+		}
+		filesToUpload = append(filesToUpload, uploadableSymlink)
```
sluongng, line 347: The current approach is to treat each symlink like a normal file by appending it to `filesToUpload`

In `func uploadFiles()` where we consume this `filesToUpload` slice, we check whether the file entry is a directory or not `uploadableFile.dir == nil` before adding it to our `interfaces.FileCache` where the file would be hardlinked into the cache dir.

Because a symlink is just a file with a different set of permission and content being the Target it should be compatible with FileCache's hardlink approach.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +403,13 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
 				if err != nil {
 					return nil, err
 				}
+				symlinkNode, err := uploadSymlinkFn(fullPath, target, info)
+				if err != nil {
+					return nil, err
+				}
+
+				txInfo.FileCount += 1
+				txInfo.BytesTransferred += int64(len(symlinkNode.GetTarget()))
```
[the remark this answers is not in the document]
bduffany: If I'm understanding the documentation of `output_symlinks` correctly, it seems like all we need to do is for each `output_path`, if it's a symlink, then just add the symlink metadata (name, target) to `output_symlinks`. I would expect the logic here to just be something like

```
if dirHelper.IsOutputPath(fqfn) {
  actionResult.OutputSymlinks = append(actionResult.OutputSymlinks, &repb.SymlinkNode{Name: trimPathPrefix(fqfn, rootDir), Target: target})
}
```

maybe @iain-macdonald can double-check my understanding since I think he implemented output_paths originally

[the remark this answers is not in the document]
sluongng: https://app.buildbuddy.io/invocation/8e025ac8-fb37-4bcd-97c0-dc4001b64c32?actionDigest=aeb0a9661b70e46c23d9f82317754063a5699656227ab28ee04fb3ceb7e36445#action is an example ActionResult executed locally and cached remotely.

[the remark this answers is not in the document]
bduffany: sorry could you elaborate? I'm not sure what I should be looking for in that invocation link, or how this change is related to local execution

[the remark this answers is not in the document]
iain-macdonald: I don't think my change for output_paths affected the symlink stuff at all, but I can take a closer look if you two need another pair of eyes. Though, I do remember that the change I made broke a project depending on some `rules_kotlin` targets, so I'd recommend that you test the example projects [here](https://github.com/bazelbuild/rules_kotlin/tree/master/examples) against your remote execution changes.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +355,17 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
+
+				if _, ok := dirHelper.FindParentOutputPath(fqfn); !ok {
+					actionResult.OutputSymlinks = append(
+						actionResult.OutputSymlinks,
+						&repb.OutputSymlink{
+							Path:   trimPathPrefix(fqfn, rootDir),
+							Target: target,
+						},
+					)
+				}
```
bduffany, line 367: The documentation of `output_symlinks` says:

```
// The output paths of the action that are symbolic links to other paths
```

If I'm understanding it correctly, "output paths" is referring to the `output_paths` field. So I don't think `FindParentOutputPath` is the right thing to do here -- instead we only want to set OutputSymlinks for paths that *exactly match* one of the `OutputPaths`.

Example:

```
input:
  - output_paths= [ /foo, /baz ]
action produces:
  /foo (directory)
    /bar (symlink)
  /baz (symlink)
```

I think the expected `OutputSymlinks` should contain just `/baz` since it exactly matches one of the output paths, but I think with this change it will return both `/bar` and `/baz`.

sluongng: Yup, that's sensible.

For better context, I was also able to dump out the requested Command from rules_go:
```
{
  "arguments":  [
    "bazel-out/k8-opt-exec-60723E07/bin/go/tools/builders/go_path/go_path-bin",
    "-manifest",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path~manifest",
    "-out",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path",
    "-mode",
    "link"
  ],
  "outputFiles":  [
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/extra.txt",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/pkg/linux_amd64/example.com/repo/cmd/bin.a",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/pkg/linux_amd64/example.com/repo/pkg/lib.a",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/pkg/linux_amd64/example.com/repo/pkg/lib/transitive.a",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/pkg/linux_amd64/example.com/repo/vendor/example.com/repo2.a",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/pkg/linux_amd64/github.com/bazelbuild/rules_go/go/tools/bzltestutil.a",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/pkg/plan9_arm/example.com/repo/cmd/bin.a",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/cmd/bin/bin.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/data.txt",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/embed_test.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/embedded_src.txt",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/external_test.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/internal_test.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/lib.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/renamed_embedded_src.txt",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/template/index.html.tmpl",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/testdata/testdata.txt",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/pkg/lib/transitive/transitive.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/example.com/repo/vendor/example.com/repo2/vendored.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/github.com/bazelbuild/rules_go/go/tools/bzltestutil/init.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/github.com/bazelbuild/rules_go/go/tools/bzltestutil/lcov.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/github.com/bazelbuild/rules_go/go/tools/bzltestutil/test2json.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/github.com/bazelbuild/rules_go/go/tools/bzltestutil/wrap.go",
    "bazel-out/k8-fastbuild/bin/tests/core/go_path/link_path/src/github.com/bazelbuild/rules_go/go/tools/bzltestutil/xml.go"
  ],
  "platform":  {
    "properties":  [
      {
        "name":  "OSFamily",
        "value":  "Linux"
      },
      {
        "name":  "container-image",
        "value":  "docker://gcr.io/bazel-public/ubuntu1804-bazel-java11@sha256:2d50853a7edbe59a99bc4141d7a03cb1068157b9766077302b46c4ec94eef151"
      }
    ]
  }
}
```

I did not have the time today but tomorrow, I plan to set this up on my Linux laptop and test + iterate from there.

bduffany: nice, are you also able to capture the local ActionResult?

enterprise/server/remote_execution/dirtools/dirtools.go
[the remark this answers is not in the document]
sluongng: Discussed with Brandon on Slack on this and shared some of the initial findings I got by running the build locally against our remote cache.

@iain-macdonald Note on the Kotlin project. Let me try running it against my change.

enterprise/server/remote_execution/dirtools/dirtools.go
sluongng: Was able to use

```
build:buildbuddy --experimental_remote_grpc_log=./grpc.log
```

in combination with `bb print` to gain very useful insight :D

enterprise/server/remote_execution/dirtools/dirtools.go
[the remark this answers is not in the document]
sluongng: Tested against both `dagger` and `trivial` in rules_kotlin example and they don't have the same problem as rules_go's go_path so both passed with/without this change applied.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +355,32 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
 				if err != nil {
```
sluongng, line 355: I was considering adding it here 

```
				if !(dirHelper.ShouldUploadFile(fqfn) || dirHelper.IsOutputPath(fqfn)) {
					continue
				}
```

But I was worried about breaking existing behavior.

It's unclear to me how often people create/include symlink outputs in their actions _without_ declaring them in `Command.output_path`, and if we should start explicitly requiring folks to declare them.

sluongng:
Added some basic test to make this easier to maintain in the long-run 🙏

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +355,32 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
+					//
+					// Set both the deprecated and the new fields to keep backward compatibility
+					// with older REAPI clients.
+					if targetInfo.IsDir() {
+						actionResult.OutputDirectorySymlinks = append(actionResult.OutputDirectorySymlinks, ospb)
+					} else {
+						actionResult.OutputFileSymlinks = append(actionResult.OutputFileSymlinks, ospb)
+					}
+				}
+				actionResult.OutputSymlinks = append(actionResult.OutputSymlinks, ospb)
```
bduffany, line 382: The documentation of `output_symlinks` says:

```
  // New in v2.1: this field will only be populated if the command
  // `output_paths` field was used, and not the pre v2.1 `output_files` or
  // `output_directories` fields.
```

So (1) based on this part: `this field will only be populated if the command `output_paths` field was used`

I think there should be a check around this line like

```
if IsOutputPath(fqfn) {
  actionResult.OutputSymlinks = append(actionResult.OutputSymlinks, ospb)
}
```

However, it looks like `IsOutputPath` is not technically checking the `action.OutputPaths` field. In workspace.go > SetTask, we are setting `outputPaths` to _either_ `output_paths` if set, else `concat(output_files, output_directories)`. This is pretty confusing IMO - we might want to refactor that. But for now we could maybe just pass in the `command` proto so that we have access to the original `output_paths` field.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +355,32 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
+				ospb := &repb.OutputSymlink{
+					Path:   trimPathPrefix(fqfn, rootDir),
+					Target: target,
+				}
+				// In case the target is a dangling symlink (the target does not exist, or inaccessible),
+				// there is no way to tell whether it's a directory or a file symlink.
+				// Dangling symlinks could be used in packaging with targets that only exist on the
+				// deployment environments.
+				//
+				// Set both OutputFileSymlinks and OutputDirectorySymlinks only when the symlinks
```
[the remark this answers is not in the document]
sluongng: Support for unresolved symlink was added to Bazel in the recent version (since 6.0.0) https://github.com/bazelbuild/bazel/commit/8056eb566371798acdec97aedde0a2c242957ef7
https://github.com/bazelbuild/bazel/issues/10298

When a symlink is unresolved, we cannot tell whether the target is a file or a directory to add it in the respective file/dir ActionResult field.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +359,32 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
+				// could be resolved.
+				if targetInfo, err := os.Stat(fqfn); err == nil {
+					// OutputFileSymlinks and OutputDirectorySymlinks are deprecated in REAPI v2.2,
```
bduffany, line 375:
```suggestion
					// OutputFileSymlinks and OutputDirectorySymlinks are deprecated in REAPI v2.1,
```

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +359,32 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
+					// but Bazel still uses them as of v6.2.0.
+					//
+					// Set both the deprecated and the new fields to keep backward compatibility
+					// with older REAPI clients.
+					if targetInfo.IsDir() {
+						actionResult.OutputDirectorySymlinks = append(actionResult.OutputDirectorySymlinks, ospb)
+					} else {
+						actionResult.OutputFileSymlinks = append(actionResult.OutputFileSymlinks, ospb)
+					}
```
bduffany, line 384: The documentation of `output_directory_symlinks` says:

```
  // The output directories of the action that are symbolic links to other
  // directories.
```

similarly for `output_file_symlinks`:

```
  // The output files of the action that are symbolic links to other files.
```

I am interpreting this to mean that "the output files of the action" means `Action.command.output_files` and "the output directories of the action" means `Action.command.output_directories`.

If that's correct then under that interpretation, we need to check that this symlink file is exactly referenced in output_files or output_directories before we add them to the ActionResult.output_file_symlinks or ActionResult.output_directory_symlinks (respectively).

(Note that this function begins directory traversal starting from the workspace root, not any of the specific output directories or output file/symlink paths, otherwise we wouldn't need an explicit check.)

enterprise/server/remote_execution/dirtools/dirtools.go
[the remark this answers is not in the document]
bduffany: The current implementation seems fine (as it gives us support for resolved symlinks which we didn't have before), but I am curious how unresolved symlinks can work it we don't set them in the ActionResult.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -31,9 +31,7 @@ import (
 
 const gRPCMaxSize = int64(4000000)
 
-var (
-	enableDownloadCompresssion = flag.Bool("cache.client.enable_download_compression", false, "If true, enable compression of downloads from remote caches")
-)
+var enableDownloadCompresssion = flag.Bool("cache.client.enable_download_compression", false, "If true, enable compression of downloads from remote caches")
```
bduffany, line 34: can you revert the noisy diffs to `var` and `0111 -> 0o111` etc (I guess they are caused by gofumpt vs gofmt or something?)

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -357,6 +359,33 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
+				// OutputFileSymlinks and OutputDirectorySymlinks are deprecated in REAPI v2.1,
+				// but Bazel still uses them as of v6.2.0.
+				//
+				// TODO(sluongng): Set OutputSymlinks when Bazel has support for it.
+				// References:
+				// - https://github.com/bazelbuild/remote-apis/blob/35aee1c4a4250d3df846f7ba3e4a4e66cb014ecd/build/bazel/remote/execution/v2/remote_execution.proto#L1071
+				// - https://github.com/bazelbuild/bazel/pull/18198
+				// - https://github.com/bazelbuild/bazel/pull/18202
+				symlinkPath := trimPathPrefix(fqfn, rootDir)
+				ospb := &repb.OutputSymlink{
```
bduffany, line 372: uber nit: maybe `symlink` instead of `ospb` which feels a little opaque (what does `pb` stand for?)

enterprise/server/remote_execution/dirtools/dirtools_test.go
```diff
@@ -10,18 +11,322 @@ import (
...
+				"fileA.txt": "a",
+			},
+			symlinkPaths: map[string]string{},
+			expectedResult: &repb.ActionResult{
+				OutputFiles: []*repb.OutputFile{
+					{
+						Path: "fileA.txt",
+						Digest: &repb.Digest{
+							SizeBytes: 1,
+							Hash:      "ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb",
```
bduffany, line 67: nit: it might be a little clearer to use `hash.String("a")` from the hash util package

enterprise/server/remote_execution/dirtools/dirtools_test.go
```diff
@@ -10,18 +11,322 @@ import (
...
+			expectedInfo: &dirtools.TransferInfo{
+				FileCount:        1,
+				BytesTransferred: 1,
+			},
+		},
+		{
+			name: "FileAndDirectory",
+			cmd: &repb.Command{
+				OutputFiles:       []string{"a/fileA.txt"},
+				OutputDirectories: []string{"a"},
```
bduffany, line 81: I might be misunderstanding but I think an `OutputDirectory` won't include an `OutputFile` as its child, otherwise the child will get uploaded twice. I think a more representative example would be to have "b" as an output directory, then in the fileContents map below, create a few files fileB.txt and fileC.txt

enterprise/server/remote_execution/dirtools/dirtools_test.go
```diff
@@ -10,18 +11,322 @@ import (
...
+			expectedResult: &repb.ActionResult{
+				OutputDirectories: []*repb.OutputDirectory{
+					{
+						Path: "a",
+						TreeDigest: &repb.Digest{
+							SizeBytes: 85,
+							Hash:      "895545df6841b7efb2e9cc903a4eac7a60c645199be059f6056817ae6feb071d",
+						},
+					},
+				},
```
bduffany, line 99: how is this TreeDigest computed? IMO it's not easy to verify that this is correct just by looking at it - can we maybe have a util func that computes it?

enterprise/server/remote_execution/dirtools/dirtools_test.go
```diff
@@ -10,18 +11,322 @@ import (
...
+				FileCount:        1,
+				BytesTransferred: 1,
+			},
+		},
+		{
+			name: "SymlinkToDirectory",
+			cmd: &repb.Command{
+				OutputFiles: []string{"a/fileA.txt"},
+				OutputDirectories: []string{
+					"a",
```
bduffany, line 157: ditto here - I think in practice we wouldn't see `a/fileA.txt` nested under an output directory. (there are also a few more occurrences of this below)

enterprise/server/remote_execution/dirtools/dirtools.go
sluongng: pb = protobuf... I think I was reusing another variable and never gave it a second thought. 🤣

`symlink` sounds good

enterprise/server/remote_execution/dirtools/dirtools_test.go
```diff
@@ -10,18 +11,322 @@ import (
...
+					},
+				},
+			},
+			expectedInfo: &dirtools.TransferInfo{
+				FileCount:        2,
+				BytesTransferred: 84,
+			},
+		},
+		{
+			name: "DanglingSymlinkIsIgnored",
```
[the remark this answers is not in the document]
sluongng: Added comment.

I think unresolved symlink is only supported as outputs and not as inputs. See https://cs.opensource.google/bazel/bazel/+/master:src/test/java/com/google/devtools/build/lib/remote/RemoteExecutionServiceTest.java;l=216-231?q=RemoteExecution&ss=bazel

Also, the flag is enabled by default since Bazel 6.0.0 so it's not an opt-in feature.

enterprise/server/remote_execution/dirtools/dirtools_test.go
bduffany: Resolving to avoid blocking the PR - might be nice to have a TreeDigest util at some point to make these test assertions a bit clearer.

bduffany (approved):
Looks great!

PR #3802, opened 2023-04-19
Add soci binary to app image

iain-macdonald:
This will be used in https://github.com/buildbuddy-io/buildbuddy/pull/3726

Longer term I would like to just call the code directly and remove this dependency, but for now this is simpler.


vadimberezniker (approved):
approving on the condition that this is something you will remove in the very near future

PR #3812, opened 2023-04-19
parse remote cache/executor options from structured command line

luluz66:
1. Parse remote cache/executor related options in event parser; and add unit tests.
2. Add remote cache/execututor options in the table <> proto conversion
   methods.

**Related issues**:
https://github.com/buildbuddy-io/buildbuddy-internal/issues/2236


server/build_event_protocol/event_parser/event_parser.go
```diff
@@ -32,27 +32,57 @@ const (
 	workflowConfiguredPriority = 5
 )
 
-func parseEnv(commandLine *command_line.CommandLine) map[string]string {
-	envVarMap := make(map[string]string)
+var (
+	optionsToParse = []string{
```
vadimberezniker, line 36: nit: you can turn this into a map for lookups (though in practice the list is fine too)

	optionsToParse = map[string]bool{
		"remote_cache": true,
		...
	}

vadimberezniker (approved):

server/build_event_protocol/event_parser/event_parser.go
luluz66: Done.

PR #3817, opened 2023-04-20
Improve workflow logging

bduffany:
- Add `X-GitHub-Delivery` header to enriched logging context so we can correlate webhook deliveries with logs
- Add some logging to make it more obvious why certain events aren't leading to workflows being started

**Related issues**: N/A


enterprise/server/workflow/service/service.go
```diff
@@ -1049,8 +1048,11 @@ func (ws *workflowService) startWorkflow(ctx context.Context, gitProvider interf
 	if err != nil {
 		return err
 	}
+	var actionsStarted []*config.Action
 	for _, action := range cfg.Actions {
 		if !config.MatchesAnyTrigger(action, wd.EventName, wd.TargetBranch) {
+			jt, _ := json.Marshal(action.Triggers)
+			log.CtxDebugf(ctx, "Action %s not matched, triggers=%+v", action.Name, string(jt))
```
vadimberezniker, line 1055: nit: %s since you're already passing a string

vadimberezniker (approved):

PR #3818, opened 2023-04-20
Move shortcut registration calls from componentWillMount to componentDidMount

iain-macdonald:
This is to fix a bug with shortcuts where a link from a synthetic workflow invocation to its sub-invocations would crash the page due to the Error thrown in duplicate keyboard shortcut detection. This is because when navigating to that link, React creates the invocation component for the sub-invocation and calls `componentWillMount()` before calling `componentWillUnmount()` on the previous, synthetic invocation. However, it doesn't call `componentDidMount()` until the previous component is unmounted, so if we register the shortcuts in there instead then there are no duplicate shortcuts registered.

### DIFF
--- a/app/shortcuts/shortcuts.tsx
+++ b/app/shortcuts/shortcuts.tsx
@@ -111,6 +111,10 @@ export class Shortcuts {
   // no longer invokable. This function will throw an error if the same
   // shortcut is registered multiple times, so if the calling component will
   // unmount and remount, shortcuts should be deregistered and reregistered.
+  //
+  // IMPORTANT: usually you'll want to register shortcuts in componentDidMount
+  // not componentWillMount because componentWillMount can be called when
+  // another component is still mounted, causing shortcut conflicts.
   public register(keyCombo: KeyCombo, action: () => void): string {
     return this.registerSequence([keyCombo], action);
   }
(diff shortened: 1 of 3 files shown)


app/shortcuts/shortcuts.tsx
```diff
@@ -141,9 +149,7 @@ export class Shortcuts {
     let shortcut = new Shortcut(keyCombo, action);
     for (let otherShortcut of this.shortcuts.values()) {
       if (shortcut.collidesWith(otherShortcut)) {
-        console.warn("Duplicate keyboard shortcut registered: " + shortcut.keyCombo);
-        // TODO(iain): throw an error here once bugs are bashed.
-        // throw new Error("Duplicate keyboard shortcut registered: " + shortcut.keyCombo);
+        throw new Error("Duplicate keyboard shortcut registered: " + shortcut.keyCombo);
```
siggisim, line 152: I'd slightly prefer keeping this a `console.warn` and us figuring a way to do some sort of client side error reporting. Don't think crashing the page does a ton of good without any way of displaying to the user what the error is (or reporting it back to us).

siggisim (approved):

app/shortcuts/shortcuts.tsx
iain-macdonald: sg, switched back to console.warn() with a TODO

siggisim (approved):

PR #3821, opened 2023-04-20
Use ANSI quotes for queries and quote table names

tempoz:
The motivation for this is that postgres does not and cannot use backticks to
indicate identifiers such as table names, so if we want to support postgres, we
need to switch to double quotes for that purpose (sql_mode="ANSI_QUOTES" in
mysql). We also need to quote table names, since we use mixed-case table names,
postgres is case sensitive, and unquoted identifiers are all converted to lower
case by postgres. This also means all strings must be specified with single
qoutes (`'`).

**Related issues**: N/A

### DIFF
--- a/server/util/db/db.go
+++ b/server/util/db/db.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -483,6 +484,10 @@ func ParseDatasource(fileResolver fs.FS, datasource string, advancedConfig *Adva
 			return nil, status.FailedPreconditionError("endpoint is required")
 		}
 
+		if ac.Driver == mysqlDriver {
+			dsn.AddParam("sql_mode", "ANSI_QUOTES")
+		}
+
 		if ac.UseAWSIAM {
 			if ac.Region == "" {
 				return nil, status.FailedPreconditionError("region is required to enable AWS IAM")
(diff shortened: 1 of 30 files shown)


tempoz:
* master
  * **PR #3821** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3821" target="_blank">Graphite</a>  👈
    * **PR #3822** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3822" target="_blank">Graphite</a> 
      * **PR #3823** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3823" target="_blank">Graphite</a> 
        * **PR #3824** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3824" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3821?utm_source=stack-comment).

server/util/db/db.go
```diff
@@ -523,6 +528,19 @@ func ParseDatasource(fileResolver fs.FS, datasource string, advancedConfig *Adva
 			return nil, fmt.Errorf("malformed db connection string")
 		}
 		driverName, connString := parts[0], parts[1]
+		switch driverName {
```
tempoz, line 531: This could obviously be an `if` right now, but the postgres driver has different logic here so it eventually is a `switch` anyway, so I just left it.

server/util/query_builder/query_builder_test.go
```diff
@@ -16,7 +16,7 @@ func normalize(t *testing.T, query string) string {
 	// The current logic in this func blindly strips whitespace, which might
 	// change the meaning of queries containing strings. So for now, just fail
 	// the test if we see any string delimiters.
-	if strings.Contains(query, "\"") || strings.Contains(query, "'") || strings.Contains(query, "`") {
+	if strings.Contains(query, "'") || strings.Contains(query, "`") {
```
tempoz, line 19: `"` is now an identifier delimiter instead of a string delimiter and is thus safe so long as none of our identifiers contain whitespace.

vadimberezniker (approved):

PR #3823, opened 2023-04-20
Update table specifications to be compatible with postgres

tempoz:
Removes display widths of ints and removes the `tinyint` specification in order
to make all of the gorm table specifications compatible with the gorm postgres
driver.

**Related issues**: N/A

PR #3822, opened 2023-04-20
Add go dependencies to support postgres

tempoz:
This PR just adds all the deps necessary for us to use the gorm postgres
driver. It is kept separate for ease of readbility.

**Related issues**: N/A

PR #3824, opened 2023-04-20
Add support for postgres db backend

tempoz:
This PR actually adds support for the postgres backend via the gorm postgres
driver.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -783,7 +783,7 @@ func (d *UserDB) GetGroupUsers(ctx context.Context, groupID string, statuses []g
 
 	q := query_builder.NewQuery(`
 			SELECT u.user_id, u.email, u.first_name, u.last_name, ug.membership_status, ug.role
-			FROM "Users" AS u INNER JOIN "UserGroups" AS ug ON u.user_id = ug.user_user_id`)
+			FROM "Users" AS u JOIN "UserGroups" AS ug ON u.user_id = ug.user_user_id`)
 	q = q.AddWhereClause(`ug.group_group_id = ?`, groupID)
 
 	o := query_builder.OrClauses{}
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -783,8 +783,8 @@ func (d *UserDB) GetGroupUsers(ctx context.Context, groupID string, statuses []g
 
 	q := query_builder.NewQuery(`
 			SELECT u.user_id, u.email, u.first_name, u.last_name, ug.membership_status, ug.role
-			FROM "Users" AS u NATURAL JOIN "UserGroups" AS ug`)
-	q = q.AddWhereClause(`u.user_id = ug.user_user_id AND ug.group_group_id = ?`, groupID)
+			FROM "Users" AS u INNER JOIN "UserGroups" AS ug ON u.user_id = ug.user_user_id`)
+	q = q.AddWhereClause(`ug.group_group_id = ?`, groupID)
 
 	o := query_builder.OrClauses{}
 	for _, s := range statuses {
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -783,7 +783,7 @@ func (d *UserDB) GetGroupUsers(ctx context.Context, groupID string, statuses []g
 
 	q := query_builder.NewQuery(`
 			SELECT u.user_id, u.email, u.first_name, u.last_name, ug.membership_status, ug.role
-			FROM "Users" AS u JOIN "UserGroups" AS ug`)
+			FROM "Users" AS u NATURAL JOIN "UserGroups" AS ug`)
 	q = q.AddWhereClause(`u.user_id = ug.user_user_id AND ug.group_group_id = ?`, groupID)
 
 	o := query_builder.OrClauses{}
--- a/enterprise/server/usage/BUILD
+++ b/enterprise/server/usage/BUILD
@@ -49,8 +49,8 @@ go_test(
     name = "usage_test_mysql",
     srcs = ["usage_test.go"],
     args = [
-        "--testenv.use_mysql",
-        "--testmysql.reuse_server",
+        "--testenv.database_type=mysql",
+        "--testenv.reuse_server",
     ],
     exec_properties = {
         "test.workload-isolation-type": "firecracker",
@@ -69,6 +69,39 @@ go_test(
         "//server/testutil/testclock",
         "//server/testutil/testenv",
         "//server/testutil/testmysql",
+        "//server/util/testing/flags",
+        "@com_github_go_redis_redis_v8//:redis",
+        "@com_github_go_sql_driver_mysql//:mysql",
+        "@com_github_stretchr_testify//assert",
+        "@com_github_stretchr_testify//require",
+        "@io_gorm_gorm//:gorm",
+        "@org_golang_x_sync//errgroup",
+    ],
+)
+
+go_test(
+    name = "usage_test_postgres",
+    srcs = ["usage_test.go"],
+    args = [
+        "--testenv.database_type=postgres",
+        "--testenv.reuse_server",
+    ],
+    exec_properties = {
+        "workload-isolation-type": "firecracker",
+        "init-dockerd": "true",
+        "recycle-runner": "true",
+    },
+    shard_count = 7,
+    tags = ["docker"],
+    deps = [
+        ":usage",
+        "//enterprise/server/backends/redis_metrics_collector",
+        "//enterprise/server/testutil/testredis",
+        "//enterprise/server/util/redisutil",
+        "//server/tables",
+        "//server/testutil/testauth",
+        "//server/testutil/testclock",
+        "//server/testutil/testenv",
         "//server/util/testing/flags",
         "@com_github_go_redis_redis_v8//:redis",
         "@com_github_go_sql_driver_mysql//:mysql",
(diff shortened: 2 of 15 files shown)


tempoz:
* master
  * **PR #3821** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3821" target="_blank">Graphite</a> 
    * **PR #3822** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3822" target="_blank">Graphite</a> 
      * **PR #3823** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3823" target="_blank">Graphite</a> 
        * **PR #3824** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3824" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3824?utm_source=stack-comment).

tempoz:
* master
  * **PR #3821** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3821" target="_blank">Graphite</a> 
    * **PR #3822** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3822" target="_blank">Graphite</a> 
      * **PR #3823** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3823" target="_blank">Graphite</a> 
        * **PR #3824** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3824" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3824?utm_source=stack-comment).

bduffany (commented):
not sure if you are already thinking about this, but I think it would be great to have an automated test on CI for this since we won't be actively exercising the postgres support ourselves (for now?)

You can take a look at how `usage_test_mysql` is implemented and see how we run a mysql server using docker (testmysql.go) - we could probably do something pretty similar for postgres, and then just have a small test that runs the GORM auto-migrate logic against an empty postgres DB.

The test would need to have the same `tags` and `exec_properties` as usage_test_mysql to make sure it runs using docker-in-firecracker on CI

server/testutil/testpostgres/testpostgres.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// Wait for the DB to start up.
+	dsn := &url.URL{
+		Scheme: "postgres",
+		User:   url.UserPassword("root", "root"),
+		Host:   net.JoinHostPort("127.0.0.1", strconv.Itoa(port)),
+		Path:   dbName,
+	}
+	// // Ignore errors logged by postgres driver when attempting to ping the DB before it is ready.
+	// gopostgres.SetLogger(&discardLogger{})
```
bduffany, line 105: nit: uncomment?

bduffany (approved):
LGTM, nice :)

would also be good to get a second review from Vadim

server/tables/tables.go
```diff
@@ -866,8 +877,32 @@ func updateInBatches(db *gorm.DB, baseQuery string, batchSize int64) error {
...
+
+	case postgresDialect:
+		updateInvocationStmt = `UPDATE "Invocations" SET invocation_uuid = (REPLACE(invocation_id, '-',''))::uuid WHERE invocation_uuid IS NULL`
+		countPrimaryKeyStmt = `
+			SELECT
+			  COUNT(*) from information_schema.table_constraints
+			WHERE
+				constraint_schema = current_schema() AND
+				constraint_type = 'PRIMARY KEY' AND
+				table_name = 'TargetStatuses'`
```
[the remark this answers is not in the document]
vadimberezniker: +1 don't need this migration for postgres here since we'll never have data w/o the uuid filled

server/util/db/db.go
```diff
@@ -234,6 +239,96 @@ func (m *mysqlDSNFormatter) Clone() dsnFormatter {
...
+}
+
+func (p *postgresDSNFormatter) AddParam(key, val string) {
+	p.params[key] = val
+}
+
+func (p *postgresDSNFormatter) String() string {
+	dsn := []string{}
+	if p.host != "" {
+		dsn = append(dsn, fmt.Sprintf("host = %s", p.host))
```
vadimberezniker, line 287: It looks like the postgres driver supports the connection string in the same format as MySQL and sqlite: https://pkg.go.dev/github.com/jackc/pgx/v5/pgconn#ParseConfig

Would it be simpler if we stuck to this format instead of using something different for postgres?

vadimberezniker (changes requested):

enterprise/server/usage/BUILD
```diff
@@ -78,3 +78,36 @@ go_test(
...
+    name = "usage_test_postgres",
+    srcs = ["usage_test.go"],
+    args = [
+        "--testenv.database_type=postgresql",
+        "--testenv.reuse_server",
+    ],
+    exec_properties = {
+        "workload-isolation-type": "firecracker",
+        "init-dockerd": "true",
+        "recycle-runner": "true",
```
bduffany, line 92: I realized that we could have only the test action run with docker-in-firecracker + runner recycling by applying a `test.` prefix to these. I made the changes in https://github.com/buildbuddy-io/buildbuddy/pull/4055 - mind applying this diff here as well?

```suggestion
        "test.workload-isolation-type": "firecracker",
        "test.init-dockerd": "true",
        "test.recycle-runner": "true",
```

enterprise/server/backends/userdb/userdb.go
```diff
@@ -783,7 +783,7 @@ func (d *UserDB) GetGroupUsers(ctx context.Context, groupID string, statuses []g
 
 	q := query_builder.NewQuery(`
 			SELECT u.user_id, u.email, u.first_name, u.last_name, ug.membership_status, ug.role
-			FROM "Users" AS u JOIN "UserGroups" AS ug`)
+			FROM "Users" AS u NATURAL JOIN "UserGroups" AS ug`)
```
vadimberezniker, line 786: Does postgres not like the query w/o this?

enterprise/server/usage/usage.go
```diff
@@ -271,72 +273,65 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, c collection
...
-				action_cache_hits = action_cache_hits + ?,
-				total_download_size_bytes = total_download_size_bytes + ?,
-				linux_execution_duration_usec = linux_execution_duration_usec + ?,
-				mac_execution_duration_usec = mac_execution_duration_usec + ?
-			WHERE
-				group_id = ?
-				AND period_start_usec = ?
-				AND region = ?
-				AND final_before_usec <= ?
+		return tx.Model(&tables.Usage{}).Where(`
```
vadimberezniker, line 298: Was this necessary because the previous query did not work for postgres?

enterprise/server/backends/userdb/userdb.go
vadimberezniker: Assuming the answer is yes, I would suggest adding an explicit `ON` clause rather than relying on implicit column matching.

enterprise/server/usage/usage.go
tempoz: That's correct, postgres has a very different syntax for ignoring conflicts.

enterprise/server/backends/userdb/userdb.go
tempoz: It does not; I can use an `ON` clause.

enterprise/server/usage/usage.go
tempoz: Oh, you're right, I'm sorry. This change did end up not being necessary and this state is a result of iterative development when fixing this transaction as a whole; I can revert this query.

server/tables/tables.go
```diff
@@ -867,7 +869,23 @@ func updateInBatches(db *gorm.DB, baseQuery string, batchSize int64) error {
 }
 
 func postMigrateInvocationUUIDForMySQL(db *gorm.DB) error {
-	updateInvocationStmt := `UPDATE "Invocations" SET invocation_uuid = UNHEX(REPLACE(invocation_id, "-","")) WHERE invocation_uuid IS NULL`
+	var updateInvocationStmt, countPrimaryKeyStmt string
+	switch db.Dialector.Name() {
```
vadimberezniker, line 873: this function is only called for mysql so there's no need for this switch

server/util/db/db.go
```diff
@@ -578,10 +670,6 @@ func ParseDatasource(fileResolver fs.FS, datasource string, advancedConfig *Adva
 			return nil, err
 		}
 
-		if ac.Endpoint == "" {
```
vadimberezniker, line 581: What's the reason for removing this check?

vadimberezniker (approved):

server/util/db/db.go
tempoz: I think it got moved and then removed from the place it was moved to; I will reinstate it.

enterprise/server/backends/userdb/userdb.go
```diff
@@ -783,8 +783,8 @@ func (d *UserDB) GetGroupUsers(ctx context.Context, groupID string, statuses []g
 
 	q := query_builder.NewQuery(`
 			SELECT u.user_id, u.email, u.first_name, u.last_name, ug.membership_status, ug.role
-			FROM "Users" AS u JOIN "UserGroups" AS ug`)
-	q = q.AddWhereClause(`u.user_id = ug.user_user_id AND ug.group_group_id = ?`, groupID)
+			FROM "Users" AS u INNER JOIN "UserGroups" AS ug ON u.user_id = ug.user_user_id`)
```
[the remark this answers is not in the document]
tempoz: Sure, I'll remove it.

PR #3825, opened 2023-04-20
Fix workflows not checking out the commit they are reporting for

maggie-lou:
Example situation this would fix:

1. Push a commit (C1) to a feature branch that prints "A" in a test
2. The triggered workflow will report commit=C1 and print "A"
3. Push a commit (C2) to the feature branch that prints "B"
4. Rerun the initial workflow
5. It should report commit=C1 and print "A". Currently, it will print "B" because the workflow always pulls the HEAD commit

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -119,16 +119,14 @@ var (
 	commitSHA          = flag.String("commit_sha", "", "Commit SHA to report statuses for.")
 	targetRepoURL      = flag.String("target_repo_url", "", "URL of the target repo.")
 	targetBranch       = flag.String("target_branch", "", "Branch to check action triggers against.")
-	// TODO(Maggie): Clean up this field - consolidate with --commit_sha
-	targetCommitSHA   = flag.String("target_commit_sha", "", "If set, target repo URL is checked out at the given commit instead of the tip of the branch.")
-	workflowID        = flag.String("workflow_id", "", "ID of the workflow associated with this CI run.")
-	actionName        = flag.String("action_name", "", "If set, run the specified action and *only* that action, ignoring trigger conditions.")
-	invocationID      = flag.String("invocation_id", "", "If set, use the specified invocation ID for the workflow action. Ignored if action_name is not set.")
-	visibility        = flag.String("visibility", "", "If set, use the specified value for VISIBILITY build metadata for the workflow invocation.")
-	bazelSubCommand   = flag.String("bazel_sub_command", "", "If set, run the bazel command specified by these args and ignore all triggering and configured actions.")
-	patchURIs         = flagutil.New("patch_uri", []string{}, "URIs of patches to apply to the repo after checkout. Can be specified multiple times to apply multiple patches.")
-	recordRunMetadata = flag.Bool("record_run_metadata", false, "Instead of running a target, extract metadata about it and report it in the build event stream.")
-	gitCleanExclude   = flagutil.New("git_clean_exclude", []string{}, "Directories to exclude from `git clean` while setting up the repo.")
+	workflowID         = flag.String("workflow_id", "", "ID of the workflow associated with this CI run.")
+	actionName         = flag.String("action_name", "", "If set, run the specified action and *only* that action, ignoring trigger conditions.")
+	invocationID       = flag.String("invocation_id", "", "If set, use the specified invocation ID for the workflow action. Ignored if action_name is not set.")
+	visibility         = flag.String("visibility", "", "If set, use the specified value for VISIBILITY build metadata for the workflow invocation.")
+	bazelSubCommand    = flag.String("bazel_sub_command", "", "If set, run the bazel command specified by these args and ignore all triggering and configured actions.")
+	patchURIs          = flagutil.New("patch_uri", []string{}, "URIs of patches to apply to the repo after checkout. Can be specified multiple times to apply multiple patches.")
+	recordRunMetadata  = flag.Bool("record_run_metadata", false, "Instead of running a target, extract metadata about it and report it in the build event stream.")
+	gitCleanExclude    = flagutil.New("git_clean_exclude", []string{}, "Directories to exclude from `git clean` while setting up the repo.")
 
 	shutdownAndExit = flag.Bool("shutdown_and_exit", false, "If set, runs bazel shutdown with the configured bazel_command, and exits. No other commands are run.")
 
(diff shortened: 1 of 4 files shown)


maggie-lou:
* master
  * **PR #3825** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3825" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3825?utm_source=stack-comment).

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1415,11 +1422,19 @@ func (ws *workspace) sync(ctx context.Context) error {
...
+
+	// Checkout the git branch
+	if checkoutRef != "" {
+		if err := git(ctx, ws.log, "checkout", "--force", checkoutRef); err != nil {
+			return err
+		}
+	}
+
+	// Checkout a specific commit on the branch and create the local branch if it doesn't already exist
+	if err := git(ctx, ws.log, "checkout", "--force", "-B", checkoutLocalBranchName, checkoutCommit); err != nil {
```
sluongng, line 1434: So when `checkoutRef != ""` that means `checkoutRef ~= pushedBranch || targetBranch`

In those cases, `checkoutLocalBranchName == pushedBranch || targetBranch`

And thus, `checkoutRef ~= checkoutLocalBranchName`

Should we simplify this then?

```suggestion
	// Checkout a specific commit on the branch and create the local branch if it doesn't already exist
	if err := git(ctx, ws.log, "checkout", "--force", "-B", checkoutLocalBranchName, checkoutCommit); err != nil {
```

---

Side note: 

We are using a sparse clone repo where objects are fetched remotely on git-checkout. So I am quite cautious in introducing new `git-checkout` calls. If we could make do with 1 checkout, it would be much preferably.

In the future, we might be able to make do without a sparse-clone by caching the git-bundle in our executor file cache(1).

(1): https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---bundle-urilturigt

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -119,15 +119,16 @@ var (
...
+	workflowID        = flag.String("workflow_id", "", "ID of the workflow associated with this CI run.")
+	actionName        = flag.String("action_name", "", "If set, run the specified action and *only* that action, ignoring trigger conditions.")
+	invocationID      = flag.String("invocation_id", "", "If set, use the specified invocation ID for the workflow action. Ignored if action_name is not set.")
+	visibility        = flag.String("visibility", "", "If set, use the specified value for VISIBILITY build metadata for the workflow invocation.")
+	bazelSubCommand   = flag.String("bazel_sub_command", "", "If set, run the bazel command specified by these args and ignore all triggering and configured actions.")
+	patchURIs         = flagutil.New("patch_uri", []string{}, "URIs of patches to apply to the repo after checkout. Can be specified multiple times to apply multiple patches.")
+	recordRunMetadata = flag.Bool("record_run_metadata", false, "Instead of running a target, extract metadata about it and report it in the build event stream.")
+	gitCleanExclude   = flagutil.New("git_clean_exclude", []string{}, "Directories to exclude from `git clean` while setting up the repo.")
```
sluongng, line 131: nit: perhaps moving the flag down into a separate block would avoid this noisy formatting diff hunk. It would also avoid future noisy hunk when you remove the flag and the indentation reset back.

enterprise/server/cmd/ci_runner/main.go
bduffany: I think at a high level, the changes we want to make are:

* `hostedrunner.go` should stop passing `target_commit_sha` and instead start passing both `commit_sha` (optional) and `target_branch` (required). This is because we currently don't have a way to fetch a commit sha from the remote (we only fetch branches currently), so I think that effectively means that `--target_branch` is always required. If you look at how `ws.fetch()` is called, it takes in a list of branches and is a NOP if no branches are passed in. So that means we don't fetch anything unless we have at least some branch name to go on.

Maybe @vadimberezniker can double-check my understanding here - there is a codepath in `hostedrunner.go` that only sets `--target_commit_sha` without setting `--pushed_branch` or `--target_branch`, but I am not sure if this codepath works, because it looks like the ci_runner only ever tries to fetch the pushed_branch or target_branch refs.

* When checking out, we can do the following:

  * If `commit_sha` is not set, then just `git checkout --force -B <target_branch>`
  * If `commit_sha` is set, then additionally reset the checked out branch to point to `commit_sha`, by just appending the commit SHA argument: `git checkout --force -B <target_branch> <commit_sha>`.

We can remove the logic that was setting the local branch name to "local", if my reasoning above is correct re. the branch name effectively being required.

maggie-lou: It works if we don't pass a `target_branch` because we `git clone` [here](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/cmd/ci_runner/main.go#L1329). So by default you stay on the master branch, even if you don't fetch any additional branches

maggie-lou: 😅 We definitely need a cleaner way to specify valid combos of parameters

maggie-lou: @sluongng @bduffany I was thinking of an edge case where the same commit is on master and the feature branch, but with different commit history, so just checking out the commit is insufficient info to get the correct git tree

Ex. 
Master: (HEAD) C1 - C2 - C3
Feature:  (HEAD) C2 - C1 - C4 - C3

Request: Checkout feature branch commit C1

If we set checkoutRef = C1, we will checkout C1 on Master branch. This will include C2. However if we had checked out C1 on the Feature branch, it would not include C2 and would include C4.

This is why I thought we might need 2 check outs - the first to get on the correct branch, and the second to checkout the commit on that branch. Does my understanding of this match what you guys are thinking?

vadimberezniker: I don't have a perfect understanding of Git, but I think the above situation is not possible. The git object commit (over which the sha is computed) includes the parent commit so even if you have the same exact change on two branches, the commit sha will be different due to having different parent shas.

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1392,14 +1393,20 @@ func (ws *workspace) sync(ctx context.Context) error {
...
-		} else {
-			checkoutRef = *targetCommitSHA
-			checkoutLocalBranchName = "local"
-		}
+		// If no branch is passed in, stay on the default branch. Don't set checkoutRef
+		checkoutLocalBranchName = "local"
+	}
+
+	// TODO(Maggie): If commit sha is not set, pull the actual sha so that it can be used in reporting
+	checkoutCommit := "HEAD"
```
bduffany: I think `HEAD` is not the right default value - HEAD points to the currently checked out commit, while we actually want "the remote branch tip of the commit we are checking out".

So I think the start point should be `origin/{*targetBranch}` if the repo is not a fork, else `fork/{*targetBranch}`.

You can use this script to see this behavior: https://gist.github.com/bduffany/da1ef7d857279e521a8dd14af199aeda

maggie-lou: Not relevant anymore now that hostedrunner always passes a target branch

maggie-lou:
@bduffany Sorry PR was in an in-between state for a bit. Addressed all the feedback and now it's ready for review

PR #3829, opened 2023-04-21
Fix incorrect request type

bduffany:
This shows up as an error in my IDE but doesn't fail the bazel typecheck test for some reason :thinking: 

**Related issues**: N/A


iain-macdonald (approved):
Weird that this doesn't fail the bazel type checker!

PR #3831, opened 2023-04-21
rules_go: simplified Go SDK config and provide a go binary alias

sluongng:
Simplified our Go SDK config and provide some context doc on why we need
to set it up the way it is. Upgrading Go version should be relatively
simpler after this.

Also provide an alias to the "go" binary in the go_sdk so we don't need
to remember what platform we are on while calling it.
It's useful to run `go fmt` or `go mod` commands using the exact Go
version Bazel use.

**Related issues**: N/A


BUILD
```diff
@@ -5,6 +5,20 @@ load("//rules/go:index.bzl", "go_sdk_tool")
...
+# binary for the current platform.
+alias(
+    name = "go_binary",
+    actual = select(
+        {
+            "@bazel_tools//src/conditions:darwin_arm64": "@go_sdk_darwin_arm64//:bin/go",
+            "@bazel_tools//src/conditions:darwin_x86_64": "@go_sdk_darwin_amd64//:bin/go",
+            "@bazel_tools//src/conditions:linux_aarch64": "@go_sdk_linux_arm64//:bin/go",
+            "@bazel_tools//src/conditions:linux_x86_64": "@go_sdk_linux_amd64//:bin/go",
+        },
```
sluongng, line 18: This could also be written as

```
        {
            "@bazel_tools//src/conditions:{}_{}".format(
                k,
                "x86_64" if v == "amd64" else "aarch64" if k == "linux" else v,
            ): "@go_sdk_{}_{}//:bin/go".format(k, v)
            for k, v in [
                (goos, goarch)
                for goos in [
                    "darwin",
                    "linux",
                ]
                for goarch in [
                    "amd64",
                    "aarch64",
                ]
            ]
        },
```

But let's not make our life harder

sluongng: Oh neat!

Do you want to merge that separately or should I pick out those hunks into this PR?

sluongng: PTAL

tempoz (approved):
LGTM

PR #3834, opened 2023-04-21
Fix re-running workflows for non-selected but authorized group

bduffany:
Fixes an issue that workflows cannot be re-run when the org that owns the workflow isn't selected, even if the authenticated user has access to the workflow.

**Related issues**: Fixed https://github.com/buildbuddy-io/buildbuddy-internal/issues/2246

### DIFF
--- a/enterprise/server/githubapp/githubapp.go
+++ b/enterprise/server/githubapp/githubapp.go
@@ -234,6 +234,34 @@ func (a *GitHubApp) GetInstallationToken(ctx context.Context, owner string) (str
 	return tok.GetToken(), nil
 }
 
+func (a *GitHubApp) GetRepositoryInstallationToken(ctx context.Context, repo *tables.GitRepository) (string, error) {
+	if err := perms.AuthorizeGroupAccess(ctx, a.env, repo.GroupID); err != nil {
+		return "", err
+	}
+	repoURL, err := gitutil.ParseGitHubRepoURL(repo.RepoURL)
+	if err != nil {
+		return "", err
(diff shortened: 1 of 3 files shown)


server/interfaces/interfaces.go
```diff
@@ -489,6 +489,10 @@ type GitHubApp interface {
 	// owner (GitHub username or org name).
 	GetInstallationToken(ctx context.Context, owner string) (string, error)
 
+	// GetRepositoryInstallationToken returns an installation for the given
```
maggie-lou, line 492: nit: "installation token"

maggie-lou (approved):

server/interfaces/interfaces.go
bduffany: done

enterprise/server/githubapp/githubapp.go
```diff
@@ -234,6 +234,34 @@ func (a *GitHubApp) GetInstallationToken(ctx context.Context, owner string) (str
...
+		return "", err
+	}
+	repoURL, err := gitutil.ParseGitHubRepoURL(repo.RepoURL)
+	if err != nil {
+		return "", err
+	}
+	var installation tables.GitHubAppInstallation
+	err = a.env.GetDBHandle().DB(ctx).Raw(`
+		SELECT *
+		FROM GitHubAppInstallations
```
vadimberezniker, line 248: quote the table name with double quotes.

vadimberezniker (approved):

PR #3835, opened 2023-04-21
Don't trigger keyboard shortcuts when typing into a text area

iain-macdonald:
Maggie noticed this when typing her name into a text box somewhere. I also added logic to reset the offset counters when shortcuts are enabled/disabled and in keypress when a text box is selected so that shortcut sequences across those events don't trigger the shortcut.

### DIFF
--- a/app/shortcuts/shortcuts.tsx
+++ b/app/shortcuts/shortcuts.tsx
@@ -144,7 +144,7 @@ export class Shortcuts {
           }
           // Don't run when typing into a text box.
           let activeElement = document.activeElement;
-          if (activeElement.tagName === "INPUT" && activeElement.type === "text") {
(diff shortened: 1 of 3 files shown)


app/shortcuts/shortcuts.tsx
```diff
@@ -138,6 +142,12 @@ export class Shortcuts {
           if (!this.preferences.keyboardShortcutsEnabled) {
             return;
           }
+          // Don't run when typing into a text box.
+          let activeElement = document.activeElement;
+          if (activeElement.tagName === "INPUT" && activeElement.type === "text") {
```
bduffany, line 147: nit: can handle textarea here too (we use those e.g. in the secrets page)

```suggestion
          if ((activeElement.tagName === "INPUT" && activeElement.type === "text") || activeElement.tagName === "TEXTAREA") {
```

bduffany (approved):

bduffany (approved):

PR #3836, opened 2023-04-21
Revert "Switch from protobufjs-cli to protoc-gen-protobufjs (#3830)"

bduffany:
This seems to break the internal build (getting an error that an action isn't generating some expected outputs - probably it's not computing the right paths when building the protos from the external OSS repo)

**Related issues**: N/A


jdhollen (approved):

bduffany:
I wound up fixing forward in https://github.com/buildbuddy-io/buildbuddy-internal/pull/2251, so closing this out for now.

PR #3840, opened 2023-04-24
Fix bug that filecache overwrite deletes existing file

bduffany:
**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2188


sluongng (commented):
Tested on my personal executor and seems to work like a charm 👍

vadimberezniker (approved):

PR #3842, opened 2023-04-24
Update recently added encryption queries to new standard.

vadimberezniker:
Also a fix a couple of other queries that I noticed have not been updated.

**Related issues**: N/A


tempoz (approved):
LGTM

Thanks for catching those others, too!

Issue #3838, opened 2023-04-21
Feature Request : Link to line in logs


karanjamutahi:
I'd like to be able to get a link to a line in the logs of a bazel invocation. 
Bonus if we can get a link to mutliline selection

siggisim:
Hey @karanjamutahi - that's a great idea! Will take a look at what it'd take to implement and keep you posted.

siggisim:
Hey @karanjamutahi - added support for this (single line selection and linking) in https://github.com/buildbuddy-io/buildbuddy/pull/3843

This should go out in this week's release (typically on Wednesdays). Should be able to get multi-line working at some point in the future as well.

Thanks again for the great feature request!

PR #3843, opened 2023-04-24
Add support for linking to an individual build log line

siggisim:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy/issues/3838


app/router/router.tsx
```diff
@@ -327,6 +327,18 @@ class Router {
...
+    return window.location.hash.split("@")[0];
+  }
+
+  getLineNumber() {
+    let hashParts = location.hash.split("@");
+    if (hashParts.length > 1) {
+      return parseInt(hashParts[1]);
+    }
+    return undefined;
+  }
```
bduffany, line 362: I feel like it's a bit error prone that `window.location.hash` no longer can be trusted, and `router.getHash()` has to be used instead.

wdyt about using a URL param instead? like `?logLine=N`

siggisim: I agree, but I'm conflicted - I almost feel like the selected tab should be in the url since it's something the server might care about (i.e for prefetching and whatever) but the selected line feels like something that truly only the client should care about (which is what belongs in the hash since it's not sent to the server).

siggisim: Would it make more sense if I renamed `getHash()` => `getTab()` and plumbed through "tab" everywhere instead of hash?

bduffany: Yeah, hiding the "hash" terminology is probably a good idea to prevent confusion.

siggisim: Yeah, it's annoying because this is like the one thing that I'm pretty sure belongs in the hash lol. Going to update the hash terminology to tab - I think that prevents a lot of confusion (then you know if you want to convert a hash to a tab, you need to strip a potential line number).

app/terminal/terminal.tsx
```diff
@@ -263,6 +264,12 @@ export default class TerminalComponent extends React.Component<TerminalProps, St
   }
 
   private scrollToEnd() {
+    let lineNumber = router.getLineNumber();
+    if (lineNumber) {
+      this.scrollToRow(lineNumber);
+      return;
+    }
```
bduffany, line 271: nit: I feel like it might be a decent idea to add line numbers to the logs at some point. Line numbers are usually 1-indexed, so wdyt about 1-indexing these instead of 0-indexing so that the URL param matches the selected line in that case?

bduffany (approved):
lgtm if we rename getHash() => getTab()

bduffany (approved):

PR #3846, opened 2023-04-25
Upgrade dependencies using "npm audit fix"

sluongng:
Ran without --force

Will upgrade dependencies with breaking changes via separate commits

**Related issues**: N/A


bduffany (approved):
Checked trends page locally; LGTM. Thanks for doing this!

PR #3852, opened 2023-04-25
Add error caching to crypter service.

vadimberezniker:
The crypter sevice is in the hot path of the cache. If there's a transient error either with the database or with the external KMS, we should avoid spamming the DB and/or the KMS service with more queries than necessary.

Note that we would only cache the error after we have exhausted multiple retries so we'd only run into this peoblem if the issue lasted >30 seconds.

**Related issues**: N/A


maggie-lou (approved):
LGTM - maybe add a small unit test on this?

bduffany (approved):

vadimberezniker:
> LGTM - maybe add a small unit test on this?

Added test

PR #3854, opened 2023-04-26
Revert "Add soci binary to app image"

iain-macdonald:
This approach ended up requiring pulling a bunch of containerd stuff into the binary as well and turned out to be more effort than just inlining the index creation code, so doing that instead.

Reverts buildbuddy-io/buildbuddy#3802


vadimberezniker (approved):
:heart:

PR #3856, opened 2023-04-26
Make the app generate soci artifacts directly

iain-macdonald:
The current implementation of the sociartifactstore uses containerd to pull images and the soci tool binary to index these images. This was a bit of a headache to get working in dev because it requires adding those binaries to the app's container, upgrade the base container image (TBD if this is still necessary), and do other stuff to get containerd working. I gave up before getting it all the way working, so I don't know what other steps were needed. Generating the ZToCs and SOCI index isn't terribly complicated, so this PR puts that logic directly in the sociartifactstore instead. Now, if sociartifacestore receives a request for artifacts for an image it doesn't have, it will pull the image, generate ZToCs for each layer, and generate the SOCI index wrapping those ZToCs on completion. All of these generate artifacts are written to the cache and the digests are returned to the caller.

This logic is a bit more complicated than before, so I filed [an issue](https://github.com/buildbuddy-io/buildbuddy-internal/issues/2262) to add some tests before running this in production.

**Related issues**: N/A


enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -174,65 +202,79 @@ func getTargetImageRef(ctx context.Context, image string, platform *rgpb.Platfor
...
-		return nil, err
-	}
-	if err = s.writeArtifactsToCache(ctx, imageRefDigest, sociIndexDigest, ztocDigests); err != nil {
+	configHashHex := configHash.Hex
+
+	// Try to only read-pull-index-write once at a time to prevent hammering
+	// the containter registry with a ton of parallel pull requests, and save
+	// apps a bunch of parallel work.
+	workKey := fmt.Sprintf("soci-artifact-store-image-%s", configHashHex)
+	respBytes, err := s.deduper.Do(ctx, workKey, func() ([]byte, error) {
```
vadimberezniker, line 243: you're ignoring this error

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -285,112 +327,196 @@ func deserializeDigest(s string) (*repb.Digest, error) {
...
+	if err != nil {
+		if t, ok := err.(*transport.Error); ok && t.StatusCode == http.StatusUnauthorized {
+			return nil, status.PermissionDeniedErrorf("could not retrieve image manifest: %s", err)
+		}
+		return nil, status.UnavailableErrorf("could not retrieve manifest from remote: %s", err)
+	}
+
+	switch remoteDesc.MediaType {
+	case types.OCIImageIndex, types.DockerManifestList:
+		return nil, status.InvalidArgumentErrorf("image index not expected in conversion request")
```
vadimberezniker, line 364: `getTargetImageRef` was responsible for checking if the target ref was an image index and then finding a single image to use. you're now passing the image directly to this function which could end up being an image index I believe?

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -285,112 +327,196 @@ func deserializeDigest(s string) (*repb.Digest, error) {
...
-	if err := pullImage(ctx, imageName); err != nil {
+	image, err := pullImage(ctx, imageName, credentials)
+	if err != nil {
 		return nil, nil, err
 	}
-	if err := runSoci(ctx, imageName); err != nil {
+	return s.indexImage(ctx, image)
+}
+
+func pullImage(ctx context.Context, image string, credentials *rgpb.Credentials) (v1.Image, error) {
```
vadimberezniker, line 339: this function is looking up the image descriptor, it's not pulling the image contents so I'd call it something else

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -174,65 +202,79 @@ func getTargetImageRef(ctx context.Context, image string, platform *rgpb.Platfor
...
+	// apps a bunch of parallel work.
+	workKey := fmt.Sprintf("soci-artifact-store-image-%s", configHashHex)
+	respBytes, err := s.deduper.Do(ctx, workKey, func() ([]byte, error) {
+		exists, err := s.blobstore.BlobExists(ctx, blobKey(configHashHex))
+		if err != nil {
+			return nil, err
+		}
+		var resp *socipb.GetArtifactsResponse
+		if exists {
+			resp, err = s.getArtifactsFromCache(ctx, configHashHex)
```
vadimberezniker, line 250: I think I missed this last time around, but you need to deal with cache eviction. If the artifacts are not in the cache anymore, you should trigger a reindexing.

vadimberezniker (changes requested):

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Whoops, fixed

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Ahh good call, added a check for the ztocs and restructured the code a bit.

There's still technically a race here wherein one (or more) artifacts could be evicted from the cache after returning from here but before the executor reads them. It's not a long window of time, maybe a few seconds, but it is possible. I don't think it's worth trying to address that now but LMK what you think.

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Renamed to getTargetImageInfo and changed to return the target image as a ctrname.Digest per your comment below.

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Ahh good call, returned to the caller. LMK if you can think of a cleaner way to manage these different components of the image.

iain-macdonald (commented):
Thanks, PTAL

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -285,112 +344,191 @@ func deserializeDigest(s string) (*repb.Digest, error) {
...
 	if err != nil {
-		return err
+		return nil, status.NotFoundErrorf("could not determine layer compression: %s", err)
+	}
+	if compressionAlgo != compression.Gzip {
+		return nil, status.UnimplementedErrorf("layer %s (%s) cannot be indexed because it is compressed with %s",
+			layerDigest.Hex, mediaType, compressionAlgo)
+	}
+
+	layerTmpFile, err := os.CreateTemp("", "layer.*")
```
vadimberezniker, line 499: We should add a GKE volume (can use `emptyDir`) for this scratch space since layers can be large. The space on the host node is not unlimited and w/o setting up your own storage, GKE won't know how much you will need. 
You should change this code to take a flag to point to a scratch directory and then separately add a volume to the GKE deployment.

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -285,112 +344,191 @@ func deserializeDigest(s string) (*repb.Digest, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	ztocDigest := repb.Digest{
+		Hash:      ztocDesc.Digest.Encoded(),
+		SizeBytes: ztocDesc.Size,
 	}
-	return s.cache.Set(ctx, resourceName(digest), bytes)
+	ztocBytes, err := ioutil.ReadAll(ztocReader)
+	s.cache.Set(ctx, resourceName(&ztocDigest), ztocBytes)
```
vadimberezniker, line 530: the cache interface has a Writer method that doesn't require providing the full buffer in one go.

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -4,52 +4,79 @@
...
 
-	// The name of the soci artifact database file
-	sociDbPath = "/var/lib/soci-snapshotter-grpc/artifacts.db"
+	// The minimum layer size for generating ZToCs. Layers smaller than this won't be
+	ztocMinLayerSize = 10 << 20 // about 10MB
+
+	// The SOCI Index build tool. !!! WARNING !!! This is embedded in both the
+	// ZToCs and SOCI index, so changing it will invalidate all previously
+	// generated and stored soci artifacts, forcing a re-pull and re-index.
+	buildToolIdentifier = "AWS SOCI CLI v0.1"
```
vadimberezniker, line 63: Should we change this to our own identifier?

vadimberezniker (approved):

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -285,112 +344,191 @@ func deserializeDigest(s string) (*repb.Digest, error) {
...
+
+func (s *SociArtifactStore) indexImage(ctx context.Context, image v1.Image) (*repb.Digest, []*repb.Digest, error) {
+	layers, err := image.Layers()
+	if err != nil {
 		return nil, nil, err
 	}
-	sociIndex, err := findSociIndex(ctx, imageRef)
+	ztocDigests := []*repb.Digest{}
+	ztocDescriptors := []ocispec.Descriptor{}
+	for _, layer := range layers {
```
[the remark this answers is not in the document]
iain-macdonald: Yup, filed a bug and added a TODO

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Sounds good, done, with another PR coming for the new volume

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Good call, done

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Yup, switched to "BuildBuddy SOCI Artifact Store v0.1"

PR #3858, opened 2023-04-27
d3-scale: upgrade to 4.0.2

sluongng:
This remove the last dependency we have over the old d3-color version.

**Related issues**: Closes https://github.com/buildbuddy-io/buildbuddy/security/dependabot/71


sluongng:
I think the old place that is using d3-scale right now is the heatmap page. @jdhollen could you help me review 🙏

jdhollen:
looks fine, thanks for handling this.

PR #3861, opened 2023-04-27
dirtools: add support for output_symlinks

sluongng:
This is a follow-up of https://github.com/buildbuddy-io/buildbuddy/pull/3800

Implement support for ActionResult.output_symlinks by checking usages of
Command.output_paths.

Added tests to verify different cases, including a hybrid case where Bazel
could be using both the new `Command.output_paths` the old fields
`Command.output_files` and `Command.output_directories`.

Added tests to verify different dangling symlink cases.

**Related issues**: N/A


vadimberezniker:
> Note: We are not in a rush to merge this, especially since Bazel only started supporting the newer Remote API version last week.
> 
> Thus currently the support for `output_symlinks` would only impact folks who use Bazel@HEAD.

Since you're not in a rush, I will take a look at this tomorrow. I'm trying to wrap something up today.

vadimberezniker (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -361,36 +361,62 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
-					if symlinkPath == expectedFile {
-						actionResult.OutputFileSymlinks = append(actionResult.OutputFileSymlinks, symlink)
-						break
+
+				// REAPI specification:
+				//   `output_symlinks` will only be populated if the command `output_paths` field
+				//   was used, and not the pre v2.1 `output_files` or `output_directories` fields.
+				//
+				// Check whether the current client is using REAPI version before or after v2.1.
+				if len(cmd.OutputPaths) > 0 && len(cmd.OutputFiles) == 0 && len(cmd.OutputDirectories) == 0 {
```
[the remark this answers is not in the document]
sluongng: Good call. I moved it out.

PTAL

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -289,6 +289,76 @@ func uploadFiles(uploader *cachetools.BatchCASUploader, fc interfaces.FileCache,
...
+	})
+
+	// REAPI specification:
+	//   `output_symlinks` will only be populated if the command `output_paths` field
+	//   was used, and not the pre v2.1 `output_files` or `output_directories` fields.
+	//
+	// Check whether the current client is using REAPI version before or after v2.1.
+	if len(cmd.OutputPaths) > 0 && len(cmd.OutputFiles) == 0 && len(cmd.OutputDirectories) == 0 {
+		// REAPI >= v2.1
+		if dirHelper.IsOutputPath(fqfn) {
```
vadimberezniker, line 315: You can flip this to `!dirHelper.IsOutputPath(fqfn)` and perform an early return.

vadimberezniker (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -289,6 +289,86 @@ func uploadFiles(uploader *cachetools.BatchCASUploader, fc interfaces.FileCache,
...
+func handleSymlink(dirHelper *DirHelper, rootDir string, cmd *repb.Command, actionResult *repb.ActionResult, directory *repb.Directory, fqfn string) error {
+	target, err := os.Readlink(fqfn)
+	if err != nil {
+		return err
+	}
+	symlink := &repb.OutputSymlink{
+		Path:   trimPathPrefix(fqfn, rootDir),
+		Target: target,
+	}
+	foundSymlink := false
```
vadimberezniker, line 303: I'm generally not a fan of using defers in this way because it's not always obvious that non-trivial code is being executed at the end of the function if you don't read the code carefully enough. You could instead define an inline function and call that instead of setting foundSymlink, e.g.

addSymlinkToDir := func() { ... }

sluongng: Great suggestion! It looks much cleaner now. 👍

PR #3869, opened 2023-04-28
tags schema take two.

jdhollen:
Turns out the gorm clickhouse driver magically supports string slices as columns, and i had simply put myself into a bad state by trying to define a non-slice column first--gorm was trying to pull data from a dead column in my by DB.  Oh well.  Back to work on this now.

**Related issues**: N/A

### DIFF
--- a/server/util/clickhouse/schema/schema.go
+++ b/server/util/clickhouse/schema/schema.go
@@ -8,9 +8,10 @@ import (
 	"strings"
 
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	// "github.com/buildbuddy-io/buildbuddy/server/util/clickhouse/array"
+	// "github.com/buildbuddy-io/buildbuddy/server/util/clickhouse/array"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
-	"github.com/lib/pq"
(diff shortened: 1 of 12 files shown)


vadimberezniker (approved):
Have you already done some performance testing to make sure querying invocations by tag using this schema is performant? Only asking because I don't know enough about Clickhouse to know the answer from just looking at this PR.

luluz66 (approved):

server/util/clickhouse/schema/schema.go
```diff
@@ -378,6 +379,16 @@ func RunMigrations(gdb *gorm.DB) error {
...
 
 func ToInvocationFromPrimaryDB(ti *tables.Invocation) *Invocation {
+	// A temporary hack to avoid writing arrays with empty strings to the DB while
+	// allowing tests with real values to pass.  This will be fixed shortly to
+	// actually trim and format array entries.
+	var tags []string
+	if len(ti.Tags) > 0 {
+		tags = strings.Split(ti.Tags, ",")
+	} else {
+		tags = make([]string, 0)
```
luluz66, line 389: nit: you can just do []string{}?

luluz66 (approved):

server/util/clickhouse/schema/schema.go
jdhollen: sure why not

PR #3871, opened 2023-04-28
Use gRPC headers for some remote bazel env vars

bduffany:
**Related issues**: N/A


vadimberezniker:
> What is the impact of this change?

I'm guessing this is to avoid cache misses due to changing env variables.

vadimberezniker (approved):

PR #3872, opened 2023-04-28
CPU time chart

luluz66:
Screenshot from 2023-04-28 09-56-14

**Related issues**: N/A

### DIFF
--- a/enterprise/app/trends/trends.tsx
+++ b/enterprise/app/trends/trends.tsx
@@ -354,17 +354,19 @@ export default class TrendsComponent extends React.Component<Props, State> {
                 separateAxis={true}
               />
 
-              <TrendsChartComponent
-                title="Saved CPU Time"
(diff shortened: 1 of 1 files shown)


enterprise/app/trends/trends.tsx
```diff
@@ -354,6 +354,18 @@ export default class TrendsComponent extends React.Component<Props, State> {
                 separateAxis={true}
               />
 
+              <TrendsChartComponent
```
jdhollen, line 357: please condition this on `capabilities.config.trendsSummaryEnabled` so that we push both of these when the data is actually available (so, wrap this in `{capabilities.config.trendsSummaryEnabled && (<TrendsChartComponent...>)`

jdhollen (changes requested):

jdhollen (approved):

PR #3873, opened 2023-04-28
fix default visibility

luluz66:
make ./buildfix.sh happy
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/testworker/BUILD
+++ b/enterprise/server/remote_execution/runner/testworker/BUILD
@@ -1,8 +1,7 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
-# gazelle:default_visibility //enterprise:__subpackages__,//enterprise/server/remote_execution/runner:__subpackages__
+# gazelle:default_visibility //enterprise/server/remote_execution/runner:__subpackages__
 package(default_visibility = [
-    "//enterprise:__subpackages__",
     "//enterprise/server/remote_execution/runner:__subpackages__",
 ])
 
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/runner/testworker/BUILD
```diff
@@ -1,6 +1,6 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
-# gazelle:default_visibility //enterprise/server/remote_execution/runner:__subpackages__
+# gazelle:default_visibility //enterprise:__subpackages__,//enterprise/server/remote_execution/runner:__subpackages__
```
jdhollen, line 3: this one and all the others below seem weird--presumably we either don't need this visibility or could just add a second package here?  what did gazelle want to update the visiblity to?

my understanding (which admittedly could be completely wrong) is that we should only need to hack this for the `@buildbuddy_internal` rule above, since gazelle can't know about external repos depending on the package.

jdhollen (changes requested):

enterprise/server/remote_execution/runner/testworker/BUILD
luluz66: gazelle changed to +package(default_visibility = ["//enterprise/server/webhooks/bitbucket:__subpackages__"]). 

@sluongng added the //enterprise:__subpackages__ in this PR https://github.com/buildbuddy-io/buildbuddy/pull/3480

jdhollen: This should either only need to be //enterprise:__subpackages__ or just the two packages that depend on it--if gazelle is recommending removing the remote_execution one then we should probably just stick with enterprise/subpackages. But in any case, we should be able to just fix these because they are internal to the repo and gazelle should understand it all.

luluz66: I removed //enterpirse:subpackages

jdhollen (approved):

PR #3888, opened 2023-05-01
Add new pebble key format that includes the encryption key ID.

vadimberezniker:
This format adds an optional encryption key segment for keys that refer to encrypted contents.

This version also makes the "remote instance hash" segment always be present even if the remote instance name is empty. Instead of omitting the segment, we use a "0" value. This simplifies the parsing logic as the encryption key is also optional.

Adding the encryption key ID to the key makes it simple to deal with transitions from having encryption enabled to off, or even if the encryption is turned off and on again.

For example, if a user has encryption enabled and then turns it off, we want previously written encrypted artifacts to be treated as non-existent. Without having the encryption metadata in the key, we would have to unmarshal and check every value during FindMissing calls which would affect the performance and complexity of the implementation. Adding the encryption key ID to the key simplifies this since the keys would be different when encryption is off and when on (or even when turned off and re-enabled with a different key).

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -217,6 +217,14 @@ func (m *v1ToV2Migrator) FromVersion() filestore.PebbleKeyVersion {
 func (m *v1ToV2Migrator) ToVersion() filestore.PebbleKeyVersion { return filestore.Version2 }
 func (m *v1ToV2Migrator) Migrate(val []byte) []byte             { return val }
 
+type v2ToV3Migrator struct{}
+
+func (m *v2ToV3Migrator) FromVersion() filestore.PebbleKeyVersion {
+	return filestore.Version2
+}
+func (m *v2ToV3Migrator) ToVersion() filestore.PebbleKeyVersion { return filestore.Version3 }
+func (m *v2ToV3Migrator) Migrate(val []byte) []byte             { return val }
+
 // Register creates a new PebbleCache from the configured flags and sets it in
 // the provided env.
 func Register(env environment.Env) error {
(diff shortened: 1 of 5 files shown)


enterprise/server/raft/filestore/filestore.go
```diff
@@ -164,6 +173,18 @@ func (pmk *PebbleKey) Bytes(version PebbleKeyVersion) ([]byte, error) {
 		partDir := PartitionDirectoryPrefix + pmk.partID
 		filePath = filepath.Join(partDir, filePath, "v2")
 		return []byte(filePath), nil
+	case Version3:
+		rih := pmk.remoteInstanceHash
+		if pmk.isolation == "ac" && rih == "" {
+			rih = "0"
```
jdhollen, line 179: uber-crazy, feel-free-to-ignore-this nit: is this value functionally impossible as a remote instance hash, or just extremely unlikely?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -217,6 +217,14 @@ func (m *v1ToV2Migrator) FromVersion() filestore.PebbleKeyVersion {
 func (m *v1ToV2Migrator) ToVersion() filestore.PebbleKeyVersion { return filestore.Version2 }
 func (m *v1ToV2Migrator) Migrate(val []byte) []byte             { return val }
 
+type v2ToV3Migrator struct{}
+
+func (m *v2ToV3Migrator) FromVersion() filestore.PebbleKeyVersion {
+	return filestore.Version2
+}
+func (m *v2ToV3Migrator) ToVersion() filestore.PebbleKeyVersion { return filestore.Version3 }
+func (m *v2ToV3Migrator) Migrate(val []byte) []byte             { return val }
```
jdhollen, line 226: sorry, slightly unrelated, when do we plan on even using this??

maggie-lou (approved):

enterprise/server/raft/filestore/filestore.go
vadimberezniker: I'm not sure actually, but if it's a possible output then it's as likely/unlikely as any other collision of the hash of (remote instance name).

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: I'm not aware of any concrete use cases at the moment.

jdhollen (approved):

PR #3889, opened 2023-05-01
CLI: Support Bazel subtractive patterns properly

siggisim:
This causes the CLI to parse Bazel's subtractive patterns syntax correctly: https://bazel.build/run/build#specifying-build-targets

For example you can run:
```
bazel build -- foo/... -foo/bar/...
```

Which will build everything in `foo/...` but excludes the subdirectory `foo/bar/...`.

Also adds a test.

### DIFF
--- a/cli/cmd/bb/bb.go
+++ b/cli/cmd/bb/bb.go
@@ -126,6 +126,9 @@ func run() (exitCode int, err error) {
 		return -1, err
 	}
 
+	// Show a picker if target argument is omitted.
+	args = picker.HandlePicker(args)
+
 	// Parse args.
 	bazelArgs, execArgs := arg.SplitExecutableArgs(args)
 	// TODO: Expanding configs results in a long explicit command line in the BB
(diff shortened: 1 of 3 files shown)


cli/cmd/bb/bb.go
```diff
@@ -126,6 +126,9 @@ func run() (exitCode int, err error) {
 		return -1, err
 	}
 
+	// Show a picker if target argument is omitted.
+	args = picker.HandlePicker(args)
```
jdhollen, line 130: no workflows implicitly depending on this, right? (we won't make anything start hanging?)

siggisim: Shouldn't change the behavior much (i.e. introduce pickers in any places where there previously weren't), we're just parsing this before splitting arguments before the `--` and after `--` so we can handle those "subtractive" patterns correctly (right now we sometimes add an extra `//...` argument where we shouldn't is all).

jdhollen (approved):

PR #3894, opened 2023-05-02
website: upgrade docusaurus to 2.4.0

sluongng:
This is a follow up of #3892.

Several changes needed to be made to accomodate for the new version:
- Deprecate colorMode.switchConfig config
- Use `className` instead of `class` in blog post's nav component.
- Several component deprecation / replacements
- BlogPostItem needed to be wrapped around by BlogPostProvider
- Tag struct in each blog post was also updated

**Related issues**: N/A

### DIFF
--- a/website/theme/BlogTagsPostsPage/index.tsx
+++ b/website/theme/BlogTagsPostsPage/index.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 
 import Layout from "@theme/Layout";
+import { BlogPostProvider } from "@docusaurus/theme-common/internal";
 import BlogPostItem from "@theme/BlogPostItem";
 import Link from "@docusaurus/Link";
 import type { Props } from "@theme/BlogTagsPostsPage";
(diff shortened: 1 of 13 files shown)


sluongng:
I have tested this locally and found everything to be working quite well.

The PR could use some cleanup(?) but I would love to get 1 round of review in first.

siggisim:
I've attempted this upgrade before, but ran into a bunch of small issues. I've committed a bunch of CSS changes to this PR that should fix a lot of them

siggisim (approved):

website/theme/BlogTagsPostsPage/index.tsx
```diff
@@ -56,14 +54,17 @@ function BlogTagsPostPage(props: Props): JSX.Element {
...
                   assets={BlogPostContent.assets}
                   metadata={BlogPostContent.metadata}
+                  content={BlogPostContent}
                   truncated>
-                  <BlogPostContent />
-                </BlogPostItem>
+                  <BlogPostItem truncated>
+                    <BlogPostContent />
+                  </BlogPostItem>
+                </BlogPostProvider>
```
siggisim, line 67: I think we somehow lost the "Read more" button here (you can see it on the prod `/blog/` page

siggisim (approved):

website/theme/BlogTagsPostsPage/index.tsx
sluongng: Fixed that, it was `s/metadata.truncated/metadata.hasTruncateMarker/`

PR #3895, opened 2023-05-02
Remove image layers from local storage after ztocs have been generated

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/sociartifactstore/sociartifactstore.go
+++ b/enterprise/server/sociartifactstore/sociartifactstore.go
@@ -508,7 +508,7 @@ func (s *SociArtifactStore) indexLayer(ctx context.Context, layer v1.Layer) (*re
 	// images sharing layers from interfering with each other.
 	tmpFileName, err := random.RandomString(10)
 	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -503,7 +503,9 @@ func (s *SociArtifactStore) indexLayer(ctx context.Context, layer v1.Layer) (*re
 			layerDigest.Hex, mediaType, compressionAlgo)
 	}
 
-	layerTmpFile, err := os.Create(filepath.Join(*layerStorage, layerDigest.Hex))
+	layerTmpFileName := filepath.Join(*layerStorage, layerDigest.Hex)
+	layerTmpFile, err := os.Create(layerTmpFileName)
```
vadimberezniker, line 507: you're better off using a random string (`random.RandomString(10)`) rather than the layer digest for the name. it's very unlikely to happen but you theoretically can get a conversion request for two images that share one or more layers which would both write to the same file.

vadimberezniker (approved):

PR #3896, opened 2023-05-02
Add support for streaming images using soci artifacts generated and stored in the app

iain-macdonald:
This PR adds a new flag, `executor.podman.enable_image_streaming` that connects the executor with the SociArtifactStore API added in [3724](https://github.com/buildbuddy-io/buildbuddy/pull/3724) and implemented in [3726](https://github.com/buildbuddy-io/buildbuddy/pull/3726) and [3856](https://github.com/buildbuddy-io/buildbuddy/pull/3856) so that the podman container code queries the SociArtifactStore for SOCI artifacts derived from the image, fetches those artifacts from the Content Addressable Store, stores them on the local filesystem where the soci-store expects them, and proceeds to pull and run the image which is now streamed. This behavior is only enabled when `executor.podman.enable_image_streaming` is true, `executor.podman.streamable_images` behavior is unaffected.


enterprise/server/cmd/executor/executor.go
```diff
@@ -100,26 +102,45 @@ func InitializeCacheClientsOrDie(cacheTarget string, realEnv *real_environment.R
 		}
 		log.Infof("Connecting to cache target: %s", cacheTarget)
 	}
+	addHealthCheck(conn, false /*=required*/, "grpc_cache_connection", realEnv.GetHealthChecker())
```
vadimberezniker, line 105: did you mean to make this "not required"?

enterprise/server/cmd/executor/executor.go
```diff
@@ -100,26 +102,45 @@ func InitializeCacheClientsOrDie(cacheTarget string, realEnv *real_environment.R
...
+}
+
+func InitializeInternalClients(internalTarget string, realEnv *real_environment.RealEnv) {
+	conn, err := grpc_client.DialTarget(internalTarget)
+	if err != nil {
+		log.Warningf("Unable to connect to app internal grpc target '%s': %s", internalTarget, err)
+		return
+	}
+	log.Infof("Connecting to cache target: %s", internalTarget)
+	addHealthCheck(conn, false /*=required*/, "grpc_internal_services_connection", realEnv.GetHealthChecker())
```
vadimberezniker, line 120: why do you want to make this "not required"?

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -443,13 +457,87 @@ func (c *podmanCommandContainer) PullImage(ctx context.Context, creds container.
...
 	ps.pulled = true
 	return nil
 }
 
+func (c *podmanCommandContainer) prepareToStreamImage(ctx context.Context) error {
+	ctx, err := prefix.AttachUserPrefixToContext(ctx, c.env)
+	if err != nil {
+		return err
+	}
+	resp, err := c.env.GetSociArtifactStoreClient().GetArtifacts(ctx, &socipb.GetArtifactsRequest{Image: c.image})
```
vadimberezniker, line 477: you are not passing platform or credentials

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -443,13 +457,87 @@ func (c *podmanCommandContainer) PullImage(ctx context.Context, creds container.
...
+			}
+			sociIndexDigest := godigest.NewDigestFromEncoded(godigest.SHA256, artifact.Digest.Hash)
+			imageDigest := godigest.NewDigestFromEncoded(godigest.SHA256, resp.ImageId)
+			if err = os.WriteFile(sociIndex(imageDigest), []byte(sociIndexDigest.String()), 0644); err != nil {
+				return err
+			}
+		}
+		blobReq.Digests = append(blobReq.Digests, artifact.Digest)
+	}
+	blobResp, err := c.env.GetContentAddressableStorageClient().BatchReadBlobs(ctx, &blobReq)
```
vadimberezniker, line 502: IIRC you said some of these artifacts can be large (hundreds of megabytes?)
BatchReadBlobs is not appropriate to use with large artifacts, they need to be retrieved via the streaming API

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -443,13 +457,87 @@ func (c *podmanCommandContainer) PullImage(ctx context.Context, creds container.
...
+}
+
+func sociBlob(digest *repb.Digest) string {
+	return sociBlobDirectory + digest.Hash
+}
+
+func resourceName(digest *repb.Digest) *rspb.ResourceName {
+	return &rspb.ResourceName{
+		Digest:         digest,
+		InstanceName:   "soci",
```
vadimberezniker, line 534: you can remove the instance name. we used to separate CAS artifacts by instance name, but no longer do.

vadimberezniker (changes requested):

enterprise/server/cmd/executor/executor.go
iain-macdonald: Yeah, this is trying to replicate the current grpc_cache_connection health checker that reports healthiness if the connection goes idle: https://github.com/buildbuddy-io/buildbuddy/blob/a8e2ae9bed50644867d1a9475a98f0621dc77754/enterprise/server/cmd/executor/executor.go#L104-L117

enterprise/server/cmd/executor/executor.go
iain-macdonald: Because for now the internal service is only use to fetch image streaming artifacts and if those are unavailable we can fall back to pulling the image slowly, so the system can still function without this connection albeit in a degraded state.

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Argh, whoops, fixed.

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: They're probably more like tens of megabytes but I haven't tried a really large image layer yet, so good idea to switch to streaming them.

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Done

iain-macdonald (commented):
Moved the internal app connection into podman.go and made it conditional on image streaming per offline discussion. This is ready for another look when you have a few minutes!

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -42,9 +52,17 @@ var (
...
-	pullTimeout      = flag.Duration("executor.podman.pull_timeout", 10*time.Minute, "Timeout for image pulls.")
+
+	// TODO(iain): delete the streamableImages flag
+	imageStreamingEnabled = flag.Bool("executor.podman.enable_image_streaming", false, "If set, all podman images are streamed using soci artifacts generated and stored in the apps.")
+	streamableImages      = flagutil.New("executor.podman.streamable_images", []string{}, "List of podman images that can be streamed using registry-stored soci artifacts. Note that if executor.podman.enable_image_streaming is set then all images are streamed using app-stored soci artifacts and the value of this flag is ignored.")
+
+	pullTimeout = flag.Duration("executor.podman.pull_timeout", 10*time.Minute, "Timeout for image pulls.")
+
+	// Currently, the app_internal target is only used for the
+	// SociArtifactStore, so its definition and use is restricted to this file.
```
vadimberezniker, line 63: You can make this a podman specific flag, like `executor.podman.soci_artifact_store_target`. I don't see a lot of benefit of keeping this flag generic.

vadimberezniker (changes requested):
one last issue re: flag name

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Sure, sounds good

server/real_environment/real_environment.go
```diff
@@ -603,3 +603,10 @@ func (r *RealEnv) GetSociArtifactStoreServer() socipb.SociArtifactStoreServer {
 func (r *RealEnv) SetSociArtifactStoreServer(s socipb.SociArtifactStoreServer) {
 	r.sociArtifactStoreServer = s
 }
+
+func (r *RealEnv) GetSociArtifactStoreClient() socipb.SociArtifactStoreClient {
```
vadimberezniker, line 607: now that you are initializing the client in podman.go, you no longer need to store it in the env. you can store internally in the podman provider struct

vadimberezniker (approved):

server/real_environment/real_environment.go
iain-macdonald: sg, done

PR #3898, opened 2023-05-02
Add webdriver test for keyboard shortcuts

iain-macdonald:
I tried to add a test for shift-c that copies the invocation page url, but couldn't get the banner to render remotely and actually pasting the link into a random text box elsewhere (on the settings page) interfered with the local clipboard and felt janky, but I'm open to suggestions.

**Related issues**: N/A


jdhollen (commented):
lmk if it passes the flakiness check, otherwise looks good.

iain-macdonald:
> lmk if it passes the flakiness check, otherwise looks good.

still getting occasional flakes due to port interference starting up the server, but nothing due to this test

jdhollen (approved):

PR #3899, opened 2023-05-03
bazelrc: support generate coverage output with lcov and genhtml

sluongng:
Used this recently to help with writing tests for #3800 and #3861.
Super helpful to identify code branches with missing tests.

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -149,6 +149,12 @@ test:webdriver-debug --test_env=DISPLAY
 # otherwise), and display verbose webdriver output in the terminal.
 test:webdriver-debug --test_output=streamed
 
+# Currently only works for go tests
+# Coverage outputs could be viewed with `genhtml`:
+#   $ genhtml -o cover bazel-out/_coverage/_coverage_report.dat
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -149,6 +149,12 @@ test:webdriver-debug --test_env=DISPLAY
 # otherwise), and display verbose webdriver output in the terminal.
 test:webdriver-debug --test_output=streamed
 
+# Currently only works for go tests
+# Coverage outputs could be viewed with `genhtml`:
```
siggisim, line 153: I also recently added support in the UI to view this on the test results page

siggisim (approved):

.bazelrc
sluongng: Yup, however I think it's rendering using the wrong version of the source file 🤔 
Perhaps it was because I did not push the changes to github just yet?

siggisim: Yeah, if you push the changes to Github it should be able to render correctly

PR #3903, opened 2023-05-03
Fix esbuild pointing to non-existent directory

sluongng:
When attempting to upgrade rules_nodejs for the internal repo, it seems
that the location of `node_modules` directory could be moved to be under
the Open Source repo (this repo).

Attempting to detect the usual 'node_modules' directory not being there
and switch to use the new path under 'external/'.

This fix is expected to be a no-op in the current repo and the current
internal repo.  However this should help unblock this internal PR:
https://github.com/buildbuddy-io/buildbuddy-internal/pull/2270

**Related issues**: N/A

### DIFF
--- a/enterprise/app/esbuild.config.mjs
+++ b/enterprise/app/esbuild.config.mjs
@@ -1,15 +1,22 @@
 import * as path from "path";
+import { existsSync } from "node:fs";
 
 // browserifyPathPlugin remaps the nodejs "path" module to "path-browserify".
 // This is needed for libsodium: https://github.com/evanw/esbuild/issues/1786
 let browserifyPathPlugin = {
   name: "path-browserify",
   setup(build) {
+    let entry = path.join(process.cwd(), "node_modules");
+    // detect if we are building from buildbuddy-internal repo and
+    // adjust 'node_modules' path accordingly
+    if (!existsSync(entry)) {
+      entry = path.join(process.cwd(), "external/com_github_buildbuddy_io_buildbuddy", "node_modules");
+    }
     build.onResolve({ filter: /^path$/ }, () => ({
       // Note: the plugin API currently doesn't have a way to just replace
       // require("path") with require("path-browserify"); we have to resolve the
       // import to an actual file path.
-      path: path.join(process.cwd(), "node_modules/path-browserify/index.js"),
+      path: entry + "/path-browserify/index.js",
     }));
   },
 };


sluongng:
While testing this from the internal repo, I ran into some weird bug like this

```
✘ [ERROR] Cannot start service: Host version "0.14.53" does not match binary version "0.15.17"

1 error
Error: The service was stopped
    at /private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:1356:25
    at /private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:678:9
    at Socket.afterClose (/private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:656:7)
    at Socket.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1358:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
```

Not sure what caused it but it seems like a `bazel clean --expunge && rm -rf node_modules` on both repos fixed it.

jdhollen:
> While testing this from the internal repo, I ran into some weird bug like this
> 
> ```
> ✘ [ERROR] Cannot start service: Host version "0.14.53" does not match binary version "0.15.17"
> 
> 1 error
> Error: The service was stopped
>     at /private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:1356:25
>     at /private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:678:9
>     at Socket.afterClose (/private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:656:7)
>     at Socket.emit (node:events:525:35)
>     at endReadableNT (node:internal/streams/readable:1358:12)
>     at processTicksAndRejections (node:internal/process/task_queues:83:21)
> ```
> 
> Not sure what caused it but it seems like a `bazel clean --expunge && rm -rf node_modules` on both repos fixed it.

Sorry for being slow.  Is this just from the versions of esbuild getting out of sync between internal and the main repo?  Is this just gonna be some weird point of suffering that we can't work around unless we do version upgrades at ~the same time?

jdhollen (commented):
tickling github into knowing that i replied to you; see my comment above.

sluongng:
> > While testing this from the internal repo, I ran into some weird bug like this
> > ```
> > ✘ [ERROR] Cannot start service: Host version "0.14.53" does not match binary version "0.15.17"
> > 
> > 1 error
> > Error: The service was stopped
> >     at /private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:1356:25
> >     at /private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:678:9
> >     at Socket.afterClose (/private/var/tmp/_bazel_sluongng/0fd4abe8e9728bd678b1511337808ad4/sandbox/darwin-sandbox/2618/execroot/buildbuddy_internal/external/npm/@bazel/esbuild/node_modules/esbuild/lib/main.js:656:7)
> >     at Socket.emit (node:events:525:35)
> >     at endReadableNT (node:internal/streams/readable:1358:12)
> >     at processTicksAndRejections (node:internal/process/task_queues:83:21)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not sure what caused it but it seems like a `bazel clean --expunge && rm -rf node_modules` on both repos fixed it.
> 
> Sorry for being slow. Is this just from the versions of esbuild getting out of sync between internal and the main repo? Is this just gonna be some weird point of suffering that we can't work around unless we do version upgrades at ~the same time?

Esbuild manages it's own go binary. And we manage esbuild using yarn and manage yarn using bazel rules_nodejs. So somewhere in this chain, while i was upgrading the 2 repos, the version of esbuild(js) and esbuild(go) get mismatched. Esbuild itself has an internal check for that, resulting in the stack trace above.

Im pretty sure it's a local development issue only. I expect it to get better when we migrate to rules_js (and pnpm) 🤞

jdhollen (approved):

enterprise/app/esbuild.config.mjs
```diff
@@ -1,15 +1,25 @@
...
   name: "path-browserify",
   setup(build) {
+    // detect if we are building from buildbuddy-internal repo and
+    // adjust 'node_modules' path accordingly
+    let entry = path.join(process.cwd(), "node_modules");
+    stat(entry, (err, _) => {
+      if (err) {
+        entry = path.join(process.cwd(), "external/com_github_buildbuddy_io_buildbuddy/node_modules");
+      }
+    });
```
bduffany, line 16: Instead of `stat`, you can use `fs.exists`. Also, I think we should use the `sync` API (`existsSync`) since there might be a race condition if the `exists` call isn't completed by the time the `onResolve` handler is invoked.

sluongng: `existsSync` is much cleaner. Good call!

PR #3905, opened 2023-05-03
Add AWS key support.

vadimberezniker:
**Related issues**: N/A


enterprise/server/backends/kms/kms.go
```diff
@@ -82,6 +106,51 @@ func (k *KMS) initGCPClient(ctx context.Context) error {
 	return nil
 }
 
+type awsKMSARN struct {
```
jdhollen, line 109: random non-actionable go style complaint, this is exactly why I wish this would be awsKmsArn

enterprise/server/backends/kms/kms.go
```diff
@@ -91,9 +160,71 @@ func (k *KMS) initLocalInsecureKMSClient(ctx context.Context) error {
...
+		defer k.awsClientsMu.Unlock()
+		key := fmt.Sprintf("%s-%s", arn.partition, arn.region)
+		client, ok := k.awsClients[key]
+		if ok {
+			return client, nil
+		}
+		if *awsCredentailsFile != "" {
+			client, err = crateAWSClientWithCreds(arn, *awsCredentailsFile)
+		} else {
+			client, err = awskms.NewClient(arn.uriLocationPrefix)
```
jdhollen, line 221: what context do we use this case in?

vadimberezniker: It would come into play if you were to run the server locally. It would pick up your AWS credentials from a "well known" location configured by the AWS SDK.

enterprise/server/backends/kms/kms.go
```diff
@@ -91,9 +160,71 @@ func (k *KMS) initLocalInsecureKMSClient(ctx context.Context) error {
...
+		}
+		k.awsClientsMu.Lock()
+		defer k.awsClientsMu.Unlock()
+		key := fmt.Sprintf("%s-%s", arn.partition, arn.region)
+		client, ok := k.awsClients[key]
+		if ok {
+			return client, nil
+		}
+		if *awsCredentailsFile != "" {
+			client, err = crateAWSClientWithCreds(arn, *awsCredentailsFile)
```
maggie-lou, line 219: typos in 'crate' and 'credentails'

enterprise/server/backends/kms/kms.go
maggie-lou: Could you add a comment about this

maggie-lou (approved):

enterprise/server/backends/kms/kms.go
vadimberezniker: Added a note to the flag description.

enterprise/server/backends/kms/kms.go
vadimberezniker: :sweat_smile:

jdhollen (approved):

PR #3906, opened 2023-05-03
Add --sync=true flag to cli

luluz66:
When --sync=true, we will
1) pass --bes_upload_mode=wait_for_upload_complete
2) pass --synchronous_write to cache_proxy in the sidecar. When
   synchronous_write flag is set, cache_proxy will not use the worker
   queue to process the write requests async.

**Related issues**: 
https://github.com/buildbuddy-io/buildbuddy-internal/issues/2108

### DIFF
--- a/cli/sidecar/sidecar.go
+++ b/cli/sidecar/sidecar.go
@@ -117,9 +117,6 @@ func isCI(args []string) bool {
 func ConfigureSidecar(args []string) []string {
 	// Disable sidecar on CI for now since the async upload behavior can cause
 	// problems if the CI runner terminates before the uploads have completed.
-	// TODO: The sidecar is just an async BES/cache proxy at the moment, but if
-	// we add more sidecar features then we should find a way to re-enable it in
-	// "synchronous" mode on CI.
 	if isCI(args) {
 		log.Debugf("CI build detected.")
 		syncFlag := arg.Get(args, "sync")
@@ -160,6 +157,7 @@ func ConfigureSidecar(args []string) []string {
 
 	if synchronousWriteFlag == "1" || synchronousWriteFlag == "true" {
 		sidecarArgs = append(sidecarArgs, "--synchronous_write")
+		sidecarArgs = append(sidecarArgs, "--build_event_server.synchronous")
 		args = append(args, "--bes_upload_mode=wait_for_upload_complete")
 	}
 
--- a/cli/sidecar/sidecar.go
+++ b/cli/sidecar/sidecar.go
@@ -95,8 +95,15 @@ func restartSidecarIfNecessary(ctx context.Context, bbCacheDir string, args []st
 	return sockPath, nil
 }
 
+func isFlagTrue(flag string) bool {
+	if flag == "1" || strings.EqualFold(flag, "true") {
+		return true
+	}
+	return false
+}
+
 func isCI(args []string) bool {
-	if os.Getenv("CI") == "1" || strings.EqualFold(os.Getenv("CI"), "true") {
+	if isFlagTrue(os.Getenv("CI")) {
 		return true
 	}
 	for _, md := range arg.GetMulti(args, "build_metadata") {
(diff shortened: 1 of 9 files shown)


cli/sidecar/sidecar.go
```diff
@@ -114,8 +121,12 @@ func ConfigureSidecar(args []string) []string {
 	// we add more sidecar features then we should find a way to re-enable it in
 	// "synchronous" mode on CI.
```
siggisim, line 122: Nit: you can remove this TODO now

siggisim (approved):

cli/sidecar/sidecar.go
```diff
@@ -146,6 +158,11 @@ func ConfigureSidecar(args []string) []string {
 		sidecarArgs = append(sidecarArgs, fmt.Sprintf("--cache_dir=%s", diskCacheDir))
 	}
 
+	if synchronousWriteFlag == "1" || synchronousWriteFlag == "true" {
+		sidecarArgs = append(sidecarArgs, "--synchronous_write")
+		args = append(args, "--bes_upload_mode=wait_for_upload_complete")
```
siggisim, line 163: I'm not sure this is enough for the BES to be sync because the sidecar fires and forgets the forwarded build events?

luluz66: For PublishBuildToolEventStream, we call stream.CloseSend() ](https://github.com/buildbuddy-io/buildbuddy/blob/93da3c661acd419d93a5bc329de216f220574c7d/server/build_event_protocol/build_event_server/build_event_server.go#L78) after we acked the everything. So if bazel is waiting for upload complete, it should be enough for Event Stream. 

But for PublishLifeCycleEvent, the server doesn't wait for the response from forwarded build events. I think in order to support the sync mode, I need to change the build_event_server to wait for the goroutines to finish before returning response behind a flag.

luluz66: ahh I see - For PublishBuildToolEventStream, the build_event_protocol server didn't check the response from the proxy.

cli/sidecar/sidecar.go
luluz66: Done

cli/sidecar/sidecar.go
luluz66: I updated the build_event_server to wait until forwarded clients acknowledged if sync mode is on.

siggisim (approved):
This looks good to me, but it might be worth having @vadimberezniker look at the waitgroup stuff since that's not my strong suit.

server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -16,6 +18,10 @@ import (
 	pepb "github.com/buildbuddy-io/buildbuddy/proto/publish_build_event"
 )
 
+var (
```
vadimberezniker, line 21: Instead of introducing a flag, can the sidecar pass an additional option to `NewBuildEventProtocolServer` to enable synchronous proxying?

server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -37,12 +43,18 @@ func NewBuildEventProtocolServer(env environment.Env) (*BuildEventProtocolServer
 }
 
 func (s *BuildEventProtocolServer) PublishLifecycleEvent(ctx context.Context, req *pepb.PublishLifecycleEventRequest) (*emptypb.Empty, error) {
+	wg := &sync.WaitGroup{}
```
vadimberezniker, line 46: nit: I would suggest using `errgroup` instead. It uses a waitgroup underneath but has a slightly more ergonomic interface . You can return `nil` error if you don't care about errors.

vadimberezniker (approved):

server/build_event_protocol/build_event_server/build_event_server.go
luluz66: Done

server/build_event_protocol/build_event_server/build_event_server.go
luluz66: Done.

PR #3908, opened 2023-05-04
Read tags from a build_metadata field

jdhollen:
**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/config/config.go
+++ b/server/build_event_protocol/config/config.go
@@ -1,0 +1,9 @@
+package config
+
+import "flag"
+
+var tagsEnabled = flag.Bool("app.tags_enabled", false, "Enable setting tags on invocations via build_metadata")
+
+func TagsEnabled() bool {
+	return *tagsEnabled
+}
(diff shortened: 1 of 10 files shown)


server/build_event_protocol/config/config.go
```diff
@@ -0,0 +1,9 @@
+package config
+
+import "flag"
+
+var tagsEnabled = flag.Bool("app.tags_enabled", false, "Enable setting tags on invocations via build_metadata")
```
vadimberezniker, line 5: any reason you can't put this flag in the only package that it's used? we have way too many `config.go` files.

vadimberezniker (changes requested):

server/build_event_protocol/config/config.go
jdhollen: Meh, I had it in my head that I'd use the same flag to control the UI, but honestly it's probably better to keep the ingestion separate from the display.

vadimberezniker (approved):

PR #3914, opened 2023-05-04
Add button to UI to trigger a workflow

maggie-lou:
Video demo: https://share.getcloudapp.com/Apuvpp0z

To encourage people to migrate from the legacy workflow product, this is only supported for workflows setup through Github App integration.

**Related issues**: Fixes buildbuddy-io/buildbuddy-internal#2080


maggie-lou:
* master
  * **PR #3914** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3914" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3914?utm_source=stack-comment).

enterprise/app/workflows/workflows.tsx
```diff
@@ -326,6 +441,39 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
...
                 <MenuItem onClick={this.onClickUnlinkMenuItem.bind(this)}>Unlink repository</MenuItem>
+                {/* The Run Workflow button is only supported for workflows configured with the Github App, not legacy workflows */}
+                {!this.props.webhookUrl && (
+                  <MenuItem
+                    onClick={this.onClickRunWorkflowMenuItem.bind(this)}
+                    disabled={this.state.isWorkflowRunning}>
+                    Run workflow
+                  </MenuItem>
+                )}
+                <Popup
```
jdhollen, line 452: Rather than double-nesting popups (which I assume shows two shades, too?), could you just inline this to expand inside of the popup that we're already showing?

enterprise/server/workflow/service/service.go
```diff
@@ -592,18 +592,11 @@ func (ws *workflowService) getWorkflowByID(ctx context.Context, workflowID strin
...
-	rwf, err := ws.getRepositoryWorkflow(ctx, groupID, parsedRepoURL)
-	if err != nil {
-		return nil, err
-	}
-
-	return rwf.Workflow, nil
+// GetLegacyWorkflowIDForGitRepository generates an artificial workflow ID so that legacy Workflow structs
+// can be created from GitRepositories to play nicely with the pre-existing architecture
+// that expects the legacy format
+func (ws *workflowService) GetLegacyWorkflowIDForGitRepository(groupID string, repoURL string) string {
```
jdhollen, line 598: this function doesn't look like it needs to be defined on the struct anymore (which would let you go back to not having `gitRepositoryWorkflow` on the struct as well and eliminate a bunch of the diffs in here if I'm not misreading)

enterprise/server/api/api_server.go
```diff
@@ -513,6 +507,11 @@ func (s *APIServer) ExecuteWorkflow(ctx context.Context, req *apipb.ExecuteWorkf
 	}
 	rsp, err := wfs.ExecuteWorkflow(ctx, r)
 	if err != nil {
+		if status.IsNotFoundError(err) {
```
jdhollen, line 510: Just confirming, all that's changed here is that this failure mode is handled by the service instead of checking before making the request, yeah?

enterprise/app/workflows/workflows.tsx
maggie-lou: I double nested the popups because I thought it would look weird to just append the 'run workflow' input form onto the existing menu. Here's what it looks like with the double popups: https://share.getcloudapp.com/xQuRXQ4r

Another option would be to move 'Run Workflow' out of the menu and into its own button

enterprise/server/workflow/service/service.go
maggie-lou: The problem is that buildbuddy_server is in the OSS package, but workflow service is in the enterprise package. buildbuddy_server can't import the workflow file to use the static function, so defined the function on the struct

enterprise/server/api/api_server.go
maggie-lou: Yeah. I realized it was making an unnecessary additional DB read before, so moved things around to eliminate it

enterprise/app/workflows/workflows.css
```diff
@@ -132,6 +132,42 @@
   margin-right: 8px;
 }
 
+.workflows-page .workflow-item .run-workflow-input {
```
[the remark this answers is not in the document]
maggie-lou: I don't have a strong preference on this - was just following the style of the pre-existing file. Removed some scoping to simplify it a bit

enterprise/app/workflows/workflows.tsx
jdhollen: Yeah, pulling it out of this menu sounds nice--it would match the UI that we have to re-trigger on an individual workflow page: 
image

enterprise/server/workflow/service/service.go
jdhollen: oops! i missed the external call, thanks.

jdhollen (commented):
ok cool, let's update the UI to what you suggested, the rest lgtm

maggie-lou:
> ok cool, let's update the UI to what you suggested, the rest lgtm

Done! Here's what it looks like now: https://share.getcloudapp.com/Apuvpp0z

repeat.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
jdhollen, line 1: ahahaha accidental commit, but also, i think you should be able to approximate what you're doing here with `--runs_per_test=50`

maggie-lou: Oops lol

jdhollen (approved):

PR #3915, opened 2023-05-04
Remove dedicated partition requirement for encryption.

vadimberezniker:
With the new pebble key format, content is implicitely separated by customer via the encryption key ID so there's no longer a need to require a dedicated storage partition.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/crypter_service/crypter_service.go
+++ b/enterprise/server/crypter_service/crypter_service.go
@@ -864,10 +864,6 @@ func (c *Crypter) enableEncryption(ctx context.Context, kmsConfig *enpb.KMSConfi
 		return err
 	}
 
-	if !c.env.GetCache().SupportsEncryption(ctx) {
-		return status.UnavailableError("enabling encryption requires an account with dedicated storage")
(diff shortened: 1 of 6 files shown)


enterprise/server/crypter_service/crypter_service.go
```diff
@@ -864,10 +864,6 @@ func (c *Crypter) enableEncryption(ctx context.Context, kmsConfig *enpb.KMSConfi
 		return err
 	}
 
-	if !c.env.GetCache().SupportsEncryption(ctx) {
```
maggie-lou, line 867: Are you going to clean this up in a future PR?

maggie-lou (approved):

enterprise/server/crypter_service/crypter_service.go
vadimberezniker: I am thinking of renaming it and re-using it to display whether a customer has a dedicated partition when performing impersonation.

jdhollen (approved):

PR #3918, opened 2023-05-05
Flip a bunch of flags that have been on in dev and prod for a long time.

jdhollen:
Other than allowing the `pattern` filter, these are all implicitly or explicitly tied to the OLAP db handle existing.

**Related issues**: N/A


luluz66 (approved):
Looks like we also use the olap database to control whether to use olap db. So it won't break local bb or on-prem users.

PR #3919, opened 2023-05-05
Actually remove dedicated partition check.

vadimberezniker:
The previous PR didn't fully do the job, woops.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache_test.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache_test.go
@@ -1948,30 +1948,12 @@ func TestEncryption(t *testing.T) {
 	rootDir := testfs.MakeTempDir(t)
 	maxSizeBytes := int64(1_000_000_000) // 1GB
 
-	// Customer has encryption enabled but partition does not support encryption.
-	{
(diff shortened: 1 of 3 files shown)


jdhollen (approved):

enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -1948,30 +1948,12 @@ func TestEncryption(t *testing.T) {
...
-		err = pc.Stop()
-		require.NoError(t, err)
-	}
-
 	// Customer has encryption enabled, but no keys are available.
 	{
 		opts := &pebble_cache.Options{
 			RootDirectory: rootDir,
 			Partitions: []disk.Partition{{
-				ID: pebble_cache.DefaultPartitionID, EncryptionSupported: true, MaxSizeBytes: maxSizeBytes,
```
maggie-lou, line 1974: Not positive what your plan is for reusing SupportsEncryption() but can you clean up the EncryptionSupported field?

maggie-lou (approved):

enterprise/server/backends/pebble_cache/pebble_cache_test.go
vadimberezniker: I will do it in a separate PR.

PR #3920, opened 2023-05-05
Strictify //enterprise/app/tap

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/app/tap/tap.tsx
+++ b/enterprise/app/tap/tap.tsx
@@ -727,10 +727,12 @@ class InnerTopBar extends React.Component<InnerTopBarProps, InnerTopBarState> {
   private tooltipPortal?: HTMLDivElement;
 
   componentWillMount() {
-    this.tooltipPortal!.style.position = "fixed";
-    this.tooltipPortal!.style.zIndex = "1";
-    this.tooltipPortal!.style.opacity = "0";
-    document.body.appendChild(this.tooltipPortal!);
+    const tooltipPortal = document.createElement("div");
+    tooltipPortal.style.position = "fixed";
+    tooltipPortal.style.zIndex = "1";
+    tooltipPortal.style.opacity = "0";
+    document.body.appendChild(tooltipPortal);
+    this.tooltipPortal = tooltipPortal;
   }
 
   componentWillUnmount() {
--- a/enterprise/app/tap/tap.tsx
+++ b/enterprise/app/tap/tap.tsx
@@ -9,6 +9,7 @@ import { invocation } from "../../../proto/invocation_ts_proto";
 import { target } from "../../../proto/target_ts_proto";
 import { Subscription } from "rxjs";
 import router from "../../../app/router/router";
+import { google as google_duration } from "../../../proto/duration_ts_proto";
 import format from "../../../app/format/format";
 import { clamp } from "../../../app/util/math";
 import { FilterInput } from "../../../app/components/filter_input/filter_input";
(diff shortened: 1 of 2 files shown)


enterprise/app/tap/tap.tsx
```diff
@@ -403,33 +409,57 @@ export default class TapComponent extends React.Component<Props, State> {
...
+
+    if (!firstStats && !secondStats) {
+      return 0;
+    } else if (!firstStats) {
+      return 1;
+    } else if (!secondStats) {
+      return -1;
+    }
+
+    if (this.getSortMode() == "target") {
```
jdhollen, line 426: I don't think this matters a ton, but putting this block (target behavior) before the nullish stats checks would make the behavior more like what it was before, right?

enterprise/app/tap/tap.tsx
```diff
@@ -610,18 +644,18 @@ export default class TapComponent extends React.Component<Props, State> {
...
 
                       let destinationUrl = `/invocation/${status.invocationId}?target=${encodeURIComponent(
-                        targetHistory.target.label
+                        targetHistory.target?.label || ""
                       )}`;
                       let title =
                         this.getColorMode() == "timing"
-                          ? `${this.durationToNum(status.timing.duration).toFixed(2)}s`
-                          : this.statusToString(status.status);
+                          ? `${(this.durationToNum(status.timing?.duration) || 0).toFixed(2)}s`
```
jdhollen, line 651: Yeah, could simplify this by making the signature for durationToNum a little cleaner.

enterprise/app/tap/tap.tsx
```diff
@@ -685,15 +719,13 @@ class InnerTopBar extends React.Component<InnerTopBarProps, InnerTopBarState> {
   private hoveredCommitInfo = React.createRef<HTMLDivElement>();
   private hoveredCommitPointer = React.createRef<HTMLDivElement>();
   // TODO(bduffany): Use a generic tooltip component.
-  private tooltipPortal: HTMLDivElement;
+  private tooltipPortal: HTMLDivElement = document.createElement("div");
```
jdhollen, line 722: ehhh, rather than doing this outside of the lifecycle, how about just doing ! on the relevant callsites?

jdhollen (changes requested):

enterprise/app/tap/tap.tsx
```diff
@@ -216,36 +216,40 @@ export default class TapComponent extends React.Component<Props, State> {
       let stats: Stat = { count: 0, pass: 0, totalDuration: 0, maxDuration: 0, avgDuration: 0 };
       for (let status of targetHistory.targetStatus) {
         stats.count += 1;
-        let duration = this.durationToNum(status.timing.duration);
+        let duration = this.durationToNum(status.timing?.duration);
```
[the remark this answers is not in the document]
iain-macdonald: Done

enterprise/app/tap/tap.tsx
iain-macdonald: Yup, done

enterprise/app/tap/tap.tsx
iain-macdonald: Done

enterprise/app/tap/tap.tsx
iain-macdonald: Done

jdhollen (approved):

PR #3922, opened 2023-05-08
Separate encrypted content on disk.

vadimberezniker:
The same digest may exist in the cache in both unencrypted and encrypted formats.

**Related issues**: N/A


maggie-lou:
Kind of unrelated to this PR, but do you think our customers will want all of their unencrpyted data deleted from the cache when they turn on encryption? Or is the plan to just let the un-encrypted entries expire after the TTL

maggie-lou (approved):

vadimberezniker:
> Kind of unrelated to this PR, but do you think our customers will want all of their unencrpyted data deleted from the cache when they turn on encryption? Or is the plan to just let the un-encrypted entries expire after the TTL

It's not practical to delete previously encrypted content when enabling encryption. We will make sure to document that enabling encryption does not affect previously written artifacts.

jdhollen (approved):

PR #3929, opened 2023-05-08
Mark last chunk so we can detect truncation.

vadimberezniker:
**Related issues**: N/A


enterprise/server/crypter_service/crypter_service.go
```diff
@@ -548,6 +548,7 @@ func (d *Decryptor) Read(p []byte) (n int, err error) {
 		// ErrUnexpectedEOF indicates that the underlying reader returned EOF
 		// before the buffer could be filled, which is expected on the last
 		// chunk.
+		lastChunk := err == io.ErrUnexpectedEOF
```
maggie-lou, line 551: Could there be an edge case where the reader perfectly fills the buffer on the read of the final chunk and doesn't return EOF? Would that prevent the final chunk from being read correctly?

vadimberezniker: Yes, good point. A file written normally would always have a non-full chunk at the end, but we also want to make sure that the file cannot be truncated after the fact. I've added a check to make sure we actually come across a "last chunk" chunk before we finish with the reader.

jdhollen (approved):

PR #3931, opened 2023-05-08
timing profile breakdown: take into account event.cat

luluz66:
1. use cat = "remote output download" for reporting download
   outputs
2. use both name = "subprocess.run" and cat = "local action execution"
   for "Executing locally"
3. Remove Objc and Swift compilation. Events with "SwiftCompile" and
   "ObjcCompile" can be in two different categories "local action execution" and
   "Remote execution process wall time". As a result, we are double accounting
   for SwiftComiple and ObjcCompile

**Related issues**:
https://github.com/buildbuddy-io/buildbuddy-internal/issues/1831

### DIFF
--- a/app/invocation/invocation_breakdown_card.tsx
+++ b/app/invocation/invocation_breakdown_card.tsx
@@ -6,6 +6,7 @@ import { getChartColor } from "../util/color";
 
 interface Props {
   durationMap: Map<string, number>;
+  durationByCategoryMap: Map<string, number>;
 }
 
 interface Datum {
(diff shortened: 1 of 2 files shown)


app/invocation/invocation_timing_card.tsx
```diff
@@ -140,6 +142,12 @@ export default class InvocationTimingCardComponent extends React.Component<Props
         } else {
           this.state.durationMap.set(event.name, event.dur);
         }
+
+        if (this.state.durationByCategoryMap.get(event.cat)) {
```
jdhollen, line 146: i think if you wanna make this a one-liner, you can do...

`set(event.cat, (this.state.durationByCategoryMap.get(event.cat) || 0) + event.dur)`

side note, this usage of a map in state is.. unusual and the map might as well be a regular member of the class (updating it this way wouldn't cause a re-render), but that's unrelated to this change.

app/invocation/invocation_breakdown_card.tsx
```diff
@@ -6,6 +6,7 @@ import { getChartColor } from "../util/color";
 
 interface Props {
   durationMap: Map<string, number>;
+  durationByCategoryMap: Map<string, number>;
```
jdhollen, line 9: seeing as these are both Map<string,number> with very similar names, it's probably worth documenting what the difference is and what can be found in each one

jdhollen (changes requested):

app/invocation/invocation_breakdown_card.tsx
luluz66: Done.

app/invocation/invocation_timing_card.tsx
luluz66: Done.

jdhollen (approved):

siggisim (approved):

PR #3935, opened 2023-05-09
Restart soci-store if it dies

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -90,33 +90,25 @@ type Provider struct {
 	streamableImages []string
 }
 
-func runSociStore(ctx context.Context) error {
-	// If the store crashed, the path must be unmounted to recover.
-	syscall.Unmount(sociStorePath, 0)
+func runSociStore(ctx context.Context) {
+	for {
+		log.Infof("Starting soci store")
+		sociStoreDir := sociStorePath
+		cmd := exec.CommandContext(ctx, "sudo", "/home/iain/soci-snapshotter-grpc/out/soci-store", sociStoreDir)
+		logWriter := log.Writer("[socistore] ")
+		cmd.Stderr = logWriter
+		cmd.Stdout = logWriter
+		cmd.Run()
 
-	log.Infof("Starting soci store")
-	sociStoreDir := sociStorePath
-	cmd := exec.CommandContext(ctx, "soci-store", sociStoreDir)
-	logWriter := log.Writer("[socistore] ")
-	cmd.Stderr = logWriter
-	cmd.Stdout = logWriter
-	if err := cmd.Start(); err != nil {
-		return status.UnavailableErrorf("could not start soci store: %s", err)
+		log.Infof("Detected soci store crash, restating")
+		// If the store crashed, the path must be unmounted to recover.
+		syscall.Unmount(sociStorePath, 0)
 	}
-
-	// To prevent podman trying to read from the soci-store directory
-	// before it's ready (which can take a few hundred ms), wait for up to
-	// one second for soci-store to start up. Note that this is happening
-	// during executor start up so it won't affect RBE times.
-	if err := disk.WaitUntilExists(context.Background(), sociStoreDir, disk.WaitOpts{}); err != nil {
-		return status.UnavailableErrorf("soci-store failed to start: %s", err)
-	}
-	return nil
 }
 
 func NewProvider(env environment.Env, imageCacheAuthenticator *container.ImageCacheAuthenticator, buildRoot string) (*Provider, error) {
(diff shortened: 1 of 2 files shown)


vadimberezniker:
Hate to disappoint you, but today healthchecks has no effect on whether or not executors pick up work. We should change that, but it's not currently the case.

vadimberezniker:
As discussed, we can put this in as a defense in case the soci store can't come back, but we can recover faster if we try to restart the process if we detect that it has crashed.

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -119,6 +121,21 @@ additionallayerstores=["/var/lib/soci-store/store:ref"]
...
 			return nil, status.UnavailableErrorf("soci-store failed to start: %s", err)
 		}
+
+		// TODO(iain): there's a concurrency bug in soci-store that causes it
+		// to die occasionally. Report the executor as unhealthy if the
+		// soci-store directory doesn't exist for any reason.
+		env.GetHealthChecker().AddHealthCheck(
+			"soci_store", interfaces.CheckerFunc(
+				func(ctx context.Context) error {
+					if _, err := os.Stat(sociStorePath); err == nil {
```
vadimberezniker, line 131: Have you verified that os.Stat() returns an error? Just want to check that a fuse error leads to a Stat error because technically the mount point is still there.

iain-macdonald: Yeah, I confirmed it returns a "transport not connected" error

iain-macdonald:
Updated to restart the store if it crashes and added a defensive check when creating the podman container, PTAL

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -112,13 +123,24 @@ additionallayerstores=["/var/lib/soci-store/store:ref"]
...
+						return nil
+					} else {
```
vadimberezniker, line 138: nit: unbalanced parenthesis in error message

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -87,17 +90,25 @@ type Provider struct {
...
+		cmd := exec.CommandContext(ctx, "soci-store", sociStoreDir)
 		logWriter := log.Writer("[socistore] ")
 		cmd.Stderr = logWriter
 		cmd.Stdout = logWriter
-		if err := cmd.Start(); err != nil {
-			return nil, status.UnavailableErrorf("could not start soci store: %s", err)
-		}
+		cmd.Run()
+
+		log.Infof("Detected soci store crash, restating")
```
vadimberezniker, line 103: nit: restarting

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -87,17 +90,25 @@ type Provider struct {
...
+		// If the store crashed, the path must be unmounted to recover.
+		syscall.Unmount(sociStorePath, 0)
```
vadimberezniker, line 105: We should add some kind of cooldown here (even if it's a simple `time.Sleep(500 * time.Millisecond)` for now) in case socistore is crash looping.

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -129,22 +151,27 @@ additionallayerstores=["/var/lib/soci-store/store:ref"]
...
 	imageIsStreamable := false
 	for _, streamableImage := range p.streamableImages {
 		if image == streamableImage {
 			imageIsStreamable = true
 			break
 		}
 	}
+	if imageIsStreamable {
+		if err := disk.WaitUntilExists(context.Background(), sociStorePath, disk.WaitOpts{}); err != nil {
+			return nil, status.UnavailableErrorf("soci-store failed to start: %s", err)
```
vadimberezniker, line 164: nit: failed to start -> not available

vadimberezniker (approved):

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Fixed

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Yeah, sounds good

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Fixed

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Fixed

PR #3937, opened 2023-05-09
Add support to run firecracker tests in a new executor pool with only bare-metal execution

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -159,6 +159,11 @@ build --experimental_inprocess_symlink_creation
 # on Docker for development.
 test --test_tag_filters=-docker
 
+# Don't build or run Firecracker tests by default, because they cannot be run on all environments
+# (Can only be run on Linux machines with bare execution)
+test --test_tag_filters=-bare
+build --build_tag_filters=-bare
+
 # Use C++17 standard for all C++ compilation
 build --host_cxxopt=-std=c++17
 
--- a/.bazelrc
+++ b/.bazelrc
@@ -155,14 +155,10 @@ build --enable_runfiles
 # Need to work around M1 Mac browser test issue https://github.com/bazelbuild/rules_webtesting/issues/438
 build --experimental_inprocess_symlink_creation
 
-# Don't run Docker tests by default, to avoid a hard dependency
-# on Docker for development.
-test --test_tag_filters=-docker
-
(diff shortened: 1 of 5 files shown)


maggie-lou:
* master
  * **PR #3937** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3937" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/3937?utm_source=stack-comment).

.bazelrc
```diff
@@ -137,6 +137,11 @@ build --experimental_inprocess_symlink_creation
 test --test_tag_filters=-docker
 build --build_tag_filters=-docker
 
+# Don't build or run Firecracker tests by default, because they cannot be run on all environments
+# (Can only be run on Linux machines with bare execution)
+test --test_tag_filters=-bare
```
bduffany, line 142: do these values accumulate? e.g. does `--test_tag_filters=-bare --test_tag_filters=-docker` mean the same thing as `--test_tag_filters=-bare,-docker` ?

maggie-lou: Hmm good point. Just combined them for simplicity

enterprise/server/workflow/service/service.go
```diff
@@ -821,6 +821,11 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 	if workflowAction.SelfHosted {
 		useSelfHostedExecutors = "true"
 	}
+	pool := ws.WorkflowsPoolName()
+	if workflowAction.Pool != "" {
+		pool = workflowAction.Pool
+	}
```
bduffany, line 827: I don't think we'll need to configure the pool for the workflow action, since the CI_runner can still run on the normal workflow pool - the RBE action that it executes just needs to be run on the bare pool.

You can set `exec_properties = {"Pool": "bare"}` on `firecracker_test` to get that to work (also with the new approach you'll need `"use-self-hosted-executors": "true"`

maggie-lou:
@bduffany Was waiting for the executor configs to go out in the release and then forgot about this for a bit

bduffany (approved):
:tada:

PR #3942, opened 2023-05-10
do invocation search in clickhouse when filtering by stats

jdhollen:
This code path is currently only used by drilldowns.  Plan is to also make tags search use this path, I'll send that separately.

Flag-guarded in case something comes up, but defaulting the flag to `true`.

Note that Clickhouse doesn't have in-progress and disconnected builds.

**Related issues**: N/A


enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -29,15 +32,21 @@ const (
...
 )
 
+var (
+	olapInvocationSearchEnabled = flag.Bool("app.olap_invocation_search_enabled", true, "If true, InvocationSearchService will query clickhouse for some queries.")
+)
+
 type InvocationSearchService struct {
 	env environment.Env
 	h   interfaces.DBHandle
+	oh  interfaces.OLAPDBHandle
```
vadimberezniker, line 42: nit: I don't think either `oh` and `h` are clear enough names for struct fields that are used throughout a package. They'd be okay as local function variables, but I'd prefer if struct names were even a tad more meaningful, like `dbh` and `olapdbh` for example.

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -48,23 +57,101 @@ func defaultSortParams() *inpb.InvocationSort {
 	}
 }
 
-func (s *InvocationSearchService) rawQueryInvocations(ctx context.Context, sql string, values ...interface{}) ([]*tables.Invocation, error) {
-	rows, err := s.h.RawWithOptions(ctx, db.Opts().WithQueryName("search_invocations"), sql, values...).Rows()
+// Clickhouse UUIDs are formatted without any dashes :(
```
vadimberezniker, line 60: It's too bad we didn't keep the original format... Can you add some bounds checks for len(uuid)?

vadimberezniker (approved):

enterprise/server/invocation_search_service/invocation_search_service.go
jdhollen: this always winds up padded to length == 32, so just added a check for that.

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -48,23 +57,108 @@ func defaultSortParams() *inpb.InvocationSort {
...
+func fixUUID(uuid string) (string, error) {
```
luluz66, line 61: nit: shall we use uuid.Base64StringToString() for conversion?

PR #3950, opened 2023-05-11
Reduce flakiness in canary metrics script by checking whether multiple metrics are unhealthy before a rollback

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/tools/check_metrics/check_metrics.go
+++ b/enterprise/tools/check_metrics/check_metrics.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
@@ -56,6 +57,9 @@ func main() {
 	}
 	promAPI := promapi.NewAPI(prometheusClient)
 
+	mu := &sync.Mutex{}
+	failedMetrics := make([]string, 0)
+
 	eg, gctx := errgroup.WithContext(context.Background())
 	for _, metric := range config.PrometheusMetrics {
 		metric := metric
@@ -65,9 +69,9 @@ func main() {
 				pollingIntervalSec = metric.PollingIntervalSeconds
 			}
 
-			maxUnhealthyCount := config.MaxUnhealthyCount
-			if metric.MaxUnhealthyCount != 0 {
-				maxUnhealthyCount = metric.MaxUnhealthyCount
+			maxUnhealthyCount := config.MaxMetricPollUnhealthyCount
+			if metric.MaxMetricPollUnhealthyCount != 0 {
+				maxUnhealthyCount = metric.MaxMetricPollUnhealthyCount
 			}
 
 			monitoringTimer := time.After(time.Duration(config.MonitoringTimeframeSeconds) * time.Second)
@@ -95,7 +99,18 @@ func main() {
 					}
 
 					if unhealthyCount >= maxUnhealthyCount {
-						return errors.New(fmt.Sprintf("%s metrics unhealthy.", metric.Name))
+						err = func() error {
+							log.Debugf("Metric %s has been consecutively unhealthy %d times. Marking as failed.", metric.Name, unhealthyCount)
+							mu.Lock()
+							defer mu.Unlock()
+
+							failedMetrics = append(failedMetrics, metric.Name)
+							if len(failedMetrics) >= config.MaxMetricFailureCount {
+								return errors.New(fmt.Sprintf("Metrics unhealthy: %v", failedMetrics))
+							}
+							return nil
+						}()
+						return err
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -95,7 +103,21 @@ func main() {
 					}
 
 					if unhealthyCount >= maxUnhealthyCount {
-						return errors.New(fmt.Sprintf("%s metrics unhealthy.", metric.Name))
+						err = func() error {
```
iain-macdonald, line 106: why not just do this stuff inline instead of wrapping in a function?

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -201,3 +223,75 @@ func relativeRangeMetricHealthy(promAPI promapi.API, metricName string, metricVa
...
+
+type TextBlock struct {
+	Type string `json:"type"`
+	Text string `json:"text"`
+}
+
+func sendMetricUnhealthySlack(metric string, isSecondaryMetric bool, maxSecondaryMetricFailureCount int) {
+	slackWebhookURL := os.Getenv("SLACK_WEBHOOK")
+	if slackWebhookURL == "" {
+		log.Warningf("No slack webhook set. Cannot send warning messages.")
```
iain-macdonald, line 245: Errorf / Fatalf?

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -95,7 +103,21 @@ func main() {
 					}
 
 					if unhealthyCount >= maxUnhealthyCount {
-						return errors.New(fmt.Sprintf("%s metrics unhealthy.", metric.Name))
+						err = func() error {
+							log.Warningf("Metric %s has been consecutively unhealthy %d times. Marking as failed.", metric.Name, unhealthyCount)
+							sendMetricUnhealthySlack(metric.Name, metric.IsSecondaryMetric, config.MaxSecondaryMetricFailureCount)
```
iain-macdonald, line 108: WDYT about gathering the status of all metrics and sending them all (including the passing ones) to slack instead of sending them off one-by-one like this?

iain-macdonald (commented):
Out of curiosity, would using a moving average for some of the spiky metrics improve their reliability? Curious about that approach in combination with or instead of this approach

maggie-lou:
> Out of curiosity, would using a moving average for some of the spiky metrics improve their reliability? Curious about that approach in combination with or instead of this approach

My understanding is that moving averages are not the best fit for canary scripts. We only want to consider metrics for the canary from after it has been re-deployed. Otherwise we might smooth out changes caused by the deploy. We also don't want the metrics script to run for too long, because if the canary is unhealthy, we want to roll back sooner than later. That means there's only a small window where we're polling the metrics, and at that point if the metrics are spiking, the moving average would spike too

enterprise/tools/check_metrics/check_metrics.go
maggie-lou: Fixed

enterprise/tools/check_metrics/check_metrics.go
maggie-lou: Changed to Errorf. Didn't want to use Fatalf because this script causes a rollback in the deploy if it fails. If the env var accidentally gets unset and the slack can't be sent, it probably shouldn't trigger a rollback

enterprise/tools/check_metrics/check_metrics.go
maggie-lou: Done

enterprise/tools/check_metrics/check_metrics.go
```diff
@@ -201,3 +235,105 @@ func relativeRangeMetricHealthy(promAPI promapi.API, metricName string, metricVa
...
+
+		if m.consecutiveUnhealthyCount > 0 {
+			// Display unhealthy metrics at the top of the message
+			msg = metricOverview + msg
+		} else {
+			msg += metricOverview
+		}
+	}
+
+	msgColor := "#36a64f" // Green
```
iain-macdonald, line 283: nit: consider defining consts for these colors

enterprise/tools/check_metrics/config.go
```diff
@@ -19,12 +22,17 @@ type PrometheusMetric struct {
...
 	// If the metric does not meet this health threshold, it is considered unhealthy
 	HealthThreshold HealthThreshold `yaml:"health_threshold"`
 	// Set if it's valid for the metric value to be 0 or missing
 	// Ex. For error log count or invocation failure rate, at any given time the value can be 0
 	IsMissingDataValid bool `yaml:"is_missing_data_valid"`
+	// Primary metric failure should immediately result in a rollback
+	// If a secondary metric fails, it should only result in a rollback if several other secondary metrics
+	// are also failing (configurable by MaxSecondaryMetricFailureCount). This is intended to reduce unnecessary
+	// rollbacks from spiky secondary metrics
+	IsSecondaryMetric bool `yaml:"is_secondary_metric"`
```
iain-macdonald, line 35: nit: feel free to shorten this to something like `is_secondary` or something if you want

iain-macdonald (approved):

PR #3954, opened 2023-05-11
Add Google's VertexAI as a suggestion backend

siggisim:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/openai/BUILD
+++ b/enterprise/server/backends/openai/BUILD
@@ -4,7 +4,6 @@ go_library(
     name = "openai",
     srcs = ["openai.go"],
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/openai",
-    visibility = ["//visibility:public"],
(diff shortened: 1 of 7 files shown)


enterprise/server/backends/openai/BUILD
```diff
@@ -6,7 +6,7 @@ go_library(
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/openai",
     visibility = ["//visibility:public"],
     deps = [
-        "//cli/log",
```
tylerwilliams, line 9: Maybe we should add a visibility rule to disallow cli imports from //server/...

tylerwilliams (approved):

enterprise/server/backends/openai/BUILD
siggisim: Will send a separate PR for that, also removed the public visibility of this rule and the vertex one so it falls back to the default //enterprise subpackages

PR #3955, opened 2023-05-11
Add UI support for multiple "Ask Buddy" backends

siggisim:
Depends on https://github.com/buildbuddy-io/buildbuddy/pull/3954

### DIFF
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -95,4 +95,7 @@ message FrontendConfig {
 
   // Whether to show encryption settings.
   bool customer_managed_encryption_keys_enabled = 31;
+
+  // True if multiple suggestions providers are configured.
+  bool multiple_suggestion_providers = 32;
 }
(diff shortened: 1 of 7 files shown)


proto/config.proto
```diff
@@ -95,4 +95,7 @@ message FrontendConfig {
 
   // Whether to show encryption settings.
   bool customer_managed_encryption_keys_enabled = 31;
+
+  // True if multiple suggestions providers are configured.
+  bool multiple_suggestion_providers = 32;
```
iain-macdonald, line 100: Might make sense to return a list of strings for this so it's a little more pluggable, WDYT?

siggisim: I thought about it, but right now I'm displaying a big warning message specific to OpenAI - so sending all of the fields needed via this frontend config seemed like overkill.

I promise I'll do that if we add a third provider :)

iain-macdonald (approved):

PR #3963, opened 2023-05-11
Render tags on invocations and support them in invocation search + invocation trends.

jdhollen:
Need to add tags to the executions table in clickhouse as well, will do that separately.

**Related issues**: N/A


proto/invocation.proto
```diff
@@ -267,6 +267,9 @@ message InvocationQuery {
 
   // The pattern for the targets built (exact match). Ex: "//..."
   string pattern = 15;
+
+  // A plaintext tag for the targets built (exact match). Ex: "my-cool-tag"
+  string tag = 16;
```
siggisim, line 272: lame question: would it be a lot harder to support repeated tags here? or a comma separated string?

siggisim (approved):

proto/invocation.proto
jdhollen: not really, but i'm out tomorrow so you're getting it monday :P

jdhollen: just kidding, love 2 write software, here it is.

siggisim (approved):

PR #3965, opened 2023-05-12
Let people say a little bit in the job application form if they want

iain-macdonald:
Also update the job listings to mention that folks might have to work with the rest of us too and refer to offsites.

**Related issues**: N/A


tylerwilliams (commented):
Nice! Thanks for doing this <3

tylerwilliams (approved):

PR #3968, opened 2023-05-12
Add some "shortcuts" for cli commands `b`=>`build`, `t`=>`test`, etc

siggisim:
**Related issues**: N/A

### DIFF
--- a/cli/parser/parser.go
+++ b/cli/parser/parser.go
@@ -78,6 +78,24 @@ var (
 		`$`)
 
 	flagShortNamePattern = regexp.MustCompile(`^[a-z]$`)
+
+	// A hardcoded list of Bazel startup args that take non-boolean values.
+	// Grabbed from the output of `bazel help startup_options` of Bazel 6.1.2.
+	// These shouldn't change often.
+	defaultStartupArgs = `
+		--bazelrc (a string; default: see description)
+		--connect_timeout_secs (an integer; default: "30")
+		--failure_detail_out (a path; default: see description)
+		--io_nice_level (an integer; default: "-1")
+		--local_startup_timeout_secs (an integer; default: "120")
+		--macos_qos_class (a string; default: "default")
+		--max_idle_secs (an integer; default: "10800")
+		--output_base (a path; default: see description)
+		--output_user_root (a path; default: see description)
+		--server_jvm_out (a path; default: see description)
+		--host_jvm_args (a string; may be used multiple times)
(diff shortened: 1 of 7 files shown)


cli/shortcuts/shortcuts.go
```diff
@@ -0,0 +1,21 @@
...
+		"b": "build",
+		"t": "test",
+		"q": "query",
+		"r": "run",
+		"f": "fix",
+	}
+)
+
+func HandleShortcuts(args []string) []string {
+	command, idx := arg.GetCommandAndIndex(args)
```
bduffany, line 16: Use `parser.GetBazelCommandAndIndex` here, otherwise I think this might fail for things like `bazel --output_base foo b blah`

siggisim: That would work for most of these, but not for `fix`. And it wouldn't work for `bazel --output_base test b blah`.

I added a change to canonicalize just startup options very early on in `bb.go`. 

In order to do this, I hardcoded just the set of startup options that take non-boolean options, which is not very large and I don't suspect will change very often. In the unlikely case these change, we can just add the new flags. Also for a new flag to cause an issue, the user would have to:
a) use this new flag 
b) use buildbuddy cli 
c) use a shortcut 
d) use the non `=` version of the flag

In order for there to be an issue, which further reduces the likelihood of this being a problem at which point we would just add the new flag.

This should solve this problem for all places we call `GetBazelCommand` and it's variants.

Let me know what you think.

tylerwilliams (approved):

cli/parser/parser.go
```diff
@@ -353,11 +371,17 @@ func getCommandLineSchema(args []string, bazelHelp BazelHelpFunc) (*CommandLineS
 	return schema, nil
 }
 
+func CanonicalizeStartupArgs(args []string) ([]string, error) {
+	return canonicalizeArgs(args, func(topic string) (string, error) {
+		return defaultStartupArgs, nil
+	}, true)
```
bduffany, line 377: would it work to pass `onlyStartupOptions` to `getCommandLineSchema` and early-return after the `schema` is initially created and populated with the parsed startup options? Then we can let `runBazelHelpWithCache` do its thing without needing to hard-code the startup_options help

siggisim: Good idea, I thought about doing something like this - but I didn't realize it was this simple. Done.

bduffany (approved):

PR #3969, opened 2023-05-12
Run buildifier on all BUILD/WORKSPACE files as part of `bb fix`

siggisim:
**Related issues**: N/A

### DIFF
--- a/cli/fix/fix.go
+++ b/cli/fix/fix.go
@@ -51,8 +51,10 @@ func walk() error {
 				return nil
 			}
 			fileName := filepath.Base(path)
-			fileNameRoot := strings.TrimSuffix(fileName, filepath.Ext(fileName))
-			if fileNameRoot != "BUILD" && fileNameRoot != "WORKSPACE" {
+			fileExt := filepath.Ext(fileName)
+			fileNameRoot := strings.TrimSuffix(fileName, fileExt)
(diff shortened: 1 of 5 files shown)


cli/fix/fix.go
```diff
@@ -30,5 +33,39 @@ func HandleFix(args []string) (exitCode int, err error) {
...
 
+	err = walk()
+	if err != nil {
+		log.Printf("Error fixing: %+v", err)
+	}
 	return 0, nil
 }
+
+func walk() error {
+	return filepath.Walk(".",
```
bduffany, line 44: nit: use `WalkDir` instead

> Walk is less efficient than WalkDir, introduced in Go 1.16, which avoids calling os.Lstat on every visited file or directory.

bduffany (approved):

cli/fix/fix.go
bduffany: actually, does buildifier support passing a dir path? wonder if we can let buildifier handle the walking

siggisim: It does, but I'd like to add other things to this `walk` in follow-up PRs and want to avoid having to walk the directory tree multiple times.

siggisim: I switched over to walkDir

bduffany (approved):

PR #3970, opened 2023-05-15
docker/podman: Override image default ENTRYPOINT

sluongng:
When using a user-provided custom container image, the ENTRYPOINT must
be override otherwise it would affect action execution.

TODO: Add tests

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2293


bduffany (commented):
This might break people if they are using ENTRYPOINT to set up the image environment, e.g. running a script to set env vars. That seems like a valid use case to me?

I wonder if we should tell people to just not set ENTRYPOINT for now, and if it becomes a common pain point then revisit.

bduffany:
A doc update seems reasonable until more people complain about this. Using ENTRYPOINT at all seems pretty uncommon.

docs/rbe-platforms.md
```diff
@@ -33,6 +33,10 @@ platform(
 
 Make sure to replace `gcr.io/YOUR:IMAGE` with your docker image url.
 
+Remote build actions will be run in your container via CMD, so you should make sure that
+the container image's ENTRYPOINT is either unset or is a wrapper that is compatible with your
+build actions' command. For more information, check [CMD and ENTRYPOINT interact](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact).
```
bduffany, line 38:
```suggestion
build actions' commands. For more information, see [Understand how CMD and ENTRYPOINT interact](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact).
```

docs/rbe-platforms.md
```diff
@@ -33,6 +33,10 @@ platform(
 
 Make sure to replace `gcr.io/YOUR:IMAGE` with your docker image url.
 
+Remote build actions will be run in your container via CMD, so you should make sure that
```
bduffany, line 36:
```suggestion
Remote build actions will be run in your container via CMD, so note that any `CMD` instructions in your Dockerfile will be ignored. ENTRYPOINT, on the other hand, is not ignored, so make sure that
```

bduffany, line 36: nit: move this paragraph to a new section called "CMD and ENTRYPOINT" just before the next section ("Passing credentials for Docker images")

docs/rbe-platforms.md
bduffany: looks like these edits got lost after moving the section?

docs/rbe-platforms.md
bduffany: looks like this edit got lost too

docs/rbe-platforms.md
```diff
@@ -41,6 +41,14 @@ You can then pass this configuration to BuildBuddy RBE with the following flag:
 
 This assumes you've placed this rule in your root BUILD file. If you place it elsewhere, make sure to update the path accordingly.
 
+### ENTRYPOINT and CMD
+
+Remote build actions will be run in your container via CMD, so you should make sure that
+the container image's ENTRYPOINT is either unset or is a wrapper that is compatible with your
+build actions' command. For more information, check [CMD and ENTRYPOINT interact](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact).
+
+You can then pass this configuration to BuildBuddy RBE with the following flag:
```
bduffany, line 50:
```suggestion
```

sluongng:
So I tried creating a Symbolic Ref in my local git repo for `main` pointing it to `master` to help switch between some repos in BazelBuild org using `main` as the default branch.
However, it seems like `main` got stale even though `master` was updated.

I accidentally run a rebase on `main` instead of `master`... then got into some conflict resolution when git stash pop automatically... It was a disaster locally.

Anyhow... I dig up the reflog and reset it to the fixup commits from your suggestion and added some minor formatting changes. Should be good to go now. 

Sorry for the noise. 🙇

PR #3971, opened 2023-05-15
Refactor common GRPC healthcheck creation to a single function

iain-macdonald:
**Related issues**: N/A


server/util/healthcheck/healthcheck.go
```diff
@@ -48,6 +50,21 @@ type HealthChecker struct {
 	shuttingDown  bool
 }
 
+func NewGRPCHealthCheck(conn *grpc.ClientConn) interfaces.CheckerFunc {
```
tylerwilliams, line 53: can you throw a comment on here? otherwise lgtm

PR #3973, opened 2023-05-15
Add explicit group ID, instance name, and platform hash to snapshot cache key

bduffany:
Snapshot isolation by group_id, instance_name, and platform hash is currently implicit in the fact that only one `Runner` can use a snapshot at a time, and runners are effectively isolated by group_id, instance_name, and platform_hash.

Eventually we'd like to be able to look up snapshots from cache without needing a pre-existing `Runner` in the runner pool. So this PR is a step towards that goal by making these properties explicit rather than implicit.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -7,7 +7,9 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/filecacheutil"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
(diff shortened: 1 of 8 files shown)


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -18,6 +20,13 @@ import (
 // Key represents a cache key pointing to a snapshot
 // manifest.
 type Key struct {
+	// InstanceName is the remote instance name associated with the snapshot.
```
bduffany, line 23: Note that the `GroupID` is not added to the `Key` struct in this PR, because in the remote snapshot case, the group ID would instead be implicit in the JWT on the outgoing gRPC context. So instead I opted to take in the `ctx` in the `*filecacheKey` funcs.

maggie-lou (approved):

vadimberezniker (approved):

PR #3976, opened 2023-05-15
Introduce `DISABLE_TARGET_TRACKING` metadata and set it for workflow PRs

siggisim:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -1610,7 +1610,8 @@ func writeBazelrc(path, invocationID string) error {
 	if *prNumber == 0 {
 		lines = append(lines, "build --build_metadata=ROLE=CI")
 	} else {
-		lines = append(lines, "build --build_metadata=ROLE=PR")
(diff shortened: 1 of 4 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1610,7 +1610,8 @@ func writeBazelrc(path, invocationID string) error {
 	if *prNumber == 0 {
 		lines = append(lines, "build --build_metadata=ROLE=CI")
 	} else {
-		lines = append(lines, "build --build_metadata=ROLE=PR")
+		lines = append(lines, "build --build_metadata=ROLE=CI")
```
bduffany, line 1613: nit: slightly cleaner to move `build --build_metadata=ROLE=CI` back to `lines[0]`, then use `if *prNumber != 0`

bduffany (approved):

enterprise/server/cmd/ci_runner/main.go
siggisim: Done

bduffany (approved):

PR #3978, opened 2023-05-15
Make keyboard-selected invocation on history page have a nice lil blue outline

iain-macdonald:
Instead of yellow left-margin and box shadow. This new UX is consistent with how the share button looks, and maybe how other shortcut-highlighted items could look in the future...

**Related issues**: N/A


bduffany (approved):

enterprise/app/history/history.tsx
```diff
@@ -280,6 +286,10 @@ export default class HistoryComponent extends React.Component<Props, State> {
       selectedInvocationIndex: newInvocationIndex,
       selectedInvocationId: newInvocationId,
     });
+    const elements = document.getElementsByClassName("selected-keyboard-shortcuts");
+    if (elements.length == 1) {
+      (elements[0] as HTMLElement).focus();
```
bduffany, line 291: not a huge deal but I think this isn't actually focusing the element since `<div>` isn't focusable by default. Since `<a>` is focusable you could focus the parent `<a>` element like `document.querySelector('a:has(.selected-keyboard-shortcuts)')?.focus();`

bduffany: https://github.com/buildbuddy-io/buildbuddy/assets/2414826/30acd145-b21d-4a00-a1cb-04c22a8ae67a

iain-macdonald: Fixed

PR #3981, opened 2023-05-16
Make workflows more robust during CAS/Execution latency spikes

bduffany:
(1) Process webhook requests in the background so that we can immediately respond to GitHub's HTTP request instead of potentially timing out waiting for CAS writes / Execute requests.

(2) Start workflow actions in parallel. This should reduce the overall workflow start latency in the case where a repo has many actions (rules_xcodeproj for example has 15 actions), especially in the case where there is a spike in CAS write latency / Execution service latency. For example, if CAS writes are taking 1s and there are 15 workflow actions, then starting the actions in serial means that the last action will be artificially delayed by 14s.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -32,6 +32,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
 	"github.com/buildbuddy-io/buildbuddy/server/util/background"
 	"github.com/buildbuddy-io/buildbuddy/server/util/bazel_request"
 	"github.com/buildbuddy-io/buildbuddy/server/util/db"
@@ -82,9 +83,10 @@ const (
 	// Number of workers to work on processing webhook events in the background.
 	webhookWorkerCount = 64
 
-	// Max number of webhook payloads to buffer in memory. This is intentionally
-	// large to help deal with spikes in cache write latency.
-	webhookWorkerTaskQueueSize = 2000
+	// Max number of webhook payloads to buffer in memory. We expect some events
+	// to occasionally be buffered during transient spikes in CAS or Execution
+	// service latency.
+	webhookWorkerTaskQueueSize = 100
 
 	// How long to wait before giving up on processing a webhook payload.
 	webhookWorkerTimeout = 30 * time.Second
@@ -184,6 +186,9 @@ func (ws *workflowService) enqueueStartWorkflowTask(ctx context.Context, gitProv
 	case ws.tasks <- t:
 		return nil
 	default:
+		alert.UnexpectedEvent(
+			"workflow_task_queue_full",
+			"Workflows are not being triggered due to the task queue being full. This may be due to elevated CAS or Execution service latency.")
 		return status.ResourceExhaustedError("workflow task queue is full")
 	}
 }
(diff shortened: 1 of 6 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -1076,22 +1158,29 @@ func (ws *workflowService) startWorkflow(ctx context.Context, gitProvider interf
...
-		if err != nil {
-			if err == ApprovalRequired {
-				log.CtxInfof(ctx, "Skipping workflow action %s (%s) %q (requires approval)", wf.WorkflowID, wf.RepoURL, action.Name)
-				if err := ws.createApprovalRequiredStatus(ctx, wf, wd, action.Name); err != nil {
-					log.CtxWarningf(ctx, "Failed to create workflow %s (%s) action status: %s", wf.WorkflowID, wf.RepoURL, err)
+
+		// Start executions in parallel to help reduce workflow start latency
+		// for repos with lots of workflow actions.
+		wg.Add(1)
+		go func() {
```
maggie-lou, line 1165: Is there any concern about load? Now that the background thread can process 2000 tasks simultaneously and each task can execute multiple actions in parallel

enterprise/server/workflow/service/service.go
```diff
@@ -1076,22 +1158,29 @@ func (ws *workflowService) startWorkflow(ctx context.Context, gitProvider interf
...
+					log.CtxInfof(ctx, "Skipping workflow action %s (%s) %q (requires approval)", wf.WorkflowID, wf.RepoURL, action.Name)
+					if err := ws.createApprovalRequiredStatus(ctx, wf, wd, action.Name); err != nil {
+						log.CtxWarningf(ctx, "Failed to create workflow %s (%s) action status: %s", wf.WorkflowID, wf.RepoURL, err)
+					}
+					return
 				}
-				continue
+				// TODO: Create a UI for these errors instead of just logging on the
+				// server.
+				log.CtxWarningf(ctx, "Failed to execute workflow %s (%s) action %q: %s", wf.WorkflowID, wf.RepoURL, action.Name, err)
```
tylerwilliams, line 1179: instead of nesting the ```if err == ApprovalRequired {``` clause under ```if err != nil``` and returning early, could you separate those two clauses?

if err == ApprovalRequired {
  // blah
} else if err != nil {
  // blah blah blah
}

server/libmain/libmain.go
```diff
@@ -395,7 +395,9 @@ func StartAndRunServices(env environment.Env) {
 		Handler: env.GetMux(),
 	}
 
+	env.GetHTTPServerWaitGroup().Add(1)
 	env.GetHealthChecker().RegisterShutdownFunction(func(ctx context.Context) error {
+		defer env.GetHTTPServerWaitGroup().Done()
```
tylerwilliams, line 400: cute

tylerwilliams (approved):
lgtm (mod the comment re simplifying logic)

enterprise/server/workflow/service/service.go
```diff
@@ -108,13 +121,86 @@ func instanceName(wf *tables.Workflow, wd *interfaces.WebhookData, workflowActio
...
+	}
+}
+
+func (ws *workflowService) runStartWorkflowTask(task *startWorkflowTask) {
+	// Keep the existing context values from the client but set a new timeout
+	// since the HTTP request has already completed at this point.
+	ctx, cancel := background.ExtendContextForFinalization(task.ctx, webhookWorkerTimeout)
+	defer cancel()
+
+	if err := ws.startWorkflow(ctx, task.gitProvider, task.webhookData, task.workflow); err != nil {
```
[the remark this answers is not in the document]
bduffany: good idea - let me send a follow-up for that

PR #3987, opened 2023-05-16
CLI: Support BUILD.yaml => BUILD.bazel translation

siggisim:
**Related issues**: N/A


cli/fix/fix.go
```diff
@@ -50,13 +51,18 @@ func walk() error {
...
 				return nil
 			}
-			runBuildifier(path)
+			fileToFormat, err := translate.Translate(path)
+			if err != nil {
+				return err
+			}
+			if fileToFormat == "" {
+				return nil
+			}
```
tylerwilliams, line 64: does runBuildifier still get hit if no translation is done?

siggisim: Only if it's a BUILD or WORKSPACE file - otherwise `fileToFormat` is set to the empty string.

tylerwilliams (approved):

PR #3991, opened 2023-05-17
go: upgrade to 1.20.4

sluongng:
Also yanked the setup from #3831 into this.

**Related issues**: N/A

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -50,32 +50,36 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_download_sdk", "go_register_toolchai
 
 go_rules_dependencies()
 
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -50,33 +50,26 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_download_sdk", "go_register_toolchai
...
+    )
+    for goos in [
+        "linux",
+        "darwin",
+    ]
+    for goarch in [
+        "amd64",
+        "arm64",
+    ]
+]
```
tylerwilliams, line 72: nit: could we just leave these written out?

sluongng: Fixed.

PR #3992, opened 2023-05-17
gazelle: make fetch_repo crash with core dump

sluongng:
The issue we are facing with firecracker microvm seems to happen more
frequently with fetch_repo binary more than any others.  This is mostly
because in a Go project running in Workflows, fetch_repo is often the
first Go binary that get executed to download external dependencies.

Let's make future crashes more verbose to help with our investigation.

**Related issues**: N/A


sluongng:
closing debug PR

PR #3994, opened 2023-05-17
scheduler_server: rm remove_stale_executors flag for good

sluongng:
This flag has been flipped to default true in #3859.
Let's remove it entirely since turning it off would cause more problems.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/pull/2278


vadimberezniker (approved):
As a general practice, we should be careful about removing flags because if they are specified via command line they will cause the application to not start at all. 

In this case, I think it's very unlikely anyone is specifying this on the command line (or at all really).

siggisim:
I know of at least one customer that sets this via `config.yaml` but I guess in that case it'll just print a warning?

vadimberezniker:
> I know of at least one customer that sets this via `config.yaml` but I guess in that case it'll just print a warning?

Yes, it's not an error when set via config.

PR #3995, opened 2023-05-17
bazelrc: provide cross remote build config

sluongng:
We have this config in our internal document to help with debugging.
Let's just make it part of the bazelrc.

I use this by setting my `user.bazelrc` to

```
build --remote_header=x-buildbuddy-api-key=<foo>

build:x --config=remote-mac
```

And subsequently invoke the build from M1 laptop with
`bazel test --config=x //server/backends/...`

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -24,15 +24,16 @@ build:cache --experimental_remote_cache_compression
 
 # Flags shared across remote configs
 build:remote-shared --remote_upload_local_results
-build:remote-shared --host_platform=@buildbuddy_toolchain//:platform
-build:remote-shared --platforms=@buildbuddy_toolchain//:platform
-build:remote-shared --crosstool_top=@buildbuddy_toolchain//:toolchain
-build:remote-shared --extra_toolchains=//:sh_toolchain
 build:remote-shared --remote_timeout=600
 build:remote-shared --remote_download_minimal
 build:remote-shared --experimental_repo_remote_exec
 build:remote-shared --jobs=100
 build:remote-shared --verbose_failures
+build:remote-shared --platforms=@buildbuddy_toolchain//:platform_linux
+build:remote-shared --host_platform=@buildbuddy_toolchain//:platform_linux
+build:remote-shared --crosstool_top=@buildbuddy_toolchain//:ubuntu_cc_toolchain_suite
+build:remote-shared --extra_toolchains=@buildbuddy_toolchain//:ubuntu_cc_toolchain
+build:remote-shared --extra_toolchains=//:sh_toolchain
 
 # Build with --config=remote to use BuildBuddy RBE.
 build:remote --config=remote-shared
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -24,26 +24,39 @@ build:cache --experimental_remote_cache_compression
...
 
+# Build with --config=remote-mac from a MacOS machine to run builds and tests
+# remotely on Linux executors.
+build:remote-mac --config=remote-shared
+build:remote-mac --config=cache
+build:remote-mac --cpu=k8
+build:remote-mac --host_cpu=k8
+build:remote-mac --host_platform=@buildbuddy_toolchain//:platform_linux
+build:remote-mac --crosstool_top=@buildbuddy_toolchain//:ubuntu_cc_toolchain_suite
+build:remote-mac --extra_toolchains=@buildbuddy_toolchain//:ubuntu_cc_toolchain
```
bduffany, line 57: It looks like the only difference between this config and the `remote-shared-linux` config is that it's missing the `--platforms` flag. Does it still work if you add `--platforms` here? If so, I think the config could be simplified a bit since the flags would be the same between both linux and mac.

.bazelrc
```diff
@@ -24,26 +24,39 @@ build:cache --experimental_remote_cache_compression
...
 build:remote-shared --remote_download_minimal
 build:remote-shared --experimental_repo_remote_exec
 build:remote-shared --jobs=100
 build:remote-shared --verbose_failures
 
+build:remote-shared-linux --config=remote-shared
+build:remote-shared-linux --host_platform=@buildbuddy_toolchain//:platform
+build:remote-shared-linux --platforms=@buildbuddy_toolchain//:platform
+build:remote-shared-linux --crosstool_top=@buildbuddy_toolchain//:toolchain
+build:remote-shared-linux --extra_toolchains=//:sh_toolchain
```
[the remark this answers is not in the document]
sluongng: Nice call. I guess we don't need to use the `@buildbuddy_toolchain//:platform` alias if it's just gonna be linux all the way down.

I sent an update that should make things less complicated.

sluongng:
Side note: pretty much `bazel test --config=x //...` works for me with the exception of `enterprise/server/remote_execution/commandutil` which has a test tagged with `exclusive` and causes Bazel to download the linux binary to run locally on MacOS. Removing `exclusive` from there made it work... but that's a different conversation to have later.

.bazelrc
```diff
@@ -44,6 +45,15 @@ build:remote-dev --config=remote-shared
...
 build:remote-dev --remote_executor=grpcs://remote.buildbuddy.dev
 
+# Build with --config=remote-mac-x to run cross platform builds
+# and tests remotely on Linux executors from a MacOS machine.
+build:mac-x --cpu=k8
+build:mac-x --host_cpu=k8
+build:remote-mac-x --config=mac-x
+build:remote-mac-x --config=remote
+build:remote-mac-x-dev --config=mac-x
+build:remote-mac-x-dev --config=remote-dev
```
bduffany, line 55: nit: maybe call this `mac-crossplatform` or even just `linux-target` (since this is really just specifying linux as the target platform)? If folks want a shorter alias like `mac-x` then they can use `user.bazelrc` for that.

bduffany (approved):

Issue #3980, opened 2023-05-16
Getting event sequence number mismatch error on publishBuildEvents


reimai:
Hi, we are using buildbuddy version _v2.12.32_ and bazel _v5.4.0_ and getting this kind of errors from time to time:

on client:

```
Error:The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=UNKNOWN, description=event sequence number mismatch: received 47325, wanted 1, cause=null}' UNKNOWN: UNKNOWN: event sequence number mismatch: received 47325, wanted 1 UNKNOWN: UNKNOWN: event sequence number mismatch: received 47325, wanted 1
```

on server:
```
2023-05-15 19:40:20.410	
   stderr   2023/05/15 16:40:20.410 WRN Missing ack: saw 47325 and wanted 1. Bailing! invocation_id=57f82ac5-3910-41a4-ba05-439f3b6a34e2 request_id=05fb796b-8554-4219-8aaf-236f6d73363b   
2023-05-15 19:40:20.410	
   stderr   2023/05/15 16:40:20.410 WRN We got over 100 build events before an event with options for invocation 57f82ac5-3910-41a4-ba05-439f3b6a34e2. Dropped the 186916 earliest event(s). invocation_id=57f82ac5-3910-41a4-ba05-439f3b6a34e2 request_id=05fb796b-8554-4219-8aaf-236f6d73363b   
2023-05-15 19:40:10.260	
   stderr   2023/05/15 16:40:10.260 WRN Missing ack: saw 47325 and wanted 1. Bailing! invocation_id=57f82ac5-3910-41a4-ba05-439f3b6a34e2 request_id=3bb859e8-9282-4f03-87cf-d9758f7fb309   
2023-05-15 19:40:10.260	
   stderr   2023/05/15 16:40:10.259 WRN We got over 100 build events before an event with options for invocation 57f82ac5-3910-41a4-ba05-439f3b6a34e2. Dropped the 186916 earliest event(s). invocation_id=57f82ac5-3910-41a4-ba05-439f3b6a34e2 request_id=3bb859e8-9282-4f03-87cf-d9758f7fb309   
2023-05-15 19:40:01.691	
[...]

PR #3998, opened 2023-05-17
Fix BES error when bazel doesn't receive all ACKs

bduffany:
If Bazel doesn't receive all the ACKs that we send after finalizing the invocation, Bazel will attempt to retry the invocation starting from the last ACK'd sequence number, plus one.

In this case, we currently return an error due to a sequence number mismatch. This PR makes it so that we just re-ACK all the events that Bazel thinks it needed to re-send, without doing any additional work.

**Related issues**: #3980


tempoz (approved):
LGTM, very clean!

tylerwilliams (approved):

PR #4007, opened 2023-05-17
coerce data to boolean in a way that makes strict mode happy.

jdhollen:
strict mode doesn't like `Boolean(foo?.bar)`, seemingly because it worries that calling `Boolean` might modify `foo` to `null`.

also, note that this one was probably fine, but i'd rather keep emitting nothing instead of `""`

**Related issues**: N/A


bduffany (approved):
lame that they don't have a special case for `Boolean()` to somehow tell the compiler that it won't modify stuff :P

PR #4019, opened 2023-05-18
render cache savings on invocations card when trends summary is enabled.

jdhollen:
more sending this out for opinions on the text and positioning of this content on the page, because there are a lot of options.  the stuff i have now is a placeholder more than anything.

It currently looks like... this: 
image

and... this:
image

**Related issues**: N/A


siggisim (approved):
This is awesome, let's start with this and we can always refine to make it nicer in the future

bduffany (approved):

PR #4021, opened 2023-05-19
Actually set the image-to-index mapping in blobstore after it's generated and add a test that exercises that functionality

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/server/main.go
+++ b/enterprise/server/cmd/server/main.go
@@ -42,6 +42,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/tasksize"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/usage"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/usage_service"
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/dsingleflight"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/webhooks/bitbucket"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/webhooks/github"
 	"github.com/buildbuddy-io/buildbuddy/server/config"
@@ -301,6 +302,8 @@ func main() {
 	if err := selfauth.Register(realEnv); err != nil {
 		log.Fatalf("%v", err)
 	}
+
+	realEnv.SetSingleFlightDeduper(dsingleflight.New(realEnv.GetDefaultRedisClient()))
 
 	libmain.StartAndRunServices(realEnv) // Returns after graceful shutdown
 }
--- a/enterprise/server/cmd/server/main.go
+++ b/enterprise/server/cmd/server/main.go
@@ -275,6 +275,9 @@ func main() {
 	if err := crypter_service.Register(realEnv); err != nil {
 		log.Fatalf("%v", err)
 	}
+	if err := dsingleflight.Register(realEnv); err != nil {
+		log.Fatalf("%v", err)
+	}
 	if err := sociartifactstore.Register(realEnv); err != nil {
 		log.Fatalf("%v", err)
 	}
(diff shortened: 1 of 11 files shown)


enterprise/server/sociartifactstore/sociartifactstore_test.go
```diff
@@ -317,3 +348,22 @@ func (r *containerRegistry) Address() string {
 func (r *containerRegistry) ImageAddress(imageName string) string {
 	return fmt.Sprintf("%s:%d/%s", r.host, r.port, imageName)
 }
+
+type deduper struct {
+	mu sync.Mutex
+}
+
+func (d *deduper) Do(_ context.Context, _ string, work func() ([]byte, error)) ([]byte, error) {
+	d.mu.Lock()
```
vadimberezniker, line 357: Why do you need a lock here?

enterprise/server/sociartifactstore/sociartifactstore_test.go
```diff
@@ -317,3 +348,22 @@ func (r *containerRegistry) Address() string {
...
+	defer d.mu.Unlock()
+	return work()
+}
+
+func contains(s []string, e string) bool {
```
vadimberezniker, line 362: you could use slices.Contains from golang.org/x/exp/slices, if you wanted to

server/real_environment/real_environment.go
```diff
@@ -610,3 +611,10 @@ func (r *RealEnv) GetSociArtifactStoreServer() socipb.SociArtifactStoreServer {
 func (r *RealEnv) SetSociArtifactStoreServer(s socipb.SociArtifactStoreServer) {
 	r.sociArtifactStoreServer = s
 }
+
+func (r *RealEnv) GetSingleFlightDeduper() interfaces.SingleFlightDeduper {
+	return r.singleFlightDeduper
+}
+func (r *RealEnv) SetSingleFlightDeduper(d interfaces.SingleFlightDeduper) {
```
vadimberezniker, line 618: you are net setting the real deduper in env anywhere

enterprise/server/sociartifactstore/sociartifactstore_test.go
```diff
@@ -290,24 +305,40 @@ func setup(t *testing.T) (*testenv.TestEnv, *SociArtifactStore, containerRegistr
...
 	go func() { _ = server.Serve(lis) }()
-	return r
+	return &registry
 }
 
 type containerRegistry struct {
 	host string
 	port int
+
+	// List of blobs which were requested using GET
```
vadimberezniker, line 340: comment out of date?

vadimberezniker (changes requested):

enterprise/server/sociartifactstore/sociartifactstore_test.go
iain-macdonald: Yup, fixed

enterprise/server/sociartifactstore/sociartifactstore_test.go
iain-macdonald: It's not needed, I just added it to be slightly more consistent with the deduper contract but am happy to just remove it instead.

enterprise/server/cmd/server/main.go
```diff
@@ -302,5 +303,7 @@ func main() {
 		log.Fatalf("%v", err)
 	}
 
+	realEnv.SetSingleFlightDeduper(dsingleflight.New(realEnv.GetDefaultRedisClient()))
```
vadimberezniker, line 306: redis is optional so you may want to make this conditional. we typically do this by adding a `Register(env)` function in the component itself.

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -105,7 +102,6 @@ func NewSociArtifactStore(env environment.Env) (error, *SociArtifactStore) {
 		cache:     env.GetCache(),
 		blobstore: env.GetBlobstore(),
 		env:       env,
```
vadimberezniker, line 104: you'll want to skip setting up the artifact store if deduper is not available in the environment (if redis is not configured)

vadimberezniker (changes requested):

enterprise/server/cmd/server/main.go
iain-macdonald: oh good call, fixed

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: fixed

vadimberezniker (approved):

PR #4035, opened 2023-05-22
Collapse the System Status row in Grafana.

vadimberezniker:
The collapsed/expanded state is saved as part of the dashboard config.

**Related issues**: N/A

### DIFF
--- a/tools/metrics/grafana/dashboards/buildbuddy.json
+++ b/tools/metrics/grafana/dashboards/buildbuddy.json
@@ -41,7 +41,297 @@
         "y": 0
       },
       "id": 23,
-      "panels": [],
+      "panels": [
+        {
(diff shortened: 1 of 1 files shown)


bduffany:
hmm, I guess `process_dashboard.py` isn't working as intended. There's some logic in there to set `collapsed: true` but it looks like it didn't apply to this panel for some reason, maybe because it is nested within another structure.

vadimberezniker:
> hmm, I guess `process_dashboard.py` isn't working as intended. There's some logic in there to set `collapsed: true` but it looks like that's not working (anymore?)

Oh yeah I completely forgot you added that.

vadimberezniker:
I merged this as part of https://github.com/buildbuddy-io/buildbuddy/pull/4037

PR #4039, opened 2023-05-22
Store the image-to-SOCI-index mapping in the cache instead of blostore

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/sociartifactstore/sociartifactstore.go
+++ b/enterprise/server/sociartifactstore/sociartifactstore.go
@@ -249,7 +249,7 @@ func (s *SociArtifactStore) GetArtifacts(ctx context.Context, req *socipb.GetArt
 	// apps a bunch of parallel work.
 	workKey := fmt.Sprintf("soci-artifact-store-image-%s", configHash.Hex)
 	respBytes, err := s.deduper.Do(ctx, workKey, func() ([]byte, error) {
-		resp, err := s.getArtifactsFromCache(ctx, configHash.Hex)
+		resp, err := s.getArtifactsFromCache(ctx, configHash)
 		if status.IsNotFoundError(err) {
 			sociIndexDigest, ztocDigests, err := s.pullAndIndexImage(ctx, targetImageRef, configHash, req.Credentials)
 			if err != nil {
--- a/enterprise/server/sociartifactstore/sociartifactstore.go
+++ b/enterprise/server/sociartifactstore/sociartifactstore.go
@@ -13,8 +13,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/awslabs/soci-snapshotter/compression"
@@ -49,7 +47,7 @@ import (
 
 var (
 	layerStorage       = flag.String("soci_artifact_store.layer_storage", "/tmp/", "Directory in which to store pulled container image layers for indexing by soci artifact store.")
-	sociIndexCacheSeed = flag.String("soci_artifact_store.cache_seed", "socicache-05222023", "If set, this seed is hashed with container image IDs to generate cache keys storing soci indexes.")
+	sociIndexCacheSeed = flag.String("soci_artifact_store.cache_seed", "socicache-05262023", "If set, this seed is hashed with container image IDs to generate cache keys storing soci indexes.")
 )
 
 const (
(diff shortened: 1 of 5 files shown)


enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -281,36 +275,44 @@ func (s *SociArtifactStore) GetArtifacts(ctx context.Context, req *socipb.GetArt
...
-func resourceName(digest *repb.Digest) *rspb.ResourceName {
+func resourceName(digest *repb.Digest, cacheType rspb.CacheType) *rspb.ResourceName {
 	return &rspb.ResourceName{
 		Digest:         digest,
 		Compressor:     repb.Compressor_IDENTITY,
-		CacheType:      rspb.CacheType_CAS,
+		CacheType:      cacheType,
 		DigestFunction: repb.DigestFunction_SHA256,
 	}
 }
```
tylerwilliams, line 293: could we use digest.NewResourceName directly here instead?

func NewResourceName(d *repb.Digest, instanceName string, cacheType rspb.CacheType, digestFunction repb.DigestFunction_Value) *ResourceName {...}

That way we'd be less likely to end up with a partially-filled-out rspb.ResourceName proto.

Even if you just called it from this method, would be an improvement:

```
func resourceName(digest *repb.Digest, cacheType rspb.CacheType) *rspb.ResourceName {
  return digest.NewResourceName(d, "" /*instnaceName*/, cacheType, repb.DigestFunction_SHA256).ToProto()
}
```

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -319,15 +321,15 @@ func (s *SociArtifactStore) getArtifactsFromCache(ctx context.Context, imageId s
...
 		}
 		if !exists {
 			return nil, status.NotFoundErrorf("ztoc %s not found in cache", ztocDigest.Hash)
 		}
 	}
-	return getArtifactsResponse(imageId, sociIndexDigest, ztocDigests), nil
+	return getArtifactsResponse(imageConfigHash, sociIndexDigest, ztocDigests), nil
 }
 
 // Serializes a repb.Digest as "<digest>/<size-bytes">
```
tylerwilliams, line 335: We could replace `serializeDigest` and `deserializeDigest` with digest.ResourceName().DownloadString and digest.func ParseDownloadResourceName(resourceName string)

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -281,36 +275,44 @@ func (s *SociArtifactStore) GetArtifacts(ctx context.Context, req *socipb.GetArt
...
 		Compressor:     repb.Compressor_IDENTITY,
-		CacheType:      rspb.CacheType_CAS,
+		CacheType:      cacheType,
 		DigestFunction: repb.DigestFunction_SHA256,
 	}
 }
 
-func (s *SociArtifactStore) getArtifactsFromCache(ctx context.Context, imageId string) (*socipb.GetArtifactsResponse, error) {
-	exists, err := s.blobstore.BlobExists(ctx, blobKey(imageId))
+func (s *SociArtifactStore) getArtifactsFromCache(ctx context.Context, imageConfigHash v1.Hash) (*socipb.GetArtifactsResponse, error) {
```
tylerwilliams, line 295: b/c this is only a couple of cache lookups I think it will be pretty fast, but mostly for FYI:

Another approach here, instead of using the environment cache object directly, would be to use a client pointing to the cache and actually store your data as repb.OutputFile / repb.ActionResult via the Bytestream Read/WRite and UpdateActionResult() RPCs, similar to how Bazel would.

Then this code could call GetActionResult and the server would automatically check the existence of all of the underlying files in the directory tree for you. Probably not important for this use case but just putting it out there :)

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Oh shoot, forgot about that, thanks

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Sounds good, done

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Hmm, good point. I added a TODO describing this a little bit, but I'll leave the implementation as is for now.

vadimberezniker (approved):

PR #4040, opened 2023-05-22
Add a flag that enables/disables running the soci snapshotter when it's needed

iain-macdonald:
I keep accidentally committing a change commenting out the call to this function when I'm developing locally.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -54,6 +54,7 @@ var (
 	cpuUsagePathTemplate = flag.String("executor.podman.cpu_usage_path_template", "/sys/fs/cgroup/cpuacct/libpod_parent/libpod-{{.ContainerID}}/cpuacct.usage", "Go template specifying a path pointing to a container's total CPU usage, in CPU nanoseconds. Templated with `ContainerID`.")
 
 	// TODO(iain): delete the streamableImages flag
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -131,7 +132,9 @@ func runSociStore(ctx context.Context) {
 
 func NewProvider(env environment.Env, imageCacheAuthenticator *container.ImageCacheAuthenticator, buildRoot string) (*Provider, error) {
 	if *imageStreamingEnabled || len(*streamableImages) > 0 {
-		go runSociStore(env.GetServerContext())
+		if *runSociSnapshotter {
```
vadimberezniker, line 135: Why is it a problem for a local development? Instead of running soci snapshotter manually could you not let the executor start it instead? Would that solve your problem?

vadimberezniker (changes requested):

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: I meant for developing the snapshotter locally and running a custom version without putting it into /usr/bin.

vadimberezniker: Would it solve your problem if you made the path to soci-store a flag so you can set it locally to whatever location you want?

vadimberezniker (approved):

PR #4043, opened 2023-05-23
Track approximate number of items in each group to enable AC sampling.

vadimberezniker:
- To approximate the number of items in a group, we calculate the distances between adjacent keys in that group and then divide the total range by the distance. We then take the median distance as a rough approximation of the group item count.

 - Every 5 seconds, sample a random group and update its approximate item count.

 - On startup, sample 10 groups so that we have some data to work with from the beginning.

 - Maintain a list of groups, sorted by item count.

 - During eviction sampling, pick a random number between 0 and the total group item count and use that number to pick a victim group. In this PR, we simply increment a counter to keep track of how often each group is picked so we can verify that this approach works as expected.

**Related issues**: N/A


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2401,7 +2610,139 @@ func (e *partitionEvictor) partitionKeyPrefix() string {
...
+			select {
+			case <-quitChan:
+				return
+			case <-time.After(*groupIDSampleFrequency):
+				if e.versionGetter.minDatabaseVersion() < filestore.Version2 {
+					continue
+				}
+				e.updateGroupIDApproxCounts()
+
+				// temporary -- log the approximate counts so that we can verify
```
tylerwilliams, line 2731: is this >80 chars? could just be gh

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1911,6 +1926,12 @@ type partitionEvictor struct {
 	casCount  int64
 	acCount   int64
 
+	// Total, approximate number of items stored under sampled group IDs.
+	groupIDApproxTotalCount int64
+	// List of sampled group IDs and their approximate item counts, sorted by
```
tylerwilliams, line 1931: nit: newline before comment

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1966,10 +1987,169 @@ func newPartitionEvictor(part disk.Partition, fileStorer filestore.Store, blobDi
...
+			return 0, status.UnknownErrorf("could not parse key %q hash: %s", string(iter.Key()), err)
+		}
+		for len(hash) < 32 {
+			hash = append(hash, 0)
+		}
+		if len(hash) > 32 {
+			hash = hash[:32]
+		}
+
+		keyAsNum := big.NewInt(0).SetBytes(hash)
```
tylerwilliams, line 2128: you could throw this in its own method to shorten this function a bit

```
func keyToInt(key pebbleKey) bigInt {
 blah
}
```

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2401,7 +2610,139 @@ func (e *partitionEvictor) partitionKeyPrefix() string {
...
+		cumulativeCount := approxCount
+		if idx > 0 {
+			cumulativeCount = e.groupIDApproxCounts[idx-1].cumulativeCount + approxCount
+		}
+		w := groupIDApproxCount{groupID: randomGroupID, count: approxCount, cumulativeCount: cumulativeCount}
+		e.groupIDApproxCounts = append(e.groupIDApproxCounts[:idx], append([]groupIDApproxCount{w}, e.groupIDApproxCounts[idx:]...)...)
+		for i := idx + 1; i < len(e.groupIDApproxCounts); i++ {
+			e.groupIDApproxCounts[i].cumulativeCount += approxCount
+		}
+	}
```
tylerwilliams, line 2710: this feels kind of error prone to me, but I cannot find fault with it.

It's the combination of for loops with offset indices (sampledGroupIdx + 1, sampledGroupIdx - 1), array indexes (data[i-1] = data[i]), temp variables and deltas being applied to things.

Some simpler version might be just removing then re-sorting the entire slice when editing it?

You know better than I whether or not the performance gains warrant this complexity, but I think personally I'd prefer starting with something simpler and only adding this if it was critical, especially if this only fires every N seconds or so.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1966,10 +1987,169 @@ func newPartitionEvictor(part disk.Partition, fileStorer filestore.Store, blobDi
...
+		cumulativeCount := int64(0)
+		for i, v := range approxCounts {
+			cumulativeCount += v.count
+			approxCounts[i].cumulativeCount = cumulativeCount
+		}
+
+		pe.groupIDApproxTotalCount = approxTotalCount
+		pe.groupIDApproxCounts = approxCounts
+	}
+
```
tylerwilliams, line 2040: is it not safe to call updateGroupIDApproxCounts here yet because pe is not fully initialized? Wondering if there's a way to dedupe this code by just calling the real method 1) on startup if necessary and 2) from our periodic timer

tylerwilliams (commented):
This is a nice implementation, and I really liked the clear and descriptive CL description. Most of my comments are nits or trying to find ways to potentially simplify a little bit / remove code.

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: It's exactly 80 chars :)

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: Went with something simpler for now. We can revisit later.

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: I was just trying to avoid doing unnecessary work since updateGroupIDApproxCounts does some things that are not necessary when we are populating the list for the first time, but i guess it doesn't really matter so I simplified it.

tylerwilliams (approved):

PR #4050, opened 2023-05-24
Fix firecracker test that won't run on the executor

maggie-lou:
This test tries to test the network connection of a firecracker VM by pinging the default route. Pinging the default route fails even when run from the executor pod itself. This may have to do with the networking setup on executors. Change the test to ping something else, because that's what the test is ultimately trying to test

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/BUILD
+++ b/enterprise/server/remote_execution/containers/firecracker/BUILD
@@ -93,7 +93,6 @@ go_test(
         "//server/util/disk",
         "//server/util/fileresolver",
         "//server/util/log",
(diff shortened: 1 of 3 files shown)


maggie-lou:
* master
  * **PR #4050** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4050" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4050?utm_source=stack-comment).

bduffany (approved):

PR #4054, opened 2023-05-24
Throw an error and disconnect the invocation when tag list is too long for mysql

jdhollen:
This will prevent us from storing any incomplete tags.  Doesn't seem worth blocking a build to me, but I'm willing to be convinced.

**Related issues**: N/A


tylerwilliams:
I think I'm ok failing here? Seems better to fail right away than to truncate user tag data and then have users wonder why it's broken / missing later on.

jdhollen:
> I think I'm ok failing here? Seems better to fail right away than to truncate user tag data and then have users wonder why it's broken / missing later on.

Sure, updated it to fail.

tylerwilliams (approved):

PR #4056, opened 2023-05-25
Make target names selectable for copying

siggisim:
Without this, trying to select a target name causes you to "drag" the link

### DIFF
--- a/app/invocation/invocation_targets_card.tsx
+++ b/app/invocation/invocation_targets_card.tsx
@@ -68,10 +68,7 @@ export default class TargetsCardComponent extends React.Component<Props, State> 
               {events
                 .slice(0, (this.props.pageSize && this.state.numPages * this.props.pageSize) || undefined)
                 .map((target) => (
-                  <Link
-                    className="target-row"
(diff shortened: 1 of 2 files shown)


app/invocation/invocation_targets_card.tsx
```diff
@@ -68,7 +68,10 @@ export default class TargetsCardComponent extends React.Component<Props, State>
               {events
                 .slice(0, (this.props.pageSize && this.state.numPages * this.props.pageSize) || undefined)
                 .map((target) => (
-                  <Link className="target-row" href={`?target=${encodeURIComponent(target.id.targetCompleted.label)}`}>
+                  <Link
+                    className="target-row"
+                    draggable={false}
```
bduffany, line 73: is `draggable={false}` needed? on Linux at least, it looks like `user-select: text;` does the trick on its own

bduffany (approved):

app/invocation/invocation_targets_card.tsx
siggisim: You're right, removed

PR #4059, opened 2023-05-25
Skip DB lookup of impossible API keys.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -15,6 +15,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/authdb/authdb.go
```diff
@@ -13,6 +14,10 @@ import (
 	grpb "github.com/buildbuddy-io/buildbuddy/proto/group"
 )
 
+const (
+	maxAPIKeyLength = 20
```
tylerwilliams, line 18: could you add a comment / link to where keys are generated?

tylerwilliams (approved):

PR #4060, opened 2023-05-25
Wait for and print BES errors when `sync` is set on CLI runs.

jdhollen:
The previous behavior waits for the upload to finish, but the errors aren't propagated back to the CLI.  With this change, error output for bad API keys, messed up metadata, etc. can be printed client-side.

**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/build_event_server/build_event_server.go
+++ b/server/build_event_protocol/build_event_server/build_event_server.go
@@ -143,6 +143,7 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 		case err := <-errCh:
 			if err == io.EOF {
 				if s.synchronous {
+					// Close the streams early so that we can Wait() for any forwarding errors.
 					closeStreamsOnce.Do(func() { closeForwardingStreams(forwardingStreams) })
 					if err := eg.Wait(); err != nil {
 						return disconnectWithErr(err)
--- a/server/build_event_protocol/build_event_server/build_event_server.go
+++ b/server/build_event_protocol/build_event_server/build_event_server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"sort"
+	"sync"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
@@ -106,6 +107,8 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 		})
 		forwardingStreams = append(forwardingStreams, fwdStream)
 	}
+	var closeStreamsOnce sync.Once
+	defer closeStreamsOnce.Do(func() { closeForwardingStreams(forwardingStreams) })
 
 	disconnectWithErr := func(e error) error {
 		if channel != nil && streamID != nil {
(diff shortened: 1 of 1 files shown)


server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -136,9 +136,16 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 	for {
 		select {
 		case <-channelDone:
+			closeForwardingStreams(forwardingStreams)
 			return disconnectWithErr(status.FromContextError(channel.Context()))
 		case err := <-errCh:
+			closeForwardingStreams(forwardingStreams)
 			if err == io.EOF {
+				if s.synchronous {
```
jdhollen, line 144: at face value it seems correct (and doesn't seem to be breaking anything) to wait here even though we've already gotten an EOF for the stream, but I was wondering if there was anything I should be worried about re: disconnecting + finalizing with an error when this stream has already effectively finished?

server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -136,9 +136,16 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
...
+					if err := eg.Wait(); err != nil {
```
luluz66, line 145: This seems to introduce a behavior change on the build event server. Forwarding streams are best effort only; and with this change we are going to fail the build with forwarding errors. Are we cool with that?

jdhollen: @tylerwilliams @bduffany any thoughts on this? There's another place lower in this file as well where we're swallowing errors, too.  If we don't want to do this, I don't see a super obvious way to plumb this info back without building an extra side-channel explicitly for this purpose, since this is just a plain ol' streaming RPC.

luluz66: oh. You are right. I missed that.

bduffany (approved):
nice

server/build_event_protocol/build_event_server/build_event_server.go
jdhollen: Okay, so, an update for you all @luluz66 @bduffany : I made this fail if forwarding fails _in synchronous mode only_ down around L172.  It's kind of difficult to test this, if i'm not totally confused?  Do you think this failure mode is overkill or does it make sense to you?  I'm trying to make things consistent but don't want to make building crazy brittle as a result.

bduffany: This seems alright to me, though I'd prefer if we could find a way to make sure that we can't forget to call `closeForwardingStreams` as this code evolves. The current approach of calling it in a bunch of places seems a bit brittle - maybe we could somehow use `defer`, possibly with a wrapper func?

server/build_event_protocol/build_event_server/build_event_server.go
bduffany: Hm I'm not sure I understand the sequence you're describing - could you elaborate?

server/build_event_protocol/build_event_server/build_event_server.go
jdhollen: I'll try to think of something clever to do here, because I agree, but the issue is that we need to directly call `eg.Wait()` in order to process and return errors, but also need the forwarding streams to be closed first.  I'm trying to find proof in documentation that calling `CloseSend` twice is OK.

jdhollen: Oh, duh, how about this--used sync.Once and now i'm only calling early in the case where it's necessary.

PR #4061, opened 2023-05-26
prevent exponentially multiplying setIntervals on pending build cards

jdhollen:
didn't bother looking up when this happened, but i noticed that my browser tabs were hitting 100% cpu, and yeah, after a minute or something the main js thread is basically constantly spinning doing this--looks like we treated a setinterval as a settimeout by accident.

**Related issues**: N/A


siggisim (approved):

siggisim:
Nice find

PR #4065, opened 2023-05-26
DNM - Testing for segfault

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/executor/executor.go
+++ b/enterprise/server/remote_execution/executor/executor.go
@@ -204,10 +204,10 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
 	if err != nil {
 		return finishWithErrFn(status.WrapErrorf(err, "error creating runner for command"))
 	}
-	//finishedCleanly := false
-	//defer func() {
-	//	go s.runnerPool.TryRecycle(ctx, r, finishedCleanly)
-	//}()
+	finishedCleanly := false
+	defer func() {
+		go s.runnerPool.TryRecycle(ctx, r, finishedCleanly)
+	}()
 
 	log.CtxInfof(ctx, "Preparing runner for task.")
 	stage.Set("pull_image")
@@ -351,7 +351,7 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
 	}
 	if cmdResult.Error == nil {
 		log.CtxInfof(ctx, "Task finished cleanly.")
-		//finishedCleanly = true
+		finishedCleanly = true
 	}
 	return false, nil
 }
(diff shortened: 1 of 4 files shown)


maggie-lou:
* master
  * **PR #4065** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4065" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4065?utm_source=stack-comment).

repeat.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+while true
```
bduffany, line 2: nit: maybe limit this to 100 or something :sweat_smile:

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -204,10 +204,10 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
...
 		return finishWithErrFn(status.WrapErrorf(err, "error creating runner for command"))
 	}
-	finishedCleanly := false
-	defer func() {
-		go s.runnerPool.TryRecycle(ctx, r, finishedCleanly)
-	}()
+	//finishedCleanly := false
+	//defer func() {
+	//	go s.runnerPool.TryRecycle(ctx, r, finishedCleanly)
+	//}()
```
bduffany, line 210: We always want to call `TryRecycle()` otherwise the VMs will accumulate and the executor will crash eventually.

What is your goal here? (I think if you want to test whether it happens on recycled runners, you could revert the changes to this file and let the runner be recycled normally)

repeat.sh
maggie-lou: Yeah... A little scary. I feel like there may need to be some volume to trigger this though because it's pretty rare. For a 2 week period I checked logs for, there were only 14 instances. I had previously tried to trigger it with a script with a 1000 for loop, and it never caused it. To be safer though I can increase the max random time so it's not hammering our executors too hard

enterprise/server/remote_execution/executor/executor.go
maggie-lou: Ah I see that TryRecycle needs to be called for runner.finalize(). The goal is for the workflow to always be executed on a clean runner, so I don't want any runners to be recycled back into the pool

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -1333,6 +1333,12 @@ func (p *pool) TryRecycle(ctx context.Context, r interfaces.Runner, finishedClea
 	if !cr.PlatformProperties.RecycleRunner {
 		return
 	}
+
+	exitEarly := true
+	if exitEarly {
+		return
+	}
```
bduffany, line 1340: IIUC if you're never recycling, then the `bazel clean --expunge` will have no effect?

maggie-lou: Oh yeah that's a good point. I guess there are a couple different experiments to run:

1. Do not care if runner is recycled or not. Run `clean --expunge` every time. Dump kernel logs and see if there's anything interesting
2. Ensure we can trigger the segfault if we're always running on clean runners. Then try to trigger it on a recycled runner. See if that's possible or not

maggie-lou: I'll start with #1 and if that doesn't provide any helpful info go on to #2

PR #4069, opened 2023-05-30
GetCapabilities: set HighApiVersion to 2.3

sluongng:
The latest version of Remote APIs is v2.3.

Let's make sure that we follow the specification and avoid future v2.4
or v3.0 clients to try using our API before we support it.

**Related issues**: https://github.com/bazelbuild/remote-apis/issues/253


bduffany (approved):

tylerwilliams:
In the past, the way we've known to add support for new features is when folks try them and find issues. With this change, we now need to stay on top of that by monitoring for new features and adding them -- I'm not opposed to this change, but I want to make sure that is being covered.

sluongng:
Agree. We have been attending the RE API meetings to stay on top of things.

PR #4070, opened 2023-05-30
util/disk: make implementation Windows friendly

sluongng:
Provide a Windows implementation to check directory's size.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2314

### DIFF
--- a/server/util/disk/disk_windows.go
+++ b/server/util/disk/disk_windows.go
@@ -1,0 +1,27 @@
+//go:build windows
+
+package disk
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/windows"
+)
+
+func GetDirUsage(path string) (*DirUsage, error) {
+	cwd, err := windows.UTF16PtrFromString(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to call UTF16PtrFromString: %v", err)
+	}
+
+	var freeBytesAvailableToCaller, totalNumberOfBytes, totalNumberOfFreeBytes uint64
+	if err := windows.GetDiskFreeSpaceEx(cwd, &freeBytesAvailableToCaller, &totalNumberOfBytes, &totalNumberOfFreeBytes); err != nil {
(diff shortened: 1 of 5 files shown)


server/util/disk/disk_windows.go
```diff
@@ -0,0 +1,27 @@
...
+	var freeBytesAvailableToCaller, totalNumberOfBytes, totalNumberOfFreeBytes uint64
+	if err := windows.GetDiskFreeSpaceEx(cwd, &freeBytesAvailableToCaller, &totalNumberOfBytes, &totalNumberOfFreeBytes); err != nil {
+		return nil, err
+	}
+	return &DirUsage{
+		TotalBytes: totalNumberOfBytes,
+		UsedBytes:  totalNumberOfBytes - totalNumberOfFreeBytes,
+		FreeBytes:  totalNumberOfFreeBytes,
+		AvailBytes: freeBytesAvailableToCaller,
+	}, nil
```
bduffany, line 26: sanity check: did you look at the return value here and make sure it looks reasonable?

bduffany (approved):

server/util/disk/disk_windows.go
sluongng: I am trying to provision a Windows machine (locally or on the cloud) to verify this.

Let's convert this to draft to prevent it from merging until then.

sluongng: This is the result when pointed to a subdir in `D:\\` drive

```
&{TotalBytes:199996993536 UsedBytes:4949659648 FreeBytes:195047333888 AvailBytes:195047333888}
```

With `D:` being like this
image

So I think it's consistent with the result on UNIX.
Added a test to make sure.

bduffany: Nice, thanks for checking this :+1:

bduffany (approved):

PR #4078, opened 2023-05-30
Change `default:false` to `default:0`

tempoz:
**Related issues**: N/A

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -220,13 +220,13 @@ type Group struct {
 	Model
 
 	SharingEnabled       bool `gorm:"default:true"`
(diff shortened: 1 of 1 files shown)


server/tables/tables.go
```diff
@@ -220,13 +220,13 @@ type Group struct {
 	Model
 
 	SharingEnabled       bool `gorm:"default:true"`
```
vadimberezniker, line 222: SharingEnabled too but `1`

vadimberezniker (approved):

server/tables/tables.go
tempoz:
```suggestion
	SharingEnabled       bool `gorm:"default:1"`
```

PR #4080, opened 2023-05-30
Make group ID sampling work for ANON data.

vadimberezniker:
Keys for ANON data don't follow the normal key format so they won't be found when sampling keys with the "GR" prefix.

To make the process work for ANON keys, we make up a random pseudo group ID for anon keys that's only used when sampling group IDs. When sampling, we check if we would have hit this group ID had it really existed in the key space and if so return the ANON group as the sampled group id.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1934,6 +1934,10 @@ type partitionEvictor struct {
 	// item count.
 	groupIDApproxCounts []groupIDApproxCount
 
+	// A random pseudo group ID we assign for ANON data so that we can sample
+	// it like any other group ID.
+	anonPseudoGroupID string
+
 	atimeBufferSize int
 	minEvictionAge  time.Duration
 }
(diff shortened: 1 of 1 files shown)


tylerwilliams:
I'd prefer to just migrate this data to a known ANON group ID and let regular old eviction do it's thing. That cool with you?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2029,6 +2034,14 @@ func (e *partitionEvictor) sampleGroupID() (string, error) {
 		return "", err
 	}
 
+	// The special ANON group doesn't follow the regular group key format, so it
+	// won't be found through regular sampling. To get around this, we make up
+	// a random group ID for it and pretend it exists in the keyspace so that
+	// it can be sampled like any other group.
+	if randomGroupID < e.anonPseudoGroupID && (key.GroupID() == "" || key.GroupID() > e.anonPseudoGroupID) {
```
tylerwilliams, line 2041: could you add a test for this functionality?

tylerwilliams (approved):
(talked offline, going to submit this and remove later if necessary)

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: The test in https://github.com/buildbuddy-io/buildbuddy/pull/4091/files already uses anon data. You could make an argument that we should be including non-anon data there as well.

tylerwilliams (approved):

PR #4082, opened 2023-05-30
Delete unused image converter and registry code.

vadimberezniker:
**Related issues**: N/A


iain-macdonald:
Few things in deps.bzl too if you don't mind cleaning those up as well

iain-macdonald (approved):

vadimberezniker:
> Few things in deps.bzl too if you don't mind cleaning those up as well

Anything obvious? Nothing is being removed from `go.mod` so I assume nothing should be removed from `deps.bzl`.

PR #4085, opened 2023-05-31
workflows: verify cross compiling Executor to target arm64

sluongng:
We do have self-hosted customers running Executors on Arm64.

Let's ensure that we could validate that at minimum, our code could
compile successfully in CI by applying a transition on the executor
binary target.

**Related issues**: N/A

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -10,6 +10,17 @@ actions:
           - "master"
     bazel_commands:
       - test //... --config=linux-workflows --test_tag_filters=-performance,-webdriver,-docker --build_metadata=TAGS=linux-workflow
+  - name: Build executor (linux_arm64)
+    container_image: ubuntu-20.04
+    triggers:
+      push:
+        branches:
+          - "master"
+      pull_request:
+        branches:
+          - "master"
+    bazel_commands:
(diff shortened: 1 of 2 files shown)


buildbuddy.yaml
```diff
@@ -10,6 +10,17 @@ actions:
...
+    container_image: ubuntu-20.04
+    triggers:
+      push:
+        branches:
+          - "master"
+      pull_request:
+        branches:
+          - "master"
+    bazel_commands:
+      - build //enterprise/server/cmd/executor:executor_arm64 --config=linux-workflows
```
bduffany, line 23: I think this should already be built as part of the "Test" workflow, no? Since we're running `test //...` and aren't setting `--build_tests_only`

bduffany (commented):
wow, I'm surprised this just works :open_mouth: nice!

sluongng:
However, after tinkering with this a bit more, I realized that we do have CGO specific deps that would have problem with this approach. So this PR either won't work at all, or gona add some burden onto current development workflow.

https://github.com/buildbuddy-io/buildbuddy-internal/issues/2452 would be a much better alternative to solve this.

PR #4089, opened 2023-05-31
Add a handful of metrics for tracking podman image pull performance with/without image streaming enabled

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -178,6 +178,9 @@ const (
 
 	// Name of service the health check is running for (Ex "distributed_cache" or "sql_primary").
 	HealthCheckName = "health_check_name"
+
+	/// Container image tag.
+	ContainerImageTag = "container_image_tag"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


server/metrics/metrics.go
```diff
@@ -2024,6 +2027,35 @@ var (
...
+	})
+
+	PodmanColdImagePullLatencyMsec = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: bbNamespace,
+		Subsystem: "podman",
+		Name:      "image_pull_latency_msec",
+		Buckets:   durationMsecBuckets(1*time.Millisecond, 100*time.Minute, 10),
+		Help:      "The latency of 'cold' podman pull requests per image, in milliseconds. 'Cold' means the image hasn't been pulled by this executor previously.",
+	}, []string{
+		ContainerImageTag,
```
vadimberezniker, line 2057: It's generally not advised to put dynamic data into labels since a basic prometheus setup will not really scale well to handle that. What are you trying to get out of this metric? i.e. what will knowing the cold pull time give you?

vadimberezniker (changes requested):

server/metrics/metrics.go
iain-macdonald: Yeah, I saw that and chatted with Tyler about it today. We don't believe this data will be too problematic to use as a label (for now) as we don't expect there to be many more than ~100 images in use and we should remove this if that number changes significantly. I want to keep track of the image pull latency per-image to observe the effect image streaming has on image pull latency and keep an eye on it for a little while until we're comfortable everything is working correctly.

If you're super concerned about it I could instead hash and bucket the image tag into a fixed set of labels, though we would lose the ability to reverse from bucket to image tag in that case, which isn't the end of the world.

tylerwilliams (approved):
LGTM (assuming vadim is OK with the image label)

vadimberezniker (approved):

PR #4090, opened 2023-05-31
strictify app/invocation: everything else

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation_artifacts_card.tsx
+++ b/app/invocation/invocation_artifacts_card.tsx
@@ -19,14 +19,14 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
     numPages: 1,
   };
 
-  handleArtifactClicked(outputUri: string, outputFilename: string, event: MouseEvent) {
+  handleArtifactClicked(outputUri: string, outputFilename: string, event: React.MouseEvent<HTMLAnchorElement>) {
     event.preventDefault();
     if (!outputUri) return false;
 
     if (outputUri.startsWith("file://")) {
       window.prompt("Copy artifact path to clipboard: Cmd+C, Enter", outputUri);
     } else if (outputUri.startsWith("bytestream://")) {
-      rpcService.downloadBytestreamFile(outputFilename, outputUri, this.props.model.getId());
+      rpcService.downloadBytestreamFile(outputFilename, outputUri, this.props.model.getId() ?? "");
     }
   }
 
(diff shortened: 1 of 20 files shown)


jdhollen:
the downside is the change is big, the upside is this is the last one

app/invocation/invocation_artifacts_card.tsx
```diff
@@ -19,14 +19,14 @@ export default class ArtifactsCardComponent extends React.Component<Props, State
...
-  handleArtifactClicked(outputUri: string, outputFilename: string, event: MouseEvent) {
+  handleArtifactClicked(outputUri: string, outputFilename: string, event: React.MouseEvent<HTMLAnchorElement>) {
     event.preventDefault();
     if (!outputUri) return false;
 
     if (outputUri.startsWith("file://")) {
       window.prompt("Copy artifact path to clipboard: Cmd+C, Enter", outputUri);
     } else if (outputUri.startsWith("bytestream://")) {
-      rpcService.downloadBytestreamFile(outputFilename, outputUri, this.props.model.getId());
+      rpcService.downloadBytestreamFile(outputFilename, outputUri, this.props.model.getId() ?? "");
```
iain-macdonald, line 29: should probably skip this if getId() is undefined no?

rules/typescript/index.bzl
```diff
@@ -10,12 +10,9 @@ def _swc(**kwargs):
     )
 
 def ts_library(name, srcs, strict = True, **kwargs):
```
iain-macdonald, line 12: mind removing the parameter too?

bduffany (approved):

rules/typescript/index.bzl
```diff
@@ -10,12 +10,9 @@ def _swc(**kwargs):
...
-    tsconfig = "//:tsconfig"
-    if strict:
-        tsconfig = "//:tsconfig_strict"
     ts_project(
         name = name,
-        tsconfig = tsconfig,
+        tsconfig = "//:tsconfig_strict",
```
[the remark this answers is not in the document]
jdhollen: I reverted this, I'll send a separate change that does the cleanup on the rule.

app/invocation/invocation_artifacts_card.tsx
jdhollen: So, yeah.  We have a lot of these now.  The production behavior of this today builds a URL containing "undefined" (i.e., this path is already unprotected).  Making interactions with invocation_model a bit more rigorous should probably be at the top of the list, but this doesn't make things worse and let's us get the bits flipped, so let's keep it separate for now.

PR #4093, opened 2023-06-01
goinit: raise inotify.max_user_watches to 1% mem available

sluongng:
We have observed that for several Workflows with larger repo, the
default max_user_watches is too low for Bazel's '--watchfs' feature to
work properly.

Let's raise the default max_user_watches according to the recommendation
in a recent Linux Kernel patch (1) where the default value should be 1%
of the memory available (which is 8GB in our case).

(1): https://lore.kernel.org/all/20201109035931.4740-1-longman@redhat.com/T/#u

**Related issues**:

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2307

### DIFF
--- a/enterprise/server/cmd/goinit/main.go
+++ b/enterprise/server/cmd/goinit/main.go
@@ -346,8 +346,6 @@ func main() {
 		}
 	}
 	die(sysctl(map[string]string{
-		// Default value, but set it explicitly to be sure.
-		"fs.inotify.max_user_instances": "128",
 		// Set to 1% of the 8GB memory.
 		// See https://github.com/torvalds/linux/blob/929ed21dfdb6ee94391db51c9eedb63314ef6847/fs/notify/inotify/inotify_user.c#L838-L844
 		// for more info
--- a/enterprise/server/cmd/goinit/main.go
+++ b/enterprise/server/cmd/goinit/main.go
@@ -58,6 +58,21 @@ func die(err error) {
 	}
 }
 
+func sysctl(conf map[string]string) error {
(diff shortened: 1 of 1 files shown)


bduffany (approved):

enterprise/server/cmd/goinit/main.go
```diff
@@ -330,6 +345,14 @@ func main() {
 			die(err)
 		}
 	}
+	die(sysctl(map[string]string{
+		// Default value, but set it explicitly to be sure.
+		"fs.inotify.max_user_instances": "128",
```
tylerwilliams, line 350: I think this is already 128 in our firecracker VMs, so probably ok to remove this?

Verified by running vmstart (`blaze run enterprise/tools/vmstart`) and:
```
~ # sysctl fs.inotify.max_user_instances
fs.inotify.max_user_instances = 128
~ #
```

tylerwilliams (approved):

sluongng:
Using the change in https://github.com/buildbuddy-io/buildbuddy/pull/4126, I was able to test this locally and verified it's working as intended.

```bash
~/work/misc/test> bazel build :all
...
INFO: From Executing genrule //:temp:
foo
128
65536
bar
Target //:temp up-to-date:
  bazel-bin/temp.txt
...
INFO: Build completed successfully, 2 total actions

~/work/misc/test> cat BUILD
genrule(
    name = "temp",
    outs = ["temp.txt"],
    cmd = """
    echo foo;
    cat /proc/sys/fs/inotify/max_user_instances;
    cat /proc/sys/fs/inotify/max_user_watches;
    echo bar;
    touch $@
    """,
)
```

PR #4099, opened 2023-06-01
Add secret expansion to config files (enterprise only) - try2.

vadimberezniker:
Any flag values in format ${SECRET:foo} are loaded from the configured secret store. Initially only GCP is supported.

In this version, we perform the expansion after YAML unmarshalling so that multi-line secrets work properly.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/configsecrets/configsecrets.go
+++ b/enterprise/server/backends/configsecrets/configsecrets.go
@@ -19,17 +19,29 @@ import (
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 )
 
+const (
+	configSecretProviderFlagName            = "config_secrets.provider"
+	configSecretsGCPProjectFlagName         = "config_secrets.gcp.project_id"
+	configSecretsGCPCredentialsFileFlagName = "config_secrets.gcp.credentials_file"
+)
+
 var (
 	// Note that all the flags here must be set on the command line and not via
 	// a config file since config file parsing depends on secret integration
 	// being already configured.
 
-	configSecretProvider = flag.String("config_secrets.provider", "", "Secrets provider to use for config variable substitution. Currently only 'gcp' is supported.")
+	configSecretProvider = flag.String(configSecretProviderFlagName, "", "Secrets provider to use for config variable substitution. Currently only 'gcp' is supported.")
 
 	// GCP flags.
-	configSecretsGCPProject         = flag.String("config_secrets.gcp.project_id", "", "GCP project from which secrets will be loaded.")
-	configSecretsGCPCredentialsFile = flag.String("config_secrets.gcp.credentials_file", "", "Credentials to use when communicating with the secrets store. If not specified, Application Default Credentials are used.")
+	configSecretsGCPProject         = flag.String(configSecretsGCPProjectFlagName, "", "GCP project from which secrets will be loaded.")
+	configSecretsGCPCredentialsFile = flag.String(configSecretsGCPCredentialsFileFlagName, "", "Credentials to use when communicating with the secrets store. If not specified, Application Default Credentials are used.")
 )
+
+func init() {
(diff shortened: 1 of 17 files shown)


tylerwilliams (approved):

tylerwilliams:
lgtm but i'll defer to @tempoz on flags lib changes

server/util/flagutil/yaml/yaml.go
```diff
@@ -599,6 +611,74 @@ func RemoveEmptyMapsFromYAMLMap(m map[string]any) map[string]any {
...
+		}
+		return os.Getenv(s)
+	})
+	if expandErr != nil {
+		return "", expandErr
+	}
+	return expandedValue, nil
+}
+
+func expandSliceValues(yamlSlice []any) error {
```
tempoz, line 638: Nit: if we have the map and slice expansions return `(map[string]any, error)` and `([]any, error)` and have `expandValue` return `(any, error)`, we can avoid code duplication by moving the `switch` statement into `expandValue` and reduce the body of the loops in `expandSlice` and `expandMap` to
```
ev, err := expandValue(v)
if err != nil {
	return nil, err
}
yaml<slice | map> = ev
```
If we want to keep all of the current `expandValue` logic in its own function (which makes sense to me), we could rename it to `expandString` and then have `expandValue` only do the switch logic for types, forwarding to `expandMap`, `expandSlice`, or `expandString` as appropriate.

server/util/flagutil/yaml/yaml.go
```diff
@@ -648,11 +728,8 @@ func RetypeAndFilterYAMLMap(yamlMap map[string]any, typeMap map[string]any, pref
 
 // OverrideFlagsFromData takes some YAML input and marshals it, then uses the
 // unmarshaled data to override the flags with names corresponding to the keys.
-func OverrideFlagsFromData(data []byte) error {
-	// expand environment variables
-	expandedData := []byte(os.ExpandEnv(string(data)))
-
-	yamlMap, node, err := getYAMLMapAndNodeFromData(expandedData)
+func OverrideFlagsFromData(data string) error {
```
tempoz, line 731: Since the expansion was moved to after the call to `yaml.Unmarshal`, there's no longer a compelling reason to change this to `string` from `[]byte`, since it is never used anywhere as a string, and it is used as `[]byte`. Can we switch it back?

tempoz (changes requested):

server/util/flagutil/yaml/yaml.go
vadimberezniker: Having this be a string does make it nicer to use in tests since there's no need to convert the test config from []byte to string.

tempoz: Fair enough; it's cosmetic either way, I suppose.

enterprise/server/backends/configsecrets/configsecrets.go
```diff
@@ -0,0 +1,81 @@
...
+	// Note that all the flags here must be set on the command line and not via
+	// a config file since config file parsing depends on secret integration
+	// being already configured.
+
+	configSecretProvider = flag.String("config_secrets.provider", "", "Secrets provider to use for config variable substitution. Currently only 'gcp' is supported.")
+
+	// GCP flags.
+	configSecretsGCPProject         = flag.String("config_secrets.gcp.project_id", "", "GCP project from which secrets will be loaded.")
+	configSecretsGCPCredentialsFile = flag.String("config_secrets.gcp.credentials_file", "", "Credentials to use when communicating with the secrets store. If not specified, Application Default Credentials are used.")
+)
```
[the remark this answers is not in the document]
tempoz: Yeah, I agree. I added the tag here: buildbuddy-io/buildbuddy/pull/4121

tempoz (approved):
LGTM; feel free to either wait for the YAMLIgnoreTag or I can put these flags in that PR if you don't want to wait.

PR #4100, opened 2023-06-01
Prototype to export Prometheus metrics

luluz66:
Expose /api/v1/metrics HTTP endpoint, which allows the customer to scrape the
Buildbuddy metrics with a Prometheus scraper when they add the following to
scrape_configs:
```
  - job_name: buildbuddy-exported
    authorization:
      type: "x-buildbuddy-api-key"
      credentials: <buildbuddy-api-key>
    metrics_path: "/api/v1/metrics"
    static_configs:
      - targets: ["host.docker.internal:8080"]
```

1. Added an interceptor to set `x-buildbuddy-api-key` header from the
   authorization header set by the Prometheus scraper. After that,
   the bb server will authenticate normally with `x-buildbuddy-api-key`.
2. Added `bbMetricsCollector` which implements `prometheus.Collector`
   interface. This collector is responsible for querying promQL to get the
   metrics for a given group ID; and converts to the Prometheus Metrics.
3. Added a registry manager which gets or create a registry per group
   ID. We uses the `promhttp.HandlerFor(..)` method to convert the metrics
   registered with the registry to plain text format that is parseable by
   the Prometheus scraper.
4. Added `GroupID` label to `metrics.InvocationCount`. Added the logic to set group id from the context in FinalizeInvocation. 

Future Improvement:
   - we could cache the promQL result in the Redis for 10 seconds; and
     serve the result from cache to mitigate the case where the customer
     scrape the metrics too often.

**Related issues**: N/A


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -814,6 +814,13 @@ func (e *EventChannel) FinalizeInvocation(iid string) error {
 	if err != nil {
 		return err
 	}
+	auth := e.env.GetAuthenticator()
+	userInfo, err := auth.AuthenticatedUser(e.ctx)
+	groupID := ""
+	if err == nil {
+		groupID = userInfo.GetGroupID()
+	}
+	ti.GroupID = groupID
```
luluz66, line 823: are there going to be edge cases where the group id is not set in the context? maybe it's safer to record the group id in the metrics but not updating the field in the db.

enterprise/server/backends/prom/prom.go
```diff
@@ -0,0 +1,143 @@
...
+		return status.InternalErrorf("failed to configure prom querier: %s", err)
+	}
+	env.SetPromQuerier(q)
+
+	m := newMetricsGroupRegistry(env)
+	env.SetMetricsGroupRegistries(m)
+	return nil
+}
+
+func (m *metricsGroupRegistryManager) GetOrCreateRegistry(groupID string) (*prometheus.Registry, error) {
```
tylerwilliams, line 74: I don't know what the overhead is for a registry -- but if Registry is going to call collector.Collect any time metrics are scraped then is there any value in holding on to the Registry object between requests?

(especially given that subsequent requests are very likely to hit another server)

Could you instead dynamically create a Registry when the scape endpoint is hit, calling your collector (which hopefully hits cached data some of the time)

enterprise/server/backends/prom/prom.go
```diff
@@ -0,0 +1,143 @@
...
+		registryByGroupID: sync.Map{},
+	}
+}
+
+func Register(env environment.Env) error {
+	q, err := newQuerier()
+	if err != nil {
+		return status.InternalErrorf("failed to configure prom querier: %s", err)
+	}
+	env.SetPromQuerier(q)
```
tylerwilliams, line 67: this I understand caching since it's a connection to prometheus

server/build_event_protocol/build_event_handler/build_event_handler.go
tylerwilliams: this feels like a little bit of a weird place to be doing this -- I'd expect we have groupID already in hand when we validate the context at the start of the invocation. Maybe we just need to hold on to it in the channel or something?

server/http/interceptors/interceptors.go
```diff
@@ -106,6 +106,18 @@ func Authenticate(env environment.Env, next http.Handler) http.Handler {
...
+		authHeader := r.Header.Get("Authorization")
+		// cannot use auth.APIKeyHeader because of circular dependency
+		apiKey := strings.TrimPrefix(authHeader, "x-buildbuddy-api-key ")
+		if apiKey != "" && r.Header.Get("x-buildbuddy-api-key") == "" {
+			r.Header.Set("x-buildbuddy-api-key", apiKey)
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
```
tylerwilliams, line 120: maybe a cleaner way to do this would be to accept an API key set as a bearer token in AuthenticatedHTTPContext -- so if someone made an HTTP request to use with:

`header: Authorization <api_key_key>`

We'd do the right thing.

tylerwilliams (commented):
Overall this looks really cool and not nearly as gnarly as you made me think it would be :)

My comments are basically about: trying to simplify some APIs a little bit (auth, env registry stuff), and then exploring caching or adding a rate limit so that it's not possible for any external scraper to DOS our prom server.

server/http/interceptors/interceptors.go
luluz66: Done. I added the support in AuthenticatedHTTPContext to support

`header: Authorization: x-buildbuddy-api-key <api_key>`

I chose to use x-buildbuddy-api-key, so that in the future if we ever want to use bearer token for HTTP we won't have a problem

enterprise/server/backends/prom/prom.go
luluz66: Done. I don't think NewRegistry() itself has any overhead. However reg.Register(collector) has some overhead as it is trying to validate the collector.

server/build_event_protocol/build_event_handler/build_event_handler.go
luluz66: I implemented a function getGroupIDForMetrics similar to what we did in execution server ([code](https://github.com/buildbuddy-io/buildbuddy/blob/83d8cbe404d069d193236dedf35d6654dc430973/enterprise/server/remote_execution/execution_server/execution_server.go#L743)). So that we are not updating the database.

We validated the group id inside `CreateInvocation` method. So we need to refactor it in order to put it in the channel.

enterprise/server/api/api_server.go
```diff
@@ -38,6 +40,7 @@ var (
 	enableAPI            = flag.Bool("api.enable_api", true, "Whether or not to enable the BuildBuddy API.")
 	enableCache          = flag.Bool("api.enable_cache", false, "Whether or not to enable the API cache.")
 	enableCacheDeleteAPI = flag.Bool("enable_cache_delete_api", false, "If true, enable access to cache delete API.")
+	enableMetricsAPI     = flag.Bool("enable_metrics_api", true, "If true, enable access to metrics API.")
```
vadimberezniker, line 43: should we make this api.enable_metrics_api? I'm not sure why the previous flag does not conform.

enterprise/server/api/api_server.go
```diff
@@ -456,6 +463,42 @@ func (s *APIServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func (s *APIServer) handleGetMetricsRequest(w http.ResponseWriter, r *http.Request) {
+	if !*enableMetricsAPI {
+		http.Error(w, "API not enabled", http.StatusNotImplemented)
+		return
+	}
+	userInfo, err := s.checkPreconditions(r.Context())
+	if err != nil {
+		log.Errorf("invalid api key to get metrics: %s", err)
```
vadimberezniker, line 477: Do we need to log every invalid key request?

vadimberezniker (approved):

tylerwilliams (approved):
I'm not 100% in love with the auth bearer token stuff but it's def good enough for an MVP :)

PR #4101, opened 2023-06-01
dedupe tags

jdhollen:
**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/invocation_format/invocation_format.go
+++ b/server/build_event_protocol/invocation_format/invocation_format.go
@@ -48,7 +48,7 @@ func ShortFormatPatterns(patterns []string) string {
 // of tags would be longer than 160 characters, an error is returned.
 func SplitAndTrimAndDedupeTags(tags string, validate bool) ([]*invocation.Invocation_Tag, error) {
 	if len(tags) == 0 {
-		return []*invocation.Invocation_Tag{}, nil
+		return nil, nil
 	}
 	splitTags := strings.Split(tags, ",")
 	totalLength := 0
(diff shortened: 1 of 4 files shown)


server/build_event_protocol/invocation_format/invocation_format.go
```diff
@@ -43,19 +43,21 @@ func ShortFormatPatterns(patterns []string) string {
...
-// and leading whitespace from each tag.  If validate it set to true and the
-// trimmed, comma-separated string containing the tags would be longer than 160
-// characters, an error is returned.
-func SplitAndTrimTags(tags string, validate bool) ([]*invocation.Invocation_Tag, error) {
+// and leading whitespace from each tag.  Duplicate entries are removed. If
+// validate is set to true and the trimmed, comma-separated, and deduped string
+// of tags would be longer than 160 characters, an error is returned.
+func SplitAndTrimAndDedupeTags(tags string, validate bool) ([]*invocation.Invocation_Tag, error) {
 	if len(tags) == 0 {
 		return []*invocation.Invocation_Tag{}, nil
```
tylerwilliams, line 51: I think you can return nil here and go will do the right thing
```suggestion
		return nil, nil
```

tylerwilliams (approved):

server/build_event_protocol/invocation_format/invocation_format.go
jdhollen: Ooo, fun

tylerwilliams (approved):

server/build_event_protocol/invocation_format/invocation_format.go
jdhollen: I don't like the idea of returning nil in one case but not another, but i understand the desire to just give back nil when the user doesn't specify any flags, so i added a little condition to the end of the function to make the case where the list parsed out to empty also return nil.

PR #4105, opened 2023-06-02
workflows/deps: fix building Gazelle from internal repo

sluongng:
We copy deps.bzl into internal repo for every commits.
However, some build patches which only exist the public repo does not
exist in the internal repo.  Moreover, the workspace name of the public
repo is `@buildbuddy//` while the internal repo load it as
`@com_github_buildbuddy_io_buildbuddy//`.

Ensure all patch references come with workspace prefix, which is
provided as an args for `install_buildbuddy_dependencies` macro.

**Related issues**: N/A

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -323,7 +323,7 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         # TODO: Remove once https://github.com/bazelbuild/bazelisk/pull/396 is
         # upstreamed
         patch_args = ["-p1"],
-        patches = ["//buildpatches:bazelisk.patch"],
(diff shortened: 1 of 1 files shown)


sluongng:
Note that `cd internal && bazel build :gazelle` would only work _after_ the `deps.bzl` file in this PR is copied over to internal. This should happen in a post-merge Github workflow.

If you want to test this PR locally, you would have to do the copy/paste yourself.

siggisim (approved):

siggisim:
Does this mean we can delete the internal buildpatches directory?

sluongng:
yup, let me create a subsequent PR to clean that up.

PR #4106, opened 2023-06-02
Add protos for nbd

bduffany:
The executor serves this `BlockDevice` service and the VM uses a little `BlockDevice` client as part of its local NBD server. The client talks to the executor to request reads/writes from a backing storage (which eventually will involve a copy-on-write mechanism implemented in the executor).

**Related issues**: N/A

### DIFF
--- a/proto/nbd.proto
+++ b/proto/nbd.proto
@@ -1,0 +1,77 @@
+syntax = "proto3";
+
+package nbd;
+
+// Remote block device service.
+//
+// Clients can use this service to read and write blocks from a remote device
+// that serves the underlying block data.
+//
+// Servers are responsible for writing and serving the block data to persistent
+// storage.
+service BlockDevice {
+  // Returns metadata for a device label. The label is a well-known value by
+  // both the client and server that corresponds to a desired device.
+  rpc Metadata(MetadataRequest) returns (MetadataResponse);
+
+  // Reads a block from a device with a given ID.
+  rpc Read(ReadRequest) returns (ReadResponse);
+
+  // Writes a block to a device with a given ID.
+  rpc Write(WriteRequest) returns (WriteResponse);
+}
+
+message DeviceMetadata {
+  // The device ID to be included in IO requests.
+  uint64 device_id = 1;
+
+  // The device label used to identify the device in metadata requests.
+  // Ex: "scratchfs"
+  string label = 2;
+
+  // The size of the block device in bytes.
+  uint64 size_bytes = 3;
+
+  // The filesystem type to be used in mount(2) operations, such as "ext4".
+  string filesystem_type = 4;
+}
+
+message MetadataRequest {
+  // The unique label of the device to fetch metadata for.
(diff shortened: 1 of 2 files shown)


proto/nbd.proto
```diff
@@ -0,0 +1,77 @@
...
+  // The stored data with the requested offset and length.
+  bytes data = 1;
+}
+
+message WriteRequest {
+  // The device to read from.
+  uint64 device_id = 1;
+
+  // The offset at which to write the data.
+  uint64 offset = 2;
```
tylerwilliams, line 71: int64

proto/nbd.proto
```diff
@@ -0,0 +1,77 @@
...
+
+message ReadRequest {
+  // The ID of the device to read from.
+  uint64 device_id = 1;
+
+  // The offset at which to start reading.
+  uint64 offset = 2;
+
+  // The number of bytes to read.
+  uint64 length = 3;
```
tylerwilliams, line 58: int64

proto/nbd.proto
```diff
@@ -0,0 +1,77 @@
...
+message DeviceMetadata {
+  // The device ID to be included in IO requests.
+  uint64 device_id = 1;
+
+  // The device label used to identify the device in metadata requests.
+  // Ex: "scratchfs"
+  string label = 2;
+
+  // The size of the block device in bytes.
+  uint64 size_bytes = 3;
```
tylerwilliams, line 33: int64

tylerwilliams: Obvi don't have a good style guide link here (goog internal only), but basically you don't want to make values like this unsigned just because you don't think they'll be negative, because you may later do math on them and get a weird result.

The read/write APIs in go are going to deal in either int64 or int anyway, so you will also probably reduce conversions this way.

https://jacobegner.blogspot.com/2019/11/unsigned-integers-are-dangerous.html

tylerwilliams (changes requested):
LGTM with some nits

proto/nbd.proto
bduffany: Nice, I hadn't thought of that before. Done!

bduffany:
FYI I got rid of `label` in favor of making `device_id` a string, serving the same purpose as `label`. `device_id` will either be `workspacefs` or `scratchfs`

proto/nbd.proto
```diff
@@ -0,0 +1,76 @@
...
+  // can refer to different underlying storage instances depending on the
+  // client/server identity.
+  // Ex: "workspacefs".
+  string name = 1;
+
+  // The size of the block device in bytes.
+  int64 size_bytes = 2;
+
+  // The filesystem type to be used in mount(2) operations, such as "ext4".
+  string filesystem_type = 3;
```
tylerwilliams, line 35: Should this be an enum, since it needs to be supported on the host side?

tylerwilliams (approved):

PR #4109, opened 2023-06-02
Add NBD client/server implementations

bduffany:
Implements the NBD server and client for the NBD service added in https://github.com/buildbuddy-io/buildbuddy/pull/4106 (Note, this PR does not wire up NBD to Firecracker just yet; instead it just adds some of the building blocks).

The NBD server runs on the executor. One server instance will be run per VM, similar to how the VFS server works. The server instance will expose two devices to the VM, named "scratchfs" and "workspacefs".

On the VM, the init process will create one `ClientDevice` for each of these two devices. A `ClientDevice` takes care of the local `/dev/nbd` connection, with an NBD backend that just proxies to the NBD server on the executor (over vsock).

The `scratchfs` device stays connected and mounted throughout the entire VM lifetime, even across pauses/resumes.  

On the other hand, `workspacefs` can be hot-swapped, as we replace action the workspace block device contents between actions. To hot-swap, the workspace device first has to be unmounted and disconnected, then the `blockio.Store` can be swapped out on the executor, then the `ClientDevice` can be re-connected and re-mounted in the VM. The executor controls this whole process, including the VM mounting/unmounting (via vmexec RPCs).

The architecture looks like this:

image

**Related issues**: N/A


enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
```diff
@@ -0,0 +1,306 @@
...
+	fstype := filesystemTypeString(d.metadata.GetFilesystemType())
+	if err := syscall.Mount(d.DevicePath(), path, fstype, syscall.MS_RELATIME, "" /*=data*/); err != nil {
+		return status.InternalErrorf("failed to mount %s (type %s) to %s: %s", d.DevicePath(), fstype, path, err)
+	}
+	log.Infof("Mounted %s (type %s) to %s", d.DevicePath(), d.metadata.GetName(), path)
+	return nil
+}
+
+// Unmount unmounts the device from its current mount path. It does not
+// disconnect the NBD device.
```
maggie-lou, line 201: It does disconnect the device right?

maggie-lou (approved):

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
bduffany: haha woops, yeah this is an outdated comment

vadimberezniker (approved):

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		fatalf("%s: read failed: %s", d.DevicePath(), err)
+	}
+	if len(res.Data) > len(p) {
+		fatalf(
+			"block device server returned too many bytes (0x%x) for read at offset=0x%x length=0x%x",
+			len(res.Data), len(p), off)
+	}
+	copy(p, res.Data)
+	return len(res.Data), nil
```
tylerwilliams, line 235: should this be:

```
bytesCopied := copy(p, res.Data)
return bytesCopied, nil
```

tylerwilliams (approved):

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
bduffany: good call, that seems slightly more conventional. Done

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
```diff
@@ -0,0 +1,299 @@
...
+	// them through directly to the host executor. The executor is responsible
+	// for guaranteeing that reads and writes are persisted.
+	return nil
+}
+
+// doWithRedial invokes a function that performs an RPC to the host server,
+// and re-dials the host server if the function fails on the first attempt.
+func (d *ClientDevice) doWithRedial(f func() error) error {
+	var lastErr error
+	for i := 1; i <= 2; i++ {
```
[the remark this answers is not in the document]
bduffany: I wanted the retry to be immediate (no delay) and only happen once, since we're just reconnecting to the new server which should be guaranteed to be running. It seemed more straightforward to use a loop than to use a retry.

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
```diff
@@ -0,0 +1,275 @@
...
+
+const (
+	// How long to wait for an NBD to become ready after it is created.
+	readyCheckTimeout = 1 * time.Second
+
+	// How often to poll for an NBD to become ready.
+	readyCheckPollInterval = 1 * time.Millisecond
+)
+
+// ClientDevice implements the BUSE interface by forwarding read and write
```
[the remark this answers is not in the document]
bduffany: Updated

PR #4110, opened 2023-06-02
Add digestFunction to key format and fix ANON groupID

tylerwilliams:
This migration adds the digest function to the stored key format and maps ANON data to a known groupID in the normal groupID format. So:

1) Pebble keys with a stored format like `PTfoo/ANON/abcd12345asdasdasd123123123asdasdasd/ac/123/EK123/v3` will be converted into keys like `PTfoo/GR74042147050500190371/abcd12345asdasdasd123123123asdasdasd/ac/123/EK123/v4`.
2) Until know, we've assumed everything is SHA256. Going forward, keys store the enum value of the digest function in their key. This allows disambiguating hash collisions that are possible with two different digest types. (For ex. `PTfoo/abcd12345asdasdasd123123123asdasdasd/1/v4` and `PTfoo/abcd12345asdasdasd123123123asdasdasd/9/v4` are different keys now)


enterprise/server/raft/filestore/filestore.go
```diff
@@ -19,12 +19,14 @@ import (
...
 
 	rfpb "github.com/buildbuddy-io/buildbuddy/proto/raft"
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	rspb "github.com/buildbuddy-io/buildbuddy/proto/resource"
 )
 
 const (
 	PartitionDirectoryPrefix = "PT"
 	groupIDPrefix            = "GR"
+	anonGroupID              = "GR74042147050500190371"
```
vadimberezniker, line 29: please add a comment explaining why we need this

enterprise/server/raft/filestore/filestore.go
```diff
@@ -276,6 +310,49 @@ func (pmk *PebbleKey) parseVersion3(parts [][]byte) error {
 	return nil
 }
 
+func (pmk *PebbleKey) parseVersion4(parts [][]byte) error {
+	digestFunctionString := "1"
```
vadimberezniker, line 314: nit: default to empty string. it may look to readers like this default value has some significance when in fact it's always overridden below.

vadimberezniker (approved):

tylerwilliams (commented):
Done

PR #4114, opened 2023-06-05
Add initial block storage layer for NBD / UFFD

bduffany:
- Add an interface for block-level storage
- Add a `File` implementation that uses `read()/write()`
- Add an `Mmap` implementation that uses `mmap()`. This is a bit experimental, but so far it seems to perform much better than `File`, with a tradeoff that it uses more memory. In a test with Bazel, the performance of nbd+Mmap is only 2% worse than our existing implementation which uses firecracker's Drive API to mount an ext4 image as a block device, compared to 20% worse for nbd+File.

In future PRs, this can be evolved to support copy-on-write (for forkable snapshots) as well as remote snapshot storage (e.g. lazily downloading blocks from cache).

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -45,7 +45,12 @@ type Mmap struct {
 	data []byte
 }
 
-func NewMmap(f *os.File) (*Mmap, error) {
+// NewMmap returns an Mmap for an existing file.
+func NewMmap(path string) (*Mmap, error) {
+	f, err := os.OpenFile(path, os.O_RDWR, 0)
+	if err != nil {
+		return nil, err
+	}
 	s, err := f.Stat()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -0,0 +1,100 @@
...
+	"io"
+	"os"
+	"syscall"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+)
+
+// Store models a block-level storage system, which is useful as a backend for a
+// Network Block Device (NBD) or userfaultfd(2).
+type Store interface {
```
tylerwilliams, line 14: Can you add some tests for this new interface and its implementations?

bduffany: Done

tylerwilliams (approved):

vadimberezniker (approved):

PR #4121, opened 2023-06-06
Add YAMLIgnoreTag for flags

tempoz:
Technical limitations of go generics demand that SecretTag and DeprecatedTag still be special-cased if we want using them to be at all pleasant, but we add YAMLIgnoreTag and use it for the config_file flag.

**Related issues**: N/A


tempoz:
* master
  * **PR #4121** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4121" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4121?utm_source=stack-comment).

server/config/config.go
```diff
@@ -12,15 +12,14 @@ import (
...
-var configPath = flag.String(pathFlagName, "/config.yaml", "The path to a buildbuddy config file")
-
-func init() {
-	// As this flag determines the YAML file we read the config from, it can't
-	// meaningfully be specified in the YAML config file.
-	flagyaml.IgnoreFlagForYAML(pathFlagName)
-}
+// As this flag determines the YAML file we read the config from, it can't
+// meaningfully be specified in the YAML config file.
+var configPath = flagutil.New(
```
vadimberezniker, line 17: Is there a need to wrap after every argument? It's inconsistent with how we typically define our flags.

vadimberezniker (approved):

server/config/config.go
tempoz: There is not, I'll swap it back. The line was just long, so I wrapped it.

PR #4123, opened 2023-06-06
Rename `bazel_exit_code` in invocation API to `exit_code_name`

bduffany:
* Drop `bazel_` prefix since this could also apply to workflow invocations if we decide to add named exit codes in the future.
* Add `_name` suffix to make it more obvious that it's a string (`exit_code` sounds like an int, and in the future we might actually want to add an `exit_code` field which is an int)

**Related issues**: N/A


bduffany:
Missed the release boat - ah well

PR #4126, opened 2023-06-06
remote_exec/runner: bump firecracker minimum memory to 200MB

sluongng:
MemSizeMB is the maximum amount of memory given to the Firecracker
microVM at run time.  This does not translate to the actual usage
comsumption however as the memory is only claimed when the page is
requested (1). So actual comsumption could be much lower.

In production, we estimate the amount of memory based on a linear
regression model trained on the historical data. However, this
estimation does not work well for new users with newly registered
executor and new build actions which are not in the trained data set.

When that happens, the original Estimated Task Size (calculated by some
heuristic in tasksize.go) is overridden by the Predicted Task Size
(value dervived from the Linear Regression model) when the task is
scheduled (2). This new estimated mem size could be less than 1MB, which
cause runner.go to use 1MB instead.

However, there are 2 errors encountered with Firecracker MicroVM when
the memory threshold is set too low.

1. Firecracker fail to load the ELF Kernel Image

```bash
[b441e327-746b-40fe-bf08-3a7c3630e535:fc_api:ERROR:src/api_server/src/parsed_request.rs:196] Received Error. Status code: 400 Bad Request. Message: Cannot load kernel due to invalid memory configuration or invalid kernel image: Kernel Loader: failed to load ELF kernel image
```

2. Our init process fail to copy file `/vmvfs` to `/mnt/vmvfs` (3)

```bash
2023/06/06 15:40:27.449 FTL die: write /mnt/vmvfs: copy_file_range: no space left on device
[    0.369707] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100
[    0.370944] CPU: 0 PID: 680 Comm: init Not tainted 5.4.0 #2
[    0.371789] Call Trace:
[    0.372176]  dump_stack+0x57/0x7a
[    0.372783]  panic+0xf6/0x2a4
[    0.373325]  do_exit.cold+0x1e/0xc5
[    0.373961]  ? schedule_hrtimeout_range_clock+0xb0/0x120
[    0.374890]  do_group_exit+0x3e/0xa0
[    0.375524]  get_signal+0x137/0x760
[    0.376151]  ? ep_poll+0x23c/0x460
[    0.376764]  do_signal+0x2f/0x260
[    0.377346]  ? wake_up_q+0x70/0x70
[    0.377865]  ? fput+0xe/0x10
[    0.378310]  exit_to_usermode_loop+0x53/0xa0
[    0.378957]  do_syscall_64+0xe5/0xf0
[    0.379531]  entry_SYSCALL_64_after_hwframe+0x44/0xa9
[    0.380295] RIP: 0033:0x403ace
[    0.380831] Code: Bad RIP value.
[    0.381434] RSP: 002b:000000c000063778 EFLAGS: 00000246 ORIG_RAX: 0000000000000119
[    0.382782] RAX: fffffffffffffffc RBX: 0000000000000003 RCX: 0000000000403ace
[    0.384032] RDX: 0000000000000080 RSI: 000000c000063840 RDI: 0000000000000003
[    0.385265] RBP: 000000c0000637c0 R08: 0000000000000000 R09: 0000000000000000
[    0.386451] R10: 000000000000270c R11: 0000000000000246 R12: 000000c000063850
[    0.387652] R13: 0000000000000000 R14: 000000c000007040 R15: 0000000000000001
[    0.388958] Kernel Offset: disabled
[    0.389557] Rebooting in 1 seconds..
INFO[0084] firecracker exited: status=0
```

Although the memory requirements may vary between different container
images that is used to create the Firecracker image. Using our default
container image `executor-docker-default`, 200MB seems to be the sweet
spot. This align with the default value in tasksize.go where we always
add an addtional 150MB to the Memory Estimation, before it was
overridden by the Model prediction result (4).

Increase minimum memsize for Firecracker MicroVM to 200MB.

(1): https://github.com/firecracker-microvm/firecracker/issues/2737#issuecomment-944272789
(2): https://github.com/buildbuddy-io/buildbuddy/blob/fb201eacd997d54c617db636fdf6c81eaccaa7e6/enterprise/server/scheduling/scheduler_server/scheduler_server.go#L385-L401
(3): https://github.com/buildbuddy-io/buildbuddy/blob/fb201eacd997d54c617db636fdf6c81eaccaa7e6/enterprise/server/cmd/goinit/main.go#L242
(4): https://github.com/buildbuddy-io/buildbuddy/blob/fb201eacd997d54c617db636fdf6c81eaccaa7e6/enterprise/server/tasksize/tasksize.go#L49-L53

**Related issues**: N/A


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -1093,7 +1093,7 @@ func (p *pool) newContainerImpl(ctx context.Context, props *platform.Properties,
 			DockerClient:           p.dockerClient,
 			ActionWorkingDirectory: p.hostBuildRoot(),
 			NumCPUs:                int64(math.Max(1.0, float64(sizeEstimate.GetEstimatedMilliCpu())/1000)),
-			MemSizeMB:              int64(math.Max(1.0, float64(sizeEstimate.GetEstimatedMemoryBytes())/1e6)),
+			MemSizeMB:              int64(math.Max(200.0, float64(sizeEstimate.GetEstimatedMemoryBytes())/1e6)),
```
bduffany, line 1096: nit: We should make this `200.0` a constant in `tasksize.go`, like `FirecrackerMinMemoryMB`

I think we should duplicate (\*) this change to `tasksize.go` (setting a minimum memory size for firecracker actions), that way the executor's local scheduler (`priority_task_scheduler.go`) gets a more accurate view of how much memory the task will take, and doesn't dequeue too many tasks at once (which all have a very low mem estimate in their TaskSize but actually wind up being assigned 200MB).

(\*) The reason to duplicate instead of move is that there might be a (rare) case where a task doesn't set `workload-isolation-type` but is assigned to an executor that is configured to run with firecracker isolation by default. `tasksize.go` won't be able to tell that the task is a firecracker task in that case.

sluongng:
Not super urgent as it only affects my laptop, so let's investigate how to bring down the memory cost instead.

sluongng:
Part of the problem here was also related to how we specify the default runner on the Executor side. This means that our scheduled task does not need to specify the isolation exec property explicitly in its Platform, which is what our scheduler uses to estimate task size.

The simple workaround was to add the firecracker isolation to the platform property to properly guide our scheduler.

Closing this now since we have shipped a lot more improvements with firecracker since.

PR #4128, opened 2023-06-07
Timing Profile: Make line charts more visible

luluz66:
Line charts and block charts shares a horizontal scroll bar, but now
have their own vertical scroll bars.

**Related issues**:
https://github.com/buildbuddy-io/buildbuddy-internal/issues/2326

[Screencast from 06-06-2023 05:34:15 PM.webm](https://github.com/buildbuddy-io/buildbuddy/assets/3977083/44990800-daa1-4e9f-9d0f-9cb67df5e75d)


luluz66:
I made the block charts taller (75%) instead of 50%. 

Screenshot from 2023-06-07 14-04-14

luluz66:
Applied some changes discussed offline:
1. make flame-chart taller: 900px instead of 600px.
2. use a slightly darker gray for the border between top and bottom
Screenshot from 2023-06-09 10-17-05

bduffany (approved):
Looks great :smiley:

Some feedback for a future iteration:

* Can maybe simplify the code as well as make it visually more intuitive by rendering two separate `<FlameChart>` instead of splitting the single flame chart (ideally we can keep the horizontal zoom/scroll in sync though so that people can easily correlate the event duration with high CPU usage etc.). The border does add some visual distinction, but not as much as I initially thought it would :thinking: 
* The top/bottom components could maybe benefit from headers like "events" / "stats" or something like that (suggestion from Siggi)

PR #4129, opened 2023-06-07
Make buildfix add newlines to the end of yaml files that don't have 'em

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/buildfix.sh
+++ b/buildfix.sh
@@ -78,6 +78,6 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
 fi
 
 # Add a newline to the end of .yaml files if not present.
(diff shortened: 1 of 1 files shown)


buildfix.sh
```diff
@@ -69,4 +69,15 @@ if ((GAZELLE)); then
   USE_BAZEL_VERSION="buildbuddy-io/$CLI_LATEST_VERSION" bazel fix
 fi
 
+SED="sed"
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  # The OS X version of sed behaves differently than the GNU version. Use gsed
+  # on OS X to make up for this. If this fails, install it with:
+  #     `brew install gnu-sed`
+  SED="gsed"
+fi
```
tylerwilliams, line 78: could we tweak the args for osx sed instead of requiring a separate install?

(I want to live in a world where new engineers don't have to install anything)

PR #4131, opened 2023-06-07
github: Uprade actions setup-go and cache

sluongng:
Got a warning from https://github.com/buildbuddy-io/buildbuddy/actions/runs/5202382221

```
Node.js 12 actions are deprecated.
Please update the following actions to use Node.js 16: actions/setup-go@v2, actions/cache@v2.

For more information see:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

**Related issues**: N/A

### DIFF
--- a/.github/workflows/checkstyle.yaml
+++ b/.github/workflows/checkstyle.yaml
@@ -23,12 +23,12 @@ jobs:
       #   uses: tj-actions/changed-files@v34.4.2
 
       - name: Setup go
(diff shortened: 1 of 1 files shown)


sluongng:
setup-go@v4 does support caching, so potentially we could remove `actions/cache` entirely after this.

bduffany (approved):

PR #4134, opened 2023-06-07
Defer invocation model creation so that it always has at least one invocation.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -23,7 +23,8 @@ export const HOSTED_BAZEL_ROLE = "HOSTED_BAZEL";
 export const InvocationStatus = invocation_status.InvocationStatus;
 
 export default class InvocationModel {
-  readonly invocations: invocation.Invocation[] = [];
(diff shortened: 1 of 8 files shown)


app/invocation/invocation_model.tsx
```diff
@@ -264,42 +268,46 @@ export default class InvocationModel {
...
+    return this.getPrimaryInvocation().invocationId;
+  }
+
+  getPrimaryInvocation(): invocation.Invocation {
+    return this.invocations[0];
+  }
+
+  getInvocations(): invocation.Invocation[] {
+    return this.invocations;
   }
```
bduffany, line 299: Instead of `getPrimaryInvocation()`, I think we could have the model expose a public readonly `invocation` instead of keeping a list of invocations. I don't think we only ever look at `invocations[0]` currently, and if we ever want more than one, then it'll make more sense to create one model per invocation rather than having a list

bduffany (approved):
LGTM because this PR feels better than what we have today - feel free to punt on my suggestion above :)

PR #4136, opened 2023-06-07
Improve display of workflow commands list

bduffany:
* Move commands to a more prominent section at the top, with improved styling
* Add a header which attempts to clarify that the invocation shown on the workflows page is for the runner itself, and is intended for use only in rare circumstances
* Remove a few details from the top section which aren't generally useful (username, hostname, OS type). These can still be found in the "Details" tab for the runner invocation.

In-progress:

image

Failed:

image

Dense mode:

image

**Related issues**: N/A


app/invocation/child_invocation_card.tsx
```diff
@@ -1,60 +1,57 @@
...
+      case "succeeded":
+        return <CheckCircle className="icon" />;
+      case "failed":
+        return <XCircle className="icon" />;
+      case "in-progress":
+        return <PlayCircle className="icon" />;
+      case "queued":
+        return <Timer className="icon" />;
+      case "not-run":
+        return <CircleSlash className="icon" />;
```
iain-macdonald, line 39: Consider adding a default case that shows some kind of error icon

app/invocation/child_invocation_card.tsx
```diff
@@ -1,60 +1,57 @@
...
-              </a>
-            ))}
-          </div>
+      <Link
+        className={`child-invocation-card status-${this.props.result.status} ${this.isClickable() ? "clickable" : ""}`}
+        href={this.isClickable() ? `/invocation/${this.props.result.invocation.invocationId}` : undefined}>
+        <div className="icon-container">{this.renderStatusIcon()}</div>
+        <div className="command">{this.props.result.invocation.bazelCommand}</div>
+        <div className="duration">
+          {typeof this.props.result.durationMillis === "number" &&
```
iain-macdonald, line 51: Can we use the question mark operator here instead of a type check? Like:
`format(duration.Millis(this.props.result.durationMillis?)`

app/invocation/child_invocations.tsx
```diff
@@ -10,94 +9,51 @@ export type ChildInvocationProps = {
 };
 
 export default class ChildInvocations extends React.Component<ChildInvocationProps> {
+  private getDurationMillis(invocation: InvocationMetadata): number | undefined {
```
iain-macdonald, line 12: Switch to just return 0 if not found and maybe rename to `getDurationMillisWithFallback` or something

iain-macdonald (approved):
niiiiiice :-)

app/invocation/child_invocation_card.tsx
bduffany: by default it'll fall through and return undefined, rendering nothing, which seems OK to me since it shouldn't happen in practice (we compute these statuses in the UI rather than the backend) - but I made it a bit less magical by adding an explicit `default:` case that returns undefined.

app/invocation/child_invocation_card.tsx
bduffany: I think we do want to keep treating the duration as optional, so that this renders nothing for commands that haven't run yet. But I changed it to `durationMillis !== undefined` which is a little more straightforward than using `typeof durationMillis === "number"`.

app/invocation/child_invocations.tsx
bduffany: Thinking about it a bit more, I'd like to keep it as `number | undefined`  so that we can distinguish between a real duration == 0 (e.g. it failed after a few microseconds or something and got rounded down to 0) and something which doesn't have a duration (e.g. `queued` / `not-run` statuses).

PR #4138, opened 2023-06-08
Use EventParser logic for repo URL, commit SHA, and role

bduffany:
Based on logging, this will cause a few metadata computations in the wild to change, but I figure it's better to consolidate this logic and help users deal with the effects of this change on a case-by-case basis, rather than having these confusing code paths for metadata computation (the confusion is that both codepaths are computing the same metadata, but one codepath would apply to GH status reporting and test grid, and the other codepath would apply to DB fields that we display in the UI).

**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/accumulator/BUILD
+++ b/server/build_event_protocol/accumulator/BUILD
@@ -7,12 +7,10 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//proto:build_event_stream_go_proto",
-        "//proto:command_line_go_proto",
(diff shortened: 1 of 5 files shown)


tylerwilliams (approved):
LGTM -- talked offline but let's proactively reach out to anyone this could possibly change results for (even tho the new results are more 'correct')

sluongng:
The affected user confirmed to me that Github Action with override `COMMIT_SHA` works as intended.

Thanks a lot! 🙏

PR #4139, opened 2023-06-08
Firecracker: Link vmvfs into init binary

bduffany:
This reduces the size of the initial RAM disk from 50MB -> 26MB

The go-fuse patch is required because it eagerly tries to read `/dev/null` in an `init()` func, and panics because it doesn't exist (as we haven't yet mounted devtempfs to /dev). I will see if I can upstream these changes and remove the patch in a follow-up.

**Related issues**: N/A


tylerwilliams (approved):
Looks like it needs buildfix

PR #4140, opened 2023-06-09
make the invocation model really only have one invocation in it.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/invocation/dense/dense_invocation_overview.tsx
+++ b/app/invocation/dense/dense_invocation_overview.tsx
@@ -7,7 +7,6 @@ import InvocationModel from "../invocation_model";
 interface Props {
   user?: User;
   model: InvocationModel;
-  invocationId: string;
 }
 export default class DenseInvocationOverviewComponent extends React.Component<Props> {
   handleUserClicked() {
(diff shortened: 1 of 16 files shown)


app/invocation/dense/dense_invocation_overview.tsx
```diff
@@ -29,10 +28,10 @@ export default class DenseInvocationOverviewComponent extends React.Component<Pr
               {isBazelInvocation ? <>Invocation</> : <>Workflow invocation</>}
             </div>
             <div className="dense-invocation-invocation-id">
-              {this.props.invocationId} ({this.props.model.getFormattedStartedDate()})
+              {this.props.model.getId()} ({this.props.model.getFormattedStartedDate()})
```
bduffany, line 31: nit: we might want to call this `getInvocationId()` - I noticed we have a convention of being more verbose about having a qualifier before things named "id," presumably to avoid confusion

bduffany (approved):
nice

app/invocation/dense/dense_invocation_overview.tsx
jdhollen: Sure, why not

PR #4141, opened 2023-06-09
Add support for encrypting API keys.

vadimberezniker:
I decided to use the same column to store either the plaintext or encrypted value of the API key to simplify the schema.

There are three flags to control the behavior of the feature and to allow a gradual rollout:
 - auth.api_key_encryption.key when the key is set, the auth code will look for keys matching both the unencrypted and unencrypted form of the key. Rolling out this flag first will allow encrypted keys to be recognized by all apps once they are in the database.

  - auth.api_key_encryption.encrypt_new_keys when this is set, all newly created keys will be writen in an encrypted format

  - auth.api_key_encryption.encrypt_old_keys on startup, any existing unencryptde keys will be encrypted

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -34,7 +34,7 @@ const (
 	apiKeyNonceLength = 6
 
 	// Encrypted API keys have this prefix in the database.
-	apiKeyEncryptedValuePrefix = "en"
+	apiKeyEncryptedValuePrefix = "en$"
 
 	apiKeyEncryptionBackfillBatchSize = 100
 )
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -2,7 +2,10 @@ package authdb
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/hex"
 	"flag"
+	"fmt"
 	"strings"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
@@ -17,6 +20,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/role"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"golang.org/x/crypto/chacha20"
 
 	akpb "github.com/buildbuddy-io/buildbuddy/proto/api_key"
 	grpb "github.com/buildbuddy-io/buildbuddy/proto/group"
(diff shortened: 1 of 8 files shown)


enterprise/server/backends/authdb/authdb.go
```diff
@@ -25,19 +29,108 @@ import (
...
+
+	// Encrypted API keys have this prefix in the database.
+	apiKeyEncryptedValuePrefix = "en$"
+
+	apiKeyEncryptionBackfillBatchSize = 100
 )
 
 var (
 	userOwnedKeysEnabled = flag.Bool("app.user_owned_keys_enabled", false, "If true, enable user-owned API keys.")
+	apiKeyEncryptionKey  = flag.String("auth.api_key_encryption.key", "", "Base64-encoded encryption key for API keys.")
```
bduffany, line 44: nit: should we mention here that this needs to be a chacha20 key?

enterprise/server/backends/authdb/authdb.go
```diff
@@ -25,19 +29,108 @@ import (
...
+			return err
+		}
+		var keysToUpdate []*tables.APIKey
+		for rows.Next() {
+			var apk tables.APIKey
+			if err := dbh.ScanRows(rows, &apk); err != nil {
+				return err
+			}
+			keysToUpdate = append(keysToUpdate, &apk)
+		}
```
bduffany, line 99: uber nit: check `rows.Err()` after the `rows.Next()` loop. (IDK if there is ever an error in practice, but the `Next()` documentation says that `Err()` needs to be consulted to distinguish between there being no more rows left and `Next()` failing)

bduffany (approved):
LGTM, just a few nits / questions

enterprise/server/backends/authdb/authdb.go
```diff
@@ -99,15 +192,86 @@ func (d *AuthDB) ClearSession(ctx context.Context, sessionID string) error {
 	return err
 }
 
+func (d *AuthDB) encryptAPIKey(apiKey string) (string, string, error) {
```
tylerwilliams, line 195: Could you add a detailed comment with an example on both `encryptAPIKey` and `decryptAPIKey` explaining how they work step by step?

(I think the code here is fine but if someone was not familiar with what we're doing here, they likely would find it pretty tough to follow)

enterprise/server/backends/authdb/authdb.go
```diff
@@ -25,19 +29,108 @@ import (
...
+		if err != nil {
+			return nil, status.InvalidArgumentErrorf("could not decode API Key encryption key: %s", err)
+		}
+		if len(key) != chacha20.KeySize {
+			return nil, status.InvalidArgumentError("API Key encryption key does not have expected length")
+		}
+		adb.apiKeyEncryptionKey = key
+		if err := adb.backfillUnencryptedKeys(); err != nil {
+			return nil, err
+		}
```
tylerwilliams, line 70: kinda freaky that this happens in NewAuthDB -- I guess nothing else from the environment is needed? Is the reason for doing it here (instead of with other migrations) because we don't want to make EncryptAPIKey public?

enterprise/server/backends/authdb/authdb.go
```diff
@@ -25,19 +29,108 @@ import (
 
 const (
 	apiKeyLength = 20
+
+	// For encrypted keys, the first N characters serve as the nonce.
+	apiKeyNonceLength = 6
+
+	// Encrypted API keys have this prefix in the database.
+	apiKeyEncryptedValuePrefix = "en$"
```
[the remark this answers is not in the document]
bduffany: Is it because clearing out the value would violate the uniqueness constraint? If so would it work to have the migration make both `value,encrypted_value` part of a unique composite index and then drop the unique index on `value`?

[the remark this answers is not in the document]
vadimberezniker: Good idea.

enterprise/server/backends/authdb/authdb.go
```diff
@@ -25,19 +29,106 @@ import (
...
+	apiKeyNonceLength = 6
+
+	apiKeyEncryptionBackfillBatchSize = 100
 )
 
 var (
 	userOwnedKeysEnabled = flag.Bool("app.user_owned_keys_enabled", false, "If true, enable user-owned API keys.")
+	apiKeyEncryptionKey  = flag.String("auth.api_key_encryption.key", "", "Base64-encoded 256-bit encryption key for API keys.")
+	encryptNewKeys       = flag.Bool("auth.api_key_encryption.encrypt_new_keys", false, "If enabled, all new API keys will be written in an encrypted format.")
+	encryptOldKeys       = flag.Bool("auth.api_key_encryption.encrypt_old_keys", false, "If enabled, all existing unencrypted keys will be encrypted on startup.")
```
bduffany, line 44: nit: should we add a note here that the old `value` column has to be manually cleared out once this is done?

bduffany (approved):
LGTM

tylerwilliams (approved):

enterprise/server/backends/authdb/authdb.go
vadimberezniker: done

PR #4145, opened 2023-06-09
Add redis cache to promQuerier

luluz66:
Add a redis cache in PromQuerier, and cache the prometheus json results to prevent querying prometheus too often. 
**Related issues**: N/A


enterprise/server/backends/prom/prom.go
```diff
@@ -104,11 +114,62 @@ func (c *bbMetricsCollector) Collect(out chan<- prometheus.Metric) {
 }
 
 func (q *promQuerier) FetchMetrics(ctx context.Context, groupID string) (model.Vector, error) {
+	resultVector, err := q.getCachedMetrics(ctx, groupID)
+	if err != nil {
+		log.Errorf("failed to get cached metrics: %s", err)
```
vadimberezniker, line 119: I would use Warningf here and below. I generally reserve Error logs for more critical failures.

vadimberezniker (approved):

PR #4149, opened 2023-06-12
Fix a bare ResourceName proto construction callsite

tylerwilliams:


server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -150,11 +150,7 @@ func (s *ActionCacheServer) GetActionResult(ctx context.Context, req *repb.GetAc
 	// Fetch the "ActionResult" object which enumerates all the files in the action.
 	d := req.GetActionDigest()
 	downloadTracker := ht.TrackDownload(d)
-	blob, err := s.cache.Get(ctx, &rspb.ResourceName{
-		Digest:       d,
-		CacheType:    rspb.CacheType_AC,
-		InstanceName: req.GetInstanceName(),
-	})
+	blob, err := s.cache.Get(ctx, rn.ToProto())
```
tylerwilliams, line 153: [I don't see why we were constructing a new ResourceName in the old code to begin with]

vadimberezniker (approved):

PR #4154, opened 2023-06-12
Firecracker: use NBD for scratch and workspace disks

bduffany:
Behind the flag `--executor.firecracker_enable_nbd`, use `mmap`-backed files for the scratch and workspace disks, served over vsock via the new NBD server.

This does not buy us anything on its own, except that it paves the way to swap out the `Store` instance with an implementation that supports copy-on-write.

**Related issues**: N/A


vadimberezniker (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -766,10 +800,24 @@ func (c *FirecrackerContainer) hotSwapWorkspace(ctx context.Context, execClient
...
+	if *enableNBD {
+		// Swap out the backing file for the workspace NBD.
+		if err := c.workspaceDevice.Close(); err != nil {
+			log.Warningf("Failed to close workspace nbd: %s", err)
+		}
+		store, err := blockio.NewMmap(c.workspaceFSPath())
+		if err != nil {
+			return status.WrapError(err, "initialize workspace NBD store")
+		}
+		c.workspaceDevice.Store = store
```
tylerwilliams, line 813: this feels sketch -- you're reaching into a *nbdserver.Device and replacing a member of it. Probably would be better / safer to have a Replace()/Load() method on the Device object and use that, or to replace the Device object entirely here

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1257,15 +1357,29 @@ func (c *FirecrackerContainer) Create(ctx context.Context, actionWorkingDir stri
...
 		return err
 	}
-	scratchFSPath := filepath.Join(c.tempDir, scratchFSName)
+	if err := os.MkdirAll(c.getChroot(), 0755); err != nil {
+		return status.InternalErrorf("failed to create chroot dir: %s", err)
+	}
+
+	scratchFSPath := c.scratchFSPath()
+	workspaceFSPath := c.workspaceFSPath()
+	// When mounting the workspace image directly as a block device (rather than
```
tylerwilliams, line 1366: nit: newline before the comment for readability

tylerwilliams (changes requested):

tylerwilliams (approved):

PR #4156, opened 2023-06-12
Run staticcheck with most analyzers as part of our go builds

iain-macdonald:
Contrary to the directions in [nogo-analyzer](https://github.com/sluongng/nogo-analyzer), I load the staticcheck dependencies after gazelle's because I want gazelle's to take precedence. This doesn't seem to cause any issues. All (53) analyzers that prevent compilation are currently disabled, but we should be able to enable most of them without too much effort, I just didn't want to do that in this PR.

**Related issues**: N/A


sluongng (approved):
LGTM.

Just want to note that I have https://github.com/sluongng/nogo-analyzer/blob/master/def.bzl which should help w. creating the nogo config json programmatically in starlark.

iain-macdonald:
> LGTM.
> 
> Just want to note that I have https://github.com/sluongng/nogo-analyzer/blob/master/def.bzl which should help w. creating the nogo config json programmatically in starlark.

I'll admit that I didn't actually try to set it up, but it looks like that would overwrite the config that we already have, unless I'm misunderstanding the write_file rule.

sluongng:
> I load the staticcheck dependencies after gazelle's because I want gazelle's to take precedence.

Yeah it's not needed, generally use which ever newer. Back when I wrote the tool, Gazelle release was 1 year late, thus that line was added to the doc?

> it looks like that would overwrite the config that we already have

Yeah it would, but I think we could convert the entire json config file into smaller starlark config in general.
Not a big deal though, we could simply do that at some point in the future.

PR #4160, opened 2023-06-13
Implement basic overlay store for copy-on-write block IO

bduffany:
Adds a `blockio.Overlay` struct that implements `blockio.Store` with basic copy-on-write functionality. This is a medium-term solution that we can use to implement snapshot sharing within a single executor. (We can also try using it for remote snapshot sharing, but it might not be feasible since it doesn't support granular chunking of the backing files.)

The `Overlay` store works similarly to overlayfs but for block devices instead of files. The "base" layer is immutable, and the "top" layer is mutable, copying blocks up from the bottom layer when doing writes. (Note, it doesn't support more than 2 layers).

The plan for using this new `Overlay` store is as follows:
* When we run a VM for the first time (with a particular set of platform properties), we can save an "immutable" snapshot to cache which can only be used as the base layer for an overlay backing NBD / UFFD.
* Then, subsequent VM runs can be forked from that immutable snapshot, and maintain their own "top" layer for writes.
* To prevent the immutable snapshots from getting too stale, we can invalidate them daily by including the current date in the cache key used to look up the snapshots from filecache.

In V1 (future PRs), we can explore using a more flexible system - instead of having just 2 big files as the backing layers, we can switch to a more granular Store where each block is backed by an individual cache artifact with a properly computed digest, so that we can use the cache for backing storage.

**Related issues**: N/A


tylerwilliams:
Talked offline but we wanted to explore:
 - changing the snaploader API to allow snaploader to manage all snapshot / layering decisions (firecracker.go should not have to care?)
 - explore doing COW at the block level so that we can represent any snapshot as simply a list of blocks

PR #4164, opened 2023-06-14
Support a "queued" invocation view and use it for queued workflows

bduffany:
* When clicking the workflow re-run button, set `async: true` in the ExecuteWorkflowRequest to avoid waiting for the invocation to be created. Instead, just wait for the action to be queued, and navigate to the invocation page with `?queued=true`.
* When `queued=true` is set in the URL on the invocation page, show a page similar to the "Invocation is in progress..." screen, but with the title "Invocation is waiting for an available worker..." instead.
* Whenever we enqueue a workflow action, publish a pending status to GitHub with the message "Queued...". The status "Details" link points to an invocation URL with `queued=true`.

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2243

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -1289,7 +1289,7 @@ func (ws *workflowService) createApprovalRequiredStatus(ctx context.Context, wf 
 func (ws *workflowService) createQueuedStatus(ctx context.Context, wf *tables.Workflow, wd *interfaces.WebhookData, actionName, invocationID string) error {
 	invocationURL := build_buddy_url.WithPath("/invocation/" + invocationID)
 	invocationURL.RawQuery = "queued=true"
-	status := github.NewGithubStatusPayload(actionName, invocationURL.String(), "Waiting for available worker...", github.PendingState)
+	status := github.NewGithubStatusPayload(actionName, invocationURL.String(), "Queued...", github.PendingState)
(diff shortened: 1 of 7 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -1281,6 +1286,18 @@ func (ws *workflowService) createApprovalRequiredStatus(ctx context.Context, wf
 	return ghc.CreateStatus(ctx, ownerRepo, wd.SHA, status)
 }
 
+func (ws *workflowService) createQueuedStatus(ctx context.Context, wf *tables.Workflow, wd *interfaces.WebhookData, actionName, invocationID string) error {
+	invocationURL := build_buddy_url.WithPath("/invocation/" + invocationID)
+	invocationURL.RawQuery = "queued=true"
+	status := github.NewGithubStatusPayload(actionName, invocationURL.String(), "Waiting for available worker...", github.PendingState)
```
siggisim, line 1292: Might this be too long to show up fully in the github UI? Should we just put Queued…?

siggisim (approved):

enterprise/server/workflow/service/service.go
bduffany: sg, done

tylerwilliams (approved):

PR #4165, opened 2023-06-14
CLI: Add some new block types to WORKSPACE.yaml parsing

siggisim:
This PR adds support for the following top level blocks in the WORKSPACE.yaml file:
`deps`, `templates`, `settings`, `bazel`, 

This allows you to specify a single file to configure your whole Bazel setup, like so:
```
workspace:
  name: myworkspace

bazel:
  version: 6.1.1

settings:
  bes_results_url: https://app.buildbuddy.io/invocation/
  bes_backend: remote.buildbuddy.io
  remote_cache: remote.buildbuddy.io
  remote_executor: remote.buildbuddy.io

templates:
  server: github/enricofoltran/simple-go-server
  app: github/Kornil/simple-ts-react-app

deps:
  - github/bazelbuild/rules_go
  - github/bazelbuild/rules_docker
  - github/bazelbuild/rules_proto
  - github/bazelbuild/bazel-gazelle
  - github/bazelbuild/rules_nodejs
```

As an escape hatch, you can still add things to your workspace file manually like so:
```
http_archive:
    name: "io_bazel_rules_go"
    sha256: "6dc2da7ab4cf5d7bfc7c949776b1b7c733f05e56edc4bcd9022bb249d2e2a996"
    urls:
        - "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.39.1/rules_go-v0.39.1.zip"
        - "https://github.com/bazelbuild/rules_go/releases/download/v0.39.1/rules_go-v0.39.1.zip"
```

As a total last resort, you can use the `raw:` block to dump whatever else you need in there:
```
raw:
  my_super_special_thing()
```

### DIFF
--- a/cli/translate/yaml/yaml.go
+++ b/cli/translate/yaml/yaml.go
@@ -2,9 +2,13 @@ package yaml
 
 import (
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/buildbuddy-io/buildbuddy/cli/add"
 	"github.com/buildbuddy-io/buildbuddy/cli/log"
 	"github.com/buildbuddy-io/buildbuddy/cli/translate/builtins"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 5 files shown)


cli/translate/yaml/yaml.go
```diff
@@ -161,3 +191,123 @@ func (y *yamlTranslator) translateLoad(m yaml.MapSlice) string {
...
+		log.Printf("skipping template %s, directory %q already exists", from, into)
+		return
+	}
+	log.Printf("grabbing template from %s and putting it into %q", from, into)
+
+	if !strings.HasPrefix(from, "github/") {
+		log.Printf("unknown template %s", from)
+	}
+
+	repo := strings.Replace(from, "github/", "https://github.com/", 1)
```
bduffany, line 302: The `github/` notation is convenient when adding deps, but I feel like requiring it to be written out as `github.com` might make the YAML format overall more convenient to work with from a usability perspective. (The URLs can be used directly instead of needing a special tool to parse them or having to manually add ".com" if you want to inspect a dependency in the browser)

cli/translate/yaml/yaml.go
```diff
@@ -161,3 +191,123 @@ func (y *yamlTranslator) translateLoad(m yaml.MapSlice) string {
...
+			for _, x := range slice {
+				key, ok := x.Key.(string)
+				if !ok {
+					log.Printf("unknown key type: %T", x.Value)
+					continue
+				}
+				value, ok := x.Value.(string)
+				if !ok {
+					log.Printf("unknown value type: %T", x.Value)
+					continue
```
bduffany, line 272: instead of printing, wdyt about returning an error?

or if we can tolerate the error, maybe use log.Warnf so that if the YAML is invalid, it's a bit more pronounced in the CLI output

bduffany (approved):

cli/translate/yaml/yaml.go
siggisim: I'd like to support both formats in both places, I've added a note for myself to do this in a follow-up

PR #4172, opened 2023-06-14
Perform flag expansion on the flag values themselves.

vadimberezniker:
This allows expansion to work for flags that are specified on the command line.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/configsecrets/BUILD
+++ b/enterprise/server/backends/configsecrets/BUILD
@@ -6,9 +6,9 @@ go_library(
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/configsecrets",
     visibility = ["//visibility:public"],
     deps = [
+        "//server/config",
         "//server/interfaces",
         "//server/util/flagutil",
-        "//server/util/flagutil/yaml",
         "//server/util/status",
         "@com_google_cloud_go_secretmanager//apiv1",
         "@com_google_cloud_go_secretmanager//apiv1/secretmanagerpb",
--- a/enterprise/server/backends/configsecrets/configsecrets.go
+++ b/enterprise/server/backends/configsecrets/configsecrets.go
@@ -9,9 +9,9 @@ import (
 	"fmt"
 
 	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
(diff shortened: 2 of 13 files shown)


tempoz (approved):
This looks good to me, though I think we'll have to document it carefully. It may be confusing for users that they need to prevent their shell from expanding secrets if they try to use this via the command line, like so:
```sh
bb run //server -- --database.data_source="postgresql://${USER}:\${SECRET:pgpass}@localhost/buildbuddy?sslmode=disable"

# or depend entirely on our expansion and skip the shell:

bb run //server -- --database.data_source='postgresql://${USER}:${SECRET:pgpass}@localhost/buildbuddy?sslmode=disable'
```

since otherwise, their shell will simply expand it to the empty string (assuming no `SECRET:<var>` environment variable exists, of course).

Edit:

Also (and this was already happening prior to your change), we should make sure it's documented that we are doing environment variable expansion twice in case the case that they are using the command line, so if they want something that looks like an env variable expansion as a literal string value, they have to do something like:

```sh
bb run //server -- --storage.tempdir="/\\\${tmp}"

# or

bb run //server -- --storage.tempdir='/\${tmp}'
```

though of course we may hope no one names a directory something as obviously problematic as `/${tmp}/`.

Finally, all of this is of course just slightly different if the user is `exec`ing programmatically and bypassing the shell, in which case they should of course account only for our environment variable expansion (which also expands secrets).

tempoz:
Actually, I was just testing with expand, and it doesn't handle escaping dollar signs at all:

https://github.com/golang/go/issues/43482

That sort of seems like a significant problem.

vadimberezniker:
> 
> That sort of seems like a significant problem.
> 

We've been calling `Expand` for a long time and have yet to run into having an issue with having $ in flag values so I don't think it's that big of a problem. If it does come up in the future, we can easily replace the `Expand` call with our own implementation.

tempoz:
> > That sort of seems like a significant problem.
> 
> We've been calling `Expand` for a long time and have yet to run into having an issue with having $ in flag values so I don't think it's that big of a problem. If it does come up in the future, we can easily replace the `Expand` call with our own implementation.

sgtm!

tylerwilliams (approved):

PR #4178, opened 2023-06-15
Make `Login`, `Auth`, and `Logout` handlers return errors instead of redirecting

siggisim:
This PR makes the HTTPAuthenticator methods `Login`, `Auth`, and `Logout` return errors rather than just immediately redirecting the browser.

This lets us reason about how each of these calls behaved when we have multiple HTTPAuthenticators.

Adds a `HandlerFunc` that converts these errors into a browser redirect.

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -25,7 +25,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lru"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
-	"github.com/buildbuddy-io/buildbuddy/server/util/request_context"
+	requestcontext "github.com/buildbuddy-io/buildbuddy/server/util/request_context"
 	"github.com/buildbuddy-io/buildbuddy/server/util/role"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/golang-jwt/jwt"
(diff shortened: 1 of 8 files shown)


enterprise/server/auth/auth.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/lru"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
-	"github.com/buildbuddy-io/buildbuddy/server/util/request_context"
+	requestcontext "github.com/buildbuddy-io/buildbuddy/server/util/request_context"
```
bduffany, line 28: nit: move after `oidc` import at the bottom

server/http/interceptors/interceptors.go
```diff
@@ -313,3 +314,13 @@ func WrapAuthenticatedExternalHandler(env environment.Env, next http.Handler) ht
 		RecoverAndAlert,
 	})
 }
+
+type HandlerFunc func(http.ResponseWriter, *http.Request) error
+
+func (f HandlerFunc) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
bduffany, line 320: nit: maybe rename `HandlerFunc` to `RedirectOnError`

so the call site is like

```
interceptors.RedirectOnError(auth.Login)
```

bduffany (approved):

server/http/interceptors/interceptors.go
siggisim: Done

PR #4181, opened 2023-06-15
cleanup: remove executor gRPC server

bduffany:
The executor no longer serves gRPC requests, so it should be safe to remove the gRPC server.

Since we're no longer depending on `grpc_server.go`, I defined a new deprecated "grpc_port" flag in executor.go to avoid breaking people who have old configs and are still setting that for whatever reason. We can remove the flag later after giving people some time to notice the deprecation warning. (I don't recall seeing any of the other flags ever being set for the executor, like `app.grpc_max_recv_msg_size_bytes`, `grpcs_port`, `internal_grpc_port` etc., but I could be mistaken)

Also update the scheduler to no longer require `port` metadata from the executor registration. After this PR is rolled out and all scheduler instances stop checking `port != 0`, we can send a follow-up to stop hard-coding the `port` field to `1` and remove it from the proto.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -941,8 +941,8 @@ func (s *SchedulerServer) GetPoolInfo(ctx context.Context, os, requestedPool, wo
 }
 
 func (s *SchedulerServer) checkPreconditions(node *scpb.ExecutionNode) error {
-	if node.GetHost() == "" || node.GetPort() == 0 {
-		return status.FailedPreconditionErrorf("Cannot register node with empty host/port: %s:%d", node.GetHost(), node.GetPort())
+	if node.GetHost() == "" {
+		return status.FailedPreconditionError("executor node registration is missing 'host' field")
 	}
 
 	return nil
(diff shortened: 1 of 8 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -965,12 +965,11 @@ func (s *SchedulerServer) RemoveConnectedExecutor(ctx context.Context, handle *e
...
 	ctx, cancel := context.WithTimeout(context.Background(), removeExecutorCleanupTimeout)
 	defer cancel()
-	addr := fmt.Sprintf("%s:%d", node.GetHost(), node.GetPort())
 	if err := s.deleteNode(ctx, node, nodePoolKey); err != nil {
-		log.CtxWarningf(ctx, "Scheduler: could not unregister node %q: %s", addr, err)
+		log.CtxWarningf(ctx, "Scheduler: could not unregister node %q: %s", node.GetExecutorId(), err)
 		return
 	}
-	log.CtxInfof(ctx, "Scheduler: unregistered worker node: %q", addr)
+	log.CtxInfof(ctx, "Scheduler: unregistered node %q", node.GetExecutorId())
```
vadimberezniker, line 972: if we are logging the host in the registration message, can we also log it here for parity? 
for macs it's sometimes easier to find things by IP

vadimberezniker (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
bduffany: Done

tylerwilliams (approved):

PR #4182, opened 2023-06-15
Move `OpenIDAuthenticator` from auth.go to oidc.go

siggisim:
This will fail tests, but I want to keep the diff small in order for Github to recognize this change as a "move".

If this looks reasonable, I will add the rest of the changes to this PR which:
- Recreates auth.go with just a few methods `Register`, `RegisterNullAuth`, `UserFromTrustedJWT`
- Moves some constants to authutils
- Will likely make the diff look terrible

Here's what the diff would look like without doing this in two stages (heinous):
https://github.com/buildbuddy-io/buildbuddy/compare/master...siggi-dev-branch-20230615-143932?body=&expand=1


tylerwilliams (approved):

siggisim:
Added the rest of the changes, if you want to look at them in isolation with a clean diff - you can see them here: https://github.com/buildbuddy-io/buildbuddy/pull/4182/commits/22f4ec3996dba983df7cd16f0d8b9d57f0127299

PR #4183, opened 2023-06-16
Remove fallback concept from `SAMLAuthenticator`

siggisim:
**Related issues**: N/A


enterprise/server/auth/auth.go
```diff
@@ -20,10 +23,35 @@ var (
...
+
+	oidc, err := oidc.NewOpenIDAuthenticator(ctx, env, *adminGroupID)
 	if err != nil {
-		return status.InternalErrorf("Authenticator failed to configure: %v", err)
+		return status.InternalErrorf("OIDC authenticator failed to configure: %v", err)
+	}
+	httpAuthenticators = append(httpAuthenticators, oidc)
+	userAuthenticators = append(userAuthenticators, oidc)
+
+	if saml.SamlEnabled(env) {
```
bduffany, line 36: nit: consider renaming to `saml.IsEnabled(env)`

enterprise/server/auth/auth.go
```diff
@@ -38,6 +66,129 @@ func RegisterNullAuth(env environment.Env) error {
 	return nil
 }
 
+type authenticator struct {
+	http         []interfaces.HTTPAuthenticator
+	user         []interfaces.UserAuthenticator
+	grpc         interfaces.GRPCAuthenticator
+	apiKey       interfaces.APIKeyAuthenticator
+	installation interfaces.InstallationAuthenticator
```
bduffany, line 74: nit: if you want, you could embed the interfaces instead of having them as fields, to avoid having to define the "passthrough" methods that delegate directly

i.e.

```
type authenticator struct {
  interfaces.GRPCAuthenticator
}

auth := &authenticator{
  ...,
  GRPCAuthenticator: oidc,
}

// can delete AuthenticatedGRPCContext() now
```

enterprise/server/saml/saml.go
```diff
@@ -55,56 +54,37 @@ var (
...
-	return a.fallback.PublicIssuers()
-}
-
 func (a *SAMLAuthenticator) SSOEnabled() bool {
-	if *certFile != "" {
-		return true
-	}
-	return a.fallback.SSOEnabled()
+	return (*certFile == "" && *cert == "")
 }
```
bduffany, line 78: haven't fully thought it through but will this method always return `true` ? since `Enabled()` returns false if these are both empty and we don't create a SAMLAuthenticator if Enabled() returns false.

enterprise/server/auth/auth.go
```diff
@@ -20,10 +23,35 @@ var (
...
 	if err != nil {
-		return status.InternalErrorf("Authenticator failed to configure: %v", err)
+		return status.InternalErrorf("OIDC authenticator failed to configure: %v", err)
+	}
+	httpAuthenticators = append(httpAuthenticators, oidc)
+	userAuthenticators = append(userAuthenticators, oidc)
+
+	if saml.SamlEnabled(env) {
+		samlAuthenticator := saml.NewSAMLAuthenticator(env)
+		httpAuthenticators = append(httpAuthenticators, samlAuthenticator)
```
bduffany, line 38: do you want the SAML authenticator to come first in these lists so that it takes precedence?

enterprise/server/auth/auth.go
siggisim: Done

enterprise/server/auth/auth.go
siggisim: Ideally order shouldn't matter, and if for some reason it does I'd like to fix it

enterprise/server/auth/auth.go
siggisim: Fancy, done

enterprise/server/saml/saml.go
siggisim: Yeah, done

enterprise/server/auth/auth.go
bduffany: ah I was mostly concerned about which error gets returned (since some codepaths depend on error codes), but after re-reading I see we're returning `errors[0]` which effectively prioritizes OIDC and keeps the same behavior we have today

bduffany (approved):

siggisim:
Yeah, agreed that the error part is probably the gnarliest bit. Probably worth fixing the SAML stuff at some point so the correct errors are returned in either case and we're not relying on ordering at all.

PR #4184, opened 2023-06-16
[RFC] workflow.proto: allow custom action

sluongng:
User may have a needs to execute custom Workflow Actions that is only
known at runtime and cannot be configured before hand.

For this use case, we could add `custom_action` field which mirror
workflow/config.BuildBuddyConfig struct. This way user could craft their
own set of actions before making the request.

When this happen, we shall ignore the `buildbuddy.yaml` inside the repo
as well as the action_names filter.

**Related issues**: N/A

### DIFF
--- a/proto/workflow.proto
+++ b/proto/workflow.proto
@@ -176,6 +176,30 @@ message ExecuteWorkflowRequest {
 
   // If true, start the workflow but do not wait for the status to be returned.
   bool async = 14;
+
+  // Optional: Custom actions to execute. Correspond to actions in
+  // buildbuddy.yaml. If this is set, action_names will be ignored.
+  repeated CustomAction custom_action = 15;
+}
+
+message CustomAction {
+  string name = 1;
+  string os = 2;
+  string arch = 3;
+  string self_hosted = 4;
+  string container_image = 5;
+  ResourceRequests resource_requests = 6;
+  string user = 7;
+  repeated string git_clean_exclude = 8;
+  string bazel_workspace_dir = 9;
+  map<string, string> env = 10;
+  repeated string bazel_commands = 11;
+}
+
+message ResourceRequests {
+  string cpu = 1;
+  string memory = 2;
+  string disk = 3;
 }
 
 message ExecuteWorkflowResponse {


sluongng:
Very much an RFC so I did not add much doc / comment.

This would help our customer construct their own release/deployment workflows dynamically based using their own calculation.

proto/workflow.proto
```diff
@@ -176,6 +176,30 @@ message ExecuteWorkflowRequest {
 
   // If true, start the workflow but do not wait for the status to be returned.
   bool async = 14;
+
+  // Optional: Custom actions to execute. Correspond to actions in
+  // buildbuddy.yaml. If this is set, action_names will be ignored.
+  repeated CustomAction custom_action = 15;
+}
+
+message CustomAction {
```
maggie-lou, line 185: What if we just moved the definition of BuildBuddyConfig.Action to proto and refactored the code to use that everywhere? That way we can guarantee the two structs don't diverge

proto/workflow.proto
```diff
@@ -176,6 +176,30 @@ message ExecuteWorkflowRequest {
 
   // If true, start the workflow but do not wait for the status to be returned.
   bool async = 14;
+
+  // Optional: Custom actions to execute. Correspond to actions in
+  // buildbuddy.yaml. If this is set, action_names will be ignored.
+  repeated CustomAction custom_action = 15;
```
maggie-lou, line 182: nit: custom_actions (plural)

proto/workflow.proto
sluongng: Im ok with that as well. 😁

maggie-lou: Another option is we could use Zoey's yaml variable interpolation logic ([Example](https://github.com/buildbuddy-io/buildbuddy-internal/blob/master/enterprise/deployment/dev/us-west1-b/executor.dev.yaml#L10)) and just have users pass in a map of variable name -> value to the API
Ex.  variables: {
     "service-name": "login",
    ...
}

The advantage for that is if there's a lot of redundant information for each action (like image, env variables, resource requests), users won't have to pass in all that data

bduffany: I forgot that @vadimberezniker actually had added a `Runner` API a while back to support remote bazel, and I suspect this has some overlap. Can you check the runner API and see whether it would work instead of custom actions?

bduffany: (I think adding an `inputs` field to our existing workflows and allowing people to specify a mapping of inputs also seems reasonable, if the runner API needs too much work)

tylerwilliams: I'd prefer exposing the runner API publicly to allowing custom workflow actions.

sluongng:
Will close this for now to focus on testing `bb remote` and Runner API instead.

PR #4185, opened 2023-06-16
put workflow history on the workflows page.

jdhollen:
it looks like this:
image

**Related issues**: N/A


enterprise/server/workflow/service/service.go
```diff
@@ -794,6 +798,153 @@ func (ws *workflowService) waitForWorkflowInvocationCreated(ctx context.Context,
...
+	q.AddWhereClause("role = ?", "CI_RUNNER")
+	q.AddWhereClause("group_id = ?", authenticatedUser.GetGroupID())
+	q.AddWhereClause("updated_at_usec > ?", timeLimitMicros)
+	q.SetOrderBy("created_at_usec", false)
+	q.SetLimit(30)
+
+	return q, nil
+}
+
+func (ws *workflowService) GetWorkflowHistory(ctx context.Context, req *wfpb.GetWorkflowHistoryRequest) (*wfpb.GetWorkflowHistoryResponse, error) {
```
jdhollen, line 822: it's not that i don't want to split this function up, it's that each phase has like 3 things going on, but i'll figure something out if folks really want me to.

jdhollen:
Broadly, this is currently sitting on the workflows page, and is only really visible to admins because we don't have an rpc path for fetching workflow info that isn't behind the admin role filter.

I'm sending this out to get comments on the implementation and feature in general, but I'd also like to hear what folks think about making something like this accessible for all users--we don't really have an equivalent page right now and this doesn't reveal much beyond invocation IDs and commit SHAs.

enterprise/server/workflow/service/service.go
```diff
@@ -794,6 +798,153 @@ func (ws *workflowService) waitForWorkflowInvocationCreated(ctx context.Context,
...
+
+	// Find the names of all of the actions for each repo.
+	q := query_builder.NewQuery(`SELECT repo_url,pattern,count(1) AS total_runs, countIf(success) AS successful_runs, toInt64(avg(duration_usec)) AS average_duration FROM Invocations`)
+	q.AddWhereClause("repo_url IN ?", repos)
+	q.AddWhereClause("role = ?", "CI_RUNNER")
+	q.AddWhereClause("group_id = ?", authenticatedUser.GetGroupID())
+	q.AddWhereClause("invocation_status = ?", int(inspb.InvocationStatus_COMPLETE_INVOCATION_STATUS))
+	q.AddWhereClause("updated_at_usec > ?", timeLimitMicros)
+	q.SetGroupBy("repo_url,pattern")
+	// XXX: No permissions check here; is this ok?
```
bduffany, line 843: The current perms check (group_id = authenticated group ID) LGTM. We don't support the `perms` field in clickhouse currently which means we're effectively using strict group RW perms, which is fine for our current use cases.

enterprise/server/workflow/service/service.go
```diff
@@ -794,6 +798,153 @@ func (ws *workflowService) waitForWorkflowInvocationCreated(ctx context.Context,
...
+		summary := &wfpb.ActionHistory_ActionHistorySummary{
+			TotalRuns:       row.TotalRuns,
+			SuccessfulRuns:  row.SuccessfulRuns,
+			AverageDuration: durationpb.New(time.Duration(row.AverageDuration) * time.Microsecond),
+		}
+		if _, ok := workflows[row.RepoUrl]; !ok {
+			workflows[row.RepoUrl] = make(map[string]*wfpb.ActionHistory)
+		}
+		workflows[row.RepoUrl][row.Pattern] = &wfpb.ActionHistory{ActionName: row.Pattern, Summary: summary}
+	}
```
bduffany, line 884: uber nit: check `rows.Err()` here - not sure if it happens in practice but the docs for `rows.Next()` say that it can fail and `rows.Err()` will contain the failure

bduffany (approved):
LGTM, mostly just nits

> Broadly, this is currently sitting on the workflows page, and is only really visible to admins because we don't have an rpc path for fetching workflow info that isn't behind the admin role filter.

I think this is fine for now - at some point in the future, we could maybe hide all the management controls for non-admins but allow developers to see the list of workflows.

enterprise/server/workflow/service/service.go
jdhollen: done here and for the history one below.

enterprise/server/workflow/service/service.go
```diff
@@ -794,6 +798,153 @@ func (ws *workflowService) waitForWorkflowInvocationCreated(ctx context.Context,
...
+		return nil, status.FailedPreconditionError("database not configured")
+	}
+
+	repos := req.GetRepoUrls()
+	authenticatedUser, err := ws.env.GetAuthenticator().AuthenticatedUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// Only fetch workflow data for stuff that has run in the last 7 days.
+	timeLimitMicros := time.Now().Add(-time.Duration(7*24) * time.Hour).UnixMicro()
```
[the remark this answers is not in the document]
jdhollen: thanks, this is nicer

enterprise/server/workflow/service/service.go
```diff
@@ -794,6 +798,153 @@ func (ws *workflowService) waitForWorkflowInvocationCreated(ctx context.Context,
...
+	if err := perms.AddPermissionsCheckToQuery(ctx, ws.env, q); err != nil {
+		return nil, err
+	}
+	authenticatedUser, err := ws.env.GetAuthenticator().AuthenticatedUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+	q.AddWhereClause("repo_url = ?", repoUrl)
+	q.AddWhereClause("pattern = ?", pattern)
+	q.AddWhereClause("branch_name IN ?", []string{"master", "main"})
```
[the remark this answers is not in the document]
jdhollen: Done

enterprise/app/workflows/workflows.tsx
```diff
@@ -122,6 +129,18 @@ class ListWorkflowsComponent extends React.Component<ListWorkflowsProps, State>
     }
   }
 
+  private fetchWorkflowHistory(repoUrls: string[]) {
+    if (!capabilities.config.workflowHistoryEnabled || repoUrls.length === 0) {
+      return;
```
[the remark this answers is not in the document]
jdhollen: Put it up at the top of the function instead, so now i'm just letting the subsequent stuff override it if it manages to get called.

enterprise/app/workflows/workflows.tsx
```diff
@@ -141,7 +160,10 @@ class ListWorkflowsComponent extends React.Component<ListWorkflowsProps, State>
     this.setState({ reposLoading: true });
     this.fetchReposRPC = rpcService.service
       .getLinkedGitHubRepos(new github.GetLinkedReposRequest())
-      .then((response) => this.setState({ reposResponse: response }))
+      .then((response) => {
+        this.setState({ reposResponse: response });
+        this.fetchWorkflowHistory(response.repoUrls);
```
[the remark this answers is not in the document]
jdhollen: Done

enterprise/app/workflows/workflows.css
```diff
@@ -425,3 +425,133 @@
...
+}
+
+.action-history-tooltip {
+  background-color: rgba(255, 255, 255, 0.95);
+  border-radius: 8px;
+  padding: 16px;
+  box-shadow: 0 3px 6px rgba(0, 0, 0, 0.16), 0 3px 6px rgba(0, 0, 0, 0.23);
+  pointer-events: none;
+  z-index: 1;
+}
```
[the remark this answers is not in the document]
jdhollen: Done

enterprise/app/workflows/action_list.tsx
```diff
@@ -0,0 +1,212 @@
...
+    return null;
+  }
+  if (invocationId === "no-data") {
+    return <div className="action-history-tooltip">No more runs in the last 7 days.</div>;
+  }
+  const entry = lookup.get(invocationId);
+
+  if (entry) {
+    return (
+      <div className="action-history-tooltip">
```
bduffany, line 94: Just double-checking that the styling still looks good for this after nesting under `.workflows-page` - I think the tooltip stuff renders in a portal so it might not be nested under `.workflows-page` anymore. (I usually hack around this by wrapping the tooltip content with a `.workflows-page` div)

bduffany (approved):

enterprise/app/workflows/action_list.tsx
jdhollen: gah, you're right, fixed.

proto/workflow.proto
```diff
@@ -120,6 +122,71 @@ message GetWorkflowsResponse {
   repeated Workflow workflow = 2;
 }
 
+message ActionHistory {
+  // The name of the action spawned by the workflow--this is stored in the
+  // 'pattern' field of the invocation.
+  string action_name = 1;
+
+  // Information about a single run of this action.
+  message ActionHistoryEntry {
```
tylerwilliams, line 131: I'm not sure if the precedent is 100% there, but wdyt about removing the redundant prefix here (and below) since these are nested inside of ActionHistory?

Would make the callsites less stuttery: `ActionHistory_ActionHistoryEntry` -> `ActionHistory_Entry`

proto/workflow.proto
```diff
@@ -120,6 +122,71 @@ message GetWorkflowsResponse {
...
+  // The repos where we expect to find workflow config--we will only check for
+  // config matching the logged-in user's currently selected organization.
+  repeated string repo_urls = 2;
+}
+
+message GetWorkflowHistoryResponse {
+  // The response context.
+  context.ResponseContext response_context = 1;
+
+  message WorkflowHistory {
```
tylerwilliams, line 182: it seems like we're sort of using repo_url as workflow_id here -- wdyt about including `id` (aka workflow ID) in the `GetWorkflowHistoryResponse` (even if we don't add it to `GetWorkflowHistoryRequest`)?

enterprise/server/workflow/service/service.go
```diff
@@ -794,6 +798,164 @@ func (ws *workflowService) waitForWorkflowInvocationCreated(ctx context.Context,
...
+		return nil, err
+	}
+	// Only fetch workflow data for stuff that has run in the last 7 days.
+	timeLimitMicros := time.Now().Add(-7 * 24 * time.Hour).UnixMicro()
+
+	// Find the names of all of the actions for each repo.
+	q := query_builder.NewQuery(`SELECT repo_url,pattern,count(1) AS total_runs, countIf(success) AS successful_runs, toInt64(avg(duration_usec)) AS average_duration FROM Invocations`)
+	q.AddWhereClause("repo_url IN ?", repos)
+	q.AddWhereClause("role = ?", "CI_RUNNER")
+	// Clickhouse doesn't have an explicit perms column, so we only do auth
```
tylerwilliams, line 841: uberest of nits: would be nice to leave a newline between `q.AddWhereClause` and the `// Clickhouse ...` comment on the next line.

tylerwilliams (approved):
Dumb nits from me; this looks real nice and it will be cool when it's visible to everyone!

proto/workflow.proto
bduffany: We removed the concept of workflow_id as part of the GitHub app stuff. Now it's just linked repositories and the repository URL (+groupID) acts as the primary key in the DB

proto/workflow.proto
jdhollen: sure, done

enterprise/server/workflow/service/service.go
jdhollen: done

PR #4189, opened 2023-06-20
Enable golang static check analyzer S1000 (Use plain channel send or receive instead of single-case select)

iain-macdonald:
**Related issues**: N/A


tylerwilliams (approved):
LGTM

PR #4197, opened 2023-06-21
Export invocation duration and remote execution queue length

luluz66:
This PR supports the bbMetricsCollector to fetch and cache multiple metrics
from prometheus. It also added support for histogram metrics.

1. Add a proto for storing intermediate result in redis cache. In order to support
multiple metrics, we need a more complicated data structure for the cached
result.
2. For histogram, we need to fetch "_count", "_sum" and "bucket" metrics
   from prometheus and organized the results by the label set.

 Metrics:
- Added a seperate invocation duration metric in metrics.go with less
   dimensions for exporting.
- Removed group_id label in invocation count metrics. Customers can use
  exported_buildbuddy_invocation_duration_usec_count instead. I planned to have
  a seperate invocation count metric with only invocation status label for
  exporting. Then I realized that it would be the same as
  invocation_duration_usec_count metric.

**Related issues**: N/A


tylerwilliams (commented):
The proto changes look reasonable, but all the code in prom.go that maps from HistogramType -> string is a little gnarly. Is it possible to determine this when we do the query? (either before, or from the results) and know that all histogram metrics are always going to also have _count, _sum, and _bucket metrics that need fetching?

luluz66:
I changed the proto message for caching purposes to be a wrapper around dto.MetricFamily(https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto)

I defned bbMetric that implements prometheus.Metric interface, which will be used in bbCollector.Collect.

Note: I need to bump prometheus/client_model from 0.3.0 to 0.4.0. Here is the change log: https://github.com/prometheus/client_model/compare/v0.3.0...v0.4.0.

enterprise/server/backends/prom/prom.go
```diff
@@ -16,28 +18,100 @@ import (
...
+	labelNames := make([]string, 0, len(m.metric.GetLabel()))
+
+	for _, label := range m.metric.Label {
+		labelNames = append(labelNames, *label.Name)
+	}
+	return prometheus.NewDesc(m.family.GetName(), m.family.GetHelp(), labelNames, nil)
+}
+
+func (m *bbMetric) Write(out *dto.Metric) error {
+	out.Label = m.metric.Label
```
vadimberezniker, line 99: Would `proto.Merge` work here?

vadimberezniker (approved):

PR #4203, opened 2023-06-23
Add Numeric flag.Value type

tempoz:
I ran into needing an int32 flag while migrating to the v2 AWS API, so I added  parameterized Numeric flag.Value type to handle all the primitive flag types go's flag library does not.

This also includes a little incidental cleanup of the flags library that I noticed while implementing numeric flags.

**Related issues**: N/A

### DIFF
--- a/server/util/flagutil/common/common.go
+++ b/server/util/flagutil/common/common.go
@@ -3,7 +3,6 @@ package common
 import (
 	"flag"
 	"reflect"
-	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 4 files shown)


tempoz:
* master
  * **PR #4203** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4203" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4203?utm_source=stack-comment).

vadimberezniker:
This feels a little bit overkill for supporting a single int32 configuration value. Could you use an int flag and cast it to int32?

tempoz:
Sure, I could cast an `int` or `int64` flag to `int32` after bounds-checking it. I just thought it would be nice to have all the numeric flags uniformly handled moving forward (including bounds-checking via the `strconv.Parse` functions) like they are in the standard `flag` lib, as well as not have to explicitly specify range constraints on flags when those constraints could be better expressed by just explicitly using the type.

vadimberezniker:
> Sure, I could cast an `int` or `int64` flag to `int32` after bounds-checking it. I just thought it would be nice to have all the numeric flags uniformly handled moving forward (including bounds-checking via the `strconv.Parse` functions) like they are in the standard `flag` lib, as well as not have to explicitly specify range constraints on flags when those constraints could be better expressed by just explicitly using the type.

I don't there's not enough need for these new flag types to warrant making this code more complex. My thinking is as follows:
 - These types of flags are uncommon and there's an easy workaround using an int flag
 - Bounds are unlikely to be a problem in practice
 - These flags will look different from other numeric flags because they won't simply be calls into `flag.Int` but will rather require using flagutil.New 

Overall I'm finding hard to justify the addition of these new types over what we have today.

tempoz:
> * Bounds are unlikely to be a problem in practice

Does this mean we would be casting to `int32` without bounds-checking? Because I basically wrote this to avoid the complexity of having to bounds-check whenever we use an `int` flag as an `int32`, but if you don't think bounds-checking them is important then I can scrap it.

tempoz:
Never mind; this just isn't that important. I'll work around it.

PR #4205, opened 2023-06-23
Allow snaploader to store and load COW artifacts

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -12,6 +12,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const (
+	// Suffix used for dirtied chunks.
+	dirtySuffix = ".dirty"
+)
+
 // Store models a block-level storage system, which is useful as a backend for a
 // Network Block Device (NBD) or userfaultfd(2).
 type Store interface {
@@ -44,7 +49,9 @@ func (f *File) SizeBytes() (int64, error) {
 
 // Mmap implements the Store interface using a memory-mapped file.
 type Mmap struct {
-	data []byte
+	data   []byte
+	mapped bool
+	path   string
 }
 
 // NewMmap returns an Mmap for an existing file.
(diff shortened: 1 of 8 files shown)


vadimberezniker (approved):

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -133,6 +193,12 @@ func (e *Hole) Close() error {
 	return nil
 }
 
+// Chunk represents a section of a larger composite store at a given offset.
+type Chunk struct {
+	Offset int64
+	Store
```
tylerwilliams, line 199: nit: feel like these should be swapped in order

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -163,13 +229,13 @@ type COW struct {
 
 // NewCOW creates a COW from the given chunks. The chunks should be open
 // initially, and will be closed when calling Close on the returned COW.
-func NewCOW(chunks []Store, chunkSizeBytes, totalSizeBytes int64, dataDir string) (*COW, error) {
+func NewCOW(chunks []*Chunk, chunkSizeBytes, totalSizeBytes int64, dataDir string) (*COW, error) {
```
tylerwilliams, line 232: I don't think calling things COW is particularly clear (because Copy On Write is really a strategy / adjective describing how something else, a blockstore, works)

What if we call this NewCOWStore or similar?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -207,6 +234,89 @@ func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotK
...
+	cf := &fcpb.ChunkedFile{
+		Name:      name,
+		Size:      size,
+		ChunkSize: cow.ChunkSizeBytes(),
+	}
+	for i, c := range cow.Chunks {
+		if _, ok := c.(*blockio.Hole); ok {
+			continue
+		}
+		if cow.Dirty(i) {
```
tylerwilliams, line 287: does snaploader need to look at each chunk like this and see that it's not a hole or could cow.Dirty() return false when the chunk is a hole (so snaploader doesnt need to know about holes)?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -207,6 +234,89 @@ func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotK
...
+			// Sync dirty chunks to make sure the underlying file is up to date
+			// before we add it to cache.
+			if err := c.Sync(); err != nil {
+				return nil, status.WrapError(err, "sync dirty chunk")
+			}
+		}
+		// TODO: It's too expensive to compute sha256 digests here for every
+		// chunk. Try out the following improvements:
```
tylerwilliams, line 295: Would be great to have a benchmark here :P so we know what too expensive means...

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -458,6 +549,31 @@ func IsEmptyOrAllZero(data []byte) bool {
 	return true
 }
 
+// expandHoles inserts explicit holes for any chunk start offsets not explicitly
+// represented in the given list. The input chunks are expected to be sorted in
+// increasing order of offset.
+func expandHoles(chunks []*Chunk, chunkSize, totalSize int64) []Store {
```
tylerwilliams, line 555: the holes idea feels a little premature and is adding a decent amount of complexity -- could we remove holes for now / treat all chunks the same and add them back later if necessary?

tylerwilliams (changes requested):

tylerwilliams (approved):

PR #4206, opened 2023-06-23
Add ClickHouse Docker test

luluz66:
1. Extract the logic to get a port from existing docker container that is shared
   by testmysql and testpostgres, and also testclickhouse into dockerutil.go
2. Add testclickhouse.go to set up a test clickhouse instance 
3. Modify target_tracket_test.go so that it can run both with clickhouse docker
and the default sql-lite.

**Related issues**: 
https://github.com/buildbuddy-io/buildbuddy-internal/issues/2011


server/testutil/testclickhouse/testclickhouse.go
```diff
@@ -0,0 +1,122 @@
...
+	options.Debug = true
+	require.NoError(t, err)
+	conn, err := clickhouse.Open(options)
+	require.NoError(t, err)
+
+	defer conn.Close()
+	for {
+		ctx := context.Background()
+		if err := conn.Ping(ctx); err != nil {
+			time.Sleep(1 * time.Second)
```
bduffany, line 91: failed pings are pretty cheap - we could ping more frequently here to help speed up the initial connection process

```suggestion
			time.Sleep(50 * time.Millisecond)
```

server/testutil/testclickhouse/testclickhouse.go
```diff
@@ -0,0 +1,122 @@
...
+			continue
+		}
+		// Drop the DB from the old test and let it be re-created via GORM
+		// auto-migration.
+		if reuseServer {
+			require.NoError(t, err)
+			defer conn.Close()
```
bduffany, line 98: `defer conn.Close()` is on line 87 as well so I think this one can be removed?

server/testutil/dockerutil/dockerutil.go
```diff
@@ -0,0 +1,30 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// FindPortByContainerNamePrefix lists running server by name, and if we find one
+// that looks like `<prefix>-$PORT>, then return the port and the container
+// name.
+func FindPortByContainerNamePrefix(t testing.TB, prefix string) (int, string) {
```
bduffany, line 16: wdyt about naming this "FindServerContainer"?

bduffany (approved):

server/testutil/testclickhouse/testclickhouse.go
luluz66: oops. i forgot to change back after debugging some connection issues.

PR #4207, opened 2023-06-23
remove shard count in usage_test_mysql and usage_test_postgres

luluz66:
When run usage_test_mysql and usage_test_postgres locally, 6/7 tests fail the 
second time because all the tests grab the same mysql port; and there seems to be
some concurrency issue when resetting databases.

We no longer need sharding since we have reuse_server set to true.
**Related issues**: N/A


bduffany:
hmm, this makes the tests take a little longer than I thought it would - mysql test takes 17s now, but before it only took 3-4s.

wdyt about using `--jobs=1` for now to work around this?

PR #4225, opened 2023-06-28
action_card: display failureDetail

sluongng:
**Related issues**:

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2449

### DIFF
--- a/app/target/action_card.tsx
+++ b/app/target/action_card.tsx
@@ -170,6 +170,19 @@ export default class ActionCardComponent extends React.Component<Props, State> {
           </div>
         )}
 
+        {(action?.failureDetail?.message?.length ?? 0) > 0 && (
+          <div className={`card ${this.props.dark ? "dark" : "light-terminal"}`}>
+            <PauseCircle className={`icon rotate-90 ${this.props.dark ? "white" : ""}`} />
+            <div className="content">
+              <TerminalComponent
+                title={<div className="title">Failure Message</div>}
+                value={action?.failureDetail?.message}
+                lightTheme={!this.props.dark}
+              />
+            </div>
+          </div>
+        )}
+
         <div className={`card ${action?.success ? "card-success" : "card-failure"}`}>
           <PlayCircle className="icon" />
           <div className="content">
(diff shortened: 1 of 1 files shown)


sluongng:
question: do we want to display the failure category and code mapping here?
In this particular example, the category + code mapping would be something like this:

```
Spawn failure: REMOTE_CACHE_FAILED (6)
```

See https://github.com/buildbuddy-io/buildbuddy/blob/f635f0927f10279f5c4311ce3a5c7d1a28c05f66/proto/failure_details.proto#L217

We could probably add it in another PR though.

sluongng:
Before

image

How it would look after

image

I wanted to incorporate the enum key in there somehow but was not able to find a nice way to do it. For example, given this and a value of `1`, I would like to get `NON_ZERO_EXIT` key as a string value 🤔.

```typescript
  export namespace Spawn {
    export enum Code {
      SPAWN_UNKNOWN = 0,

      /**
       * See the SpawnResult.Status Java enum for definitions of the following
       * Spawn failure codes.
       */
      NON_ZERO_EXIT = 1,
      TIMEOUT = 2,

      /**
       * Note: Spawn OUT_OF_MEMORY leads to a BUILD_FAILURE exit_code because the
       * build tool itself did not run out of memory.
       */
      OUT_OF_MEMORY = 3,
      EXECUTION_FAILED = 4,
      EXECUTION_DENIED = 5,
      REMOTE_CACHE_FAILED = 6,
      COMMAND_LINE_EXPANSION_FAILURE = 7,
      EXEC_IO_EXCEPTION = 8,
      INVALID_TIMEOUT = 9,
      ...
    }
  }
```

bduffany (approved):
Nice!

> I wanted to incorporate the enum key in there somehow but was not able to find a nice way to do it. For example, given this and a value of 1, I would like to get NON_ZERO_EXIT key as a string value thinking.

If you want to do this (either this PR or a separate one is fine) we have an example here of how to build an inverse mapping from numbers to values: https://github.com/buildbuddy-io/buildbuddy/blob/64f2a9bc25d15b0fe9378fe66f9c30e1a8f098f7/app/invocation/invocation_execution_util.tsx#L10-L12

sluongng:
oh wow... Hmm, I don't know if I want to do that, given that there are multiple categories under failure_details.proto, each having its own enum. 🙈 

I will create a different PR to update the proto first, then perhaps I could narrow things down to 1-2 relevant categories and stick with them.

PR #4230, opened 2023-06-28
Add Pebble Cache to benchmark cache test

luluz66:
Add Pebble Cache and Distributed Pebble Cache to benchmark cache test.

Also change the log level to error. Some warning logs make the benchmark result difficult to read. 
**Related issues**: N/A


vadimberezniker (approved):
Why are there so many warnings? Maybe we could fix them?

luluz66:
> Why are there so many warnings? Maybe we could fix them?

All the warnings are from disk_cache.go. "could not retrieve filesystem stats: no such file or directory"

PR #4231, opened 2023-06-28
Update atime at FindMissing in pebble cache

luluz66:
Calls sendAtimeUpdate in FindMissing instead of reader

Benchmark test: [before](https://app.buildbuddy.io/invocation/3ff0a8a7-fb9e-4f9e-a1ce-51a994fcc65c?target=%2F%2Fenterprise%2Fserver%2Ftest%2Fperformance%2Fcache%3Acache_test) and  [after](https://app.buildbuddy.io/invocation/3467f3fb-9869-49f2-9cae-d8701ded0697?target=%2F%2Fenterprise%2Fserver%2Ftest%2Fperformance%2Fcache%3Acache_test)

**Related issues**:
https://github.com/buildbuddy-io/buildbuddy-internal/issues/2450


luluz66 (commented):
This change made FindMissing slower because we read the proto. 

I wonder if the `olderThanThreshold` check is necessary in `sendATimeUpdate` since we also check it in `updateAtime`.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2930,7 +2931,6 @@ func (p *PebbleCache) reader(ctx context.Context, iter pebble.Iterator, r *rspb.
 		}
 		return nil, err
 	}
-	p.sendAtimeUpdate(key, fileMetadata)
```
siggisim, line 2933: Are there some artifacts where we don't call FindMissing and we only call Read on? Like AC entries for example?

I kinda feel like we need to bump atime in both places, but I might be wrong.

vadimberezniker:
> This change made FindMissing slower because we read the proto.
> 
> I wonder if the `olderThanThreshold` check is necessary in `sendATimeUpdate` since we also check it in `updateAtime`.

I wonder if we can use this trick to avoid deserializing the full proto: https://github.com/buildbuddy-io/buildbuddy/blob/64f2a9bc25d15b0fe9378fe66f9c30e1a8f098f7/server/util/bazel_request/bazel_request.go#L48

Having the check in `sendATimeUpdate` allows us to avoid using up space in the channel for unnecessary updates. It's a cheap check so removing it wouldn't really buy you anything.

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: I agree, I don't see a reason to remove the atime update from reads.

luluz66:
> > This change made FindMissing slower because we read the proto.
> > I wonder if the `olderThanThreshold` check is necessary in `sendATimeUpdate` since we also check it in `updateAtime`.
> 
> I wonder if we can use this trick to avoid deserializing the full proto:
> 
> https://github.com/buildbuddy-io/buildbuddy/blob/64f2a9bc25d15b0fe9378fe66f9c30e1a8f098f7/server/util/bazel_request/bazel_request.go#L48
> 
I had a impl to do this, but the performance is unfortunately worse: https://app.buildbuddy.io/invocation/173adf2d-de5a-43d4-8473-9131be8ec718?target=%2F%2Fenterprise%2Fserver%2Ftest%2Fperformance%2Fcache%3Acache_test#@79. Probably because the last access usec is the second to the last field, so it didn't save much.

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: +1 to this -- let's keep this one!

vadimberezniker:
> > > This change made FindMissing slower because we read the proto.
> > > I wonder if the `olderThanThreshold` check is necessary in `sendATimeUpdate` since we also check it in `updateAtime`.
> > 
> > 
> > I wonder if we can use this trick to avoid deserializing the full proto:
> > https://github.com/buildbuddy-io/buildbuddy/blob/64f2a9bc25d15b0fe9378fe66f9c30e1a8f098f7/server/util/bazel_request/bazel_request.go#L48
> 
> I had a impl to do this, but the performance is unfortunately worse: https://app.buildbuddy.io/invocation/173adf2d-de5a-43d4-8473-9131be8ec718?target=%2F%2Fenterprise%2Fserver%2Ftest%2Fperformance%2Fcache%3Acache_test#@79. Probably because the last access usec is the second to the last field, so it didn't save much.

This is a bit surprising. Can you double check your implementation? The allocations should not be higher than the proto version. 
Can you also re-run the benchmarks locally? You're not going to get consistent numbers running benchmarks on remote executors.

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: Done

luluz66 (commented):
> This is a bit surprising. Can you double check your implementation? The allocations should not be higher than the proto version. Can you also re-run the benchmarks locally? You're not going to get consistent numbers running benchmarks on remote executors.

Yeah. I fixed the implementation. The protowire change is now slightly slower than the master branch; but faster than looking up the entire proto. The benchmark result can be found at https://docs.google.com/document/d/1EtTaCd12bdgURIr8vqJXxNtuA0vdowo1BwiV3HQ8obI/edit.

vadimberezniker:
> > This is a bit surprising. Can you double check your implementation? The allocations should not be higher than the proto version. Can you also re-run the benchmarks locally? You're not going to get consistent numbers running benchmarks on remote executors.
> 
> Yeah. I fixed the implementation. The protowire change is now slightly slower than the master branch; but faster than looking up the entire proto. The benchmark result can be found at https://docs.google.com/document/d/1EtTaCd12bdgURIr8vqJXxNtuA0vdowo1BwiV3HQ8obI/edit.

Thanks. Speed wise both versions seem acceptable. The protowire version is appealing due to no increase in allocations. 
We should add a comment to the proto wrt the tag number. 

PR looks good to me, but let @tylerwilliams opine as well.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1298,6 +1299,43 @@ func (p *PebbleCache) lookupFileMetadata(ctx context.Context, iter pebble.Iterat
...
+			return 0, err
+		}
+		if iter.SeekGE(keyBytes) && bytes.Equal(iter.Key(), keyBytes) {
+			return getLastAccessUsec(iter.Value()), nil
+		}
+	}
+	return 0, status.NotFoundErrorf("key %q not found", key)
+}
+
+func getLastAccessUsec(b []byte) int64 {
```
tylerwilliams, line 1315: nit: comment why this is this way and link to the proto tag field definition (or explain that lastAccessUsecTag matches it)

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1434,10 +1472,11 @@ func (p *PebbleCache) FindMissing(ctx context.Context, resources []*rspb.Resourc
 		}
 
 		unlockFn := p.locker.RLock(key.LockID())
-		found, err := p.iterHasKey(iter, key)
-		if !found || err != nil {
+		lastAccessUsec, err := p.lookupLastAccessTime(iter, key)
+		if err != nil {
 			missing = append(missing, r.GetDigest())
 		}
+		p.sendAtimeUpdate(key, lastAccessUsec)
```
tylerwilliams, line 1479: I don't think we need to sendAtimeUpdate if the file was missing. (correct me if i'm wrong?)

tylerwilliams (changes requested):

tylerwilliams (approved):

PR #4237, opened 2023-06-29
Don't call setInterval with larger than a day

siggisim:
Fun quirk I found when working on auth stuff, if you call setInterval with larger than a 32 bit integer - you get an overflow and the setInterval function gets spammed:
https://stackoverflow.com/questions/39007056/is-there-a-maximum-delay-limit-for-window-setinterval


tylerwilliams (approved):
lol nice

PR #4239, opened 2023-06-30
Pull some more stuff out of oidc.go into claims.go and cookie.go

siggisim:
These methods are used in both oidc and github auth.

We pull the following methods into cookie.go:
- SetLoginCookie
- ClearLoginCookie

And the following methods into claims.go
- ClaimsFromContext
- AuthContextFromClaims

### DIFF
--- a/server/util/claims/BUILD
+++ b/server/util/claims/BUILD
@@ -10,6 +10,7 @@ go_library(
         "//server/environment",
         "//server/interfaces",
         "//server/tables",
+        "//server/util/authutil",
         "//server/util/capabilities",
         "//server/util/flagutil",
         "//server/util/lru",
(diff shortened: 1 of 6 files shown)


tylerwilliams (approved):

server/util/claims/BUILD
```diff
@@ -10,6 +10,7 @@ go_library(
         "//server/environment",
```
bduffany, line 10: uber nit (non blocking) - do we want these 2 pkgs under //enterprise?

bduffany (approved):

server/util/claims/BUILD
siggisim: Good point, will move them there - but after all this to keep the diffs clean and not screw up my changes too much

PR #4241, opened 2023-06-30
Add blockio support for UFFD

bduffany:
Provide an API that will let the UFFD handler map a VM page address to a memory address within the COWStore.

**Related issues**: N/A


enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -220,12 +235,50 @@ func NewCOWStore(chunks []*Chunk, chunkSizeBytes, totalSizeBytes int64, dataDir
...
+// GetPageAddress returns a memory address for the given byte offset which can
+// be used to handle a page fault with userfaultfd.
+//
+// If reading a lazily mmapped chunk, this will cause the chunk to be mmapped
+// so that the returned address is valid.
+//
+// If write is set to true, and the page is not dirty, then a copy is first
+// performed so that the returned chunk can be written to without modifying
+// readonly chunks.
+func (c *COWStore) GetPageAddress(address uintptr, write bool) (uintptr, error) {
```
maggie-lou, line 254: nit: address -> storeOffset?

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -220,12 +235,50 @@ func NewCOWStore(chunks []*Chunk, chunkSizeBytes, totalSizeBytes int64, dataDir
...
+	if write && !c.dirty[chunkStartOffset] {
+		if err := c.copyChunk(chunkStartOffset); err != nil {
+			return 0, status.WrapError(err, "copy chunk")
+		}
+	}
+	chunk := c.chunks[chunkStartOffset]
+	if chunk == nil {
+		// No data (yet); map into our static zero-filled buf. Note that this
+		// can only happen for reads, since for writes we call copyChunk above.
+		ptr := uintptr(unsafe.Pointer(&c.zeroBuf[0]))
```
maggie-lou, line 266: nit: What do you think of adding a helper function for uintptr(unsafe.Pointer() so it's more clear that is just getting the start address of an array?

maggie-lou (approved):

PR #4246, opened 2023-07-01
Fix CLI offline mode: return default capabilities if unable to communicate with server

siggisim:
Fixes: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2316

### DIFF
--- a/cli/cache_proxy/cache_proxy.go
+++ b/cli/cache_proxy/cache_proxy.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"os"
 	"time"
@@ -24,6 +25,7 @@ import (
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	rspb "github.com/buildbuddy-io/buildbuddy/proto/resource"
+	smpb "github.com/buildbuddy-io/buildbuddy/proto/semver"
 	bspb "google.golang.org/genproto/googleapis/bytestream"
 )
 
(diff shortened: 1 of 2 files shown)


cli/cache_proxy/cache_proxy.go
```diff
@@ -370,3 +374,40 @@ func (qw *queueWorker) RemoteWriteBlocked(ctx context.Context, wreq *bspb.WriteR
 	}
 	return qw.handleWriteRequest(req)
 }
+
+func defaultCapabilities() *repb.ServerCapabilities {
+	var compressors = []repb.Compressor_Value{repb.Compressor_IDENTITY, repb.Compressor_ZSTD}
+	return &repb.ServerCapabilities{
+		// Support bazel 2.0 -> 2.3
```
brentleyjones, line 381:
```suggestion
		// Support REAPI 2.0 -> 2.3
```

cli/cache_proxy/cache_proxy.go
```diff
@@ -123,7 +125,9 @@ func (p *CacheProxy) GetCapabilities(ctx context.Context, req *repb.GetCapabilit
 			log.Infof("Remote capabilities: %s", string(b))
 		}
```
bduffany, line 126:
```suggestion
		}
		return res
```

bduffany: (or better yet, invert the logic so that this clause is checking `err != nil` and doing the fallback in that case - generally `err == nil` feels a bit harder to read than `err != nil` since the latter is so much more common)

siggisim: Good catch, inverted the logic and added the missing return. PTAL

bduffany (approved):

PR #4247, opened 2023-07-01
Populate user's githubToken if logging in via githubauth

siggisim:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/githubauth/githubauth.go
+++ b/enterprise/server/githubauth/githubauth.go
@@ -22,11 +22,12 @@ import (
 )
 
 const (
-	githubIssuer   = "https://github.com"
-	contextUserKey = "auth.githubuser"
-	loginPath      = "/login/github/"
-	authPath       = "/auth/github/"
-	jwtDuration    = 24 * 365 * time.Hour
+	githubIssuer      = "https://github.com"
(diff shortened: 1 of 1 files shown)


enterprise/server/githubauth/githubauth.go
```diff
@@ -211,28 +220,28 @@ func (a *githubAuthenticator) SSOEnabled() bool {
 	return false
 }
 
-func (a *githubAuthenticator) authenticateUser(w http.ResponseWriter, r *http.Request) (*claims.Claims, *github.GithubUserResponse, error) {
+func (a *githubAuthenticator) authenticateUser(w http.ResponseWriter, r *http.Request) (*claims.Claims, *github.GithubUserResponse, *tables.Session, error) {
```
bduffany, line 223: nit: it might be a bit cleaner to have a struct like

```
type authenticatedGitHubUser struct {
    Profile *github.GithubUserResponse
    AccessToken string
}
```

and then return that from this func / store it in ctx under `auth.githubuser`

bduffany (approved):

enterprise/server/githubauth/githubauth.go
siggisim: Much cleaner, done!

PR #4255, opened 2023-07-05
Enable golang staticcheck analyzer S1019 (Simplify make call by omitting redundant arguments)

iain-macdonald:
**Related issues**: N/A


tylerwilliams:
I don't really care about this one -- feels like 6 of one, half dozen of the other. Don't feel it really improves clarity much. Can we skip?

PR #4259, opened 2023-07-06
nogo_config.json: programatically generate the json config

sluongng:
Use skylib's write_file to programatically create the json config file
from starlark structs.

Also adjusted the `external/` regex to support Windows path separator.

**Related issues**: N/A


BUILD
```diff
@@ -1,14 +1,80 @@
...
+                    "exclude_files": {
+                        ".*_cgo_gotypes\\.go": "cgo internal files",
+                    },
+                },
+            } | {
+                analyzer: {
+                    "only_files": {
+                        "/server/": "Server code",
+                        "/enterprise/server/": "Enterprise server code",
+                    },
```
sluongng, line 49: Not quite sure why we don't apply these checks to the rest of the code base. I.e. `//cli/...` code is being excluded

Probably worth investigating in the future

iain-macdonald (approved):

PR #4263, opened 2023-07-06
workflows: allow fetching private repo dependencies matching target repo host

bduffany:
Tested with a private repo. The only issue I ran into was that the GH app also needs to be configured with access to the private repo, and the error message is confusing if it doesn't have access (the error message is just "repo not found").

Decided to use a credential helper instead of the `url.{}.insteadOf` trick, to avoid storing credentials on disk.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2173


siggisim (approved):

sluongng:
😍 this is amazing!

tylerwilliams (approved):

PR #4264, opened 2023-07-06
CLI: Don't apply .bazelrc when running `bazel help` to collect metadata

siggisim:
If you apply .bazelrc options and you have `common` flags in your .bazelrc to the `bazel help` command, which may be invalid.

This causes problems until this Bazel fix is in (but will continue to cause problems on old versions):
https://github.com/bazelbuild/bazel/pull/18609


bduffany (approved):
woops, good catch

PR #4266, opened 2023-07-06
CLI: Support relative paths for explicit .bazelrc flags

siggisim:
**Related issues**: N/A

### DIFF
--- a/cli/parser/parser.go
+++ b/cli/parser/parser.go
@@ -853,8 +853,8 @@ func consumeRCFileArgs(args []string, workspaceDir string) (newArgs []string, rc
 				continue
 			}
 
-			if !filepath.IsAbs(bazelrcArg) {
-				bazelrcArg = filepath.Join(workspaceDir, bazelrcArg)
+			if a, err := filepath.Abs(bazelrcArg); err == nil {
+				bazelrcArg = a
 			}
 
 			explicitBazelrcPaths = append(explicitBazelrcPaths, bazelrcArg)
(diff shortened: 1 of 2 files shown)


cli/parser/parser.go
```diff
@@ -852,6 +852,11 @@ func consumeRCFileArgs(args []string, workspaceDir string) (newArgs []string, rc
 				encounteredDevNullBazelrc = true
 				continue
 			}
+
+			if !filepath.IsAbs(bazelrcArg) {
+				bazelrcArg = filepath.Join(workspaceDir, bazelrcArg)
+			}
```
bduffany, line 858: (since bazel can be invoked from a subdir of the workspace)

```suggestion
			if a, err := filepath.Abs(bazelrcArg); err == nil {
				bazelrcArg = a
			}
```

siggisim: Looks like this fails the test. I feel like we need the current working directory in there somehow, right?

siggisim: Oh looks like that’s what Abs does. Hmm…

bduffany: yeah good question, I wasn't actually sure how bazel deals with that.

siggisim:
Closing this because I think https://github.com/buildbuddy-io/buildbuddy/pull/4272 was the real culprit here. Relative paths already seem to work, and the test just failed because the .bazelrc file was written to a testfs directory instead of the workspace root.

PR #4267, opened 2023-07-06
CLI: Only set sync flag if we need a sidecar

siggisim:
Before this change, we'd unnecessarily start the sidecar if `--sync=true` is set (which it is if the CI=true environment variable is set) but BES / cache flags were not set.

### DIFF
--- a/cli/sidecar/sidecar.go
+++ b/cli/sidecar/sidecar.go
@@ -155,14 +155,15 @@ func ConfigureSidecar(args []string) []string {
 		sidecarArgs = append(sidecarArgs, fmt.Sprintf("--cache_dir=%s", diskCacheDir))
 	}
 
+	// If there aren't any sidecar arguments, we don't need to spin up a sidecar at all.
+	if len(sidecarArgs) == 0 {
+		return args
+	}
+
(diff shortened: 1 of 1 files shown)


bduffany (approved):

cli/sidecar/sidecar.go
```diff
@@ -155,16 +155,17 @@ func ConfigureSidecar(args []string) []string {
...
 
+	// If there aren't any sidecar arguments, we don't need to spin up a sidecar at all.
+	if len(sidecarArgs) == 0 {
+		return args
+	}
+
 	if synchronousWriteFlag == "1" || synchronousWriteFlag == "true" {
 		sidecarArgs = append(sidecarArgs, "--synchronous_write")
 		sidecarArgs = append(sidecarArgs, "--bes_synchronous")
 		args = append(args, "--bes_upload_mode=wait_for_upload_complete")
```
bduffany, line 166: Unrelated to this PR, but it'd be nice if we could enable this flag whenever CI=true

siggisim: Not a bad idea, let me think about the best way to do this. 

I think a suggestion in the UI could be a good idea because then it helps non-CLI folks as well

PR #4268, opened 2023-07-06
Protos for audit logging.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/proto/auditlog.proto
+++ b/proto/auditlog.proto
@@ -60,10 +60,13 @@ message Entry {
   // Additional metadata describing the change performed.
   // Only one field will be populated depending on the event type.
   message ResourceState {
-    APIKeyState api_key = 1;
-    GroupState group = 2;
-    GroupUserState group_users = 3;
-    InvocationState invocation = 4;
+    api_key.CreateApiKeyRequest create_api_key = 1;
+    api_key.GetApiKeysRequest get_api_keys = 2;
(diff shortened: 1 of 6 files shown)


proto/auditlog.proto
```diff
@@ -0,0 +1,94 @@
...
+}
+
+message GetAuditLogsRequest {
+  context.RequestContext request_context = 1;
+
+  string page_token = 2;
+  google.protobuf.Timestamp timestamp_after = 3;
+  google.protobuf.Timestamp timestamp_before = 4;
+  ResourceID resource = 5;
+  string method = 6;
```
tylerwilliams, line 86: we probably could omit these for now

tylerwilliams (approved):

proto/auditlog.proto
```diff
@@ -0,0 +1,120 @@
...
+
+message AuthenticationInfo {
+  // Populated for events on behalf of authenticated users.
+  AuthenticatedUser user = 1;
+  // Populated for events on behalf of authenticated API keys.
+  AuthenticatedAPIKey api_key = 2;
+  string client_ip = 3;
+}
+
+enum ResourceType {
```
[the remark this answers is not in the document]
vadimberezniker: I prefer top level. Simpler to reference from go code.

proto/auditlog.proto
vadimberezniker: done

PR #4269, opened 2023-07-06
Clickhouse schema for audit logs.

vadimberezniker:
**Related issues**: N/A


server/util/clickhouse/schema/schema.go
```diff
@@ -283,6 +289,44 @@ func (t *TestTargetStatus) TableOptions() string {
 	return fmt.Sprintf("ENGINE=%s ORDER BY (group_id, repo_url, commit_sha, label, invocation_uuid)", getEngine())
 }
 
+type AuditLog struct {
+	AuditLogID    string
+	GroupID       string `gorm:"primaryKey;"`
```
luluz66, line 294: If I remember correctly, gorm-clickhouse doesn't support "primaryKey", so the sort key is the primary key. Maybe we can drop this?

luluz66 (approved):

server/util/clickhouse/schema/schema.go
vadimberezniker: Yeah, you're right. The clickhouse gorm automigrator doesn't support setting primary key.

PR #4270, opened 2023-07-06
Service for logging audit events.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/auditlog/auditlog.go
+++ b/enterprise/server/auditlog/auditlog.go
@@ -1,0 +1,257 @@
+package auditlog
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
+	"github.com/buildbuddy-io/buildbuddy/server/util/authutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/clickhouse/schema"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/query_builder"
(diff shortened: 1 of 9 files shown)


server/environment/environment.go
```diff
@@ -162,4 +162,6 @@ type Env interface {
 	SetSingleFlightDeduper(interfaces.SingleFlightDeduper)
 	GetPromQuerier() interfaces.PromQuerier
 	SetPromQuerier(interfaces.PromQuerier)
+	GetAuditLog() interfaces.AuditLog
+	SetAuditLog(log interfaces.AuditLog)
```
tylerwilliams, line 166: could we call this GetAuditLogger/SetAuditLogger instead?

thinking logger is the object, log is the method we call on it

enterprise/server/auditlog/auditlog.go
```diff
@@ -0,0 +1,263 @@
...
+	ui, err := l.env.GetUserDB().GetUser(ctx)
+	if err != nil {
+		return status.WrapError(err, "could not lookup user")
+	}
+
+	request = clearRequestContext(request)
+
+	var requestBytes []byte
+	if request != nil {
+		rp, err := l.wrapRequestProto(request)
```
tylerwilliams, line 158: idk if you want to do any validation here to ensure that the request proto type matches the resource ID type (and bail if it does not)

server/interfaces/interfaces.go
```diff
@@ -1160,3 +1163,8 @@ type PromQuerier interface {
 type ConfigSecretProvider interface {
 	GetSecret(ctx context.Context, name string) ([]byte, error)
 }
+
+type AuditLog interface {
```
tylerwilliams, line 1167: nit: AuditLogger

tylerwilliams (changes requested):

tylerwilliams (approved):

PR #4271, opened 2023-07-07
nogo: make config json consistent

sluongng:
There are several changes here:

- Disable fieldalignment completely (instead of excluding all files from
  being checked).

- Remove nilness config. The analyzer was removed from an earlier change.

- Switch from including only files under `sever/` and
  `enterprise/server/` to excluding all `external/` files. This will
  include Go files from other directories such as `cli/`.

- Programatically replace all unix path separators with a regex that
  matches both Windows and Unix separators.

- Add comment with example on how to check the rendered json config
  file.

- Add missing configs for some analyzers,
  effectively ignore all external dependencies during linting.

**Related issues**: N/A

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -7,6 +7,9 @@ load("//rules/go:index.bzl", "go_sdk_tool")
 
 package(default_visibility = ["//visibility:public"])
 
+# Rendered JSON result could be checked by doing:
+#   bazel build //:no_go_config
+#   cat bazel-bin/no_go_config.json | jq .
 write_file(
     name = "nogo_config",
     out = "nogo_config.json",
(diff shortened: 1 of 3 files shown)


BUILD
```diff
@@ -7,66 +7,77 @@ load("//rules/go:index.bzl", "go_sdk_tool")
...
+                        },
                     },
-                }
-                for analyzer in ANALYZERS
+                    "copylocks": {
+                        "exclude_files": {
+                            "external/": "third_party",
+                            # TODO(sluongng): this should be fixed on rules_go side
+                            # https://github.com/bazelbuild/rules_go/issues/3619
+                            "cgo/github.com/shirou/gopsutil/": "third_party cgo",
```
iain-macdonald, line 41: This is getting kind of complicated. Could you add this exclude to all analyzers and put `copylocks` in the list below, and remove all the loops and stuff above here in favor of a static definition which would be easier to understand?

BUILD
```diff
@@ -77,35 +88,39 @@ nogo(
...
-        "@org_golang_x_tools//go/analysis/passes/stringintconv:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/structtag:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/tests:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/testinggoroutine:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/unmarshal:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/unreachable:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/unsafeptr:go_default_library",
-        "@org_golang_x_tools//go/analysis/passes/unusedresult:go_default_library",
-        "@com_github_nishanths_exhaustive//:exhaustive",
+    deps = ["@com_github_nishanths_exhaustive//:exhaustive"] + [
```
iain-macdonald, line 91: This doesn't feel necessary to me, could we just leave it as a list of analyzers without the programmatic creation?

BUILD
sluongng: Oh this is a brilliant idea!

BUILD
sluongng: My original intention was to have both the analyzer name in the config file and the package name here to come from the same Starlark array. I.e.

```
MY_ANALYZERS = [
    "asmdecl",
    "assign",
    ...
]
```

However I realized that the "analyzer's name" and the "package's name" are only _almost_ the same. I.e. `copylock` vs `copylocks` 🙄. So I had to use 2 arrays instead and things turned ugly.

Your suggestion makes a lot more sense than my mumbo jumbo 👍.
Fixed.

BUILD
```diff
@@ -77,35 +74,36 @@ nogo(
...
```
iain-macdonald, line 77: Any reason to concatenate two arrays here rather than just make `exhaustive` an entry in the longer array?

iain-macdonald (approved):

BUILD
sluongng: I was planning on adding a few more checks into the mix in the future, so it's good to maintain a separate list vs the checks in golang/x/tools.

However, that's not needed for now. I pushed a change to return the check to it's original place.

PR #4274, opened 2023-07-07
vfs: provide noop impl for Windows

sluongng:
Provide no-op implementation for `vfs` and `vfs_server` on Windows

**Related issues**: N/A


sluongng:
The relationship from executor down to `hanwen/go-fuse/v2` packages.

image

After this PR, the expectation is compiling Executor on Windows would no longer require FUSE.
We could investigate using https://github.com/winfsp/winfsp and https://github.com/cloud-hypervisor/cloud-hypervisor(supports Windows MicroVM) in the future, but that's a long story from now.

sluongng:
image

both `//enterprise/server/util/vfs_server` and `//enterprise/server/remote_execution/workspace` could be compiled on Windows after this PR.

The error before this PR looks something like this

image

bduffany (approved):

PR #4278, opened 2023-07-07
add a workflows page for developers (i.e., not admins)

jdhollen:
**Related issues**: N/A


enterprise/app/workflows/workflows_user.tsx
```diff
@@ -0,0 +1,140 @@
...
+export type WorkflowsUserProps = {
+  user: User;
+};
+
+type State = {
+  workflowHistoryLoading: boolean;
+  workflowHistoryResponse: workflow.GetWorkflowHistoryResponse | null;
+};
+
+export default class WorkflowsUserComponent extends React.Component<WorkflowsUserProps> {
```
bduffany, line 24: I feel like it would be simpler to hide the few controls that are admin specific, rather than copying this component code. I think it's just
- The controls section on the RHS of each workflow item
- The "Link a repository" button

wdyt?

jdhollen: Honestly, I don't know.  I don't really like forking but I also don't really like showing / hiding.  Maybe it's a bit premature to split things, though..

Another consideration is just what RPCs do we want to have available / to whom.  Right now I'm just taking the results from GetHistory on the developer page and showing those (things that aren't actively running don't show).  Maybe this is just straight bad and we could expose the other two get calls for developers, too?

bduffany: It seems fine to me to allow developers to get linked repositories / workflows, just not modify them. @siggisim wdyt?

enterprise/app/workflows/workflows.tsx
```diff
@@ -289,60 +281,70 @@ class ListWorkflowsComponent extends React.Component<ListWorkflowsProps, State>
...
-            submitLabel="Okay"
-            destructive
-            onSubmit={() => this.setState({ showCleanWorkflowWarning: false })}>
-            <p>Are you sure you want to run the workflow in a clean container?</p>
-            <p>
-              This will prevent all existing workflow containers from being reused by other workflow runs, making them
-              slower, so this flag is not encouraged.
-            </p>
-          </SimpleModalDialog>
+          {isAdmin && (
```
bduffany, line 296: nit: for simplicity, we can keep the modals rendering even if `!isAdmin`, since the controls to show them are invisible and the worst case is that the modal is somehow visible but the RPC is denied.

enterprise/app/workflows/workflows.tsx
```diff
@@ -459,7 +461,7 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
       <div className="workflow-item container">
         <div className="workflow-item-column">
           <div className="workflow-item-row">
-            {this.state.runWorkflowActionStatuses && (
+            {this.props.isAdmin && this.state.runWorkflowActionStatuses && (
```
bduffany, line 464: nit: runWorkflowActionStatuses won't be populated since the "run" control is hidden, so we can leave this as-is too

enterprise/server/workflow/service/service.go
```diff
@@ -421,11 +421,11 @@ func (ws *workflowService) providerForRepo(u *url.URL) (interfaces.GitProvider,
...
-func (ws *workflowService) GetWorkflows(ctx context.Context, req *wfpb.GetWorkflowsRequest) (*wfpb.GetWorkflowsResponse, error) {
+func (ws *workflowService) GetWorkflows(ctx context.Context, protoContext *ctxpb.RequestContext) (*wfpb.GetWorkflowsResponse, error) {
 	if err := ws.checkPreconditions(ctx); err != nil {
 		return nil, err
 	}
-	groupID, err := perms.AuthenticateSelectedGroupID(ctx, ws.env, req.GetRequestContext())
+	groupID, err := perms.AuthenticateSelectedGroupID(ctx, ws.env, protoContext)
 	if err != nil {
 		return nil, err
 	}
```
bduffany, line 431: I should probably remove this `AuthenticateSelectedGroupID` func at some point - the new way to get the authenticated user's selected group ID is

```
u, err := perms.AuthenticatedUser(ctx)
if err != nil {
    return err
}
groupID := u.GetGroupID()
```

that way you can avoid the awkward `protoContext` param

bduffany (approved):

enterprise/app/workflows/workflows_user.tsx
jdhollen: Alright, just one file it is.

PR #4279, opened 2023-07-07
Audit logs for organization updates.

vadimberezniker:
**Related issues**: N/A


enterprise/server/backends/userdb/userdb.go
```diff
@@ -499,22 +541,11 @@ func (d *UserDB) UpdateGroupUsers(ctx context.Context, groupID string, updates [
...
-						DELETE FROM "APIKeys"
-						WHERE user_id = ? AND group_id = ?`,
-					update.GetUserId().GetId(),
-					groupID).Error; err != nil {
+				if err := d.removeUserFromGroup(ctx, tx, update.GetUserId().GetId(), groupID); err != nil {
 					return err
 				}
 			case grpb.UpdateGroupUsersRequest_Update_ADD:
-				if err := d.addUserToGroup(tx, update.GetUserId().GetId(), groupID); err != nil {
+				if err := d.addUserToGroup(ctx, tx, update.GetUserId().GetId(), groupID, role.FromProto(update.GetRole())); err != nil {
```
bduffany, line 548: super duper nit: Today we're always passing UNKNOWN (i.e. unset) from the UI here which means FromProto today would always return Developer, which is a slight behavior change from today where this is effectively always `role.Default`

To preserve the previous behavior we might want to update `FromProto` to return `role.Default` whenever role is UNKNOWN. 

This doesn't really matter because `role.Default` is hard-coded as Developer today but at some point we might want to make this configurable.

bduffany (approved):

PR #4280, opened 2023-07-07
config: Propagate parse failures upward

minor-fixes:
This change returns the parse error upward when a config fails to parse, instead of early-returning no error. This fixes an issue where a malformed config results in a confusing failure to launch:

```
2023/07/07 22:32:10.053 INF BuildBuddy v2.12.49 (933606adb4ed2c807eab7a9edcc27782eb0e1d10) compiled with go1.20.5 X:nocoverageredesign
2023/07/07 22:32:10.054 INF Reading buildbuddy config from '/local/config.yaml'
2023/07/07 22:32:10.062 INF Auto-migrating DB
2023/07/07 22:32:10.078 INF Successfully configured sqlite database.
2023/07/07 22:32:10.079 WRN No authentication will be configured: rpc error: code = Internal desc = OIDC authenticator failed to configure: rpc error: code = FailedPrecondition desc = No auth providers specified in config!
2023/07/07 22:32:10.080 FTL rpc error: code = Internal desc = Invalid Remote Execution Redis config.
```

This is happening becuase the parse failure essentially causes the config to be discarded, at which point default values are used, which happens to result in an invalid configuration.

With this change, Buildbuddy will still fail to launch, but will print the actual config parse error instead.

Tested: Built an image with this change and used to debug a config that wouldn't launch properly after updating (spoiler: turns out it's always been malformed, and is only now properly validated)

**Related issues**: N/A

### DIFF
--- a/server/config/config_test.go
+++ b/server/config/config_test.go
@@ -225,4 +225,29 @@ secret_struct_slice_flag:
 		require.NoError(t, err)
 		require.Equal(t, []secretHolder{{Secret: "FIRST\nSECRET"}, {Secret: "SECOND\nSECRET"}}, *secretStructSliceFlag)
 	}
+
+}
+
+func TestLoadFromData(t *testing.T) {
+	// Can successfully parse int config item
+	{
+		flags := replaceFlagsForTesting(t)
+		_ = flags.Int("must_be_a_number", 0, "")
+		err := config.LoadFromData(strings.TrimSpace(`
+must_be_a_number: 4
+	`))
+		require.NoError(t, err)
+	}
+
+	// Parse error when config value causes type mismatch
+	{
+		flags := replaceFlagsForTesting(t)
+		_ = flags.Int("must_be_a_number", 0, "")
+		err := config.LoadFromData(strings.TrimSpace(`
+must_be_a_number: "not a string, like this"
+	`))
+		require.Error(t, err)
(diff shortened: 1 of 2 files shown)


vadimberezniker (approved):
Thanks for finding this.

Looks I unintentionally introduced this in https://github.com/buildbuddy-io/buildbuddy/pull/4172

minor-fixes:
No problem! I rebased onto master, and added a test which fails on master but now passes - PTAL

server/config/config_test.go
```diff
@@ -225,4 +225,16 @@ secret_struct_slice_flag:
 		require.NoError(t, err)
 		require.Equal(t, []secretHolder{{Secret: "FIRST\nSECRET"}, {Secret: "SECOND\nSECRET"}}, *secretStructSliceFlag)
 	}
+
+	// Parse errors are propagated
```
vadimberezniker, line 229: Can you move this testcase to a standalone function? Although the bug was introduced as part of the "expansion" feature, the bug itself has nothing to do with it.

vadimberezniker (changes requested):

server/config/config_test.go
minor-fixes: Done - moved to `TestLoadFromData`, with both a parse-working and parse-broken test. Does this naming look ok, tests follow the conventions you all prefer?

vadimberezniker (approved):

PR #4284, opened 2023-07-10
Add audit logs for API key operations.

vadimberezniker:
**Related issues**: N/A


tylerwilliams (changes requested):
LGTM but adding @bduffany so we have multiple sets of eyes on any auth changes

enterprise/server/backends/authdb/authdb.go
```diff
@@ -8,6 +8,9 @@ import (
 	"fmt"
 	"strings"
 
+	akpb "github.com/buildbuddy-io/buildbuddy/proto/api_key"
```
bduffany, line 11: All the diffs in this file (and the BUILD file) look like NOPs - revert?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -480,6 +480,9 @@ func (s *BuildBuddyServer) GetApiKeys(ctx context.Context, req *akpb.GetApiKeysR
 			VisibleToDevelopers: k.VisibleToDevelopers,
 		})
 	}
+	if al := s.env.GetAuditLog(); al != nil {
+		al.Log(ctx, auditlog.GroupAPIKeyResourceID(""), auditlog.ListAPIKeys, req)
```
bduffany, line 484: for my own understanding - are we using a convention that an empty resource ID always represents the list of all the resources? (if so, are there other resource types where that convention would not work due to empty string being valid, and maybe lead to confusion?)

bduffany: hm actually since the operation name is ListAPIKeys I think this is fine.

bduffany (approved):

server/buildbuddy_server/buildbuddy_server.go
vadimberezniker: This is basically setting resource_type = API_KEY w/o setting an id. This will allow the "list API keys" operations to be returned if the user queries all logs for API Key resources.

tylerwilliams (approved):

PR #4286, opened 2023-07-11
server/util/db: allow compilation without cgo

sluongng:
Provide an alternative compilcation option when build without cgo on
some platforms. In particular, remove sqlite support when cgo is off.

**Related issues**: N/A


tempoz (approved):
LGTM

PR #4289, opened 2023-07-11
proto: update a few proto from Bazel's master branch

sluongng:
Cherry picked a few changes from Bazel's latest proto into our repo.

These were taken from commit
https://github.com/bazelbuild/bazel/commit/57ae1ff52ffbb2494ac9d43c1e49c2ba9425495e

**Related issues**: N/A


proto/remote_execution.proto
```diff
@@ -352,12 +368,16 @@ service ContentAddressableStorage {
   // Clients can use this API before uploading blobs to determine which ones are
   // already present in the CAS and do not need to be uploaded again.
   //
-  // Servers SHOULD increase the lifetimes of the referenced blobs if necessary and
-  // applicable.
+  // Servers SHOULD increase the lifetimes of the referenced blobs if necessary
+  // and applicable.
```
bduffany, line 372: any idea why these comment formatting diffs are happening? this looks different than the upstream proto: https://github.com/bazelbuild/remote-apis/blob/068363a3625e166056c155f6441cfb35ca8dfbf2/build/bazel/remote/execution/v2/remote_execution.proto#L347-L348

is it because of clang? we have `// clang-format off` at the top of the file, so clang should be disabled for this file at least...

sluongng: 🤔 strange... I recall only running `build-fix.sh` over the file.

Let me check if I could restore these.

sluongng: ah, what I did was 

```
cp ${bazel}/*.proto ${buildbuddy}/proto/
./buildfix.sh
...
git restore -p
```

This causes a problem because the formatting header would only be restored _AFTER_ the clang-format run.
Let me restore those hunks manually before running the linter

sluongng: I manually cherry pick a fix for this.

Could you please give it another look? 🙈

bduffany (approved):
Nice, thank you for updating these :)

PR #4298, opened 2023-07-12
rules_go + gazelle: upgrade to v0.41.0 and add googleapis

sluongng:
In the latest version of rules_go and gazelle,
the implicit googleapis dependency has been removed in favor of explicit
dependency. See the release notes for more details:

https://github.com/bazelbuild/rules_go/releases/tag/v0.41.0

**Related issues**: N/A


tempoz (approved):
LGTM; is there anything people might need to know about the large block of `# gazelle:resolve` statements in `BUILD`? Is that list expected to be exhaustive of problematic `googleapis` protos, or might we need to add more similar statements if we depend on more google protos?

sluongng:
> is there anything people might need to know about the large block of # gazelle:resolve statements in BUILD?

Great question.

Initially I was experimenting with removing the `gazelle:proto disable` directive we have in the root BUILD file.
That would allow Gazelle to generate and edit `proto` files / targets, thus such mapping would be needed.

However, with the way we are organizing proto files, we have multiple `proto package` within the same `bazel package` (directory). Hence, it's a use case the Gazelle's proto language extension does not support. So removing `gazelle:proto disable` did not affect how we manage the proto BUILD targets. Thus, I have reinstated the `gazelle:proto disable` directive and removed the block of `gazelle:resolve`.

We would still need the resolve directives when using Gazelle within `go_repository` though, see `go-gax` as one example of how to do that.

PR #4300, opened 2023-07-12
Limit the max number of build events that can be buffered in memory

bduffany:
This is a safe guard to protect the apps until https://github.com/buildbuddy-io/buildbuddy-internal/issues/749 is fixed.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/749


tylerwilliams (commented):
I have a fix for this I'll send when I'm online

bduffany:
> I have a fix for this I'll send when I'm online

sgtm - closing

PR #4302, opened 2023-07-12
Remove sorting in pebble_cache.FindMissing

luluz66:
Don't sort the input resource names in FindMissing. This has the side effect of 
changing the order of the input slice; and benchmark tests didn't show that it
impact performances after removing the sort.
**Related issues**: N/A


vadimberezniker (approved):
I vaguely recall Tyler saw some perf differences with sorting in place so it's worth waiting for him to chime in, but if you're not seeing anything on the benchmarks then removing this sounds fine to me.

tylerwilliams (approved):

PR #4309, opened 2023-07-13
app/executors: improve OS/Arch UI

sluongng:
When helping a customer troubleshoot their executor pools,
it was not clear visually which pool was which between multiple
"Default" pools with different OS/Arch combination.

Move the OS/Arch information to their dedicated lines to better display
different attributes of a pool.

**Related issues**: N/A

### DIFF
--- a/enterprise/app/executors/BUILD
+++ b/enterprise/app/executors/BUILD
@@ -20,6 +20,7 @@ ts_library(
         "//proto:bazel_config_ts_proto",
         "//proto:scheduler_ts_proto",
         "@npm//@types/react",
+        "@npm//lucide-react",
         "@npm//react",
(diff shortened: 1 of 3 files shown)


sluongng:
image

Quick preview of how it would look like with this PR

siggisim:
This feels like it takes up a lot of vertical space, maybe we could add them as little chips after the pool name?

bduffany:
Pushed a commit to fix some layout/alignment issues - 
image

(Feel free to remove the commit, just figured it was easier than making a bunch of PR comments)

sluongng:
That looks so much better wow. Nicely done!

bduffany (approved):

PR #4312, opened 2023-07-13
Populate client IP in audit log entries.

vadimberezniker:
**Related issues**: N/A


server/util/current_request/current_request.go
```diff
@@ -0,0 +1,12 @@
+package current_request
```
tylerwilliams, line 1: nit: maybe call this client_ip and change `func ClientIP` to `func Get`

tylerwilliams: (or move other current_request stuff here if you think that's a better fit)

vadimberezniker: renamed to clientip.

tylerwilliams (approved):

PR #4314, opened 2023-07-13
Disable streaming private images

iain-macdonald:
This PR:
- Adds a new flag, `--executor.podman.enable_private_image_streaming` which controls whether images requiring authentication are streamed or not. Note that this flag only takes effect if `--executor.podman.enable_image_streaming` (which now controls streaming public images) is also true.
- Removes the unused `--executor.podman.streamable_images` flag because it was making the code more confusing than necessary.
- Refactors the image-streaming flow control a little bit so that checking state via flags and booleans is more consistent. Now, per-image streamability is controlled in the Container (which is narrowly scoped to a single podman operation) by just a boolean that is provided to the Container and has a different name than the flag (for easier searching). And image streaming is controlled in the Provider (which is more broadly scoped to the server) using the command-line flags. The provider translates the value of the flags into per-image streaming in `NewContainer`. Also, checking the nil-ness of sociArtifactStore client is removed because that is one-to-one with the flag *imageStreamingEnabled.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -115,6 +115,7 @@ type Provider struct {
 	env                     environment.Env
 	imageCacheAuth          *container.ImageCacheAuthenticator
 	buildRoot               string
+	imageStreamingEnabled   bool
 	sociArtifactStoreClient socipb.SociArtifactStoreClient
 	sociStoreKeychainClient sspb.LocalKeychainClient
 }
@@ -190,6 +191,7 @@ additionallayerstores=["/var/lib/soci-store/store:ref"]
 	return &Provider{
 		env:                     env,
 		imageCacheAuth:          imageCacheAuthenticator,
+		imageStreamingEnabled:   *imageStreamingEnabled,
 		sociArtifactStoreClient: sociArtifactStoreClient,
 		sociStoreKeychainClient: sociStoreKeychainClient,
 		buildRoot:               buildRoot,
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -223,27 +218,23 @@ func initializeSociStoreKeychainClient(env environment.Env, target string) (sspb
...
 
-func (p *Provider) NewContainer(ctx context.Context, image string, options *PodmanOptions) (container.CommandContainer, error) {
-	imageIsStreamable := false
-	for _, streamableImage := range p.streamableImages {
-		if image == streamableImage {
-			imageIsStreamable = true
-			break
-		}
+func (p *Provider) NewContainer(ctx context.Context, image string, imageIsPublic bool, options *PodmanOptions) (container.CommandContainer, error) {
+	var imageIsStreamable bool
```
vadimberezniker, line 222: You can simplify this to
```
imageIsStreamable := *imageStreamingEnabled && (imageIsPublic || *privateImageStreamingEnabled)
```

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -223,27 +218,23 @@ func initializeSociStoreKeychainClient(env environment.Env, target string) (sspb
...
+	if imageIsPublic {
+		imageIsStreamable = *imageStreamingEnabled
```
vadimberezniker, line 224: nit: instead of directly depending on the flag value in 2 places, you can set a Provider property (imageStreamingEnabled) in NewProvider and reference that in NewContainer.

vadimberezniker (approved):

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Done

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Done

PR #4318, opened 2023-07-14
bytebufferpool: store pointer to slice to reduce allocation

sluongng:
sync.Pool should be storing pointers instead of struct.
A Go slice is a struct underneath (with one of the fields being the
pointer to an array).

Swap out sync.Pool usage in bytebufferpool with a `safeSyncPool` struct
that provides type-safe access to pool's values.
Change bytebufferpool.Pool to use this generic wrapper with type being
pointer to byte slice.

**Related issues**: N/A


server/util/bytebufferpool/bytebufferpool.go
```diff
@@ -5,23 +5,48 @@ import (
 	"sync"
 )
 
+// safeSyncPool is a wrapper around `sync.Pool` that provides type safety
+// using Go generics.
+type safeSyncPool[T any] struct {
```
vadimberezniker, line 10: This feels a little overkill for internal usage, but it's fine.
We could move this somewhere else in the repo and use it everywhere we use sync.Pool, then it would feel less wasteful. Of course we don't need to do that in this PR.

vadimberezniker (approved):
In practice the slice allocations here are not going to make any kind of dent in overall performance, but I see how it's worth making the change in order to be able to enable this vet check.

iain-macdonald (approved):

PR #4323, opened 2023-07-14
Adds gopackagesdriver

tempoz:
This adds a `gopackagesdriver` that can be specified to the `gopls` instance that an IDE is connected to to prevent the need for symlinking go protos to get auto-completion and go-to-definition working for protos. This even works if the proto files are not currently in the cache (though it takes a bit).

I had to restrict the sandbox test to mac via the `go:build` directive instead of the `target_compatible_with` bazel property to prevent it failing to build when the `gopackagesdriver` ran, blocking it from use.

**Related issues**: N/A


tempoz:
* master
  * **PR #4323** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4323" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4323?utm_source=stack-comment).

enterprise/server/remote_execution/containers/sandbox/sandbox_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build darwin
```
sluongng, line 1:
```suggestion
//go:build darwin && !ios
```

This should help you simplify the generated BUILD file for this package.
We are never building our binary targeting iOS anyway 🤔

tools/gopackagesdriver
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+SCRIPT_DIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
+cd $SCRIPT_DIR/..
```
sluongng, line 3: Nit: this is a bit different from the default recommendation https://github.com/bazelbuild/rules_go/blob/master/tools/gopackagesdriver.sh

I think it's because you are planning to use the packages driver within another script, but it's unclear in the context of this PR why these are needed and how they would be used.

sluongng (approved):

enterprise/server/remote_execution/containers/sandbox/sandbox_test.go
sluongng: you might need to re-run Gazelle locally after applying the suggestion.

enterprise/server/remote_execution/containers/sandbox/BUILD
```diff
@@ -24,12 +24,21 @@ go_test(
...
+            "@com_github_stretchr_testify//assert",
+        ],
+        "@io_bazel_rules_go//go/platform:ios": [
+            ":sandbox",
+            "//enterprise/server/remote_execution/container",
+            "//proto:remote_execution_go_proto",
+            "//server/testutil/testfs",
+            "@com_github_stretchr_testify//assert",
+        ],
+        "//conditions:default": [],
```
sluongng, line 42: Review note (to my future self): When the test is built for Linux, rules_go will inject an `empty.go` file into both internal and external test binary, effectively creating 2 no-ops tests on non-darwin platforms.

Bazel would still display that the test was executed on Linux, but it does not run any actual Go tests.

tools/gopackagesdriver
tempoz:
```suggestion
# go-to-definition can take us to files outside the bazel workspace where bazel
# cannot be run, so cd back to the workspace root.
SCRIPT_DIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
cd $SCRIPT_DIR/..
```

PR #4325, opened 2023-07-17
filecache: log potential unlink error

sluongng:
We have received a report from a customer with self-hosted executor
regarding cleanup left over.

Let's start tracking if there is any failures during unlink syscall.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/filecache/filecache.go
+++ b/enterprise/server/remote_execution/filecache/filecache.go
@@ -93,12 +93,13 @@ func evictFn(value interface{}, reason lru.EvictionReason) {
 	// TODO(sluongng): consider using Go generic to avoid type assertion.
 	v, ok := value.(*entry)
 	if !ok {
-		log.Warningf("could not cast filecache's value to entry struct: %v", value)
+		log.Errorf("Unexpected filecache entry type %T", value)
 		return
 	}
 
-	err := syscall.Unlink(v.value)
-	log.Errorf("failed to unlink filecache entry %q: %s", v.value, err)
+  if err := syscall.Unlink(v.value); err != nil {
+	  log.Errorf("Failed to unlink filecache entry %q: %s", v.value, err)
+  }
 	if reason == lru.SizeEviction {
 		age := time.Since(time.UnixMicro(v.addedAtUsec)).Microseconds()
 		metrics.FileCacheLastEvictionAgeUsec.Set(float64(age))
--- a/enterprise/server/remote_execution/filecache/filecache.go
+++ b/enterprise/server/remote_execution/filecache/filecache.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/metrics"
+	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
 	"github.com/buildbuddy-io/buildbuddy/server/util/disk"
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -90,12 +90,18 @@ func sizeFn(value interface{}) int64 {
...
-	if v, ok := value.(*entry); ok {
-		syscall.Unlink(v.value)
-		if reason == lru.SizeEviction {
-			age := time.Since(time.UnixMicro(v.addedAtUsec)).Microseconds()
-			metrics.FileCacheLastEvictionAgeUsec.Set(float64(age))
-		}
+	// TODO(sluongng): consider using Go generic to avoid type assertion.
+	v, ok := value.(*entry)
+	if !ok {
+		log.Warningf("could not cast filecache's value to entry struct: %v", value)
```
bduffany, line 96:
```suggestion
		log.Errorf("Unexpected filecache entry type %T", value)
```

enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -90,12 +90,18 @@ func sizeFn(value interface{}) int64 {
...
+		return
+	}
+
+	err := syscall.Unlink(v.value)
+	log.Errorf("failed to unlink filecache entry %q: %s", v.value, err)
```
bduffany, line 101: (missing `err != nil` check)

```suggestion
  if err := syscall.Unlink(v.value); err != nil {
	  log.Errorf("Failed to unlink filecache entry %q: %s", v.value, err)
  }
```

enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -90,12 +90,18 @@ func sizeFn(value interface{}) int64 {
 }
 
 func evictFn(value interface{}, reason lru.EvictionReason) {
-	if v, ok := value.(*entry); ok {
-		syscall.Unlink(v.value)
-		if reason == lru.SizeEviction {
-			age := time.Since(time.UnixMicro(v.addedAtUsec)).Microseconds()
-			metrics.FileCacheLastEvictionAgeUsec.Set(float64(age))
-		}
+	// TODO(sluongng): consider using Go generic to avoid type assertion.
```
bduffany, line 93: I'm not sure this is the right place for this TODO - LRU is used pretty extensively in the codebase. I would omit this for now.

sluongng: Agree

enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -90,12 +90,18 @@ func sizeFn(value interface{}) int64 {
...
-	if v, ok := value.(*entry); ok {
-		syscall.Unlink(v.value)
-		if reason == lru.SizeEviction {
-			age := time.Since(time.UnixMicro(v.addedAtUsec)).Microseconds()
-			metrics.FileCacheLastEvictionAgeUsec.Set(float64(age))
-		}
+	v, ok := value.(*entry)
+	if !ok {
+		log.Errorf("Unexpected filecache entry type %T", value)
+		return
```
tylerwilliams, line 96: you can use alert.UnexpectedEvent for stuff like this

siggisim:
We can add this if you'd like, but I'm pretty sure the real issue here is https://github.com/buildbuddy-io/buildbuddy-internal/issues/2142

bduffany (approved):

PR #4326, opened 2023-07-17
runner: allow setting podID using environment variable

sluongng:
When cgroup v2 or other container orchestration is used, it's possible
that podID is not set in `/proc/*/cpuset` and become empty.

Let's provide a way for user to explicitly set podID from K8s container
definition via environment variable.

For example, one could use the following config

```yaml

  containers:
    - name: buildbuddy-executor
      ...
      env:
      - name: K8S_POD_ID
        valueFrom:
          fieldRef:
            fieldPath: metadata.uid
```

to set this value.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -140,6 +140,9 @@ var (
 )
 
 func k8sPodID() (string, error) {
+	if podID := os.Getenv("K8S_POD_ID"); podID != "" {
+		return podID, nil
+	}
 	if _, err := os.Stat("/proc/1/cpuset"); err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -140,6 +140,9 @@ var (
 )
 
 func k8sPodID() (string, error) {
+	if podID := os.Getenv("K8S_POD_ID"); podID != "" {
```
bduffany, line 143: could you put this in `server/resources/resources.go` and add a getter like `GetK8sPodID()` just so all the env vars are in one place (MY_POOL, MY_NODENAME, MY_HOSTNAME etc.)

bduffany (approved):

PR #4329, opened 2023-07-17
usage: remove reliance on unique index for upsert query

bduffany:
We're planning to update the Usage table schema with new Label columns (e.g. `origin={internal, external}, client={bazel, executor}`), and we might also add more labels in the future (such as `owner={bb, self_hosted}`).

These new labels effectively need to be part of the current unique index on (group_id, region, period_start_usec).

But there are a few problems with just adding these fields to the unique index:
- GORM can't migrate unique indexes for us, and doing a manual migration each time we want to add a new label is error prone. The consequence of getting this migration wrong would be potentially double-counting usage data.
- Even if GORM could migrate these unique indexes for us, our current upsert logic would be error prone. If the Create() fails, that means there is already a usage row with the given primary key (group_id, region, period_start_usec, labels[0], ..., labels[N]). Therefore, we fall back to an UPDATE query, in which the current app may not yet know about the new label columns (while we are doing an app rollout). So the app could potentially apply an update to multiple rows, since the UPDATE will apply to all rows matching the labels that the app does know about. This again results in overcounting.
- Unique indexes are limited to 3072 bytes in size, so this approach may not be scalable to more label columns. (This is a minor point though since we might not ever hit this limit.)

To work around these problems with unique indexes, this PR makes the following changes:
- Stop depending on the unique index for the upsert logic. Instead, within a transaction (on the application side), manually query for existing rows matching the key, acquiring a lock on those rows using SELECT ... FOR UPDATE. Then, do the update only if both of the following are true: (a) the SELECT query matched only a single row, and (b) the row that was matched does not have unrecognized columns with values set (e.g. a new app started writing a new label column that we don't yet know about). Generally speaking, we should never run into this "unrecognized fields" case as long as we roll out new label columns properly - release 1 should start reading the new field, then release 2 should start writing the new field. But this new approach protects us in case we aren't perfect about rolling out the labels, or have to roll back the apps after we have already written new usage rows.
- Since we no longer depend on the unique index, remove the unique index from the schema and instead add a non-unique index so that queries can still be fast. Note that GORM will not delete the unique index for us. Instead, we have to manually delete it, which I plan to do after the next release.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/usage/usage.go
+++ b/enterprise/server/usage/usage.go
@@ -11,6 +11,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
 	"github.com/buildbuddy-io/buildbuddy/server/util/authutil"
 	"github.com/buildbuddy-io/buildbuddy/server/util/db"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
@@ -18,7 +19,6 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/timeutil"
 	"github.com/go-redis/redis/v8"
-	"gorm.io/gorm/clause"
 
 	usage_config "github.com/buildbuddy-io/buildbuddy/enterprise/server/usage/config"
 )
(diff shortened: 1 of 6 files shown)


enterprise/server/usage/usage.go
```diff
@@ -273,28 +273,80 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, c collection
...
+		if err != nil {
+			return status.WrapError(err, "schema")
+		}
+
+		existingRowCount := 0
+		for rows.Next() {
+			existingRowCount++
+			fields := map[string]any{}
+			if err := tx.ScanRows(rows, &fields); err != nil {
+				return status.WrapError(err, "scan row")
```
tylerwilliams, line 304: I don't think the WrapErrors are adding much value, can we remove these?

enterprise/server/usage/usage.go
```diff
@@ -273,28 +273,80 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, c collection
...
+					continue
+				}
+				if _, ok := schema.FieldsByDBName[f]; !ok {
+					unsupportedField = f
+					unsupportedFieldValue = v
+					break
+				}
+			}
+			if unsupportedField != "" {
+				log.Warningf("Existing usage row contains unsupported column %q with nonempty value %q; dropping usage data.", unsupportedField, unsupportedFieldValue)
```
tylerwilliams, line 320: could we use alert.UnexpectedEvent here too?

tylerwilliams: alert.UnexpectedEvent("usage_update_dropped",

tylerwilliams (changes requested):

vadimberezniker (approved):

enterprise/server/usage/usage.go
bduffany: I'll remove in this case since DB errors are usually more obvious, but I've sort of started to form a habit of adding these because I've gotten bitten lately by error messages in the logs being hard to decipher, e.g. https://github.com/buildbuddy-io/buildbuddy/pull/4342

When this happens it's a bit unfortunate because it adds another release cycle before we can be sure about where an error is coming from.

enterprise/server/usage/usage.go
bduffany: good catch - done

tylerwilliams (approved):

PR #4330, opened 2023-07-17
executor: allow using overlayfs for action workspaces

bduffany:
Adds a new flag `--executor.workspace.overlayfs_enabled` which enables overlayfs for action workspaces and another flag `--executor.workspace.overlayfs_anonymous_only` to enable it only for anonymous users (which may be preferable when already using group-level filecache isolation for authenticated users). This provides an alternative to reflinking to get copy-on-write for action inputs.

The behavior of this flag is:
- If overlayfs is enabled for the current task, then before running any actions in the workspace, "convert" the workspace to an overlayfs mount.
- "Convert" does the following: 
  - Rename the existing workspace path to `{path}.lower` - inputs will go in this dir and cannot be overwritten by actions
  - Create `{path}.work` (for overlayfs temp data) and `{path}.upper` (for changes)
  - Mount the overlayfs to the pre-existing `{path}`
- This "conversion" approach allows the changes to the existing `workspace` package to be minimal. All we have to do is hardlink inputs to the `{path}.lower` dir. Everywhere else (e.g. action working directory, uploading action outputs), we can still use `workspace.Path()` to get the action working directory, which points to the overlayfs mount dir and is where the action should be reading and writing files.
- After running an action, we perform an "apply" operation which applies the changes from the upper dir to the lower dir, without actually modifying any of the files in the lower dir. (This can also be thought of as "flattening" the layers.) This is done for 2 reasons:
  - Since we use `dirHelper` for downloading inputs and uploading outputs, it simplifies the code so that `dirHelper` can always just be working from `lowerdir` instead of downloading inputs from lowerdir and uploading from the mount dir.
  - It ensures that when we hardlink new inputs into the lowerdir for the next action (in the case of runner recycling), the changes are guaranteed to be visible (i.e. not "masked" by something in the upperdir).
- Note: It might seem a bit "unsafe" to be messing with files in the upper and lower dirs, but this sort of thing is not discouraged in the manual pages for overlayfs. The specification for overlayfs is clear e.g. about what file deletion markers look like, so as long as we're following the specification, we should be OK.

Benchmark results (from GCP VM matching current executor configuration)

```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 3.10GHz
BenchmarkOverlayfs/OverlayEnabled_ReadFile_1B-16                  132554              7824 ns/op
BenchmarkOverlayfs/OverlayDisabled_ReadFile_1B-16                 177919              6671 ns/op
BenchmarkOverlayfs/OverlayEnabled_ReadFile_1KiB-16                149977              7982 ns/op
BenchmarkOverlayfs/OverlayDisabled_ReadFile_1KiB-16               169365              6887 ns/op
BenchmarkOverlayfs/OverlayEnabled_ReadFile_4KiB-16                133058              8929 ns/op
BenchmarkOverlayfs/OverlayDisabled_ReadFile_4KiB-16               150868              7860 ns/op
BenchmarkOverlayfs/OverlayEnabled_ReadFile_8KiB-16                117721             10076 ns/op
BenchmarkOverlayfs/OverlayDisabled_ReadFile_8KiB-16               131685              9157 ns/op
BenchmarkOverlayfs/OverlayEnabled_ReadFile_50KiB-16                50624             23179 ns/op
BenchmarkOverlayfs/OverlayDisabled_ReadFile_50KiB-16               53535             22065 ns/op
BenchmarkOverlayfs/OverlayEnabled_ReadFile_100KiB-16               34107             35225 ns/op
BenchmarkOverlayfs/OverlayDisabled_ReadFile_100KiB-16              35740             33563 ns/op
BenchmarkOverlayfs/OverlayEnabled_WriteFile_1B-16                  38896             29462 ns/op
BenchmarkOverlayfs/OverlayDisabled_WriteFile_1B-16                 53720             21583 ns/op
BenchmarkOverlayfs/OverlayEnabled_WriteFile_1KiB-16                39532             29479 ns/op
BenchmarkOverlayfs/OverlayDisabled_WriteFile_1KiB-16               52353             21704 ns/op
BenchmarkOverlayfs/OverlayEnabled_WriteFile_4KiB-16                39133             29495 ns/op
BenchmarkOverlayfs/OverlayDisabled_WriteFile_4KiB-16               53686             21603 ns/op
BenchmarkOverlayfs/OverlayEnabled_WriteFile_8KiB-16                36180             32434 ns/op
BenchmarkOverlayfs/OverlayDisabled_WriteFile_8KiB-16               46580             24317 ns/op
BenchmarkOverlayfs/OverlayEnabled_WriteFile_50KiB-16               20928             57340 ns/op
BenchmarkOverlayfs/OverlayDisabled_WriteFile_50KiB-16              24051             49096 ns/op
BenchmarkOverlayfs/OverlayEnabled_WriteFile_100KiB-16              14228             84356 ns/op
BenchmarkOverlayfs/OverlayDisabled_WriteFile_100KiB-16             15501             74746 ns/op
```

Results from a simple build (overlayfs is about 20% slower):

```
build //server/util/...
1 local executor
---
anonymous user, overlayfs enabled
warmup build: done
build 1: 104.5s
build 2: 106.3s
---
authenticated user, overlayfs disabled
warmup build: done
build 1: 84.5s
build 2: 87.1s
```

I checked to see if there are any knobs available that would help improve performance - I found the [`volatile`](https://www.redhat.com/sysadmin/container-volatile-overlay-mounts) mount option but it didn't measurably affect performance (it seems to be more applicable to some niche workloads which do a lot of `fsync` operations - the article mentions "Certain tools like RPM call for a sync after every file is written to disk").

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2441


enterprise/server/remote_execution/overlayfs/overlayfs_linux.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	for _, e := range entries {
+		if err := os.RemoveAll(filepath.Join(o.UpperDir, e.Name())); err != nil {
+			return status.WrapError(err, "remove upper dir entry")
+		}
+	}
+	return nil
+}
+
+func forceDir(path string, perm fs.FileMode) error {
```
tylerwilliams, line 179: maybe could call this forceMkdir?

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -69,6 +69,9 @@ var (
 	// can't be added to the pool and must be cleaned up instead.
 	maxRunnerMemoryUsageBytes = flag.Int64("executor.runner_pool.max_runner_memory_usage_bytes", 0, "Maximum memory usage for a recycled runner; runners exceeding this threshold are not recycled.")
 	podmanWarmupDefaultImages = flag.Bool("executor.podman.warmup_default_images", true, "Whether to warmup the default podman images or not.")
+
+	overlayfsEnabled  = flag.Bool("executor.workspace.overlayfs_enabled", false, "Enable overlayfs for executor workspaces. This provides a stronger guarantee that the executor's local cache cannot be modified by actions. Only available on Linux, and requires the executor to have mount() permissions as well as the overlayfs kernel module to be enabled.")
+	overlayfsAnonOnly = flag.Bool("executor.workspace.overlayfs_anonymous_only", false, "Restrict usage of overlayfs to unauthenticated requests only.")
```
tylerwilliams, line 74: I'm not sure I understand why we need this flag? It seems like we want:

- a flag to flip this on/off for anon
- a platform property so anyone can turn this on (if it's enabled)

enterprise/server/remote_execution/workspace/workspace.go
```diff
@@ -270,8 +298,23 @@ func (ws *Workspace) UploadOutputs(ctx context.Context, cmd *repb.Command, execu
...
+		if ws.overlay != nil {
+			// When overlayfs is enabled, apply the changes from upperdir to
+			// lowerdir, since dirHelper's root dir is configured as the
+			// lowerdir and it needs to see the output files.
+			//
+			// Optimization: if recycling is not enabled, then at this point the
+			// runner should be removed and cannot affect any files in the
+			// workspace anymore, so it is safe to rename the outputs files in
+			// upperdir here rather than copying.
+			recyclingEnabled := platform.IsTrue(platform.FindValue(ws.task.GetCommand().GetPlatform(), platform.RecycleRunnerPropertyName))
```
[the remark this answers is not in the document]
bduffany: Without this optimization, we have to do full copies of all output files in the action instead of being able to hardlink like we do today. I think the IO savings will be worth the small amount of extra complexity.

The comment here frames it as an optimization but when I was originally writing the code I thought of it more as handling an edge case - hardlinking is our preferred/simplest way of copying files, but when runner recycling is enabled we have to fall back to a clunky/inefficient full-copy for safety reasons.

enterprise/server/remote_execution/runner/runner.go
tylerwilliams: can we do this in a followup? I think we should test this first, on our own stuff, on anon only, to begin with. (I'm fine if you want to add this functionality later in a followup)

enterprise/server/remote_execution/workspace/workspace.go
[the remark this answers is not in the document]
tylerwilliams: ok.

enterprise/server/remote_execution/runner/runner.go
bduffany: sg - kept the anon-only flag and removed the executor-wide one.

tylerwilliams (approved):

PR #4337, opened 2023-07-17
Add cdc chunking support to PebbleCache.

luluz66:
When chunking is enabled, during write, we will first attempt to
chunk it, and then compress each individual chunk, and finally encrypt the chunk
if encryption is enabled.  If the data is compressed, we will first decompress it
before we chunk it.

If there is only one single chunk, we will write the chunk inline,
similar to how we write small files inline when chunking is disabled.
If there are multiple chunks, we will write a file record in Pebble. which include
all the resource names that points to the chunks. Every chunks will be
stored inline in Pebble.

Added unit tests.

This is based on Tyler's PR:
https://github.com/buildbuddy-io/buildbuddy/pull/3457

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1860,10 +1860,11 @@ func (p *PebbleCache) newCDCCommitedWriteCloser(ctx context.Context, fileRecord 
 			return err
 		}
 		if cdcw.numChunks == 1 {
-			// When there is only one single chunk, we want to store the original file
-			// record with the original key instead of computed digest from the
-			// chunkData. This is because the chunkData can be compressed or encrypted,
-			// so the digest computed from it will be different from the original digest.
+			// When there is only one single chunk, we want to store the original
+			// file record with the original key instead of computed digest from
+			// the chunkData. This is because the chunkData can be compressed or
+			// encrypted, so the digest computed from it will be different from
+			// the original digest.
 			return cdcw.writeRawChunk(cdcw.fileRecord, cdcw.key, cdcw.firstChunk)
 		}
 		now := p.clock.Now().UnixMicro()
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1933,22 +1933,18 @@ func (cdcw *cdcWriter) writeChunkWhenMultiple(chunkData []byte) error {
 	if err != nil {
 		return err
 	}
-	r := &rspb.ResourceName{
-		Digest:         d,
-		InstanceName:   cdcw.fileRecord.GetIsolation().GetRemoteInstanceName(),
-		CacheType:      cdcw.fileRecord.GetIsolation().GetCacheType(),
-		DigestFunction: cdcw.fileRecord.GetDigestFunction(),
-		Compressor:     cdcw.fileRecord.GetCompressor(),
-	}
-	if cdcw.shouldCompress && r.Compressor == repb.Compressor_IDENTITY {
+
+	r := digest.NewResourceName(d, cdcw.fileRecord.GetIsolation().GetRemoteInstanceName(), cdcw.fileRecord.GetIsolation().GetCacheType(), cdcw.fileRecord.GetDigestFunction())
+	if cdcw.shouldCompress && cdcw.fileRecord.GetCompressor() == repb.Compressor_IDENTITY {
 		// we need to compress the chunk, but this data hasn't been compressed yet.
 		// so we need to set the resource name to identity to signal to the nested
 		// writer to compress it.
-		r.Compressor = repb.Compressor_IDENTITY
+		r.SetCompressor(repb.Compressor_IDENTITY)
 	} else {
-		r.Compressor = repb.Compressor_ZSTD
+		r.SetCompressor(repb.Compressor_ZSTD)
 	}
-	fileRecord, err := p.makeFileRecord(ctx, r)
+	rn := r.ToProto()
+	fileRecord, err := p.makeFileRecord(ctx, rn)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1793,6 +1804,196 @@ func (p *PebbleCache) Reader(ctx context.Context, r *rspb.ResourceName, uncompre
...
+	fileRecord, err := p.makeFileRecord(ctx, r)
+
+	if err != nil {
+		return err
+	}
+	key, err := p.fileStorer.PebbleKey(fileRecord)
+	if err != nil {
+		return err
+	}
+	cdcw.writtenChunks = append(cdcw.writtenChunks, r)
```
tylerwilliams, line 1961: maybe worth a comment why this happens first :)

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1793,6 +1804,196 @@ func (p *PebbleCache) Reader(ctx context.Context, r *rspb.ResourceName, uncompre
...
+	cdcw.eg.Go(func() error {
+		return cdcw.writeRawChunk(fileRecord, key, chunkData)
+	})
+	return nil
+}
+
+func (cdcw *cdcWriter) Write(buf []byte) (int, error) {
+	// If the bytes being written are compressed, we decompress them in
+	// order to generate CDC chunks, then compress those chunks.
+	if cdcw.isCompressed {
```
tylerwilliams, line 1972: i wonder if we'll want to do this if the file is gonna be less than 1 chunk size? maybe some way to skip this code path?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2997,6 +3208,39 @@ type readCloser struct {
 	io.Closer
 }
 
+// newChunkedReader returns a reader to read chunked content.
+// When shouldDecompress is true, the content readed is decompressed.
```
tylerwilliams, line 3212:
```suggestion
// When shouldDecompress is true, the content read is decompressed.
```

tylerwilliams (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1793,6 +1804,199 @@ func (p *PebbleCache) Reader(ctx context.Context, r *rspb.ResourceName, uncompre
...
+		// file-level metadata entry and a chunk entry into pebble.
+		cdcw.firstChunk = make([]byte, len(chunkData))
+		copy(cdcw.firstChunk, chunkData)
+		return nil
+	}
+
+	if cdcw.numChunks == 2 {
+		if err := cdcw.writeChunkWhenMultiple(cdcw.firstChunk); err != nil {
+			return err
+		}
```
vadimberezniker, line 1898: Should we clear out `cdcw.firstChunk` here? I believe the data is no longer needed after this.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1793,6 +1804,199 @@ func (p *PebbleCache) Reader(ctx context.Context, r *rspb.ResourceName, uncompre
...
+	if err != nil {
+		return err
+	}
+	key, err := p.fileStorer.PebbleKey(fileRecord)
+	if err != nil {
+		return err
+	}
+
+	// We use cdcw.writtenChunks for the file-level metadata, and this needs to
+	// be in the orther. Otherwise, when we read the file, the chunks will be
```
vadimberezniker, line 1958: I think you meant to write "in order" here.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2744,6 +2959,8 @@ func (e *partitionEvictor) deleteFile(key filestore.PebbleKey, version filestore
 		}
 	case storageMetadata.GetInlineMetadata() != nil:
 		break
+	case storageMetadata.GetChunkedMetadata() != nil:
+		break
```
vadimberezniker, line 2963: Do we need to skip updating the partition size in this case?

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: Done

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: I set the storedSizeBytes in the filemetadata is zero if it's a file level entry that include chunked metadata

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1793,6 +1805,200 @@ func (p *PebbleCache) Reader(ctx context.Context, r *rspb.ResourceName, uncompre
...
+			return cdcw.writeRawChunk(cdcw.fileRecord, cdcw.key, cdcw.firstChunk)
+		}
+		now := p.clock.Now().UnixMicro()
+
+		md := &rfpb.FileMetadata{
+			FileRecord:      fileRecord,
+			StorageMetadata: cdcw.Metadata(),
+			// The chunks the file record pointed are stored seperately and are
+			// not evicted when this entry is evicted. Therefore, the stored
+			// size bytes should be zero to avoid double counting.
```
tylerwilliams, line 1876: these too

tylerwilliams (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1793,6 +1805,200 @@ func (p *PebbleCache) Reader(ctx context.Context, r *rspb.ResourceName, uncompre
...
+			return err
+		}
+		if err := cdcw.eg.Wait(); err != nil {
+			return err
+		}
+		if cdcw.numChunks == 1 {
+			// When there is only one single chunk, we want to store the original file
+			// record with the original key instead of computed digest from the
+			// chunkData. This is because the chunkData can be compressed or encrypted,
+			// so the digest computed from it will be different from the original digest.
```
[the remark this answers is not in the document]

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: Done

vadimberezniker (approved):

tylerwilliams (approved):

PR #4341, opened 2023-07-18
firecracker_test: switch to use mirror.gcr.io

sluongng:
It seems like we are experiencing serious flake because our test would
pull image from marketplace.gcr.io which requires credentials.

Side-step the credential issue completely by pulling from GCP's mirror
of Docker's public container registry "docker.io".

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
@@ -45,11 +45,11 @@ import (
 )
 
 const (
-	busyboxImage = "docker.io/library/busybox:latest"
+	busyboxImage = "mirror.gcr.io/library/busybox:latest"
 	// Alternate image to use if getting rate-limited by docker hub
 	// busyboxImage = "gcr.io/google-containers/busybox:latest"
 
-	ubuntuImage              = "marketplace.gcr.io/google/ubuntu2004"
+	ubuntuImage              = "mirror.gcr.io/library/ubuntu:20.04"
 	imageWithDockerInstalled = "gcr.io/flame-public/executor-docker-default:enterprise-v1.6.0"
 
 	// Minimum memory needed for a firecracker VM. This may need to be increased


dockerfiles/executor_image/data/registries.conf
```diff
@@ -7,15 +7,3 @@ prefix = "docker.io"
...
-# hand-copied those to quay, using skopeo copy --all ...
-[[registry]]
-prefix="docker.io/library"
-location="quay.io/libpod"
-
-# 2021-03-23 these are used in buildah system tests, but not (yet?) listed in the global shortnames.conf.
-[aliases]
-busybox="docker.io/library/busybox"
-ubuntu="docker.io/library/ubuntu"
-php="docker.io/library/php"
```
bduffany, line 21: cc @luluz66 any idea if this change would break any existing use cases?

sluongng: I think this was taken as-is from an older version of https://github.com/containers/podman/blob/main/test/registries.conf

I think the removed part is only relevant to RedHat / Podman-specific setup.

luluz66: This looked fine to me; but probably run a few builds in dev to make sure we didn't break anything.

luluz66 (approved):

dockerfiles/executor_image/data/registries.conf
bduffany: Can you revert this specific diff since it's unrelated to the PR? We can revisit later if it's needed, but would prefer to avoid unnecessary churn here.

tylerwilliams: +1 -- this looks fine to move to another CL

sluongng: Reverted that chunk. Will move to another PR

sluongng: created https://github.com/buildbuddy-io/buildbuddy/pull/4350

PR #4347, opened 2023-07-18
Secret manager tweaks:

vadimberezniker:
- Don't use "REPLACE INTO" which is mysql-specific
  - Don't include user_id in key when updating secrets since any user in the org should be able to update the same key
  - Return a boolean to indicate whether it's a new secret to facilitate proper audit logging

**Related issues**: N/A

### DIFF
--- a/enterprise/server/secrets/secrets.go
+++ b/enterprise/server/secrets/secrets.go
@@ -12,6 +12,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
 	"github.com/buildbuddy-io/buildbuddy/server/util/query_builder"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"gorm.io/gorm"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	skpb "github.com/buildbuddy-io/buildbuddy/proto/secrets"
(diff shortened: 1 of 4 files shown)


enterprise/server/secrets/secrets.go
```diff
@@ -150,15 +151,45 @@ func (s *SecretService) UpdateSecret(ctx context.Context, req *skpb.UpdateSecret
...
 	}
-	err = dbHandle.DB(ctx).Exec(`REPLACE INTO "Secrets" (user_id, group_id, name, value, perms) VALUES (?, ?, ?, ?, ?)`,
-		u.GetUserID(), u.GetGroupID(), req.GetSecret().GetName(), req.GetSecret().GetValue(), secretPerms.Perms).Error
+
+	newSecret := false
+	err = dbHandle.Transaction(ctx, func(tx *gorm.DB) error {
+		rows, err := tx.Raw(`
+			SELECT * 
+			FROM "Secrets" 
+			WHERE group_id = ? AND name = ?
```
tylerwilliams, line 162: does this need to be SELECT FOR UPDATE to lock?

vadimberezniker: we add `dbHandle.SelectForUpdateModifier()` below

tylerwilliams: ah got it, forgot that was db specific. thanks!

tylerwilliams (approved):

PR #4351, opened 2023-07-19
api_keys: make the key hidden by default and provide a copy button

sluongng:
**Related issues**: N/A

### DIFF
--- a/enterprise/app/api_keys/api_keys.tsx
+++ b/enterprise/app/api_keys/api_keys.tsx
@@ -1,4 +1,4 @@
-import { Key } from "lucide-react";
+import { Check, Copy, Eye, EyeOff, Key } from "lucide-react";
 import React from "react";
 import { User } from "../../../app/auth/auth_service";
 import capabilities from "../../../app/capabilities/capabilities";
@@ -13,6 +13,8 @@ import Dialog, {
 import TextInput from "../../../app/components/input/input";
 import Spinner from "../../../app/components/spinner/spinner";
 import Modal from "../../../app/components/modal/modal";
+import alert_service from "../../../app/alert/alert_service";
+import { copyToClipboard } from "../../../app/util/clipboard";
 import errorService from "../../../app/errors/error_service";
 import { CancelableRpc } from "../../../app/service/rpc_service";
 import { BuildBuddyError } from "../../../app/util/errors";
@@ -461,10 +463,7 @@ export default class ApiKeysComponent extends React.Component<ApiKeysComponentPr
               <div className="api-key-capabilities">
                 <span>{describeCapabilities(key)}</span>
               </div>
-              <div className="api-key-value">
-                <Key className="icon" />
-                <span>{key.value}</span>
-              </div>
+              <ApiKeyField value={key.value} />
               {this.props.user.canCall("updateApiKey") && (
                 <OutlinedButton className="api-key-edit-button" onClick={this.onClickUpdate.bind(this, key)}>
                   Edit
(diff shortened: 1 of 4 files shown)


sluongng:
Some preview

image
image

vadimberezniker:
It would be cool if we could fetch the API key value via a separate RPC call. This would make the audit logs more useful for API keys.

sluongng:
Agree. We could later bind that rpc into these new button 😁

vadimberezniker:
> Agree. We could later bind that rpc into these new button grin

Let me know if you're interested in this or if you'd like me to add this RPC :)

enterprise/app/api_keys/api_keys.tsx
```diff
@@ -576,3 +573,73 @@ function newFormState<T extends ApiKeyFields>(request: T): FormState<T> {
...
+
+    return (
+      <div className="api-key-value">
+        <button className="api-key-value-copy" onClick={this.handleCopyClick.bind(this)}>
+          {isCopied ? <Check style={{ stroke: "green" }} className="icon" /> : <Copy className="icon" />}
+        </button>
+        <button className="api-key-value-hide" onClick={this.toggleHideValue.bind(this)}>
+          {hideValue ? <Eye className="icon" /> : <EyeOff className="icon" />}
+        </button>
+        <span>{displayValue}</span>
```
bduffany, line 641: nit: I think it is more conventional to have the value on the left and the controls on the right.

Otherwise looks great :)

enterprise/app/api_keys/api_keys.tsx
```diff
@@ -576,3 +573,73 @@ function newFormState<T extends ApiKeyFields>(request: T): FormState<T> {
...
+
+  // onClick handler function for the copy button
+  private handleCopyClick() {
+    const copyTextToClipboard = async (text: string) => {
+      if ("clipboard" in navigator) {
+        return await navigator.clipboard.writeText(text);
+      } else {
+        return document.execCommand("copy", true, text);
+      }
+    };
```
bduffany, line 610: We have an existing util, `app/util/clipboard.ts` for this

enterprise/app/api_keys/api_keys.tsx
```diff
@@ -576,3 +573,73 @@ function newFormState<T extends ApiKeyFields>(request: T): FormState<T> {
...
+
+    // Asynchronously call copyTextToClipboard
+    copyTextToClipboard(this.state.value).then(() => {
+      // If successful, update the isCopied state value
+      this.setState({ isCopied: true });
+      setTimeout(() => {
+        this.setState({ isCopied: false });
+      }, 1500);
```
bduffany, line 618: nit: rather than using state, you can also show a "success" banner, like `alert.success("Copied!")` - that would simplify the logic here since you don't need a timeout or dedicated UI state

sluongng:
> Let me know if you're interested in this or if you'd like me to add this RPC :)

@vadimberezniker it would take me a while to get to it as I am building up a backlog of tasks 🙈 
So if you manage to do it first, please do!

Take a quick peek at what this would entail, I think it will include introducing a `ListKey` RPC that would only return the keys' names and modifying the existing API to take in a key name as a filter, then eventually making the filter mandatory. 🤔

enterprise/app/api_keys/api_keys.tsx
sluongng: No strong opinion here. I will move it to the right to see how it look

enterprise/app/api_keys/api_keys.tsx
sluongng: let me try using it 🤔

enterprise/app/api_keys/api_keys.tsx
sluongng: I still prefer having a ui state here to turn the button to a green check. 
It's a visual signal to tell people to stop clicking.
Alerting would require the same state to deduplicate multiple clicks within a small time window.

Would adding an alert plus the green check button be overkill? Let's find out! 🤔

sluongng:
image

Right-hand side looks good 👍

enterprise/app/api_keys/api_keys.tsx
```diff
@@ -576,3 +575,64 @@ function newFormState<T extends ApiKeyFields>(request: T): FormState<T> {
...
+  isCopied: false,
+  hideValue: true,
+  displayValue: "************",
+};
+
+class ApiKeyField extends React.Component<ApiKeyFieldProps, ApiKeyFieldState> {
+  state = ApiKeyFieldDefaultState;
+
+  componentDidMount() {
+    this.setState({ value: this.props.value });
```
bduffany, line 601: since you're copying `this.props.value` directly to state and `this.state.value` is never modified, I think you could remove `value` from state and instead always reference `this.props.value`

bduffany (approved):

enterprise/app/api_keys/api_keys.tsx
```diff
@@ -576,3 +575,64 @@ function newFormState<T extends ApiKeyFields>(request: T): FormState<T> {
...
+  }
+
+  // onClick handler function for the copy button
+  private handleCopyClick() {
+    copyToClipboard(this.state.value);
+    this.setState({ isCopied: true }, () => {
+      alert_service.success("Copied api key to clipboard");
```
bduffany, line 608:
```suggestion
      alert_service.success("Copied API key to clipboard");
```

enterprise/app/api_keys/api_keys.tsx
sluongng: fixed

PR #4354, opened 2023-07-19
Revert "usage: remove reliance on unique index for upsert query"

tylerwilliams:
Reverts buildbuddy-io/buildbuddy#4329


vadimberezniker (approved):

tylerwilliams:
reverting this because we saw continuing unexpected_events in dev

PR #4355, opened 2023-07-19
Add a CDC Chunker.

luluz66:
Add a Chunker that implements io.WriteCloser.

When we create a Chunker, we passes two write functions to indicate how
to write the data when there is a single chunk versus when there are
multiple chunks. For example, when there is a single chunk, we want to
write it inline directly, but when there are multiple chunks, we want to
write a file level record that includes pointers to individual chunks.

Note: The fastcdc-go library is not thread safe because when
fastcdc.NewChunker is called it will change the global variable table.

So I included a patch that don't change the table when the seed is not
set (or equals zero). This should be functionally the same, since
x ^ 0 = x.

**Related issues**: N/A


enterprise/server/util/chunker/chunker.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"context"
+	"io"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/jotfs/fastcdc-go"
+)
+
+type WriteFunc func([]byte) error
```
tylerwilliams, line 11: I'm sure you considered this -- but asking so I understand. WDYT about including the chunk index in this function, and then having a single chunkWriteFunction like:

type WriteFunc func(chunkIndex int, data []byte) error

It would make this library code a little simpler to handle, at the expense of maybe more complexity in the pebble cache calling code where you'll want to distinguish these two cases and need to detect if there are multiple writes or not.

enterprise/server/util/chunker/chunker.go
```diff
@@ -0,0 +1,122 @@
...
+	pr, pw := io.Pipe()
+	c := &Chunker{
+		ctx:                  ctx,
+		pr:                   pr,
+		pw:                   pw,
+		done:                 make(chan struct{}, 0),
+		singleChunkWriteFn:   singleChunkWriteFn,
+		multipleChunkWriteFn: multipleChunkWriteFn,
+	}
+	cdcOpts := fastcdc.Options{
```
tylerwilliams, line 69: nit: could you write a comment here to explain why these are the options you picked and if they would ever change?

enterprise/server/util/chunker/chunker.go
```diff
@@ -0,0 +1,122 @@
...
+		AverageSize: averageSize,
+		MaxSize:     averageSize * 4,
+	}
+
+	chunker, err := fastcdc.NewChunker(pr, cdcOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
```
tylerwilliams, line 80: should we find a way to make this goroutine die if context is cancelled? (or are you sure it will not leak if a request dies mid-flight?)

enterprise/server/util/chunker/chunker.go
luluz66: I was using the default min and max; i don't think i need to specify them.

enterprise/server/util/chunker/chunker.go
luluz66: if pw is cancelled, then pr will return an error; and as a result this for loop will be exited.

vadimberezniker (approved):

enterprise/server/util/chunker/chunker.go
tylerwilliams: might be worth specifying these still just in case the underlying library decides to change them?

enterprise/server/util/chunker/chunker.go
tylerwilliams: sgtm

tylerwilliams (approved):
one nit, otherwise LGTM

enterprise/server/util/chunker/chunker.go
luluz66: Done.

PR #4359, opened 2023-07-19
Increase timeout and resource-measurement buffers in commandutil_test

iain-macdonald:
This test worked fine locally due to the "exclusive" tag that prevented it from competing with stuff locally, but it runs into problems running remotely when the host is under heavier load because the timeouts are kind of tight and the resource measurement isn't 100% accurate. This change makes it pass for me locally with --runs_per_test=100, but it's not perfect.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/commandutil/commandutil_test.go
+++ b/enterprise/server/remote_execution/commandutil/commandutil_test.go
@@ -210,7 +210,7 @@ func TestRun_SubprocessInOwnProcessGroup_Timeout(t *testing.T) {
 		// way to make the setpgid system call.
 		testfs.RunfilePath(t, "enterprise/server/remote_execution/commandutil/test_binary/test_binary_/test_binary"),
 	}}
-	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
 	defer cancel()
 
 	res := commandutil.Run(ctx, cmd, wd, nil /*=statsListener*/, &container.Stdio{})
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/commandutil/commandutil_test.go
```diff
@@ -253,7 +253,7 @@ func TestRun_EnableStats_RecordsCPUStats(t *testing.T) {
 	require.Equal(t, "", string(res.Stderr))
 	require.Equal(t, 0, res.ExitCode)
 	require.GreaterOrEqual(
-		t, res.UsageStats.GetCpuNanos(), int64(2e9),
+		t, res.UsageStats.GetCpuNanos(), int64(1500e6),
```
bduffany, line 256: If you wanna go even lower like 1000e6 or something, I think that'd be fine too - would rather have looser confidence in this code than have it be flaky

enterprise/server/remote_execution/commandutil/commandutil_test.go
```diff
@@ -284,21 +284,21 @@ func TestRun_EnableStats_ComplexProcessTree_RecordsStatsFromAllChildren(t *testi
 	}}
 
 	ctx := context.Background()
-	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
```
bduffany, line 288: tbh I'm not sure why I added this timeout, this LGTM but we can also just remove this and let the default test timeout do its thing.

bduffany (approved):

PR #4360, opened 2023-07-19
Fixes race condition and panic in service_test

tempoz:
Does not fix test itself, errors out with connection refused. This uncovers the error, though.

As a side note, it might be nice to have a way to gracefully shut *everything* down before changing the flags, as that was why we were encountering the race condition before.

**Related issues**: N/A


tempoz:
* master
  * **PR #4360** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4360" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4360?utm_source=stack-comment).

tylerwilliams (approved):

bduffany (changes requested):
The port ":8080" appearing in the test failure error message suggests that the app is no longer trying to run on a random port, but instead is now trying to run on the default local port (8080)

tempoz:
> The port ":8080" appearing in the test failure error message suggests that the app is no longer trying to run on a random port, but instead is now trying to run on the default local port (8080)

Yup, the test env was getting set up before the flags were being set, which meant that once I started setting the url in the workflow service only once at the beginning, it would use 8080 every time.

bduffany (approved):

PR #4362, opened 2023-07-20
Remove redsync dependency

bduffany:
Use the redlock code from https://redis.io/topics/distlock directly instead of via the redsync library, which recently caused us some problems (it made a breaking API change, changing `ErrFailed` => something else, without bumping the major version).

**Related issues**: N/A


enterprise/server/util/redisutil/redisutil.go
```diff
@@ -200,26 +209,32 @@ type redlock struct {
 // (b) It does not retry failed locking attempts.
 //
 // See https://redis.io/topics/distlock
```
tylerwilliams, line 211: is this comment still correct in light of your changes?

bduffany: yep, it's still using the same underlying implementation (lock=atomically set key to `fixedRandomValue` if not exists; unlock=atomically delete key but only if it's still equal to `fixedRandomValue`), so the limitations are still the same.

tylerwilliams (approved):
LGTM but please get @vadimberezniker's approval before submitting

PR #4374, opened 2023-07-21
Allow evicting more than 1 key before resampling.

vadimberezniker:
`evictSingleKey` is the extracted inner loop from `evict` w/o changes. 

**Related issues**: N/A


tylerwilliams (approved):
LGTM

PR #4377, opened 2023-07-22
Don't refresh entire eviction pool on every eviction

tylerwilliams:
Profiling + metrics show that eviction can be a bottleneck. A major contributor is refreshing the atime of every item in the pool on every eviction.

This CL stops doing that -- instead the approxlru will refresh an item before deleting and skip deletion if the sample's atime differs between the pool value and the live value.

### DIFF
--- a/server/util/approxlru/approxlru.go
+++ b/server/util/approxlru/approxlru.go
@@ -251,6 +251,9 @@ func (l *LRU[T]) evictSingleKey() (*Sample[T], error) {
 		}
 		if sample.Timestamp != timestamp {
 			log.Warningf("Evictor skipping %q; atime has changed %s -> %s", sample.Key, sample.Timestamp, timestamp)
+			// Update the timestamp to the new value so this sample can be
+			// removed later when the pool is trimmed.
+			sample.Timestamp = timestamp
 			continue
 		}
 
--- a/server/util/approxlru/approxlru.go
+++ b/server/util/approxlru/approxlru.go
@@ -217,22 +217,7 @@ func (l *LRU[T]) resampleK(k int) error {
 		}
 	}
 
-	filtered := make([]*Sample[T], 0, len(l.samplePool))
-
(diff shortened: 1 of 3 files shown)


server/util/approxlru/approxlru.go
```diff
@@ -250,13 +235,30 @@ func (l *LRU[T]) resampleK(k int) error {
...
 		l.mu.Lock()
 		oldLocalSizeBytes := l.localSizeBytes
 		oldGlobalSizeBytes := l.globalSizeBytes
 		l.mu.Unlock()
 
-		log.Infof("Evictor attempting to evict %q (last accessed %s)", sample.Key, time.Since(sample.Timestamp))
-		skip, err := l.onEvict(l.ctx, sample)
+		skip, timestamp, err := l.onRefresh(l.ctx, sample.Key)
+		if err != nil {
+			log.Warningf("Could not refresh timestamp for %q: %s", sample.Key, err)
```
vadimberezniker, line 246: Infof

server/util/approxlru/approxlru.go
```diff
@@ -250,13 +235,30 @@ func (l *LRU[T]) resampleK(k int) error {
...
+			continue
+		}
+		if skip {
+			continue
+		}
+		if sample.Timestamp != timestamp {
+			log.Warningf("Evictor skipping %q; atime has changed %s -> %s", sample.Key, sample.Timestamp, timestamp)
```
vadimberezniker, line 253: Infof

vadimberezniker (approved):

server/util/approxlru/approxlru.go
tylerwilliams: Done.

server/util/approxlru/approxlru.go
tylerwilliams: Done.

PR #4379, opened 2023-07-24
lru: use go generic

sluongng:
Make use of go generic to make LRU package type-safe.

This restricts `key`'s type from `interface{} / any` to `string`, which is what we use throughout the entire code base.
It also lets us eliminate the type switch inside `keyHash()` func, further simplifying our code.

**Related issues**: N/A

### DIFF
--- a/server/interfaces/interfaces.go
+++ b/server/interfaces/interfaces.go
@@ -962,26 +962,26 @@ type XcodeLocator interface {
 }
 
 // LRU implements a Least Recently Used cache.
-type LRU interface {
+type LRU[V any] interface {
 	// Inserts a value into the LRU. A boolean is returned that indicates
 	// if the value was successfully added.
-	Add(key, value interface{}) bool
+	Add(key string, value V) bool
 
 	// Inserts a value into the back of the LRU. A boolean is returned that
 	// indicates if the value was successfully added.
-	PushBack(key, value interface{}) bool
+	PushBack(key string, value V) bool
 
 	// Gets a value from the LRU, returns a boolean indicating if the value
 	// was present.
-	Get(key interface{}) (interface{}, bool)
+	Get(key string) (V, bool)
 
 	// Returns a boolean indicating if the value is present in the LRU.
-	Contains(key interface{}) bool
+	Contains(key string) bool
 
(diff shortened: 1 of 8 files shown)


.bazelrc
```diff
@@ -33,7 +33,7 @@ build:remote-shared --remote_upload_local_results
 build:remote-shared --remote_timeout=600
 build:remote-shared --remote_download_minimal
 build:remote-shared --experimental_repo_remote_exec
-build:remote-shared --jobs=100
+build:remote-shared --jobs=250
```
sluongng, line 36:
```suggestion
build:remote-shared --jobs=100
```

tylerwilliams (approved):

server/interfaces/interfaces.go
```diff
@@ -962,26 +962,26 @@ type XcodeLocator interface {
 }
 
 // LRU implements a Least Recently Used cache.
-type LRU interface {
+type LRU[V any] interface {
 	// Inserts a value into the LRU. A boolean is returned that indicates
 	// if the value was successfully added.
-	Add(key, value interface{}) bool
+	Add(key string, value *V) bool
```
bduffany, line 968: If we want an LRU holding a primitive type then we have to pass in pointers, vs the plain old primitive values. This feels slightly clunky? Also, strings in particular are already pointers in a sense, so the extra pointer dereference might be inefficient.

wdyt about using `V` instead of `*V` for all of these and then at each call site declaring `LRU[*myStruct]` instead of the current `LRU[myStruct]`? (not sure if there's a best practice here...)

sluongng: I tried that at first.

However there are some tricky part such as `Get(key string) (*V, bool)` where turning `*V -> V` here would make it impossible to return nil value. 🤔 
And we want to avoid `Add()` in one data type and `Get()` in another.

Some popular open-source container types such as https://github.com/zyedidia/generic/blob/master/btree/btree.go would get around it by returning a default value instead of nil.

```go
func (t *Tree[K, V]) search(x *node[K, V], key K, height int) (V, bool) {
	...
	var v V
	return v, false
}
```

I could explore this direction in the next iteration

bduffany (approved):
nice :)

PR #4380, opened 2023-07-24
cmd/executor: make network config linux exclusive

sluongng:
Package `server/util/networking` only works on Linux.

**Related issues**: N/A


bduffany (approved):

bduffany (commented):
nit: do you mind adding `target_compatible_with` to `server/util/networking` to encode the fact that it's linux-only?

sluongng:
I will do that in a future PR. There is still a dependency path left through `runner -> podman -> networking` that needs to be decoupled. Working on it right now 👨‍💻 

image

PR #4381, opened 2023-07-24
Small styling tweaks for API key value

bduffany:
* Decrease font size to allow full API key to be displayed, instead of truncating w/ ellipsis
* Use unicode bullet `•` instead of `*` (looks a bit nicer)
* Reuse `<OutlinedButton>` w/ `icon-button` class for controls (had to remove `disabled` state + tweaked logic, since the default `[disabled]` styling for `<OutlinedButton>` shows a 'cancel' cursor which felt a bit weird)

**Related issues**: N/A

### DIFF
--- a/enterprise/app/api_keys/api_keys.css
+++ b/enterprise/app/api_keys/api_keys.css
@@ -86,10 +86,11 @@
 
 .api-keys .api-key-value {
   font-family: monospace;
-  font-size: 16px;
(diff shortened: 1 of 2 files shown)


bduffany:
> Do you know if there is a way to avoid the ellipsis `...`?

Fixed by changing the `flex-basis` of `api-key-value` to `fit-content`

PR #4382, opened 2023-07-24
runner: make podman and firecracker linux only

sluongng:
Exclude podman and firecracker runners on darwin and windows.

**Related issues**: N/A


enterprise/server/remote_execution/runner/runner_linux.go
```diff
@@ -0,0 +1,173 @@
...
+			dockerSocket = fmt.Sprintf("unix://%s", dockerSocket)
+		}
+		p.runnerClients.dockerClient, err = dockerclient.NewClientWithOpts(
+			dockerclient.WithHost(dockerSocket),
+			dockerclient.WithAPIVersionNegotiation(),
+		)
+		if err != nil {
+			return status.FailedPreconditionErrorf("Failed to create docker client: %s", err)
+		}
+	}
```
bduffany, line 51: there is a lot of code duplicated between `runner_linux.go` and `runner_nolinux.go`, like this highlighted section.

Do you mind moving these pieces of logic to functions in `docker.go` / `podman.go` etc.

Ideally we could have a `container.Provider` interface with 2 methods `Init()` (one-time setup) / `New()` (creates a `Container` instance for a task) but for now it'd also be fine to just move this code to top-level `Init` / `New` functions in the appropriate package. (Eventually I'd like to move to the Provider interface approach and then only bind providers that are flag-enabled and compatible with the current platform. That would let us remove hacks like `firecracker_darwin.go` since the container provider setup for darwin would not depend on firecracker at all.)

sluongng: I am not happy with the current state of this PR either. 🥲
It's very MVP-ish (thus draft status).
Let me take a look at the interface `Init() / New()` approach. Sounds reasonable.

> (Eventually I'd like to move to the Provider interface approach and then only bind providers that are flag-enabled and compatible with the current platform. That would let us remove hacks like firecracker_darwin.go since the container provider setup for darwin would not need to depend on firecracker at all.)

Agree. The final state should be VFS, firecracker, podman are linux-exclusive dependencies.
And our darwin (and potentially windows) executor would not have to build with them.

Though it's gonna take a while as some of these refactoring changes are quite clunky to ship all at once.

bduffany: > Though it's gonna take a while as some of these refactoring changes are quite clunky to ship all at once.

If you decide to go with the interface approach, one way to hopefully reduce clunk factor could be to add the interface definition first + env / runner boilerplate to use the new interface, and then add the interface implementations one-by-one as separate PRs

sluongng:
@bduffany PTAL. Thanks to the previous refactoring, this PR looks so much cleaner now 🙏

enterprise/server/remote_execution/runner/BUILD
```diff
@@ -4,17 +4,17 @@ package(default_visibility = ["//enterprise:__subpackages__"])
 
 go_library(
     name = "runner",
-    srcs = ["runner.go"],
+    srcs = [
+        "runner.go",
+        "runner_linux.go",
+        "runner_nolinux.go",
```
bduffany, line 10: The `sandbox` isolation type is darwin-specific, so we probably want separate `_darwin` and `_windows` files.

enterprise/server/remote_execution/runner/runner_linux.go
```diff
@@ -0,0 +1,37 @@
...
+	podmanProvider, err := podman.NewProvider(p.env, p.imageCacheAuth, *rootDirectory)
+	if err != nil {
+		log.Warningf("Failed to initialize podman container provider: %s", err)
+	}
+	if podmanProvider != nil {
+		providers[platform.PodmanContainerType] = podmanProvider
+	}
+	providers[platform.FirecrackerContainerType] = firecracker.NewProvider(p.env, p.imageCacheAuth, *rootDirectory)
+	providers[platform.SandboxContainerType] = &sandbox.Provider{}
```
bduffany, line 33: remove (this is darwin-specific)

enterprise/server/remote_execution/runner/runner_linux.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/platform"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+)
+
+func (p *pool) initContainerProviders() {
+	providers := make(map[platform.ContainerType]container.Provider)
+	dockerProvider, err := docker.NewProvider(p.env, p.imageCacheAuth, p.hostBuildRoot())
+	if err != nil {
+		log.Warningf("Failed to initialize docker container provider: %s", err)
+	}
```
bduffany, line 21: This should return an error that propagates up and causes the executor to fail initialization. Otherwise the executor could be in a state where it appears healthy but is unable to run tasks.

(Same for all the log.Warnings elsewhere)

enterprise/server/remote_execution/runner/BUILD
sluongng: Did not know about this. Good call!

enterprise/server/remote_execution/runner/runner_linux.go
sluongng: Makes sense. Fixed!

sluongng:
@bduffany I realized that I was missing the VFS stuff in my last iteration so I added them here.

If you want, I could split them into a separate PR as well.

enterprise/server/remote_execution/runner/runner_linux.go
```diff
@@ -0,0 +1,124 @@
...
+	if r.VFS != nil {
+		if err := r.VFS.Unmount(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if r.VFSServer != nil {
+		r.VFSServer.Stop()
+	}
+	return errs
+}
```
bduffany, line 124: nit: `[]error` feels a bit unconventional as a return value in go. If you want to `log.Error()` and return the last non-nil error (e.g. `if err != nil { lastErr = err } ... return lastErr`), I think that works.

sluongng: This is mostly because of the current way error handling works in `*commandRunner.Remove()`, which collect a slice of errors and then returns an `errSlice`, which are multiple errors concat into 1.

So I fixed `removeVFS` to return errSlice as well, mostly so that we could preserve the data of all errors in the slice, which could be valuable when troubleshooting.

sluongng: nvm I see the problem 🙈 
will fix!

bduffany (approved):

PR #4388, opened 2023-07-25
Revert "deps.bzl: remove unused attributes (#4316)"

bduffany:
This reverts commit d02a2e64eec8b96ff74607668ad35174b8dbbf19. The original commit breaks the "Push dev" workflow, resulting in the following errors when trying to build smarter device manager:

```
external/com_gitlab_arm_research_smarter_device_manager/nvidia-server.go:82:39: cannot use m.server (variable of type *"gitlab.com/arm-research/smarter/smarter-device-manager/vendor/google.golang.org/grpc".Server) as *"google.golang.org/grpc".Server value in argument to pluginapi.RegisterDevicePluginServer
external/com_gitlab_arm_research_smarter_device_manager/nvidia-server.go:122:44: cannot use conn (variable of type *"gitlab.com/arm-research/smarter/smarter-device-manager/vendor/google.golang.org/grpc".ClientConn) as *"google.golang.org/grpc".ClientConn value in argument to pluginapi.NewRegistrationClient
external/com_gitlab_arm_research_smarter_device_manager/server.go:79:39: cannot use m.server (variable of type *"gitlab.com/arm-research/smarter/smarter-device-manager/vendor/google.golang.org/grpc".Server) as *"google.golang.org/grpc".Server value in argument to pluginapi.RegisterDevicePluginServer
external/com_gitlab_arm_research_smarter_device_manager/server.go:118:44: cannot use conn (variable of type *"gitlab.com/arm-research/smarter/smarter-device-manager/vendor/google.golang.org/grpc".ClientConn) as *"google.golang.org/grpc".ClientConn value in argument to pluginapi.NewRegistrationClient
```

Verify with `bazel build @com_gitlab_arm_research_smarter_device_manager//:smarter-device-manager`

**Related issues**: N/A

PR #4389, opened 2023-07-25
smarter-device-manager: create an alias to validate internal build

sluongng:
A previous change in d02a2e64eec8b96ff74607668ad35174b8dbbf19 broke the
"manual" build that depends on this Go package.

Revert the relevant buildd directive.

Add an alias for the external dep to guard our future changes against
it.

**Related issues**: replace #4388


bduffany (approved):

sluongng:
Reverted https://github.com/buildbuddy-io/buildbuddy/pull/4388 and rebase the change in this PR on top of the revert commit.

PR #4392, opened 2023-07-25
Fix race condition when merging diff snapshots with uffd enabled

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -496,9 +496,7 @@ func (s *COWStore) copyChunkIfNotDirty(chunkStartOffset int64) (err error) {
 	}
 	// Once we've created a copy, we no longer need the source chunk.
 	defer func() {
-		if src != nil {
-			src.Close()
-		}
+		src.Close()
 	}()
 
 	b := s.copyBufPool.Get().(*[]byte)
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -201,8 +201,10 @@ type Chunk struct {
 //
 // A COWStore can be created either by splitting a file into chunks, or loading
 // chunks from a directory containing artifacts exported by a COWStore instance.
+//
+// A COWStore supports concurrent reads/writes, as long as they are to different chunks
 type COWStore struct {
-	mu sync.RWMutex
+	mu sync.RWMutex // Protects chunks and dirty
 
 	// Chunks is a mapping of chunk offset to Store implementation.
 	chunks map[int64]*Chunk
@@ -278,12 +280,16 @@ func (c *COWStore) GetPageAddress(offset uintptr, write bool) (uintptr, error) {
 			return 0, status.WrapError(err, "copy chunk")
 		}
 	}
+
+	c.mu.RLock()
 	chunk := c.chunks[chunkStartOffset]
+	c.mu.RUnlock()
 	if chunk == nil {
 		// No data (yet); map into our static zero-filled buf. Note that this
 		// can only happen for reads, since for writes we call copyChunkIfNotDirty above.
 		return memoryAddress(c.zeroBuf) + chunkRelativeAddress, nil
 	}
+
 	// Non-empty chunk; initialize the lazy mmap (if applicable) and return
 	// the offset relative to the memory start address.
 	mm, ok := chunk.Store.(*Mmap)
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #4392** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4392" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4392?utm_source=stack-comment).

buildbuddy.yaml
```diff
@@ -77,7 +77,7 @@ actions:
         branches:
           - "*"
     bazel_commands:
-      - test //... --config=linux-workflows --test_tag_filters=+bare --build_tag_filters=+bare
+      - test //... --config=linux-workflows --test_tag_filters=+bare --build_tag_filters=+bare --runs_per_test=200 --notest_keep_going
```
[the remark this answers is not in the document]
maggie-lou: Yeah sorry pr isn't finalized yet - marked as draft

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -52,6 +53,7 @@ func (f *File) SizeBytes() (int64, error) {
 // Mmap implements the Store interface using a memory-mapped file. This allows processes to read/write to the file as if it
 // was memory, as opposed to having to interact with it via I/O file operations.
 type Mmap struct {
+	mu     sync.RWMutex
```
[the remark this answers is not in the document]
maggie-lou: In mergeDiffSnapshot we have a bunch of concurrent goroutines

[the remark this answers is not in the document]
bduffany: for the concurrent goroutines issue, one option is to make sure goroutines don't cross chunk boundaries. Maybe each goroutine could process N chunks (e.g. N=16 or something) and use errgroup / SetLimit to make sure there are still only 4 workers active at a time like we have today

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -446,6 +452,14 @@ func (s *COWStore) calculateChunkSize(startOffset int64) int64 {
 }
 
 func (s *COWStore) copyChunk(chunkStartOffset int64) (err error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
bduffany, line 456: I think it will hurt performance to have this whole func locked since the goroutines in MergeDiffSnapshot will be blocked waiting for the full copy to complete. We probably want just the map accesses to be locked

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -337,18 +340,19 @@ func (c *COWStore) WriteAt(p []byte, off int64) (int, error) {
...
 		// On each iteration, write to one chunk, first copying the readonly
 		// chunk if needed.
-		if !c.dirty[chunkOffset] {
-			// If we're newly dirtying a chunk, copy.
-			if err := c.copyChunk(chunkOffset); err != nil {
-				return 0, status.WrapError(err, "failed to copy chunk")
-			}
+		if err := c.copyChunk(chunkOffset); err != nil {
+			return 0, status.WrapError(err, "failed to copy chunk")
 		}
```
bduffany, line 345: imo the old code was slightly clearer - you could use the new `Dirty()` func to ensure `c.dirty` access is locked, or to keep the code less indented you could rename `copyChunk` to `copyChunkIfDirty`

enterprise/server/remote_execution/blockio/blockio.go
maggie-lou: Done

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -201,6 +202,8 @@ type Chunk struct {
 // A COWStore can be created either by splitting a file into chunks, or loading
```
bduffany, line 202: nit: maybe add a comment that COWStore supports concurrent reads and writes as long as the reads/writes are to different chunks.

bduffany (approved):
Awesome!

enterprise/server/remote_execution/blockio/blockio.go
maggie-lou: Done - previously I had only made the blockio code that was used in mergeDiffSnapshot concurrency-safe, so made it concurrency-safe everywhere

maggie-lou:
@bduffany - Caught another sneaky race condition with copyBuf. Could you take a look at my new changes?

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -445,33 +472,43 @@ func (s *COWStore) calculateChunkSize(startOffset int64) int64 {
...
 
 	if src == nil {
 		// We had no data at this offset; nothing to copy.
 		return nil
 	}
+	// Once we've created a copy, we no longer need the source chunk.
+	defer func() {
+		if src != nil {
+			src.Close()
+		}
```
bduffany, line 501: nit: can remove the nil check here

bduffany (approved):

PR #4395, opened 2023-07-25
Make FindMissing chunk aware

luluz66:
- find missing should check all the chunks
- use findMissing in contains

**Related issues**: N/A


tylerwilliams (commented):
LGTM

PR #4399, opened 2023-07-26
[UFFD] Copy entire chunks when handling page faults with uffd handler

maggie-lou:
To optimize performance of the UFFD handler, copy an entire chunk of memory when handling page faults, as opposed to just one page. This will reduce the overhead of having to make lots of copy calls for small memory ranges

A chunk size of 1000 memory pages yielded the best performance results in my testing. Some test results:

**Without UFFD:**
Initial pause: ~1s
Subsequent pauses: ~220ms
LoadSnapshot: ~420ms
Unpause: ~420ms

**UFFD loading 1 page at a time:**
Initial pause: ~2.9s
Subsequent pause: ~2.4s
Initial LoadSnapshot: ~2s
Subsequent LoadSnapshot: ~1.3s
Initial Unpause: ~2s
Subsequent Unpause: ~1.3s

**UFFD Loading 512KB chunks:**
Initial pause: ~2.7s
Subsequent pause: ~2.3s
LoadSnapshot: ~565ms
Unpause: ~565ms

**[Chosen] UFFD Loading 1000 page chunks:**
Initial pause: ~2.7s
Subsequent pause: ~2s
LoadSnapshot: ~190ms
Unpause: ~190ms

**UFFD Loading 1500 page chunks:**
Initial pause: ~2.7s
Subsequent pause: ~2s
LoadSnapshot: ~225ms
Unpause: ~225ms

### DIFF
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -262,6 +262,21 @@ func NewCOWStore(chunks []*Chunk, chunkSizeBytes, totalSizeBytes int64, dataDir 
 	}, nil
 }
 
+func (c *COWStore) GetRelativeOffsetFromChunkStart(offset uintptr) uintptr {
+	chunkStartOffset := c.chunkStartOffset(int64(offset))
+	chunkRelativeAddress := offset - uintptr(chunkStartOffset)
+	return chunkRelativeAddress
+}
+
+func (c *COWStore) GetChunkStartAddressAndSize(offset uintptr, write bool) (uintptr, int64, error) {
+	chunkStartOffset := c.chunkStartOffset(int64(offset))
+	chunkStartAddress, err := c.GetPageAddress(uintptr(chunkStartOffset), write)
+	if err != nil {
+		return 0, 0, err
+	}
+	return chunkStartAddress, c.chunkSizeBytes, nil
+}
+
 // GetPageAddress returns the memory address for the given byte offset into the store.
 //
 // This memory address can be used to handle a page fault with userfaultfd.
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -272,15 +272,14 @@ func (c *COWStore) GetRelativeOffsetFromChunkStart(offset uintptr) uintptr {
 	return chunkRelativeAddress
 }
 
-// GetChunkStartAddressAndSize returns the start address of the chunk containing the input offset,
-// and the chunk size
-func (c *COWStore) GetChunkStartAddressAndSize(offset uintptr, write bool) (uintptr, int64, error) {
+// GetChunkStartAddress returns the start address of the chunk containing the input offset.
+func (c *COWStore) GetChunkStartAddress(offset uintptr, write bool) (uintptr, error) {
 	chunkStartOffset := c.chunkStartOffset(int64(offset))
 	chunkStartAddress, err := c.GetPageAddress(uintptr(chunkStartOffset), write)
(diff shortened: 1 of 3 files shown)


maggie-lou:
I'm going to do some additional investigation why the Pause latency is so high

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -126,7 +126,7 @@ const (
 	minScratchDiskSizeBytes = 64e6
 
 	// Chunk size to use when creating COW images from files.
-	cowChunkSizeBytes = 512 * 1024
+	cowChunkSizeInPages = 1000
```
bduffany, line 129: The optimal chunk size might be different from the optimal UFFD copy size, since we'll be copying data to/from cache at chunk granularity. Can we have 2 separate knobs for these? Maybe cowChunkSizeBytes / uffdCopySizeInPages

bduffany: (Also, if we have 2 separate knobs for chunkSize vs uffdCopySize, it might change how the tests perform; may be worth re-running just to double check that 1000 pages is still the optimal number)

maggie-lou: I'm not sure if this is worth it for now. Right now, the logic is simple because we always copy entire chunks. If we make the copy size != chunk size, we will also need to track which addresses we've already copied (you can't copy the same address twice)

Increasing the copy size changes the speed by ms. Copying 500 pages takes ~1ms, copying 1000 pages takes 2ms. If fetching remote blocks becomes the bottleneck, it seems like we might as well just copy the entire chunk to avoid having to make more round trips

bduffany: I think Tyler recently found that 10MB was the optimal-ish size for transferring files between the executor and the app - 4MB isn't too far off so keeping this for now sgtm

bduffany (approved):

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -262,6 +262,27 @@ func NewCOWStore(chunks []*Chunk, chunkSizeBytes, totalSizeBytes int64, dataDir
 	}, nil
 }
 
+// GetRelativeOffsetFromChunkStart returns the relative offset from the beginning of the chunk
+//
+// Ex. Let's say there are 100B chunks. For input offset 205, chunkStartOffset would be 200,
+// and this would return 5
```
tylerwilliams, line 268: Can you format these lines so they are less than or equal to 80 chars?

enterprise/server/remote_execution/containers/firecracker/firecracker.go
tylerwilliams: Yeah these were the (pretty rough) numbers:

```
	- can upload a 1GB file, compressed, serially, in 16seconds []
	- can upload the same 1GB file but chunked into 10 100M parts, in 1.75 seconds [500MB/sec]
	- can upload the same 1GB file but chunked into 100 10M parts, in 0m2.489s
	- can upload the same 1GB file but chunked into 1000 1M parts, in 11.618s
```

tylerwilliams (approved):
1 nit

bduffany:
> these were the (pretty rough) numbers

oh, I completely misread originally, it looks like larger chunk sizes are better in terms of cache throughput. This might be even more reason to have 2 separate knobs (UFFD copy size and chunk file size), assuming UFFD copies take longer for larger chunks? but either way I think it's fine to do this much later when we start taking a closer look at performance

PR #4400, opened 2023-07-26
Add support for usage labels

bduffany:
Adds usage label columns and logic to store the column values in the DB, but does not yet add the logic to compute the label values from the client request.

Columns must be flag-enabled one-by-one after all apps are updated with the new label schema.

**Related issues**: N/A


bduffany:
Chatted with Tyler - will see if we can simplify this by only flushing collection period rows to the DB instead of usage period rows. That would let us only ever have to INSERT rows to the DB and also make it so we don't have to worry about backwards/forwards compatibility as much.

enterprise/server/usage/usage.go
```diff
@@ -414,6 +495,14 @@ func countsToMap(tu *tables.UsageCounts) (map[string]int64, error) {
 	return counts, nil
 }
 
+func labelsFromJSON(j string) (*tables.UsageLabels, error) {
```
vadimberezniker, line 498: I think this is unused.

vadimberezniker (approved):

enterprise/server/usage/usage.go
bduffany: removed

bduffany:
I am going to submit this after https://github.com/buildbuddy-io/buildbuddy/pull/4403 rolls out to avoid changing too much at once.

PR #4403, opened 2023-07-26
usage: flush to DB in 1 minute intervals

bduffany:
The goal of this PR is to simplify usage data flushing, particularly since we're adding new usage label and count columns soon.

By flushing at 1 minute intervals, we no longer have a distinction between "usage periods" and "collection periods," which makes the code much simpler - each key in redis corresponds to a single row in the DB, rather than multiple "collection period" entries in redis mapping to the same "usage period" row in the DB.

One advantage of this is that we never have to run UPDATE queries when flushing - only INSERT. With updates, we have to worry about accidentally updating rows we shouldn't (causing misattribution). If we're only inserting, then the worst that happens is that we potentially don't record some label values that the current app doesn't yet support (during a rollout), which is fine.

This change should be forward compatible with old apps in the deployment that might try to flush at 1-hour granularity. The DB will be in a slightly weird state in that some usage rows will reflect the old "usage period + collection period" flushing mechanism with 1 hour granularity, and others will have 1 minute granularity, but the data in aggregate should still be correct.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/usage/usage.go
+++ b/enterprise/server/usage/usage.go
@@ -276,7 +276,7 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, p period, co
 		Region:          ut.region,
 	}
 	dbh := ut.env.GetDBHandle()
-	return dbh.TransactionWithOptions(ctx, db.Opts().WithQueryName("upsert_usage"), func(tx *db.DB) error {
+	return dbh.TransactionWithOptions(ctx, db.Opts().WithQueryName("insert_usage"), func(tx *db.DB) error {
 		log.Debugf("Flushing usage counts for key %+v", pk)
 
 		// First check whether the row already exists. Make sure to select for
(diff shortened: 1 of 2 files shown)


enterprise/server/usage/usage.go
```diff
@@ -292,200 +292,97 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, c collection
...
-			}
-			if unsupportedField != "" {
-				alert.UnexpectedEvent("usage_update_dropped", "Usage update transaction aborted since existing usage row contains unsupported column %q = %q (key = %+v)", unsupportedField, unsupportedFieldValue, pk)
-				return nil
-			}
-		}
-		if err := rows.Err(); err != nil {
-			return err
+		if err == nil {
+			log.Warningf("Usage update dropped since the row already exists; this should only happen if redis locking has failed.")
```
tylerwilliams, line 300: do you want to make this an unexpected event?

I think we want to treat usage data failing as a pageable offense :)

bduffany: I was on the fence about this since (a) in this case, we're not actually 'dropping' usage (the warning is somewhat confusingly worded), we're just skipping an update because it's a dupe, and (b) I figured we don't want to treat the redis locking stuff as a hard dependency. But I guess it doesn't hurt to alert here, since if redis goes down we're probably getting alerted anyway :sweat_smile:

tylerwilliams (approved):

PR #4412, opened 2023-07-27
Retrieve API key on demand.

vadimberezniker:
For backwards compatibility, we skip retrieving the API key value if it's alraedy been returned as part of the list request.

**Related issues**: N/A


enterprise/app/api_keys/api_keys.tsx
```diff
@@ -596,25 +597,53 @@ class ApiKeyField extends React.Component<ApiKeyFieldProps, ApiKeyFieldState> {
...
   private copyTimeout: number | undefined;
+  private value: string | undefined;
+
+  componentDidMount() {
+    if (this.props.apiKey.value) {
+      this.value = this.props.apiKey.value
+    }
+  }
+
+  private async retrieveValue(fn: (val: string) => void) {
```
bduffany, line 608: nit: since you're using `async/await` here, this func will return `Promise<string>`, so you could write

```
return this.value
```

instead of

```
fn(this.value);
return;
```

then invoke the func like

```
const value = this.retrieveValue().then((val: string) => { ...});
```

instead of passing in a callback arg.

bduffany (approved):

enterprise/app/api_keys/api_keys.tsx
vadimberezniker: Done

vadimberezniker:
> FYI I think we also return API keys in `GetBazelConfig` in case you want to update that as well.

Thanks, completely forgot about that code path.

app/docs/setup_code.tsx
```diff
@@ -53,11 +55,12 @@ export default class SetupCodeComponent extends React.Component<Props, State> {
 
   componentDidUpdate(prevProps: Props) {
     if (this.props.bazelConfigResponse !== prevProps.bazelConfigResponse) {
-      this.setConfigResponse(this.props.bazelConfigResponse);
+      this.setConfigResponse(this.props.bazelConfigResponse!);
```
bduffany, line 58: nit: since `this.props.bazelConfigResponse` is optional (based on the `Props` type) we might want to keep `setConfigResponse()` supporting undefined response values (`?:`) and drop the `!` assertion here (`fetchAPIKeyValue` probably needs a small tweak as well)

bduffany (approved):

app/docs/setup_code.tsx
vadimberezniker: done

app/docs/setup_code.tsx
```diff
@@ -197,8 +202,32 @@ export default class SetupCodeComponent extends React.Component<Props, State> {
...
+        })
+        .catch((e) => error_service.handleError(e));
+    }
+  }
+
   onChangeCredential(e: React.ChangeEvent<HTMLSelectElement>) {
-    this.setState({ selectedCredentialIndex: Number(e.target.value) });
+    const selectedIndex = Number(e.target.value);
+    this.fetchAPIKeyValue(this.state.bazelConfigResponse!, selectedIndex);
+    this.setState({ selectedCredentialIndex: selectedIndex });
```
sluongng, line 230: We might need to implement a hide/show button for this page (Quickstart), otherwise, we will generate an Audit Log entry every time this page is loaded. 🤔 

Might not be within the scope of this PR though. WDYT?

vadimberezniker: This seems fine for now. I could be wrong but I don't expect this page to be accessed very frequently. If we see any customer confusion wrt this we can revisit it.

PR #4414, opened 2023-07-27
Fix up github auth redirects

siggisim:
This PR fixes two issues with the current Github auth setup relating to redirects:
1) This uses the standard `Redirect-Url` redirect cookie instead of the more specific `Github-Redirect-Url` so that Github logins with a redirect_url parameter set are correctly redirected after the user logs in.
2) If Github auth is enabled, and a user hits the `/auth/github/link/` endpoint but is not already logged in - redirect them to the github login flow instead of to `/`.

### DIFF
--- a/server/backends/github/github.go
+++ b/server/backends/github/github.go
@@ -20,6 +20,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/http/interceptors"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
 	"github.com/buildbuddy-io/buildbuddy/server/util/authutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/cookie"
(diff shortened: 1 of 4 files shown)


server/backends/github/github.go
```diff
@@ -244,9 +245,11 @@ func (c *OAuthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	_, err := perms.AuthenticatedUser(r.Context(), c.env)
 	if err != nil {
 		// If not logged in to the app (e.g. when installing directly from
-		// GitHub), log in first, then come back here to complete the
-		// installation.
-		loginURL := fmt.Sprintf("/?redirect_url=%s", url.QueryEscape(r.URL.String()))
+		// GitHub), redirect to the account creation flow flow.
```
bduffany, line 248:
```suggestion
		// GitHub), redirect to the account creation flow.
```

bduffany (approved):

PR #4415, opened 2023-07-27
Add a flag that logs the output of podman pull operations

iain-macdonald:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2523

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -48,6 +48,9 @@ import (
 )
 
 var (
+	// TODO(#2523): remove this flag.
+	podmanPullLogLevel = flag.String("executor.podman.pull_log_level", "", "Level at which to log `podman pull` command output. Should be one of the standard log levels, all lowercase.")
+
 	// Note: to get these cgroup paths, run a podman container like
 	//     podman run --rm --name sleepy busybox sleep infinity
 	// then look at the output of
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -765,13 +768,26 @@ func (c *podmanCommandContainer) pullImage(ctx context.Context, creds container.
...
 	// Use server context instead of ctx to make sure that "podman pull" is not killed when the context
 	// is cancelled. If "podman pull" is killed when copying a parent layer, it will result in
 	// corrupted storage.  More details see https://github.com/containers/storage/issues/1136.
 	pullCtx, cancel := context.WithTimeout(c.env.GetServerContext(), *pullTimeout)
 	defer cancel()
-	pullResult := runPodman(pullCtx, "pull", &container.Stdio{}, podmanArgs...)
+	stdio := &container.Stdio{}
+	if *podmanPullLogLevel != "" {
+		stdio = &container.Stdio{
+			Stdin:  nil,
```
vadimberezniker, line 785: nit: you don't need to specify keys with nil values

vadimberezniker (approved):

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Ah, thanks, removed it

PR #4416, opened 2023-07-27
Add protos for CreateRepo

siggisim:
This proto will be used to create Git repositories in the UI, optionally with a "template" repo that is used to populate the repo's initial content.


proto/repo.proto
```diff
@@ -0,0 +1,53 @@
...
+message CreateRepoRequest {
+  // The request context.
+  context.RequestContext request_context = 1;
+
+  // The name of the repo, i.e. "buildbuddy".
+  string name = 2;
+
+  // The organization that the repo should be created under ex. "buildbuddy-io".
+  // If this is empty, the repo will be created under the authenticated Github
+  // User's account.
```
tylerwilliams, line 16: could we require this to always be set (client picks either repo or puts in user acct name)?

proto/repo.proto
```diff
@@ -0,0 +1,53 @@
...
+  string organization = 3;
+
+  // A short description of the repo, ex. "BuildBuddy is an open source Bazel
+  // build event viewer, result store, remote caching, and remote build
+  // execution platform."
+  string description = 4;
+
+  // An existing git repository to use as a template for the contents of the new
+  // repo. ex.
+  // "https://github.com/buildbuddy-io/go-typescript-starter-template.git"
```
tylerwilliams, line 26: nit: maybe say what we do with the template / how it's used?

proto/repo.proto
```diff
@@ -0,0 +1,53 @@
...
+  // If true, the repository will be created as a private repo.
+  bool private = 6;
+
+  // The platform on which the repo should be hosted. Currently only Github is
+  // supported, but in the future we'll likely support more git providers.
+  GitHost host = 7;
+
+  // The Github app installation id that should be used to create the
+  // repository. Only required for repositories that are hosted on Github.
+  int64 installation_id = 8;
```
tylerwilliams, line 38: later we can wrap this kinda stuff into a provider specific message

tylerwilliams (approved):
just nits

proto/repo.proto
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// The platform on which the repository should be hosted.
+enum GitHost {
```
tylerwilliams, line 42: oh maybe call this GitProvider instead of GitHost (might be less confusing if we ever talk to a enterprise hosted git provider)

siggisim: Done

proto/repo.proto
siggisim: Yeah I'm down to do that once we have multiple providers

proto/repo.proto
siggisim: Done

proto/repo.proto
siggisim: There's a subtle distinction between `organization` and `owner`.

I decided to go with `organization` because it I need to feed it into this method which has takes either an organization or an empty string if you want to create it for the authenticated user instead https://github.com/google/go-github/blob/c8c34d01f65e98954a092333edf4ecdf3120c36e/github/repos.go#L428

I can switch to `owner` if you prefer, but in that case I'd also need a boolean / enum telling me whether or not the owner is an organization (Github weirdly has different endpoints for creating a [repo under an organization](https://docs.github.com/en/free-pro-team@latest/rest/repos/repos?apiVersion=2022-11-28#create-an-organization-repository) or under the [authenticated user](https://docs.github.com/en/free-pro-team@latest/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-for-the-authenticated-user) which I believe is the root of this complexity).

tylerwilliams: alright i'm happy with your simple approach for now, we can revisit if it needs more later :)

siggisim: And yes, the org endpoint blows up if you try to pass the username as the org name - that was the first thing I tried haha

PR #4417, opened 2023-07-27
Define custom buckets for eviction histogram.

vadimberezniker:
Less buckets in <24 hour range, more buckets in 1-7 day range.

**Related issues**: N/A


tylerwilliams (approved):

tylerwilliams:
this rocks

PR #4418, opened 2023-07-27
Implement `CreateRepo`

siggisim:
CreateRepo currently does 3 things:
- Creates a github repo with the correct owner
- If a template is specified, copies the contents of that template repo into the new repo
- "Links" the newly created repo in order to enable BuildBuddy Workflows


enterprise/server/githubapp/githubapp.go
```diff
@@ -23,15 +25,19 @@ import (
...
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/config"
 	"github.com/golang-jwt/jwt"
 	"github.com/google/go-github/v43/github"
 	"golang.org/x/oauth2"
 
 	gh_webhooks "github.com/buildbuddy-io/buildbuddy/enterprise/server/webhooks/github"
 	ghpb "github.com/buildbuddy-io/buildbuddy/proto/github"
+	repb "github.com/buildbuddy-io/buildbuddy/proto/repo"
```
tylerwilliams, line 36: how bout rppb here too since we use repb everywhere else for remote execution proto?

enterprise/server/githubapp/githubapp.go
```diff
@@ -586,6 +592,98 @@ func (a *GitHubApp) GetAccessibleGitHubRepos(ctx context.Context, req *ghpb.GetA
...
+	}
+	if tu.GithubToken == "" {
+		return nil, status.UnauthenticatedError("github account link is required")
+	}
+
+	// Pick the right client based on the request (organization or user).
+	githubClient, token, err := a.newAuthenticatedClient(ctx, tu.GithubToken)
+	if req.Organization != "" {
+		githubClient, token, err = a.newInstallationClient(ctx, req.InstallationId)
+	}
```
tylerwilliams, line 608: you're shadowing the first err here with the second before checking the first -- is this intentional?

enterprise/server/githubapp/githubapp.go
```diff
@@ -586,6 +592,98 @@ func (a *GitHubApp) GetAccessibleGitHubRepos(ctx context.Context, req *ghpb.GetA
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// If we have a template, copy the template contents to the new repo
+	if req.Template != "" {
+		tmpDir := fmt.Sprintf("%s/template-repo/%d/%s/", os.TempDir(), req.InstallationId, req.Name)
+		cloneTemplate(tmpDir, token, req.Template, *repo.CloneURL)
+	}
```
tylerwilliams, line 627: Outside the host ID file, we don't let the apps write files outside of the cache and (for remote_asset_api) a staging area. Could you use  "github.com/buildbuddy-io/buildbuddy/server/util/scratchspace" for this (just make the func tempDir() func public and replace os.TempDir() with it)?

Should be a drop in, and will guarantee your temp dir is on an ssd.

enterprise/server/githubapp/githubapp.go
```diff
@@ -586,6 +592,98 @@ func (a *GitHubApp) GetAccessibleGitHubRepos(ctx context.Context, req *ghpb.GetA
...
+		Private:     github.Bool(req.Private),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// If we have a template, copy the template contents to the new repo
+	if req.Template != "" {
+		tmpDir := fmt.Sprintf("%s/template-repo/%d/%s/", os.TempDir(), req.InstallationId, req.Name)
+		cloneTemplate(tmpDir, token, req.Template, *repo.CloneURL)
```
tylerwilliams, line 626: nit: check the error returned by cloneTemplate

enterprise/server/githubapp/githubapp.go
```diff
@@ -586,6 +592,98 @@ func (a *GitHubApp) GetAccessibleGitHubRepos(ctx context.Context, req *ghpb.GetA
...
+	// Make a temporary directory for the template
+	err := os.MkdirAll(tmpDir, os.ModePerm)
+	if err != nil {
+		return err
+	}
+	dir, err := ioutil.TempDir(tmpDir, "repo")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(dir)
```
tylerwilliams, line 657: maybe misunderstanding this -- but I think you want to use https://pkg.go.dev/os#MkdirTemp instead here?

enterprise/server/githubapp/githubapp.go
```diff
@@ -586,6 +592,98 @@ func (a *GitHubApp) GetAccessibleGitHubRepos(ctx context.Context, req *ghpb.GetA
...
+		URL:  srcURL,
+		Auth: auth,
+	})
+	if err != nil {
+		return err
+	}
+
+	// Create a new remote and push to it
+	remote, err := gitRepo.CreateRemote(&config.RemoteConfig{
+		Name: "new-repo",
```
tylerwilliams, line 674: should this be 'origin' or something? I'm a bit ig when it comes to git

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1468,5 +1468,9 @@ func (s *BuildBuddyServer) GetAuditLogs(ctx context.Context, request *alpb.GetAu
 }
 
 func (s *BuildBuddyServer) CreateRepo(ctx context.Context, request *repb.CreateRepoRequest) (*repb.CreateRepoResponse, error) {
-	return nil, status.UnimplementedError("Not implemented")
+	gh := s.env.GetGitHubApp()
+	if gh == nil {
+		return nil, status.FailedPreconditionError("github app not enabled in the server")
+	}
```
tylerwilliams, line 1474: if you wanted to be consistent with other handlers you could do:

```
	if gh == nil {
		return nil, status.UnimplementedError("Github service not configured")
	}
```

tylerwilliams (changes requested):

enterprise/server/githubapp/githubapp.go
siggisim: yeah, i don't really care about that original error if we're supposed to be using an installationClient

could make this an if/else but this allows me to avoid having to create all 3 variables before the if else.

down to do this some other way if you have a better idea

enterprise/server/githubapp/githubapp.go
siggisim: a remote called `origin` already exists (that's the original template repo that we cloned)

i originally had this deleting `origin` and recreating this new remote as `origin`, but figured it was cleaner / less steps to just create a new remote with a new name and push to that one

server/buildbuddy_server/buildbuddy_server.go
siggisim: done, i just copied the ones above me - but i went ahead and updated those too

enterprise/server/githubapp/githubapp.go
siggisim: done i think? seems like os.MkdirTemp and ioutil.TempDir do the same thing afaict, but one less import is nice

enterprise/server/githubapp/githubapp.go
siggisim: done - this is the one linter i wish we had enabled

enterprise/server/githubapp/githubapp.go
siggisim: done - i was going to ask you where i should put these temp files haha

enterprise/server/githubapp/githubapp.go
siggisim: done - did this in other places but forgot to update this one

enterprise/server/githubapp/githubapp.go
tylerwilliams: Yeah they deprecated ioutil and moved all that to os

enterprise/server/githubapp/githubapp.go
```diff
@@ -586,6 +592,101 @@ func (a *GitHubApp) GetAccessibleGitHubRepos(ctx context.Context, req *ghpb.GetA
...
+		Username: "github",
+		Password: token,
+	}
+	gitRepo, err := git.PlainClone(dir, false, &git.CloneOptions{
+		URL:  srcURL,
+		Auth: auth,
+	})
+	if err != nil {
+		return err
+	}
```
bduffany, line 673: should we maybe do this clone as a remote action (or add a TODO to do that soon?) Seems like these clones would cause bursts of CPU/disk usage (due to decompressing / unpacking git objects) and increase our cache tail latency.

The action could be a simple shell script (like `Arguments: []string{"bash", "-c", "git clone ... && git push ..."`), setting the token via env-var-overrides header like we do for workflows.

enterprise/server/githubapp/githubapp.go
```diff
@@ -586,6 +592,101 @@ func (a *GitHubApp) GetAccessibleGitHubRepos(ctx context.Context, req *ghpb.GetA
...
+	defer os.RemoveAll(dir)
+
+	// Clone the template into the directory
+	auth := &githttp.BasicAuth{
+		Username: "github",
+		Password: token,
+	}
+	gitRepo, err := git.PlainClone(dir, false, &git.CloneOptions{
```
bduffany, line 667: Should we set `Depth: 1` here? That would limit things a bit so that we don't have to clone the full history of the template repo, and instead just the latest commit. (I guess there are pros and cons here)

server/util/scratchspace/scratchspace.go
```diff
@@ -63,5 +63,5 @@ func tempDir() string {
 //	    os.Remove(f.Name())
 //	}()
 func CreateTemp(pattern string) (*os.File, error) {
-	return os.CreateTemp(tempDir(), pattern)
+	return os.CreateTemp(TempDir(), pattern)
```
bduffany, line 66: imo it would be better to use `scratchspace.CreateTemp()` as the top level dir, then create whatever dirs you want under that. That way we don't ever have to worry about naming collisions and we don't have to expose a new public API.

If you wanted a fixed directory name so that you could inspect it locally, you could still inspect it using something like

`cd /tmp/buildbuddy/template-repo-*/ && git status (etc...)`

siggisim: I tried this out, but `scratchspace.CreateTemp()` creates a new temporary file rather than creating a new temporary directory.

enterprise/server/githubapp/githubapp.go
siggisim: Added a TODO. My biggest concern with doing that would be latency (I don't want a new user to sit and wait in a queue in this onboarding flow) and I think these templates should be relatively small at least to start. Maybe we could add some sort of task priority system in the future like Interactive / Non-interactive that would give these tasks priority.

enterprise/server/githubapp/githubapp.go
siggisim: I tried this out originally, but I was getting some git `missing necessary objects` errors when pushing to the new remote.

enterprise/server/githubapp/githubapp.go
```diff
@@ -730,28 +831,28 @@ func (a *GitHubApp) newAppClient(ctx context.Context) (*github.Client, error) {
...
-		return nil, err
+		return nil, "", err
 	}
 	return a.newAuthenticatedClient(ctx, token.GetToken())
 }
 
 // newAuthenticatedClient returns a GitHub client authenticated with the given
 // access token.
-func (a *GitHubApp) newAuthenticatedClient(ctx context.Context, accessToken string) (*github.Client, error) {
+func (a *GitHubApp) newAuthenticatedClient(ctx context.Context, accessToken string) (*github.Client, string, error) {
```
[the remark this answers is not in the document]
siggisim: Done

enterprise/server/githubapp/githubapp.go
siggisim: Done

server/util/scratchspace/scratchspace.go
bduffany: We can add `scratchspace.MkdirTemp()` func (as the analogue of `os.MkdirTemp`)

siggisim: Done

bduffany (approved):

tylerwilliams (approved):

PR #4421, opened 2023-07-28
Fill eviction sample pool when approxlru is first created.

vadimberezniker:
Adjust eviction test to check that evicted entries are older than 95% of the remaining entries rather than strictly comparing to the oldest entry.

**Related issues**: N/A


tylerwilliams (approved):
Thanks for finding and fixing this!

PR #4423, opened 2023-07-28
docker: use ContainerAttach to capture std{out,err}

sluongng:
Our previous approach was prone to race conditions between the container
starting up and our call to `ContainerLogs` API. In the case where we call
the log API before container actually started, we get no outputs.  This
resulted in flaky tests observed in our CI.

Switch to use `ContainerAttach` API to capture the container outputs as
they are streamed out to reliably obtain the logs.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -22,6 +22,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/docker/docker/pkg/stdcopy"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
@@ -135,8 +136,19 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		result.Error = wrapDockerErr(err, "failed to create docker container")
 		return result
 	}
-
 	cid := createResponse.ID
+
+	hijackedResp, err := r.client.ContainerAttach(ctx, cid, dockertypes.ContainerAttachOptions{
+		Stream: true,
+		Stdout: true,
+		Stderr: true,
+	})
+	if err != nil {
+		result.Error = wrapDockerErr(err, "failed to attach to docker container")
+		return result
+	}
+	defer hijackedResp.Close()
+
 	err = r.client.ContainerStart(ctx, cid, dockertypes.ContainerStartOptions{})
 	if err != nil {
 		result.Error = wrapDockerErr(err, "failed to start docker container")
@@ -160,48 +172,33 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		}()
 	}()
 
-	r.copyContainerLogs(ctx, cid, result)
-	if result.Error != nil {
-		return result
-	}
-
-	// The container should not be running at this point since we're done copying
-	// logs, but we need to call ContainerWait to get the exit code.
-	statusCh, errCh := r.client.ContainerWait(ctx, cid, dockercontainer.WaitConditionNotRunning)
-
-	select {
-	case err := <-errCh:
-		result.Error = wrapDockerErr(err, "container did not exit cleanly")
-	case s := <-statusCh:
-		exitedCleanly = true
-		if s.Error != nil {
-			result.Error = wrapDockerErr(status.UnknownError(s.Error.Message), "failed to get container status")
-			return result
+	eg := &errgroup.Group{}
+	eg.Go(func() error {
+		var stdout, stderr bytes.Buffer
+		_, err := stdcopy.StdCopy(&stdout, &stderr, hijackedResp.Reader)
+		result.Stdout = stdout.Bytes()
+		result.Stderr = stderr.Bytes()
+		return wrapDockerErr(err, "failed to copy docker container output")
+	})
+	eg.Go(func() error {
+		statusCh, errCh := r.client.ContainerWait(ctx, cid, dockercontainer.WaitConditionNotRunning)
+		select {
+		case err := <-errCh:
+			return wrapDockerErr(err, "container did not exit cleanly")
+		case s := <-statusCh:
+			exitedCleanly = true
+			if s.Error != nil {
+				return wrapDockerErr(status.UnknownError(s.Error.Message), "failed to get container status")
+			}
+			result.ExitCode = int(s.StatusCode)
+			return nil
 		}
-		result.ExitCode = int(s.StatusCode)
+	})
+	if err := eg.Wait(); err != nil {
+		result.Error = err
 	}
-	return result
-}
 
-func (r *dockerCommandContainer) copyContainerLogs(ctx context.Context, cid string, result *interfaces.CommandResult) {
-	logOptions := dockertypes.ContainerLogsOptions{
-		ShowStdout: true,
-		ShowStderr: true,
-		Follow:     true,
-	}
-	logs, err := r.client.ContainerLogs(ctx, cid, logOptions)
-	if err != nil {
-		result.Error = wrapDockerErr(err, "failed to get docker container logs")
-		return
-	}
-	err = copyOutputs(logs, result, &container.Stdio{})
-	if closeErr := logs.Close(); closeErr != nil {
-		log.Warningf("Failed to close docker logs: %s", closeErr)
-	}
-	if err != nil {
-		result.Error = wrapDockerErr(err, "failed to read docker container logs")
-		return
-	}
+	return result
 }
 
 func wrapDockerErr(err error, contextMsg string) error {
@@ -299,7 +296,7 @@ func (r *dockerCommandContainer) hostConfig(workDir string) *dockercontainer.Hos
 		Resources: dockercontainer.Resources{
 			Devices: devices,
 			Ulimits: []*units.Ulimit{
-				&units.Ulimit{Name: "nofile", Soft: defaultDockerUlimit, Hard: defaultDockerUlimit},
+				{Name: "nofile", Soft: defaultDockerUlimit, Hard: defaultDockerUlimit},
 			},
 		},
 	}
--- a/enterprise/server/remote_execution/containers/docker/docker_test.go
+++ b/enterprise/server/remote_execution/containers/docker/docker_test.go
@@ -197,8 +197,8 @@ func TestDockerRun_Timeout_StdoutStderrStillVisible(t *testing.T) {
 	res := c.Run(ctx, cmd, workDir, container.PullCredentials{})
 
 	assert.True(
-		t, status.IsDeadlineExceededError(res.Error),
-		"expected DeadlineExceeded error, got: %s", res.Error)
+		t, status.IsUnavailableError(res.Error),
+		"expected UnavailableError error, got: %s", res.Error)
 	assert.Less(
 		t, res.ExitCode, 0,
 		"if timed out, exit code should be < 0 (unset)")
--- a/enterprise/server/remote_execution/containers/docker/BUILD
+++ b/enterprise/server/remote_execution/containers/docker/BUILD
@@ -23,6 +23,7 @@ go_library(
         "@com_github_docker_go_units//:go-units",
         "@org_golang_google_grpc//codes",
         "@org_golang_google_grpc//status",
+        "@org_golang_x_sync//errgroup",
     ],
 )
 


sluongng:
Another fix direction here is to retry again after the container have successfully exit

```
diff --git a/enterprise/server/remote_execution/containers/docker/docker.go b/enterprise/server/remote_execution/containers/docker/docker.go
index 47f31400a..4ce28009e 100644
--- a/enterprise/server/remote_execution/containers/docker/docker.go
+++ b/enterprise/server/remote_execution/containers/docker/docker.go
@@ -160,7 +160,7 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
                }()
        }()

-       r.copyContainerLogs(ctx, cid, result)
+       written := r.copyContainerLogs(ctx, cid, result)
        if result.Error != nil {
                return result
        }
@@ -168,7 +168,6 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
        // The container should not be running at this point since we're done copying
        // logs, but we need to call ContainerWait to get the exit code.
        statusCh, errCh := r.client.ContainerWait(ctx, cid, dockercontainer.WaitConditionNotRunning)
-
        select {
        case err := <-errCh:
                result.Error = wrapDockerErr(err, "container did not exit cleanly")
@@ -180,10 +179,15 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
                }
                result.ExitCode = int(s.StatusCode)
        }
+
+       if written == 0 {
+               r.copyContainerLogs(ctx, cid, result)
+       }
+
        return result
 }
```

This is somewhat cleaner, but less eager

bduffany:
I think the "right" way to do it might be to switch from using the `ContainerLogs` API to the `ContainerAttach` API. It looks like that's how the docker CLI implements `docker run` - https://github.com/docker/cli/blob/e0e27724390cbce21cf6d67568972a4227b07382/cli/command/container/run.go#L177

I think I might not have done it that way originally because the `hijackedIOStreamer` stuff looked slightly complex, but it might be the way to go in order to avoid flakiness.

sluongng:
Agree.

I was checking how Gitlab Runner is doing it and seems like attach is the way to go https://sourcegraph.com/gitlab.com/gitlab-org/gitlab-runner/-/blob/executors/docker/internal/exec/exec.go?L68

Let me explore that in the next iteration.

bduffany:
> Agree.
> 
> I was checking how Gitlab Runner is doing it and seems like attach is the way to go https://sourcegraph.com/gitlab.com/gitlab-org/gitlab-runner/-/blob/executors/docker/internal/exec/exec.go?L68
> 
> Let me explore that in the next iteration.

Nice, it looks like the `hijackedIOStreamer` stuff that the docker CLI is probably not necessary - seems like we can just use the response returned by ContainerAttach directly, and just call `stdcopy.StdCopy(stdoutbuf, stderrbuf, resp.Reader)`.

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -14,15 +14,17 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/docker/pkg/stdcopy"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+
```
bduffany, line 20: nit: can you merge this import section with the one below?

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -160,50 +173,37 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		}()
 	}()
 
-	r.copyContainerLogs(ctx, cid, result)
-	if result.Error != nil {
+	var stdout, stderr bytes.Buffer
+	eg := &errgroup.Group{}
+	eg.Go(func() error {
+		_, errCopy := stdcopy.StdCopy(&stdout, &stderr, hijackedResp.Reader)
+		return errCopy
```
bduffany: Wrap the error here instead of after `eg.Wait()`, since this func is the one that is copying container logs

+nit: can use the more conventional `err` here since there are no conflicting `err` variables in scope.

```suggestion
		_, err := stdcopy.StdCopy(&stdout, &stderr, hijackedResp.Reader)
		return wrapDockerError(err, "failed to copy container logs")
```

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -160,50 +173,37 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
...
+	})
+	eg.Go(func() error {
+		statusCh, errCh := r.client.ContainerWait(ctx, cid, dockercontainer.WaitConditionNotRunning)
+		select {
+		case err := <-errCh:
+			result.Error = wrapDockerErr(err, "container did not exit cleanly")
```
bduffany: `return wrapDockerErr(...)` here, since we are setting `result.Error` based on the return value below. (I think there might be a race condition otherwise)

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -160,50 +173,37 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
...
+			if s.Error != nil {
+				return wrapDockerErr(status.UnknownError(s.Error.Message), "failed to get container status")
+			}
+			result.ExitCode = int(s.StatusCode)
+		}
+
+		return nil
+	})
+	if err := eg.Wait(); err != nil {
+		result.Error = wrapDockerErr(err, "failed to copy docker container logs")
```
bduffany: `result.Error = err` to avoid double-wrapping.

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -160,50 +173,36 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
...
+			containerTerminated = true
+			if s.Error != nil {
+				return wrapDockerErr(status.UnknownError(s.Error.Message), "failed to get container status")
+			}
+			result.ExitCode = int(s.StatusCode)
+			return nil
+		}
+	})
+	if err := eg.Wait(); err != nil {
+		result.Error = wrapDockerErr(err, "failed to run docker container")
```
bduffany, line 197: avoid double-wrapping the error:

```suggestion
		result.Error = err
```

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -160,50 +173,36 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		}()
 	}()
 
-	r.copyContainerLogs(ctx, cid, result)
-	if result.Error != nil {
+	var stdout, stderr bytes.Buffer
+	eg := &errgroup.Group{}
+	eg.Go(func() error {
+		_, errCopy := stdcopy.StdCopy(&stdout, &stderr, hijackedResp.Reader)
+		return wrapDockerErr(errCopy, "failed to copy docker container output")
```
bduffany, line 180: nit: `err` is slightly more conventional here since there are no other error variables in scope.

```suggestion
		_, err := stdcopy.StdCopy(&stdout, &stderr, hijackedResp.Reader)
		return wrapDockerErr(err, "failed to copy docker container output")
```

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -160,50 +173,36 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
...
-	case s := <-statusCh:
-		exitedCleanly = true
-		if s.Error != nil {
-			result.Error = wrapDockerErr(status.UnknownError(s.Error.Message), "failed to get container status")
-			return result
-		}
-		result.ExitCode = int(s.StatusCode)
-	}
+	result.Stdout = stdout.Bytes()
+	result.Stderr = stderr.Bytes()
```
bduffany, line 202: we want to run this code if `err != nil` as well, to get partial output (useful for debugging).

enterprise/server/remote_execution/containers/docker/docker_test.go
```diff
@@ -197,8 +197,8 @@ func TestDockerRun_Timeout_StdoutStderrStillVisible(t *testing.T) {
 	res := c.Run(ctx, cmd, workDir, container.PullCredentials{})
 
 	assert.True(
-		t, status.IsDeadlineExceededError(res.Error),
-		"expected DeadlineExceeded error, got: %s", res.Error)
+		t, status.IsUnavailableError(res.Error),
```
bduffany, line 200: I think the expected error code is DeadlineExceeded - if the container impl is returning Unavailable then that's probably a bug

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -123,7 +125,7 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		result.Error = err
 		return result
 	}
-	createResponse, err := r.client.ContainerCreate(
+	createResp, err := r.client.ContainerCreate(
```
bduffany, line 128: nit: revert

```suggestion
	createResponse, err := r.client.ContainerCreate(
```

enterprise/server/remote_execution/containers/docker/docker.go
```diff
@@ -135,21 +137,32 @@ func (r *dockerCommandContainer) Run(ctx context.Context, command *repb.Command,
 		result.Error = wrapDockerErr(err, "failed to create docker container")
 		return result
 	}
+	cid := createResp.ID
+
```
bduffany, line 141: we can use `containerName` instead of `cid` for the attach request to simplify a tiny bit.

sluongng: I will stick with cid to show that the Attach call depends on ContainerCreate call and consistent with the other docker client API calls.

enterprise/server/remote_execution/containers/docker/docker_test.go
sluongng: Hmm, I am not sure why we were initially expecting `DeadlineExceeded` since this is not a GRPC call. Docker client is REST over HTTP or UDS. 🤔 

The error I am getting here is

```
                Error:          Received unexpected error:
                                rpc error: code = Unavailable desc = container did not exit cleanly:
```

which correlates to the `Wait` call getting context canceled without a terminated container, which seems to be the correct behavior here?

sluongng: So this is the source of the Wait API

```
func (cli *Client) ContainerWait(ctx context.Context, containerID string, condition container.WaitCondition) (<-chan container.WaitResponse, <-chan error) {
	if versions.LessThan(cli.ClientVersion(), "1.30") {
		return cli.legacyContainerWait(ctx, containerID)
	}

	resultC := make(chan container.WaitResponse)
	errC := make(chan error, 1)

	query := url.Values{}
	if condition != "" {
		query.Set("condition", string(condition))
	}

	resp, err := cli.post(ctx, "/containers/"+containerID+"/wait", query, nil, nil)
	if err != nil {
		defer ensureReaderClosed(resp)
		errC <- err
		return resultC, errC
	}

	go func() {
		defer ensureReaderClosed(resp)

		body := resp.body
		responseText := bytes.NewBuffer(nil)
		stream := io.TeeReader(body, responseText)

		var res container.WaitResponse
		if err := json.NewDecoder(stream).Decode(&res); err != nil {
			// NOTE(nicks): The /wait API does not work well with HTTP proxies.
			// At any time, the proxy could cut off the response stream.
			//
			// But because the HTTP status has already been written, the proxy's
			// only option is to write a plaintext error message.
			//
			// If there's a JSON parsing error, read the real error message
			// off the body and send it to the client.
			_, _ = io.ReadAll(io.LimitReader(stream, containerWaitErrorMsgLimit))
			errC <- errors.New(responseText.String())  // <--- RETURNED HERE
			return
		}

		resultC <- res
	}()

	return resultC, errC
}
```

it seems like the error was returned by the json decoder failure?

sluongng: https://github.com/moby/moby/blob/6f2e0fac19b502e63aad68fd3d2d6678c4889022/client/container_wait.go#L60C8-L72

sluongng: Im thinking of doing this
```
		case err := <-errCh:
			if err.Error() == "" {
				err = context.DeadlineExceeded
			}
			return wrapDockerErr(err, "container did not exit cleanly")
```

But Im not sure if it's the right thing to do 🤔 
Perhaps the right fix here is to patch Docker CLI to return the json decode error? 🤔

bduffany: Ah sorry, I remembered that the error code here doesn't matter, as long as we just return some error in this case.

The executor always checks to see whether the context deadline exceeded, and if so, it overrides the command execution error to have DeadlineExceeded status required by the RBE protocol: https://github.com/buildbuddy-io/buildbuddy/blob/d0e9b2a25b12befdbe32577cbd06a324677fcb8e/enterprise/server/remote_execution/executor/executor.go#L368-L372

So I think returning Unavailable is fine :)

bduffany (approved):
Nice

PR #4425, opened 2023-07-28
Use same CLI version in buildfix and checkstyle

bduffany:
buildfix.sh produces a nonempty diff today, presumably because the CLI version in checkstyle doesn't match the one in buildfix.

**Related issues**: N/A


sluongng (approved):
I was wondering where the changes in my local diff came from 🤣

Thanks for fixing this!

PR #4426, opened 2023-07-28
Add a basic grpc proxy for tests

bduffany:
I'm working on a small change that allows executors to retry task update streams so that tasks aren't affected as much during app rollouts. The test proxy will help us test this, because it will let us control how RPCs are routed to apps (by RPC name). Specifically, we want to make sure that the test connects to app 1 for PublishOperation streaming, then we can kill app 1 and make sure the execution still succeeds (as it should auto-reconnect to app 2).

**Related issues**: N/A

### DIFF
--- a/server/testutil/testgrpc/testgrpc.go
+++ b/server/testutil/testgrpc/testgrpc.go
@@ -14,21 +14,46 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-// StreamDirector decides how to connect a client request to a backend.
-type StreamDirector func(ctx context.Context, fullMethodName string) (ctxOut context.Context, conn *grpc.ClientConn, err error)
-
 // Proxy is a basic in-process gRPC L7 proxy for use in tests.
+//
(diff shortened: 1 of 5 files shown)


server/testutil/testgrpc/testgrpc.go
```diff
@@ -0,0 +1,86 @@
...
+	"testing"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/mwitkow/grpc-proxy/proxy"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
```
tylerwilliams, line 16: can you add a comment with an example usage of these new methods?

bduffany: Done

vadimberezniker (approved):

PR #4419, opened 2023-07-28
Implement uffd handler to load memory for a firecracker vm

maggie-lou:
**Related issues**: N/A

PR #4428, opened 2023-07-28
Add locking to blockio::Mmap

maggie-lou:
When UFFD is enabled, the uffd handler runs in the background to serve page faults. There is a race condition when (1) the firecracker VM receives a Remove() request and it tries to Close() the blockio store and all its Mmap chunks at the same time as (2) the UFFD handler tries to serve a page fault by allocating memory in a Mmap chunk.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -53,6 +53,7 @@ func (f *File) SizeBytes() (int64, error) {
 // Mmap implements the Store interface using a memory-mapped file. This allows processes to read/write to the file as if it
 // was memory, as opposed to having to interact with it via I/O file operations.
 type Mmap struct {
+	mu     sync.RWMutex
 	data   []byte
 	mapped bool
 	closed bool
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
      * **PR #4428** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4428" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4428?utm_source=stack-comment).

bduffany:
Can we instead orchestrate the COWStore Close() operations and the UFFD Shutdown() operation so that they happen in the correct order? we probably shouldn't call Close() while the UFFD handler is in the process of serving requests.

I really don't want to add locking to the storage layer, since then we'd have no way to know whether we're touching memory/disk that we shouldn't be touching, at the same time that the guest is trying to access it. Today, the race detector flags this to us, but adding a lock around all IO operations would mask the issue.

bduffany:
Also, do you have an invocation link for the race condition? I would've thought that once we Pause the VM, firecracker should guarantee that there are no more memory accesses. Or is this because in `Run()` we don't pause, we're just calling `StopVMM` (which I think just sends SIGTERM to firecracker?)

maggie-lou:
@bduffany Yeah that's fair. I'll update the PR

Here's an [invocation link](https://app.buildbuddy.io/invocation/221ccd74-f056-4b9d-9f0a-c88dd757bce4). It looks like in this particular case, it's happening because the test cleanup calls [Remove().](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go#L324). Regardless I agree that we should check whether UFFD is still handling requests before we Close() the Mmap

PR #4433, opened 2023-07-31
Test gazelle's fetch_repo in firecracker

sluongng:
DO NOT MERGE

Demonstrate that fetch_repo binary could be run without issues in firecracker.

```
> bazel build --config=remote-target-linux test:all
```

Use this to validate Go crashes on FC runner.

**Related issues**: N/A


sluongng:
We have fixed the go crash issue a while ago with correct kernel version.

PR #4435, opened 2023-07-31
runner: refactor container provider

sluongng:
Move the container provider initiations of each isolation type to their
respective packages. This should aid future PR that would update `runner.go`
to support different platforms easier (i.e. Windows).

Also introduce podman flags that alias docker's.

**Related issues**: N/A


sluongng:
@bduffany Sorry for the long delay with this PR.
I could not move the interface up to `interfaces` package as it would create some cyclic dependency situation.
So I decided to simply do it in `runner` instead.

If you prefer to have this PR broken up into smaller PRs for easier review, I could do that as well.

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -71,6 +73,12 @@ var (
 
 	sociArtifactStoreTarget = flag.String("executor.podman.soci_artifact_store_target", "", "The GRPC url to use to access the SociArtifactStore GRPC service.")
 	sociStoreKeychainPort   = flag.Int("executor.podman.soci_store_keychain_port", 1989, "The port on which the soci-store local keychain service is exposed, for sharing credentials for streaming private container images.")
+	podmanNetwork           = flagtypes.Alias[string]("executor.docker_network", "executor.podman_network")
+	podmanCapAdd            = flagtypes.Alias[string]("docker_cap_add", "podman_cap_add")
+	podmanDevices           = flagtypes.Alias[[]container.DockerDeviceMapping]("executor.docker_devices", "executor.podman_devices")
+	podmanVolumes           = flagtypes.Alias[[]string]("executor.docker_volumes", "executor.podman_volumes")
+	podmanRuntime           = flag.String("executor.podman.runtime", "", "Enables running podman with other runtimes, like gVisor (runsc).")
+	podmanEnableStats       = flag.Bool("executor.podman.enable_stats", false, "Whether to enable cgroup-based podman stats.")
```
bduffany, line 81: I feel like it's confusing to have docker flags that affect podman behavior, and podman flags that affect docker behavior... I guess today, the docker flags already affect podman behavior, but I think we should avoid making the situation more confusing, and just continue to reference the `docker` flags for these (might need to use `flagutil.GetValueForFlagName()`)

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -540,22 +526,23 @@ func (r *commandRunner) cleanupCIRunner(ctx context.Context) error {
 	return res.Error
 }
 
-type ContainerProvider func(ctx context.Context, props *platform.Properties, st *repb.ScheduledTask, state *rnpb.RunnerState, workDir string) (*container.TracedCommandContainer, error)
+type ContainerProvider interface {
+	NewContainer(context.Context, *platform.Properties, *repb.ScheduledTask, *rnpb.RunnerState, string) (container.CommandContainer, error)
+}
```
bduffany, line 531: could this go in `enterprise/server/remote_execution/container/container.go` and call it `Provider`? So it's referenced like `container.Provider`

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -540,22 +526,23 @@ func (r *commandRunner) cleanupCIRunner(ctx context.Context) error {
 	return res.Error
 }
 
-type ContainerProvider func(ctx context.Context, props *platform.Properties, st *repb.ScheduledTask, state *rnpb.RunnerState, workDir string) (*container.TracedCommandContainer, error)
+type ContainerProvider interface {
+	NewContainer(context.Context, *platform.Properties, *repb.ScheduledTask, *rnpb.RunnerState, string) (container.CommandContainer, error)
```
bduffany, line 530: nit: maybe call this `New()` to avoid stutter? `containerProvider.New()` vs. `containerProvider.NewContainer()`

bduffany, line 530: nit: there are quite a few args to this func, maybe pack into a struct? (if it's too much though we can always do this later)

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -540,22 +526,23 @@ func (r *commandRunner) cleanupCIRunner(ctx context.Context) error {
...
+}
 
 type PoolOptions struct {
 	// ContainerProvider is an optional implementation overriding
 	// newContainerImpl.
-	ContainerProvider ContainerProvider
+	CtrProvider ContainerProvider
```
bduffany, line 536: the old name `ContainerProvider` seemed better to me - I think generally we avoid abbreviations for public fields

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -629,6 +591,29 @@ func NewPool(env environment.Env, opts *PoolOptions) (*pool, error) {
...
+	}
+	if podmanProvider != nil {
+		providers[platform.PodmanContainerType] = podmanProvider
+	}
+	providers[platform.FirecrackerContainerType] = firecracker.NewProvider(p.env, p.imageCacheAuth, *rootDirectory)
+	providers[platform.SandboxContainerType] = &sandbox.Provider{}
+	providers[platform.BareContainerType] = &bare.Provider{}
+
+	p.containerProviders = providers
+}
```
bduffany, line 615: do you plan on splitting this into OS-specific files in a separate PR? (I think that would let us delete `firecracker_darwin.go` which would be awesome)

enterprise/server/remote_execution/runner/runner_test.go
```diff
@@ -419,18 +419,20 @@ func TestRunnerPool_DefaultSystemBasedLimits_CanAddAtLeastOneRunner(t *testing.T
 	assert.Equal(t, 1, pool.PausedRunnerCount())
 }
 
+type DiskOnlyCtrProvider struct{}
```
bduffany, line 422: nit: carry over the comment here

```suggestion
// Returns containers that only consume disk resources when paused (like firecracker).
type DiskOnlyCtrProvider struct{}
```

enterprise/server/remote_execution/runner/runner.go
sluongng: I will defer it to later, mostly because I am fairly bad at naming things and we already have a bunch of `*Options` structs.

enterprise/server/remote_execution/runner/runner.go
sluongng: Yup, it will be another PR. This change was made so that darwin build could be green meanwhile.

enterprise/server/remote_execution/containers/podman/podman.go
sluongng: This is a bit tricky. Because the initialization order between podman.go and docker.go, I would need to move these values out of `var` block 🤔 

Let me look into it.

sluongng: I removed the package vars and call flagutil when init the provider.

sluongng:
I pushed the OS-specific change to https://github.com/buildbuddy-io/buildbuddy/pull/4382, but it probably needs a rebase after this merged before review-able. Much cleaner though.

enterprise/server/remote_execution/container/container.go
```diff
@@ -57,6 +57,10 @@ type DockerDeviceMapping struct {
 	CgroupPermissions string `yaml:"cgroup_permissions" usage:"cgroup permissions that should be assigned to device."`
 }
 
+type Provider interface {
+	New(context.Context, *platform.Properties, *repb.ScheduledTask, *rnpb.RunnerState, string) (CommandContainer, error)
```
bduffany, line 61:
```suggestion
// Provider constructs new CommandContainer instances
// for a specific isolation type.
//
// This approach is used instead of package-level
// `New` funcs because it allows the executor to avoid
// depending directly on platform-specific container
// implementations.
type Provider interface {
	New(context.Context, *platform.Properties, *repb.ScheduledTask, *rnpb.RunnerState, string) (CommandContainer, error)
```

bduffany (approved):

PR #4436, opened 2023-07-31
Finish handling open requests in UFFD handler before completing Stop()

maggie-lou:
**Related issues**: N/A


maggie-lou:
* master
      * **PR #4436** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4436" target="_blank">Graphite</a>  👈
      * **PR #4428** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4428" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4436?utm_source=stack-comment).

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -113,6 +119,8 @@ func (h *Handler) Start(ctx context.Context, socketPath string, memoryStore *blo
 // the UFFD handler over a unix socket. It also sends GuestRegionUFFDMapping data so the handler knows
 // how to resolve the page faults
 func (h *Handler) receiveSetupMsg(ctx context.Context) (*setupMessage, error) {
+	defer h.lis.Close()
```
maggie-lou, line 122: We can close this immediately. From the documentation: "After Firecracker sends the payload (i.e mem mappings and file descriptor), no other communication happens on the UDS socket (or otherwise) between Firecracker and the page fault handler process."

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -177,8 +188,16 @@ func (h *Handler) handle(ctx context.Context, memoryStore *blockio.COWStore) err
...
+		case <-h.quitChan:
+			return nil
+		default:
+		}
+
 		// Poll UFFD for messages
-		_, pollErr := unix.Poll(pollFDs, -1)
+		// Poll blocks until a message is ready to be read. A timeout is necessary so that if the handler receives
+		// a shutdown message on quitChan while Poll is waiting, Poll will exit and restart the loop to check quitChan
+		_, pollErr := unix.Poll(pollFDs, 500 /* timeout in ms */)
```
bduffany, line 200: This timeout means that `Remove()` can take 500ms which seems a bit too long if we can avoid it.

It looks like a pattern for "canceling" a poll is to create a file descriptor that you control (in golang you can use `os.Pipe()` to get read/write ends of a pipe and call `Fd()` to get the file descriptor. Then you can add the output end of the pipe to the `pollFDs` list, and to cancel, write something to the input end of the pipe.

See https://stackoverflow.com/a/4049050

maggie-lou: Ah very clever! Updated

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -162,14 +178,21 @@ func (h *Handler) receiveSetupMsg(ctx context.Context) (*setupMessage, error) {
 }
 
 func (h *Handler) handle(ctx context.Context, memoryStore *blockio.COWStore) error {
+	h.wg.Add(1)
+	defer h.wg.Done()
```
bduffany, line 182: nit: this works, but instead of using a wait group it would be slightly simpler / more idiomatic to use a `quit` channel

```go
defer close(h.quit)
```

then in `Stop()`:

```go
<-h.quit
```

I realized that in `setupUFFDHandler` we have a bug though -

```go
	h, err := uffd.NewHandler()
	if err != nil {
		return status.WrapError(err, "create uffd handler")
	}
	c.uffdHandler = h
	sockAbsPath := filepath.Join(c.getChroot(), uffdSockName)
	if err := h.Start(ctx, sockAbsPath, c.memoryStore); err != nil {
		return status.WrapError(err, "start uffd handler")
	}
	return nil
```

we should move `c.uffdHandler = h` to after the `Start()` block, so that we don't call `c.uffdHandler.Stop()` in `Remove()` unless the handler was started.

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -267,8 +300,16 @@ func pageStartAddress(addr uint64, pageSize int) uintptr {
...
-	h.lis.Close()
+func (h *Handler) Stop() error {
+	log.Info("UFFD handler beginning shut down")
+	_, err := h.earlyTerminationWriter.Write([]byte{0})
+	if err != nil {
+		return status.WrapError(err, "write to early terminator")
+	}
+	h.wg.Wait()
+	h.earlyTerminationReader.Close()
+	h.earlyTerminationWriter.Close()
```
bduffany, line 311: sorry missed this - I think the `h.wg.Wait()` needs to go after the Close() here, since the Close() is what awakens the uffd poll?

maggie-lou: h.earlyTerminationWriter.Write() is what awakens the uffd poll.

io.Pipe.Write() writes to the pipe immediately - it does not wait for Close() to flush the data. I put the h.wg.Wait() after because I didn't want to close the earlyTerminationReader before it could read the early termination message

enterprise/server/remote_execution/uffd/uffd.go
maggie-lou: Modified to use a quit channel and moved setting the uffdHandler

Also added an extra check to Stop() so that it won't break if we do happen to call it without calling Start() first

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -267,8 +301,27 @@ func pageStartAddress(addr uint64, pageSize int) uintptr {
 	return uintptr(addr & ^(uint64(pageSize) - 1))
 }
 
-func (h *Handler) Stop() {
-	h.lis.Close()
+func (h *Handler) Stop() error {
+	if h.earlyTerminationWriter == nil && h.earlyTerminationReader == nil && h.quitChan == nil {
```
bduffany, line 305:
```suggestion
	if h.earlyTerminationWriter == nil || h.earlyTerminationReader == nil || h.quitChan == nil {
```

bduffany (approved):

PR #4439, opened 2023-08-01
Fix anon group sampling.

vadimberezniker:
We need to use the stored anon group ID when doing sample, not the logical "ANON" group ID.

**Related issues**: N/A


tylerwilliams (approved):
Ahhhh

PR #4441, opened 2023-08-01
Check if UsageStats is nil before referencing in rbeperf

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/tools/rbeperf/rbeperf.go
+++ b/enterprise/tools/rbeperf/rbeperf.go
@@ -702,14 +702,12 @@ func (g *runner) run(ctx context.Context) error {
 		remoteOutputUploadHist.Add(remoteStats.uploadOutputsDuration.Milliseconds())
 		remoteTotalHist.Add(remoteStats.totalDuration.Milliseconds())
 
-		if res.RemoteStats.UsageStats == nil {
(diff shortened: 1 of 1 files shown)


enterprise/tools/rbeperf/rbeperf.go
```diff
@@ -702,6 +702,12 @@ func (g *runner) run(ctx context.Context) error {
 		remoteOutputUploadHist.Add(remoteStats.uploadOutputsDuration.Milliseconds())
 		remoteTotalHist.Add(remoteStats.totalDuration.Milliseconds())
 
+		if res.RemoteStats.UsageStats == nil {
+			log.Warningf("Missing remote usage stats for %q", res.CommandName)
```
bduffany, line 706: This will happen pretty often because we use polling-based stats recording, and the first poll attempt might happen just after the task completes. Not sure how often these are happening but maybe we don't want to treat this as an error

bduffany (approved):
LGTM - unrelated to this PR (and not sure if you're already thinking about this), but we probably want to somehow incorporate the executor's total CPU usage, and not just the aggregate measured UsageStats. The measured UsageStats will always under-report because they don't include podman overhead / workspace setup overhead etc. Instead, we could probably report the CPU per task as `total executor node CPU-seconds used / num tasks executed` for a given time slice.

PR #4444, opened 2023-08-01
Add gRPC origin header

bduffany:
This will apply to all `executor -> app` requests (as long as the dial is done with `grpc_client.DialTarget()`). The executor will also set an env var to set the origin header for both the "outer" workflow invocation BES stream request as well as the inner one, along with any cache requests on the inner invocation.

**Related issues**: N/A


server/rpc/interceptors/interceptors.go
```diff
@@ -46,6 +47,10 @@ func init() {
 	}
 }
 
+func ClientOrigin() string {
```
tylerwilliams, line 50: can you comment this?

tylerwilliams (approved):

PR #4445, opened 2023-08-01
Add internal API for fetching target-level invocation events

siggisim:
**Related issues**: N/A


proto/target.proto
```diff
@@ -137,3 +137,41 @@ message GetTargetResponse {
   // The pagination token to retrieve the next page of results.
   string next_page_token = 4;
 }
+
+message Artifact {
+  string name = 1;
+  string bytestream_url = 2;
+}
+
+message Target {
```
bduffany, line 146: think we might also need actionExecuted here (stderr etc for failed tests)

siggisim: i think we could just leave this in the GetInvocation response the way we get it now

it's hard to match to a target, and according to the BES proto we'll only get "important ones". i've only ever seen one in a response

proto/target.proto
```diff
@@ -137,3 +137,41 @@ message GetTargetResponse {
...
+}
+
+message GetTargetRequest {
+  // Optional, if set only targets with this label will be returned. 
+  // If not set, all targets will be returned.
+  string target_label = 1;
+
+  // Optional, if set only targets with this status will be returned.
+  // If not set, all targets will be returned.
+  api.v1.Status status = 2;
```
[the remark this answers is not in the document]
siggisim: i'd like to see if we can get away without it - ideally the UI just shows the TestGroups it gets from the server each in a card

bduffany (approved):

PR #4446, opened 2023-08-01
Stop the UFFD handler from running in the background when the VM is paused

maggie-lou:
Stop the UFFD handler from running in the background when the VM is paused so it isn't wasting resources. Also add support to the UFFD handler so that it can be stopped and restarted using the same uffd object

This also fixes a race condition. For some reason, there can still be page faults when the VM is paused. I wonder if there is some lightweight process that continues to run on the paused VM to listen for API requests or something

The race condition occurs when the background UFFD thread tries to serve a page fault and read from a blockio chunk. At the same time, in MergeDiffSnapshot we are creating copies of dirty chunks and unmapping the old chunks. The race occurs when we try to read from a chunk that is simultaneously being unmapped

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/BUILD
+++ b/enterprise/server/remote_execution/containers/firecracker/BUILD
@@ -136,6 +136,7 @@ go_test(
     tags = [
         "bare",  # Firecracker tests must be run with bare execution so they aren't nested within another container
         # TODO: enable on CI once we fix race conditions and timeouts
+        "manual",
         "no-sandbox",  # Firecracker is not compatible with Bazel's sandbox environment
     ],
     target_compatible_with = [
--- a/enterprise/server/remote_execution/containers/firecracker/BUILD
+++ b/enterprise/server/remote_execution/containers/firecracker/BUILD
@@ -136,7 +136,6 @@ go_test(
     tags = [
         "bare",  # Firecracker tests must be run with bare execution so they aren't nested within another container
         # TODO: enable on CI once we fix race conditions and timeouts
-        "manual",
         "no-sandbox",  # Firecracker is not compatible with Bazel's sandbox environment
     ],
     target_compatible_with = [
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -586,6 +586,13 @@ func (c *FirecrackerContainer) pauseVM(ctx context.Context) error {
 		log.CtxErrorf(ctx, "Error pausing VM: %s", err)
 		return err
 	}
+
+	if c.uffdHandler != nil {
+		if err := c.uffdHandler.Stop(); err != nil {
+			return status.WrapError(err, "failed to stop uffd handler")
+		}
+	}
+
 	// Now that we've paused the VM, it's a good time to Sync the NBD backing
 	// files. This is particularly important when the files are backed with an
 	// mmap. The File backing the mmap may differ from the in-memory contents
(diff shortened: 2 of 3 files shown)


maggie-lou:
* master
      * **PR #4436** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4436" target="_blank">Graphite</a> 
        * **PR #4446** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4446" target="_blank">Graphite</a>  👈
      * **PR #4428** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4428" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4446?utm_source=stack-comment).

bduffany (commented):
in `Pause()`, we call `Remove()` which should already stop the UFFD handler, so I'm not sure this PR is needed?

(I know this is confusing - I have gotten bitten by this before :sweat_smile: )

maggie-lou:
> in `FirecrackerContainer::Pause()`, we call `FirecrackerContainer::Remove()` which should already stop the UFFD handler / set it to nil, so I'm not sure this PR is needed?
> 
> (I know this is confusing - I have gotten bitten by this before 😅 )

It sucks because the ordering goes pauseVM(), CreateSnapshot(), mergeDIffSnapshots(), remove(). The UFFD handler needs to be running for CreateSnapshot, but there's a race condition if it runs at the same time as mergeDiffSnapshots.

bduffany (commented):
>  the ordering goes pauseVM(), CreateSnapshot(), mergeDIffSnapshots(), remove()

Ahh, makes sense.

> For some reason, there can still be page faults when the VM is paused. I wonder if there is some lightweight process that continues to run on the paused VM to listen for API requests or something

By stopping the UFFD handler, are we possibly causing issues for whatever it is that's waiting for the fault to be resolved? I'm a little worried that this will bite us later as a mystifying bug that takes a long time to trace back to this issue :sweat_smile: 

Could it just be the VM itself accessing some memory, but we don't finish responding to the request by the time the fault is resolved? I wonder if firecracker somehow re-triggers these faults when the VM wakes up :thinking:

maggie-lou:
> > the ordering goes pauseVM(), CreateSnapshot(), mergeDIffSnapshots(), remove()
> 
> Ahh, makes sense.
> 
> > For some reason, there can still be page faults when the VM is paused. I wonder if there is some lightweight process that continues to run on the paused VM to listen for API requests or something
> 
> By stopping the UFFD handler, are we possibly causing issues for whatever it is that's waiting for the fault to be resolved? I'm a little worried that this will bite us later as a mystifying bug that takes a long time to trace back to this issue 😅
> 
> Could it just be the VM itself accessing some memory, but we don't finish responding to the request by the time the fault is resolved? I tests still pass etc., I wonder if firecracker is somehow re-triggering these unresolved faults when the VM wakes up 🤔

Yeah... I am a little nervous about those unknown page faults too. The good news is that we always call Remove() when pausing a VM and LoadSnapshot() when unpausing. LoadSnapshot will listen for a new uffd object from firecracker. So even if the original VM state gets corrupted, it should be okay?

(Side note: This is reminding me that we might a UFFD.Pause() vs UFFD.Stop(), where the Stop() actually closes the uffd file descriptor because right now it's never getting closed)

To be safer, I guess we could move the uffd.Stop() until after CreateSnapshot(), to guarantee that the VM is healthy up until the point where we've created a snapshot. After that point we can always reliably use the snapshot to load up a new healthy VM. I put it in pauseVM() because that seemed like a natural point to stop the UFFD handler, and if that function is reused we can guarantee that the UFFD handler won't cause more problems in the background, whereas you'd need a lot more context to understand why uffd.Stop() is after CreateSnapshot()

Another hypothesis I had for why there'd continue to be page faults was if the host machine re-allocated those virtual memory addresses to another process while the firecracker VM was paused. But from my research it seems like the guest machine continues to hold onto the memory

maggie-lou:
Replaced with this PR: https://github.com/buildbuddy-io/buildbuddy/pull/4505

PR #4449, opened 2023-08-02
fastcopy: make darwin implementation cgo free

sluongng:
Switch fastcopy implementation on Darwin from custom-CGO to make use of
the API in golang.org/x/sys/unix package.

References: https://github.com/containerd/continuity/blob/25762ef7c27a37645197ed6d1d708e55a230e35d/fs/copy_darwin.go#L27

**Related issues**: N/A


bduffany (commented):
do you mind also adding tests for this?

- dest should have same contents as original
- mutating dest should not affect the original (on darwin only for now)
- 'already exists' errors should be ignored (tbh we probably shouldn't ignore these but I guess our current code probably depends on it somehow :P )

sluongng:
@bduffany good call. I added the tests. PTAL 🙏

bduffany (approved):

PR #4450, opened 2023-08-02
Add logging for executions missing invocation ID

bduffany:
We think that https://github.com/buildbuddy-io/buildbuddy-internal/issues/2533 is happening because the execution is missing an invocation ID. We depend on invocation ID in order to get failed action results for a particular invocation.

This will let us know whether the request metadata is missing entirely or if bazel possibly has a bug where it sets the metadata but just doesn't set the invocation ID in some cases, or if we have a bug somehow in the way that we're extracting the invocation ID.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2533


sluongng (approved):
LGTM.

PR #4456, opened 2023-08-02
Transparently retry PublishOperation streams

bduffany:
The executor currently has 2 streams to the app that it maintains for each task, each of which can be broken when the apps roll out and cause a task retry:
- LeaseTask
- PublishOperation

This PR allows us to transparently retry PublishOperation, which should roughly cut in half the number of times that a task has to be retried because of an app rollout / disconnect.

In a later PR we can retry LeaseTask as well, though this is more complex.

---

Description of the changes in this PR:
- There are two stream methods in a `PublishOperation` stream client: `Send()` publishes a status update, and `CloseAndRecv()` terminates the stream and waits for the server to ack that it has received all messages.
- Note, `Send()` does not guarantee that the server has ack'd the message. So if `Send()` fails, it's possible that _multiple_ messages actually failed to send.
- Luckily though, only the _last sent_ message matters, since newer messages effectively replace the execution status of older messages. e.g. if we drop both the IN_PROGRESS and COMPLETE messages, we only need to re-publish the COMPLETE message, since that reflects the current status of the task and the IN_PROGRESS update is no longer relevant.
- When sending a message, we buffer the last message sent, so that we can re-send it if needed.
- We re-send the buffered message if either Send() or CloseAndRecv() fails. It's especially important to re-send if CloseAndRecv() fails, because CloseAndRecv() succeeding is what guarantees that the server has acked the message. (`Send()` does not guarantee that the server received the message.)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/executor/executor.go
+++ b/enterprise/server/remote_execution/executor/executor.go
@@ -2,6 +2,7 @@ package executor
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"strings"
 	"syscall"
@@ -30,15 +31,17 @@ import (
 	rspb "github.com/buildbuddy-io/buildbuddy/proto/resource"
 )
 
-const (
+var (
 	// Messages are typically sent back to the client on state changes.
 	// During very long build steps (running large tests, linking large
 	// objects, etc) no progress can be returned for a very long time.
 	// To ensure we keep the connection alive, we start a timer and
 	// just repeat the last state change message after every
 	// execProgressCallbackPeriod. If this is set to 0, it is disabled.
-	execProgressCallbackPeriod = 60 * time.Second
+	execProgressCallbackPeriod = flag.Duration("executor.task_progress_publish_interval", 60*time.Second, "How often tasks should publish progress updates to the app.")
+)
 
+const (
 	// 7 days? Forever. This is the duration returned when no max duration
 	// has been set in the config and no timeout was set in the client
 	// request. It's basically the same as "no-timeout".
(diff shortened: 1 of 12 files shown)


vadimberezniker:
> I think there might be a race condition here:
> 
> * Stream 1 publishes IN_PROGRESS message.
> * Stream 1 is broken due to a transient network error.
> * Executor detects that stream 1 is broken. It retries, starting a new PublishOperation stream (stream 2).
> * Due to retry logic, executor re-publishes the IN_PROGRESS message on stream 2.
> * Shortly after, the task completes. Executor then publishes COMPLETED on stream 2.
> * Server receives {IN_PROGRESS, COMPLETED} from stream 2, and marks the task complete.
> * Server has not yet detected that stream 1 is broken. It receives the original {IN_PROGRESS} message from stream 1, and marks the task as IN_PROGRESS.
> * The task is now stuck in IN_PROGRESS state.
> 
> Will see if I can reproduce and think about a possible fix.

At least for the DB, we ignore updates if the task is already marked completed:
```
	dbErr := s.env.GetDBHandle().TransactionWithOptions(ctx, db.Opts().WithQueryName("upsert_execution"), func(tx *db.DB) error {
		var existing tables.Execution
		if err := tx.Where("execution_id = ?", executionID).First(&existing).Error; err != nil {
			return err
		}
		return tx.Model(&existing).Where("execution_id = ? AND stage != ?", executionID, repb.ExecutionStage_COMPLETED).Updates(execution).Error
	})
```

tylerwilliams (approved):

bduffany:
There's some weird interaction between the new `AppShutdown` test and `TestSaturateTaskQueue` which causes the app shutdown test to fail. It doesn't reproduce remotely because those tests are run in different shards, but it reproduces locally when running with `--test_output=streamed`. Looking into it...

PR #4459, opened 2023-08-02
Add support for communicating with firecracker dockerd over tcp

siggisim:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/goinit/main.go
+++ b/enterprise/server/cmd/goinit/main.go
@@ -159,7 +159,7 @@ func startDockerd(ctx context.Context, enableTCP bool) error {
 
 	args := []string{}
 	if enableTCP {
-		args = append(args, "-H", "tcp://0.0.0.0:2375")
+		args = append(args, "-H", "fd://", "-H", "tcp://0.0.0.0:2375")
 	}
 
 	cmd := exec.CommandContext(ctx, "dockerd", args...)
--- a/enterprise/server/cmd/goinit/main.go
+++ b/enterprise/server/cmd/goinit/main.go
@@ -146,7 +146,7 @@ func copyFile(src, dest string, mode os.FileMode) error {
 	return nil
 }
 
-func startDockerd(ctx context.Context, enableTCP bool) error {
+func startDockerd(ctx context.Context) error {
 	// Make sure we can locate both docker and dockerd.
 	if _, err := exec.LookPath("docker"); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


enterprise/server/cmd/goinit/main.go
```diff
@@ -156,7 +157,12 @@ func startDockerd(ctx context.Context) error {
 
 	log.Infof("Starting dockerd")
 
-	cmd := exec.CommandContext(ctx, "dockerd")
+	args := []string{}
+	if enableTCP {
```
bduffany, line 161: in `waitForDockerd` below, should probably also check `docker ps` using the TCP port if enabled, otherwise the unix socket might be ready but the TCP port might not be.

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -769,6 +769,110 @@ func TestFirecrackerRunWithDocker(t *testing.T) {
...
+
+	path := filepath.Join(workDir, "world.txt")
+	if err := os.WriteFile(path, []byte("world"), 0660); err != nil {
+		t.Fatal(err)
+	}
+	cmd := &repb.Command{
+		Arguments: []string{"bash", "-c", `
+			set -e
+			# Discard pull output to make the output deterministic
+			docker -H tcp://0.0.0.0:2375 pull ` + busyboxImage + ` &>/dev/null
```
bduffany, line 790: if this doesn't work, might need 127.0.0.1 here - IIRC 0.0.0.0 only works as a serving address (meaning "any IP"), but could be wrong

bduffany (approved):

enterprise/server/cmd/goinit/main.go
siggisim: Done

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
siggisim: Done, but I still can't seem to get these tests pass - I've tried a bunch of things but they always seem to fail when setting things up (before we even get to this part) with:
```
firecracker_test.go:820: rpc error: code = Internal desc = failed to sync workspace: Firecracker VM crashed: rpc error: code = DeadlineExceeded desc = docker init timed out after 30s
```

Example: https://app.buildbuddy.io/invocation/57e4a583-bcdf-43df-a898-a5857075da56

Kind of at my wit's end on what else to try :(

bduffany: the VM logs might have some interesting dockerd output - you can get them with `--config=firecracker-debug`

siggisim: Thanks to your suggestion, setting `--tls=false` seemed to do the trick!

tylerwilliams (approved):

bduffany (approved):

PR #4460, opened 2023-08-02
Create Iterator after Lock

luluz66:
**Related issues**: N/A
We should create the iterator after we lock the key. Otherwise, when a goroutine
1 tries to write a entry to pebble and another goroutine tries to read the same
entry. If the second goroutine obtains the iterator before the first go routine
finish writes, even when the second goroutine reads after the first goroutine
finished writing, it won't see the entry.

In https://github.com/buildbuddy-io/buildbuddy/pull/4337, I added some cases In
TestCompression_ParallelRequests where we are writing and reading the same
resource in different goroutines (i.e. when the testSize is small). This test 
become flakey due to this issue. After I applied this fix, the test is de-flaked.


vadimberezniker (approved):
Nice find!

tylerwilliams (approved):
Nice find, thanks for fixing this!

PR #4464, opened 2023-08-02
Fix potential data race in ExecuteWorkflow

bduffany:
**Related issues**: N/A


tylerwilliams (approved):

maggie-lou (approved):

enterprise/server/workflow/service/service.go
```diff
@@ -549,17 +549,17 @@ func (ws *workflowService) ExecuteWorkflow(ctx context.Context, req *wfpb.Execut
 	for actionName, action := range actions {
 		action := action
 		actionName := actionName
-		wg.Add(1)
+		actionStatus := &wfpb.ExecuteWorkflowResponse_ActionStatus{
+			ActionName: actionName,
+		}
+		actionStatuses = append(actionStatuses, actionStatus)
```
vadimberezniker, line 555: It still feels a little iffy to me that you are reading and writing state across goroutines w/o locking. I would throw a mutex around it.

PR #4468, opened 2023-08-03
app/terminal: make search result count white

sluongng:
The terminal is black so currently the search result count inherits it's
color from the parent.  Make the text white so it's readable.

**Related issues**: N/A

### DIFF
--- a/app/terminal/terminal.css
+++ b/app/terminal/terminal.css
@@ -128,6 +128,13 @@
   align-items: center;
   padding: 0 8px;
   color: white;
+}
+
+.terminal.light-terminal .search-result-count {
+  color: black;
+}
+
+.terminal .no-results {
   opacity: 0.3;
 }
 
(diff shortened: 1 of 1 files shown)


sluongng:
Current
image
image


After
image
image

bduffany:
what browser are you using? on Chrome it shows up white already

image

sluongng:
strange... I am using chrome 🤔

app/terminal/terminal.css
```diff
@@ -127,9 +127,7 @@
   display: flex;
   align-items: center;
   padding: 0 8px;
-}
-
-.terminal .no-results {
+  color: white;
   opacity: 0.3;
 }
```
bduffany, line 132: - Keep `opacity: 0.3` within `.no-results` (i.e. only when "No results" is displayed), otherwise the text is too light when displaying "1 of 3" etc.
- For light terminal theme, text should be black

```suggestion
  color: white;
}

.terminal.light-terminal .search-result-count {
  color: black;
}

.terminal .no-results {
  opacity: 0.3;
}
```

bduffany (approved):

PR #4480, opened 2023-08-04
Remove unused write_token auth.

vadimberezniker:
**Related issues**: N/A


tylerwilliams (approved):

bduffany:
LGTM, might be worth checking with Siggi though in case he is aware of anyone that relies on basic auth (for on-prem etc.). I think that was the concern when I tried to remove this before.

bduffany (approved):

siggisim:
> LGTM, might be worth checking with Siggi though in case he is aware of anyone that relies on basic auth (for on-prem etc.). I think that was the concern when I tried to remove this before.

I think this write token is old enough that nobody is using it anymore - if they are, I'm happy to tell them to switch to API keys.

PR #4483, opened 2023-08-04
Fix `bb fix -mode diff`

bduffany:
`bb fix -mode diff` fails with "failed to parse startup options: invalid options syntax: -mode"

**Related issues**: N/A


sluongng (approved):
LGTM

PR #4484, opened 2023-08-04
gazelle: fix missing lint

sluongng:
**Related issues**: N/A


bduffany (commented):
This is fixed in https://github.com/buildbuddy-io/buildbuddy/pull/4482

bduffany (approved):
Actually the fix in this PR is probably more "idiomatic" (marking with `&& !android`) so let's submit this one instead of my fix :)

PR #4490, opened 2023-08-04
Add initial GetTarget implementation

bduffany:
Making the minimal changes needed to serve only the first page of targets to the UI.

Subsequent PRs:
- Don't buffer all events into the Invocation, and instead stream directly into EventIndex
- Have the UI render the returned TargetGroups
- Have the UI use GetTarget to fetch subsequent data pages for target listings
- Have the UI use the GetTarget endpoint to fetch data for the target page
- Add tests

**Related issues**: N/A


siggisim:
proto changes look reasonable to me here, I'll let @tylerwilliams take a look at the go code

server/build_event_protocol/event_index/event_index.go
```diff
@@ -0,0 +1,127 @@
...
+	TargetsByStatus            map[cmnpb.Status][]*trpb.Target
+	NamedSetOfFilesByID        map[string]*bespb.NamedSetOfFiles
+	// Events which aren't indexed by target and are instead returned in the
+	// top-level invocation proto.
+	TopLevelEvents []*inpb.InvocationEvent
+
+	rootCauseLabels map[string]bool
+}
+
+func Build(in *inpb.Invocation) *Index {
```
tylerwilliams, line 30: I don't think we want this API to take a full invocation because that implies we're buffering a full invocation in memory, which we want to stop doing.

Instead can we parse events as we read them from blobstore (or from the invocation stream, if you're accumulating these eventually)?

Like, the caller of event_index can use:
```
func streamRawInvocationEvents(env environment.Env, ctx context.Context, streamID string, callback invocationEventCB) error
```

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -105,9 +107,36 @@ func (s *BuildBuddyServer) GetInvocation(ctx context.Context, req *inpb.GetInvoc
 		return nil, err
 	}
 
+	if *paginateInvocations {
```
tylerwilliams, line 110: if paginateInvocations is set, we should not call `LookupInvocation` anymore

server/target/target.go
```diff
@@ -36,7 +43,18 @@ const (
 	coverageCommand = "coverage"
 
 	// The number of distinct commits returned in GetTargetHistoryResponse.
-	targetPageSize = 20
+	targetHistoryPageSize = 20
+
+	// The number of targets returned in the GetTarget response.
```
tylerwilliams, line 48: nit: in a single page of the GetTarget response.

server/target/target.go
```diff
@@ -36,7 +43,18 @@ const (
...
+	targetPageSize = 12
+
+	// When returning a paginated list of all targets in an invocation with
+	// files expanded, stop returning targets after this many files have been
+	// expanded. This is a "soft" limit because the file list is not truncated,
+	// only the target list. So for example if a target contains 101 files, all
+	// 101 files will be returned, but only that target will be returned in the
+	// target page.
+	targetFilesSoftLimit = 100
```
tylerwilliams, line 57: Do we need this?

tylerwilliams (changes requested):

server/build_event_protocol/event_index/event_index.go
bduffany: I plan to do this in a follow-up PR, since it will involve tinkering with the LookupInvocation API in a  way that I want to think through carefully.

server/buildbuddy_server/buildbuddy_server.go
bduffany: that's my plan with the TODO - just want to get this smaller PR submitted first

proto/target.proto
```diff
@@ -165,7 +168,7 @@ message TargetGroup {
   // Targets in the group. If this is empty but next_page_token is set, this
   // means that the invocation is still in progress and that more targets may be
   // fetched later.
-  repeated Target target = 2;
+  repeated Target targets = 2;
```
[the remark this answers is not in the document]
bduffany: I added this field in the last PR and meant to call it `targets`, this is more of a typo fix.

server/target/target.go
bduffany: Done

server/target/target.go
bduffany: I think so - some targets can have a ton of files and some can have barely any. We should ideally paginate by number of files directly but the GetTarget API doesn't make that very easy, currently. We can maybe add a separate GetFiles endpoint later

bduffany (commented):
(woops, replied to the earlier comments too soon)

tylerwilliams (approved):
LGTM (thanks for the changes)

PR #4494, opened 2023-08-07
config: fix FOSS local config not setting up cache

sluongng:
Ensure that the local disk cache is used when we run the FOSS server
locally with `bazel run server`.

**Related issues**: N/A

### DIFF
--- a/config/buildbuddy.local.yaml
+++ b/config/buildbuddy.local.yaml
@@ -2,6 +2,11 @@ app:
   log_error_stack_traces: true
 storage:
   ttl_seconds: 86400 # One day in seconds.
+  disk:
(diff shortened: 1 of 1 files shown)


bduffany (approved):

tylerwilliams:
Does this affect the docker image too and have we tested that? (I want to avoid having the docker image run with a confusing / easily broken configuration. If we're only after fixing bazel run, another approach is to add args params to that go_binary target)

sluongng:
@tylerwilliams the container image uses `config/buildbuddy.release.yaml` which is not affected by this (it has `root_directory` defined).

This local.yaml file is only used when we run the server locally with `bazel run server`.

PR #4505, opened 2023-08-07
[UFFD] Ensure firecracker VM and UFFD handler are shut down before saving snapshots to the cache

maggie-lou:
Fix a race condition where the UFFD handler is still trying to access the snapshots in the background at the same time that we're trying to merge them and save them to the cache

Also inline most of the logic from SaveSnapshot() into Pause(), so it's more clear that we're doing a lot more than just saving a snapshot (i.e. we're shutting down the VM and cleaning up resources).

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -603,17 +603,7 @@ func (c *FirecrackerContainer) pauseVM(ctx context.Context) error {
 	return nil
 }
 
-// SaveSnapshot pauses the VM and takes a snapshot, saving the snapshot to
-// cache.
-//
-// If the VM is resumed after this point, the VM state and memory snapshots
-// immediately become invalid and another call to SaveSnapshot is required. This
-// is because we don't save a copy of the disk when taking a snapshot, and
-// instead reuse the same disk image across snapshots (for performance reasons).
-// Resuming the VM may change the disk state, causing the original VM state to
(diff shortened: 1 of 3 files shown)


maggie-lou:
* master
  * **PR #4505** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4505" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4505?utm_source=stack-comment).

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1934,15 +1884,106 @@ func (c *FirecrackerContainer) Pause(ctx context.Context) error {
...
+
+	// Finish cleaning up VM resources
+	if err = c.Remove(ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *FirecrackerContainer) snapshotDetails(ctx context.Context) (snapshotType string, memSnapshotName string, baseSnapshotUnpackDir string, err error) {
```
bduffany, line 1936: uber nit: it might be cleaner to pack up the 3 return values into a struct, so it could just return `(*snapshotDetails, error)`

```suggestion
func (c *FirecrackerContainer) snapshotDetails(ctx context.Context) (snapshotType string, memSnapshotName string, baseSnapshotUnpackDir string, err error) {
```

bduffany (approved):

PR #4509, opened 2023-08-08
Bump default resource allocation for workflows

bduffany:
Our workflows feel a bit sluggish currently and have been occasionally hitting OOM errors. Bumping from 3CPU/7.5GB mem => 4CPU/10GB mem should improve the overall experience and also might help alleviate a bit of pressure on the workflow executors that we suspect is caused by running too many VMs concurrently.

**Related issues**: N/A

### DIFF
--- a/docs/workflows-config.md
+++ b/docs/workflows-config.md
@@ -179,7 +179,9 @@ build or fetch these dependencies.
 
 :::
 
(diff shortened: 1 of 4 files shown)


tylerwilliams:
Can we close this one for now?

PR #4512, opened 2023-08-08
Render TargetGroups in the UI

bduffany:
- Render TargetGroups as cards in the UI
- Fix a few server-side bugs
- Increase the build event limit when rendering paginated invocations.

Note: this does not yet render the artifacts card or the target page. That'll be a separate PR.

**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -1321,7 +1321,15 @@ func streamRawInvocationEvents(env environment.Env, ctx context.Context, streamI
 // in memory.
 func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*inpb.Invocation, error) {
 	var events []*inpb.InvocationEvent
-	inv, err := LookupInvocationWithCallback(ctx, env, iid, maxEventCount, func(event *inpb.InvocationEvent) error {
+	inv, err := LookupInvocationWithCallback(ctx, env, iid, func(event *inpb.InvocationEvent) error {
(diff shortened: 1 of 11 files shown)


siggisim (commented):
Font-end stuff looks reasonable to me. Will let @tylerwilliams take a look at the backend.

Is there an easy way to turn this on/off if we find out it's busted in some way?

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -1321,7 +1321,7 @@ func streamRawInvocationEvents(env environment.Env, ctx context.Context, streamI
 // in memory.
 func LookupInvocation(env environment.Env, ctx context.Context, iid string) (*inpb.Invocation, error) {
 	var events []*inpb.InvocationEvent
-	inv, err := LookupInvocationWithCallback(ctx, env, iid, func(event *inpb.InvocationEvent) error {
+	inv, err := LookupInvocationWithCallback(ctx, env, iid, maxEventCount, func(event *inpb.InvocationEvent) error {
```
tylerwilliams, line 1324: a possibly nicer / more idiomatic way to do this would be to have your callback return a bool indicating if you want to continue. Then track maxEventCount in your callback

tylerwilliams (changes requested):

bduffany:
> Is there an easy way to turn this on/off if we find out it's busted in some way?

This is all behind `--app.paginate_invocations=true` - the UI only looks at whether `target_groups` is nonempty but that'll only be the case if the flag is set

tylerwilliams (approved):

PR #4514, opened 2023-08-08
Make APIKeys nonce column unique by default.

vadimberezniker:
Adding the unique index conditionally confuses GORM auto-migrator causing it to issue a no-op alter column statement.

To get around this, backfill the nonce column for all users before the gorm migrator runs so that it can add the unique index.

**Related issues**: N/A

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -350,7 +350,7 @@ type APIKey struct {
 	// The API key token used for authentication.
 	Value          string `gorm:"default:NULL;uniqueIndex:api_key_value_composite_index;index:api_key_unencrypted_value_index;"`
 	EncryptedValue string `gorm:"default:'';uniqueIndex:api_key_value_composite_index;index:api_key_encrypted_value_index;"`
-	Nonce          string `gorm:"default:NULL;"`
+	Nonce          string `gorm:"default:NULL;uniqueIndex:api_key_nonce_index"`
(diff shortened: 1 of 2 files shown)


server/tables/tables.go
```diff
@@ -834,6 +834,26 @@ func PreAutoMigrate(db *gorm.DB) ([]PostAutoMigrateLogic, error) {
...
+		// gorm adds the unique index.
+		if !m.HasColumn(&APIKey{}, "nonce") {
+			stmt := `ALTER TABLE "APIKeys" ADD COLUMN "nonce" varchar(255) DEFAULT NULL`
+			if err := db.Exec(stmt).Error; err != nil {
+				return nil, err
+			}
+		}
+		backfill := `
+			UPDATE "APIKeys"
+			SET nonce = SUBSTR(value, 1, 6)
```
maggie-lou, line 849: Is the idea that we're filling 'nonce' with arbitrary unique data that will later get overwritten by backfillUnencryptedKeys()?

vadimberezniker: We use the first 6 characters of the key as the nonce, the only difference is that we were previously only populating this column when encryption was enabled. I've also removed the redundant update in backfillUnencryptedKeys.

maggie-lou (approved):

tylerwilliams (approved):

PR #4515, opened 2023-08-09
.bazelrc: switch to use common

sluongng:
Recent Bazel release let us use `common` to apply flags to all
applicable subcommands. Make use of this to help ensure `query` or
`fetch` would use the same flags as `build`

**Related issues**: N/A

### DIFF
--- a/template-user.bazelrc
+++ b/template-user.bazelrc
@@ -6,7 +6,7 @@
 ################
 # Windows only #
 ################
-# startup --output_user_root=D:/tmp
+# common --output_user_root=D:/tmp
 
 ########################
 ## API key based auth ##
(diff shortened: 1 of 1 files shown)


sluongng:
It seems like we gona need to adjust `cli/parser/parser.go` to the new `common` logic

.bazelrc
```diff
@@ -1,175 +1,175 @@
...
-#   - Tell Bazel to use Linux-compatible toolchains to build
-build:target-linux --cpu=k8
-build:target-linux --host_cpu=k8
+#   - Tell Bazel to common binaries for linux
+#   - Tell Bazel to use Linux-compatible toolchains to common
+common:target-linux --cpu=k8
+common:target-linux --host_cpu=k8
 
-# Build with --config=remote-target-linux* to build on remote Linux Executors.
+# Build with --config=remote-target-linux* to common on remote Linux Executors.
```
sam-bristow, line 60: Looks like the find/replace got a bit carries away in a few places...

sluongng:
The parsing logic has been solved in #5263 so we should be able to switch everything to `common`

bduffany (approved):

PR #4518, opened 2023-08-09
Add digest.Compute benchmark

bduffany:
```
BenchmarkDigestCompute/SHA256/1-8                2624952               455.7 ns/op
BenchmarkDigestCompute/BLAKE3/1-8                 713340              1601 ns/op
BenchmarkDigestCompute/SHA256/10-8               2636890               463.4 ns/op
BenchmarkDigestCompute/BLAKE3/10-8                812553              1558 ns/op
BenchmarkDigestCompute/SHA256/100-8              2019846               587.7 ns/op
BenchmarkDigestCompute/BLAKE3/100-8               904671              2409 ns/op
BenchmarkDigestCompute/SHA256/1000-8              515154              2291 ns/op
BenchmarkDigestCompute/BLAKE3/1000-8              546352              3174 ns/op
BenchmarkDigestCompute/SHA256/10000-8              55933             19288 ns/op
BenchmarkDigestCompute/BLAKE3/10000-8             238008              8266 ns/op
BenchmarkDigestCompute/SHA256/100000-8              6258            188327 ns/op
BenchmarkDigestCompute/BLAKE3/100000-8             49518             23969 ns/op
```

**Related issues**: N/A

### DIFF
--- a/server/remote_cache/digest/digest_test.go
+++ b/server/remote_cache/digest/digest_test.go
@@ -1,6 +1,9 @@
 package digest
 
 import (
+	"bytes"
+	"crypto/rand"
+	"fmt"
 	"regexp"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


sluongng (approved):
LGTM.

nit: it would be nice to have a separate BUILD target (perhaps `manual` tagged) to run this with `-test.bench=.` in args.

bduffany:
> LGTM.
> 
> nit: it would be nice to have a separate BUILD target (perhaps `manual` tagged) to run this with `-test.bench=.` in args.

I have done this in the past - defining a separate `{package}_benchmark_test` target - one downside I noticed is that gazelle doesn't seem to work on the second target (with `test.bench`), but haven't looked into whether there is a good workaround

PR #4522, opened 2023-08-09
Do less fake-keyboard probing during fc boot

tylerwilliams:
From https://github.com/firecracker-microvm/firecracker/blob/main/docs/api_requests/actions.md#intel-and-amd-only-sendctrlaltdel:

```
at boot time, the Linux driver for i8042 spends a few tens of milliseconds probing the device. This can be disabled by using these kernel command line parameters:

i8042.noaux i8042.nomux i8042.nopnp i8042.dumbkbd
```


bduffany (approved):
nice find

tylerwilliams:
I think we can optionally also add 8250.nr_uarts=0 if we don't need the console (aka in production?)

maggie-lou (approved):

bduffany:
> I think we can optionally also add 8250.nr_uarts=0 if we don't need the console (aka in production?)

tried it locally, looks like it is needed for `tools/vmstart` to work, but otherwise seems safe to disable. Maybe we can add an "EnableConsole" to VMConfig and only set it in vmstart

PR #4525, opened 2023-08-09
Redact `EXPLICIT_COMMAND_LINE`

siggisim:
Explicitly redacts `EXPLICIT_COMMAND_LINE` in the BuildMetadata event so it can be rendered in the UI with `<redacted>` for any command line options that should be redacted.

In all other events, it's just dropped so we don't have to deal with parsing it.

Fixes: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2540

### DIFF
--- a/server/util/redact/redact.go
+++ b/server/util/redact/redact.go
@@ -2,6 +2,7 @@ package redact
 
 import (
 	"context"
+	"encoding/json"
 	"reflect"
 	"regexp"
 	"strings"
@@ -38,6 +39,7 @@ const (
 	buildMetadataOptionPrefix = "--build_metadata="
 	allowEnvPrefix            = "ALLOW_ENV="
 	allowEnvListSeparator     = ","
+	explicitCommandLineName   = "EXPLICIT_COMMAND_LINE"
 )
 
 var (
@@ -187,9 +189,18 @@ func stripRemoteHeadersFromCmdLine(tokens []string) {
 	}
 }
 
+func stripExplicitCommandLineFromCmdLine(tokens []string) {
+	for i, token := range tokens {
+		if strings.HasPrefix(token, buildMetadataOptionPrefix+explicitCommandLineName+"=") {
+			tokens[i] = ""
+		}
+	}
+}
+
 func redactCmdLine(tokens []string) {
 	stripURLSecretsFromCmdLine(tokens)
 	stripRemoteHeadersFromCmdLine(tokens)
+	stripExplicitCommandLineFromCmdLine(tokens)
 }
 
 func stripURLSecretsFromFile(file *bespb.File) *bespb.File {
@@ -212,6 +223,16 @@ func stripRepoURLCredentialsFromBuildMetadata(metadata *bespb.BuildMetadata) {
 		if val := metadata.Metadata[repoURLKey]; val != "" {
 			metadata.Metadata[repoURLKey] = gitutil.StripRepoURLCredentials(val)
 		}
+	}
+	if m, ok := metadata.Metadata[explicitCommandLineName]; ok {
+		var commandLine []string
+		json.Unmarshal([]byte(m), &commandLine)
+		redactCmdLine(commandLine)
+		commandLineJSON, err := json.Marshal(commandLine)
+		if err != nil {
+			metadata.Metadata[explicitCommandLineName] = ""
+		}
+		metadata.Metadata[explicitCommandLineName] = string(commandLineJSON)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


server/util/redact/redact.go
```diff
@@ -213,6 +224,16 @@ func stripRepoURLCredentialsFromBuildMetadata(metadata *bespb.BuildMetadata) {
 			metadata.Metadata[repoURLKey] = gitutil.StripRepoURLCredentials(val)
 		}
 	}
+	if m, ok := metadata.Metadata[explicitCommandLineName]; ok {
+		var commandLine []string
+		json.Unmarshal([]byte(m), &commandLine)
```
bduffany, line 229: (this can return an error - it's safe to ignore but worth clarifying it with `_`)

```suggestion
		_ = json.Unmarshal([]byte(m), &commandLine)
```

server/util/redact/redact.go
```diff
@@ -213,6 +224,16 @@ func stripRepoURLCredentialsFromBuildMetadata(metadata *bespb.BuildMetadata) {
...
+		redactCmdLine(commandLine)
+		commandLineJSON, err := json.Marshal(commandLine)
+		if err != nil {
+			metadata.Metadata[explicitCommandLineName] = ""
+		}
+		metadata.Metadata[explicitCommandLineName] = string(commandLineJSON)
```
bduffany, line 235: I think this code is equivalent since the line after the err check runs unconditionally:

```suggestion
		commandLineJSON, _ := json.Marshal(commandLine)
		metadata.Metadata[explicitCommandLineName] = string(commandLineJSON)
```

server/util/redact/redact_test.go
```diff
@@ -282,6 +301,7 @@ func TestRedactMetadata_BuildMetadata_StripsURLSecrets(t *testing.T) {
 	})
 
 	assert.Equal(t, "https://github.com/buildbuddy-io/metadata_repo_url", buildMetadata.Metadata["REPO_URL"])
+	assert.Equal(t, `["--remote_header=\u003cREDACTED\u003e","--foo=SAFE"]`, buildMetadata.Metadata["EXPLICIT_COMMAND_LINE"])
```
bduffany, line 304: I think you can use `<` and `>` since they are plain ascii chars?

```suggestion
	assert.Equal(t, `["--remote_header=<REDACTED>","--foo=SAFE"]`, buildMetadata.Metadata["EXPLICIT_COMMAND_LINE"])
```

server/util/redact/redact.go
```diff
@@ -187,9 +189,18 @@ func stripRemoteHeadersFromCmdLine(tokens []string) {
 	}
 }
 
+func stripExplicitCommandLineFromCmdLine(tokens []string) {
+	for i, token := range tokens {
+		if strings.HasPrefix(token, buildMetadataOptionPrefix+explicitCommandLineName+"=") {
+			tokens[i] = ""
```
bduffany, line 195: clearing out the arg (instead of dropping) might cause issues later, since a literal `""` argument is treated by bazel as an error, but seems fine for now since we won't actually render the `""` in the UI. just something to be aware of in case we see errors in the future

```
ERROR: Skipping '': invalid target name '': empty target name
```

bduffany (approved):

server/util/redact/redact_test.go
siggisim: I had this initially, but it seems the JSON marshalling / unmarshalling code converts these into `\u003c` and `\u003e`. They're rendered correctly in the UI so it doesn't seem like a huge deal (there's a way to enable this escaping, but I'd kinda rather leave it enabled because I'm sure there's some reason for it).

What I was really looking for was an assert.Equal that behaves more like `strings.EqualFold()` but couldn't find anything.

server/util/redact/redact.go
siggisim: Ack, this should just be for UI rendering - we can switch to filtering if it causes issues.

server/util/redact/redact_test.go
bduffany: > I'd kinda rather leave it enabled because I'm sure there's some reason for it

ah I bet it's to support dumping these into HTML without the `<` being treated as a tag or something like that.

bduffany (approved):

server/util/redact/redact_test.go
bduffany: huh, TIL: https://codepen.io/bduffs/pen/LYXoKoP

PR #4526, opened 2023-08-09
Deflake quota_test

bduffany:
- Change `newQuotaManager` so that it does not return until it has successfully established the pubsub connection, otherwise quota updates that occur before the pubsub connection is established can be lost.
- Remove the `time.Sleep` and explicitly wait for the pubsub update to propagate (via a channel)

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2559


luluz66 (approved):

tylerwilliams:
Thanks for fixing this!

PR #4527, opened 2023-08-09
Account for metadata size

luluz66:
**Related issues**: N/A
Account for metadata size behind a flag.

When the flag is on, when we send size update, we will account for the metadata
size.

The plan is that when we do the migration for CDC, I'll set includeMetadataSize
to true for the dest cache.


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -918,6 +925,7 @@ func (p *PebbleCache) processSizeUpdates() {
 
 	for edit := range p.edits {
 		e := evictors[edit.partID]
+		log.Infof("update size: %d", edit.delta)
```
vadimberezniker, line 928: forgot to remove this

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1720,12 +1729,24 @@ func (p *PebbleCache) deleteFileAndMetadata(ctx context.Context, key filestore.P
 		return status.FailedPreconditionErrorf("Unnown storage metadata type: %+v", storageMetadata)
 	}
 
-	if md.GetStoredSizeBytes() > 0 {
-		p.sendSizeUpdate(partitionID, key.CacheType(), -1*md.GetStoredSizeBytes())
+	delta := md.GetStoredSizeBytes()
+	if p.includeMetadataSize {
+		delta = getTotalSizeBytes(md)
 	}
+	p.sendSizeUpdate(partitionID, key.CacheType(), -1*delta)
```
vadimberezniker, line 1736: What do you think about changing sendSizeUpdate to take `md` and 'op' (add or delete) as parameters so that this logic lives in a single place?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1676,7 +1684,8 @@ func (p *PebbleCache) deleteMetadataOnly(ctx context.Context, key filestore.Pebb
 	if err := db.Delete(fileMetadataKey, pebble.NoSync); err != nil {
 		return err
 	}
-	p.sendSizeUpdate(fileMetadata.GetFileRecord().GetIsolation().GetPartitionId(), key.CacheType(), -1*fileMetadata.GetStoredSizeBytes())
+	mdSize := int64(proto.Size(fileMetadata))
```
vadimberezniker, line 1687: The old code is not necessarily wrong. deleteMetadataOnly is called if metadata entry is present but the file on disk is missing in which case not counting StoredSizeBytes could lead to incorrect accounting. But same can be said in the other direction. Sticking to the same calculation as the rest of the code might make things simpler.

luluz66: I moved the calculation in the sendSizeUpdate.

enterprise/server/backends/pebble_cache/pebble_cache.go

vadimberezniker (commented):
While we're at it, we should also account for the size of the keys as well since they're not negligible.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2846,6 +2876,11 @@ func (e *partitionEvictor) sample(ctx context.Context, k int) ([]*approxlru.Samp
 			atime := time.UnixMicro(fileMetadata.GetLastAccessUsec())
 			age := e.clock.Since(atime)
 
+			sizeBytes := fileMetadata.GetStoredSizeBytes()
+			if e.includeMetadataSize {
+				sizeBytes = getTotalSizeBytes(fileMetadata)
```
vadimberezniker, line 2881: this should also include key size

vadimberezniker (approved):

PR #4528, opened 2023-08-10
Switch to `--remote_exec_header` for platform overrides

bduffany:
`--remote_header=` applies to both `--remote_cache` and `--remote_executor` requests, which adds unnecessary overhead to cache requests.

Platform override headers only need to be set for `Execute` requests, so switch our usage + recommendation in docs to `--remote_exec_header`.

**Related issues**: N/A


server/util/redact/redact_test.go
```diff
@@ -149,15 +149,15 @@ func TestRedactMetadata_OptionsParsed_StripsURLSecretsAndRemoteHeaders(t *testin
...
+			"--remote_exec_header=x-buildbuddy-platform.container-registry-password=TOPSECRET",
 			"--bes_header=foo=TOPSECRET",
 			"--build_metadata=PATTERN=@//foo,NAME=@bar,SECRET=TOPSECRET@",
 			"--some_other_flag=SUBFLAG=@//foo",
 		},
 		ExplicitCmdLine: []string{
 			"213wZJyTUyhXkj381312@explicit",
 			"--flag=@repo//package",
-			"--remote_header=x-buildbuddy-platform.container-registry-password=TOPSECRET_EXPLICIT",
+			"--remote_exec_header=x-buildbuddy-platform.container-registry-password=TOPSECRET_EXPLICIT",
```
sluongng, line 160: For this test I think we want to test for both `--remote_header` and `--remote_exec_header` right?

bduffany: good catch - done

sluongng (approved):
LGTM

PR #4532, opened 2023-08-10
[Firecracker] Support local snapshot sharing with firecracker vms

maggie-lou:
**Related issues**: N/A


maggie-lou:
* master
  * **PR #4532** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4532" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4532?utm_source=stack-comment).

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -556,28 +564,6 @@ func (c *FirecrackerContainer) State(ctx context.Context) (*rnpb.ContainerState,
 	return state, nil
 }
 
-func (c *FirecrackerContainer) unpackBaseSnapshot(ctx context.Context) (string, error) {
```
maggie-lou, line 559: @bduffany This function is used to get the base snapshot before merging the diff snapshot on top of it. Is there a reason we need to unpack the base snapshot to a separate dir when it's already been unpacked to chroot when loading the snapshot?

Now that we're sharing snapshots, there can be a race condition. If VM A and B start from a shared base snapshot, and then VM A merges its diff snapshots on top of it and saves it to file cache, when B tries to read the base snapshot from file cache in this function, it will read A's merged snapshot files. This will cause corruption for B's snapshot merge

maggie-lou: The overwrite happens because we aren't actually generating the digests of the snapshot files. We're setting digest = hashStrings(groupID, s.InstanceName, s.PlatformHash, s.ConfigurationHash, name). So even though the snapshot files are changing, it's overwriting the old entries

We'll need to change this for remote snapshot sharing to actually generating a digest to prevent this overwriting behavior, but the question still stands of whether we need to unpack to a separate dir

bduffany: > Is there a reason we need to unpack the base snapshot to a separate dir when it's already been unpacked to chroot when loading the snapshot?

I don't know the full context since I didn't write this code originally, but it seems like we can remove it :+1:

> Now that we're sharing snapshots, there can be a race condition. If VM A and B start from a shared base snapshot, and then VM A merges its diff snapshots on top of it and saves it to file cache, when B tries to read the base snapshot from file cache in this function, it will read A's merged snapshot files. This will cause corruption for B's snapshot merge

Makes sense.

> The overwrite happens because we aren't actually generating the digests of the snapshot files. We're setting digest = hashStrings(groupID, s.InstanceName, s.PlatformHash, s.ConfigurationHash, name). So even though the snapshot files are changing, it's overwriting the old entries
>
> We'll need to change this for remote snapshot sharing to actually generating a digest to prevent this overwriting behavior

Hmm if I'm understanding things correctly, I think we need to do this for local snapshot sharing to work too, not just remote? Otherwise, I think we could hit the following race condition (to make it easier to understand, let's pretend that the GroupID, InstanceName, PlatformHash, ConfigurationHash are all the same between both VMs, and so the artifact hashes are just hashes of the file names. Also, pretend that we don't hash the filenames and just use them as the cache keys directly)

```
- VM A: calls CacheSnapshot()
  - Stores filecache entry `"vmstate.snap" => <vmstate file>`
- VM B (interrupting VM A's thread) calls CacheSnapshot
  - Stores filecache entry `"vmstate.snap" => <vmstate file>`
  - Stores filecache entry `"manifest.json" => <manifest file>`
- VM A: resuming CacheSnapshot()
  - Stores filecache entry `"manifest.json" => <manifest file>`
```

But the manifest that VM A stored now points to the `vmstate.snap` that VM B wrote, which is incorrect. This would be fixed by keying vmstate.snap under a proper content digest instead of just the file name.

We also need to do this for `containerfs.ext4` although we should make sure to only compute that digest once since it's expensive and the containerfs should be immutable. Since the container image ref is part of the platform which is part of the cache key, we _technically_ don't need to do that in this PR, but should eventually.

Also, we already compute proper digests for the COW chunks - in snaploader.go see `cacheCOW` (:moneybag: :cow2: ) I think it's just `vmstate.snap` and `containerfs.ext4` that we have to worry about.

tylerwilliams: just responding to this part: `Is there a reason we need to unpack the base snapshot to a separate dir when it's already been unpacked to chroot when loading the snapshot?` -- the chroots are meant to be isolated directories that are removed after the firecracker instance running inside finishes. So doing anything in a chroot directory that is not tied to the life of the instance might cause issues?

bduffany: Today, `unpackBaseSnapshot` already unpacks to `filepath.Join(c.getChroot(), "base")` - and the CreateSnapshot stuff should be happening after pause but before remove - so I think doing our mergeDiffSnapshot stuff inside the chroot is probably fine

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -423,7 +424,14 @@ func NewContainer(ctx context.Context, env environment.Env, imageCacheAuth *cont
 		if err != nil {
 			return nil, err
 		}
-		c.snapshotKey, err = snaploader.NewKey(task, cd.GetHash(), c.id)
+
+		// TODO(Maggie): Once local snapshot sharing is stable, remove runner ID
+		// from the snapshot key
+		runnerID := c.id
+		if *enableNBD && *enableLocalSnapshotSharing {
+			runnerID = "SHARED"
```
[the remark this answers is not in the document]
maggie-lou: Smart - good call!

enterprise/server/remote_execution/containers/firecracker/firecracker.go
maggie-lou: Adding digest computation [here](https://github.com/buildbuddy-io/buildbuddy/pull/4574) to address the race condition you flagged Brandon. Still debugging something on that PR, so just going to merge this one first to make rebasing easier

PR #4533, opened 2023-08-10
Lessen read amplification during eviction sampling

tylerwilliams:
Basically the idea is: files are laid out in a random order, because they are written in "digest" order, so rather than seeking all around to read k random digests, just seek once and read k digests in order.

Hopefully this reduces the number of blocks we need to read and random seeks we do against the backing disk.


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2784,6 +2789,47 @@ func (e *partitionEvictor) sampleGroup() {
...
+	// Attempt AC sampling (but do not error out if it fails, instead fall
+	// through to CAS sampling)
+	if sampleType == rspb.CacheType_AC {
+		gid, err := e.randomGroupForEvictionSampling()
+		if err != nil {
+			if !status.IsNotFoundError(err) {
+				log.Warningf("no groups to sample for %q: %s", e.part.ID, err)
+			}
+		} else {
+			if rk, err := e.randomKey(64, gid, sampleType); err == nil {
```
vadimberezniker, line 2820: Do you need both this and the randomKey call below? If you leave gid as empty if not doing AC sampling you can just have one randomKey call at the end of the function.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2795,46 +2841,34 @@ func (e *partitionEvictor) sample(ctx context.Context, k int) ([]*approxlru.Samp
...
-			} else if !status.IsNotFoundError(err) {
-				log.Warningf("no groups to sample for %q: %s", e.part.ID, err)
+		if !iter.Valid() {
+			// This should only happen once per call to sample(), or
+			// occasionally more if we've exhausted the iter.
+			randomKey, err := e.randomKeyForEvictionSampling()
+			if err != nil {
+				return nil, err
 			}
+			iter.SeekGE(randomKey)
```
vadimberezniker, line 2865: Not sure if this was intentional but you are skipping the first key since you call SeekGE and then immediately call iter.Next(). Shouldn't really matter in practice since we're picking random keys anyway, but might be worth documenting that this is intentional.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: PTAL

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2784,6 +2789,42 @@ func (e *partitionEvictor) sampleGroup() {
...
+	// Attempt AC sampling (but do not error out if it fails, instead fall
+	// through to CAS sampling)
+	if sampleType == rspb.CacheType_AC {
+		gid, err = e.randomGroupForEvictionSampling()
+		if err != nil && !status.IsNotFoundError(err) {
+			log.Warningf("no groups to sample for %q: %s", e.part.ID, err)
+		}
+	}
+
+	// If this is a CAS sample batch, or if AC sampling has failed above,
```
vadimberezniker, line 2823: You can remove this comment now.

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: looks good

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go

PR #4534, opened 2023-08-10
Compaction tweaks

siggisim:
- Start compaction sooner: L0CompactionThreshold = 2
- Start parallel compactions sooner: L0CompactionConcurrency = 2
- Raise max compactions limit: MaxConcurrentCompactions = 12
- Slow down the rate at which parallel slots are added: CompactionDebtConcurrency = 10 << 30 (10GB of compaction debt per slot)

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -433,6 +433,7 @@ func ensureDefaultPartitionExists(opts *Options) {
 func defaultPebbleOptions(el *pebbleEventListener) *pebble.Options {
 	// These values Borrowed from CockroachDB.
 	opts := &pebble.Options{
+        // The amount of L0 read-amplification necessary to trigger an L0 compaction.
 		L0CompactionThreshold:    2,
 		MaxConcurrentCompactions: 12,
 		MemTableSize:             64 << 20, // 64 MB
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -432,15 +432,21 @@ func ensureDefaultPartitionExists(opts *Options) {
...
+		MaxConcurrentCompactions: 12,
 		MemTableSize:             64 << 20, // 64 MB
 		EventListener: pebble.EventListener{
 			WriteStallBegin: el.WriteStallBegin,
 			WriteStallEnd:   el.WriteStallEnd,
 			DiskSlow:        el.DiskSlow,
 		},
 	}
+
+	opts.Experimental.L0CompactionConcurrency = 2
```
tylerwilliams, line 446:
```suggestion
	// The threshold of L0 read-amplification at which compaction concurrency
	// is enabled (if CompactionDebtConcurrency was not already exceeded).
	// Every multiple of this value enables another concurrent
	// compaction up to MaxConcurrentCompactions.
	opts.Experimental.L0CompactionConcurrency = 2
```

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -432,15 +432,21 @@ func ensureDefaultPartitionExists(opts *Options) {
...
+	opts.Experimental.CompactionDebtConcurrency = 10 << 30
```
tylerwilliams, line 447:
```suggestion
	// CompactionDebtConcurrency controls the threshold of compaction debt
	// at which additional compaction concurrency slots are added. For every
	// multiple of this value in compaction debt bytes, an additional
	// concurrent compaction is added. This works "on top" of
	// L0CompactionConcurrency, so the higher of the count of compaction
	// concurrency slots as determined by the two options is chosen.
	opts.Experimental.CompactionDebtConcurrency = 10 << 30
```

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -432,15 +432,21 @@ func ensureDefaultPartitionExists(opts *Options) {
 // defaultPebbleOptions returns default pebble config options.
 func defaultPebbleOptions(el *pebbleEventListener) *pebble.Options {
 	// These values Borrowed from CockroachDB.
-	return &pebble.Options{
-		MaxConcurrentCompactions: 6,
+	opts := &pebble.Options{
+		L0CompactionThreshold:    2,
```
tylerwilliams, line 436:
```suggestion
        // The amount of L0 read-amplification necessary to trigger an L0 compaction.
		L0CompactionThreshold:    2,
```

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: this is 2 and L0CompactionThreshold is 2, so parallel compactions will start immediately. We could make this 4 or something for a more gradual ramp up?

tylerwilliams (approved):

PR #4535, opened 2023-08-10
Use "ON CONFLICT" instead of trying and failing to insert for an Invocation

tempoz:
~~Since Postgres returns an unhelpful `InFailedSQLTransaction` on a duplicate key conflict in the invocation db `Create` in the case of a pre-existing invocation attempt, we instead use Postgres and SQLite's "Upsert" syntax to perform this insertion/update. Now errors will all only be real errors, so there is no need to discriminate (or fail to discriminate) amongst them. As there is no need for more than one SQL query with the `RETURNING` clause supported by both SQLite and Postgres, we do not need to use a transaction unless we are using MySQL.~~

~~Added a function in `db.go` to allow us to update all fields programmatically and still do custom update logic, which gorm does not smoothly support.~~

Added tests to confirm the behavior. Switched the corresponding MySQL query to use the database's `NowFunc` to make testing the time-based constraints possible

EDIT: Now we just do nothing if we get a conflict on `INSERT` / `Create`. If no rows were inserted and there's no error, that means we hit a conflict and need to update instead.

**Related issues**: buildbuddy-io/buildbuddy#4466

### DIFF
--- a/server/backends/invocationdb/invocationdb.go
+++ b/server/backends/invocationdb/invocationdb.go
@@ -16,12 +16,17 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/retry"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 
 	aclpb "github.com/buildbuddy-io/buildbuddy/proto/acl"
 	akpb "github.com/buildbuddy-io/buildbuddy/proto/api_key"
 	inspb "github.com/buildbuddy-io/buildbuddy/proto/invocation_status"
 	telpb "github.com/buildbuddy-io/buildbuddy/proto/telemetry"
 	uidpb "github.com/buildbuddy-io/buildbuddy/proto/user_id"
+)
+
+const (
+	upsertInvocationQueryName = "upsert_invocation"
 )
 
 type InvocationDB struct {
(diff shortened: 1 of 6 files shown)


tempoz:
* master
  * **PR #4535** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4535" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4535?utm_source=stack-comment).

server/backends/invocationdb/invocationdb.go
```diff
@@ -40,10 +45,42 @@ func getACL(i *tables.Invocation) *aclpb.ACL {
...
+					gorm.Expr(`"Invocations".invocation_status <> ?`, int64(inspb.InvocationStatus_COMPLETE_INVOCATION_STATUS)),
+					gorm.Expr(`"Invocations".updated_at_usec > ?`, d.NowFunc().Add(time.Hour*-4).UnixMicro()),
+				},
+			},
+			DoUpdates: updates,
+		},
+		clause.Returning{},
+	).Create(ti)
+	return result.RowsAffected > 0, result.Error
+}
```
bduffany, line 71: This code feels difficult to grok because it is using some of the more advanced GORM functionality (`clause.OnConflict{}`, `clause.Returning{}`, Session, etc), which itself maps to some advanced-ish SQL.

Would it work to do a simple transaction like this?

```
Transaction("upsert_invocation", func (tx) {
  // Get existing row; use SELECT ... FOR UPDATE construct
  // to lock the row during the transaction.
  // Include iid in WHERE clause, and maybe attempt number
  // if we know this is not the first attempt.
  existing := &Invocation{}
  err := tx.Raw(`SELECT * FROM Invocations WHERE ... ` +db.SelectForUpdateModifier()).Take(existing).Error

  // check err; but allow NotFound
  /* ... */
  
  // If we know the current attempt number, check to make sure
  // we weren't preempted by a more recent attempt
  if (ourAttempt > 0 && existing != nil && existing.Attempt > ourAttempt) {
    return // a newer attempt exists; do nothing.
  }
  if (!existing) {
    // Create empty invocation row for (iid, attempt)
    /* ... */
  }
  tx.Updates(invocation)
})
```

and then maybe wrap this with a retry+backoff in case the lock failed.

tempoz: Could I just add comments to make the clause stuff easier to understand? It doesn't feel simpler to me to add all of this application logic and a transaction and then have to add a retry and backoff as well. The clauses are pretty straightforward, imo, and I'm happy to add comments explaining them.

tempoz: I'd write the raw SQL for it but we *need* the ORM to ensure code maintainability as we add/remove columns from tables.

tempoz:
Now we just do nothing if we get a conflict on `INSERT` / `Create`. If no rows were inserted and there's no error, that means we hit a conflict and need to update instead.

bduffany (approved):

PR #4536, opened 2023-08-11
Render Target page using GetTarget API

bduffany:
**Related issues**: N/A


proto/target.proto
```diff
@@ -159,6 +159,12 @@ message Target {
   // Test results associated with the target. The whole event is included
   // because the ID contains useful info, such as configuration.
   repeated build_event_stream.BuildEvent test_result_events = 6;
+
+  // Test summary associated with the target.
+  build_event_stream.TestSummary test_summary = 7;
+
+  // ActionCompleted events associated with the target.
+  repeated build_event_stream.BuildEvent action_events = 8;
```
bduffany, line 167: @siggisim I know we chatted earlier about computing this in the UI, but since we don't return the raw Completed events to the UI anymore, I think we have to compute it on the server

siggisim (approved):

proto/target.proto
siggisim: As long as the server code isn't too complicated, I'm cool with it

PR #4539, opened 2023-08-11
Move API key caching to authdb.

vadimberezniker:
The caching needs to be adjusted for subdomain handling and it makes more sense for the caching to be an internal detail of AuthDB.

**Related issues**: N/A


enterprise/server/backends/authdb/authdb.go
```diff
@@ -37,23 +42,86 @@ const (
...
+	data         interfaces.APIKeyGroup
+	expiresAfter time.Time
+}
+
+// apiKeyGroupCache is a cache for API Key -> Group lookups. A single Bazel invocation
+// can generate large bursts of RPCs, each of which needs to be authed.
+// There's no need to go to the database for every single request as this data
+// rarely changes.
+type apiKeyGroupCache struct {
+	// Note that even though we base this off an LRU cache, every entry has a hard expiration
```
tylerwilliams, line 68: nit: 80 chars?

enterprise/server/backends/authdb/authdb.go
```diff
@@ -37,23 +42,86 @@ const (
...
 
+	apiKeyGroupCache *apiKeyGroupCache
+
 	// Nil if API key encryption is not enabled.
 	apiKeyEncryptionKey []byte
 }
 
 func NewAuthDB(env environment.Env, h interfaces.DBHandle) (*AuthDB, error) {
 	adb := &AuthDB{env: env, h: h}
+	if *apiKeyGroupCacheTTL != time.Duration(0) {
```
tylerwilliams, line 118: if apiKeyGroupCacheTTL is a time.Duration you should just be able to compare it to 0?

enterprise/server/backends/authdb/authdb.go
```diff
@@ -296,10 +372,22 @@ func (d *AuthDB) GetAPIKeyGroupFromAPIKey(ctx context.Context, apiKey string) (i
 		}
 		return nil, err
 	}
+	if d.apiKeyGroupCache != nil {
+		metrics.APIKeyLookupCount.With(prometheus.Labels{metrics.APIKeyLookupStatus: "cache_miss"}).Inc()
+		d.apiKeyGroupCache.Add(apiKey, akg)
+	} else {
+		metrics.APIKeyLookupCount.With(prometheus.Labels{metrics.APIKeyLookupStatus: "invalid_key"}).Inc()
```
tylerwilliams, line 379: it this in the right place? It seems like it will fire any time apiKeyGroupCache is nil.

tylerwilliams (approved):

enterprise/server/backends/authdb/authdb.go
vadimberezniker: done

enterprise/server/backends/authdb/authdb.go
vadimberezniker: done

enterprise/server/backends/authdb/authdb.go
vadimberezniker: woops, this was also broken in the old code

PR #4546, opened 2023-08-14
Small improvements/fixes to replay_invocation

bduffany:
* When pressing Ctrl+C, actually cancel the script (before, it would keep running in the background)
* Print a progress message every 10K events so we know the script isn't stuck.

**Related issues**: N/A


tempoz (approved):
LGTM

PR #4548, opened 2023-08-14
Deflake usage_test_postgres

bduffany:
Switch to table-level locking for writing usage rows, since only one app should be writing to the table at a time anyway.

The SELECT FOR UPDATE does not cause the non-existent row to be locked in postgres, and in mysql it happens to work but I don't know why. Locking non-existent rows generally seems like a bad idea - https://stackoverflow.com/questions/17068686/how-do-i-lock-on-an-innodb-row-that-doesnt-exist-yet

**Related issues**: N/A

### DIFF
--- a/enterprise/server/usage/usage.go
+++ b/enterprise/server/usage/usage.go
@@ -328,9 +328,18 @@ func (ut *tracker) flushCounts(ctx context.Context, groupID string, p period, la
 		json := string(b)
 		log.Infof("Flushing usage row: %s", json)
 
-		// First check whether the row already exists. Make sure to select for
-		// update in order to lock the row.
-		err := tx.Raw(`
+		// Lock the table so that other apps cannot attempt to insert usage rows
+		// with the same key. Note that this locking should not affect
+		// performance since only one app should be writing to the DB at a time
+		// anyway.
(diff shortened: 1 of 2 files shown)


vadimberezniker (approved):

tempoz:
If you're willing to migrate the `Usages` table to using `region`, `group_id`, `period_start_usec`, `origin`, and `client` as a composite primary key (and shorten, say, `origin` and `client` to 128 characters apiece to avoid going over MySQL's primary key limit of 3072 bytes), you can do this without the full table lock or the `SELECT` (or the transaction) in a single `Create`:
```go
result := dbh.DB(ctx).Clauses(clause.OnConflict{DoNothing: true}.Create(tu)
```
and then check `result.RowsAffected` to see if an `INSERT` actually occurred. IMO the main benefit of this approach would be not locking the table, since locking a table is not generally great for performance. If we're not worried about write performance in the `Usages` table, then all of that would be unnecessary.

bduffany:
@tempoz Only one app is supposed to write to this table at a time, so write performance isn't an issue. I didn't want to use a primary key / unique index because it's too easy to mess up the DB migration when adding more label columns.

tempoz:
Cool, then this is great!

tempoz (approved):

PR #4549, opened 2023-08-14
Add logging whenever gRPC sends a keepalive ping

bduffany:
**Related issues**: N/A

### DIFF
--- a/server/util/grpc_client/grpc_client.go
+++ b/server/util/grpc_client/grpc_client.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/rpc/interceptors"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"google.golang.org/grpc"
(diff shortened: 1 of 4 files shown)


server/util/grpc_client/grpc_client.go
```diff
@@ -78,6 +82,12 @@ func CommonGRPCClientOptions() []grpc.DialOption {
 
 			// If true, client sends keepalive pings even with no active RPCs.
 			PermitWithoutStream: true,
+
+			PingCallback: func(connectionID uint64) {
+				if *logKeepalivePings {
+					log.Debugf("Sending keepalive ping for connection %d", connectionID)
```
tylerwilliams, line 88: do you want this to be at a higher level than debug so we can see it in prod logs?

bduffany: this is happening pretty frequently in dev (tens of times per day) so I think it's fine to just have it at debug level

tylerwilliams (approved):

PR #4557, opened 2023-08-15
Allow oauth login to work for subdomains.

vadimberezniker:
Set domain on cookies to allow cookies to be shared across subdomains and domain. Auth flow will happen on the domain auth endpoint and redirect back to the subdomain after auth.

**Related issues**: N/A

### DIFF
--- a/server/endpoint_urls/build_buddy_url/build_buddy_url.go
+++ b/server/endpoint_urls/build_buddy_url/build_buddy_url.go
@@ -2,6 +2,7 @@ package build_buddy_url
 
 import (
 	"net/url"
+	"strings"
 
 	flagtypes "github.com/buildbuddy-io/buildbuddy/server/util/flagutil/types"
 )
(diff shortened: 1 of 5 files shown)


siggisim (approved):

server/util/cookie/cookie.go
```diff
@@ -23,10 +26,19 @@ var (
 	httpsOnlyCookies = flag.Bool("auth.https_only_cookies", false, "If true, cookies will only be set over https connections.")
 )
 
+func cookieDomain() string {
+	domain := ""
+	if subdomain.Enabled() {
+		domain = build_buddy_url.Domain()
+	}
+	return domain
+}
```
bduffany, line 35: When switching from one domain to another, I wonder if we should detect cookies from the old domain (`app.buildbuddy.dev` / `app.buildbuddy.io`) and ignore/delete those? See https://buildbuddy-corp.slack.com/archives/C01D5GHRJ59/p1692108833397279

vadimberezniker: Updated PR to clear non-domain bound cookies when setting domain bound cookies.

server/endpoint_urls/build_buddy_url/build_buddy_url.go
```diff
@@ -15,3 +16,12 @@ func WithPath(path string) *url.URL {
 func String() string {
 	return buildBuddyURL.String()
 }
+
+func Domain() string {
```
tylerwilliams, line 20: nit: add a comment with an example?

PR #4564, opened 2023-08-15
Start populating usage label columns

bduffany:
To be merged after this week's release, to avoid changing too much at once (this week's release also contains https://github.com/buildbuddy-io/buildbuddy/pull/4521)

**Related issues**: N/A

### DIFF
--- a/server/util/usageutil/usageutil.go
+++ b/server/util/usageutil/usageutil.go
@@ -4,10 +4,56 @@ import (
 	"context"
 
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/bazel_request"
+	"github.com/buildbuddy-io/buildbuddy/server/util/pbwireutil"
+	"google.golang.org/grpc/metadata"
+)
+
+// Client label constants.
+const (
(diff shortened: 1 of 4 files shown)


vadimberezniker (approved):

server/util/usageutil/usageutil.go
```diff
@@ -4,10 +4,56 @@ import (
...
+	if len(vals) == 0 {
+		return ""
+	}
+	return vals[0]
+}
+
+func clientLabel(ctx context.Context) string {
+	// Note: we avoid deserializing the RequestMetadata proto here since
+	// proto deserialization is too expensive to run on every request.
+	b := bazel_request.GetRequestMetadataBytes(ctx)
```
tylerwilliams, line 37: Do we do this with metadata in multiple places? Would it make sense to deserialize the full proto once in the interceptor and throw fields in the context so we don't need to reparse?

tylerwilliams (approved):

server/util/usageutil/usageutil.go
bduffany: Yeah, that would be a nice improvement - will look at that when I have more time.

PR #4570, opened 2023-08-16
Fix uffd.Stop() hanging when firecracker initialization fails

bduffany:
**Related issues**: N/A


maggie-lou:
To simplify this, what if we set h.lis = nil at the end of receiveSetupMsg() and added
if h.lis != nil {
    h.lis.Close()
} in Stop() 

That way we don't need 2 channels

bduffany:
> To simplify this, what if we set h.lis = nil at the end of receiveSetupMsg() and added if h.lis != nil { h.lis.Close() } in Stop()
> 
> That way we don't need 2 channels

I think that would hit a race condition in the case where we call Remove() while we are still waiting to receive the setup message, (e.g. due to context timeout).

maggie-lou (approved):

PR #4575, opened 2023-08-17
platform: allow env-overrides in base64

sluongng:
Currently we parse env-overrides request headers with `,` to separate
between different pair of variable's name and value. This approach works
for simple value string but when folks want to pass a JSON containing
commas characters within, the parse would fail.

Provide `env-overrides-base64` as a work around for more complex values.
User could encode the `k=v` pair in base64 and pass it to the new
header. Our server will attempt to decode the pair and append it to
existing overrides.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/platform/platform.go
+++ b/enterprise/server/remote_execution/platform/platform.go
@@ -2,6 +2,7 @@ package platform
 
 import (
 	"context"
+	"encoding/base64"
 	"flag"
 	"fmt"
 	"runtime"
@@ -81,6 +82,7 @@ const (
 	disablePredictedTaskSizePropertyName = "debug-disable-predicted-task-size"
 	extraArgsPropertyName                = "extra-args"
 	EnvOverridesPropertyName             = "env-overrides"
+	EnvOverridesBase64PropertyName       = "env-overrides-base64"
 	podmanImageStreamingPropertyName     = "podman-enable-image-streaming"
 	IncludeSecretsPropertyName           = "include-secrets"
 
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/platform/platform.go
```diff
@@ -226,6 +228,15 @@ func ParseProperties(task *repb.ExecutionTask) *Properties {
 	// Only Enable VFS if it is also enabled via flags
 	vfsEnabled := boolProp(m, enableVFSPropertyName, false) && *enableVFS
 
+	envOverrides := stringListProp(m, EnvOverridesPropertyName)
+	for _, prop := range stringListProp(m, EnvOverridesBase64PropertyName) {
+		b, err := base64.StdEncoding.DecodeString(prop)
+		// Ignore decode errors
```
tylerwilliams, line 234: probably shouldn't ignore these -- makes it hard for a user to tell why their override isn't working

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -81,6 +82,7 @@ const (
 	disablePredictedTaskSizePropertyName = "debug-disable-predicted-task-size"
 	extraArgsPropertyName                = "extra-args"
 	EnvOverridesPropertyName             = "env-overrides"
+	EnvOverridesBase64PropertyName       = "env-overrides-base64"
```
tylerwilliams, line 85: in the docs for this, when you write them, worth documenting which takes precedence if both set the same value

enterprise/server/remote_execution/platform/platform.go
bduffany: `ParseProperties` returns `*Properties` instead of `*Properties, error` (my mistake) and there are a lot of callsites, so I think that would result in this PR being pretty big - we might want to do that separately

sluongng: I could send a separate PR to change the return and return a `status.FailedPreconditionError` when the base64 decode fails.

enterprise/server/remote_execution/platform/platform.go
sluongng: Will add a doc for this in a follow-up PR. I am thinking between https://www.buildbuddy.io/docs/secrets or https://www.buildbuddy.io/docs/rbe-microvms#using-custom-images but not sure yet which one is best.

bduffany (approved):

PR #4577, opened 2023-08-17
Support repo default branch names in code editor

siggisim:
Currently we assume the (sometimes incorrect) default branch name.

### DIFF
--- a/enterprise/app/code/code.tsx
+++ b/enterprise/app/code/code.tsx
@@ -34,6 +34,7 @@ interface Props {
 interface State {
   commitSHA: string;
   repoResponse: any;
+  treeResponse: any;
   treeShaToExpanded: Map<string, boolean>;
   treeShaToChildrenMap: Map<string, any[]>;
   fullPathToModelMap: Map<string, any>;
@@ -55,6 +56,10 @@ interface State {
 }
 
 const LOCAL_STORAGE_STATE_KEY = "code-state-v1";
+var repoLoaded: (value: any) => void;
+var repoLoadedPromise = new Promise<any>(function (resolve) {
+  repoLoaded = resolve;
+});
 
 // TODO(siggisim): Add links to the code editor from anywhere we reference a repo
 // TODO(siggisim): Add branch / workspace selection
@@ -63,6 +68,7 @@ export default class CodeComponent extends React.Component<Props, State> {
   state: State = {
     commitSHA: "",
     repoResponse: undefined,
+    treeResponse: undefined,
     treeShaToExpanded: new Map<string, boolean>(),
     treeShaToChildrenMap: new Map<string, any[]>(),
     fullPathToModelMap: new Map<string, any>(),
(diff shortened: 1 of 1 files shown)


enterprise/app/code/code.tsx
```diff
@@ -55,6 +56,10 @@ interface State {
 }
 
 const LOCAL_STORAGE_STATE_KEY = "code-state-v1";
+var repoLoaded: (value: any) => void;
+var repoLoadedPromise = new Promise<any>(function (resolve) {
+  repoLoaded = resolve;
+});
```
bduffany, line 62: nit: can you make these fields of CodeComponent? otherwise I think there could be problems when switching between this page and other pages

enterprise/app/code/code.tsx
```diff
@@ -143,14 +149,16 @@ export default class CodeComponent extends React.Component<Props, State> {
       return;
     }
 
+    let repoResponse = await repoLoadedPromise;
```
bduffany, line 152: I think `componentDidMount()` is not supposed to do async stuff, because by the time the promise completes, the component might have already unmounted.

I think it would be slightly better to have a `componentDidUpdate(prevProps, prevState)` that looks at prevState and checks `if (!prevState.repoResponse && this.state.repoResponse) { doAllThisLogicBelow() }` - that would also let you get rid of the promise

enterprise/app/code/code.tsx
```diff
@@ -261,24 +269,28 @@ export default class CodeComponent extends React.Component<Props, State> {
...
+    repoLoaded(repoResponse);
+    let commit = this.state.commitSHA || repoResponse.data.default_branch;
     this.octokit!.request(`/repos/${this.currentOwner()}/${this.currentRepo()}/git/trees/${commit}`).then(
-      (response: any) => {
-        console.log(response);
-        this.updateState({ repoResponse: response, commitSHA: response.data.sha });
+      (treeResponse: any) => {
+        console.log(treeResponse);
+        this.updateState({ repoResponse: repoResponse, treeResponse: treeResponse, commitSHA: treeResponse.data.sha });
       }
```
bduffany, line 294: nit, the serial requests here will probably be slow (because the GH API is generally pretty slow) - if you want, you could do

```
let [repoResponse, treeResponse] = await Promise.all([
  this.octokit!.request(`GET /repos/${this.currentOwner()}/${this.currentRepo()}`),
  this.octokit!.request(`/repos/${this.currentOwner()}/${this.currentRepo()}/git/trees/${commit}`),
])
this.updateState({ reposResponse, treeResponse })

```

siggisim: You need the result from the first request to make the second request if the commitSha isn't specified, i.e.
```
    let commit = this.state.commitSHA || repoResponse.data.default_branch;
```

enterprise/app/code/code.tsx
siggisim: Removed them

enterprise/app/code/code.tsx
siggisim: Great idea - I was able to remove the promise by moving things to `componentDidUpdate`.

I still needed to call this method in `componentDidMount` though because `componentDidUpdate` isn't called on initial render (and if state is stored in localstorage - everything is ready and loaded into state on initial render.)

Definitely feels much cleaner though.

bduffany (approved):

PR #4581, opened 2023-08-17
Make raft splits happen via raft command

tylerwilliams:


enterprise/server/raft/store/store.go
```diff
@@ -1753,7 +1753,11 @@ func (s *Store) SplitCluster(ctx context.Context, req *rfpb.SplitClusterRequest)
 
 	s.log.Infof("splitlog: acquired split lease")
 
-	if err := bringup.CloneCluster(ctx, s.apiClient, bootStrapInfo, req.GetRange().GetRangeId(), newIDs.rangeID); err != nil {
+	cloneReq := rbuilder.NewBatchBuilder().Add(&rfpb.CloneClusterRequest{
```
vadimberezniker, line 1756: Will this hold up properly if the CloneClusterRequest is replayed from the raft log? I don't believe Clone is idempotent.

PR #4590, opened 2023-08-18
Add a timeout when waiting for 'OK' from firecracker guest over vsock.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/util/vsock/vsock.go
+++ b/enterprise/server/util/vsock/vsock.go
@@ -140,6 +140,10 @@ func dialHostToGuest(ctx context.Context, socketPath string, port uint32) (net.C
 	if n != len(fcConnectString) {
 		return nil, status.InternalErrorf("HostDial failed: wrote %d bytes, expected %d", n, len(fcConnectString))
 	}
+	// Firecracker should normally be listening on v.sock as soon as
+	// the VM is started, but occasionally we might call this function
+	// before firecracker starts listening. Set a short timeout so that
+	// we catch this early and retry instead of hanging.
 	conn.SetReadDeadline(time.Now().Add(1 * time.Second))
 	rsp, err := bufio.NewReaderSize(conn, 32).ReadString('\n')
 	if err != nil {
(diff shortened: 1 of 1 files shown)


enterprise/server/util/vsock/vsock.go
```diff
@@ -140,10 +140,12 @@ func dialHostToGuest(ctx context.Context, socketPath string, port uint32) (net.C
 	if n != len(fcConnectString) {
 		return nil, status.InternalErrorf("HostDial failed: wrote %d bytes, expected %d", n, len(fcConnectString))
 	}
+	conn.SetReadDeadline(time.Now().Add(1 * time.Second))
```
bduffany, line 143: Trying to make sure I understand how this works - does this sound right to you?

* Firecracker should start listening on v.sock as soon as the VM is started, even before we start the vmexec server, and should normally be able to reply to these CONNECT messages immediately
* When the CONNECT does get accepted by firecracker, but vmexec is not yet listening, firecracker will just terminate the connection (based on [this](https://github.com/firecracker-microvm/firecracker/blob/main/docs/vsock.md#host-initiated-connections)
* Sometimes we get unlucky though and firecracker doesn't start listening to v.sock by the time we call this func, so the Read() just hangs?
* This PR will make it so that if we happen to get unlucky, we'll just fail instead of waiting longer than we need to?

Also, in the case where firecracker terminates the connection (due to the vmexec server not yet listening) - I guess SimpleGRPCDial will retry? I see there's a backoff config but not 100% sure if the backoff is just retrying the bufDialer func or something else

bduffany, line 143:
```suggestion
	// Firecracker should normally be listening on v.sock as soon as
	// the VM is started, but occasionally we might call this function
	// before firecracker starts listening. Set a short timeout so that
	// we catch this early and retry instead of timing out.
	conn.SetReadDeadline(time.Now().Add(1 * time.Second))
```

bduffany (approved):
LGTM

tylerwilliams (approved):

PR #4593, opened 2023-08-18
Use compression.NewZstdDecompressor to decompress when cdc is enabled

luluz66:
**Related issues**: N/A
To fix dev migration error unexpected EOF

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1864,6 +1864,18 @@ func (p *PebbleCache) newCDCCommitedWriteCloser(ctx context.Context, fileRecord 
 		shouldCompress: shouldCompress,
 		isCompressed:   isCompressed,
 	}
+	var wc, decompressor io.WriteCloser
+	wc = cdcw
+
+	if isCompressed {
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1864,16 +1864,34 @@ func (p *PebbleCache) newCDCCommitedWriteCloser(ctx context.Context, fileRecord
...
 	if err != nil {
 		return nil, err
 	}
 	cdcw.chunker = chunker
 
-	cwc := ioutil.NewCustomCommitWriteCloser(cdcw)
+	cwc := ioutil.NewCustomCommitWriteCloser(wc)
 	cwc.CloseFn = db.Close
 	cwc.CommitFn = func(bytesWritten int64) error {
+		if decompressor != nil {
```
vadimberezniker, line 1889: Is it necessary to close the decompressor here? It will be closed when Close() is called on  the CustomCommitWriteCloser.

luluz66: Yes. If the decompressor is not closed; then the decompressor can write to cdcw when cdcw is already closed.

vadimberezniker (approved):

PR #4595, opened 2023-08-21
Add protos for GitHub requests we currently make

siggisim:
These protos contain just the minimal set of request / response data necessary to render the UI. 

This is not an attempt to recreate the entire Github API an all of its functionality, its just the bits that we actually need.

### DIFF
--- a/proto/github.proto
+++ b/proto/github.proto
@@ -180,3 +180,279 @@ message UnlinkRepoRequest {
 message UnlinkRepoResponse {
   context.ResponseContext response_context = 1;
 }
+
+
+// GET /user/installations
+message GetGithubUserInstallationsRequest {
+  context.RequestContext request_context = 1;
+}
+
+message UserInstallation {
+  int64 id = 1;
+  string login = 2;
+  string url = 3;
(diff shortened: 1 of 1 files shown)


proto/github.proto
```diff
@@ -180,3 +180,277 @@ message UnlinkRepoRequest {
...
+
+message UserInstallationPermissions {
+  string administration = 1;
+  string repository_hooks = 2;
+  string pull_requests = 3;
+}
+
+message GetGithubUserInstallationsResponse {
+  context.ResponseContext response_context = 1;
+  repeated UserInstallation installation = 2;
```
bduffany, line 205: uber nit: we have "repeated fields should be plural" in our style guide https://www.buildbuddy.dev/codebase#proto but I know there are various opinions on this :P

my 2c is that the plural naming makes it easier to read/write code using the generated APIs since you can tell just by the name whether it's an array or not.

bduffany (approved):

proto/github.proto
siggisim: Done

PR #4601, opened 2023-08-21
Restrict visible groups when accessing customer subdomain.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -154,6 +154,7 @@ export class AuthService {
         )
       ),
       isImpersonating: Boolean(rpcService.requestContext.impersonatingGroupId),
+      subdomainGroupID: response.subdomainGroupId,
     });
   }
 
--- a/enterprise/app/org/org_access_denied.tsx
+++ b/enterprise/app/org/org_access_denied.tsx
@@ -1,0 +1,42 @@
+import { User } from "../../../app/auth/user";
+import React from "react";
+import FilledButton from "../../../app/components/button/button";
+import authService from "../../../app/auth/auth_service";
+import router from "../../../app/router/router";
+
+export type Props = {
+  user: User;
+};
+
+export default class OrgAccessDeniedComponent extends React.Component<Props> {
+  handleImpersonateClicked() {
+    const params = new URLSearchParams(window.location.search);
+    const sourceUrl = params.get("source_url");
+    if (sourceUrl) {
+      router.navigateTo(sourceUrl);
+    }
+    authService.enterImpersonationMode(this.props.user.subdomainGroupID);
+  }
+
+  render() {
+    return (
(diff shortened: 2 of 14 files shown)


app/auth/auth_service.ts
```diff
@@ -154,6 +154,7 @@ export class AuthService {
         )
       ),
       isImpersonating: Boolean(rpcService.requestContext.impersonatingGroupId),
+      subdomainGroupID: response.subdomainGroupId,
```
bduffany, line 157: nit: maybe this would fit better in the `config` proto since it's a property of the domain and not the user? It feels slightly error-prone to put this in the User object if the user is not necessarily authorized to access this group ID.

If we are worried about an extra DB request though in the backend, maybe this can be a field on the AuthService?

bduffany (approved):

enterprise/app/org/org_access_denied.tsx
```diff
@@ -0,0 +1,42 @@
...
+            <div className="titles">
+              <div className="title">Access denied</div>
+            </div>
+            <div className="details">You are not authorized to access this site.</div>
+            {this.props.user?.subdomainGroupID && (
+              <div>
+                <FilledButton onClick={this.handleImpersonateClicked.bind(this)} className="impersonate-button">
+                  Impersonate owner
+                </FilledButton>
+              </div>
```
bduffany, line 35: probably want to hide this behind `this.props.user?.canCall('getInvocationOwner')` so we don't show the button if the user can't impersonate. (I know the RPC-based perms feel a little clunky so if you want you could also add a method like `canImpersonate()` to the User class which just calls that under the hood for now)

app/auth/auth_service.ts
vadimberezniker: It's related to the user: `subdomainGroupId` is only populated if the user is allowed to impersonate.

enterprise/app/org/org_access_denied.tsx
vadimberezniker: subdomainGroupID is not populated if the user is not allowed to impersonate

bduffany (approved):

server/util/claims/claims.go
```diff
@@ -143,6 +144,9 @@ func ClaimsFromSubID(ctx context.Context, env environment.Env, subID string) (*C
 		for _, g := range u.Groups {
 			if g.Group.GroupID == c.GetGroupId() {
 				eg = c.GetGroupId()
+				if sd := subdomain.Get(ctx); sd != "" && (g.Group.URLIdentifier == nil || *g.Group.URLIdentifier != sd) {
```
tylerwilliams, line 147: doing a lot here

tylerwilliams (approved):

PR #4604, opened 2023-08-22
Disable credential helper for 'git fetch'

bduffany:
Disables the git credential helper for `git fetch`. It's unclear why sometimes we are getting 'invalid username or password' on certain `git fetch` commands, but this PR will let us rule out whether it's because of the cred helper.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -1546,7 +1546,7 @@ func (ws *workspace) fetch(ctx context.Context, remoteURL string, branches []str
 	if _, err := git(ctx, io.Discard, "remote", "add", remoteName, authURL); err != nil && !isRemoteAlreadyExists(err) {
 		return status.UnknownErrorf("Command `git remote add %q <url>` failed.", remoteName)
 	}
-	fetchArgs := append([]string{"fetch", "--filter=blob:none", "--force", remoteName}, branches...)
+	fetchArgs := append([]string{"-c", "credential.helper=", "fetch", "--filter=blob:none", "--force", remoteName}, branches...)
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1546,7 +1546,7 @@ func (ws *workspace) fetch(ctx context.Context, remoteURL string, branches []str
 	if _, err := git(ctx, io.Discard, "remote", "add", remoteName, authURL); err != nil && !isRemoteAlreadyExists(err) {
 		return status.UnknownErrorf("Command `git remote add %q <url>` failed.", remoteName)
 	}
-	fetchArgs := append([]string{"fetch", "--filter=blob:none", "--force", remoteName}, branches...)
+	fetchArgs := append([]string{"fetch", "--config=credential.helper=", "--filter=blob:none", "--force", remoteName}, branches...)
```
sluongng, line 1549: I think it needs to be `git --config=credential.helper= fetch` instead of `git fetch --config=credential.helper=`

`--config / -c` is a global git flag not specific to a git subcommand. We might change that as well in `git-clone` if needed.

bduffany: ah thank you, I was a little perplexed why this didn't work :)

bduffany: seems `fetch` only supports the short form (-c`) - `git --config=credential.helper=` didn't work

bduffany: The `clone` command is working fine, so will leave that as-is.

sluongng (approved):
LGTM

PR #4613, opened 2023-08-23
Don't allow using subdomains in default subdomains list.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -18,6 +18,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/query_builder"
 	"github.com/buildbuddy-io/buildbuddy/server/util/role"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/buildbuddy-io/buildbuddy/server/util/subdomain"
 
 	akpb "github.com/buildbuddy-io/buildbuddy/proto/api_key"
 	grpb "github.com/buildbuddy-io/buildbuddy/proto/group"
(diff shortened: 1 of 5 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -446,11 +469,13 @@ func (s *BuildBuddyServer) UpdateGroup(ctx context.Context, req *grpb.UpdateGrou
 	urlIdentifier := strings.TrimSpace(req.GetUrlIdentifier())
 
 	if urlIdentifier != "" {
-		if group, err = userDB.GetGroupByURLIdentifier(ctx, urlIdentifier); group != nil && group.GroupID != req.GetRequestContext().GetGroupId() {
-			return nil, status.InvalidArgumentError("URL is already in use")
-		} else if err != nil && gstatus.Code(err) != gcodes.NotFound {
+		valid, err := s.validateURLIdentifier(ctx, req, urlIdentifier)
```
bduffany, line 472: should we add this validation to `CreateGroup` as well?

also, I wonder if the validation code would fit better in userdb - for better or worse, a lot of the validation logic has ended up there (e.g. the OwnedDomain check), and it's much better tested than this buildbuddy_server package is.

vadimberezniker: good point, done

enterprise/server/backends/userdb/userdb.go
```diff
@@ -199,11 +200,47 @@ func getUserGroup(tx *db.DB, userID string, groupID string) (*tables.UserGroup,
 	return userGroup, nil
 }
 
+func (d *UserDB) validateURLIdentifier(ctx context.Context, groupID string, urlIdentifier string) (bool, error) {
```
tylerwilliams, line 203: nit: could you add a comment?

tylerwilliams (approved):

bduffany (approved):

PR #4614, opened 2023-08-23
Allow users to optionally link a GCP account

siggisim:
We request the GCP scope if the `link_gcp_for_group` url param is set in the auth flow.

If this param is set, we create secrets with the account email and refresh tokens.

### DIFF
--- a/enterprise/server/gcplink/gcplink.go
+++ b/enterprise/server/gcplink/gcplink.go
@@ -17,31 +17,35 @@ import (
 )
 
 const (
-	Issuer                      = "https://accounts.google.com"
-	scope                       = "https://www.googleapis.com/auth/cloud-platform"
-	paramName                   = "link_gcp_for_group"
-	cookieName                  = "link-gcp-for-group"
-	cookieDuration              = 1 * time.Hour
+	Issuer         = "https://accounts.google.com"
+	scope          = "https://www.googleapis.com/auth/cloud-platform"
+	linkParamName  = "link_gcp_for_group"
+	linkCookieName = "link-gcp-for-group"
+	cookieDuration = 1 * time.Hour
+	// These can be used to call gcloud auth activate-refresh-token $CLOUDSDK_AUTH_ACCOUNT $CLOUDSDK_AUTH_REFRESH_TOKEN
 	refreshTokenEnvVariableName = "CLOUDSDK_AUTH_REFRESH_TOKEN"
 	authAccountEnvVariableName  = "CLOUDSDK_AUTH_ACCOUNT"
 )
 
+// Returns true if the request contains either a gcp link url param or cookie.
 func IsLinkRequest(r *http.Request) bool {
-	return (r.URL.Query().Get(paramName) != "" && cookie.GetCookie(r, cookie.AuthIssuerCookie) == Issuer) ||
-		cookie.GetCookie(r, cookieName) != ""
+	return (r.URL.Query().Get(linkParamName) != "" && cookie.GetCookie(r, cookie.AuthIssuerCookie) == Issuer) ||
+		cookie.GetCookie(r, linkCookieName) != ""
 }
 
+// Redirects the user to the auth flow with a request for the GCP scope.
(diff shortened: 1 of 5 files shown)


siggisim:
Here's the frontend change: https://github.com/buildbuddy-io/buildbuddy/pull/4615

enterprise/server/gcplink/gcplink.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const (
+	Issuer                      = "https://accounts.google.com"
+	scope                       = "https://www.googleapis.com/auth/cloud-platform"
+	paramName                   = "link_gcp_for_group"
+	cookieName                  = "link-gcp-for-group"
+	cookieDuration              = 1 * time.Hour
+	refreshTokenEnvVariableName = "CLOUDSDK_AUTH_REFRESH_TOKEN"
+	authAccountEnvVariableName  = "CLOUDSDK_AUTH_ACCOUNT"
```
bduffany, line 26: would it make sense to prefix these with `GOOGLE`? CLOUDSDK might be ambiguous if we support other cloud providers

bduffany (commented):
LGTM but would be good to get Tyler's review as well

enterprise/server/gcplink/gcplink.go
tylerwilliams: wanna add that link here as a comment explaining that these must match so that gcloud will pick them up?

enterprise/server/gcplink/gcplink.go
```diff
@@ -0,0 +1,83 @@
...
+		cookie.GetCookie(r, cookieName) != ""
+}
+
+func RequestAccess(w http.ResponseWriter, r *http.Request, authUrl string) {
+	authUrl = strings.ReplaceAll(authUrl, "scope=", fmt.Sprintf("scope=%s+", scope))
+	cookie.SetCookie(w, cookieName, r.URL.Query().Get(paramName), time.Now().Add(cookieDuration), true)
+	http.Redirect(w, r, authUrl, http.StatusTemporaryRedirect)
+}
+
+func LinkForGroup(env environment.Env, w http.ResponseWriter, r *http.Request, email, refreshToken string) error {
```
tylerwilliams, line 40: i could live without the others but for this publicly exported method it'd be nice to have a doc string

enterprise/server/gcplink/gcplink.go
```diff
@@ -0,0 +1,83 @@
...
+	ctxpb "github.com/buildbuddy-io/buildbuddy/proto/context"
+	skpb "github.com/buildbuddy-io/buildbuddy/proto/secrets"
+	requestcontext "github.com/buildbuddy-io/buildbuddy/server/util/request_context"
+)
+
+const (
+	Issuer                      = "https://accounts.google.com"
+	scope                       = "https://www.googleapis.com/auth/cloud-platform"
+	paramName                   = "link_gcp_for_group"
+	cookieName                  = "link-gcp-for-group"
```
tylerwilliams, line 23: lol I guess i'm pedantic but i'd call these linkParamName and linkCookieName just so the code below is a little clearer (yeah i know it's in the gcplink package but it's still a little disorienting to read GetCookie(r, cookieName) and be like... what is cookieName haha)

tylerwilliams (approved):
minor nits but lgtm, thanks for making the effort to pull most of this out into a diff package

enterprise/server/gcplink/gcplink.go
siggisim: Added a comment explaining what to do with it, but it's actually not the variable from the link (the naming is just derived from it).

enterprise/server/gcplink/gcplink.go
siggisim: Added for all of 'em

enterprise/server/gcplink/gcplink.go
siggisim: Done

PR #4617, opened 2023-08-23
Don't try to persist empty artifacts to blobstore

tylerwilliams:


siggisim (approved):

vadimberezniker (approved):
In a separate PR, we should also make sure that the serve path also works in case someone tries to load an empty digest.

tempoz (approved):

PR #4620, opened 2023-08-23
Don't overwrite artifacts if they already exist

tylerwilliams:


bduffany (approved):

server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -486,6 +486,10 @@ func (r *statsRecorder) Stop() {
 }
 
 func persistArtifact(ctx context.Context, env environment.Env, uri *url.URL, path string) error {
+	alreadyExists, err := env.GetBlobstore().BlobExists(ctx, path)
+	if err == nil && alreadyExists {
+		return nil
+	}
```
bduffany, line 492: this seems ok but not sure if we'll ever hit this case since I think we store these per-invocation?

tempoz:
I understand that we ended up not needing to do this in this case, but in case we ever do need to this sort of non-overwriting write in the future, it's important to know that this style of "check for existence and then write"-based write introduces a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) (time-of-check to time-of-use) error. Since it ended up being a pain to figure out how to do it right in this case, I'm leaving this comment here so that no one has to figure this out again if we decide we need something like this.

For blobstore, we should instead leverage:
- For GCS: the [`ObjectHandle.If`](https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.If) GCP call with [`Conditions.DoesNotExist`](https://pkg.go.dev/cloud.google.com/go/storage#Conditions) 
- For AWS: the [`if-none-match`](https://stackoverflow.com/questions/12654828/amazon-s3-avoid-overwriting-objects-with-the-same-name) header, by creating a struct implementing [`BuildMiddleware`](https://pkg.go.dev/github.com/aws/smithy-go/middleware#BuildMiddleware) setting that header to `*` and then [`Add`](https://pkg.go.dev/github.com/aws/smithy-go/middleware#BuildStep.Add)ing it to the [`Stack.Build`](https://pkg.go.dev/github.com/aws/smithy-go/middleware#Stack) step via a function with the signature [`func(*middleware.Stack) error`](https://pkg.go.dev/github.com/aws/smithy-go/middleware#Stack), and then appending that function to [`Options.APIOptions`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#Options) via [`Uploader.ClientOptions`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#Uploader) via the `opts...` of type `func(*Uploader)` in the arguments to [`Uploader.Upload`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#Uploader.Upload). An existing blob will return a `403`. If that sounds complicated, it's because it is, but basically it will look like:
```go
import(
  "github.com/aws/aws-sdk-go-v2/service/s3"
  "github.com/aws/smithy-go/middleware"

  s3manager "github.com/aws/aws-sdk-go-v2/feature/s3/manager"
  smithyhttp "github.com/aws/smithy-go/transport/http"
)

...

type struct NoOverwrite {}

func (n *NoOverwrite) ID() string {
}

func (n *NoOverwrite) HandleBuild(ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler) (middleware.BuildOutput, middleware.Metadata, error) {
  req, ok := in.Request.(*smithyhttp.Request)
  if !ok {
    return middleware.BuildOutput{}, middleware.Metadata{}, fmt.Errorf("unexpected request middleware type %T", in.Request)
  }
  req.Request.Header.Add("if-none-match", "*")
  return next.HandleBuild(ctx, in)
}

...

u.Upload(ctx, input, func(u *s3manager.Uploader) {
  u.ClientOptions = append(u.ClientOptions, func(o *s3.Options) {
    o.APIOptions = append(o.APIOptions, func(s *middleware.Stack) {
      s.Build.Add(
        &NoOverwrite{},
        middleware.After,
      ),
    }),
  }),
})
```
- For Azure: the `if-none-match` header again, albeit indirectly. When calling [`UploadBufferToBlockBlob`](https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob#UploadBufferToBlockBlob), in [`UploadToBlockBlobOptions.AccessConditions`](https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob#UploadToBlockBlobOptions), in [`BlobAccessConditions.ModifiedAccessConditions`](https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob#BlobAccessConditions) , we set [`ModifiedAccessConditions.IfNoneMatch`](https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob#ModifiedAccessConditions) to [`ETagAny`](https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob#ETag). THis will once again return a `403` if the file already exists.
- For disk: Use [`O_CREATE | O_EXCL`](https://pkg.go.dev/os#pkg-constants) when calling [`OpenFile`](https://pkg.go.dev/os#OpenFile).

PR #4626, opened 2023-08-24
Don't use digest size passed from client for decryptor.

vadimberezniker:
**Related issues**: N/A


tylerwilliams (approved):
Thanks for adding a test!

PR #4627, opened 2023-08-24
Only write secrets that have been set in repo flow

siggisim:
**Related issues**: N/A

### DIFF
--- a/enterprise/app/repo/repo.tsx
+++ b/enterprise/app/repo/repo.tsx
@@ -199,6 +199,12 @@ export default class RepoComponent extends React.Component<RepoComponentProps, R
     return s ? s.split(",") : [];
   }
 
+  getUnsetSecrets() {
+    return this.getSecrets()
+      .filter((s) => !this.state.secretsResponse?.secret.map((s) => s.name).includes(s))
+      .filter((s) => gcpRefreshTokenKey != s);
+  }
+
   async handleCreateClicked() {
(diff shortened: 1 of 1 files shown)


enterprise/app/repo/repo.tsx
```diff
@@ -199,6 +199,12 @@ export default class RepoComponent extends React.Component<RepoComponentProps, R
     return s ? s.split(",") : [];
   }
 
+  getUnsetSecrets() {
+    return this.getSecrets()
+      .filter((s) => !this.state.secretsResponse?.secret.map((s) => s.name).includes(s))
+      .filter((s) => gcpRefreshTokenKey != s);
```
bduffany, line 205: took me a minute to make sense of this line, so maybe worth a comment -

```suggestion
      // Exclude the GCP refresh token from secrets that need to be set
      // up manually by the user, since we set this up as part of GCP
      // account linking.
      .filter((s) => gcpRefreshTokenKey != s);
```

after writing this out though, it feels a bit confusing to me - why do we use a secret for the GCP account link instead of a plain old table like we do for oauth?

bduffany (approved):

enterprise/app/repo/repo.tsx
siggisim: I think that works alright for GCP, but want to be able to support arbitrary deployment targets - i.e AWS, fly.io, TestFlight, etc.

PR #4628, opened 2023-08-24
commandutil refactoring to support windows

sluongng:
Refactor how commandutil package start and kill a process.
In particular, introduce a wrapper struct `process` to store additional information about the command execution.
This new struct provides several methods, acting as hooks into different stages of the command execution:
- preStart
- start
- postStart
- monitor
- wait
- cleanUp

Some of these hooks could be overridden to implement platform-specific behaviors.
For example, we customize Windows `preStart` and `postStart` to initialize a Windows' Job Object and add the started process to the Job Object.
At the `cleanUp` stage, we close the handle on the Job Object to terminate all sub-processes.


enterprise/server/remote_execution/commandutil/commandutil_windows.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (p *process) postStart() error {
+	processHandle, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.PROCESS_TERMINATE, false, uint32(p.cmd.Process.Pid))
+	if err != nil {
+		return nil
+	}
+	defer windows.CloseHandle(processHandle)
+
+	return windows.AssignProcessToJobObject(p.jobHandle, processHandle)
```
bduffany, line 38: does this work if the process has exited by the time we get here? (can test this by adding a 100 ms delay before calling this, and running a NOP process - whatever the `sleep 0` equivalent is on Windows or something)

I wonder if there is a cleaner way to do this, like assign the process to the job at creation time or something

enterprise/server/remote_execution/commandutil/commandutil_windows.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func createWindowsJobObject() (windows.Handle, error) {
+	job, err := windows.CreateJobObject(nil, nil)
+	if err != nil {
+		return windows.InvalidHandle, fmt.Errorf("failed to create job object: %w", err)
```
bduffany, line 44: can this be nil instead of InvalidHandle? (same below)

enterprise/server/remote_execution/commandutil/commandutil_windows.go
```diff
@@ -0,0 +1,89 @@
...
+// an error along the way, it proceeds to kill subsequent pids in the tree. It
+// returns the last error encountered, if any.
+//
+// For Windows, see
+// https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects
+// and https://learn.microsoft.com/en-us/windows/win32/procthread/nested-jobs
+// for more details.
+func (p *process) killProcessTree() error {
+	return windows.CloseHandle(p.jobHandle)
+}
```
bduffany, line 72: wow, this part is much simpler than unix :)

does it work even if subprocesses "fork" into their own jobs? I think that's the reason I added the "KillProcessTree" thing to begin with - on unix, processes could fork into their own process group, and then killing the process group would not properly clean up all processes.

enterprise/server/remote_execution/commandutil/commandutil_windows.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return job, nil
+}
+
+// killProcessTree kills the given pid as well as any descendant processes.
+//
+// It tries to kill as many processes in the tree as possible. If it encounters
+// an error along the way, it proceeds to kill subsequent pids in the tree. It
+// returns the last error encountered, if any.
+//
```
bduffany, line 65: nit: remove this part since this is more describing the unix-specific behavior

enterprise/server/remote_execution/commandutil/commandutil_unix.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// LookupCredential adds credentials to the cmd by resolving a "USER[:GROUP]" string
+// to a credential with both uid and gid populated. Both numeric IDs and non-numeric
+// names can be  specified for either USER or GROUP. If no group is specified, then
+// the user's primary group is used.
+//
+// NOTE: This function does not authenticate that the user is part of the
+// specified group.
+func LookupCredential(cmd *exec.Cmd, spec string) error {
```
bduffany, line 77: nit: maybe call it `setCredential` since it's now setting the credential on the cmd instead of returning it?

enterprise/server/remote_execution/commandutil/commandutil_windows.go
sluongng: Done

enterprise/server/remote_execution/commandutil/commandutil_unix.go
sluongng: Done.

enterprise/server/remote_execution/commandutil/commandutil_windows.go
sluongng: This is a really good call.

I tested it and it failed with sleep.

I think the correct way to do this would be to create a suspended process, assign it to a job object, and then resume the process. https://devblogs.microsoft.com/oldnewthing/20130405-00/?p=4743
Let me look into fixing it.

sluongng: It's fixed

enterprise/server/remote_execution/commandutil/commandutil_windows.go
sluongng: > Operations performed on a job object affect all processes associated with the job object. Examples include enforcing limits such as working set size and process priority or terminating all processes associated with a job.

From https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects

It's also used by other CI vendors such as https://github.com/buildkite/agent/blob/8faa01f13fd92d8a730f1c04bf457ead05c173b1/process/signal_windows.go#L76-L79

I did not set up a specific test for it, mostly because I don't wanna dive too deep into PowerShell arcane magic. But I did put some basic tests in place for us to replicate in the future if anything needs to be reproduced.

sluongng:
@bduffany PTAL 🙇

sluongng:
Worth noting that this depends on the future revert of the revert in https://github.com/buildbuddy-io/buildbuddy/pull/4938

Because we do need platform-specific config for windows.

bduffany (approved):

PR #4629, opened 2023-08-24
Monitor route '/api/v1/metrics' 

luluz66:
**Related issues**: N/A


bduffany (approved):

server/http/interceptors/interceptors.go
```diff
@@ -255,6 +255,9 @@ func routeLabel(r *http.Request) string {
 	if strings.HasPrefix(path, "/invocation/") {
 		return "/invocation/[...]"
 	}
+	if strings.HasPrefix(path, "/api/v1/metrics") {
```
bduffany, line 258: nit:

```suggestion
	if path == "/api/v1/metrics" {
```

PR #4631, opened 2023-08-24
Don't include ToolDetails in executor -> app request metadata

bduffany:
This causes the app to incorrectly identify requests as being from Bazel.

This is problematic for usage labeling, but also we have a few other places in the app that call `bazel_request.GetVersion()` and then do version-specific behavior (bytestream short-circuiting and AC write capabilities), when in reality they should not be treating the client as bazel at all.

An alternative approach is to update all these callsites to explicitly check for ExecutorDetails, but that feels a bit more error prone, as any other callsites we add in the future would have to care about the executor case. Plus, the app shouldn't care about which bazel client started an Action, especially since Action merging is a thing, so the action result will potentially be delivered to multiple different bazel clients. So it seems simpler to stop sending up ToolDetails to the app.

**Related issues**: N/A


tylerwilliams (approved):
in `interceptors.go` we propagate                 "build.bazel.remote.execution.v2.requestmetadata-bin" from header -> ctx and ctx -> header. Does that need to change at all or is this enough?

bduffany:
> in `interceptors.go` we propagate "build.bazel.remote.execution.v2.requestmetadata-bin" from header -> ctx and ctx -> header. Does that need to change at all or is this enough?

For what I'm trying to fix in this PR, we shouldn't need to change `interceptors.go` because the RequestMetadata used for the executor -> app requests comes only from the [ExecutionTask](https://github.com/buildbuddy-io/buildbuddy/blob/3d34429c494b62a13de6e4a320effbad9990229d/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go#L300), which is created by the execution server.

Eventually though I think it would be nice to only propagate the invocation ID from ctx -> header (outgoing RPCs) since all of the other metadata is pure overhead currently (i.e. just a simple `x-invocation-id` header or something so we don't even have to deserialize any protos)

tylerwilliams:
> > in `interceptors.go` we propagate "build.bazel.remote.execution.v2.requestmetadata-bin" from header -> ctx and ctx -> header. Does that need to change at all or is this enough?
> 
> For what I'm trying to fix in this PR, we shouldn't need to change `interceptors.go` because the RequestMetadata used for the executor -> app requests comes only from the [ExecutionTask](https://github.com/buildbuddy-io/buildbuddy/blob/3d34429c494b62a13de6e4a320effbad9990229d/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go#L300), which is created by the execution server.
> 
> Eventually though I think it would be nice to only propagate the invocation ID (and nothing else) from ctx -> header (outgoing RPCs) since all of the other metadata is pure overhead currently at least for the apps - i.e. just a simple `x-invocation-id` header or something so we don't have to do any deserialization for internal app -> app requests.

I'm pro making this change :) The less we have to carry around on every request the better.

PR #4635, opened 2023-08-24
Hide "go to org" and "create org" on subdomain UI.

vadimberezniker:
**Related issues**: N/A


siggisim (commented):
Should we just have these take you to the new subdomain when you switch orgs?

PR #4637, opened 2023-08-24
Fix UFFDIO_COPY size within last memory chunk

bduffany:
In my local testing this fixes the "possible stuck request" issue that we've seen with `firecracker_test_blockio`.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/uffd/uffd.go
+++ b/enterprise/server/remote_execution/uffd/uffd.go
@@ -268,11 +268,6 @@ func resolvePageFault(uffd uintptr, faultingRegion uint64, src uint64, size uint
 	}
 	_, _, errno := syscall.Syscall(syscall.SYS_IOCTL, uffd, UFFDIO_COPY, uintptr(unsafe.Pointer(&copyData)))
 	if errno != 0 {
-		if errno == unix.ENOENT {
-			// The faulting process changed its virtual memory layout simultaneously with an outstanding UFFDIO_COPY
-			// The UFFDIO_COPY will have aborted and the handler should continue to serve page faults
-			return 0, nil
-		}
 		return 0, status.InternalErrorf("UFFDIO_COPY failed with errno(%d)", errno)
 	}
 	return int64(copyData.Copy), nil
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -268,11 +268,6 @@ func resolvePageFault(uffd uintptr, faultingRegion uint64, src uint64, size uint
 	}
 	_, _, errno := syscall.Syscall(syscall.SYS_IOCTL, uffd, UFFDIO_COPY, uintptr(unsafe.Pointer(&copyData)))
 	if errno != 0 {
-		if errno == unix.ENOENT {
-			// The faulting process changed its virtual memory layout simultaneously with an outstanding UFFDIO_COPY
-			// The UFFDIO_COPY will have aborted and the handler should continue to serve page faults
-			return 0, nil
-		}
```
tylerwilliams, line 275: i'm not sure we want to remove this? 

https://man7.org/linux/man-pages/man2/ioctl_userfaultfd.2.html

bduffany: I have a feeling that these were only happening because of the last-chunk issue, but will wait for @maggie-lou to weigh in since I think she added this check originally. In my testing it seems like these errors are unrecoverable and so we shouldn't drop them silently or else things will just hang rather than failing.

tylerwilliams: Maybe leave a comment here, and/or catch this error and return a meaningful failure back up the stack so we know when something is really borked and what happened rather than having to puzzle it out?

bduffany: It gets caught on the following line (271) - now we treat it the same as other errors, rather than checking for ENOENT specifically

tylerwilliams: sgtm

tylerwilliams (approved):
LGTM

enterprise/server/remote_execution/uffd/uffd.go
bduffany: Will go ahead and submit this for now - if we hit ENOENT in the future and we're sure that the right thing to do is silently drop it, we can revert this specific diff chunk.

PR #4638, opened 2023-08-24
Fix default group selection for subdomains.

vadimberezniker:
Perform the selection logic after filtering the list of groups.

**Related issues**: N/A

### DIFF
--- a/server/buildbuddy_server/buildbuddy_server.go
+++ b/server/buildbuddy_server/buildbuddy_server.go
@@ -231,7 +231,7 @@ func (s *BuildBuddyServer) CancelExecutions(ctx context.Context, req *inpb.Cance
 	return &inpb.CancelExecutionsResponse{}, nil
 }
 
-func filterGroupRoles(ctx context.Context, groupRoles []*tables.GroupRole) []*tables.GroupRole {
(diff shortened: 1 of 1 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -231,14 +231,22 @@ func (s *BuildBuddyServer) CancelExecutions(ctx context.Context, req *inpb.Cance
 	return &inpb.CancelExecutionsResponse{}, nil
 }
 
-func makeGroups(ctx context.Context, groupRoles []*tables.GroupRole) []*grpb.Group {
-	r := make([]*grpb.Group, 0)
+func filterGroupRoles(ctx context.Context, groupRoles []*tables.GroupRole) []*tables.GroupRole {
```
tylerwilliams, line 234: could we call this filterGroupsToSubdomain or something so it's clearer what it's filtering to?

(In general there's a lot of groups-specific code in this file now which maybe needs a better group-specific home...)

tylerwilliams (approved):

server/buildbuddy_server/buildbuddy_server.go
vadimberezniker: done

PR #4642, opened 2023-08-25
ci_runner,executor: remove os-specific syscall

sluongng:
Replace the os-specific syscall in ci_runner and executor by using an os-specific source file, or by switching to an alternative implementation.


sluongng:
Rebase on top of origin/master without the changes in #4628 
This should allow this PR to land earlier while I find a proper way to test the other one.

bduffany (approved):

PR #4643, opened 2023-08-25
Clean up net namespaces in firecracker test

bduffany:
If tests get interrupted for whatever reason, then the `t.Cleanup()` might not run and clean up the networking stuff associated with the VM. This PR makes sure we can recover from that - this is especially important for CI where we can't easily run `tools/firecracker_clean_networking.sh` - occasionally I have to run this locally to recover from aborted tests.

**Related issues**: N/A


enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -105,6 +106,10 @@ func getTestEnv(ctx context.Context, t *testing.T) *testenv.TestEnv {
 	err := os.Setenv("REGISTRY_AUTH_FILE", filepath.Join(tmp, "auth.json"))
 	require.NoError(t, err)
 
+	// Clean up any lingering networking changes from previous test runs.
+	err = networking.DeleteNetNamespaces(ctx)
```
tylerwilliams, line 110: is this safe? I would think this would introduce a data race if multiple fc tests happen to be running concurrently on the same machine, but maybe I haven't thought it through all the way.

bduffany: The firecracker tests have to run exclusively today because of our "vmIdx" thing that starts at 0 and assigns IPs deterministically from the index. (The bare pool is configured with exclusive_task_scheduling).

tylerwilliams: I don't think we want to add more code that depends on firecracker tests being run exclusively -- sort of going in the wrong direction you know? But I'm happy to approve this temporarily if you:
- add a TODO to remove it
- ideally also take a look in a followup at fixing the vmIdx stuff in a better way?

bduffany: adding a TODO for now - will try to make the vmIdx stuff concurrency-safe (across executor processes) in a separate PR

tylerwilliams (approved):

PR #4647, opened 2023-08-25
Add a route to select a specific group ID.

vadimberezniker:
/org/GR123?path=/foo will set the selected group to GR123 and redirect to /foo.

When switching groups on a subdomain, we need a way to redirect to a specific group on the app subdomain if the group does not have a URL identifier set.

**Related issues**: N/A

### DIFF
--- a/enterprise/app/root/root.tsx
+++ b/enterprise/app/root/root.tsx
@@ -94,6 +94,12 @@ export default class EnterpriseRootComponent extends React.Component {
 
   componentDidMount() {
     errorService.register();
+    if (this.state.path.startsWith("/org/GR")) {
+      const groupId = window.location.pathname.split("/").pop();
+      if (groupId && groupId.startsWith("GR")) {
+        authService.redirectToGroupId(groupId);
+      }
+    }
   }
 
   handlePathChange() {
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -176,6 +176,18 @@ export class AuthService {
     rpcService.requestContext.groupId = this.user?.selectedGroup?.id || "";
   }
 
+  redirectToGroupId(groupId: string) {
+    const newURL = new URL(window.location.href)
+    const path = newURL.searchParams.get("path")
+    if (!path) {
+      throw new Error("path parameter missing")
+    }
+    newURL.searchParams.delete("path")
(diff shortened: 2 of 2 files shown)


enterprise/app/root/root.tsx
```diff
@@ -94,6 +94,12 @@ export default class EnterpriseRootComponent extends React.Component {
 
   componentDidMount() {
     errorService.register();
+    if (this.state.path.startsWith("/org/GR")) {
+      const groupId = window.location.pathname.split("/").pop();
+      if (groupId && groupId.startsWith("GR")) {
+        authService.redirectToGroupId(groupId);
+      }
+    }
```
bduffany, line 102: It feels a bit odd to fully render the <Root> component, only to have it do a redirect once it has mounted. Switching groups always triggers a full page reload / server-side navigation today, so it might be cleaner to handle this redirect on the server?

vadimberezniker: How would we set the selected group ID if we do the redirect server-side?

vadimberezniker: What if we put the selected group ID into a cookie instead of local storage? Was there a reason we chose to store it inside local storage in the first place?

bduffany: Eventually I think putting the group ID in a cookie would be a good idea so that we always have it available on the initial page request. That would let us do some nice things like server-side data prefetching.

Separate from that though, I'm not sure I understand this part

> When switching groups on a subdomain, we need a way to redirect to a specific group on the app subdomain if the group does not have a URL identifier set.

By "switching groups on a subdomain" do you mean switching using the group picker? Does the current `setSelectedGroupId` func work for that? (Instead of `window.location.reload` can we set window.location = `/group?id=${groupID}&dest=${encodeURIComponent(currentURLExcludingDomain())}`)

vadimberezniker: > Does the current `setSelectedGroupId` func work for that? (Instead of `window.location.reload` can we set window.location = `/group?id=${groupID}&dest=${encodeURIComponent(currentURLExcludingDomain())}`)

setSelectedGroupId relies on local storage which is isolated to the current site (subdomain). If you're on foo.buildbuddy.io and set the group ID, then redirect to app.buildbuddy.io the group ID won't be visible. If we use cookies, we can pass information across subdomains.

bduffany: Gotcha. I think using a cookie for the selected group ID would be ideal.

PR #4649, opened 2023-08-26
Store selected group ID in cookie instead of local storage.

vadimberezniker:
Having the group ID in a cookie will allow group switching to work properly between subdomain and app w/o having to resort to additional hacks.

**Related issues**: N/A

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -27,13 +27,14 @@ export class AuthService {
   register() {
     if (!capabilities.auth) return;
     // Set initially preferred group ID from cookie.
-    const preferredGroupId = this.getCookie(SELECTED_GROUP_ID_COOKIE) ||
-        window.localStorage.getItem(SELECTED_GROUP_ID_LOCAL_STORAGE_KEY) ||
-        "";
+    const preferredGroupId =
+      this.getCookie(SELECTED_GROUP_ID_COOKIE) ||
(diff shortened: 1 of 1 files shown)


bduffany (approved):

vadimberezniker:
I'll wait for @siggisim's LGTM as well, just in case.

app/auth/auth_service.ts
```diff
@@ -24,8 +26,11 @@ export class AuthService {
 
   register() {
     if (!capabilities.auth) return;
-    // Set initially preferred group ID from local storage.
-    rpcService.requestContext.groupId = window.localStorage.getItem(SELECTED_GROUP_ID_LOCAL_STORAGE_KEY) || "";
+    // Set initially preferred group ID from cookie.
+    rpcService.requestContext.groupId =
+      this.getCookie(SELECTED_GROUP_ID_COOKIE) ||
+      window.localStorage.getItem(SELECTED_GROUP_ID_LOCAL_STORAGE_KEY) ||
+      "";
```
bduffany, line 33: (thought of this after approving)

we might want to proactively populate the cookie after this line, so that if we remove the localStorage fallback later, there's a better chance of the cookie being available.

siggisim (approved):
Agree with Brandon - might be nice to set the cookie if not set to make the crossover smoother

PR #4650, opened 2023-08-28
platform: make ParseProperties return error

sluongng:
This is a follow up from #4575.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/tasksize/tasksize.go
+++ b/enterprise/server/tasksize/tasksize.go
@@ -121,7 +121,12 @@ func (s *taskSizer) Get(ctx context.Context, task *repb.ExecutionTask) *scpb.Tas
 	if !*useMeasuredSizes {
 		return nil
 	}
-	props := platform.ParseProperties(task)
+	props, err := platform.ParseProperties(task)
+	if err != nil {
+		// TODO(sluongng): reject tasks that fail validation so users could catch errors sooner
+		log.CtxInfof(ctx, "Failed to parse task properties: %s", err)
+		return nil
+	}
 	// If a task size is explicitly requested, measured task size is not used.
 	if props.EstimatedComputeUnits != 0 {
 		return nil
@@ -178,7 +183,10 @@ func (s *taskSizer) Update(ctx context.Context, cmd *repb.Command, md *repb.Exec
 	}
 	statusLabel := "ok"
 	defer func() {
-		props := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+		props, err := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+		if err != nil {
+			log.CtxWarningf(ctx, "Failed to parse task properties: %s", err)
+		}
 		groupID, _ := s.groupKey(ctx)
 		metrics.RemoteExecutionTaskSizeWriteRequests.With(prometheus.Labels{
 			metrics.TaskSizeWriteStatusLabel: statusLabel,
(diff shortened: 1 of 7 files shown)


enterprise/server/tasksize/tasksize.go
```diff
@@ -178,7 +182,10 @@ func (s *taskSizer) Update(ctx context.Context, cmd *repb.Command, md *repb.Exec
 	}
 	statusLabel := "ok"
 	defer func() {
-		props := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+		props, err := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+		if err != nil {
+			log.CtxWarningf(ctx, "Failed to parse task properties: %s", err)
+		}
```
sluongng, line 188: This will mostly be a no-op, but adding error handling here just to future proof

enterprise/server/tasksize/tasksize.go
```diff
@@ -322,7 +329,16 @@ func testSize(testSize string) (int64, int64) {
...
-	props := platform.ParseProperties(task)
+	props, err := platform.ParseProperties(task)
+	if err != nil {
+		log.Warningf("Failed to parse task properties, using default estimation: %s", err)
+
+		return applyMinimums(&scpb.TaskSize{
+			EstimatedMemoryBytes:   DefaultMemEstimate,
+			EstimatedMilliCpu:      DefaultCPUEstimate,
+			EstimatedFreeDiskBytes: DefaultFreeDiskEstimate,
+		})
```
sluongng, line 340: This skips the "less than Max value" check and relies on the default value to always be less than Max.

enterprise/server/tasksize/tasksize.go
```diff
@@ -121,7 +121,11 @@ func (s *taskSizer) Get(ctx context.Context, task *repb.ExecutionTask) *scpb.Tas
 	if !*useMeasuredSizes {
 		return nil
 	}
-	props := platform.ParseProperties(task)
+	props, err := platform.ParseProperties(task)
+	if err != nil {
+		log.CtxWarningf(ctx, "Failed to parse task properties: %s", err)
```
bduffany, line 126: let's switch these to `log.Infof()` since they are user errors - I think we are starting to move towards treating Warnings as real issues, e.g. w/ the app canary process

also can you add a TODO to reject the task in this case? we probably shouldn't accept tasks that fail validation, so that users can catch errors soooner.

enterprise/server/tasksize/tasksize.go
```diff
@@ -322,7 +329,16 @@ func testSize(testSize string) (int64, int64) {
 // from the task such as test size and estimated compute units, but does not use
 // information about historical task executions.
 func Estimate(task *repb.ExecutionTask) *scpb.TaskSize {
-	props := platform.ParseProperties(task)
+	props, err := platform.ParseProperties(task)
+	if err != nil {
+		log.Warningf("Failed to parse task properties, using default estimation: %s", err)
```
bduffany, line 334: Let's make this a log.Info too, otherwise it'll affect app rollout if people are using invalid values

enterprise/server/tasksize/tasksize.go
sluongng: fixed

enterprise/server/tasksize/tasksize.go
sluongng: fixed

enterprise/server/tasksize_model/tasksize_model.go
```diff
@@ -193,7 +193,11 @@ func (m *Model) Predict(ctx context.Context, task *repb.ExecutionTask) *scpb.Tas
 
 	// Don't use predicted task sizes for Firecracker tasks for now, since task
 	// sizes are used as hard limits on allowed resources.
-	props := platform.ParseProperties(task)
+	props, err := platform.ParseProperties(task)
+	if err != nil {
+		log.Warningf("Failed to parse task properties: %s", err)
```
bduffany, line 198:
```suggestion
		log.Infof("Failed to parse task properties: %s", err)
```

enterprise/server/tasksize/tasksize.go
```diff
@@ -178,7 +183,10 @@ func (s *taskSizer) Update(ctx context.Context, cmd *repb.Command, md *repb.Exec
 	}
 	statusLabel := "ok"
 	defer func() {
-		props := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+		props, err := platform.ParseProperties(&repb.ExecutionTask{Command: cmd})
+		if err != nil {
+			log.CtxWarningf(ctx, "Failed to parse task properties: %s", err)
```
bduffany, line 188: (forgot to push changes?)

```suggestion
			log.CtxInfof(ctx, "Failed to parse task properties: %s", err)
```

bduffany (approved):

bduffany:
my mistsake, didn't realize auto-merge was on. Will send a fix.

PR #4653, opened 2023-08-28
docs: document env-overrides* headers

sluongng:
Provide documentation for the env-overrides headers as a way to supply
ephemeral secrets to builds.

**Related issues**: N/A

### DIFF
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -72,8 +72,8 @@ as normal Bazel actions shown above.
 
 For secrets that have a short Time To Live (TTL), BuildBuddy supports setting
 environment variables via special headers passed at the Bazel command line.
-Using `env-overrides` is more secure than setting environment variables with Bazel,
-as it prevents the secrets from being stored in the remote cache.
+Headers are more secure than setting environment variables with Bazel,
+as they are not stored in the remote cache.
 
 ```bash
 # For simple secrets
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -72,7 +72,8 @@ as normal Bazel actions shown above.
 
 For secrets that have a short Time To Live (TTL), BuildBuddy supports setting
 environment variables via special headers passed at the Bazel command line.
-
+Using `env-overrides` is more secure than setting environment variables with Bazel,
+as it prevents the secrets from being stored in the remote cache.
 
 ```bash
 # For simple secrets
@@ -115,10 +116,12 @@ These secrets will be set as environment variables at action execution time,
 overriding the default environment variables on your container image as well as
 the environment variables set by Bazel as part of the action configuration.
 
-Using `env-overrides` is more secure than setting environment variables with Bazel,
-as it prevents the secrets from being stored in the remote cache. However,
-these secrets *may* be cached as part of action results if not properly handled.
-For example, avoid printing secret values to the command line.
+:::warning
+
+Secrets may be cached as part of action results if not properly handled.
+Avoid printing secret values to the console or storing them in action outputs.
+
+:::
 
 Secrets that are passed through `env-overrides` or `env-overrides-base64` headers
 are not subjected to `include-secrets` control documented above.
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -68,6 +68,52 @@ repository.
 Workflow secrets are accessed via environment variables, in the same way
 as normal Bazel actions shown above.
 
+## Sort-lived secrets
+
+For secrets that have a short Time To Live(TTL) duration, or ephemeral secrets,
+BuildBuddy provides special headers that could be set via Bazel command line.
+
+```bash
+# For simple secrets
+--remote_exec_header=x-buildbuddy-platform.env-overrides=VAR_A=value_a,VAR_B=val_b
+
+## At execution:
+> echo $VAR_A
+value_a
+> echo $VAR_B
+val_b
+
+---
+
+# For more complex secrets
+
+## First encode the secrets using base64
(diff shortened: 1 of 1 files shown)


docs/secrets.md
```diff
@@ -68,6 +68,52 @@ repository.
 Workflow secrets are accessed via environment variables, in the same way
 as normal Bazel actions shown above.
 
+## Sort-lived secrets
+
+For secrets that have a short Time To Live(TTL) duration, or ephemeral secrets,
+BuildBuddy provides special headers that could be set via Bazel command line.
```
bduffany, line 74:
```suggestion
## Short-lived secrets

For secrets that have a short Time To Live (TTL), BuildBuddy supports setting
environment variables via special headers passed at the Bazel command line.

```

docs/secrets.md
```diff
@@ -68,6 +68,52 @@ repository.
...
+> echo $VAR_A
+value_a
+> echo $VAR_B
+val_b
+
+---
+
+# For more complex secrets
+
+## First encode the secrets using base64
```
bduffany, line 90:
```suggestion
## First encode the secrets using base64,
## making sure to include the entire 'KEY=VALUE' pair
```

docs/secrets.md
```diff
@@ -68,6 +68,52 @@ repository.
...
+## Sort-lived secrets
+
+For secrets that have a short Time To Live(TTL) duration, or ephemeral secrets,
+BuildBuddy provides special headers that could be set via Bazel command line.
+
+```bash
+# For simple secrets
+--remote_exec_header=x-buildbuddy-platform.env-overrides=VAR_A=value_a,VAR_B=val_b
+
+## At execution:
```
bduffany, line 80:
```suggestion
## At execution time:
```

docs/secrets.md
```diff
@@ -68,6 +68,52 @@ repository.
...
+# For more complex secrets
+
+## First encode the secrets using base64
+> echo -n 'VAR_C={"a": 1, "b", 2}' | base64
+> echo -n 'VAR_D=asdfa!@#C,+{}' | base64
+
+## then use the base64-encoded strings in the `env-overrides-base64` header, comma separated.
+--remote_exec_header=x-buildbuddy-platform.env-overrides-base64=VkFSX0M9eyJhIjogMSwgImIiLCAyfQ==,VkFSX0Q9YXNkZmEhQCNDLCt7fQ==
+
+## At execution:
```
bduffany, line 97:
```suggestion
## At execution time:
```

docs/secrets.md
```diff
@@ -68,6 +68,52 @@ repository.
...
+> echo $VAR_C
+{"a": 1, "b", 2}
+> echo $VAR_D
+asdfa!@#C,+{}
+```
+
+> Note: If 2 values are given to the same variables, the last value will be used.
+> `env-overrides` values will be applied BEFORE `env-overrides-base64` and thus,
+> the latter would take priority.
```
bduffany, line 106: I think in docusaurus you can use `:::note` and it will be rendered a bit more nicely - I guess this is rendered as a blockquote currently

```suggestion
:::note

If multiple values are given with the same variable name, the last value will be used.
If a variable is specified in both `env-overrides` and `env-overrides-base64`,
`env-overrides-base64` will take priority.

:::
```

docs/secrets.md
```diff
@@ -68,6 +68,52 @@ repository.
...
+
+These secrets will be set as environment variables at Actions' Execution Time,
+overriding the default environment variables on your container image as well as
+the environment variables set in your Action's protobuf object. For this reason,
+using these secrets does not affect Action Cache Entry while could affect Action's
+result and could lead to Cache poisoning if miss used.
```
bduffany, line 112:
```suggestion
These secrets will be set as environment variables at action execution time,
overriding the default environment variables on your container image as well as
the environment variables set by Bazel.

Using `env-overrides` is more secure than setting environment variables with Bazel,
as it prevents the secrets from being stored in the remote cache. However,
these secrets *may* be cached as part of action results if not properly handled.
For example, avoid printing secret values to the command line.
```

bduffany (approved):
LGTM - some spelling nits and minor grammar nits

docs/secrets.md
bduffany: Also I would maybe move the note about the security benefit up to the first paragraph - otherwise people might not understand why they would use these.

```
For secrets that have a short Time To Live (TTL), BuildBuddy supports setting
environment variables via special headers passed at the Bazel command line.
Using these headers is more secure than setting environment variables with Bazel,
as environment variables are stored in the remote cache while headers are not.
```

then maybe down here, we can just have this warning

```
:::warning

Secrets *may* be cached as part of action results if not properly handled.
Avoid printing secret values to the console or storing them in action outputs.

:::
```

docs/secrets.md
sluongng: Oh nice! this is exactly what I was looking for.

docs/secrets.md
```diff
@@ -68,6 +68,64 @@ repository.
 Workflow secrets are accessed via environment variables, in the same way
 as normal Bazel actions shown above.
 
+## Short-lived secrets
+
+For secrets that have a short Time To Live (TTL), BuildBuddy supports setting
+environment variables via special headers passed at the Bazel command line.
+Using `env-overrides` is more secure than setting environment variables with Bazel,
+as it prevents the secrets from being stored in the remote cache.
```
bduffany, line 76:
```suggestion
Headers are more secure than setting environment variables with Bazel,
as they are not stored in the remote cache.
```

bduffany (approved):

PR #4574, opened 2023-08-17
[Firecracker] Add support to compute full digest of snapshot files

maggie-lou:
This fixes a race condition with local snapshot sharing (described at https://github.com/buildbuddy-io/buildbuddy/pull/4532) and will be needed for remote snapshot sharing

It also adds unit test coverage to ensure that when UFFD is enabled, the snaploader doesn't overwrite memory snapshot files it shouldn't

PR #4654, opened 2023-08-28
Allow new firecracker runners to start from snapshots

bduffany:
Building off of https://github.com/buildbuddy-io/buildbuddy/pull/4532 and https://github.com/buildbuddy-io/buildbuddy/pull/4574, this PR makes it so that runners can start from snapshot without having to be matched to a runner in the runner pool.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -200,6 +200,15 @@ func (l *FileCacheLoader) GetSnapshot(ctx context.Context, key *fcpb.SnapshotKey
 	if err := proto.Unmarshal(buf, manifest); err != nil {
 		return nil, status.UnavailableErrorf("failed to unmarshal snapshot manifest: %s", status.Message(err))
 	}
+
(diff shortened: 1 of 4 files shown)


maggie-lou (approved):

tylerwilliams (approved):

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -284,6 +293,28 @@ func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotK
 	return &Snapshot{key: key, manifest: manifest}, nil
 }
 
+func (l *FileCacheLoader) checkAllArtifactsExist(ctx context.Context, manifest *fcpb.SnapshotManifest) error {
```
tylerwilliams, line 296: can't wait till this can just be a call to GetActionResult

bduffany: agreed - maybe another good reason to migrate filecache -> disk_cache and remove filecache? since we'd need a local AC for that to work

PR #4658, opened 2023-08-28
Add client label header

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
+++ b/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
@@ -21,6 +21,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
+	"github.com/buildbuddy-io/buildbuddy/server/util/usageutil"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/text/language"
 	"golang.org/x/text/message"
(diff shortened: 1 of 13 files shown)


enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -294,6 +295,7 @@ func (q *PriorityTaskScheduler) EnqueueTaskReservation(ctx context.Context, req
 }
 
 func (q *PriorityTaskScheduler) propagateExecutionTaskValuesToContext(ctx context.Context, execTask *repb.ExecutionTask) context.Context {
+	ctx = usageutil.InternalUsageContext(ctx)
```
tylerwilliams, line 298: I think a comment would be useful here (and below as well, about what this does and why you're calling it?)

server/rpc/interceptors/interceptors.go
```diff
@@ -54,6 +57,12 @@ func ClientOrigin() string {
 	return *origin
 }
 
+// PropagateLabelsContextKey is a context key that, if present in the context,
+// causes the configured origin and client headers to propagate to the outgoing
+// context. These headers will also be propagated as-is across chained RPCs to
+// other BuildBuddy servers if applicable.
+type PropagateLabelsContextKey struct{}
```
tylerwilliams, line 64: Do we need this context key? Or would it be sufficient to run:

```
		if *origin != "" {
			ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-origin", *origin)
		}
		if *client != "" {
			ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-client", *client)
		}
```

when InternalUsageContext() is called?

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
bduffany: Done - also renamed `InternalUsageContext` since the "Internal" part might be confusing (e.g. BYOX would actually be external-executor). wdyt?

server/rpc/interceptors/interceptors.go
bduffany: Good call :) done

tylerwilliams (approved):

PR #4659, opened 2023-08-28
Rework group picker behavior on subdomains.

vadimberezniker:
The previous UX was not great as it made switching groups difficult.

Now the group picker works and redirects to the appropriate subdomain, if necessary.

Create org also works on subdomain and redirects to the new URL after creation.

**Related issues**: N/A

### DIFF
--- a/enterprise/app/org/create_org.tsx
+++ b/enterprise/app/org/create_org.tsx
@@ -19,8 +19,8 @@ export default class CreateOrgComponent extends OrgForm<grp.CreateGroupRequest> 
     return new grp.CreateGroupRequest({ ...DEFAULT_VALUES, ...values });
   }
   async submitRequest() {
-    const rsp = await rpcService.service.createGroup(this.state.request);
-    await authService.setSelectedGroupId(rsp.id, rsp.url);
+    const response = await rpcService.service.createGroup(this.state.request);
+    await authService.setSelectedGroupId(response.id, response.url);
     router.navigateTo(Path.settingsPath);
   }
 
--- a/enterprise/app/org/create_org.tsx
+++ b/enterprise/app/org/create_org.tsx
@@ -19,8 +19,8 @@ export default class CreateOrgComponent extends OrgForm<grp.CreateGroupRequest> 
     return new grp.CreateGroupRequest({ ...DEFAULT_VALUES, ...values });
   }
   async submitRequest() {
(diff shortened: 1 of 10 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -231,17 +231,12 @@ func (s *BuildBuddyServer) CancelExecutions(ctx context.Context, req *inpb.Cance
...
-			}
-		}
-		filtered = append(filtered, gr)
+func getGroupUrl(g *tables.Group) string {
+	gURL := build_buddy_url.String()
+	if g.URLIdentifier != nil && *g.URLIdentifier != "" {
+		gURL = replaceURLSubdomainForGroup(build_buddy_url.String(), g)
 	}
-	return filtered
+	return gURL
```
tylerwilliams, line 239: https://media.tenor.com/kcHmyiZ6q8AAAAAd/gurl-please-eye-roll.gif

tylerwilliams (approved):
backend looks OK but I'll defer to @bduffany on the frontend!

server/buildbuddy_server/buildbuddy_server.go
bduffany: :joy:

enterprise/app/org/create_org.tsx
```diff
@@ -19,8 +19,8 @@ export default class CreateOrgComponent extends OrgForm<grp.CreateGroupRequest>
     return new grp.CreateGroupRequest({ ...DEFAULT_VALUES, ...values });
   }
   async submitRequest() {
-    const { id } = await rpcService.service.createGroup(this.state.request);
-    await authService.setSelectedGroupId(id);
+    const rsp = await rpcService.service.createGroup(this.state.request);
+    await authService.setSelectedGroupId(rsp.id, rsp.url);
     router.navigateTo(Path.settingsPath);
```
bduffany, line 24: uber nit: this may be a losing battle since we are all used to Go, but for the JS code it's a bit unconventional to abbreviate stuff :P

```suggestion
    const response = await rpcService.service.createGroup(this.state.request);
    await authService.setSelectedGroupId(response.id, response.url);
    router.navigateTo(Path.settingsPath);
```

bduffany (approved):

PR #4663, opened 2023-08-28
Use hermetic_cc_toolchain to statically link against musl

tempoz:
Adds `hermetic_cc_toolchain` as the new toolchain for C and C++ compilation. Patches `soci-snapshotter` to depend on `zlib`. Adds build options to `.bazelrc` to perform static linking and an [ASLR-able binary](https://stackoverflow.com/questions/61553723/whats-the-difference-between-statically-linked-and-not-a-dynamic-executable).

Edits: Does not work on Mac. Could not build PIE-able code without crashing the race detector during analysis, so no ASLR.

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -39,13 +39,7 @@ build:remote-shared --jobs=100
 build:remote-shared --verbose_failures
 build:remote-shared --platforms=@buildbuddy_toolchain//:platform_bare_linux
 build:remote-shared --host_platform=@buildbuddy_toolchain//:platform_bare_linux
-# build:remote-shared --platforms=@hermetic_cc_toolchain//toolchain/platform:linux_amd64
-# build:remote-shared --host_platform=@hermetic_cc_toolchain//toolchain/platform:linux_amd64
-# build:remote-shared --crosstool_top=@buildbuddy_toolchain//:ubuntu_cc_toolchain_suite
-# build:remote-shared --extra_toolchains=@buildbuddy_toolchain//:ubuntu_cc_toolchain
 build:remote-shared --extra_toolchains=//:sh_toolchain
-# build:remote-shared --host_javabase=@bazel_tools//tools/jdk:absolute_javabase
-# build:remote-shared --define=ABSOLUTE_JAVABASE=/path/to/my/jdk
 
 # Build with --config=remote to use BuildBuddy RBE.
 build:remote --config=remote-shared
--- a/.bazelrc
+++ b/.bazelrc
@@ -41,8 +41,8 @@ build:remote-shared --platforms=@buildbuddy_toolchain//:platform_bare_linux
 build:remote-shared --host_platform=@buildbuddy_toolchain//:platform_bare_linux
 # build:remote-shared --platforms=@hermetic_cc_toolchain//toolchain/platform:linux_amd64
 # build:remote-shared --host_platform=@hermetic_cc_toolchain//toolchain/platform:linux_amd64
-build:remote-shared --crosstool_top=@buildbuddy_toolchain//:ubuntu_cc_toolchain_suite
-build:remote-shared --extra_toolchains=@buildbuddy_toolchain//:ubuntu_cc_toolchain
+# build:remote-shared --crosstool_top=@buildbuddy_toolchain//:ubuntu_cc_toolchain_suite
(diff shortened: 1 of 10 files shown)


tempoz:
* master
  * **PR #4663** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4663" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4663?utm_source=stack-comment).

buildpatches/com_github_awslabs_soci_snapshotter.patch
```diff
@@ -0,0 +1,34 @@
+diff --git a/MODULE.bazel b/MODULE.bazel
+new file mode 100644
+index 0000000..7ab2087
+--- /dev/null
++++ b/MODULE.bazel
+@@ -0,0 +1,6 @@
++module(
++    name = "soci_snapshotter",
++    version = "0.0.5",
```
iain-macdonald, line 9: I just bumped this to 0.0.7, do you mind updating it here as well?

tylerwilliams:
really excited to try this

buildpatches/com_github_awslabs_soci_snapshotter.patch
tempoz: done!

tempoz:
Overview of final changes:
- We add the `hermetic_cc_toolchain` and then register it for linux hosts (and only linux hosts, since the mac toolchain is missing headers we need).
- We set `--enable_platform_specific_config` so that we can specify a linux-only default configuration in the `.bazelrc`.
- We enable cpp toolchain resolution and disable detection of a default cpp toolchain on linux hosts to ensure we always use the correct toolchain and don't ever fall back to the default host one.
- We specify that `go` targets should be built statically for linux hosts
- We specify that all C targets be built with `-fPIC` for linux hosts, which was important for making static linking work; otherwise, we see `ld` failures during the linking step when attempting to relocate symbols.
- We set the remote platforms to use a non-clang version of the linux platforms in `buildbuddy_toolchain`.
- We patch the gazelle-generated bazel files for the `soci_snapshotter` to depend explicitly on `zlib` using bazel modules. The hermetic build environment does not contain `zlib`, whereas the default toolchain was just pulling from whatever `zlib` was installed on the host machine.
- We disable `pure` for the `goinit` target, as it caused the resulting binary to require an ELF interpreter, which was obviously not present on the initial ramdisk and caused the kernel to fail on startup with an `ENOENT` error code.
- We add the hermetic cpp toolchain to our tests that run bazel and need to locate a cpp toolchain, because the environment(?) is preventing them from locating the default system one. I would like to, in the future, see if I can figure out why this is happening and how to prevent it.
- We make some minor changes to the firecracker tests to make them behave better when sharding (which was helpful when testing) and to increase the memory allocated to prevent the race detector experiencing OOM errors.

I saw no appreciable performance difference when building remotely. Tensorflow built in 4096 seconds with the statically linked binary and 4067 seconds on master on my desktop with the same settings.

tempoz:
This is ready for review, but ideally we merge buildbuddy-io/buildbuddy-toolchain#12 first, so we can point our buildbuddy-toolchain `http_archive` back at the main branch.

Also, merging is "blocked" (or at least should be) by #4910, since the current CLI cannot build this PR, which would disrupt the workflows of anyone who is using the CLI on the team.

tylerwilliams (approved):

.bazelrc
```diff
@@ -147,6 +145,15 @@ build:auto-release --config=remote
...
 build:auto-release -c opt
 
+build:linux --incompatible_enable_cc_toolchain_resolution
+build:linux --action_env BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+build:linux --@io_bazel_rules_go//go/config:static=true
+# build:linux --@io_bazel_rules_go//go/config:linkmode=pie
+build:linux --copt=-fPIC
+# build:linux --linkopt=-static
+# build:linux --linkopt=-static-pie
+# build:linux --action_env=GOTAGS=-tags=osusergo,netgo
```
tylerwilliams, line 155: can we remove the commented lines here if they're not necessary?

tempoz: yup, fine with me, I just left them so if I go back and do pie stuff I remember the flags needed for it.

PR #4671, opened 2023-08-29
Firecracker: ignore old-style persisted snapshots when snapshot sharing is enabled

bduffany:
I'm not really sure if old snapshots (non-shareable) and new snapshots (shareable) are compatible, but don't want to have to spend any time testing it - so when snapshot sharing is enabled, just ignore old persisted snapshots.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -431,6 +431,10 @@ func (p *Provider) New(ctx context.Context, props *platform.Properties, task *re
 			InitDockerd:       props.InitDockerd,
 			EnableDockerdTcp:  props.EnableDockerdTCP,
 		}
+	} else if *snaputil.EnableLocalSnapshotSharing {
+		// When local snapshot sharing is enabled, reject old-style persisted
+		// snapshots, since these aren't shareable (i.e. COW-formatted).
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -329,6 +329,10 @@ func (p *Provider) New(ctx context.Context, props *platform.Properties, task *re
 			InitDockerd:       props.InitDockerd,
 			EnableDockerdTcp:  props.EnableDockerdTCP,
 		}
+	} else if *snaploader.EnableLocalSnapshotSharing {
+		// When local snapshot sharing is enabled, reject old-style persisted
+		// snapshots, since these aren't shareable (i.e. COW-formatted).
+		return nil, status.UnavailableError("ignoring persisted snapshot; this functionality has been replaced by local snapshot sharing")
```
maggie-lou, line 335: Should this return nil? Or should it just use the default vmConfig like when savedState == nil

bduffany: These VMs are eagerly added to the pool on executor startup and if we return an error then we just skip adding it. So I think returning an error makes sense here

maggie-lou (approved):

PR #4676, opened 2023-08-30
Fix workflows using stale GitHub app tokens

bduffany:
This was probably causing the "invalid username or password" errors.

**Related issues**: N/A


vadimberezniker (approved):

tylerwilliams (approved):

bduffany:
Switched from "remove then add" to "try adding but if it already exists then update the URL."

The problem with removing is that it drops some config options on the existing remote. This caused tests to fail. (The dropped options were `partialclonefilter = blob:none` and `promisor = true`)

PR #4680, opened 2023-08-30
use output from ResourceNameFromProto to fill DigestFunction in FileRecords

jdhollen:
This fixes an issue where GetCacheMetadata was failing because the provided ResourceName didn't have a digest hash function set.

In the future, GetCacheMetadata should be updated to explicitly provide a HashFunction, but this code still shouldn't ignore the side effects of ResourceNameFromProto.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/distributed/distributed.go
+++ b/enterprise/server/backends/distributed/distributed.go
@@ -611,7 +611,7 @@ func (c *Cache) Metadata(ctx context.Context, r *rspb.ResourceName) (*interfaces
 			c.log.Debugf("Metadata(%q) not found on peer %s", cacheproxy.ResourceIsolationString(r), peer)
 			continue
 		}
-		c.log.Warningf("Metadata(%q) lookup failed on peer %s: (err: %+v)", cacheproxy.ResourceIsolationString(r), peer, err)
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
Nice catch, LGTM (and maybe see what @vadimberezniker thinks before submitting?)

enterprise/server/backends/distributed/distributed.go
```diff
@@ -611,6 +611,7 @@ func (c *Cache) Metadata(ctx context.Context, r *rspb.ResourceName) (*interfaces
 			c.log.Debugf("Metadata(%q) not found on peer %s", cacheproxy.ResourceIsolationString(r), peer)
 			continue
 		}
+		c.log.Debugf("Metadata(%q) lookup failed on peer %s: (err: %+v)", cacheproxy.ResourceIsolationString(r), peer, err)
```
vadimberezniker, line 614: nit: for errors we typically use %v as there's generally no need to know the structure of the error message

vadimberezniker (approved):

enterprise/server/backends/distributed/distributed.go
jdhollen: ok! done

PR #4682, opened 2023-08-30
Avoid recomputing digests for unchanged snapshot chunks

bduffany:
Otherwise if the disk is say 20GB, we are potentially burning through 20GB worth of digest CPU computations on every VM run even if only a few MB worth of files changed.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -126,12 +126,29 @@ type CacheSnapshotOptions struct {
 	WorkspaceFSPath string
 
 	// Labeled map of chunked artifacts backed by blockio.COWStore storage.
-	ChunkedFiles map[string]*blockio.COWStore
+	ChunkedFiles map[string]*ChunkedFile
 }
 
 type UnpackedSnapshot struct {
 	// ChunkedFiles holds any chunked files that were part of the snapshot.
-	ChunkedFiles map[string]*blockio.COWStore
+	ChunkedFiles map[string]*ChunkedFile
+}
+
+type ChunkedFile struct {
+	*blockio.COWStore
+
+	// digests caches any known digests for *non-dirty* chunks, keyed by offset.
+	digests map[int64]*repb.Digest
+}
+
+func (cf *ChunkedFile) chunkDigest(chunk *blockio.Chunk) (*repb.Digest, error) {
+	// If we already know the original chunk digest and the chunk hasn't
+	// changed, return the original digest.
+	if d := cf.digests[chunk.Offset]; d != nil && !cf.COWStore.Dirty(chunk.Offset) {
+		return d, nil
+	}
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -126,12 +126,32 @@ type CacheSnapshotOptions struct {
...
-	ChunkedFiles map[string]*blockio.COWStore
+	ChunkedFiles map[string]*ChunkedFile
+}
+
+type ChunkedFile struct {
+	*blockio.COWStore
+
+	// digests caches any known digests for *non-dirty* chunks, keyed by offset.
+	digests map[int64]string
+}
```
bduffany, line 142: I decided to put this in a snaploader-specific struct instead of adding it to COWStore because right now COWStore is pure I/O and doesn't care about digests etc. which seems more specific to caching. But if that feels somewhat arbitrary, I'm open to changing this. I also wasn't sure whether `ConvertFileToCOW` should eagerly compute the digest. We'll need the digest eventually but if the chunks change by the time we add them to cache then the computation would have been wasteful.

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -126,12 +126,32 @@ type CacheSnapshotOptions struct {
...
 	// ChunkedFiles holds any chunked files that were part of the snapshot.
-	ChunkedFiles map[string]*blockio.COWStore
+	ChunkedFiles map[string]*ChunkedFile
+}
+
+type ChunkedFile struct {
+	*blockio.COWStore
+
+	// digests caches any known digests for *non-dirty* chunks, keyed by offset.
+	digests map[int64]string
```
maggie-lou, line 141: Why not just cache the full digest including size? So we don't always have to recompute it

enterprise/server/remote_execution/snaploader/snaploader.go
maggie-lou: I don't really see the benefit of ConvertFileToCOW eagerly computing the digests. If anything it seems better that we push it off until after the workflow is finished if computing the digests ends up being slow

maggie-lou (approved):

maggie-lou:
* master
  * **PR #4682** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4682" target="_blank">Graphite</a>  👈
    * **PR #4687** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4687" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4682?utm_source=stack-comment).

maggie-lou:
Sorry - I accidentally force pushed a stale version of the branch using graphite. Should be fixed now

PR #4683, opened 2023-08-30
record the digest size to prometheus metrics in pebble cache Writer

luluz66:
**Related issues**: N/A
When CDC is enabled, we are recording the chunk size instead of the size of the
artifacts in DiskCacheAddedFileSizeBytes. This PR moves this metric up in the
call stack in the Writer to make things a bit simpler. Otherwise, we should only
record the metrics when we are writing entries with single chunks or file-level
entries.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -2103,6 +2103,10 @@ func (p *PebbleCache) Writer(ctx context.Context, r *rspb.ResourceName) (interfa
 	if err != nil {
 		return nil, err
 	}
+	sizeBytes := r.GetDigest().GetSizeBytes()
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2103,6 +2103,10 @@ func (p *PebbleCache) Writer(ctx context.Context, r *rspb.ResourceName) (interfa
 	if err != nil {
 		return nil, err
 	}
+	sizeBytes := r.GetDigest().GetSizeBytes()
```
vadimberezniker, line 2106: For AC entries the digest size does not equal the content size. Could we skip updating the metric for chunks and only update it for non-chunk entries?

vadimberezniker (changes requested):

PR #4685, opened 2023-08-30
Set encryption bit from selected group when authenticating by sub

jdhollen:
**Related issues**: N/A

### DIFF
--- a/server/util/claims/claims.go
+++ b/server/util/claims/claims.go
@@ -174,18 +174,23 @@ func ClaimsFromSubID(ctx context.Context, env environment.Env, subID string) (*C
 func userClaims(u *tables.User, effectiveGroup string) *Claims {
 	allowedGroups := make([]string, 0, len(u.Groups))
 	groupMemberships := make([]*interfaces.GroupMembership, 0, len(u.Groups))
+	cacheEncryptionEnabled := false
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):

server/util/claims/claims.go
```diff
@@ -174,18 +174,23 @@ func ClaimsFromSubID(ctx context.Context, env environment.Env, subID string) (*C
...
 	groupMemberships := make([]*interfaces.GroupMembership, 0, len(u.Groups))
+	cacheEncryptionEnabled := false
 	for _, g := range u.Groups {
 		allowedGroups = append(allowedGroups, g.Group.GroupID)
 		groupMemberships = append(groupMemberships, &interfaces.GroupMembership{
 			GroupID: g.Group.GroupID,
 			Role:    role.Role(g.Role),
 		})
+		if g.Group.GroupID == effectiveGroup {
+			cacheEncryptionEnabled = g.Group.CacheEncryptionEnabled
```
vadimberezniker, line 185: You can save this for a separate PR, but we should also fix the value during impersonation. ClaimsFromSubID creates a synthetic Group entry that does not have any of the group attributes filled in apart from the GroupID

vadimberezniker (approved):

PR #4688, opened 2023-08-31
Make SAML login work on subdomains.

vadimberezniker:
Uses a modified version of samlsp.CookieRequestTracker that sets cookies with a domain value that allows the cookies to be visible both on app.buildbuddy.io and foo.buildbuddy.io

Wraps the samlsp.CookieSessionProvider with similiar functionality. New cookies will have domain set such that cookies will be visible on both app.buildbuddy.io and foo.buildbuddy.io. The wraper will also clear out old cookies with the old domain value.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/saml/saml.go
+++ b/enterprise/server/saml/saml.go
@@ -113,14 +113,14 @@ func (t CookieRequestTracker) TrackRequest(w http.ResponseWriter, r *http.Reques
 // StopTrackingRequest stops tracking the SAML request given by index, which is a string
 // previously returned from TrackRequest
 func (t CookieRequestTracker) StopTrackingRequest(w http.ResponseWriter, r *http.Request, index string) error {
-	cookie, err := r.Cookie(t.NamePrefix + index)
+	c, err := r.Cookie(t.NamePrefix + index)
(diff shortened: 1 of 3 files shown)


enterprise/server/saml/saml.go
```diff
@@ -51,6 +56,146 @@ var (
...
+		SameSite: t.SameSite,
+		Secure:   t.ServiceProvider.AcsURL.Scheme == "https",
+		Path:     t.ServiceProvider.AcsURL.Path,
+	})
+
+	return trackedRequest.Index, nil
+}
+
+// StopTrackingRequest stops tracking the SAML request given by index, which is a string
+// previously returned from TrackRequest
```
tylerwilliams, line 114: nit: <80

tylerwilliams (approved):
Is it possible to upstream any of this so it's less for us to maintain and less brittle if we update the SAML lib?

tylerwilliams:
@siggisim just making sure you get a chance to see this if you want to

siggisim:
> @siggisim just making sure you get a chance to see this if you want to

this is cool with me

PR #4689, opened 2023-08-31
Use BuildEvents directly instead of wrapping stuff with InvocationEvents.

jdhollen:
Gonna test this a bit longer, but this fixes some weirdness where we were doing comparisons of things that were literally _always_ new instances.  Do you remember if there was a particularly good reason for this?

**Related issues**: N/A


bduffany:
> Do you remember if there was a particularly good reason for this?

It was just so I didn't have to refactor that card :sweat_smile:

bduffany (approved):

PR #4694, opened 2023-08-31
Remove podman images that hit the input/output error bug

iain-macdonald:
This PR expands the image cleanup logic used to clean up podman images affected by https://github.com/containers/storage/issues/1136 to also apply to images affected by https://github.com/buildbuddy-io/buildbuddy-internal/issues/2570. This will not fix the latter bug, but it should ensure that subsequent actions have a better chance of succeeding because they'll re-pull the image and, I hope, correct the race that's causing the bug.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2570q

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -972,12 +972,12 @@ func (c *podmanCommandContainer) killContainerIfRunning(ctx context.Context) err
 func (c *podmanCommandContainer) maybeCleanupCorruptedImages(ctx context.Context, result *interfaces.CommandResult) error {
 	stderr := string(result.Stderr)
 	if result.ExitCode == podmanInternalExitCode && storageErrorRegex.MatchString(stderr) {
-		log.Debug("podman detected storage corruption, removing image")
+		log.CtxInfo(ctx, "podman detected storage corruption, removing image")
 		result.Error = status.UnavailableError("a storage corruption occurred")
 		result.ExitCode = commandutil.NoExitCode
 		return removeImage(ctx, c.image)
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -957,19 +962,27 @@ func (c *podmanCommandContainer) killContainerIfRunning(ctx context.Context) err
...
-	if result.ExitCode != podmanInternalExitCode {
-		return nil
+	stderr := string(result.Stderr)
+	if result.ExitCode == podmanInternalExitCode && storageErrorRegex.MatchString(stderr) {
+		log.Debug("podman detected storage corruption, removing image")
+		result.Error = status.UnavailableError("a storage corruption occurred")
+		result.ExitCode = commandutil.NoExitCode
+		return removeImage(ctx, c.image)
+	} else if result.ExitCode == podmanCannotInvokeExitCode && inputOutputErrorRegex.MatchString(stderr) {
+		log.Debug("podman detected input/output error, removing image")
```
bduffany, line 980: nit: I think these are worth logging at info level so we can tell how often we're hitting these in prod

also probably worth using `Ctx` funcs so we can associate the logs with invocation IDs

```suggestion
		log.CtxInfo(ctx, "podman detected storage corruption, removing image")
		result.Error = status.UnavailableError("a storage corruption occurred")
		result.ExitCode = commandutil.NoExitCode
		return removeImage(ctx, c.image)
	} else if result.ExitCode == podmanCannotInvokeExitCode && inputOutputErrorRegex.MatchString(stderr) {
		log.CtxInfo(ctx, "podman detected input/output error, removing image")
```

bduffany (approved):

PR #4702, opened 2023-09-05
Use transient store for podman actions

bduffany:
Without this flag, we see up to a few seconds of extra latency per action when running lots of actions concurrently. It can take up to 8 seconds to run `sleep 1` :open_mouth: to fix this, we need to set _both_ `--transient-store` and `--net=none`. This PR sets `--transient-store` on all actions, and separately we'll try to migrate users onto setting `dockerNetwork=off` in their actions.

See https://github.com/containers/podman/issues/19824

**Related issues**: N/A


iain-macdonald (approved):

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -927,6 +927,9 @@ func (c *podmanCommandContainer) State(ctx context.Context) (*rnpb.ContainerStat
 func runPodman(ctx context.Context, subCommand string, stdio *container.Stdio, args ...string) *interfaces.CommandResult {
 	command := []string{
 		"podman",
+		// Use transient store to reduce contention.
+		// See https://github.com/containers/podman/issues/19824
+		"--transient-store",
```
tylerwilliams, line 932: do you think it's a good idea to put this behind a flag so we can test in dev for a little longer? (rather than just one day)

bduffany: yeah, good call - Done

PR #4705, opened 2023-09-05
Make splits happen via raft command

tylerwilliams:
Make splits happen via raft command.

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -824,12 +824,6 @@ func (sm *Replica) printRange(r pebble.Reader, iterOpts *pebble.IterOptions, tag
 	}
 }
 
-// REMEMBER: no raft commands may runwhile this one is running!
-// xxx Request contains the new ClusterID and RangeID
-// xxx Find a split point and compute new range descriptors
-// xxx Add a peer with the new ClusterID
-// xxx Update peer's range descriptor and snapshot it
-// xxx Update this peer's range descriptor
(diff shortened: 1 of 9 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -819,6 +824,74 @@ func (sm *Replica) printRange(r pebble.Reader, iterOpts *pebble.IterOptions, tag
 	}
 }
 
+// REMEMBER: no raft commands may runwhile this one is running!
+// xxx Request contains the new ClusterID and RangeID
```
vadimberezniker: Clean up the comment.

enterprise/server/raft/store/store.go
```diff
@@ -1062,6 +1058,97 @@ func (s *Store) TransferLeadership(ctx context.Context, req *rfpb.TransferLeader
...
+		default:
+			if rd := s.lookupRange(clusterID); rd != nil {
+				return nil
+			}
+			time.Sleep(10 * time.Millisecond)
+		}
+	}
+}
+
+func (s *Store) SnapshotCluster(ctx context.Context, clusterID uint64) error {
```
vadimberezniker, line 1075: I'd add a comment to clarify that this a local operation (i.e. it doesn't affect other nodes)

vadimberezniker (approved):

enterprise/server/raft/store/store.go
tylerwilliams: Done.

enterprise/server/raft/replica/replica.go
tylerwilliams: Done.

PR #4698, opened 2023-09-05
Refactor uffd to take a snaploader store instead of a blockio store

maggie-lou:
**Related issues**: N/A

PR #4706, opened 2023-09-05
[Remote Snapshot Sharing] Refactor firecracker copy-on-write files to dynamically initialize chunks

maggie-lou:
To prepare for remote snapshot sharing, refactor Snaploader's copy-on-write interface to lazy-fetch chunks. This will let us dynamically fetch a remote chunk, rather than having to pre-fetch all chunks when unpacking the COW files

New flow:
1. Firecracker uses snaploader. ConvertFileToLazyCOW() anytime it wants to create a COW file
2. This creates a blockio.COWStore with snaploader.LazyMmap chunks
3. Anytime we need to access a chunk - i.e. to read or write to it, LazyMmap first calls init(), which will load the chunk from the local or remote cache if needed
4. It then delegates the actual read/write operation to the nested blockio.Mmap struct which implements the logic of actually modifying the chunk's data

I tried to preserve the existing structure of blockio. All logic around copy-on-write or reading/writing to a chunk lives in blockio. All logic around loading chunks from the cache and memoizing digests lives in snaploader


maggie-lou:
* master
  * **PR #4682** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4682" target="_blank">Graphite</a> 
    * **PR #4706** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4706" target="_blank">Graphite</a>  👈
    * **PR #4698** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4698" target="_blank">Graphite</a> 
    * **PR #4687** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4687" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4706?utm_source=stack-comment).

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -128,29 +130,199 @@ type CacheSnapshotOptions struct {
...
+		//}
+		return err
+	}
+
+	err = cf.AddChunk(chunkStartOffset, chunk)
+	if err != nil {
+		return status.WrapError(err, "add chunk to store")
+	}
+
+	delete(cf.unloadedChunks, chunkStartOffset)
```
maggie-lou, line 189: @bduffany One change I made here - If we dynamically fetch a chunk, we'll delete the digest from the memoized map. That means that even if we don't write to the chunk, we'll still have to recompute the digest. 

I thought it was cleaner this way, because otherwise the memoized map (offset -> digest) would contain some valid data and some invalid digests if we had written to that offset. Seems error-prone if someone accidentally tried to use that memoized map somewhere else without realizing it hadn't been synchronized with writes. Also seems wasteful to recompute the digest every time we write to a chunk if we wanted to keep the digest map always up-to-date

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -355,7 +355,7 @@ func (p *Provider) New(ctx context.Context, props *platform.Properties, task *re
 type FirecrackerContainer struct {
 	id          string // a random GUID, unique per-run of firecracker
 	vmIdx       int    // the index of this vm on the host machine
-	loader      snaploader.Loader
+	loader      *snaploader.FileCacheLoader
```
bduffany, line 358: is this change needed?

enterprise/server/remote_execution/nbd/nbdserver/nbdserver.go
```diff
@@ -18,13 +18,13 @@ import (
...
+	*blockio.COWStore
 	Metadata *nbdpb.DeviceMetadata
 }
 
-func NewExt4Device(store blockio.Store, name string) (*Device, error) {
+func NewExt4Device(store *blockio.COWStore, name string) (*Device, error) {
 	return &Device{
-		Store: store,
+		COWStore: store,
 		Metadata: &nbdpb.DeviceMetadata{
```
bduffany, line 28: are these changes needed?

maggie-lou:
Rebased off my refactors and it's way simpler now 🎉  Opened a new PR for this here: https://github.com/buildbuddy-io/buildbuddy/pull/4794

PR #4711, opened 2023-09-06
docs/troubleshooting-rbe: add TCP connection tunning

sluongng:
This was one of the issue came up recently when a customer of ours try
to connect to us through an EC2 instance backed by AWS NAT Gateway.

It's a known issue that NAT Gateway enforce connection drop after 350
seconds idling. (1)

(1): https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-troubleshooting.html#nat-gateway-troubleshooting-timeout

**Related issues**: N/A

### DIFF
--- a/docs/troubleshooting-rbe.md
+++ b/docs/troubleshooting-rbe.md
@@ -59,14 +59,14 @@ This typically happens when there is a proxy or gateway (e.g. AWS NAT Gateway) i
 When this happens, try the following Linux network settings:
 
 ```bash
-# Lowered from default 7200
+# Lowered from default value: 7200
 sudo sysctl -w net.ipv4.tcp_keepalive_time=180
-# Using default value
-sudo sysctl -w net.ipv4.tcp_keepalive_intvl=75
-# Using default value
-sudo sysctl -w net.ipv4.tcp_keepalive_probes=9
+# Lowered from default value: 75
+sudo sysctl -w net.ipv4.tcp_keepalive_intvl=60
+# Lowered from default value: 9
+sudo sysctl -w net.ipv4.tcp_keepalive_probes=2
 ```
 
-This will cause the Linux kernel to send keepalive probes more eagerly, before the proxy/gateway in the middle detects and drops the idle connection.
+This will cause the Linux kernel to send keepalive probes earlier and more frequently, before the proxy/gateway in the middle detects and drops the idle connection.
 
 The optimal values may depend on specific network conditions, but try these values as a starting point. Please [contact us](/contact/) if you have any questions / concerns.
--- a/docs/troubleshooting-rbe.md
+++ b/docs/troubleshooting-rbe.md
@@ -40,12 +40,12 @@ If you'd like to add Mac executors to your BuildBuddy Cloud account, please [con
 
 ## WARNING: Remote Cache: UNAVAILABLE: io exception
 
-This error occurs when Bazel JVM have a hard time maintaining the TCP connection with BuildBuddy remote cache.
+This error may occur when Bazel fails to properly maintain a long-running TCP connection to BuildBuddy.
 
-To validate this error, user could use Bazel's flag `--remote_grpc_log=grpc.log` to capture the grpc traffic
-from Bazel JVM and BuildBuddy. The log file will be in protobuf format. to convert it to JSON format, user
-could use our [BB CLI](/docs/cli) and run `bb print --grpc_log=<path-to-file>/grpc.log`. In side the log, we
-expect to see network errors such as
+To check whether this is the case, try running Bazel with `--remote_grpc_log=grpc.log` to capture the gRPC traffic
+between Bazel and BuildBuddy. The log file will be in protobuf format. To convert it to JSON format, download the [BuildBuddy CLI](/docs/cli) and run `bb print --grpc_log=<path-to-file>/grpc.log`.
+
+In the log, you may see network errors such as the following:
 
 ```json
   "status":  {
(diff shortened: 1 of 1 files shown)


docs/troubleshooting-rbe.md
```diff
@@ -37,3 +37,35 @@ If you'd like to add Mac executors to your BuildBuddy Cloud account, please [con
...
+```bash
+sudo sysctl -w net.ipv4.tcp_keepalive_time=180
+sudo sysctl -w net.ipv4.tcp_keepalive_intvl=60
+sudo sysctl -w net.ipv4.tcp_keepalive_probes=25
+```
+
+so the keepalive routines in Linux kernel would send keepalive probe more frequently.
+
+The exact values could be different, subjected to user's specific network condition.
+Please [contact us](/contact/) if you have any questions / concerns.
```
bduffany, line 71: (The GitHub suggestion UI wasn't happy because this section contains triple-backticks, but here are my suggested edits below - feel free to copy and paste the raw markdown if this all looks good)

---

This error may occur when Bazel fails to properly maintain a long-running TCP connection to BuildBuddy.

To check whether this is the case, try running Bazel with `--remote_grpc_log=grpc.log` to capture the gRPC traffic
between Bazel and BuildBuddy. The log file will be in protobuf format. To convert it to JSON format, download the [BuildBuddy CLI](/docs/cli) and run `bb print --grpc_log=<path-to-file>/grpc.log`.

In the log, you may see network errors such as the following:

```json
  "status":  {
    "code":  14,
    "message":  "io.netty.channel.unix.Errors$NativeIoException: readAddress(..) failed: Connection reset by peer"
  },
```

This typically happens when there is a proxy or gateway (e.g. AWS NAT Gateway) in between Bazel and BuildBuddy that is terminating idle connections too quickly.

When this happens, try the following Linux network settings:

```bash
sudo sysctl -w net.ipv4.tcp_keepalive_time=180
sudo sysctl -w net.ipv4.tcp_keepalive_intvl=60
sudo sysctl -w net.ipv4.tcp_keepalive_probes=25
```

This will cause the Linux kernel to send keepalive probes more frequently.

The optimal values may depend on specific network conditions, but try these values as a starting point. Please [contact us](/contact/) if you have any questions / concerns.

docs/troubleshooting-rbe.md
```diff
@@ -37,3 +37,35 @@ If you'd like to add Mac executors to your BuildBuddy Cloud account, please [con
...
+
+This is usually encountered by customers who run operate some proxy or gateway (i.e. AWS NAT Gateway) in between
+Bazel JVM and BuildBuddy server.
+
+When this happen, user should consider tunning their Linux network settings like this:
+
+```bash
+sudo sysctl -w net.ipv4.tcp_keepalive_time=180
+sudo sysctl -w net.ipv4.tcp_keepalive_intvl=60
+sudo sysctl -w net.ipv4.tcp_keepalive_probes=25
```
vadimberezniker, line 65: This value seems unnecessarily high. This is effectively saying don't consider the connection dead until 180+60*25=28 minutes after sending the first keepalive packet. I would aim for something closer to 5-10 minutes.

sluongng: Here is the kernel default https://github.com/torvalds/linux/blob/7ba2090ca64ea1aa435744884124387db1fac70f/include/net/tcp.h#L155-L157

which is
```
7200
75
9
```

which total to roughly more than 2 hours. Another point to note is that on the Client machine, anything could be running, not just Bazel. So we probably want to stick with the default as much as possible.

I think the goal here is not to extend the total keepalive time but to shorten the ping interval to bypass the proxy/gateway limit in the middle. So how about we use the default value for `tcp_keepalive_intvl` and `tcp_keepalive_probes`, but still mention them in the doc for users' awareness? That would make it 

```bash
sudo sysctl -w net.ipv4.tcp_keepalive_time=180
# Using default value
sudo sysctl -w net.ipv4.tcp_keepalive_intvl=75
# Using default value
sudo sysctl -w net.ipv4.tcp_keepalive_probes=9
```

which equate to 180 + 75 * 9 > 14 mins.

docs/troubleshooting-rbe.md
sluongng: Thanks for this edit.

I pushed a newer version with a few minor adjustments based on Vadim's comment below.

docs/troubleshooting-rbe.md
vadimberezniker: The kernel defaults were set 20+ years ago when connectivity was very different. It would be very unusual on modern day networks and modern day PCs to not get a response to a keepalive ping in 5 minutes, let alone 15.

sluongng: @vadimberezniker Alright. Let's do `180 + 60 * 2 = 300s = 5m` then.

docs/troubleshooting-rbe.md
```diff
@@ -37,3 +37,36 @@ If you'd like to add Mac executors to your BuildBuddy Cloud account, please [con
...
+
+When this happens, try the following Linux network settings:
+
+```bash
+# Lowered from default value: 7200
+sudo sysctl -w net.ipv4.tcp_keepalive_time=180
+# Lowered from default value: 75
+sudo sysctl -w net.ipv4.tcp_keepalive_intvl=60
+# Lowered from default value: 9
+sudo sysctl -w net.ipv4.tcp_keepalive_probes=2
```
vadimberezniker, line 67: I would raise this to 4 or 5. My math was a little handwavy earlier :-P 
Once keepalive is triggered, the initial time (180) is not used anymore. When keepalive first comes into effect, the time is 180+60\*2, but on subsequent checks the timeout becomes only 60*2 before connection is considered dead.

vadimberezniker (approved):

bduffany (approved):

docs/troubleshooting-rbe.md
sluongng: bump the number to 5. 👍

PR #4712, opened 2023-09-06
Run the executor under tini in the release image too.

bduffany:
Analogue of https://github.com/buildbuddy-io/buildbuddy-internal/pull/2536/files but for the release image.

Also updates the `run_executor_image` script since `container_image` targets cannot be run directly, like `go_image` targets can.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/executor/BUILD
+++ b/enterprise/server/cmd/executor/BUILD
@@ -122,9 +122,20 @@ container_image(
 #
 # This target can be run locally with enterprise/tools/run_executor_image.sh
 go_image(
-    name = "executor_image",
+    name = "executor_go_image",
     base = ":base_image",
     binary = ":executor",
+    tags = ["manual"],
+)
+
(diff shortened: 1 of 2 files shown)


sluongng:
We might be able to clean up the redundant image in https://github.com/buildbuddy-io/buildbuddy-internal/blob/1e4b95b1958cf3fa65e2a078f1c97f28aa2de754/enterprise/BUILD#L79-L89 then?

bduffany:
> We might be able to clean up the redundant image in https://github.com/buildbuddy-io/buildbuddy-internal/blob/1e4b95b1958cf3fa65e2a078f1c97f28aa2de754/enterprise/BUILD#L79-L89 then?

Maybe :) It's referencing a different file path for the executor binary, so I'm not quite sure how we'd consolidate that.

vadimberezniker (approved):

PR #4713, opened 2023-09-06
cli/picker: skip using picker when some flags are provided

sluongng:
In case where Bazel target patterns are provided via a flag instead of
args, don't use our picker.

**Related issues**: N/A

### DIFF
--- a/cli/picker/picker.go
+++ b/cli/picker/picker.go
@@ -21,6 +21,19 @@ func HandlePicker(args []string) []string {
 	// If the command is build, test, or query without a specified target - apply to all targets.
 	command := arg.GetCommand(args)
 
+	bzlArgs := arg.GetBazelArgs(args)
+	for _, bzlArg := range bzlArgs {
+		// Skip using the picker if the user has specified a query file.
+		if strings.Contains(command, "query") && strings.Contains(bzlArg, "--query_file") {
+			return args
+		}
+
+		// Skip using the picker if the user has specified a target pattern file.
+		if command == "build" && strings.Contains(bzlArg, "--target_pattern_file") {
+			return args
+		}
+	}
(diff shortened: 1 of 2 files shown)


cli/test/integration/cli/cli_test.go
```diff
@@ -295,6 +295,18 @@ func TestTerminalOutput(t *testing.T) {
 	require.Contains(t, term.Render(), "\x1b[32mINFO")
 }
 
+func TestNoPickerUsed(t *testing.T) {
```
bduffany, line 298: nit: it's a little hard to understand the name of this test out of context.

```suggestion
func TestTargetPatternFile(t *testing.T) {
```

bduffany (approved):

cli/picker/picker.go
```diff
@@ -21,6 +21,19 @@ func HandlePicker(args []string) []string {
...
+	for _, bzlArg := range bzlArgs {
+		// Skip using the picker if the user has specified a query file.
+		if strings.Contains(command, "query") && strings.Contains(bzlArg, "--query_file") {
+			return args
+		}
+
+		// Skip using the picker if the user has specified a target pattern file.
+		if command == "build" && strings.Contains(bzlArg, "--target_pattern_file") {
+			return args
+		}
```
bduffany, line 34: sorry I misread this originally - instead of Contains, should do `==` OR `startswith --target_pattern_file=`

sluongng: I think startswith makes sense. The flag could be set as `--target_pattern_file=targets.txt` OR `--target_pattern_file targets.txt --` (this happens before `--` so targets.txt won't be parsed as build target pattern).

Let me send a fix.

PR #4716, opened 2023-09-06
Enable golang staticcheck SA4006 -- A value assigned to a variable is never read before being overwritten. Forgotten error check or dead code?

iain-macdonald:
**Related issues**: N/A


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1414,7 +1414,8 @@ func (p *PebbleCache) blobDir() string {
 }
 
 func (p *PebbleCache) lookupFileMetadataAndVersion(ctx context.Context, iter pebble.Iterator, key filestore.PebbleKey) (*rfpb.FileMetadata, filestore.PebbleKeyVersion, error) {
-	ctx, spn := tracing.StartSpan(ctx)
+	// Skip lint to ensure the correct ctx is used if it's needed.
+	ctx, spn := tracing.StartSpan(ctx) //nolint:SA4006
```
tylerwilliams, line 1418: I don't know how you feel about this, but one way we could fix this is by changing the arg name from ctx to rawCtx or something else, then code would be:

ctx, spn := tracing.StartSpan(rawCtx)

This would remove the nead to have `//nolint:SA4006` everywhere.

enterprise/server/raft/store/store.go
```diff
@@ -1864,7 +1864,7 @@ func addLocalRangeEdits(oldLeft, newLeft *rfpb.RangeDescriptor, b *rbuilder.Batc
 	if err != nil {
 		return err
 	}
-	b = b.Add(cas)
+	*b = *b.Add(cas)
```
tylerwilliams, line 1867: I think this can just be `b.Add(cas)`?

The returned object is only returned for chaining -- but it's mutating the underlying object either way.

tylerwilliams: Edit: same below.

tylerwilliams (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: SGTM

enterprise/server/raft/store/store.go
iain-macdonald: Ah yeah, fixed

tylerwilliams (approved):
LGTM

Cool with the nolint or rawctx approach, whatever you like better!

PR #4721, opened 2023-09-07
add backend support for smaller time increments in drilldowns.

jdhollen:
I'm not seeing an appreciable difference in performance for this.  I'll fix the axis / timestamp rendering in the frontend separately (this is flag-guarded)

Before: Screenshot 2023-09-06 at 11 11 25 PM
After: 
Screenshot 2023-09-06 at 11 12 05 PM

**Related issues**: N/A

### DIFF
--- a/enterprise/server/invocation_stat_service/invocation_stat_service.go
+++ b/enterprise/server/invocation_stat_service/invocation_stat_service.go
@@ -39,6 +39,7 @@ var (
 	useTimezoneInHeatmapQueries    = flag.Bool("app.use_timezone_in_heatmap_queries", true, "If enabled, use timezone instead of 'timezone offset' to compute day boundaries in heatmap queries.")
 	invocationSummaryAvailableUsec = flag.Int64("app.invocation_summary_available_usec", 0, "The timstamp when the invocation summary is available in the DB")
 	tagsInDrilldowns               = flag.Bool("app.fetch_tags_drilldown_data", true, "If enabled, DrilldownType_TAG_DRILLDOWN_TYPE can be returned in GetStatDrilldownRequests")
+	finerDrilldownTimeIncrements   = flag.Bool("app.finer_drilldown_time_increments", false, "If enabled, split drilldowns into smaller time buckets when the user has a smaller date range selected.")
 )
 
 type InvocationStatService struct {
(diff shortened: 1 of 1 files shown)


jdhollen:
feh, i need to make this round to the "right" values based on the bucket size we picked (i.e., :05, :30, etc.).  this is why you don't code late.  i'll fix it in the morning.

jdhollen:
okay, all set!

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -556,6 +557,19 @@ func (i *InvocationStatService) getMetricBuckets(ctx context.Context, table stri
 	return metricBuckets, metricArrayStr, nil
 }
 
+func getTimeIncrement(duration time.Duration) time.Duration {
```
tylerwilliams, line 560: little comment here would be nice for the non-drilldown-initiated

enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -594,15 +608,27 @@ func getTimestampBuckets(q *stpb.TrendQuery, requestContext *ctxpb.RequestContex
...
-	midnightOnStartDate := time.Date(start.Year(), start.Month(), start.Day(), 0, 0, 0, 0, start.Location())
-	current := midnightOnStartDate.AddDate(0, 0, 1)
-
 	var timestampBuckets []int64
 	timestampBuckets = append(timestampBuckets, start.UnixMicro())
-	for current.Before(end) {
-		timestampBuckets = append(timestampBuckets, current.UnixMicro())
-		current = current.AddDate(0, 0, 1)
+	// When the queried time range is less than eight days, we show smaller buckets.
+	if *finerDrilldownTimeIncrements && time.Duration(endSec-startSec)*time.Second <= (8*24)*time.Hour {
```
tylerwilliams, line 614: you could clean this up a bit like:

```
const 8days = 8*24*time.Hour
queriedRangeDuration := end.Sub(start)
if *finerDrilldownTimeIncrements && queriedRangeDuration < 8days {
...
```

tylerwilliams (approved):
lgtm (but 2 small things)

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: sure yes done

enterprise/server/invocation_stat_service/invocation_stat_service.go
jdhollen: Fixed the function and flag name to suck a little less, comment is hopefully mostly redundant now but added it anyway.

PR #4723, opened 2023-09-07
dockerfile/executor: upgrade podman and skopeo

sluongng:
Upgrade to latest available version on Debian 11 repositories.

**Related issues**: N/A


sluongng:
Upgrading podman to >4.6.0 will give us a more verbose logging into [an issue reported in Slack](https://buildbuddy-corp.slack.com/archives/C0495TM9UUE/p1694073625339359)

I also bump skopeo while at it. If there is any objection, I could skip it for now.

tylerwilliams (approved):

PR #4728, opened 2023-09-07
Add ability for server admins to create impersonation API keys.

vadimberezniker:
These keys are short-lived (1h) and come with a more descriptive audit log event.

**Related issues**: N/A


tylerwilliams (approved):
LGTM but please get @bduffany's approval before submitting :)

enterprise/server/backends/authdb/authdb.go
```diff
@@ -537,11 +539,12 @@ func (d *AuthDB) createAPIKey(db *db.DB, userID, groupID, label string, caps []a
 		return nil, err
 	}
 	keyPerms := int32(0)
-	if userID == "" {
+	if ak.UserID == "" {
 		keyPerms = perms.GROUP_READ | perms.GROUP_WRITE
 	} else {
 		keyPerms = perms.OWNER_READ | perms.OWNER_WRITE
 	}
+	ak.Perms = keyPerms
```
bduffany, line 547: nit: consider cloning the API key or passing it by value, since it's easy to think that this returns the newly created key without modifying the original

```suggestion
	{
		clone := *ak
		ak = *clone
	}
	ak.Perms = keyPerms
```

enterprise/server/backends/authdb/authdb.go
```diff
@@ -748,6 +789,8 @@ func (d *AuthDB) GetAPIKeyForInternalUseOnly(ctx context.Context, groupID string
 		SELECT * FROM "APIKeys"
 		WHERE group_id = ?
 		AND (user_id IS NULL OR user_id = '')
+		AND impersonation = 0
```
bduffany, line 792: Would there be any benefit to having this predicate be `WHERE impersonation = <is_impersonating>` instead?

bduffany (approved):

enterprise/server/backends/authdb/authdb.go
vadimberezniker: Updated to pass by value.

enterprise/server/backends/authdb/authdb.go
vadimberezniker: No, I don't see a case at this time where we'd want to (internally) pick a impersonation key over a different group key.

PR #4736, opened 2023-09-08
Modify impersonation api key permission check.

vadimberezniker:
The API will be called after impersonation is already in effect in which case we can't rely on checking admin group membership.

Instead check the impersonation bit, which can only be set for authenticated admins.

**Related issues**: N/A


bduffany (approved):

vadimberezniker:
Will wait for LGTM from @tylerwilliams before merging.

tylerwilliams (approved):

PR #4738, opened 2023-09-08
When compressed data is requested, don't compare the size of the compressed files

luluz66:
**Related issues**: N/A
The size of compressed files can be different when we enable cdc, because we
concatenated zstd compressed chunks.

In this PR, when we request compressed data, we will decompress a certain percent (configuratble) the content of the data and then compare the decompressed size.

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -448,8 +448,9 @@ type doubleReader struct {
 	bytesReadSrc  int
 	bytesReadDest int
 	lastSrcErr    error
+	lastDestErr   error
 
-	decompressBuf    bytes.Buffer
+	decompressBuf    *bytes.Buffer
 	decompressor     io.WriteCloser
 	decompressSrcErr error
 }
@@ -489,6 +490,7 @@ func (d *doubleReader) Read(p []byte) (n int, err error) {
 				dstErr = io.EOF
 			}
 			d.bytesReadDest += dstN
+			d.lastDestErr = dstErr
 			return nil
 		})
 	}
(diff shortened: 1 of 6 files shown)


tylerwilliams (approved):

vadimberezniker (changes requested):

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -510,16 +571,41 @@ func (mc *MigrationCache) Reader(ctx context.Context, r *rspb.ResourceName, unco
 	eg := &errgroup.Group{}
 	var dstErr error
 	var destReader io.ReadCloser
+	decompressBuffer := &bytes.Buffer{}
```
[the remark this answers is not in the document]
luluz66: Used a pipe reader/write instead.

enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -549,11 +645,31 @@ func (mc *MigrationCache) Reader(ctx context.Context, r *rspb.ResourceName, unco
...
+		r:            r,
+		decompressor: decompressor,
+		pw:           pw,
+		pr:           pr,
+		mu:           sync.Mutex{},
+		done:         make(chan struct{}, 0),
+	}
+
+	go func() {
+		defer close(dr.done)
```
vadimberezniker, line 660: will this goroutine ever exit if shouldDecompressAndVerify is false?

vadimberezniker (approved):

enterprise/server/backends/migration_cache/migration_cache.go
luluz66: I added a check to only launch the goroutine when shouldDecompressAndVerify is true

PR #4739, opened 2023-09-08
[Release] Fix version tag and commit not getting embedded correctly

maggie-lou:
The commit and version keep getting set to UNKNOWN in version.go, ruining our app version metric. There seems to be a bug where x_defs does not consistently embed the data during releases.

In this commit we try another approach where we use a genrule that explicitly sets stamp = 1 and directly reads from bazel-out/stable-status.txt (described here: https://github.com/bazelbuild/bazel/issues/4942#issuecomment-1506315647)

There's some additional context at the bottom of this thread: https://buildbuddy-corp.slack.com/archives/C01D5GHRJ59/p1694026279817119

### DIFF
--- a/server/version/BUILD
+++ b/server/version/BUILD
@@ -1,15 +1,35 @@
 load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
+genrule(
+    name = "populate_version",
+    outs = ["version.txt"],
+    cmd_bash = """
+        version=$$(grep ^STABLE_VERSION_TAG bazel-out/stable-status.txt | cut -d' ' -f2); \
+        echo "$$version" > $@;
+    """,
+    stamp = 1,
+)
+
+genrule(
+    name = "populate_commit",
+    outs = ["commit.txt"],
+    cmd_bash = """
+        commit=$$(grep ^STABLE_COMMIT_SHA bazel-out/stable-status.txt | cut -d' ' -f2); \
+        echo "$$commit" > $@;
+    """,
+    stamp = 1,
+)
+
 go_library(
     name = "version",
     srcs = ["version.go"],
+    embedsrcs = [
+        "version.txt",
+        "commit.txt",
+    ],
     importpath = "github.com/buildbuddy-io/buildbuddy/server/version",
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #4739** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4739" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4739?utm_source=stack-comment).

server/version/BUILD
```diff
@@ -1,15 +1,35 @@
 load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
+genrule(
+    name = "populate_version",
+    outs = ["version.txt"],
+    cmd_bash = """
+        version=$$(grep ^STABLE_VERSION_TAG bazel-out/stable-status.txt | cut -d' ' -f2); \
+        echo "$$version" > $@;
```
bduffany, line 9: nit: to avoid a newline after the version, do:

```suggestion
        printf '%s' $$version > $@;
```

server/version/BUILD
```diff
@@ -1,15 +1,35 @@
...
+    """,
+    stamp = 1,
+)
+
+genrule(
+    name = "populate_commit",
+    outs = ["commit.txt"],
+    cmd_bash = """
+        commit=$$(grep ^STABLE_COMMIT_SHA bazel-out/stable-status.txt | cut -d' ' -f2); \
+        echo "$$commit" > $@;
```
bduffany, line 19:
```suggestion
        printf '%s' $$commit > $@;
```

bduffany (approved):
sanity check: if we do a dev push twice, does the second app push show up as a NOP?

maggie-lou:
> sanity check: if we do a dev push twice with `--stamp`, does the second app push become a NOP?

Unfortunately this doesn't fix the 'apps being double released problem'. If you do a dev push twice with the same target, the second one is a NOP. In our release flow we use two different targets, and the apps are being re-pushed. More context [here](https://buildbuddy-corp.slack.com/archives/C01D5GHRJ59/p1694543456364689)

PR #4741, opened 2023-09-08
Start redirecting users to subdomain even on the main site.

vadimberezniker:
This applies to both the group picker and for creating new orgs.

**Related issues**: N/A

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -198,9 +198,8 @@ export class AuthService {
 
     this.setCookie(SELECTED_GROUP_ID_COOKIE, groupId);
 
-    // If we're on a subdomain and the new group is on a different subdomain then
-    // we have to use a redirect.
-    if (capabilities.config.customerSubdomain && new URL(groupURL).hostname != window.location.hostname) {
+    // If the new group is on a different subdomain then we have to use a redirect.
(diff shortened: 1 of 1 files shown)


app/auth/auth_service.ts
```diff
@@ -198,9 +198,8 @@ export class AuthService {
 
     this.setCookie(SELECTED_GROUP_ID_COOKIE, groupId);
 
-    // If we're on a subdomain and the new group is on a different subdomain then
-    // we have to use a redirect.
-    if (capabilities.config.customerSubdomain && new URL(groupURL).hostname != window.location.hostname) {
+    // If the new group is on a different subdomain then we have to use a redirect.
+    if (capabilities.config.subdomainsEnabled && new URL(groupURL).hostname != window.location.hostname) {
```
bduffany, line 202: quick sanity check, can the groupURL be empty here? (We have groups without url_identifier set, but not sure if groupURL would just be `app.buildbuddy.io` in that case)

bduffany (approved):

app/auth/auth_service.ts
vadimberezniker: The backend sets the URL to `app.buildbuddy.io` if the group doesn't have an identifier.

PR #4743, opened 2023-09-11
Support specifying which soci-store to run via a flag

iain-macdonald:
**Related issues**: [2570](https://github.com/buildbuddy-io/buildbuddy-internal/issues/2570)

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -7071,19 +7071,19 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
     )
 
     http_file(
-        name = "com_github_buildbuddy_io_soci_snapshotter-soci-store-v0.0.8-linux-amd64",
+        name = "com_github_buildbuddy_io_soci_snapshotter-soci-store-linux-amd64",
         urls = ["https://storage.googleapis.com/buildbuddy-tools/binaries/soci-snapshotter/soci-store-v0.0.8-linux-amd64"],
         sha256 = "978febe41046d9338e766ddd0a8f3f46d798b7f5b7a537840f4d9c2f814a759a",
         executable = True,
-        downloaded_file_path = "soci-store-v0.0.8-linux-amd64",
+        downloaded_file_path = "soci-store",
     )
 
     http_file(
-        name = "com_github_buildbuddy_io_soci_snapshotter-soci-store-v0.0.9-linux-amd64-race",
+        name = "com_github_buildbuddy_io_soci_snapshotter-soci-store-linux-amd64-race",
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -64,8 +64,11 @@ var (
 
 	// TODO(iain): delete executor.podman.run_soci_snapshotter flag once
 	// the snapshotter doesn't need root permissions to run.
-	sociStoreLogLevel            = flag.String("executor.podman.soci_store_log_level", "", "The level at which the soci-store should log. Should be one of the standard log levels, all lowercase.")
-	runSociStoreLocally          = flag.Bool("executor.podman.run_soci_store", true, "If true, runs the soci store locally if needed for image streaming.")
+	sociStoreLogLevel   = flag.String("executor.podman.soci_store_log_level", "", "The level at which the soci-store should log. Should be one of the standard log levels, all lowercase.")
+	runSociStoreLocally = flag.Bool("executor.podman.run_soci_store", true, "If true, runs the soci store locally if needed for image streaming.")
+
+	// TODO(https://github.com/buildbuddy-io/buildbuddy-internal/issues/2570): remove this flag
+	runSociStoreRaceDetector     = flag.Bool("executor.podman.dev_only_soci_store_race_detector", false, "If true, runs the soci store with race detection enabled (up to 10x slower). Only for us in dev.")
```
vadimberezniker, line 71: instead of adding a separate flag, what do you think about adding a flag for the path to the soci binary?

deps.bzl
```diff
@@ -7078,6 +7078,14 @@ def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
         downloaded_file_path = "soci-store-v0.0.8-linux-amd64",
     )
 
+    http_file(
+        name = "com_github_buildbuddy_io_soci_snapshotter-soci-store-v0.0.9-linux-amd64-race",
```
vadimberezniker, line 7082: remove the version from the rule & downloaded file name to make your life easier

vadimberezniker (changes requested):

enterprise/server/remote_execution/containers/podman/podman.go
iain-macdonald: Sounds good to me, I went ahead and combined it with the run-locally flag.

deps.bzl
iain-macdonald: Good call, thanks!

vadimberezniker (approved):

PR #4746, opened 2023-09-11
Add IP rules service for manipulating and checking IP rules.

vadimberezniker:
There's a new group-level settings that determines whether IP rules are applied.

Currently, IP rules are a simple unordered list of allowed IP ranges.

IP rules are cached to avoid having to load them for every request.

Next PR hooks up the service for enforcement.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/iprules/BUILD
+++ b/enterprise/server/iprules/BUILD
@@ -1,10 +1,11 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
+package(default_visibility = ["//enterprise:__subpackages__"])
+
 go_library(
     name = "iprules",
     srcs = ["iprules.go"],
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/iprules",
-    visibility = ["//visibility:public"],
     deps = [
         "//proto:iprules_go_proto",
         "//server/environment",
@@ -19,8 +20,6 @@ go_library(
         "//server/util/status",
     ],
 )
-
-package(default_visibility = ["//enterprise:__subpackages__"])
 
 go_test(
     name = "iprules_test",
--- a/enterprise/server/iprules/BUILD
+++ b/enterprise/server/iprules/BUILD
@@ -1,0 +1,41 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "iprules",
+    srcs = ["iprules.go"],
+    importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/iprules",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//proto:iprules_go_proto",
+        "//server/environment",
+        "//server/interfaces",
+        "//server/tables",
+        "//server/util/alert",
+        "//server/util/authutil",
+        "//server/util/clientip",
+        "//server/util/lru",
+        "//server/util/perms",
+        "//server/util/role",
+        "//server/util/status",
+    ],
+)
+
+package(default_visibility = ["//enterprise:__subpackages__"])
+
+go_test(
+    name = "iprules_test",
+    srcs = ["iprules_test.go"],
+    deps = [
+        ":iprules",
+        "//enterprise/server/testutil/enterprise_testauth",
+        "//enterprise/server/testutil/enterprise_testenv",
+        "//proto:context_go_proto",
+        "//proto:iprules_go_proto",
+        "//server/environment",
+        "//server/testutil/testauth",
+        "//server/util/clientip",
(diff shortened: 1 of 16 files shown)


enterprise/server/iprules/BUILD
```diff
@@ -0,0 +1,41 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "iprules",
+    srcs = ["iprules.go"],
+    importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/iprules",
+    visibility = ["//visibility:public"],
```
tylerwilliams, line 7: did you mean for this to be public?

enterprise/server/iprules/iprules.go
```diff
@@ -0,0 +1,261 @@
...
+	irpb "github.com/buildbuddy-io/buildbuddy/proto/iprules"
+)
+
+var (
+	enableIPRules = flag.Bool("auth.ip_rules.enable", false, "If true, IP rules will be checked during auth.")
+	cacheTTL      = flag.Duration("auth.ip_rules.cache_ttl", 5*time.Minute, "Duration of time IP rules will be cached in memory.")
+)
+
+const (
+	cacheSize = 100_000
```
tylerwilliams, line 30: nit: comment

enterprise/server/iprules/iprules.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (s *Service) Authorize(ctx context.Context) error {
+	u, err := perms.AuthenticatedUser(ctx, s.env)
+	if authutil.IsAnonymousUserError(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
```
tylerwilliams, line 190: Idk if this is much better but you could.
```
	if err != nil {
    	if authutil.IsAnonymousUserError(err) {
	    	return nil
    	}
		return err
	}
```

Do we want anon users returning no error here?

enterprise/server/iprules/iprules.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (s *Service) AddRule(ctx context.Context, req *irpb.AddRuleRequest) (*irpb.AddRuleResponse, error) {
+	u, err := perms.AuthenticatedUser(ctx, s.env)
+	if err != nil {
+		return nil, err
+	}
+	if err := authutil.AuthorizeGroupRole(u, req.GetRequestContext().GetGroupId(), role.Admin); err != nil {
+		return nil, err
+	}
```
tylerwilliams, line 211: could you have a 

`func (s *Service) checkPreconditions(ctx, reqCtx) error` method that does this since it's the same check in all 3 places? (Add|Update|Delete)

proto/iprules.proto
```diff
@@ -0,0 +1,56 @@
...
+  context.RequestContext request_context = 1;
+
+  IPRule rule = 2;
+}
+
+message AddRuleResponse {
+  context.ResponseContext response_context = 1;
+
+  IPRule rule = 2;
+}
```
tylerwilliams, line 26: This is pretty close to Update -- could we get away with just having Update?

proto/iprules.proto
```diff
@@ -0,0 +1,56 @@
+syntax = "proto3";
```
tylerwilliams, line 1: very minor nit but next time it would be nice to send just the proto first :)

tylerwilliams (changes requested):

enterprise/server/iprules/iprules.go
```diff
@@ -0,0 +1,261 @@
...
+		if err := s.env.GetDBHandle().DB(ctx).ScanRows(rows, r); err != nil {
+			return nil, err
+		}
+		_, ipNet, err := net.ParseCIDR(r.CIDR)
+		if err != nil {
+			alert.UnexpectedEvent("unparsable CIDR rule", "rule %q", r.CIDR)
+			continue
+		}
+		allowed = append(allowed, ipNet)
+	}
```
bduffany, line 144: uber nit: check `rows.Err()`

bduffany (approved):

enterprise/server/iprules/iprules.go
vadimberezniker: For anon users, we don't have a group to check against. Anon users will be handled by regular auth checks.

proto/iprules.proto
vadimberezniker: Having Add and Update as separate operations avoids the need to have condition DB logic in the implementation for INSERT vs UPDATE. I prefer having separate operations.

tylerwilliams (approved):

PR #4749, opened 2023-09-12
Use a single block device for the VM rootfs.

bduffany:
Rather than using a readonly containerfs using the firecracker drive API and a read/write NBD overlay on top of that, use a single NBD for the rootfs that preserves the readonly property of the containerfs via its copy-on-write behavior.

This has a few benefits:
- Simpler - there are fewer block devices in total. Instead of containerfs + scratchfs, it's now just a single rootfs. This means we don't need to set up the overlay filesystem in the VM as well as think about caching both artifacts separately.
- Better I/O performance - the root FS will no longer be an overlayfs and will instead be an ext4 fs. The ext4 driver should be faster since it won't need to check both the top and bottom layer for every read/write. This also will allow us to run docker using the overlay2 driver. Before, it had to use the slower vfs driver, because the rootfs was already an overlayfs, and overlay-on-overlay is not supported.
- More flexible - customers can now run more interesting workloads which themselves use overlayfs, now that the root device is using ext4 instead of overlayfs.

**Related issues**: N/A


enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -363,9 +364,25 @@ func (c *COWStore) ReadAt(p []byte, off int64) (int, error) {
...
+				dataSize = max(0, remainder)
+			}
+			if dataSize > 0 {
+				if _, err := readFullAt(chunk, p[:dataSize], chunkRelativeOffset); err != nil {
+					return n, err
+				}
+			}
+			for off := dataSize; off < int64(readSize); off++ {
+				p[off] = 0
+			}
```
bduffany, line 385: @maggie-lou I know we talked a little bit about always making things chunk-aligned so that we wouldn't have to handle the last-chunk case here like this, but as I started to do the chunk alignment I started to second-guess whether that approach would cause more problems than it solved. e.g. it breaks a reasonable assumption that `size(file) == size(fileConvertedToCOW)` which could cause subtle bugs later.

That said, if this lazy approach seems too complex or seems like it will make the caching stuff more difficult, we can rethink this

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -314,23 +342,81 @@ func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotK
...
+		return nil, status.WrapError(err, "convert image to COW")
+	}
+	// Add the COW to cache. This will also compute chunk digests.
+	cf := &ChunkedFile{COWStore: cow}
+	if _, err := l.cacheCOW(ctx, "rootfs.ext4", cf); err != nil {
+		return nil, err
+	}
+	log.Debugf("Converted containerfs to COW in %s", time.Since(start))
+	return cf, nil
+}
```
bduffany, line 379: I just realized that we can simplify all of this if we make all fields optional in `Snapshot` (i.e. don't require a memory / vmstate file). Then, the containerfs artifact can just be a snapshot and reuse most of the existing code for getting snapshots from cache, validating snapshots etc. wdyt?

enterprise/server/cmd/goinit/main.go
```diff
@@ -452,6 +471,81 @@ func main() {
...
+	if errno != 0 {
+		return status.InternalErrorf("EXT4_IOC_RESIZE_FS: errno %s", errno)
+	}
+	return nil
+}
+
+// startRootfsClientProcess starts a background process that acts as the NBD
+// client for the rootfs. It creates the rootfs NBD device and handles all IO
+// requests to it. This function returns when the rootfs is mounted or the
+// rootfs
```
tylerwilliams, line 502: don't leave us hanging!

enterprise/server/cmd/goinit/main.go
```diff
@@ -452,6 +471,81 @@ func main() {
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	errCh := make(chan error)
+	go func() { errCh <- cmd.Wait() }()
+	for {
+		// TODO(bduffany): Do something better than file polling here.
+		_, err := os.Stat("/.rootfs_ready")
```
tylerwilliams, line 514: could we make "./rootfs_ready" a const?

enterprise/server/cmd/goinit/main.go
```diff
@@ -452,6 +471,81 @@ func main() {
 	syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
 }
 
+// Resizes the rootfs ext4 filesystem to fill the available space on the block
+// device.
```
tylerwilliams, line 475: nit: update comment since it resizes to `size`?

tylerwilliams (commented):
you want to send just the goinit change first and get that out of the way?

enterprise/server/cmd/goinit/main.go
```diff
@@ -452,6 +471,81 @@ func main() {
...
+func resizeExt4FS(mountPath string, size int64) error {
+	s := &syscall.Statfs_t{}
+	if err := syscall.Statfs(mountPath, s); err != nil {
+		return status.InternalErrorf("statfs %s: %s", mountPath, err)
+	}
+	blocks := size / s.Bsize
```
maggie-lou, line 481: What if size is not a multiple of Bsize? Do we need to +1 for the last block

enterprise/server/cmd/goinit/main.go
```diff
@@ -452,6 +471,81 @@ func main() {
...
+	cmd := exec.Command(os.Args[0], append(os.Args[1:], "--rootfs_client")...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	errCh := make(chan error)
+	go func() { errCh <- cmd.Wait() }()
+	for {
+		// TODO(bduffany): Do something better than file polling here.
```
maggie-lou, line 513: I see it's a TODO so no pressure to do it now, but one potential implementation is using unix.Poll on a file descriptor like we do [here](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/uffd/uffd.go#L227)

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -363,9 +364,25 @@ func (c *COWStore) ReadAt(p []byte, off int64) (int, error) {
 				p[i] = 0
 			}
 		} else {
-			if _, err := readFullAt(chunk, p[:readSize], chunkRelativeOffset); err != nil {
+			s, err := chunk.SizeBytes()
```
maggie-lou, line 367: Could you rename SizeBytes to something like DataSize? The different types of `sizes` is a bit confusing

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -363,9 +364,25 @@ func (c *COWStore) ReadAt(p []byte, off int64) (int, error) {
...
+			if err != nil {
 				return n, err
 			}
+			// Chunk might have less data available than the calculated size
+			// if this was the last chunk when we resized. If so then fill
+			// the range from [dataSize, readSize) with 0s.
+			dataSize := int64(readSize)
+			if remainder := s - chunkRelativeOffset; dataSize > remainder {
```
maggie-lou, line 375: nit: Change to `if readSize > remainder` - I find that more intuitive to understand

enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -523,6 +551,16 @@ func (s *COWStore) copyChunkIfNotDirty(chunkStartOffset int64) (err error) {
...
 	}()
 
+	// note: the src chunk might be smaller than the dst chunk if we resized
+	// and now we're copying the last chunk, since resizing is done lazily.
+	size, err := src.SizeBytes()
+	if err != nil {
+		return err
+	}
+	if size > dstSize {
+		return status.InternalErrorf("chunk source size %d is greater than dest size %d; this is a bug", size, dstSize)
```
maggie-lou, line 561: lol after reading the word `size` so many times it's starting to look really weird to me

enterprise/server/remote_execution/snaploader/snaploader.go
maggie-lou: Yeah I like reusing the existing architecture. It's confusing that we have a couple functions that are very similar but slightly different

bduffany:
> you want to send just the goinit change first and get that out of the way?

Yeah, let me mark this as draft and split this up into a few smaller PRs. The `blockio` changes also make sense as a standalone PR

tylerwilliams (approved):
LGTM but please get @maggie-lou's approval before submitting :)

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -902,6 +917,36 @@ func (c *FirecrackerContainer) initScratchImage(ctx context.Context, path string
...
+	}()
+	// Resize the containerfs to match the disk size requested for the action.
+	// Note, this operation doesn't affect the original containerfs, and is also
+	// cheap (just updates size metadata).
+	curSize, err := cf.SizeBytes()
+	if err != nil {
+		return status.WrapError(err, "get container image size")
+	}
+	newSize := curSize + minScratchDiskSizeBytes + c.vmConfig.ScratchDiskSizeMb*1e6
+	if _, err := cf.Resize(newSize); err != nil {
```
maggie-lou, line 942: If there is an error here, the defer won't close the cf

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -107,24 +115,25 @@ func (s *Snapshot) GetVMConfiguration() *fcpb.VMConfiguration {
...
 type CacheSnapshotOptions struct {
-	// The following fields are all required.
 	VMConfiguration     *fcpb.VMConfiguration
 	VMStateSnapshotPath string
 	KernelImagePath     string
 	InitrdImagePath     string
-	ContainerFSPath     string
 
 	// MemSnapshotPath is the memory snapshot file path. It is required if the
 	// memory file is not represented as a ChunkedFile.
```
maggie-lou, line 131: nit: remove/modify this comment

maggie-lou (commented):
Mostly LGTM other than that 1 small bug

PR #4752, opened 2023-09-12
Add metrics to track the number of chunks in pebble cache when cdc is enabled

luluz66:
**Related issues**: N/A

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -2224,6 +2224,16 @@ var (
 		GroupID,
 	})
 
+	PebbleCacheNumCDCChunks = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: bbNamespace,
+		Subsystem: "remote_cache",
+		Name:      "pebble_cache_num_cdc_chunks",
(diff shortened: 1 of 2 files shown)


server/metrics/metrics.go
```diff
@@ -2224,6 +2224,16 @@ var (
 		GroupID,
 	})
 
+	PebbleCacheNumCDCChunks = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: bbNamespace,
+		Subsystem: "remote_cache",
+		Name:      "pebble_cache_num_cdc_chunks",
```
vadimberezniker, line 2230: Should we call this something like chunks_per_file_count? The name sounds like it's counting total number of chunks in the cache.

vadimberezniker (changes requested):

server/metrics/metrics.go
luluz66: Good point. I named it like pebble_cache_num_chunks_per_file instead. Trying to avoid "_count" at the end b/c it might sound like a Count metrics instead of a Histogram.

vadimberezniker (approved):

PR #4766, opened 2023-09-13
Incorporate IP access check into interceptors.

vadimberezniker:
- For HTTP, check all requests except GetUser which will be used to inform the frontend as to whether user has access to the target group.

   I removed the HTTP Path prefix stripping from the interceptors so that it's always possible for interceptors to know the requested URI.

 - For gRPC all requests are checked.

**Related issues**: N/A

### DIFF
--- a/server/role_filter/role_filter.go
+++ b/server/role_filter/role_filter.go
@@ -3,7 +3,7 @@ package role_filter
 import (
 	"context"
 	"flag"
-	"strings"
+	"path"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/util/authutil"
@@ -202,9 +202,7 @@ func ServerAdminOnlyRPCs() []string {
 func AuthorizeRPC(ctx context.Context, env environment.Env, rpcName string) error {
 	// Strip off the service name since it varies depending on whether it was
 	// accessed via protolet or gRPC.
-	if si := strings.LastIndex(rpcName, "/"); si >= 0 {
-		rpcName = rpcName[si+1:]
-	}
+	rpcName = path.Base(rpcName)
 
 	if stringSliceContains(RoleIndependentRPCs(), rpcName) {
 		return nil
--- a/server/role_filter/role_filter.go
+++ b/server/role_filter/role_filter.go
@@ -3,6 +3,7 @@ package role_filter
 import (
 	"context"
 	"flag"
+	"strings"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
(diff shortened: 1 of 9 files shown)


server/rpc/interceptors/interceptors.go
```diff
@@ -195,6 +195,28 @@ func roleAuthUnaryServerInterceptor(env environment.Env) grpc.UnaryServerInterce
...
+	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		if irs := env.GetIPRulesService(); irs != nil {
+			if err := irs.Authorize(stream.Context()); err != nil {
+				return err
+			}
+		}
+		return handler(srv, stream)
+	}
+}
+
```
bduffany, line 219: have you tried running `cacheload` against a local server w/ IP auth enabled to see whether it significantly affects throughput?

vadimberezniker: I have not, but is there a reason to believe it would affect perf? Apart from loading IP rules from DB on a cache miss, the rest of the operations are fast.

tylerwilliams: I agree it would be nice to do a quick check here just in case. It's relatively cheap to do and will let us know we didn't miss something silly (like if the cache were busted somehow and we were reading from DB every time or something)

server/role_filter/role_filter.go
```diff
@@ -199,6 +200,12 @@ func ServerAdminOnlyRPCs() []string {
 // If the RPC accesses any specific resources within the org, further
 // authorization checks may be needed beyond this coarse-grained filter.
 func AuthorizeRPC(ctx context.Context, env environment.Env, rpcName string) error {
+	// Strip off the service name since it varies depending on whether it was
+	// accessed via protolet or gRPC.
+	if si := strings.LastIndex(rpcName, "/"); si >= 0 {
+		rpcName = rpcName[si+1:]
+	}
```
tylerwilliams, line 207: could you use path.Base for this?

tylerwilliams (approved):

server/role_filter/role_filter.go
vadimberezniker: done

server/rpc/interceptors/interceptors.go
vadimberezniker: 95% latency, in milliseconds:
rule_check_latency

bduffany (approved):

server/rpc/interceptors/interceptors.go
vadimberezniker: and now it's down to 0.01 ms after fixing the unnecessary group lookup

PR #4769, opened 2023-09-14
bazelrc: remote-target-linux use platform in output dir, stop repo RE

sluongng:
This flag changes the name of output directories under `bazel-out/*` to
contain the platform information. This means that if a user were to
switch between a local build and a remote build
using `--config=remote-target-linux`, they would be able to retain the
output from both configs.

Also disabling `experimental_repo_remote_exec`. This flag is meant to
allow repository rules, when marked with `remotable = True` to be
executed remotely with RBE. The main purpose of this is to allow
toolchains to probe remote execution platforms information to setup
appropriate toolchains and paths. We do not use this effect in our
repository anywhere, however it will make us re-run all repository rules
when toggle between `--config=remote-target-linux` build and local build.

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -32,7 +32,6 @@ build:cache --experimental_remote_cache_compression
 build:remote-shared --remote_upload_local_results
 build:remote-shared --remote_timeout=600
 build:remote-shared --remote_download_minimal
-build:remote-shared --experimental_repo_remote_exec
 build:remote-shared --jobs=100
 build:remote-shared --verbose_failures
 build:remote-shared --platforms=@buildbuddy_toolchain//:platform_linux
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -68,8 +67,10 @@ build:target-linux --host_cpu=k8
 #   bazel test --config=x //...
 build:remote-target-linux --config=target-linux
 build:remote-target-linux --config=remote
+build:remote-target-linux --experimental_platform_in_output_dir
 build:remote-target-linux-dev --config=target-linux
 build:remote-target-linux-dev --config=remote-dev
+build:remote-target-linux-dev --experimental_platform_in_output_dir
```
bduffany, line 73: would it make sense to put this in `build:target-linux`?

.bazelrc
```diff
@@ -32,7 +32,6 @@ build:cache --experimental_remote_cache_compression
 build:remote-shared --remote_upload_local_results
 build:remote-shared --remote_timeout=600
 build:remote-shared --remote_download_minimal
-build:remote-shared --experimental_repo_remote_exec
```
bduffany, line 35: not sure if there was a reason we added this originally, maybe @siggisim remembers - but at a glance it seems OK to remove

bduffany (approved):

.bazelrc
siggisim: Just wanted to try it out

.bazelrc
sluongng: I would be better to put it in "build" as default. But I was trying to be most conservative here.

Could move to target-linux as well.

PR #4770, opened 2023-09-14
Add a resize function to COWStore

bduffany:
Adds a `Resize()` function which allows right-padding the COWStore with zeroes. This will let us merge the containerfs and scratchfs into a single disk - instead of layering an empty scratchfs onto the containerfs, we can just resize the containerfs.

The right-padding is implemented by just updating a single int (the totalSize), so that the right-padding is purely virtual / lazy. It avoids accessing any of the underlying chunks, which means we preserve our current copy-on-write behavior while also not needing to eagerly fetch any chunks from cache (particularly the last chunk).

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/blockio/blockio.go
+++ b/enterprise/server/remote_execution/blockio/blockio.go
@@ -350,7 +350,8 @@ func (c *COWStore) ReadAt(p []byte, off int64) (int, error) {
 	n := 0
 	for len(p) > 0 {
 		chunkRelativeOffset := off % c.chunkSizeBytes
-		readSize := int(c.chunkSizeBytes - chunkRelativeOffset)
+		chunkCalculatedSize := c.calculateChunkSize(chunkOffset)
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/blockio/blockio.go
```diff
@@ -349,27 +349,68 @@ func (c *COWStore) ReadAt(p []byte, off int64) (int, error) {
 	chunkOffset := c.chunkStartOffset(off)
 	n := 0
 	for len(p) > 0 {
```
maggie-lou, line 351: Not introduced in your PR, but do we want to return an error if we try to read bytes beyond the max size of the store?

maggie-lou (approved):
It's a lot so thanks for adding all the comments

enterprise/server/remote_execution/blockio/blockio.go
bduffany: good catch, will send a separate one for that

tylerwilliams (approved):

PR #4771, opened 2023-09-14
Use GetUser to communicate if user can't access selected group.

vadimberezniker:
Use it to communicate if the user has selected a group that they have access to but cannot access due to IP rules. Also use the same mechanism for communicating subdomain access issues.

Update UI to show a minimal sidebar if user doesn't have access to the selceted group so that the user still has access to the group switcher.

**Related issues**: N/A

### DIFF
--- a/enterprise/app/sidebar/sidebar.tsx
+++ b/enterprise/app/sidebar/sidebar.tsx
@@ -33,6 +33,7 @@ import router, { Path } from "../../../app/router/router";
 import rpcService from "../../../app/service/rpc_service";
 import rpc_service from "../../../app/service/rpc_service";
 import { grp } from "../../../proto/group_ts_proto";
+import { user } from "../../../proto/user_ts_proto";
 
 interface Props {
   user?: User;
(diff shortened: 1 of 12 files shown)


enterprise/app/sidebar/sidebar.tsx
```diff
@@ -154,80 +155,87 @@ export default class SidebarComponent extends React.Component<Props, State> {
...
-            <SidebarLink selected={this.isCodeSelected()} href={Path.codePath} title="Code">
-              <Code className="icon" />
-              <span className="sidebar-item-text">Code</span>
-            </SidebarLink>
-          )}
-          <SidebarLink selected={this.isSetupSelected()} href={Path.setupPath} title="Quickstart">
-            <Terminal className="icon" />
-            <span className="sidebar-item-text">Quickstart</span>
-          </SidebarLink>
+          {this.props.user?.selectedGroupAccess === user.SelectedGroup.Access.ALLOWED && (
```
bduffany, line 158: Maybe Siggi has thoughts here as well, but to avoid introducing another visibility dimension for each sidebar item, we could also just hide all of the sidebar items if the user doesn't have access.

siggisim: Could we just show the full sidebar? If they click on anything other than the group picker, presumably they'll get the same "Access not allowed" page. Not sure how nice a mostly empty sidebar would look anyway

vadimberezniker: The access check was only being performed once on login. I updated the check to happen on every navigation and reverted the change to hide sidebar items.

bduffany (approved):

PR #4772, opened 2023-09-14
Set go 1.21 in go.mod

bduffany:
We have it in WORKSPACE already, but set it to 1.21 in go.mod so that we can start using the new language features without editor warnings.

**Related issues**: N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/buildbuddy-io/buildbuddy
 
-go 1.20
+go 1.21
 
 replace (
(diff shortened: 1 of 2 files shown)


sluongng (approved):
Lgtm.

I think we might need to upgrade internal as well, but that could be done separately.

bduffany:
@sluongng we don't have an internal `go.mod` file, but I sent https://github.com/buildbuddy-io/buildbuddy-internal/pull/2647 to upgrade the WORKSPACE version

PR #4777, opened 2023-09-14
Add convenience `flag` package that supports tagging.

vadimberezniker:
It has the same API as the standard package, but allows you to specify optional tags eliminating the need to have a mixture of `flag.Foo` and `flagutil.New` definitions.

It also includes Slice and Struct types to remove the need for the bare `flagutil.New` calls.

**Related issues**: N/A


tylerwilliams (approved):
looks fine to me

tempoz (approved):
LGTM

PR #4778, opened 2023-09-14
Add a flag to goinit to allow mounting a single rootfs disk.

bduffany:
Split out from #4749.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/goinit/main.go
+++ b/enterprise/server/cmd/goinit/main.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/nbd/nbdclient"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/vsock"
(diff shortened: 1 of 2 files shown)


enterprise/server/cmd/goinit/main.go
```diff
@@ -452,6 +475,80 @@ func main() {
 	syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
 }
 
+// Resizes the ext4 filesystem mounted at the given path to the given size.
+func resizeExt4FS(mountPath string, size int64) error {
+	s := &syscall.Statfs_t{}
+	if err := syscall.Statfs(mountPath, s); err != nil {
+		return status.InternalErrorf("statfs %s: %s", mountPath, err)
+	}
+	blocks := size / s.Bsize
```
bduffany, line 484: @maggie-lou to answer your question from the other PR, if the ext4 FS isn't perfectly aligned to the ext4 block size (4096) then this division will mean we have a few bytes of leftover data at the end of the file, which seems OK.

tylerwilliams (approved):

enterprise/server/cmd/goinit/main.go
maggie-lou: Worst case it would be 4095B but agree that probably nothing would come up in practice

PR #4782, opened 2023-09-14
Support specifying version numbers when using `bb add`

siggisim:
Previously bb add would always just fetch the latest version of a dependency

### DIFF
--- a/cli/add/add.go
+++ b/cli/add/add.go
@@ -238,7 +238,7 @@ type RegistryResponse struct {
 	LatestReleaseWithModuleSnippet    string           `json:"latest_release_with_module_snippet"`
 	Disambiguation                    []Disambiguation `json:"disambiguation"`
 	Repo                              Repo             `json:"repo"`
-	Releases						  []Release 	   `json:"releases"`
+	Releases                          []Release        `json:"releases"`
 }
 
 type Disambiguation struct {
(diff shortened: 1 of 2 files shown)


cli/add/add.go
```diff
@@ -105,44 +109,56 @@ func HandleAdd(args []string) (int, error) {
...
-func GenerateSnippet(module string, resp *RegistryResponse) string {
-	versionKey := fmt.Sprintf("[https://registry.build/%s@%s]", module, resp.LatestReleaseWithWorkspaceSnippet)
+func GenerateSnippet(module, version string, resp *RegistryResponse) string {
+	versionKey := fmt.Sprintf("[https://registry.build/%s@%s]", module, version)
 	header := fmt.Sprintf(headerTemplate, versionKey)
 	footer := fmt.Sprintf(footerTemplate, versionKey)
-	return fmt.Sprintf("\n%s\n\n%+v\n\n%s\n", header, strings.TrimSpace(resp.WorkspaceSnippet), footer)
+	snippet := resp.WorkspaceSnippet
+	for _, r := range resp.Releases {
+		if r.Name == "v"+version || r.Name == version {
```
tylerwilliams, line 156: 👀

tylerwilliams (approved):

cli/add/add.go
siggisim: Haha it's not strictly required, just adds some flexibility in specifying version names

PR #4784, opened 2023-09-15
Fix 0 chunk error

luluz66:
**Related issues**: N/A
Run the server locally with --@io_bazel_rules_go//go/config:race, turn on
cdc with pebble cache, and then run a build (e.g. //app/...)  against the local 
bb server triggers DATA RACE and 0 chunk errors

I added a test where we cancel the context before writing any data. It can reproduce both the data race and 0 chunk error. 
- Add mutex in `cdcWriter` to protect writtenChunks and numChunks.
- In commit, instead of calling `cdcw.Close()` we call `cdcw.chunker.Close()`. In `cdcw.Close()` we ignore the error from `cdcw.chunker.Close()`. It causes us to attempt to write 0 chunks if we cancel the context before any data is written. Returning `cdcw.chunker.Close()` in `cdcw.Close()` doesn't work because we call cdcw.Close() twice: once in CommitFn and once when we close the writer returned by pebble cache. The second time Close() will return a "context canceled" error.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1847,7 +1847,7 @@ type cdcWriter struct {
 	firstChunk []byte
 	fileType   rfpb.FileMetadata_FileType
 
-	mu *sync.Mutex // protects writtenChunks, numChunks, firstChunk, fileType
+	mu sync.Mutex // protects writtenChunks, numChunks, firstChunk, fileType
 	eg *errgroup.Group
 }
 
@@ -1867,7 +1867,6 @@ func (p *PebbleCache) newCDCCommitedWriteCloser(ctx context.Context, fileRecord 
 		fileRecord:     fileRecord,
 		shouldCompress: shouldCompress,
 		isCompressed:   isCompressed,
-		mu:             &sync.Mutex{},
 	}
 	var wc, decompressor io.WriteCloser
 	wc = cdcw
@@ -1905,17 +1904,15 @@ func (p *PebbleCache) newCDCCommitedWriteCloser(ctx context.Context, fileRecord 
 		}
 		cdcw.mu.Lock()
 		defer cdcw.mu.Unlock()
-		isSingleChunk := cdcw.numChunks == 1
 
-		if isSingleChunk {
-			firstChunk := cdcw.firstChunk
+		if cdcw.numChunks == 1 {
 			cdcw.fileType = rfpb.FileMetadata_COMPLETE_FILE_TYPE
 			// When there is only one single chunk, we want to store the original
 			// file record with the original key instead of computed digest from
 			// the chunkData. This is because the chunkData can be compressed or
 			// encrypted, so the digest computed from it will be different from
 			// the original digest.
-			return cdcw.writeRawChunk(cdcw.fileRecord, cdcw.key, firstChunk)
+			return cdcw.writeRawChunk(cdcw.fileRecord, cdcw.key, cdcw.firstChunk)
 		}
 		now := p.clock.Now().UnixMicro()
 
@@ -1943,10 +1940,10 @@ func (p *PebbleCache) newCDCCommitedWriteCloser(ctx context.Context, fileRecord 
 func (cdcw *cdcWriter) writeChunk(chunkData []byte) error {
 	cdcw.mu.Lock()
 	defer cdcw.mu.Unlock()
-	cdcw.numChunks++
-	numChunks := cdcw.numChunks
 
-	if numChunks == 1 {
+	cdcw.numChunks++
+
+	if cdcw.numChunks == 1 {
 		// We will wait to write the first chunk until either cdcw.Commit() is
 		// called or the second chunk is encountered.
 		// In the former case, there is only one chunk, we don't want to write a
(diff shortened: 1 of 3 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1840,12 +1840,14 @@ type cdcWriter struct {
...
+	chunker         *chunker.Chunker
+	writtenChunks   []*rspb.ResourceName
+	writtenChunksMu *sync.Mutex // protects writtenChunks
 
-	numChunks  int
-	firstChunk []byte
-	fileType   rfpb.FileMetadata_FileType
+	numChunks   int
+	numChunksMu *sync.Mutex // protects numChunks
+	firstChunk  []byte
```
vadimberezniker, line 1849: it looks like firstChunk and fileType are also potentially written/read by multiple goroutines?

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1847,6 +1847,7 @@ type cdcWriter struct {
 	firstChunk []byte
 	fileType   rfpb.FileMetadata_FileType
 
+	mu *sync.Mutex // protects writtenChunks, numChunks, firstChunk, fileType
```
vadimberezniker, line 1850: nit: it's more idiomatic to use non-pointer types for mutexes

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1934,9 +1943,14 @@ func (p *PebbleCache) newCDCCommitedWriteCloser(ctx context.Context, fileRecord
 }
 
 func (cdcw *cdcWriter) writeChunk(chunkData []byte) error {
+	cdcw.mu.Lock()
```
vadimberezniker, line 1946: IIUC, there's no real concurrency to worry about since writeChunk is only called by 1 goroutine and Commit will only need to access the mutex protected data when writeChunk is guaranteed to not be called. I think this means you could get away with locking the mutex for the entire duration of writeChunk, i.e.

```
cdcw.my.Lock()
defer cdcw.mu.Unlock()
``` 
at the top of the function

I think you can do something similiar in Commit after `cdcw.eg.Wait()`

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1847,6 +1847,7 @@ type cdcWriter struct {
 	firstChunk []byte
 	fileType   rfpb.FileMetadata_FileType
 
+	mu sync.Mutex // protects writtenChunks, numChunks, firstChunk, fileType
```
vadimberezniker, line 1850: This might be a personal preference, but I like to organize struct fields such that protected variables are below the mutex that protects them, i.e.

```
   eg *errgroup.Group

    mu sync.Mutex
	writtenChunks []*rspb.ResourceName
	numChunks  int
	firstChunk []byte
	fileType   rfpb.FileMetadata_FileType
```

Though you have a comment which accomplishes the same goal.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
tylerwilliams: +1

tylerwilliams (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: Done

luluz66 (commented):
I removed the context cancellation in the chunker, which can make things a lot simpler. 

Otherwise, there can be a scenario that causes race conditions on the size updates channel ([example test failure](https://app.buildbuddy.dev/invocation/65444f9b-37e0-42ab-9003-265d04dbf1db?target=%2F%2Fenterprise%2Fserver%2Fbackends%2Fpebble_cache%3Apebble_cache_test&targetStatus=6#4@52))

Vadim came up with a scenario where there are some go routines in the wild after Commit() returns. 
- `Chunker.Close` is Called. 
- At the same time, Chunker is about to call `writeChunkFn`
- `cdcw.eg.Wait()` is called but it returns immediately because there is no writes are in progress
- writeChunkFn spawns a new goroutine.

enterprise/server/util/chunker/chunker.go
```diff
@@ -86,5 +81,10 @@ func New(ctx context.Context, averageSize int, writeChunkFn WriteFunc) (*Chunker
 		}
 	}()
 
+	go func() {
+		<-ctx.Done()
+		pw.Close()
```
vadimberezniker, line 86: You should call CloseWithErr to be safe to avoid a scenario where the reader thinks the pipe was closed properly (i.e. something like `pw.CloseWithErr(context.Canceled)`

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2026,8 +2032,24 @@ func (cdcw *cdcWriter) Write(buf []byte) (int, error) {
...
+	}
+	if err := cdcw.eg.Wait(); err != nil {
+		return err
+	}
+	return closeErr
+}
+
 func (cdcw *cdcWriter) Close() error {
-	defer cdcw.chunker.Close()
+	if cdcw.isChunkerClosed {
```
vadimberezniker, line 2049: should this condition be inverted?

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: oops. Done.

PR #4789, opened 2023-09-15
Roll-forward "Upgrade bb toolchain to get `dockerNetwork=off`"

bduffany:
Added `test.dockerNetwork=bridge` to the webdriver test to fix it. Also change `dockerNetwork` => `test.dockerNetwork` so that only the test actions get networking, not the actions that build the tests.

**Related issues**: N/A


tempoz (approved):
LGTM!

PR #4793, opened 2023-09-15
Run NBD clients in dedicated processes

bduffany:
This will prevent other goroutines in the init / vmexec binaries from deadlocking with the NBD handler (due to the binary + handler contending for the same OS thread resource). This should reduce the occurrence of these occasional "possible stuck request" bugs.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
+++ b/enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
@@ -1,15 +1,18 @@
 package nbdclient
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io"
 	"net"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/Merovius/nbd"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/vsock"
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
```diff
@@ -27,8 +30,184 @@ const (
...
+	log.CtxInfof(ctx, "nbdclient child process started")
+	// fds 3 and 4 are the files passed to us via ExtraFiles.
+	in := os.NewFile(3, "in")
+	out := os.NewFile(4, "out")
+	scanner := bufio.NewScanner(in)
+
+	var device *ClientDevice
+	var mountPath string
+
+	for scanner.Scan() {
```
maggie-lou, line 151: I might be misunderstanding how this works, but could there be a race condition if we haven't sent commands to this background process fast enough? i.e. If the input FD hasn't received any commands yet, then scanner.Scan would return immediately

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
```diff
@@ -27,8 +30,184 @@ const (
...
+	// Close input reader and output writer handles; these are only used by the
+	// child.
+	inR.Close()
+	outW.Close()
+
+	scanner := bufio.NewScanner(outR)
+	h := &Process{name: name, input: inW, output: scanner}
+	go func() {
+		err := cmd.Wait()
+		h.err = status.InternalErrorf("child exited: %s", err)
```
maggie-lou, line 93: Do we need to check for nil err here?

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
bduffany: Linux pipes work similarly to buffered channels in Go - the `read()` syscall that underlies `scanner.Scan()` will block until there is data available in the pipe buffer, and `write()` on the other end of the pipe will either add to the pipe buffer, or if the buffer is full, it waits until there's available capacity.

enterprise/server/remote_execution/nbd/nbdclient/nbdclient.go
bduffany: The nbdclient process should stay running forever so we want to treat any exit here as an error (added a clarifying comment)

maggie-lou (approved):

PR #4795, opened 2023-09-15
Install amazon-ecr-credential-helper in executor

raylu:
Since `docker-credential-gcr` is installed, I figure installing the cred helper for AWS ECR would be good too.

I tried to test this but I wasn't able to pull from gcr.io/cloud-marketplace. Even after changing to docker:11 though, I ran into an error installing docker-ce.


sluongng:
Thanks for creating this change.
This LGTM, the additional apt install clean up is much appreciated.

@bduffany wdyt?

bduffany (approved):
Looks good, thanks!

PR #4797, opened 2023-09-16
Stick to even-numbered local hour steps even when dst goes by

jdhollen:
**Related issues**: N/A


tylerwilliams (approved):
good comment :)

PR #4799, opened 2023-09-18
Swap order of podman and firecracker in default isolation type list

bduffany:
When the `default_isolation_type` flag is unspecified, it currently defaults to the first of `docker, firecracker, podman, none`.

This PR changes the ordering to be `docker, podman, firecracker, none`, swapping podman and firecracker. This way, we don't have to explicitly set `default_isolation_type=podman` when both podman and firecracker are enabled. It also intuitively makes more sense this way since docker and podman are the same "class" of isolation (OCI container).

I hit this while trying to run a local executor within a GCP VM while trying to reproduce an issue, and had both podman and firecracker enabled. I was surprised to see that build actions were trying to use firecracker by default.

**Related issues**: N/A


sluongng (approved):
LGTM

PR #4802, opened 2023-09-18
Make edit_instance_template use the same go.mod as everything else

siggisim:
This way we don't need to create / maintain a go.work file.

Have a separate PR to update instruction in the internal playbooks.


tylerwilliams (approved):

vadimberezniker (approved):
For context, the reason it was this way originally because there was some proto version conflict.

PR #4803, opened 2023-09-18
Rename `install_buildbuddy_dependencies` => `install_go_mod_dependencies`

siggisim:
This change:
 1) Allows us to have different functions for different language ecosystems (i.e. install_package_json_dependencies, etc)
2) Makes it easier for automated tooling to generate this function name (i.e `install_%s_dependencies` where %s is the name of the file where the dependencies came from). 

I have a separate change do to the same in the internal repo.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -42,12 +42,12 @@ http_archive(
     ],
 )
 
-load(":deps.bzl", "install_buildbuddy_dependencies")
+load(":deps.bzl", "install_go_mod_dependencies")
 
 # Install gazelle and go_rules dependencies after ours so that our go module versions take precedence.
 
-# gazelle:repository_macro deps.bzl%install_buildbuddy_dependencies
-install_buildbuddy_dependencies()
+# gazelle:repository_macro deps.bzl%install_go_mod_dependencies
+install_go_mod_dependencies()
 
 load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
 load("@io_bazel_rules_go//go:deps.bzl", "go_download_sdk", "go_register_toolchains", "go_rules_dependencies")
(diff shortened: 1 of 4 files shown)


deps.bzl
```diff
@@ -1,8 +1,8 @@
 load("@bazel_gazelle//:deps.bzl", "go_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
 
-# bazelisk run //:gazelle -- update-repos -from_file=go.mod -to_macro=deps.bzl%install_buildbuddy_dependencies
-def install_buildbuddy_dependencies(workspace_name = "buildbuddy"):
+# bazelisk run //:gazelle -- update-repos -from_file=go.mod -to_macro=deps.bzl%install_go_mod_dependencies
+def install_go_mod_dependencies(workspace_name = "buildbuddy"):
```
tylerwilliams, line 5: I think there are some http_archives and http_files that we might wanna put in a go_static_dependencies target or something if this is gonna be go_mod only

WORKSPACE
```diff
@@ -193,6 +193,9 @@ http_archive(
 
 load("@com_google_googletest//:googletest_deps.bzl", "googletest_deps")
 
+# gazelle:repository_macro deps.bzl%install_go_mod_dependencies
+install_go_mod_dependencies()
```
vadimberezniker, line 197: Why is this being called twice?

vadimberezniker (changes requested):

deps.bzl
siggisim: Pulled those out into their own function - there's no harm in having them together, but this makes it clearer which ones are generated and which ones are manual.

WORKSPACE
siggisim: Removed

tylerwilliams (approved):

vadimberezniker (approved):

PR #4808, opened 2023-09-18
Generate a `WORKSPACE` file when running `bb fix` if one isn't present

siggisim:
**Related issues**: N/A

### DIFF
--- a/cli/fix/fix.go
+++ b/cli/fix/fix.go
@@ -1,17 +1,26 @@
 package fix
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/buildbuddy-io/buildbuddy/cli/add"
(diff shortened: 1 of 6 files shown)


cli/fix/fix.go
```diff
@@ -24,25 +33,39 @@ func HandleFix(args []string) (exitCode int, err error) {
...
 	}
 
+	_, _, err = workspace.CreateWorkspaceFileIfNotExists()
+	if err != nil {
+		return 1, err
+	}
+
+	err = walk()
+	if err != nil {
+		log.Printf("Error fixing: %+v", err)
```
tylerwilliams, line 43: %s

tylerwilliams (approved):

cli/fix/fix.go
siggisim: Done

PR #4809, opened 2023-09-18
[Remote Snapshot] Refactor snaploader to represent snapshot metadata using an ActionResult

maggie-lou:
For remote snapshot sharing, we will need to represent snapshot metadata as an ActionResult in order to save it to the action cache.

When fetching action results, our existing architecture will validate that all referenced digests have not been evicted from the cache. We can take advantage of this to validate that our snapshots are valid and all necessary artifacts are still in the cache

Also remove the DeleteSnapshot function. When we are storing snapshot artifacts remotely, we can rely on the remote cache to evict LRU elements. Snapshot chunks can be reused by other snapshots, so we won't want to delete all associated artifacts

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -1,7 +1,6 @@
 package snaploader
 
 import (
-	"bufio"
 	"context"
 	"flag"
 	"fmt"
@@ -80,13 +79,7 @@ func fileDigest(filePath string) (*repb.Digest, error) {
 		return nil, err
 	}
 	defer file.Close()
-
-	fileReader := bufio.NewReader(file)
-	d, err := digest.Compute(fileReader, repb.DigestFunction_BLAKE3)
-	if err != nil {
-		return nil, err
-	}
-	return d, nil
+	return digest.Compute(file, repb.DigestFunction_BLAKE3)
 }
 
 // Snapshot holds a snapshot manifest along with the corresponding cache key.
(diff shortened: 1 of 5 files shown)


maggie-lou:
* master
  * **PR #4805** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4805" target="_blank">Graphite</a> 
    * **PR #4809** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4809" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4809?utm_source=stack-comment).

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -219,6 +204,94 @@ func (l *FileCacheLoader) GetSnapshot(ctx context.Context, key *fcpb.SnapshotKey
...
+	if err != nil {
+		return nil, status.UnavailableErrorf("failed to read chunked file tree: %s", status.Message(err))
+	}
+	tree := &repb.Tree{}
+	if err := proto.Unmarshal(buf, tree); err != nil {
+		return nil, status.WrapError(err, "unmarshall chunked file tree")
+	}
+	return tree, nil
+}
+
```
bduffany, line 294: I think it might be good to have this code eventually if we want to share some file-downloading code w/ `dirtools.go`, but thinking about it a bit more, since the snapshotting stuff is very specialized (e.g. we lazily download everything, we have this special "chunk_size" prop...) I wonder if realistically we'll get around to that any time soon.

wdyt about keeping this code in our back pocket until we have a more solid plan for consolidating this with `dirtools.go`, and maybe just serializing the manifest proto to a single `[]byte` field in the ActionResult for now (e.g. `stdout_raw`), just to keep things simple?

bduffany (approved):
LGTM modulo nit

PR #4810, opened 2023-09-18
add a flag in migration cache to clear pebble cache on start up

luluz66:
**Related issues**: N/A
Need to delete the cdc cache to restart migration in prod/dev

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -177,7 +177,7 @@ type Options struct {
 
 	Clock clockwork.Clock
 
-	ClearCacheBeforeMigration bool
+	ClearCacheOnStartup bool
 }
 
 type sizeUpdate struct {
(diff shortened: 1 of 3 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -176,6 +176,8 @@ type Options struct {
 	IncludeMetadataSize bool
 
 	Clock clockwork.Clock
+
+	ClearCacheBeforeMigration bool
```
vadimberezniker, line 180: You can omit 'BeforeMigration' from the option name, it doesn't need to be tied to migrations.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: i changed to clear_cache_on start_up

PR #4826, opened 2023-09-19
[Remote Snapshot] Add support for remote snapshot sharing

maggie-lou:
If remote snapshot sharing is enabled for firecracker VMs, the snapshot loader should try to fetch artifacts from the local file cache first, and fall back to the remote cache if they're missing locally. 

If remote snapshot sharing is enabled, the remote cache should serve as the source of truth for the snapshot manifest that VMs should use. Manifests will no longer be stored locally. If any artifact referenced by the manifest has been evicted from the cache, there will be no snapshots available, and runners will have to start a cold VM


maggie-lou:
* master
  * **PR #4826** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4826" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4826?utm_source=stack-comment).

enterprise/server/remote_execution/snaploader/snaploader_test.go
```diff
@@ -81,67 +230,68 @@ func TestPackAndUnpackChunkedFiles(t *testing.T) {
...
+			if i == 0 {
+				rn := digest.NewResourceName(c.GetDigest(), "", rspb.CacheType_CAS, repb.DigestFunction_BLAKE3).ToProto()
+				err = env.GetCache().Delete(ctx, rn)
+				require.NoError(t, err)
+				break
+			}
+		}
 	}
+
+	// Even though the assets still exist in the local cache, the remote cache
```
maggie-lou, line 291: Is this the desired behavior? My understanding coming out of our meeting was that we wanted one source of truth for what snapshot VMs should be using

enterprise/server/remote_execution/snaploader/snaploader_test.go
```diff
@@ -81,67 +231,133 @@ func TestPackAndUnpackChunkedFiles(t *testing.T) {
...
+		}
+	}
+	for _, f := range snapMetadata.GetChunkedFiles() {
+		for i, c := range f.GetChunks() {
+			if i == 0 {
+				rn := digest.NewResourceName(c.GetDigest(), "", rspb.CacheType_CAS, repb.DigestFunction_BLAKE3).ToProto()
+				err = env.GetCache().Delete(ctx, rn)
+				require.NoError(t, err)
+				break
+			}
```
bduffany, line 288: since you're breaking on `i == 0` (first iteration) you could remove the `i == 0` check and break unconditionally at the end of the loop. Same for the above loop

enterprise/server/remote_execution/snaploader/snaploader_test.go
bduffany: I think it's fine to keep it like this for now and then see how it performs in dev.

I'm betting that we're going to need some sort of configurable "max age" value after which snapshots have to be re-fetched from the remote cache, like 3 hours or something. I think this is mainly going to be needed for users that don't keep their build graphs super lean, which is probably pretty common. Our build graph for example is relatively heavyweight since we only have an 89% AC hit rate, so I suspect we'll have ~gigabytes of cache churn on every commit pushed to master.

Maybe we should start storing the snapshot creation time somewhere in the snapshot metadata, so that we can see how much of a correlation there is between snapshot age and workflow execution time.

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -52,24 +52,21 @@ func NewKey(task *repb.ExecutionTask, configurationHash, runnerID string) (*fcpb
...
+// manifestDigest returns the digest for the snapshot manifest
 //
-// We always want runners to use the newest manifest (and corresponding
-// snapshot), so they should overwrite any existing manifest when saving
-// snapshots so that newer runners will read from the newer version
-func manifestFileCacheKey(ctx context.Context, env environment.Env, s *fcpb.SnapshotKey) (*repb.FileNode, error) {
+// Because we always want runners to use the newest manifest/snapshot, this
+// doesn't actually create a digest of the manifest contents. It just takes a
+// hash of shared properties, for which the snapshot should be shared
+func manifestDigest(ctx context.Context, env environment.Env, s *fcpb.SnapshotKey) (*repb.Digest, error) {
```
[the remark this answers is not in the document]
maggie-lou: Renamed to `manifestKey` for now

enterprise/server/remote_execution/snaploader/snaploader_test.go
maggie-lou: Storing snapshot age is a good idea. I can look into that when we start to do more performance improvements

enterprise/server/remote_execution/snaputil/snaputil.go
```diff
@@ -0,0 +1,124 @@
...
+
+func GetBytes(ctx context.Context, localCache interfaces.FileCache, bsClient bytestream.ByteStreamClient, d *repb.Digest, instanceName string, tmpDir string) ([]byte, error) {
+	randStr, err := random.RandomString(10)
+	if err != nil {
+		return nil, err
+	}
+	tmpPath := filepath.Join(tmpDir, fmt.Sprintf("%s.%s.tmp", d.Hash, randStr))
+	defer func() {
+		if err := os.Remove(tmpPath); err != nil {
+			log.Warningf("Failed to remove temp file in snaputil::GetBytes: %s", err)
```
bduffany, line 63:
```suggestion
			log.CtxWarningf(ctx, "Failed to remove temp file in snaputil::GetBytes: %s", err)
```

enterprise/server/remote_execution/snaputil/snaputil.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return err
+	}
+	tmpPath := filepath.Join(localCache.TempDir(), fmt.Sprintf("%s.%s.tmp", d.Hash, randStr))
+	if err := os.WriteFile(tmpPath, b, 0777); err != nil {
+		return err
+	}
+	defer func() {
+		if err := os.Remove(tmpPath); err != nil {
+			log.Warningf("Failed to remove temp file: %s", err)
```
bduffany, line 107:
```suggestion
			log.CtxWarningf(ctx, "Failed to remove temp file: %s", err)
```

bduffany (approved):

PR #4827, opened 2023-09-19
Adding more context to returned error in chunker and pebble cache

luluz66:
**Related issues**: N/A


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1984,13 +1984,13 @@ func (cdcw *cdcWriter) writeRawChunk(fileRecord *rfpb.FileRecord, key filestore.
 	}
 	_, err = wcm.Write(chunkData)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to write raw chunk: %s", err)
```
vadimberezniker, line 1987: any reason not to use status.Blah() everywhere?

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go

PR #4828, opened 2023-09-20
commit then close

luluz66:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1986,11 +1986,11 @@ func (cdcw *cdcWriter) writeRawChunk(fileRecord *rfpb.FileRecord, key filestore.
 	if err != nil {
 		return status.InternalErrorf("failed to write raw chunk: %s", err)
 	}
-	if err := wcm.Close(); err != nil {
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1986,12 +1986,12 @@ func (cdcw *cdcWriter) writeRawChunk(fileRecord *rfpb.FileRecord, key filestore.
 	if err != nil {
 		return status.InternalErrorf("failed to write raw chunk: %s", err)
 	}
-	if err := wcm.Close(); err != nil {
-		return status.InternalErrorf("failed to close while writing raw chunk: %s", err)
-	}
 	if err := wcm.Commit(); err != nil {
 		return status.InternalErrorf("failed to commit while writing raw chunk: %s", err)
 	}
+	if err := wcm.Close(); err != nil {
```
vadimberezniker, line 1992: Can you put the Close in a defer so that it's called even if the write or commit fail?

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: Done

vadimberezniker (approved):

PR #4830, opened 2023-09-20
auth: log auth errors

sluongng:
Right now we are only returning the first error for all the
authenticator errors. However, since we always configure OIDC
authenticator, only OIDC authenticator error is returned.

This is fine from user perspective, but from the sysadmin / operation
perspective, we have no way to verify whether the issue is coming from
SAML or Github authenticator to troubleshoot.

Log out the errors to help aid troubleshooting.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/auth/auth.go
+++ b/enterprise/server/auth/auth.go
@@ -94,7 +94,7 @@ func (a *authenticator) Login(w http.ResponseWriter, r *http.Request) error {
 		}
 	}
 	if len(errors) > 0 {
-		log.Errorf("login failed: %s", errors)
+		log.Infof("login failed: %s", errors)
 		return errors[0]
 	}
 	return status.NotFoundErrorf("No authenticator registered to handle login path: %s", r.URL.Path)
@@ -115,7 +115,7 @@ func (a *authenticator) Auth(w http.ResponseWriter, r *http.Request) error {
 		}
 	}
 	if len(errors) > 0 {
-		log.Errorf("auth failed: %s", errors)
+		log.Infof("auth failed: %s", errors)
 		return errors[0]
 	}
 	return status.NotFoundErrorf("No authenticator registered to auth path: %s", r.URL.Path)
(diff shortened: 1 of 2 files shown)


bduffany (changes requested):
we should log these at INFO level to prevent failing the canary step during the next deploy

enterprise/server/auth/auth.go
```diff
@@ -93,6 +94,7 @@ func (a *authenticator) Login(w http.ResponseWriter, r *http.Request) error {
 		}
 	}
 	if len(errors) > 0 {
+		log.Infof("login failed: %s", errors)
```
bduffany, line 97:
```suggestion
		log.Infof("Failed login attempt: %s", errors)
```

enterprise/server/auth/auth.go
```diff
@@ -113,6 +115,7 @@ func (a *authenticator) Auth(w http.ResponseWriter, r *http.Request) error {
 		}
 	}
 	if len(errors) > 0 {
+		log.Infof("auth failed: %s", errors)
```
bduffany, line 118:
```suggestion
		log.Infof("Failed to authorize HTTP request: %s", errors)
```

enterprise/server/auth/auth.go
```diff
@@ -146,6 +149,7 @@ func (a *authenticator) FillUser(ctx context.Context, user *tables.User) error {
 		}
 	}
 	if len(errors) > 0 {
+		log.Infof("fill user failed: %s", errors)
```
bduffany, line 152:
```suggestion
		log.Infof("Failed to fill authenticated user fields: %s", errors)
```

enterprise/server/auth/auth.go
```diff
@@ -164,6 +168,7 @@ func (a *authenticator) AuthenticatedUser(ctx context.Context) (interfaces.UserI
 		}
 	}
 	if len(errors) > 0 {
+		log.Infof("authenticated user failed: %s", errors)
```
bduffany, line 171:
```suggestion
		log.Infof("Failed to authenticate user: %s", errors)
```

bduffany (approved):

PR #4836, opened 2023-09-20
Make impersonation play better with subdomains.

vadimberezniker:
Use a cookie instead of session storage to store the impersonated group ID so that the value can be set across subdomains.

**Related issues**: N/A


tylerwilliams (approved):
lgtm but please get @bduffany's review too

bduffany (approved):

PR #4837, opened 2023-09-20
Add an option to rate limit writes in migration cache

luluz66:
**Related issues**: N/A
In early stages of migration while we are backfilling from source cache, we need
the ability to rate limit writes to destination cache to avoid overwhelming the
destination cache which causes Write latency goes up and compaction debt goes up
for example.

Also in this PR, I used mc.Writer() to implement Set() method to make the logic
simpler and also allow us to write data in background in the Set() method.

### DIFF
--- a/enterprise/server/backends/migration_cache/config.go
+++ b/enterprise/server/backends/migration_cache/config.go
@@ -21,9 +21,8 @@ type MigrationConfig struct {
 	// CopyChanFullWarningIntervalMin controls how often we should log when the copy chan is full
 	CopyChanFullWarningIntervalMin int64 `yaml:"copy_chan_full_warning_interval_min"`
 	MaxCopiesPerSec                int   `yaml:"max_copies_per_sec"`
-	// Whether we rate limit Write(). If so, we will write to destination cache
-	// in the background and use MaxCopiesPerSec to control the speed.
-	RateLimitWriteToDest bool `yaml:"rate_limit_writes"`
+	// AsyncDestWrites controls whether we write to destination cache in the background
+	AsyncDestWrites bool `yaml:"async_dest_writes"`
 }
 
 type CacheConfig struct {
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -752,6 +754,12 @@ func (mc *MigrationCache) Writer(ctx context.Context, r *rspb.ResourceName) (int
 		return nil, err
 	}
 
+	if mc.rateLimitWriteToDest {
+		// We will write to the destination cache in the background
+		mc.sendNonBlockingCopy(ctx, r, false /*=onlyCopyMissing*/)
```
vadimberezniker, line 759: onlyCopyMissing should be true to avoid enqueing copies for digests that already exist

enterprise/server/backends/migration_cache/config.go
```diff
@@ -21,6 +21,9 @@ type MigrationConfig struct {
 	// CopyChanFullWarningIntervalMin controls how often we should log when the copy chan is full
 	CopyChanFullWarningIntervalMin int64 `yaml:"copy_chan_full_warning_interval_min"`
 	MaxCopiesPerSec                int   `yaml:"max_copies_per_sec"`
+	// Whether we rate limit Write(). If so, we will write to destination cache
+	// in the background and use MaxCopiesPerSec to control the speed.
+	RateLimitWriteToDest bool `yaml:"rate_limit_writes"`
```
vadimberezniker, line 26: nit: i'd call this something like `async_dest_writes` since the rate limiting is only one of the effects of the change.

vadimberezniker (approved):

enterprise/server/backends/migration_cache/migration_cache.go
luluz66: Talked offline. Somewhere in the system checks whether the digest exists in the system or not prior to Write() is called. So we don't need to set onlyCopyMissing = true.

PR #4839, opened 2023-09-20
Log some more stuff about https://github.com/buildbuddy-io/buildbuddy-internal/issues/2673

iain-macdonald:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2673

### DIFF
--- a/enterprise/server/sociartifactstore/sociartifactstore.go
+++ b/enterprise/server/sociartifactstore/sociartifactstore.go
@@ -233,6 +233,7 @@ func getTargetImageInfo(ctx context.Context, image string, platform *rgpb.Platfo
 }
 
 func (s *SociArtifactStore) GetArtifacts(ctx context.Context, req *socipb.GetArtifactsRequest) (*socipb.GetArtifactsResponse, error) {
+	ctx = log.EnrichContext(ctx, "group_id", getGroupIdForDebugging(ctx, s.env))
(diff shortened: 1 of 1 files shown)


enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -327,12 +328,17 @@ func (s *SociArtifactStore) getArtifactsFromCache(ctx context.Context, imageConf
 }
 
 func (s *SociArtifactStore) pullAndIndexImage(ctx context.Context, imageRef ctrname.Digest, configHash v1.Hash, credentials *rgpb.Credentials) (*repb.Digest, []*repb.Digest, error) {
-	log.Infof("soci artifacts not found, generating them for image %s", imageRef.DigestStr())
+	groupId := getGroupIdForDebugging(ctx, s.env)
+	log.Infof("soci artifacts not found, generating them for group %s / image %s", groupId, imageRef.DigestStr())
```
tylerwilliams, line 332: this is fine, just for future reference another way to handle this is to call

ctx = log.EnrichContext(ctx, "group_id", getGroupId())

somewhere high up in the chain of processing here when you have groupID, and then these log statements can become:

log.CtxInfo("soci artifacts not found, generating for image: %s", imageStr)

and the log will include the groupID you attached above.

tylerwilliams (approved):

PR #4851, opened 2023-09-21
Close pr on cancel in chunker

luluz66:
**Related issues**: N/A
Close pr instead of pw in the chunker on context cancellation.

pw.Write() returns "io: read/write on closed pipe" if pw is already closed.
However, if we close the pr with ctx.Err(), then pw.Write() will return the
given error. 

We also need to let Chunker.Close() return an error if the context
is cancelled, so that Commit() in pebble cache will return early and not writing
0 chunk to pebble.

### DIFF
--- a/enterprise/server/util/chunker/chunker.go
+++ b/enterprise/server/util/chunker/chunker.go
@@ -3,6 +3,7 @@ package chunker
 import (
 	"context"
 	"io"
+	"sync"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/jotfs/fastcdc-go"
@@ -17,6 +18,9 @@ type Chunker struct {
 
 	done chan struct{}
 
+	mu  sync.Mutex // protexts err
+	err error
+
 	writeChunkFn WriteFunc
 }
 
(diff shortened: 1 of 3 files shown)


enterprise/server/util/chunker/chunker.go
```diff
@@ -24,13 +24,22 @@ func (c *Chunker) Write(buf []byte) (int, error) {
 	return c.pw.Write(buf)
 }
 
-func (c *Chunker) Close() error {
+func (c *Chunker) Close() (bool, error) {
+	var ctxErr error
```
vadimberezniker, line 28: I think this is still racy because the context may be cancelled after you check the context error but before you return. To avoid the race, you'd need to use a (mutex protected) variable to indicate whether the chunker has been closed already

vadimberezniker (changes requested):

enterprise/server/util/chunker/chunker.go
```diff
@@ -24,13 +24,22 @@ func (c *Chunker) Write(buf []byte) (int, error) {
...
+	select {
+	case <-c.ctx.Done():
+		ctxErr = c.ctx.Err()
+	default:
+		break
+	}
+
 	if err := c.pw.Close(); err != nil {
-		return status.InternalErrorf("failed to close chunker: %s", err)
+		return false, status.InternalErrorf("failed to close chunker: %s", err)
```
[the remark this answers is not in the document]
luluz66: Done. 
Though I feel a bit wierd in closeChunkerAndWait() to set `cdcw.isChunkerClosed = true` without checking the returned error.

enterprise/server/util/chunker/chunker.go
luluz66: Done. I actually used a variable to save for the error encountered. I noticed that in the chunker's for loop, when we encounter error we close the pr with that error. If we don't call chunker.Write(..) that error won't be surfaced. And if there are errors encountered in the for loop, we don't want to proceed with Commit(..)

enterprise/server/util/chunker/chunker.go
```diff
@@ -72,19 +79,36 @@ func New(ctx context.Context, averageSize int, writeChunkFn WriteFunc) (*Chunker
...
+				}
 				return
 			}
 		}
 	}()
 
 	go func() {
 		<-ctx.Done()
-		pw.CloseWithError(ctx.Err())
+		pr.CloseWithError(ctx.Err())
```
vadimberezniker, line 106: move `pr.CloseWithError(ctx.Err())` below the Lock()

otherwise it's possible that the other goroutine will run first and update c.err before this one has a chance to

vadimberezniker (approved):

PR #4859, opened 2023-09-22
[Remote Snapshot] Make manual test to print performance of different snaploader configurations

maggie-lou:
Results:

(remote_snapshot_sharing=true) Pausing a new VM took 9.713178852s.
(remote_snapshot_sharing=false) Pausing a new VM took 806.808369ms.

(remote_snapshot_sharing=true) Pausing a VM that had started from a snapshot and had no non-metadata changes took 11.158405497s.
(remote_snapshot_sharing=false) Pausing a VM that had started from a snapshot and had no non-metadata changes took 877.628348ms.

(remote_snapshot_sharing=true) Pausing a VM that had started from a snapshot and had execution-related changes took 13.230709017s.
(remote_snapshot_sharing=false) Pausing a VM that had started from a snapshot and had execution-related changes took 884.535171ms.

(remote_snapshot_sharing=true) Unpausing a VM that is fully cached in filecache took 358.3608ms.
(remote_snapshot_sharing=false) Unpausing a VM that is fully cached in filecache took 327.293934ms.

(remote_snapshot_sharing=true) Unpausing a VM where ~30% of artifacts were evicted from filecache and must be fetched remotely. Took 667.515344ms.

(remote_snapshot_sharing=true) Unpausing a VM where all artifacts were evicted from filecache and must be fetched remotely. Took 1.882710614s.


maggie-lou:
* master
  * **PR #4826** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4826" target="_blank">Graphite</a> 
    * **PR #4859** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4859" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4859?utm_source=stack-comment).

enterprise/server/remote_execution/containers/firecracker/BUILD
```diff
@@ -159,6 +161,59 @@ go_test(
...
+        "//server/util/status",
+        "//server/util/testing/flags",
+        "@com_github_stretchr_testify//assert",
+        "@com_github_stretchr_testify//require",
+        "@org_golang_google_genproto_googleapis_bytestream//:bytestream",
+        "@org_golang_x_sync//errgroup",
+    ],
+)
+
+go_test(
```
bduffany, line 183: it's kind of a pain having these separate test targets because Gazelle only manages the `firecracker_test` deps for us and we have to manage the other ones manually. I'm hoping to remove the `_blockio` target as soon as we launch the copy-on-write stuff.

Ideally we could split this into a separate test file like `firecracker_benchmark_test.go` and I think that should be enough for gazelle to manage it for us, not 100% sure though. (getTestEnv etc. could be moved into a separate testutil package)

Another option is to rename the test case to `BenchmarkFirecrackerSnapshotting` and then we can run it with `--test_filter=Benchmark --test_arg=-test.bench=. --test_arg=-test.benchtime=1x` - the downside though is that we can't run it automatically as part of our Benchmark action on CI

maggie-lou: Good point. To simplify, I just removed conditionally running this test in our regular test suite.

bduffany: This test seems pretty heavyweight (takes > 1min to run from looking at the logs) - I think we should keep it manually run for now

maggie-lou: If I split it into a separate test file, gazelle tries to add it as an additional `src` to the `firecracker_test` target, which won't let us set custom tags on it. To simplify, I just added a new flag to control whether this is run, and we can manually set it.

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -564,6 +550,356 @@ func TestFirecracker_LocalSnapshotSharing(t *testing.T) {
...
+		containersToCleanup = append(containersToCleanup, c)
+		err = container.PullImageIfNecessary(ctx, env, c, container.PullCredentials{}, opts.ContainerImage)
+		require.NoError(t, err)
+		err = c.Create(ctx, opts.ActionWorkingDirectory)
+		require.NoError(t, err)
+	}
+
+	// Pausing a new VM
+	for _, enableRemote := range []bool{true, false} {
+		{
```
bduffany, line 762: nit: the extra `{` within the loop is unnecessary, since for-loops create their own scope already (here and below)

bduffany (approved):
LGTM, looks like the PR just needs a rebase

PR #4860, opened 2023-09-22
Add an option in migration config to allow for multiple workers to copy in parallel

luluz66:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/migration_cache/migration_cache.go
+++ b/enterprise/server/backends/migration_cache/migration_cache.go
@@ -43,6 +43,7 @@ type MigrationCache struct {
 	quitChan chan struct{}
 
 	maxCopiesPerSec             int
+	numCopyWorkers              int
 	copyChan                    chan *copyData
 	copyChanFullWarningInterval time.Duration
 	numCopiesDropped            *int64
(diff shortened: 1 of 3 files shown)


enterprise/server/backends/migration_cache/migration_cache.go
```diff
@@ -889,27 +891,31 @@ type copyData struct {
...
+				if err := rateLimiter.Wait(context.Background()); err != nil {
+					return err
+				}
+				select {
+				case <-mc.quitChan:
+					// Drain copy channel on shutdown
+					// We cannot close the channel before draining because there's a chance some cache requests may be
+					// concurrently queued as the cache is shutting down, and if we try to enqueue a copy to the closed
+					// channel it will panic
+					for len(mc.copyChan) > 0 {
```
vadimberezniker, line 907: This for loop won't work properly if there are multiple goroutines. You can move it out of the worker code.

vadimberezniker (changes requested):

enterprise/server/backends/migration_cache/migration_cache.go
luluz66: Done.

vadimberezniker (approved):

PR #4864, opened 2023-09-25
tsconfig: update rootdirs for target-linux

sluongng:
In #4769, we make it so that `--config=target-linux` would store outputs
in a new dir with name includes the platform information.

Include this new dir schema in tsconfig.

**Related issues**: N/A

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -19,6 +19,8 @@
       "bazel-bin",
       "bazel-out/k8-opt/bin",
       "bazel-out/k8-fastbuild/bin",
+      "bazel-out/platform_linux-fastbuild/bin",
+      "bazel-out/platform_linux-opt/bin",
       "bazel-out/darwin-opt/bin",
       "bazel-out/darwin-fastbuild/bin",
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -19,6 +19,7 @@
       "bazel-bin",
       "bazel-out/k8-opt/bin",
       "bazel-out/k8-fastbuild/bin",
+      "bazel-out/platform_linux-fastbuild/bin",
```
bduffany, line 22: can you double-check that this works with `-c opt`? you might need to add this:

```suggestion
      "bazel-out/platform_linux-opt/bin",
      "bazel-out/platform_linux-fastbuild/bin",
```

bduffany (approved):

tsconfig.json
sluongng: Good point. I pushed a fix.

sluongng:
Dev builds are disconnected and docker_test seems flaky. Both are unrelated to this PR.

Merging.

PR #4874, opened 2023-09-25
Automatically include client identity in intra-server RPCs.

vadimberezniker:
grpc_client.Dial is a new function that takes an env and automatically adds a client identity interceptor, if necessary.

grpc_client.DialSimple is the former grpc_client.DialTarget func that can be used in non-server code.

**Related issues**: N/A

### DIFF
--- a/server/util/grpc_client/grpc_client.go
+++ b/server/util/grpc_client/grpc_client.go
@@ -6,16 +6,28 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/rpc/interceptors"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/google"
 	"google.golang.org/grpc/keepalive"
 )
 
-// DialTarget handles some of the logic around detecting the correct GRPC
+// DialSimple handles some of the logic around detecting the correct GRPC
 // connection type and applying relevant options when connecting.
-func DialTarget(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
-	dialOptions := CommonGRPCClientOptions()
+//
+// NOTE: Clients within BuildBuddy servers (i.e. executor) should use the Dial
+// variant with an env argument.
+func DialSimple(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
+	return Dial(nil, target, extraOptions...)
+}
+
+// Dial is similar to DialSimple, but it adds additional interceptors
+// (such as client identity) based on the specified environment.
+//
+// Outside of BuildBuddy servers, DialSimple may be used instead.
+func Dial(env environment.Env, target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
+	dialOptions := CommonGRPCClientOptions(env)
 	dialOptions = append(dialOptions, extraOptions...)
 	u, err := url.Parse(target)
 	if err == nil {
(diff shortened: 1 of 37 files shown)


server/util/grpc_client/grpc_client.go
```diff
@@ -6,16 +6,28 @@ import (
...
+// NOTE: Clients within BuildBuddy servers (i.e. executor) should use the Dial
+// variant with an env argument.
+func DialSimple(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
+	return Dial(nil, target, extraOptions...)
+}
+
+// Dial is similar to DialSimple, but it adds additional interceptors
+// (such as client identity) based on the specified environment.
+//
+// Outside of BuildBuddy servers, DialSimple may be used instead.
```
bduffany, line 28: I wonder if the "Simple" vs. "non-Simple" naming will be confusing? wdyt about

```
DialSimple, Dial -> Dial, DialInternal
```

vadimberezniker: I renamed Dial to DialInterservice, how does that sound?

bduffany (approved):

server/util/grpc_client/grpc_client.go
```diff
@@ -6,16 +6,28 @@ import (
...
+// variant with an env argument.
+func DialSimple(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
+	return DialInterservice(nil, target, extraOptions...)
+}
+
+// DialInterservice is similar to DialSimple, but it adds additional interceptors
+// (such as client identity) based on the specified environment.
+//
+// Outside of BuildBuddy servers, DialSimple may be used instead.
+func DialInterservice(env environment.Env, target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
```
tylerwilliams, line 29: nit: can you keep comments to < 80 chars and add some detail here about cases where each of these is appropriate? Want to make it crystal clear which one should be used for new cases (can we replace 'may' with 'should' and clarify?)

server/util/grpc_client/grpc_client.go
```diff
@@ -6,16 +6,28 @@ import (
...
-// DialTarget handles some of the logic around detecting the correct GRPC
+// DialSimple handles some of the logic around detecting the correct GRPC
 // connection type and applying relevant options when connecting.
-func DialTarget(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
-	dialOptions := CommonGRPCClientOptions()
+//
+// NOTE: Clients within BuildBuddy servers (i.e. executor) should use the DialInterservice
+// variant with an env argument.
+func DialSimple(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
+	return DialInterservice(nil, target, extraOptions...)
```
tylerwilliams, line 22: passing nil as env (and then passing it all the way down to getClientIdentityAdder()) feels a little bit error prone to me. If the major difference between these two is whether or not the client identity interceptor is installed -- could we instead go back to having a

dialOptions := CommonGRPCClientOptions()

and then appending one more interceptor to it if env is set?

server/util/grpc_client/grpc_client.go
```diff
@@ -6,16 +6,28 @@ import (
...
 
-// DialTarget handles some of the logic around detecting the correct GRPC
+// DialSimple handles some of the logic around detecting the correct GRPC
 // connection type and applying relevant options when connecting.
-func DialTarget(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
-	dialOptions := CommonGRPCClientOptions()
+//
+// NOTE: Clients within BuildBuddy servers (i.e. executor) should use the DialInterservice
+// variant with an env argument.
+func DialSimple(target string, extraOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
```
tylerwilliams, line 21: Is there a strong reason to rename DialTarget to DialSimple?

I think DialTarget is clearer and maybe doesn't need to change -- we could just introduce the DialInterservice and use that where necessary?

That should reduce the number of changes in this CL significantly

tylerwilliams (changes requested):

server/util/grpc_client/grpc_client.go
vadimberezniker: done

server/util/grpc_client/grpc_client.go
vadimberezniker: I wanted people to think twice when dialing to make sure they are using the appropriate function. The change is intentionally disruptive so that people don't keep using DialTarget for everything out of momentum.

server/util/grpc_client/grpc_client.go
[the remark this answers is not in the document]
vadimberezniker: Done, though I still think DialInterservice is more descriptive.

server/util/grpc_client/grpc_client.go
vadimberezniker: done

tylerwilliams (approved):

PR #4876, opened 2023-09-26
Add scripts to test/bench firecracker locally

bduffany:
Testing requires `mount()` privileges once we enable VBD. Add some scripts to make this slightly less painful.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/test.sh
+++ b/enterprise/server/remote_execution/containers/firecracker/test.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+set -e
+
+# Firecracker tests have to be run as root since they use FUSE mounts.
+# This is a convenience script intended for local development, which
+# builds the test as the current user and runs it as root.
+#
+# Bazel args can be provided to this script.
+# Args after "--" are passed to the 'go test' command, e.g. "-test.run=MyTestFilter"
+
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/containers/firecracker/test.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+set -e
+
+# Firecracker tests have to be run as root since they use FUSE mounts.
```
bduffany, line 4: This script probably seems overkill, but the alternatives kind of suck -

* `bazel run <target> --run_under=sudo` didn't seem to work, the test just hangs for some reason
* `bazel test <target> --run_under=sudo` doesn't work because the input is not a terminal and it needs to ask for the root pass. `SUDO_ASKPASS` helper is an option but seems hard to implement in a secure way.
* `sudo bazel` results in weird "permission denied" errors when trying to read source files, also kind of sucks because it uses a totally separate Bazel cache namespace

maggie-lou:
Do you want to add a reference to this script somewhere? Maybe at the top of the firecracker tests or in the README

maggie-lou (approved):

bduffany:
> Do you want to add a reference to this script somewhere? Maybe at the top of the firecracker tests or in the README

Good call, done

PR #4877, opened 2023-09-26
Fix impersonation bar styling

bduffany:
* Enable flex-wrap
* Make it less verbose on mobile and hide the API key button (it's probably impractical to actually use an API key on mobile)

Screenshot (simulating `iPhone 12 Pro` screen dimensions)

image

**Related issues**: N/A

### DIFF
--- a/app/root/root.css
+++ b/app/root/root.css
@@ -1551,7 +1551,7 @@ code .comment {
 
 @media (max-width: 800px) {
   .hide-on-mobile {
-    display: none;
+    display: none !important;
   }
 
   .login-box {
--- a/enterprise/app/root/root.tsx
+++ b/enterprise/app/root/root.tsx
@@ -127,20 +127,27 @@ class ImpersonationComponent extends React.Component<ImpersonationProps, Imperso
   render() {
     return (
       <div className="impersonation-toolbar">
-        <AlertCircle className="icon black" />
-        <span>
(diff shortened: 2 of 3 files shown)


app/root/root.css
```diff
@@ -1550,7 +1550,7 @@ code .comment {
 }
 
 @media (max-width: 800px) {
-  .hide-on-mobile {
+  .hide-on-mobile.hide-on-mobile {
```
jdhollen, line 1553: this is one of the rare situations where i think !important actually makes sense--this way only works until someone makes a three-deep selector that sets display somewhere else.

enterprise/app/root/root.tsx
```diff
@@ -127,20 +127,27 @@ class ImpersonationComponent extends React.Component<ImpersonationProps, Imperso
   render() {
     return (
       <div className="impersonation-toolbar">
-        <AlertCircle className="icon black" />
-        <span>
-          Authenticated as a member of <b>{this.props.user.selectedGroupName()}</b> ({this.props.user.selectedGroup?.id}
-          ). Proceed with caution.
```
jdhollen, line 133: You don't like "proceed with caution"? :)

jdhollen: Oh oh, prefixed it instead.

jdhollen (approved):

app/root/root.css
bduffany: Done

PR #4879, opened 2023-09-26
Add firecracker stress-ng benchmark

bduffany:
**Related issues**: N/A


maggie-lou (approved):

tylerwilliams (approved):

bduffany:
will revisit this if we have more concrete plans for improving disk I/O. For now, these tests are limited in usefulness because they aren't being run automatically.

PR #4882, opened 2023-09-26
IP rule impersonation fixes.

vadimberezniker:
- Don't enforce IP rules for GetGroup API as it's now used for impersonation.

 - AuthorizeGroup should check the impersonation bit, like Authorize does.

**Related issues**: N/A


bduffany (approved):

tylerwilliams (approved):
LGTM

PR #4883, opened 2023-09-26
Fix disk size calculation in `DirectoryToImageAutoSize`

bduffany:
`du -k` uses 1K = 1024 bytes (IEC), not 1000 bytes (SI).

This fixes an issue that `alpine` fails to be converted to ext4.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/util/ext4/ext4.go
+++ b/enterprise/server/util/ext4/ext4.go
@@ -19,6 +19,9 @@ const (
 	// image. The functions in this package which create disk images will fail
 	// if the provided size is any smaller.
 	MinDiskImageSizeBytes = 225e3
+
+	// The number of bytes in one IEC kilobyte (K).
+	iecKilobyte = 1024
 )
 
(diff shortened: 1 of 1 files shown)


enterprise/server/util/ext4/ext4.go
```diff
@@ -111,7 +111,7 @@ func DiskSizeBytes(ctx context.Context, inputDir string) (int64, error) {
 	if err != nil {
 		return 0, status.InternalErrorf("du output %q did not match 'SIZE  /file/path': %s", out, err)
 	}
-	return int64(s * 1000), nil
+	return int64(s * 1024), nil
```
maggie-lou, line 114: nit: Change to a constant so it's more clear what 1024 represents?

maggie-lou (approved):

enterprise/server/util/ext4/ext4.go
bduffany: done

PR #4887, opened 2023-09-26
Allow executor host_id to be manually specified via config option / flag

siggisim:
Fixes: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2728

### DIFF
--- a/server/util/uuid/uuid.go
+++ b/server/util/uuid/uuid.go
@@ -3,6 +3,7 @@ package uuid
 import (
 	"context"
 	"encoding/hex"
+	"flag"
 	"io"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


server/util/uuid/uuid.go
```diff
@@ -26,6 +27,8 @@ var (
 
 	failsafeID     string
 	failsafeIDOnce sync.Once
+
+	manualHostID = flag.String("executor.host_id", "", "Optional: Allows for manual specification of an executor's host id. If not set, a random UUID will be used.")
```
bduffany, line 31: does k8s have an easy way to map these flags to e.g. pod ID? wonder if we should support env vars too like we do for MY_POD_ID etc

siggisim: For env vars, you could do this in your config.yaml
```
executor:
  host_id: ${YOUR_ENVIRONMENT_VARIABLE}
```

bduffany (approved):

PR #4889, opened 2023-09-27
[UFFD] Fix bug where we try to map an address outside of the VM's memory range

maggie-lou:
To reduce the number of round trips, when the UFFD handler receives a page fault and fetches a chunk from the store to serve the page fault, we copy the entire chunk as opposed to just 1 page.

We had a bug where sometimes a chunk would map to addresses that fell out of the memory range allocated to the VM. The handler would get errno(2) and fail.

Ex. Let's say the VM's entire memory range is 100B from addresses 0 - 99. The uffd object will be registered for addresses 0-99. If one chunk is 200B, there will be an error when we try to copy to addresses 100-200.

### DIFF
--- a/enterprise/server/remote_execution/uffd/uffd.go
+++ b/enterprise/server/remote_execution/uffd/uffd.go
@@ -246,17 +246,39 @@ func (h *Handler) handle(ctx context.Context, memoryStore *copy_on_write.COWStor
 			}
 			return status.InternalErrorf("read event from uffd failed with errno(%d)", err)
 		}
+
+		mapping, err := guestMemoryAddrToMapping(uintptr(guestFaultingAddr), mappings)
+		if err != nil {
+			return err
+		}
+
 		guestPageAddr := pageStartAddress(guestFaultingAddr, pageSize)
+		if guestPageAddr < mapping.BaseHostVirtAddr {
+			// Make sure we only try to map addresses that fall within the valid
+			// guest memory ranges
+			guestPageAddr = mapping.BaseHostVirtAddr
+		}
 
 		// Find the memory data in the store that should be used to handle the page fault
-		faultStoreOffset, err := guestMemoryAddrToStoreOffset(guestPageAddr, mappings)
-		if err != nil {
-			return status.WrapError(err, "translate to store offset")
-		}
-		relOffset := memoryStore.GetRelativeOffsetFromChunkStart(faultStoreOffset)
+		faultStoreOffset := guestMemoryAddrToStoreOffset(guestPageAddr, *mapping)
+
+		// To reduce the number of UFFD round trips, try to copy the entire
+		// chunk containing the faulting address
 		hostAddr, allocatedLen, err := memoryStore.GetChunkStartAddressAndSize(uintptr(faultStoreOffset), false /*=write*/)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #4889** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4889" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4889?utm_source=stack-comment).

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -248,27 +248,55 @@ func (h *Handler) handle(ctx context.Context, memoryStore *copy_on_write.COWStor
...
+			hostAddr += invalidBytesAtChunkStart
+			allocatedLen -= int64(invalidBytesAtChunkStart)
+		}
+		// Check for data above the valid memory range
+		lastValidAddress := mapping.BaseHostVirtAddr + mapping.Size - 1
+		copyEndAddress := destAddr + uintptr(allocatedLen)
+		if copyEndAddress > lastValidAddress {
+			invalidBytesAtChunkEnd := int64(copyEndAddress - lastValidAddress)
+			allocatedLen -= invalidBytesAtChunkEnd
+		}
```
bduffany, line 291: it looks like there might be an off-by-one error in computing `invalidBytesAtChunkEnd` here because `lastValidAddress` is an inclusive end index (start + length - 1) but `mappingEndAddr` is an exclusive end index (start + length), and when those two things are subtracted I think the result is off by one:

```suggestion
		mappingEndAddr := mapping.BaseHostVirtAddr + mapping.Size
		copyEndAddr := destAddr + uintptr(allocatedLen)
		if copyEndAddr > mappingEndAddr {
			invalidBytesAtChunkEnd := int64(copyEndAddr - mappingEndAddr)
			allocatedLen -= invalidBytesAtChunkEnd
		}
```

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -248,27 +248,55 @@ func (h *Handler) handle(ctx context.Context, memoryStore *copy_on_write.COWStor
...
+		// If copying the entire chunk would map data falling outside the valid
+		// guest memory range, only copy the valid parts of the chunk
+		relOffset := memoryStore.GetRelativeOffsetFromChunkStart(faultStoreOffset)
+		destAddr := guestPageAddr - relOffset
+		// Check for data below the valid memory range
+		if destAddr < mapping.BaseHostVirtAddr {
+			invalidBytesAtChunkStart := mapping.BaseHostVirtAddr - destAddr
+			destAddr = mapping.BaseHostVirtAddr
+			hostAddr += invalidBytesAtChunkStart
+			allocatedLen -= int64(invalidBytesAtChunkStart)
```
bduffany, line 283: nit: maybe rename `allocatedLen` to `copyLen` or `copySize`? The "allocated" part feels slightly unclear

enterprise/server/remote_execution/uffd/uffd.go
maggie-lou: Ah thanks :) Saved me from having to find this

tylerwilliams (approved):

PR #4892, opened 2023-09-27
Move replay_action to OSS, with some improvements

bduffany:
Improvements:
* Migrate off of deprecated `github.com/golang/protobuf/proto` package.
* Add `-n` and `-jobs` flags to allow replaying the action several times, optionally in parallel.
* Print action stdout and stderr (only when `n=1`, to avoid spamming the console)
* Instead of traversing the input directory tree using chained RPCs, use `GetTree`.
* Instead of having to pass `-skip_copy_files` on replay attempts after the first attempt, use `FindMissingBlobs` to check whether blobs need to be re-uploaded. Adds a `FindMissingBatcher` to transparently batch the digest existence checks.

**Related issues**: N/A

### DIFF
--- a/enterprise/tools/replay_action/replay_action.go
+++ b/enterprise/tools/replay_action/replay_action.go
@@ -1,0 +1,271 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"flag"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
+	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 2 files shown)


enterprise/tools/replay_action/replay_action.go
```diff
@@ -0,0 +1,440 @@
...
+	Digest       *repb.Digest
+	ResponseChan chan findMissingResponse
+}
+
+type findMissingResponse struct {
+	Missing bool
+	Error   error
+}
+
+type FindMissingBatcherOpts struct {
```
tylerwilliams, line 309: I don't think this should be required for replay_action to work, should it? Would it be sufficient to just send single find missing requests?

bduffany: For actions with lots of inputs (which are often the ones I want to replay) the round-trip overhead of the FindMissing requests is very high. For example I tried replaying one of our heavier test actions and when fully cached it takes 20s for the FindMissing checks (without batching), vs ~2s with batching enabled.

The goal here was mainly just to save time & frustration when trying to debug - I would use the `-skip_copy_files` flag to save time but it was frustrating because I would keep forgetting to set it.

tylerwilliams: Sgtm

tylerwilliams (approved):
LGTM

PR #4894, opened 2023-09-27
Don't single-flight-dedupe reading soci artifacts from cache

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/sociartifactstore/sociartifactstore.go
+++ b/enterprise/server/sociartifactstore/sociartifactstore.go
@@ -244,32 +244,36 @@ func (s *SociArtifactStore) GetArtifacts(ctx context.Context, req *socipb.GetArt
 	}
 
 	resp, err := s.getArtifactsFromCache(ctx, configHash)
-	if status.IsNotFoundError(err) {
-		log.CtxDebugf(ctx, "soci artifacts for image %s missing from cache: %s", targetImageRef.DigestStr(), err)
-		// Try to only pull-index-write once at a time to prevent hammering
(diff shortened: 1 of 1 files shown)


enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -243,36 +243,33 @@ func (s *SociArtifactStore) GetArtifacts(ctx context.Context, req *socipb.GetArt
...
 	}
 
-	// Try to only read-pull-index-write once at a time to prevent hammering
-	// the containter registry with a ton of parallel pull requests, and save
-	// apps a bunch of parallel work.
-	workKey := fmt.Sprintf("soci-artifact-store-image-%s", configHash.Hex)
-	respBytes, err := s.deduper.Do(ctx, workKey, func() ([]byte, error) {
-		resp, err := s.getArtifactsFromCache(ctx, configHash)
-		if status.IsNotFoundError(err) {
+	resp, err := s.getArtifactsFromCache(ctx, configHash)
```
vadimberezniker, line 246: remove the extra level of nesting

if err == nil {
  return resp, nil
}
if err != nil && !status.IsNotFoundError(err) {
  return nil, err
}

...the rest...

vadimberezniker (changes requested):

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: nice, done

vadimberezniker (approved):

PR #4899, opened 2023-09-28
Support fetching executors from multiple regions

siggisim:
Screenshot 2023-09-27 at 5 23 16 PM

Decided to just fetch executors in parallel from each region and display them for now. We can add a dropdown once the number of regions gets higher.

Probably still need to set a CORS header that includes these regions, but that's tough to test locally.


bduffany (approved):
LGTM modulo `bb fix`

tylerwilliams (approved):

PR #4900, opened 2023-09-28
Ensure client sets the digest type when necessary

tylerwilliams:
This makes blake3 action results viewable and all files downloadable.

### DIFF
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -365,7 +365,7 @@ export default class InvocationModel {
   }
 
   getDigestFunctionDir() {
-    if (this.optionsMap.get("digest_function").toLowerCase() == "blake3") {
+    if (this.optionsMap.get("digest_function")?.toLowerCase() == "blake3") {
       return "blake3/";
     }
     return "";
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -364,6 +364,13 @@ export default class InvocationModel {
     return this.optionsMap.get("remote_instance_name") ?? "";
   }
 
+  getDigestFunctionDir() {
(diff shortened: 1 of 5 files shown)


app/invocation/invocation_model.tsx
```diff
@@ -364,6 +364,13 @@ export default class InvocationModel {
     return this.optionsMap.get("remote_instance_name") ?? "";
   }
 
+  getDigestFunctionDir() {
```
jdhollen, line 367: i will admit that i like this better without the slash, but i feel like i'm gonna cause a bug by asking for it.

jdhollen (approved):

app/invocation/invocation_model.tsx
```diff
@@ -364,6 +364,13 @@ export default class InvocationModel {
...
+    if (this.optionsMap.get("digest_function").toLowerCase() == "blake3") {
```
jdhollen, line 368: (just need to add a ? after the get() here to make tsc happy)

app/invocation/invocation_model.tsx
bduffany: I also like it better without the slash, but am fine with keeping this as-is, since eventually I'd like to get rid of this func anyway and replace all of our manual URL construction with a cache util package

bduffany (approved):

app/invocation/invocation_model.tsx
tylerwilliams: same, but it makes concatenation logic at the callsites gnarlier and I assume whomever makes the url-lib will fix this one.

app/invocation/invocation_model.tsx
tylerwilliams: Done.

PR #4902, opened 2023-09-28
bazelrc: support windows specific configs

sluongng:
Also ensure that C++17 is passed to compiler for all configurations:
exec(host_cxxopt) and others(cxxopt).

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,6 @@
+build --enable_platform_specific_config
+query --enable_platform_specific_config
+
 # Skymeld merges the analysis phase and execution phase.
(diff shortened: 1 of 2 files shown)


sluongng:
Few small changes since last review:

- I stop using `common` in favor of `build + query` because our `bb` CLI parser does not support `common` correctly

- Only make runfiles the default on Linux and Mac. On Windows, enabling runfiles will stop tests from running with `cannot shorten the path enough`

sluongng:
Raised the runfiles issue here https://github.com/bazelbuild/bazel/issues/19710

PR #4905, opened 2023-09-28
Store the image-to-SOCI-index mapping as an ActionResult

iain-macdonald:
This PR changes the format of the image-to-SOCI-index that's stored in the ActionCache. Previously, a cache entry was written using the rehashed, salted image's config hash as the key and the serialized digest of the SOCI index as the value. However, this had a few drawbacks including increased difficulty debugging the ActionCache entries using the ActionCacheServer APIs and requiring multiple trips to the cache to unpack a SOCI index. This PR changes the image-to-index mapping to be stored as an ActionResult with all of the SOCI artifacts (index and ZToCs) stored as OutputFiles with a NodeProperty denoting their type. This should address both shortcomings mentioned previously.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/sociartifactstore/sociartifactstore.go
+++ b/enterprise/server/sociartifactstore/sociartifactstore.go
@@ -297,7 +297,11 @@ func (s *SociArtifactStore) getArtifactsFromCache(ctx context.Context, imageConf
 			InlineStdout: true,
 		})
 	if err != nil {
-		recordOutcome("soci_index_pointer_contains_error")
+		if status.IsNotFoundError(err) {
+			recordOutcome("action_result_missing")
+		} else {
+			recordOutcome("action_result_error")
+		}
 		return nil, err
 	}
 	// The SOCI Index digest is serialized in the stdout field so we can
@@ -308,7 +312,7 @@ func (s *SociArtifactStore) getArtifactsFromCache(ctx context.Context, imageConf
 	}
 	sociIndexResourceName, err := digest.ParseDownloadResourceName(string(actionResult.StdoutRaw))
 	if err != nil {
-		recordOutcome("malformed_soci_index_pointer")
+		recordOutcome("soci_index_pointer_malformed")
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -450,16 +436,27 @@ func (s *SociArtifactStore) indexImage(ctx context.Context, image v1.Image, conf
...
+	// - The digest of all artifacts (ZToCs and SOCI Index) as OutputFiles
+	var artifacts []*repb.OutputFile
+	artifacts = append(artifacts, &repb.OutputFile{Digest: indexResourceName.GetDigest()})
+	for _, ztocDigest := range ztocDigests {
+		artifacts = append(artifacts, &repb.OutputFile{Digest: ztocDigest})
+	}
+	req := repb.UpdateActionResultRequest{
+		ActionDigest: sociIndexCacheKey,
+		ActionResult: &repb.ActionResult{
+			StdoutRaw:   []byte(serializedIndexResourceName),
```
vadimberezniker, line 455: I don't like overloading Stdout like this. What if we stored it as another output file, using path or node_properties to identify the type of data?

vadimberezniker (changes requested):

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: SG, done

enterprise/server/sociartifactstore/sociartifactstore.go
```diff
@@ -285,59 +289,52 @@ func sociIndexKey(h v1.Hash) (*repb.Digest, error) {
...
+		} else if outputFile.NodeProperties.Properties[0].Value == ztocArtifactType {
+			ztocDigests = append(ztocDigests, outputFile.Digest)
+		} else {
+			recordOutcome("action_result_bad_node_prop_value")
+			return nil, status.InternalErrorf(
+				"malformed action result, unrecognized node property value: %s",
+				outputFile.NodeProperties.Properties[0].Value)
 		}
 	}
 	recordOutcome("cached")
```
vadimberezniker, line 336: add a check here to verify that sociIndexDigest and ztocDigests were populated

vadimberezniker (approved):

enterprise/server/sociartifactstore/sociartifactstore.go
iain-macdonald: Added a check for the SOCI index. It's technically valid (but not very useful) for all the ZToCs to be missing if all of the image layers are below the minimum layer size, so I omitted that chekc.

vadimberezniker (approved):

PR #4907, opened 2023-09-28
Fix UFFD race condition between Wait and Stop

maggie-lou:
We call Wait in a background thread in firecracker to check if the handler has shut down early due to an error. We call Stop() when we are explicitly telling the handler to shut down, for example if we are pausing the VM. There is a race condition where Stop might try to set handleDoneChan=nil at the same time that the background Wait thread is trying to read it.

### DIFF
--- a/enterprise/server/remote_execution/uffd/uffd.go
+++ b/enterprise/server/remote_execution/uffd/uffd.go
@@ -340,7 +340,6 @@ func (h *Handler) Stop() error {
 
 	// Wait for handle() to finish processing open requests
 	<-h.handleDoneChan
-	h.handleDoneChan = nil
 
 	h.earlyTerminationReader.Close()
 	h.earlyTerminationWriter.Close()


maggie-lou:
* master
  * **PR #4907** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4907" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4907?utm_source=stack-comment).

maggie-lou:
Ultimately the goal is that if the UFFD handler fails, we want to immediately cancel the VM context so it doesn't waste time trying to continue to run. 

Alternatively, what if we added a `cancelVmCtx context.CancelCauseFunc` field to the UFFD handler? (Instead of introducing a background thread that listens on a channel and could cause race conditions like this one)

Here is the [PR](https://github.com/buildbuddy-io/buildbuddy/pull/4886/files) where the race condition was introduced

tylerwilliams (approved):

PR #4908, opened 2023-09-28
Make bytebufferpool less error-prone.

vadimberezniker:
When buffers are returned, look at the capacity instead of the length to make sure the buffers are returned to the right bucket. There's at least one codepath where the buffer is returned with a smaller length.

Since we're now looking at the capacity instead of the length, we can automatically return buffers of the correct length instead of asking callers to do it.

**Related issues**: N/A


tylerwilliams (approved):
This looks pretty great

PR #4910, opened 2023-09-28
Fix cli to work with `--enable_platform_specific_config`

tempoz:
Since we do a custom parse of the `.bazelrc`, we need to handle the `--enable_platform_specific_config` specially, as it is supposed to be equivalent to `--config=<os>`. This PR substitutes the last instance of this flag, if it is set to true, with the appropriate `--config` option immediately before we do `--config` substitution.

### DIFF
--- a/cli/parser/parser.go
+++ b/cli/parser/parser.go
@@ -683,6 +683,8 @@ func ExpandConfigs(args []string) ([]string, error) {
 	return CanonicalizeArgs(args)
 }
 
+// Mirroring the behavior here:
+// https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/ConfigExpander.java#L41
 func getBazelOS() string {
 	switch runtime.GOOS {
 	case "linux":
--- a/cli/parser/parser.go
+++ b/cli/parser/parser.go
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	workspacePrefix = `%workspace%/`
+	workspacePrefix                  = `%workspace%/`
+	enablePlatformSpecificConfigFlag = "enable_platform_specific_config"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


tempoz:
* master
  * **PR #4910** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4910" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4910?utm_source=stack-comment).

cli/parser/parser.go
```diff
@@ -682,6 +683,23 @@ func ExpandConfigs(args []string) ([]string, error) {
 	return CanonicalizeArgs(args)
 }
 
+func getBazelOS() string {
```
siggisim, line 686: I'll let Brandon have a look at the expansion logic since this is his baby, but could we add a link to the Bazel source here in a comment https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/ConfigExpander.java#L41 so we can attempt to keep these in sync?

tempoz: Sounds great!

```suggestion
// Mirroring the behavior here:
// https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/ConfigExpander.java#L41
func getBazelOS() string {
```

sluongng:
Not directly related but I was hoping that I could use this

```
# .bazelrc
common --enable_platform_specific_config
```

however our BB cli parser, afaik, does not handle the new `common` logic in `.bazelrc`.

So most likely we gonna need to do something like 

```
# .bazelrc
build --enable_platform_specific_config
fetch --enable_platform_specific_config
query --enable_platform_specific_config
```

cli/parser/parser.go
```diff
@@ -732,6 +752,18 @@ func expandConfigs(workspaceDir string, args []string) ([]string, error) {
...
+	enable, enableIndex, enableLength := arg.FindLast(args, "enable_platform_specific_config")
+	noEnable, noEnableIndex, noEnableLength := arg.FindLast(args, "noenable_platform_specific_config")
+	if enableIndex > noEnableIndex {
+		if enable == "true" || enable == "yes" || enable == "1" || enable == "" {
+			args = concat(args[:enableIndex], []string{"--config", getBazelOS()}, args[enableIndex+enableLength:])
+		}
+	} else if noEnableIndex > enableIndex {
+		if noEnable == "false" || noEnable == "no" || noEnable == "0" {
+			args = concat(args[:noEnableIndex], []string{"--config", getBazelOS()}, args[noEnableIndex+noEnableLength:])
+		}
```
sluongng, line 764: I could be understanding this wrongly, but validating `noEnable` value might not be needed.

```
> bazel build --noenable_platform_specific_config=no //...
ERROR: --noenable_platform_specific_config=no :: Unexpected value after boolean option: --noenable_platform_specific_config=no
```

tempoz: Oh, excellent. I didn't realize that that was how `no` flags worked. Thanks, fixed!

siggisim (approved):

PR #4920, opened 2023-09-29
[WIP] Add locking to copy on write structs due to concurrent access

maggie-lou:
FUSE serves requests in goroutines (https://github.com/hanwen/go-fuse/blob/master/fuse/server.go#L497). This has introduced a race condition.


maggie-lou:
* master
  * **PR #4920** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4920" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4920?utm_source=stack-comment).

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -686,40 +693,56 @@ func (m *Mmap) initMap() error {
...
+	if !m.safeReadMapped() {
 		if err := m.initMap(); err != nil {
 			return 0, err
 		}
 	}
-	if err := checkBounds("read", int64(len(m.data)), p, off); err != nil {
+	dataLen, err := m.SizeBytes()
+	if err != nil {
+		return 0, err
+	}
```
bduffany, line 704: if we move the bounds check after `RLock` / `RUnlock` then I think we can go back to using `len(m.data)` directly, which seems a bit simpler and doesn't result in the extra locking overhead incurred by the SizeBytes() call. Same in WriteAt below.

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -686,40 +693,56 @@ func (m *Mmap) initMap() error {
...
+	if err := checkBounds("write", dataLen, p, off); err != nil {
 		return 0, err
 	}
+	m.mu.Lock()
 	m.lazyDigest = nil
 	copy(m.data[int(off):int(off)+len(p)], p)
+	m.mu.Unlock()
```
bduffany, line 730: nit: defer the unlock

```suggestion
	m.mu.Lock()
	defer m.mu.Unlock()
	m.lazyDigest = nil
	copy(m.data[int(off):int(off)+len(p)], p)
```

enterprise/server/remote_execution/copy_on_write/copy_on_write_test.go
```diff
@@ -64,6 +65,43 @@ func TestMmap_Digest(t *testing.T) {
 
 }
 
+func TestMmap_Concurrency(t *testing.T) {
+	s, _ := newMmap(t)
```
bduffany, line 69:
```suggestion
	s, err := newMmap(t)
	require.NoError(t, err)
```

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -584,11 +585,15 @@ func ConvertFileToCOW(ctx context.Context, env environment.Env, filePath string,
...
 type Mmap struct {
 	ctx                context.Context
 	env                environment.Env
 	remoteInstanceName string
 
+	mu sync.RWMutex
+
 	Offset int64
 
 	data       []byte
```
bduffany, line 599: nit: move the mutex above the fields that it's protecting

(also, `dataDir` can be moved up next to `Offset` probably)

```suggestion
	Offset int64

	mu         sync.RWMutex
	data       []byte
```

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -392,12 +392,13 @@ func (s *COWStore) copyChunkIfNotDirty(chunkStartOffset int64) (err error) {
...
 	dst, err := s.initDirtyChunk(chunkStartOffset, dstChunkSize)
 	if err != nil {
+		s.mu.Unlock()
 		return status.WrapError(err, "initialize dirty chunk")
 	}
-
-	s.mu.Lock()
 	src := s.chunks[chunkStartOffset]
 	s.chunks[chunkStartOffset] = dst
 	s.dirty[chunkStartOffset] = true
```
bduffany, line 404: If we move this logic into `initDirtyChunk` and have that func return `src, dst, err`, then we can make this a bit safer by doing the `Lock() / defer Unlock()` at the top of `initDirtyChunk` and removing the locking here.

enterprise/server/remote_execution/copy_on_write/copy_on_write_test.go
maggie-lou: The second return arg is a path, not an error

bduffany (approved):
nice!

maggie-lou:
Hey @bduffany can you take another look? I realized that more of the reading / writing behavior should be atomic, though unfortunately it means that we need to lock around larger areas of code

I looked into using lockmap for finer grained locking, but don't think it will work in this case because we need to lock across ranges of chunks. For example let's say we have (A) a write to chunks (2, 3, 4) and  (B) a read on chunks (1, 2, 3). If  A finishes writing to chunk 2 but hasn't started writing to chunk 3, we don't want B to be able to read the modified chunk 2 and an unmodified chunk 3.

bduffany:
> For example let's say we have (A) a write to chunks (2, 3, 4) and (B) a read on chunks (1, 2, 3). If A finishes writing to chunk 2 but hasn't started writing to chunk 3, we don't want B to be able to read the modified chunk 2 and an unmodified chunk 3.

For VBD, we need to provide the same atomicity guarantees as what is provided by `read()` / `write()` for regular Unix files, since the block device is being exposed to Firecracker as a plain old file.

I think files don't have atomicity guarantees across overlapping ranges of bytes like that (see https://stackoverflow.com/a/35595686). So for example when trying to read() and write() the same range in a regular file concurrently, the reader might see some of the writer's output interleaved in the result.

So I think it's up to Firecracker's block device driver (not us) to order the reads and writes in such a way that it prevents race conditions with overlapping byte ranges.

bduffany (approved):
I think we can merge this as-is since it's better than what we have now in terms of correctness, and it's better to have it be correct than broken. Separately I can check whether the lock around Write() has a significant performance impact, and if so we can do the per-chunk locking like we discussed offline

PR #4926, opened 2023-09-29
Extend X-Forwarded-For parsing.

vadimberezniker:
- Support single-IP format used by NGINX.
 - Don't trust X-Forwarded-For header unless explicitly enabled by flag.
 - Fall back to peer IP for RPCs if header is absent/not trusted.

**Related issues**: N/A

### DIFF
--- a/server/util/clientip/clientip.go
+++ b/server/util/clientip/clientip.go
@@ -3,9 +3,15 @@ package clientip
 import (
 	"context"
 	"strings"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/flag"
 )
 
 const ContextKey = "clientIP"
+
+var (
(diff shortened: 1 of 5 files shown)


server/util/clientip/clientip.go
```diff
@@ -15,10 +21,18 @@ func Get(ctx context.Context) string {
...
-	if len(ips) < 2 {
+	if !*trustXForwardedForHeader || header == "" {
 		return ctx, false
 	}
+
+	ips := strings.Split(header, ",")
+
+	// If there's only a single IP in the header, return it directly.
+	// This handles the header format set by NGINX.
+	if len(ips) == 1 {
```
tylerwilliams, line 32: would be nice to add a test for this

tylerwilliams (approved):

server/util/clientip/clientip.go
vadimberezniker: added

PR #4932, opened 2023-10-02
Add container type to ImageCacheToken to prevent sharing tokens across container types

iain-macdonald:
This prevents a theoretical bug where we accidentally share tokens for the same image across isolation types. This shouldn't happen, but Brandon and I noticed it as a possibility in https://github.com/buildbuddy-io/buildbuddy/pull/4869

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -166,6 +166,11 @@ type FileSystemLayout struct {
 
 // CommandContainer provides an execution environment for commands.
 type CommandContainer interface {
+
+	// Returns the isolation type of this command container, for example:
+	// "docker" or "podman".
+	ContainerType() platform.ContainerType
+
 	// Run the given command within the container and remove the container after
 	// it is done executing.
 	//
@@ -250,7 +255,7 @@ func PullImageIfNecessary(ctx context.Context, env environment.Env, ctr CommandC
 	if err != nil {
 		return err
 	}
-	cacheToken, err := NewImageCacheToken(ctx, env, creds, imageRef)
+	cacheToken, err := NewImageCacheToken(ctx, env, creds, imageRef, ctr.ContainerType())
 	if err != nil {
 		return status.WrapError(err, "create image cache token")
 	}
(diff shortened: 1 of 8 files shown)


bduffany (commented):
I think we shouldn't partition these tokens by isolation type, because all of the container types are doing the same authorization check - passing the image credentials to the remote registry. (My main concern here isn't efficiency / reusing the tokens as much as possible, but more so just making the auth simpler and less error-prone.)

Rather than making this change I think we should simplify the OCI auth process. The current contract for `PullImage()` is that it **must** always authorize with the remote registry and then **optionally** download the image from the remote registry (i.e. the download can be skipped if it's cached). This contract is a bit too subtle/error-prone IMO and could be simplified - particularly the "must always authorize" part (it's too easy to write `if isCached { return nil }` at the start of the func, which is wrong because it skips the auth check).

To simplify, we could remove the "must always authorize" part, and have the authorization be handled _externally_ to the container interface (rather than internally), using an explicit authenticator that uses that `crane` library func you found that does a HEAD request for the container image using the user's credentials. This way, registry auth is handled in one place, and each container impl doesn't have to worry about auth.

To make this simplification I think we'd want to get rid of ImageCacheAuthenticator + ImageCacheToken and instead have a simpler interface like this:

```
// OCIAuthenticator authenticates OCI image registry credentials.
type OCIAuthenticator struct {
  // Authenticate returns an error if the given image pull credentials are not valid.
  // It does so by making a HEAD request to the remote registry.
  // As an optimization to avoid excessive overhead due to registry round-trips,
  // the result of this function may be cached internally for up to 5 minutes.
  Authenticate(ctx context.Context, imageRef, user, password string) error
}
```

then `PullImageIfNecessary` could be simplified to just

```
err := ociAuth.Authenticate(...)
// check err...
err = container.PullImage(...) // internally can choose to do nothing if the image is cached
// check err...
```

iain-macdonald:
That sounds good, I'll pursue the more in-depth refactoring you suggested and we discussed. Stay tuned!

PR #4934, opened 2023-10-02
[WIP] Fix runner pool shutdown race

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -1048,6 +1048,10 @@ func (p *pool) take(ctx context.Context, key *rnpb.RunnerKey) (*commandRunner, e
 	p.mu.Lock()
 	defer p.mu.Unlock()
 
+	if p.isShuttingDown {
+		return nil, status.UnavailableErrorf("Could not take a runner from the pool because the executor is shutting down.")
+	}
+
 	log.CtxInfof(ctx, "Looking for match for %q in runner pool %s", keyString(key), p)
 	taskKeyBytes, err := proto.Marshal(key)
 	if err != nil {
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -371,7 +371,7 @@ func (r *commandRunner) Run(ctx context.Context) *interfaces.CommandResult {
 	case ready:
 		break
 	case removed:
-		return commandutil.ErrorResult(status.UnavailableErrorf("Not starting new task since executor is shutting down"))
+		return commandutil.ErrorResult(shutdownError("run task"))
 	default:
 		return commandutil.ErrorResult(status.InternalErrorf("unexpected runner state %d; this should never happen", r.state))
 	}
@@ -581,7 +581,7 @@ func (p *pool) checkAddPreconditions(r *commandRunner) *labeledError {
 
 	if p.isShuttingDown {
 		return &labeledError{
-			status.UnavailableError("pool is shutting down; new runners cannot be added."),
+			shutdownError("add new runner"),
 			"pool_shutting_down",
 		}
 	}
@@ -901,7 +901,7 @@ func (p *pool) Get(ctx context.Context, st *repb.ScheduledTask) (interfaces.Runn
 		PersistentWorkerKey: effectivePersistentWorkerKey(props, task.GetCommand().GetArguments()),
 	}
 	if props.RecycleRunner {
-		r := p.takeWithRetry(ctx, key)
+		r, err := p.takeWithRetry(ctx, key)
 		if r != nil {
 			p.mu.Lock()
 			r.task = task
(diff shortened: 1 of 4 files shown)


maggie-lou:
* master
  * **PR #4934** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4934" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4934?utm_source=stack-comment).

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -1048,6 +1069,10 @@ func (p *pool) take(ctx context.Context, key *rnpb.RunnerKey) (*commandRunner, e
 	p.mu.Lock()
 	defer p.mu.Unlock()
 
+	if p.isShuttingDown {
```
maggie-lou, line 1072: This line should fix the panic. The other changes are optimizations so that we exit a little earlier on shut down

bduffany (commented):
I haven't communicated this super well, but `runner.go` is becoming very unwieldy so I would like to work towards removing code from it rather than adding to it (it's on my TODO list to simplify the locking, remove the co-dependence of runner + runnerPool, and also move the persistent worker stuff into a separate package).

There's another approach which I think would be a little simpler and less error prone while also not touching `runner.go` - we already have `TracedCommandContainer` which wraps all container operations with tracing. I think we could augment `TracedCommandContainer` to do two additional things (and maybe rename it to `ContainerWrapper` or something more general):

1. Wrap each container delegation call with a mutex Lock/Unlock, to prevent container operations from ever being called concurrently.
2. Before each operation (after acquiring the mutex), check whether `Remove()` has been called, and if so, return `status.UnavailableError("container has been removed, likely due to executor shutdown")`

e.g. `Exec()` would look like this

```golang
func (c *TracedCommandContainer) Exec(ctx context.Context) error {
  ctx, span := tracing.StartSpan(ctx)
  defer span.End()
  c.mu.Lock()
  defer c.mu.Unlock()
  if (c.removed) {
    return status.UnavailableError("container has been removed, likely due to executor shutdown")
  }
  return c.Exec(ctx)
}
```

maggie-lou:
> I haven't communicated this super well, but `runner.go` is becoming very unwieldy so I would like to work towards removing code from it rather than adding to it (it's on my TODO list to simplify the locking, remove the co-dependence of runner + runnerPool, and also move the persistent worker stuff into a separate package).
> 
> There's another approach which I think would be a little simpler and less error prone while also not touching `runner.go` - we already have `TracedCommandContainer` which wraps all container operations with tracing. I think we could augment `TracedCommandContainer` to do two additional things (and maybe rename it to `ContainerWrapper` or something more general):
> 
> 1. Wrap each container delegation call with a mutex Lock/Unlock, to prevent container operations from ever being called concurrently.
> 2. Before each operation (after acquiring the mutex), check whether `Remove()` has been called, and if so, short-circuit with an error.
> 
> e.g. `Exec()` would look like this
> 
> ```go
> func (c *TracedCommandContainer) Exec(ctx context.Context) error {
>   ctx, span := tracing.StartSpan(ctx)
>   defer span.End()
>   c.mu.Lock()
>   defer c.mu.Unlock()
>   if (c.removed) {
>     return status.UnavailableError("container has been removed, likely due to executor shutdown")
>   }
>   return c.Exec(ctx)
> }
> ```

I actually found another bug that might be the root cause of these errors and I think would still be required for your solution. In pool.Shutdown, we should mark the runner.state=removed [before we unlock](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/runner/runner.go#L1252). Right now we're doing it in runner.Remove(). Otherwise there could be a race condition where we've added all these runners in a list to remove, but they're still eligible to be used by new tasks coming in.

enterprise/server/remote_execution/runner/runner.go
maggie-lou: Looking into this more, I actually don't think this is the fix. This is just another optimization so we shut down earlier on shut down

bduffany:
> I actually found another bug that might be the root cause of these errors and I think would still be required for your solution. In pool.Shutdown, we should mark the runner.state=removed [before we unlock](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/runner/runner.go#L1252). Right now we're doing it in runner.Remove(). Otherwise there could be a race condition where we've added all these runners in a list to remove, but they're still eligible to be used by new tasks coming in.

We set `p.runners = activeRunners` before unlocking, so I think that should never happen? https://github.com/buildbuddy-io/buildbuddy/blob/4d2b250c40df69f6edda130d8d937345ef958b59/enterprise/server/remote_execution/runner/runner.go#L1214

maggie-lou:
> We set `p.runners = activeRunners` before unlocking, so I think that should never happen?
> 
> https://github.com/buildbuddy-io/buildbuddy/blob/4d2b250c40df69f6edda130d8d937345ef958b59/enterprise/server/remote_execution/runner/runner.go#L1214

Hmmm good point. I missed that (Reason why your refactor would be a good idea). There still must be a bug in the runner code though, that I haven't been able to find. Your solution would enforce preventing Exec() and Remove() from simultaneously happening and fix the panics, but they should never be simultaneously called in the first place

PR #4942, opened 2023-10-03
Store chunks outside of pebble

luluz66:
**Related issues**: N/A
Store chunks outside of pebble if they are bigger than maxInlineFileSizeBytes. 
When tested locally, the eviction latency doesn't spike, the eviction rate don't
plateaued and can get to about the same as the current rate in production, no
compaction debt accumulated.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -2002,17 +2002,17 @@ func (cdcw *cdcWriter) writeChunk(chunkData []byte) error {
 func (cdcw *cdcWriter) writeRawChunk(fileRecord *rfpb.FileRecord, key filestore.PebbleKey, chunkData []byte) error {
 	ctx := cdcw.ctx
 	p := cdcw.pc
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2172,21 +2172,25 @@ func (p *PebbleCache) Writer(ctx context.Context, r *rspb.ResourceName) (interfa
...
+		// Files smaller than averageChunkSizeBytes are highly like to only
+		// have one chunk, so we skip cdc-chunking step.
 		return p.newCDCCommitedWriteCloser(ctx, fileRecord, key, shouldCompress, isCompressed)
 	}
 
+	return p.newWrappedWriter(ctx, fileRecord, key, shouldCompress, rfpb.FileMetadata_COMPLETE_FILE_TYPE)
+}
+
+// newWrappedWriter returns an interfaces.CommittedWriteCloser that on Write,
+// it will
```
tylerwilliams, line 2185:
```suggestion
// newWrappedWriter returns an interfaces.CommittedWriteCloser that on Write
// will:
```

tylerwilliams (approved):

PR #4945, opened 2023-10-04
distributed: fix potential panic

sluongng:
When replication factor is higher than the actual peer count, such as
running 1 pod with replication factor set to 3, we would run into:

```
panic: runtime error: slice bounds out of range [:3] with capacity 0
```

To avoid this, let's make sure that we only sub-slicing when the
replication factor was set correctly.

Also adjust the default replication factor to be inline with our
documentation "Must be >=1".

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/distributed/distributed.go
+++ b/enterprise/server/backends/distributed/distributed.go
@@ -397,10 +397,7 @@ func dedupe(in []string) []string {
 func (c *Cache) readPeers(d *repb.Digest) *peerset.PeerSet {
 	peers := c.consistentHash.GetAllReplicas(d.GetHash())
 	var primaryPeers, secondaryPeers []string
-	if len(peers) < c.config.ReplicationFactor {
-		primaryPeers = []string{}
-		secondaryPeers = []string{}
-	} else {
+	if len(peers) >= c.config.ReplicationFactor {
 		primaryPeers = peers[:c.config.ReplicationFactor]
 		secondaryPeers = peers[c.config.ReplicationFactor:]
 	}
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/distributed/distributed.go
```diff
@@ -390,13 +390,25 @@ func dedupe(in []string) []string {
...
-		allPrimaryPeers := extendedPeerList[:c.config.ReplicationFactor]
-		allSecondaryPeers := extendedPeerList[c.config.ReplicationFactor:]
+		var allPrimaryPeers, allSecondaryPeers []string
+		if len(extendedPeerList) < c.config.ReplicationFactor {
+			allPrimaryPeers = []string{}
+			allSecondaryPeers = []string{}
+		} else {
+			allPrimaryPeers = extendedPeerList[:c.config.ReplicationFactor]
+			allSecondaryPeers = extendedPeerList[c.config.ReplicationFactor:]
+		}
```
tylerwilliams, line 411: I think it would make more sense to move this check (that replication factor <= peer count) to the health check -- that way the cache emits a warning and does not become healthy if misconfigured.

(that's the `func (c *Cache) Check(ctx context.Context) error {` method)

This check should happen even if the nodelist is hardcoded btw.

enterprise/server/backends/distributed/distributed.go
```diff
@@ -390,24 +396,29 @@ func dedupe(in []string) []string {
 // peers are returned in random order.
 func (c *Cache) readPeers(d *repb.Digest) *peerset.PeerSet {
 	peers := c.consistentHash.GetAllReplicas(d.GetHash())
-	primaryPeers := peers[:c.config.ReplicationFactor]
-	secondaryPeers := peers[c.config.ReplicationFactor:]
+	var primaryPeers, secondaryPeers []string
+	if len(peers) >= c.config.ReplicationFactor {
```
tylerwilliams, line 400: would be nice to add a tiny comment here explaining this logic

// To prevent a panic if replication is misconfigured to be higher than peer count.

tylerwilliams (approved):

PR #4949, opened 2023-10-04
Prevent container.Remove() being called while other operations are in progress

bduffany:
Ensure that we can't call `container.Remove()` while other operations are in progress, and also make operations fail immediately after the container is already removed.

This will likely fix https://github.com/buildbuddy-io/buildbuddy-internal/issues/2554, and at minimum it should give us better information as to whether a container operation is failing because of `Remove()` having already been called (on executor shutdown).

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -482,9 +482,6 @@ func (t *TracedCommandContainer) Exec(ctx context.Context, command *repb.Command
 	ctx, span := tracing.StartSpan(ctx, trace.WithAttributes(t.implAttr))
 	defer span.End()
 
-	log.Infof("Exec()...")
-	defer log.Infof("Exec done!")
-
 	t.mu.RLock()
 	defer t.mu.RUnlock()
 	if t.removed {
@@ -524,10 +521,8 @@ func (t *TracedCommandContainer) Remove(ctx context.Context) error {
 	ctx, span := tracing.StartSpan(ctx, trace.WithAttributes(t.implAttr))
 	defer span.End()
 
-	log.Infof("Remove()")
-
-	// Get an *exclusive* lock here to ensure any other pending operations are
-	// completed before we remove.
+	// Get an *exclusive* lock here to ensure any other concurrent operations
+	// are completed before we remove.
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	if t.removed {
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -33,12 +33,21 @@ const (
 	// Time window over which to measure CPU usage when exporting the milliCPU
 	// used metric.
 	cpuUsageUpdateInterval = 1 * time.Second
+
+	// Exit code used when returning an error instead of an actual exit code.
+	// TODO: fix circular dependency with commandutil and reference that const
+	// instead.
+	noExitCode = -2
 )
 
 var (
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/runner/runner.go
```diff
@@ -437,14 +437,14 @@ func (r *commandRunner) Remove(ctx context.Context) error {
...
-			errs = append(errs, err)
-		}
 		if r.stopPersistentWorker != nil {
 			if err := r.stopPersistentWorker(); err != nil {
 				errs = append(errs, err)
 			}
 		}
+		if err := r.Container.Remove(ctx); err != nil {
+			errs = append(errs, err)
+		}
```
bduffany, line 447: This had to be moved down in order to get the persistent worker's long-running `Exec()` call to return before we can call `Remove()` on it; otherwise there's a deadlock.

enterprise/server/remote_execution/container/container.go
```diff
@@ -401,69 +410,157 @@ func GetPullCredentials(env environment.Env, props *platform.Properties) (PullCr
...
 }
 
 func (t *TracedCommandContainer) Remove(ctx context.Context) error {
 	ctx, span := tracing.StartSpan(ctx, trace.WithAttributes(t.implAttr))
 	defer span.End()
+
+	log.Infof("Remove()")
+
+	// Get an *exclusive* lock here to ensure any other pending operations are
+	// completed before we remove.
```
tylerwilliams, line 530: nit: s/pending/concurrent/g

tylerwilliams (approved):

enterprise/server/remote_execution/container/container.go
```diff
@@ -401,69 +410,157 @@ func GetPullCredentials(env environment.Env, props *platform.Properties) (PullCr
...
+	}
+
 	return t.Delegate.Create(ctx, workingDir)
 }
 
 func (t *TracedCommandContainer) Exec(ctx context.Context, command *repb.Command, opts *Stdio) *interfaces.CommandResult {
 	ctx, span := tracing.StartSpan(ctx, trace.WithAttributes(t.implAttr))
 	defer span.End()
+
+	log.Infof("Exec()...")
```
maggie-lou, line 485: Did you intend to leave in these debug logs?

enterprise/server/remote_execution/container/container.go
```diff
@@ -401,69 +410,157 @@ func GetPullCredentials(env environment.Env, props *platform.Properties) (PullCr
...
+	}
+
 	return t.Delegate.Pause(ctx)
 }
 
 func (t *TracedCommandContainer) Remove(ctx context.Context) error {
 	ctx, span := tracing.StartSpan(ctx, trace.WithAttributes(t.implAttr))
 	defer span.End()
+
+	log.Infof("Remove()")
```
maggie-lou, line 527: Here too?

maggie-lou (approved):

enterprise/server/remote_execution/container/container.go
bduffany: woops, thanks for catching this!

enterprise/server/remote_execution/container/container.go
bduffany: removed

enterprise/server/remote_execution/container/container.go
bduffany: sg, done

PR #4952, opened 2023-10-04
Use total RAM instead of free RAM for resource estimation limits

siggisim:
**Related issues**: N/A

### DIFF
--- a/server/resources/resources.go
+++ b/server/resources/resources.go
@@ -62,7 +62,7 @@ func setSysRAMBytes() {
 	}
 	mem := gosigar.Mem{}
 	mem.Get()
-	allocatedRAMBytes = int64(mem.ActualFree)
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):

server/resources/resources.go
```diff
@@ -62,7 +62,7 @@ func setSysRAMBytes() {
 	}
 	mem := gosigar.Mem{}
 	mem.Get()
-	allocatedRAMBytes = int64(mem.ActualFree)
+	allocatedRAMBytes = int64(mem.Total)
```
bduffany, line 65: this LGTM, but a slightly better default for darwin might involve subtracting the current `wired memory` usage. That value seems to hover at around 1.5 - 2GB and can't be swapped to disk unlike other types of memory.

bduffany (approved):

PR #4958, opened 2023-10-05
Mark fc blockio test as manual, instead of commenting out

maggie-lou:
This will make it easier to continue testing with the target even if we still don't want it automatically enabled on CI

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/BUILD
+++ b/enterprise/server/remote_execution/containers/firecracker/BUILD
@@ -122,62 +122,63 @@ go_test(
 #
 # TODO(bduffany): once blockio is fully enabled, set blockio flag(s) on
 # firecracker_test and remove this one.
-#go_test(
-#    name = "firecracker_test_blockio",
-#    timeout = "long",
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/BUILD
```diff
@@ -122,62 +122,63 @@ go_test(
...
+go_test(
+    name = "firecracker_test_blockio",
+    timeout = "long",
+    srcs = ["firecracker_test.go"],
+    args = [
+        "--executor.firecracker_enable_vbd=true",
+        "--executor.firecracker_enable_merged_rootfs=true",
+        "--executor.firecracker_enable_uffd=true",
+        "--executor.enable_local_snapshot_sharing=true",
+        "--executor.enable_remote_snapshot_sharing=true",
```
bduffany, line 134: do you mind disabling this flag until we improve the Write performance? for me, the test takes 2X as long with this enabled. (5min vs 2min 30s)

bduffany (approved):

PR #4962, opened 2023-10-05
Revert "Revert "Use hermetic_cc_toolchain to statically link against musl" (#4953)"

tempoz:
Appended the username to the `/tmp/zig-cache` directory to prevent potential persmissions issues on shared machines.


tempoz:
* master
  * **PR #4962** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4962" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4962?utm_source=stack-comment).

tools/bazel
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
tylerwilliams, line 1: is this going to work with bb / bazelisk?

tools/bazel
```diff
@@ -0,0 +1,5 @@
...
+
+export HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX=/tmp/zig-cache-$USER
```
sluongng, line 3: I would like to avoid using tools/bazel if possible as it's not customizable/hard to customize.

Could we set this in `.bazelrc` to some default values, then let folks override it in their `user.bazelrc` with their own value?

tempoz: No, not really. It needs to be a per-user directory to avoid failures on multi-user machines, and the `.bazelrc` cannot do any substitution. I, too, wished to avoid `//bazel/tools` and searched for another solution for hours. If you know of one, let me know! I don't love this, either.

tools/bazel
tempoz: Yep, I tested with both!

tools/bazel
tylerwilliams: Could we patch the toolchain to use the directory we want?

Related but I still prefer a directory in {user}/.caches/bazel/zig-cache so that all build stuff is in one place and easily removed together rather than spread across the system. Can we use this dir?

tempoz: Notably, people can already customize this value in their `user.bazelrc` with `build --action_env HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX=/path/to/directory`. All this does is set a sane default for that environment variable.

tempoz: Fixed.

sluongng (commented):
I know there are a lot of in-flights changes.

Feel free to ignore my comments where they are outdated.
I would love to take a look after as well.

As mentioned, I cherry-picked a few hunks to my own branch over at  https://github.com/buildbuddy-io/buildbuddy/compare/master...sluongng/test-static and found everything to be working as expected. Great job!

.bazelrc
```diff
@@ -143,9 +207,24 @@ build:release-mac --config=release-shared
...
+build:auto-release-static -c opt
+
+# On multi-user machines, specify HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX in your env
+# to avoid permissions errors.
+build:static --action_env HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX
+build:static --incompatible_enable_cc_toolchain_resolution
+build:static --action_env BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+build:static --@io_bazel_rules_go//go/config:static=true
+build:static --copt=-fPIC
+build:static --action_env BUILDBUDDY_TOOLCHAIN_OVERRIDE_USE_EXTERNAL_CC=1
```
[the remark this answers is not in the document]
tempoz: Thank you! `repo_env` turned out to be exactly what I needed! Fixed ^-^

.bazelrc
```diff
@@ -143,9 +207,24 @@ build:release-mac --config=release-shared
...
+build:auto-release-static --config=remote-static
+build:auto-release-static --remote_instance_name=buildbuddy-io/buildbuddy/auto-release
+build:auto-release-static -c opt
+
+# On multi-user machines, specify HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX in your env
+# to avoid permissions errors.
+build:static --action_env HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX
+build:static --incompatible_enable_cc_toolchain_resolution
+build:static --action_env BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+build:static --@io_bazel_rules_go//go/config:static=true
```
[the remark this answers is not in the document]
tempoz: I think we only want to build one version of the binary at a time.

cli/testutil/testcli/testcli.go
```diff
@@ -99,7 +99,37 @@ func CombinedOutput(cmd *exec.Cmd) ([]byte, error) {
...
+
+register_toolchains(
+    "@zig_sdk//toolchain:linux_amd64_musl",
+    "@zig_sdk//toolchain:linux_arm64_musl",
+    "@zig_sdk//toolchain:darwin_amd64",
+    "@zig_sdk//toolchain:darwin_arm64",
+    "@zig_sdk//toolchain:windows_amd64",
+    "@zig_sdk//toolchain:windows_arm64",
+)
+`,
```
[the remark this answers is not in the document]
tempoz: `repo_env` fixed this.

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -195,10 +201,6 @@ func getTestEnv(ctx context.Context, t *testing.T, opts envOpts) *testenv.TestEn
 	fc.WaitForDirectoryScanToComplete()
 	env.SetFileCache(fc)
 
-	// Some tests need iptables which is in /usr/sbin.
-	err = os.Setenv("PATH", os.Getenv("PATH")+":/usr/sbin")
-	require.NoError(t, err)
-
```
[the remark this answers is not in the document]
tempoz: I was having a LOT of trouble with the firecracker tests and was trying to clean things up to make it easier to figure out what was happening. Agreed, these changes belong elsewhere.

enterprise/server/cmd/goinit/BUILD
```diff
@@ -25,6 +25,14 @@ go_library(
...
     name = "goinit",
     embed = [":goinit_lib"],
-    pure = "on",
-    static = "on",
+    # "pure" causes the init file to require an ELF interpreter which the kernel
+    # does not have access to if we are bulding statically.
+    pure = select({
+        "//:static": "auto",
+        "//conditions:default": "on",
+    }),
```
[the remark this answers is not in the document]
tempoz: The problem shows up if you use `--@io_bazel_rules_go//go/config:static=true`, in my experience. Not sure why we'd make a test to confirm that a build state we don't use doesn't work.

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -1282,7 +1284,7 @@ func TestFirecrackerRun_ReapOrphanedZombieProcess(t *testing.T) {
 		ActionWorkingDirectory: workDir,
 		VMConfiguration: &fcpb.VMConfiguration{
 			NumCpus:           1,
-			MemSizeMb:         2500,
+			MemSizeMb:         3500,
```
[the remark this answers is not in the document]
tempoz: This is from when I was trying to do PIE stuff and the race detector was running OOM; can probably be removed now.

enterprise/server/test/integration/ci_runner/ci_runner_test.go
```diff
@@ -131,9 +161,39 @@ actions:
...
+    "@zig_sdk//toolchain:linux_amd64_musl",
+    "@zig_sdk//toolchain:linux_arm64_musl",
+    "@zig_sdk//toolchain:darwin_amd64",
+    "@zig_sdk//toolchain:darwin_arm64",
+    "@zig_sdk//toolchain:windows_amd64",
+    "@zig_sdk//toolchain:windows_arm64",
+)
+`,
+		"BUILD":   `sh_binary(name = "exit", srcs = ["exit.sh"])`,
+		"exit.sh": `exit "$1"`,
```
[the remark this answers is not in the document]
tempoz: `repo_env` fixed this.

platforms/BUILD
```diff
@@ -23,3 +23,95 @@ config_setting(
         "@platforms//cpu:arm64",
     ],
 )
+
+load("@io_buildbuddy_buildbuddy_toolchain//:rules.bzl", "UBUNTU20_04_IMAGE")
```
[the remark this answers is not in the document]
tempoz: Done!

enterprise/server/cmd/goinit/BUILD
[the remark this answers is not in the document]
sluongng: Hmm, I was confused which line was which when I made the comment above.

I was thinking that we could have a test to help confirm that our binary was statically linked, like using `ldd <binary>` to verify the binary as we discussed in Slack.
This will help us prevent future regressions where our binary could accidentally become dynamically linked 🤔 .

It's probably just a "nice to have" nit and should not block this PR though. Please feel free to ignore.

.bazelrc
```diff
@@ -143,9 +207,24 @@ build:release-mac --config=release-shared
...
+build:static --copt=-fPIC
```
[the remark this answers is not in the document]
tempoz: This causes `store_test` to mysteriously time out, so I'm leaving it as-is.

platforms/BUILD
```diff
@@ -23,3 +26,91 @@ config_setting(
...
+platform(
+    name = "buildbuddy_rbe_darwin_arm64_with_clang",
+    constraint_values = ["@bazel_tools//tools/cpp:clang"],
+    parents = [":buildbuddy_rbe_darwin_arm64"],
+)
+
+platform(
+    name = "buildbuddy_rbe_darwin_arm64_default",
+    parents = [":buildbuddy_rbe_darwin_arm64"] if "//:static" else [":buildbuddy_rbe_darwin_arm64_with_clang"],
+)
```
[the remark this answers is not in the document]
tempoz: awesome, thanks! I incorporated this now.

server/util/redact/redact_test.go
```diff
@@ -110,7 +110,7 @@ func TestRedactMetadata_StructuredCommandLine(t *testing.T) {
 		{"some_url", "https://token@foo.com", "https://foo.com"},
 		{"remote_default_exec_properties", "container-registry-username=SECRET_USERNAME", "container-registry-username=<REDACTED>"},
 		{"remote_default_exec_properties", "container-registry-password=SECRET_PASSWORD", "container-registry-password=<REDACTED>"},
-		{"host_platform", "@buildbuddy_toolchain//:platform", "@buildbuddy_toolchain//:platform"},
+		{"host_platform", "//platforms:local_config_platform", "//platforms:local_config_platform"},
```
sluongng, line 113: this test could be left as-is.

I hope that we could upstream some changes in the `//platforms` dir to `@buildbuddy_toolchain` eventually.

terraform/eks-cluster/commands/run.sh
```diff
@@ -22,8 +22,8 @@ bazel build server \
   --remote_instance_name=$(date +%s) \
   --host_cpu=k8 --cpu=k8 \
   --crosstool_top=@buildbuddy_toolchain//:ubuntu1604_cc_toolchain_suite \
-  --host_platform=@buildbuddy_toolchain//:platform_linux \
-  --platforms=@buildbuddy_toolchain//:platform_linux \
+  --host_platform=//platforms:buildbuddy_rbe_linux_x86_64_default \
+  --platforms=//platforms:buildbuddy_rbe_linux_x86_64_default \
```
sluongng, line 26: Not only we don't need to update this, but we probably need to rework a lot of these flags for it to work correctly now that we are no longer using `cc_toolchain_suite` 🤔 

I would prefer to leave this out and address this in a separate PR.

tools/dev_qa.py
```diff
@@ -76,9 +76,9 @@
                 --bes_backend=remote.buildbuddy.dev \
                 --bes_results_url=https://app.buildbuddy.dev/invocation/ \
                 --remote_timeout=10m \
-                --extra_execution_platforms=@buildbuddy_toolchain//:platform \
-                --host_platform=@buildbuddy_toolchain//:platform \
-                --platforms=@buildbuddy_toolchain//:platform \
+                --extra_execution_platforms=//platforms:local_config_platform \
+                --host_platform=//platforms:local_config_platform \
+                --platforms=//platforms:local_config_platform \
```
sluongng, line 81: same problem here, we need to rework some of our flags usages (i.e. removing crosstool_top in favor of extra_toolchain).

Could you please leave this hunk out of your PR? I could take a look separately.

.bazelrc
```diff
@@ -37,9 +37,8 @@ build:remote-shared --remote_timeout=600
 build:remote-shared --remote_download_minimal
 build:remote-shared --jobs=100
 build:remote-shared --verbose_failures
-# TODO(sluongng): split these into separate configs for each target platform
-build:remote-shared --platforms=//platforms:linux_x86_64
-build:remote-shared --extra_execution_platforms=//platforms:linux_x86_64
+build:remote-shared --platforms=//platforms:local_config_platform
+build:remote-shared --extra_execution_platforms=//platforms:local_config_platform
```
sluongng, line 41: I don't think this should work for MacOS?

Effectively here, we were saying: if you gonna do a remote build, you _MUST_ be targeting linux x86_64. I set it up that way because:
1. Most of our executors are linux x86_64, we have a very limited number of macos executors
2. Our MacOS RBE does not work. There are a few problems with toolchains and Go SDK (see https://github.com/buildbuddy-io/buildbuddy/pull/5153)

In the future, I think we gonna do something like this
```
build:remote-shared --platforms=//platforms:linux_x86_64,//platforms:macos_arm64
build:remote-shared --extra_execution_platforms=//platforms:linux_x86_64,//platforms:macos_arm64
```

where we will be building multiple platforms remotely by default, instead of only a single platform that is compatible with the user's current host platform.

.bazelrc
```diff
@@ -1,5 +1,5 @@
 build --enable_platform_specific_config=true
-query --enable_platform_specific_config=true
+# query --enable_platform_specific_config=true
```
sluongng, line 2: any particular reason for this to be commented out?

.bazelrc
```diff
@@ -129,6 +128,13 @@ build:auto-release --config=remote
 build:auto-release --remote_instance_name=buildbuddy-io/buildbuddy/auto-release
 build:auto-release -c opt
 
+# On multi-user machines, specify HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX in your
+# repo_env in user.bazelrc to avoid permissions errors.
```
sluongng, line 132: We should add a note here to clarify that this currently only works with Linux x86_64 and not with MacOS.
It will help curious folks navigate these configs a bit easier rather than running into some of our un-supported use cases.

.bazelrc
```diff
@@ -129,6 +128,13 @@ build:auto-release --config=remote
...
+build:static --repo_env BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+build:static --extra_toolchains=@zig_sdk//toolchain:linux_amd64_musl,@zig_sdk//toolchain:linux_arm64_musl
```
sluongng, line 134: Did you have a chance to test this on an ARM64 machine?
If not, I will make a conservative call to leave it out until we can verify that it works.

.bazelrc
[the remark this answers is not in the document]
sluongng: Hmm could you please add a TODO there for us to revisit?

Setting a global PIC compiler flag feels redundant to me.

sluongng (commented):
This PR looks very good. Things are clean and easy to review.

I only have a few more nits left.
Afterward I think we should be able to merge and iterate incrementally.

BUILD
```diff
@@ -209,6 +208,11 @@ config_setting(
     values = {"define": "release=true"},
 )
 
+config_setting(
+    name = "static",
```
sluongng, line 212: Nit: could this be moved to under `platforms/configs`?

.bazelrc
tempoz: There's a bug in the bb cli where if there is no matching config for the platform, the query fails. We don't have linux-specific query configs, so it doesn't work. Doesn't belong in this PR, though.

.bazelrc
tempoz: kk, sounds good. Thanks for the explanation.

.bazelrc
tempoz: Good point.

.bazelrc
tempoz: Yeah, I did not. I just added this in case someone might find it useful, but it's fair that we're not currently supporting an arm64 static build.

BUILD
tempoz: It's not really platform-related, though? Like, obviously any config_setting can be in any publicly visible bazel file in any repository, so the answer to your question is yes, but the `static` go build setting is absolutely independent of the platform which makes putting it there kind of odd to me.

platforms/BUILD
```diff
@@ -14,31 +16,40 @@ platform(
     ],
     exec_properties = {
         "OSFamily": "Linux",
-        "container-image": "docker://gcr.io/flame-public/rbe-ubuntu20-04:latest",
+        "container-image": "docker://" + UBUNTU20_04_IMAGE,
```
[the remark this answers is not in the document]
tempoz: Good point!

terraform/eks-cluster/commands/run.sh
tempoz: 👍

tools/dev_qa.py
tempoz: 👍

server/util/redact/redact_test.go
tempoz: 👍

.bazelrc
```diff
@@ -129,6 +128,16 @@ build:auto-release --config=remote
 build:auto-release --remote_instance_name=buildbuddy-io/buildbuddy/auto-release
 build:auto-release -c opt
 
+# Static build configuration. MacOS is currently unsupported.
+# On multi-user machines, specify HERMETIC_CC_TOOLCHAIN_CACHE_PREFIX in your
+# repo_env in user.bazelrc to avoid permissions errors.
+build:static --repo_env BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+build:static --extra_toolchains=@zig_sdk//toolchain:linux_amd64_musl
+# amd64 is not yet tested, so we leave this toolchain commented out for now.
```
sluongng, line 136:
```suggestion
# arm64 is not yet tested, so we leave this toolchain commented out for now.
```

enterprise/server/cmd/goinit/BUILD
```diff
@@ -29,6 +29,14 @@ go_library(
...
+    # static = select({
+    #    "//:static": "on",
+    #    "//conditions:default": "auto",
+    # }),
```
sluongng, line 41: Hmm in case of NOT having `—config=static`, we would still want to produce the original binary with `static = “on”` here right? Or is it not needed at all?

platforms/BUILD
```diff
@@ -14,31 +17,40 @@ platform(
...
-    exec_properties = {
-        "Arch": "arm64",
-        "OSFamily": "darwin",
-    },
+    constraint_values = ["@platforms//cpu:arm64"],
+    parents = [":macos"],
+)
+
+platform(
+    name = "local_config_platform",
```
sluongng, line 53: Nit: don’t need to address this now but we should discuss when/how this could be used later for a potential better name for this. It’s fairly similar to `@local_config_platform//:host` with very subtle differences.

sluongng (approved):
Tiny nits left but LGTM overall.

enterprise/server/cmd/goinit/BUILD
tempoz: It is not needed and will, in fact, cause the firecracker tests to fail.

platforms/BUILD
tempoz: Sure, I named it that because it uses the `@local_config_platform` "package" to fill in its values, but happy to rename it when we upstream this stuff to buildbuddy-toolchain.

PR #4963, opened 2023-10-05
Clean up context_based_shutdown flag

bduffany:
This is now the default, and we don't explicitly enable it anymore.

**Related issues**: N/A


tylerwilliams (approved):
nice

PR #4964, opened 2023-10-05
WIP Simplify contexts

maggie-lou:
Right now in firecracker we create a second context for the VM - `vmCtx`. We also have several points in the code ([monitorVMContext](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/containers/firecracker/firecracker.go#L1853), [remove](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/containers/firecracker/firecracker.go#L2029)) where we manually cancel one context if the other context has been canceled. To simplify, we should just use one context everywhere.

Not only will this simplify things, but the monitorVMContext function was also causing a race condition.

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -412,10 +412,6 @@ type FirecrackerContainer struct {
 	mountWorkspaceFile bool
 
 	cleanupVethPair func(context.Context) error
-	vmCtx           context.Context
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #4964** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4964" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4964?utm_source=stack-comment).

PR #4970, opened 2023-10-06
app/invocation_tabs: highlight Execution while viewing action

sluongng:
"action" tab is a child tab that could be accessed via either "Cache"
tab (by clicking on the mnemonic of Cache Request), or via "Execution"
tab.

Let's default it to highlight the Execution tab while viewing "action"
details. If no RBE is enabled, highlight "Cache" tab instead.

**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation_tabs.tsx
+++ b/app/invocation/invocation_tabs.tsx
@@ -26,10 +26,15 @@ export type TabId =
   | "suggestions"
   | "raw"
   | "execution"
-  | "fetches"
-  | "action";
+  | "fetches";
 
 export function getTabId(tab: string): TabId {
+  if (tab === "#action") {
+    // Treat the action tab as the execution tab if executions are enabled,
+    // otherwise treat it as the cache tab.
+    return this.props.executionsEnabled ? "execution" : "cache";
+  }
+
   return (tab.substring(1) as TabId) || "all";
 }
 
--- a/app/invocation/invocation_tabs.tsx
+++ b/app/invocation/invocation_tabs.tsx
@@ -30,9 +30,8 @@ export type TabId =
 
 export function getTabId(tab: string): TabId {
   if (tab === "#action") {
-    // Treat the action tab as the execution tab if executions are enabled,
-    // otherwise treat it as the cache tab.
-    return this.props.executionsEnabled ? "execution" : "cache";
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_tabs.tsx
```diff
@@ -43,8 +43,12 @@ export function getActiveTab({ tab, role, denseMode }: TabsContext): TabId {
 
 export default class InvocationTabsComponent extends React.Component<InvocationTabsProps> {
   private renderTab(id: TabId, { label, href }: { label: string; href?: string }) {
+    const activeTab = getActiveTab(this.props);
```
bduffany, line 46: I think the correct fix here is to remove `action` from the `TabId` enum (since "action" is not actually a tab), then add this code at the start of `getTabId`:

```typescript
if (tab === "#action") {
  return "execution";
}
```

then you should be able to revert the changes to the TabsComponent here.

sluongng: Great suggestion. Sent a fix accordingly

I won't make the refactoring s/tab/hash/ because, in the propagated props, we are naming it `tab`. We could do it in a separate future PR though.

app/invocation/invocation_tabs.tsx
```diff
@@ -26,10 +26,15 @@ export type TabId =
...
   | "execution"
-  | "fetches"
-  | "action";
+  | "fetches";
 
 export function getTabId(tab: string): TabId {
+  if (tab === "#action") {
+    // Treat the action tab as the execution tab if executions are enabled,
+    // otherwise treat it as the cache tab.
+    return this.props.executionsEnabled ? "execution" : "cache";
```
bduffany, line 35: `this` won't work here since this is a top-level function and not a class method, need to plumb through the `executionsEnabled` prop. For simplicity, I think it's also fine to just return `"execution"` here and have no tabs highlighted when RBE is not enabled.

sluongng: Oh, I did not realize that we need it to be a class method 😮‍💨 

I pushed the fix.

bduffany (approved):

sluongng:
@bduffany removing `| "actions"` from TabId would cause a type check error in `invocation.tsx`, so I reverted that change for now to avoid additional refactoring.

PR #4974, opened 2023-10-06
[WIP] Add per-runner locks

maggie-lou:
Add a per-runner lock for executors.

Previously all runners in a pool were sharing one lock. Because certain operations (like container.Remove() or container.Pause()) are slow and we only had one lock, we would sometimes unlock before starting these operations, then relock, to not block the entire pool. However we still want some guarantees of atomicity on the runner-level - i.e. we don't want runner.Run() and runner.RemoveInBackground() to be interleaved, as this can cause race conditions.

**Related issues**: N/A


maggie-lou:
* master
  * **PR #4974** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4974" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/4974?utm_source=stack-comment).

PR #4975, opened 2023-10-06
Add popup auth flow behind a flag

siggisim:
Per https://developers.google.com/identity/oauth2/web/guides/use-code-model#initialize_a_code_client

> Key Point: For the best balance of usability and security using the Popup mode UX flow with Authorization code model is recommended.

I've also noticed that many (most?) well designed apps like Figma, etc use this popup flow instead of the redirect flow.

Finally, it makes the repo flow logic much simpler since you can keep state around while authenticating / linking different Github organizations.

Much of the popup logic is roughly based on: https://github.com/auth0/auth0-spa-js/blob/aa35732585113973d5481fb66ac0434ea7e90657/src/utils.ts#L86

Added behind a flag so we can test it out in dev and easily switch back to the redirect flow if we bump into issues.


app/auth/auth_service.ts
```diff
@@ -103,6 +104,10 @@ export class AuthService {
   }
 
   handleLoggedIn(response: user.GetUserResponse) {
+    if (window.opener) {
+      console.log("Posting success message to " + window.location.origin);
```
jdhollen, line 108: Remove?

app/errors/error_service.ts
```diff
@@ -21,6 +21,10 @@ export class ErrorService {
       return;
     }
     alertService.error(String(error));
+    if (window.opener) {
+      console.log("Posting error message to " + window.location.origin);
```
jdhollen, line 25: Same

app/auth/auth_service.ts
```diff
@@ -103,6 +104,10 @@ export class AuthService {
...
+      window.opener?.postMessage("", window.location.origin);
```
bduffany, line 109: nit

```suggestion
      window.opener.postMessage("", window.location.origin);
```

jdhollen (approved):

app/errors/error_service.ts
```diff
@@ -21,6 +21,10 @@ export class ErrorService {
...
+      window.opener?.postMessage(String(error), window.location.origin);
```
bduffany, line 26:
```suggestion
      window.opener.postMessage(String(error), window.location.origin);
```

app/util/popup.ts
```diff
@@ -0,0 +1,50 @@
+export default {
+  open: (url: string, timeout = 60, width = 400, height = 600) => {
```
bduffany, line 2: (I don't have much experience with popup windows)

Does the timeout mean the user only gets 1 minute to log in and then the popup closes? Wondering if we should make it a tad longer like 5 min or something

bduffany (approved):

app/util/popup.ts
siggisim: I took the 60s timeout from the Auth0 library, but I agree it feels a bit short. Bumped it up to 5 min.

app/errors/error_service.ts
siggisim: I removed the log statements per Jim's suggestions, so the if became useless - so I kept the question mark

app/errors/error_service.ts
siggisim: Done

app/auth/auth_service.ts
siggisim: I removed the log statements per Jim's suggestions, so the if became useless - so I kept the question mark

app/auth/auth_service.ts
siggisim: Initially thought these might be useful for debugging bad behaviors, but they don't really end up telling you much other than whether or not the window was spawned from a popup - which you could check yourself by printing `window.opener` in the console.

Done.

PR #4979, opened 2023-10-07
Update repo flow to use popups

siggisim:
This is still a bit of a work in progress, but this state is definitely better than what's currently live in dev.

This PR does a few things, happy to break it up if it's a pain to review:
- Switches to using popups instead of redirects for all linking / auth
  - This allows us to remove like 3 or 4 buttons that were needed for all sorts of edge cases
  - Moves the edge case handling to when either the create repo button is clicked or the deploy button is clicked
-  Adds some query parameters which allows us to render a nice title with an image if present
- Adds a cookie with the last github app installation id so we can automatically select it if it was just installed

Screenshot:
Screenshot 2023-10-06 at 5 12 51 PM

### DIFF
--- a/enterprise/app/repo/repo.css
+++ b/enterprise/app/repo/repo.css
@@ -1,6 +1,6 @@
 .create-repo-page {
-  font-family: system-ui, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto,
-  Oxygen, Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
+  font-family: system-ui, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell, Fira Sans,
+    Droid Sans, Helvetica Neue, sans-serif;
   display: flex;
   align-items: center;
   justify-content: center;
@@ -9,7 +9,7 @@
 }
 
 .create-repo-page .card {
-  box-shadow: 0 1px 3px 0 rgba(0,0,0,.1),0 1px 2px -1px rgba(0,0,0,.1);
+  box-shadow: 0 1px 3px 0 rgba(0, 0, 0, 0.1), 0 1px 2px -1px rgba(0, 0, 0, 0.1);
   border: 1px solid hsl(240 5.9% 90%);
 }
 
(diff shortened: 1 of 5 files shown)


enterprise/app/repo/repo.css
```diff
@@ -1,16 +1,24 @@
 .create-repo-page {
+  font-family: system-ui, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell, Fira Sans,
+    Droid Sans, Helvetica Neue, sans-serif;
```
bduffany, line 3: not sure if you wanted this to match the main app font, but we're using Open Sans currently

```suggestion
  font-family: "Open Sans", sans-serif;
```

enterprise/app/repo/repo.tsx
```diff
@@ -118,16 +131,57 @@ export default class RepoComponent extends React.Component<RepoComponentProps, R
     this.fetchSecrets();
   }
 
+  async loginAndLinkGithub() {
+    if (!this.props.user) {
+      await this.loginToGithub();
```
bduffany, line 136: if `loginToGithub` fails then it will continue to `linkGithubAccount`, because `loginToGithub` does `catch` on the promise. Is that OK?

enterprise/app/repo/repo.tsx
```diff
@@ -239,6 +304,16 @@ export default class RepoComponent extends React.Component<RepoComponentProps, R
   }
 
   async handleDeployClicked(repoResponse: repo.CreateRepoResponse) {
+    let isGCPDeploy = this.getSecrets().includes(gcpRefreshTokenKey);
+    let needsGCPLink =
+      isGCPDeploy && !this.state.secretsResponse?.secret.map((s) => s.name).includes(gcpRefreshTokenKey);
+
+    if (isGCPDeploy && needsGCPLink) {
+      await this.linkGoogleCloud()
```
bduffany, line 312: if there's an error then it will continue due to the `catch` - is that OK here too?

enterprise/app/repo/repo.css
siggisim: I'd like to try using this system-ui stack that we use for our marketing page. I messed around with making this the default for our whole app UI as well, but wanted to wait a bit and test it out before taking the leap.

enterprise/app/repo/repo.tsx
siggisim: Ah, I didn't quite know how catch interacted with await. I've updated this to remove the catches and added a single try/catch to protect the whole thing.

Side note, I think async/await is provides a such nice syntax for flattening deeply nested serial promises - but the try catch syntax makes it look so ugly :(

bduffany (approved):

PR #4981, opened 2023-10-09
Add an endpoint for fetching projects for linked gcp account

siggisim:
This will allow the user to pick which project they'd like to deploy into

### DIFF
--- a/enterprise/server/gcplink/gcplink.go
+++ b/enterprise/server/gcplink/gcplink.go
@@ -14,9 +14,11 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/keystore"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/util/cookie"
+	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 
 	ctxpb "github.com/buildbuddy-io/buildbuddy/proto/context"
+	gcpb "github.com/buildbuddy-io/buildbuddy/proto/gcp"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	skpb "github.com/buildbuddy-io/buildbuddy/proto/secrets"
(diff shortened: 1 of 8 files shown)


enterprise/server/gcplink/gcplink.go
```diff
@@ -146,3 +149,106 @@ func makeTokenExchangeRequest(refreshToken string) (string, error) {
 type accessTokenResponse struct {
 	AccessToken string `json:"access_token"`
 }
+
+func GetGCPProject(env environment.Env, ctx context.Context, request *gcpb.GetGCPProjectRequest) (*gcpb.GetGCPProjectResponse, error) {
+	groupID := ""
+	if reqCtx := requestcontext.ProtoRequestContextFromContext(ctx); reqCtx != nil {
+		groupID = reqCtx.GetGroupId()
+	}
```
tylerwilliams, line 157: is there a reason this isn't GetGroupID()? is it because these requests ignore the group picker

tylerwilliams (approved):

enterprise/server/gcplink/gcplink.go
siggisim: Na, I think I just cribbed it from the wrong place - will fix.

enterprise/server/gcplink/gcplink.go
```diff
@@ -146,3 +149,106 @@ func makeTokenExchangeRequest(refreshToken string) (string, error) {
...
+
+	secretService := env.GetSecretService()
+	if secretService == nil {
+		return nil, status.FailedPreconditionError("secret service not available")
+	}
+	envVars, err := secretService.GetSecretEnvVars(ctx, groupID)
+	if err != nil {
+		return nil, err
+	}
+	envVars, err = ExchangeRefreshTokenForAuthToken(ctx, envVars, true)
```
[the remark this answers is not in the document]
siggisim: Agreed, but I feel like the right way to do that would be to add a method to the SecretService to allow you to just fetch a single environment variable - but that felt a little out of the scope of this PR. I've added a TODO

enterprise/server/gcplink/gcplink.go
siggisim: Fixed

PR #4988, opened 2023-10-10
Retry LeaseTask on app shutdown

bduffany:
Implements the reconnection changes described in the proto docs added here: https://github.com/buildbuddy-io/buildbuddy/pull/4983

The executor changes are behind the flag `executor.enable_lease_reconnect` which is set to true by default. The app's behavior is gated by the client's reported value of this flag, as well as the app flag `remote_execution.lease_reconnect_grace_period` being greater than 0 (defaults to 1s)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -21,6 +21,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/perms"
+	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
 	"github.com/go-redis/redis/v8"
@@ -29,6 +30,7 @@ import (
 	"google.golang.org/grpc/peer"
 	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	remote_execution_config "github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/config"
@@ -43,12 +45,12 @@ var (
 	defaultPoolName              = flag.String("remote_execution.default_pool_name", "", "The default executor pool to use if one is not specified.")
 	sharedExecutorPoolGroupID    = flag.String("remote_execution.shared_executor_pool_group_id", "", "Group ID that owns the shared executor pool.")
 	requireExecutorAuthorization = flag.Bool("remote_execution.require_executor_authorization", false, "If true, executors connecting to this server must provide a valid executor API key.")
+	leaseInterval                = flag.Duration("remote_execution.lease_duration", 10*time.Second, "How long before a task lease must be renewed by the executor client.")
+	leaseGracePeriod             = flag.Duration("remote_execution.lease_grace_period", 10*time.Second, "How long to wait for the executor to renew the lease after the TTL duration has elapsed.")
+	leaseReconnectGracePeriod    = flag.Duration("remote_execution.lease_reconnect_grace_period", 1*time.Second, "How long to delay re-enqueued tasks in order to allow the previous lease holder to renew its lease (following a server shutdown).")
 )
 
 const (
-	leaseInterval    = 10 * time.Second
-	leaseGracePeriod = 10 * time.Second
-
 	// This number controls how many reservations the scheduler will
 	// enqueue (across executor nodes) for each task. Typically this
 	// is 2 -- we've increased it to 3 because each executor will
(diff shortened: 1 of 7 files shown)


enterprise/server/test/integration/remote_execution/remote_execution_test.go
```diff
@@ -1759,7 +1759,98 @@ func TestAppShutdownDuringExecution(t *testing.T) {
...
+		})
+	}
+	err := eg.Wait()
+	require.NoError(t, err)
+
+	// Make sure we only ever started a single task execution per command (tasks
+	// should not be re-executed just because the app goes down).
+	tasksStartedCount := testmetrics.CounterValue(t, metrics.RemoteExecutionTasksStartedCount) - initialTasksStartedCount
+	require.Equal(t, float64(len(cmds)), tasksStartedCount, "no tasks should have been retried")
+}
```
bduffany, line 1785: this stuff was not actually added, it's just identical to the last few lines of code in `TestAppShutdownDuringExecution_LeaseTaskRetried`

vadimberezniker (commented):
Change looks great. 

Does the retry need to be limited to app shutdown? It seems like we should be able to use this approach if the connection breaks between executor and app for any reason. I'll grant you that app shutdown is the most common cause for issues, but if it's not much work it seems like it would be good to handle disconnects as a general problem.

bduffany:
> Does the retry need to be limited to app shutdown? It seems like we should be able to use this approach if the connection breaks between executor and app for any reason. I'll grant you that app shutdown is the most common cause for issues, but if it's not much work it seems like it would be good to handle disconnects as a general problem.

Mind if we do that in a separate PR? It will take a bit of refactoring, because the general disconnect case is different from the shutdown case - in the general case, we fail to ping the server (so the error handling logic is significantly different), vs in the shutdown case we successfully ping the server and then disconnect and reattach the stream on our own terms.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1443,7 +1495,18 @@ func (s *SchedulerServer) LeaseTask(stream scpb.Scheduler_LeaseTaskServer) error
 			}
 		}
 
+		// If shutting down, generate a reconnect token. This indicates to the
```
vadimberezniker, line 1498: Instead of handing out this token at shut down time you could generate a "lease token" at the time the lease is obtained and use it in a similar manner. This matters if you plan to generalize this to non-shutdown cases since you won't have the opportunity to send a "reconnect" token in those cases.

vadimberezniker (approved):

enterprise/server/scheduling/task_leaser/task_leaser.go
```diff
@@ -42,20 +45,51 @@ func NewTaskLeaser(env environment.Env, executorID string, taskID string) *TaskL
...
+			return nil, err
+		}
+		if rsp.GetReconnectToken() == "" {
+			break
+		}
+		// If the reconnect token is set, the ping was successful but the server
+		// is shutting down. To transfer the lease to a different app: terminate
+		// the current lease stream, begin a new lease, and re-send the ping but
+		// with the reconnect token set. The app guarantees that our retry will
+		// succeed as long as we reconnect within the grace period.
```
tylerwilliams, line 75: comments <= 80 chars

tylerwilliams (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
tylerwilliams: ah good point

tylerwilliams:
> > Does the retry need to be limited to app shutdown? It seems like we should be able to use this approach if the connection breaks between executor and app for any reason. I'll grant you that app shutdown is the most common cause for issues, but if it's not much work it seems like it would be good to handle disconnects as a general problem.
> 
> It doesn't need to be limited to app shutdown, but mind if we do that in a separate PR? It will take a bit of refactoring (and ideally adding a different type of test), because the general disconnect case is different from the shutdown case - in the general case, either the `stream.Send()` or `stream.Recv()` fails (on the executor), vs in the shutdown case we roundtrip the req successfully but then disconnect (on our own terms) and reattach the stream with a reconnect_token set.

Maybe it is better to handle all disconnects the same way?

enterprise/server/scheduling/scheduler_server/scheduler_server.go
bduffany: Done, except I kept the name `reconnect_token` for clarity (I don't think this will be useful for anything other than reconnecting the same lease?).

Also, we can no longer use `token != ""` as a signal that the server is shutting down, so I changed things so the server returns an explicit shutdown error instead, which the client checks for when `Recv()` fails.

bduffany:
> Maybe it is better to handle all disconnects the same way?

Now that I'm having the server return an error instead of using `reconnect_token` as the signal that the server is shutting down, this is a little more doable. Will ping once I add the commit for that

bduffany:
OK, now handling all disconnects the same way. Whenever we get an `Unavailable` error while pinging the server, we retry, but only if the server sent us a `reconnect_token`.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
bduffany: Slight modification to my last comment - the client no longer explicitly checks for the shutdown error, and now retries whenever it sees any `Unavailable` error. We now return an `Unavailable` error when the server is shutting down, and we might also see `Unavailable` in some other disconnect scenarios as well.

vadimberezniker (approved):

PR #4989, opened 2023-10-10
Filter out any window messages that aren't from buildbuddy

siggisim:
This prevents plugins like [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi/related) from unintentionally posting window messages that interfere with the flow.


tempoz (approved):
LGTM!

PR #4990, opened 2023-10-10
cli: avoid setting empty API key header in upload/download commands

bduffany:
**Related issues**: N/A

### DIFF
--- a/cli/download/download.go
+++ b/cli/download/download.go
@@ -107,7 +107,7 @@ func downloadFile(uri string) error {
 	}
 
 	ctx := context.Background()
-	if apiKey, err := storage.ReadRepoConfig("api-key"); err == nil {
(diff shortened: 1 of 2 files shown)


cli/download/download.go
```diff
@@ -107,7 +107,7 @@ func downloadFile(uri string) error {
 	}
 
 	ctx := context.Background()
-	if apiKey, err := storage.ReadRepoConfig("api-key"); err == nil {
+	if apiKey, _ := storage.ReadRepoConfig("api-key"); apiKey != "" {
```
tylerwilliams, line 110: Seems like it'd be more correct to check err != nil && str != ""

bduffany: done

tylerwilliams (approved):

PR #4991, opened 2023-10-10
Switch from XHR to Fetch API

bduffany:
This is in preparation for switching the timing profile to use a streaming HTTP request for loading and parsing the JSON profile. XHR supports streaming but it's very ugly.

* Switch from XHR to Fetch
* Remove `storedEncoding` as a method param (since it's pretty niche), and instead pass the stored-encoding header directly via the standard `RequestInit` object
* Improve type safety of fetch() method - instead of `Promise<any>`, returns `Promise<ArrayBuffer>` when setting request type to `arraybuffer`, `Promise<string>` when setting request type to `"text"`, etc.

**Related issues**: N/A


siggisim (approved):

bduffany:
(Will wait to submit this until after tomorrow's release)

siggisim:
Maybe this will make it easier for us to one day support streaming protolet RPCs 🤞

PR #5001, opened 2023-10-11
Fix URL weirdness when entering impersonation.

vadimberezniker:
The navigateTo call was going through the access check which tried to manipulate the URL prior to impersonation. Instead, redirect directly to the target URL.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2780

**Related issues**: N/A

### DIFF
--- a/app/auth/auth_service.ts
+++ b/app/auth/auth_service.ts
@@ -224,10 +224,17 @@ export class AuthService {
   }
 
   // Enters impersonation for the given group, which may either be a group ID or a URL identifier.
-  async enterImpersonationMode(query: string) {
+  async enterImpersonationMode(query: string, { redirectUrl = ""}: { redirectUrl?: string} = {}) {
     const request = grp.GetGroupRequest.create(query.startsWith("GR") ? { groupId: query } : { urlIdentifier: query });
     const response = await rpc_service.service.getGroup(request);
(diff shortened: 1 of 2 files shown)


app/auth/auth_service.ts
```diff
@@ -224,10 +224,17 @@ export class AuthService {
...
+  async enterImpersonationMode(query: string, { redirectUrl = ""}: { redirectUrl?: string} = {}) {
     const request = grp.GetGroupRequest.create(query.startsWith("GR") ? { groupId: query } : { urlIdentifier: query });
     const response = await rpc_service.service.getGroup(request);
     this.setCookie(IMPERSONATING_GROUP_ID_COOKIE, response.id, { maxAge: 0 });
+
+    // If we have an explicit redirect URL, navigate there directly.
+    if (redirectUrl) {
+      window.location.href = redirectUrl
+      return
+    }
```
bduffany, line 236: does this play nicely with the subdomain logic below? (should redirectUrl be modified to have the correct subdomain?)

bduffany (approved):

app/auth/auth_service.ts
vadimberezniker: The destination subdomain should only be different when "go to org" is used in which case redirectUrl would not be set.

PR #5002, opened 2023-10-11
Refactor task_router and add a new routing property that routes actions to executors that most recently executed similar actions (based on the first bazel output)

iain-macdonald:
**Related issues**: [222](https://github.com/buildbuddy-io/buildbuddy-internal/issues/222)

### DIFF
--- a/enterprise/server/scheduling/task_router/task_router.go
+++ b/enterprise/server/scheduling/task_router/task_router.go
@@ -2,8 +2,6 @@ package task_router
 
 import (
 	"context"
-	"crypto/sha256"
-	"fmt"
 	"math/rand"
 	"strings"
 	"time"
@@ -87,7 +85,6 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
 	params := getRoutingParams(ctx, tr.env, cmd, remoteInstanceName)
 	strategy := tr.selectRouter(params)
 	if strategy == nil {
-		log.Info("no routing strategy applied to the routing params")
 		return nodes
 	}
 
@@ -142,7 +139,6 @@ func (tr *taskRouter) MarkComplete(ctx context.Context, cmd *repb.Command, remot
 	params := getRoutingParams(ctx, tr.env, cmd, remoteInstanceName)
 	strategy := tr.selectRouter(params)
 	if strategy == nil {
-		log.Info("no routing strategy applied to the routing params")
 		return
 	}
 	preferredNodeLimit, routingKey, err := strategy.RoutingInfo(params)
@@ -249,7 +245,7 @@ func (runnerRecycler) routingKey(params routingParams) (string, error) {
 	if err != nil {
 		return "", status.InternalErrorf("failed to marshal Command: %s", err)
 	}
-	parts = append(parts, fmt.Sprintf("%x", sha256.Sum256(b)))
+	parts = append(parts, hash.Bytes(b))
 
 	return strings.Join(parts, "/"), nil
 }
@@ -260,10 +256,18 @@ func (s runnerRecycler) RoutingInfo(params routingParams) (int, string, error) {
 	return nodeLimit, key, err
 }
 
-// affinityRouter attempts to route execution actions to execuction nodes that
-// have some affinity with the provided routing parameters. That affinity is
-// determined through a combination of platform properties and the first
-// bazel-request output, which should uniquely identify an action.
+// affinityRouter generates Redis routing keys based on:
+//   - remoteInstanceName
+//   - groupID
+//   - platform properties
+//   - and the name of the first action output
+//
+// Because only a single action can generate a given output in Bazel, this key
+// uniquely identifies an action and is stable even if the action's inputs
+// change. The intent of using this routing key is to route successive actions
+// whose inputs have changed to nodes which previously executed that action to
+// increase the local-cache hitrate, as it's likely that for large actions most
+// of the input tree is unchanged.
 type affinityRouter struct{}
 
 func (affinityRouter) Applies(params routingParams) bool {
@@ -290,7 +294,7 @@ func (affinityRouter) routingKey(params routingParams) (string, error) {
 	if err != nil {
 		return "", status.InternalErrorf("failed to marshal Command: %s", err)
 	}
-	parts = append(parts, fmt.Sprintf("%x", sha256.Sum256(b)))
+	parts = append(parts, hash.Bytes(b))
 
 	// Add the first output as the final part of the routing key. This should
 	// uniquely identify a bazel action and is an attempt to route actions to
@@ -300,7 +304,7 @@ func (affinityRouter) routingKey(params routingParams) (string, error) {
 	if firstOutput == "" {
 		return "", status.InternalError("routing key requested for action with no outputs")
 	}
-	parts = append(parts, fmt.Sprintf("%x", hash.String(firstOutput)))
+	parts = append(parts, hash.String(firstOutput))
 
 	return strings.Join(parts, "/"), nil
 }
--- a/enterprise/server/scheduling/task_router/task_router.go
+++ b/enterprise/server/scheduling/task_router/task_router.go
@@ -267,11 +267,8 @@ func (s runnerRecycler) RoutingInfo(params routingParams) (int, string, error) {
 type affinityRouter struct{}
 
 func (affinityRouter) Applies(params routingParams) bool {
-	if !platform.IsTrue(platform.FindValue(params.cmd.GetPlatform(), platform.OutputAffinityRoutingPropertyName)) {
-		return false
-	}
-	o, err := getFirstOutput(params.cmd)
-	return o != "" && err == nil
+	return platform.IsTrue(platform.FindValue(params.cmd.GetPlatform(), platform.AffinityRoutingPropertyName)) &&
+		getFirstOutput(params.cmd) != ""
 }
 
 func (affinityRouter) preferredNodeLimit(params routingParams) int {
@@ -299,9 +296,9 @@ func (affinityRouter) routingKey(params routingParams) (string, error) {
 	// uniquely identify a bazel action and is an attempt to route actions to
 	// executor nodes that are warmed up (with inputs and OCI images) for this
 	// action.
-	firstOutput, err := getFirstOutput(params.cmd)
-	if err != nil {
-		return "", err
+	firstOutput := getFirstOutput(params.cmd)
+	if firstOutput == "" {
+		return "", status.InternalError("routing key requested for action with no outputs")
 	}
 	parts = append(parts, fmt.Sprintf("%x", hash.String(firstOutput)))
 
@@ -314,13 +311,13 @@ func (s affinityRouter) RoutingInfo(params routingParams) (int, string, error) {
 	return nodeLimit, key, err
 }
 
-func getFirstOutput(cmd *repb.Command) (string, error) {
+func getFirstOutput(cmd *repb.Command) string {
 	if len(cmd.OutputPaths) > 0 && cmd.OutputPaths[0] != "" {
-		return cmd.OutputPaths[0], nil
+		return cmd.OutputPaths[0]
 	} else if len(cmd.OutputFiles) > 0 && cmd.OutputFiles[0] != "" {
-		return cmd.OutputFiles[0], nil
+		return cmd.OutputFiles[0]
 	} else if len(cmd.OutputDirectories) > 0 && cmd.OutputDirectories[0] != "" {
-		return cmd.OutputDirectories[0], nil
+		return cmd.OutputDirectories[0]
 	}
-	return "", status.InternalError("routing key requested for action with no outputs")
+	return ""
 }
--- a/enterprise/server/scheduling/task_router/task_router.go
+++ b/enterprise/server/scheduling/task_router/task_router.go
@@ -2,6 +2,7 @@ package task_router
 
 import (
 	"context"
+	"flag"
 	"math/rand"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -61,9 +64,11 @@ func New(env environment.Env) (interfaces.TaskRouter, error) {
 	if rdb == nil {
 		return nil, status.FailedPreconditionError("Redis is required for task router")
 	}
+	strategies := []routingStrategy{runnerRecyclingRoutingStrategy{}, outputAffinityRoutingStrategy{}}
```
tylerwilliams, line 67: tbd if we need multiple of these, but maybe it's nice so we can have one for workflows and one for other actions?

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -128,61 +136,118 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+	return b
+}
+
+// A routingStrategy encapsulates a strategy for performing task routing based
+// on a set of provided routingParams. Currently this interface is somewhat
+// coupled with the Redis reading and writing logic in the task_router, but
+// that could be uncoupled in the future when other routing strategies that
+// arent Redis-list-based are added.
+type routingStrategy interface {
+
```
tylerwilliams, line 214: nit: rm newline

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -128,61 +136,118 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+	}
+	return b
+}
+
+// A routingStrategy encapsulates a strategy for performing task routing based
+// on a set of provided routingParams. Currently this interface is somewhat
+// coupled with the Redis reading and writing logic in the task_router, but
+// that could be uncoupled in the future when other routing strategies that
+// arent Redis-list-based are added.
+type routingStrategy interface {
```
tylerwilliams, line 213: I think usually you'd name this something with an -er, like Writer, Reader, Router, or whatever. Optionally, you can prefix with I, like IRouter or something.

I'm hesitant to say this is wrong, but I've most commonly seen interfaces be public and have public methods, so `type Router interface {` with `Applies() bool` and `Info()`...

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -128,61 +136,118 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+	// provided routing parameters. The preferredNodeLimit is the number of
+	// preferred executor nodes that should be used, and the routing Key is the
+	// Redis key where the list of preferred executor nodes are stored.
+	routingInfo(params routingParams) (int, string, error)
+}
+
+// The runnerRecyclingRoutingStrategy is a routing strategy that attempts to
+// "recycle" warm execution nodes when possible.
+type runnerRecyclingRoutingStrategy struct {
+}
```
tylerwilliams, line 229: nit: can move this up to the line above so you just have:

```type runnerRecyclingRoutingStrategy struct {}```

File: enterprise/server/scheduling/task_router/task_router.go [hunk 5]
```diff
@@ -128,61 +136,118 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+	// Returns the routing info (preferredNodeLimit and routingKey) for the
+	// provided routing parameters. The preferredNodeLimit is the number of
+	// preferred executor nodes that should be used, and the routing Key is the
+	// Redis key where the list of preferred executor nodes are stored.
+	routingInfo(params routingParams) (int, string, error)
+}
+
+// The runnerRecyclingRoutingStrategy is a routing strategy that attempts to
+// "recycle" warm execution nodes when possible.
+type runnerRecyclingRoutingStrategy struct {
```
tylerwilliams, line 228: runnerRecyclingRoutingStrategy is long -- maybe just runnerRecycling? Similarly for OutputAffinity{}.

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -195,15 +260,66 @@ func (tr *taskRouter) routingKey(ctx context.Context, cmd *repb.Command, remoteI
...
+
+	platform := params.cmd.GetPlatform()
+	if platform == nil {
+		platform = &repb.Platform{}
+	}
+	b, err := proto.Marshal(platform)
+	if err != nil {
+		return "", status.InternalErrorf("failed to marshal Command: %s", err)
+	}
+	parts = append(parts, fmt.Sprintf("%x", sha256.Sum256(b)))
```
tylerwilliams, line 300: you can use hash.String() for this instead of sha256.Sum256

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -195,15 +260,66 @@ func (tr *taskRouter) routingKey(ctx context.Context, cmd *repb.Command, remoteI
...
+	firstOutput := ""
+	if len(params.cmd.OutputPaths) > 0 && params.cmd.OutputPaths[0] != "" {
+		firstOutput = params.cmd.OutputPaths[0]
+	} else if len(params.cmd.OutputFiles) > 0 && params.cmd.OutputFiles[0] != "" {
+		firstOutput = params.cmd.OutputFiles[0]
+	} else if len(params.cmd.OutputDirectories) > 0 && params.cmd.OutputDirectories[0] != "" {
+		firstOutput = params.cmd.OutputDirectories[0]
+	} else {
+		return "", status.InternalError("routing key requested for action with no outputs")
+	}
```
tylerwilliams, line 315: i'd move this to its own function `getFirstOutput(cmd *repb.Command) string {`

then you can simplify this code and the `applies()` function above can instead become:

`len(getFirstOutput(cmd)) > 0`

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -128,61 +136,118 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+	copy(out, nodes)
+	return out
+}
+
+func minInt(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
```
tylerwilliams, line 206: somewhat embarrassing to say but I think after go 1.21 this is a builtin now.

tylerwilliams:
originally was gonna say that it seems overkill to have this routing strategy selection logic, but I do think you're right we'll always have at least 2 strats: one for workflows and one for reg tasks, so maybe it's ok?

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -41,6 +41,9 @@ const (
 type taskRouter struct {
 	env environment.Env
 	rdb redis.UniversalClient
+
+	// The ordered list of routingStrategies to use for task routing.
```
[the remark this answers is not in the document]
iain-macdonald: Yeah, thought that was clear with "ordered list" but made it even more explicit.

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Per discussion, my thinking was that we'll probably want to at least experiment with other routing strategies in the future, so clarifying this logic is worth a shot.

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Ah, I was wondering about that. Changed.

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Hmmm, if that's more idiomatic then okay :-P

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Done

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: SG, changed to `runnerRecycler` and `affinityRouter`

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Done

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Whoops, thanks

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Good call, done

enterprise/server/test/integration/task_router/task_router_test.go
```diff
@@ -106,6 +112,113 @@ func TestTaskRouter_RankNodes_DefaultNodeLimit_ReturnsOnlyLatestNodeMarkedComple
 	require.Equal(t, executorID2, ranked[0].GetExecutorID())
 
 	requireNotAlwaysRanked(1, executorID1, t, router, ctx, cmd, instanceName)
+	requireNonSequential(t, ranked[1:])
+}
+
+func TestTaskRouter_RankNodes_AffinityRouting(t *testing.T) {
```
tylerwilliams, line 118: I know it's sort of obvious (after reviewing the code) but it'd be nice to add a test for the expected happy path which is:

1) action runs
2) action inputs change
3) action runs again

to ensure you're hitting the same node

(apologies if i'm just missing the test where this happens)

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -76,20 +84,23 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
 
-	if cmd == nil {
-		return nodes
-	}
-	preferredNodeLimit := getPreferredNodeLimit(cmd)
-	if preferredNodeLimit == 0 {
+	params := getRoutingParams(ctx, tr.env, cmd, remoteInstanceName)
+	strategy := tr.selectRouter(params)
+	if strategy == nil {
+		log.Info("no routing strategy applied to the routing params")
```
tylerwilliams, line 90: in prod we run with

```  log_level: "info"```

so this might log quite often (for all actions which neither are workflows or have your new prop set) -- is that right and what you want?

**Review by tylerwilliams — CHANGES_REQUESTED** [review]

File: enterprise/server/scheduling/task_router/task_router.go (hunk 11 above)
Comment (iain-macdonald) [reply enterprise/server/scheduling/task_router/task_router.go]: Oh yeah, I had this here from an earlier iteration where there was a fallback, but it doesn't make sense now so I removed it.

File: enterprise/server/scheduling/task_router/task_router.go [hunk 12]
```diff
@@ -128,61 +139,150 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
-		parts = append(parts, remoteInstanceName)
+	platform := params.cmd.GetPlatform()
+	if platform == nil {
+		platform = &repb.Platform{}
+	}
+	b, err := proto.Marshal(platform)
+	if err != nil {
+		return "", status.InternalErrorf("failed to marshal Command: %s", err)
 	}
+	parts = append(parts, fmt.Sprintf("%x", sha256.Sum256(b)))
```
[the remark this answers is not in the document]
iain-macdonald: Added `hash.Bytes()` (to prevent bytes -> string -> bytes conversion) and used that.

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -128,61 +139,150 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+	return strings.Join(parts, "/"), nil
+}
+
+func (s runnerRecycler) RoutingInfo(params routingParams) (int, string, error) {
+	nodeLimit := s.preferredNodeLimit(params)
+	key, err := s.routingKey(params)
+	return nodeLimit, key, err
+}
+
+// affinityRouter attempts to route execution actions to execuction nodes that
```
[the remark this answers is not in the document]
iain-macdonald: Superseded by below

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -128,61 +139,150 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+// have some affinity with the provided routing parameters. That affinity is
+// determined through a combination of platform properties and the first
+// bazel-request output, which should uniquely identify an action.
```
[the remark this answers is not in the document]
iain-macdonald: Your version's pretty good :-)  just made a couple minor tweaks

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -192,18 +292,32 @@ func (tr *taskRouter) routingKey(ctx context.Context, cmd *repb.Command, remoteI
...
 
+	// Add the first output as the final part of the routing key. This should
+	// uniquely identify a bazel action and is an attempt to route actions to
+	// executor nodes that are warmed up (with inputs and OCI images) for this
+	// action.
+	firstOutput := getFirstOutput(params.cmd)
+	if firstOutput == "" {
+		return "", status.InternalError("routing key requested for action with no outputs")
+	}
+	parts = append(parts, fmt.Sprintf("%x", hash.String(firstOutput)))
```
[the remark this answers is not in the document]
iain-macdonald: Whoops, fixed

enterprise/server/test/integration/task_router/task_router_test.go
iain-macdonald: Good call, expanded this test case to do that

iain-macdonald (commented):
Thanks for the comments, PTAL

tylerwilliams (approved):

tylerwilliams:
Can't wait to see this land :)

bduffany (approved):

PR #5003, opened 2023-10-11
Add a gcp project picker to the repo flow

siggisim:
Screenshot:
Screenshot 2023-10-11 at 2 24 28 PM

### DIFF
--- a/app/picker/picker.tsx
+++ b/app/picker/picker.tsx
@@ -1,17 +1,19 @@
 import { Search } from "lucide-react";
 import React from "react";
 import { Subscription } from "rxjs";
-import pickerService, { Picker } from "./picker_service";
+import Modal from "../components/modal/modal";
+import pickerService, { PickerModel } from "./picker_service";
 
 interface State {
-  picker?: Picker;
-  isVisible?: boolean;
+  picker?: PickerModel;
+  isVisible: boolean;
   search: string;
   selectedIndex: number;
 }
 
-export default class PickerComponent extends React.Component<{}, State> {
+export default class Picker extends React.Component<{}, State> {
   state: State = {
+    isVisible: false,
     search: "",
     selectedIndex: 0,
   };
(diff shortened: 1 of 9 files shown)


app/picker/picker.tsx
```diff
@@ -0,0 +1,103 @@
...
+import pickerService, { Picker } from "./picker_service";
+
+interface State {
+  picker?: Picker;
+  isVisible?: boolean;
+  search: string;
+  selectedIndex: number;
+}
+
+export default class PickerComponent extends React.Component<{}, State> {
```
bduffany, line 13: nit: the stuff under app/components doesn't use the "Component" suffix

bduffany (commented):
I think you can simplify this by using `<Modal>`:

```jsx
<Modal
    isVisible={this.state.isPickerVisible}
    onRequestClose={() => this.setState({isPickerVisible: false})>
  <Picker
      options={this.state.pickerOptions}
      onSelect={selection => { this.setState({ selection, isPickerVisible: false})} /> 
</Modal>
```

Benefits:
- `Picker` becomes more generally useful, e.g. we want to render it inline instead of it always rendering in a Modal
- No need for the global picker service
- No need to render `<Picker>` in the root component

siggisim:
> I think you can simplify this by having the Picker render inline rather than in a modal, and then instead using `<Modal>` from `app/components/modal/modal.tsx`
> 
> ```js
> <Modal
>   isVisible={this.state.isPickerVisible}
>   onRequestClose={() => this.setState({ isPickerVisible: false })} >
>     <Picker
>       options={this.state.pickerOptions}
>       onSelect={selection => { this.setState({ selection, isPickerVisible: false}); } /> 
> </Modal>
> ```
> 
> Benefits:
> 
> * Can get rid of the custom CSS for showing the backdrop behind the picker (Modal should already handle this)
> * Inherit other nice behaviors from `<Modal>` such as dismissing with Esc
> * `Picker` becomes more generally useful, e.g. we might have a form where it makes sense to render this inline in the form
> * No need for the global picker service which adds a bit of complexity / indirection
> * No need to render `<Picker>` in the root component

I've moved the Picker component to use a `Modal` instead of creating its own shade. That gets rid of the custom CSS and the nice modal behaviors like support for `Esc` (though I feel the focus behavior is a bit wonky - i.e. if you click on the shade when you've got another component focused, it doesn't seem to close the modal - it requires another click on the shade).

I like having the picker service, because it allows you to just say "have the user pick one of these options and give me the result in a promise" rather than having to build a whole fancy dom structure and manage state any time you want the user to pick something. Kind of like the old `let result = input("Enter your input")` api but a bit nicer. I'm down to separate this out into more reusable components once we have the need for more customization / reuse in other places.

bduffany (approved):

PR #5005, opened 2023-10-12
deprecate app.user_management_enabled flag

sluongng:
This has been set to true by default for a while now.
Let's clean it up.

**Related issues**: N/A

### DIFF
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -48,7 +48,7 @@ message FrontendConfig {
   // Whether or not the usage page is enabled.
   bool usage_enabled = 15;
 
-  reserved 16; // DEPRECATED user_management_enabled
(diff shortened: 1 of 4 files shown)


proto/config.proto
```diff
@@ -49,7 +49,8 @@ message FrontendConfig {
   bool usage_enabled = 15;
 
   // Whether or not user management is enabled.
-  bool user_management_enabled = 16;
+  // DEPRECATED: This is now always true.
+  bool user_management_enabled = 16 [deprecated = true];
```
bduffany, line 53: nit: this can be removed rather than deprecating.

```
reserved 16;
```

bduffany (approved):

bduffany (approved):

PR #5006, opened 2023-10-12
gossip: fix flaky and race test

sluongng:
It seems like the timeout on these tests could be a bit low when they
are scheduled on resource limitted Executors.

Upon crash, the test will also trigger race detection of accessing the
variable during `t.Fatalf()` while writing to it in an unfinished
goroutine. Avoid the race detection by acquiring the mutex prior to
`t.Fatalf()`.

**Related issues**: N/A

### DIFF
--- a/server/gossip/gossip_test.go
+++ b/server/gossip/gossip_test.go
@@ -177,8 +177,10 @@ func TestUserQuery(t *testing.T) {
 	}()
 
 	select {
-	case <-time.After(3 * time.Second):
+	case <-time.After(5 * time.Second):
+		mu.Lock()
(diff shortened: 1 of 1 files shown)


bduffany:
It's really surprising that `t.Fatalf` is not safe for concurrent access, do you have an example invocation showing the data race logs?

sluongng:
@bduffany yeah take a look at https://buildbuddy.buildbuddy.io/invocation/8f71ecae-b655-4e9b-a58e-204c34add0fe?target=%2F%2Fserver%2Fgossip%3Agossip_test&targetStatus=6

sluongng:
closing this for now since it's no longer relevant

PR #5014, opened 2023-10-12
Remove AC startup deletion code

tylerwilliams:
Remove AC startup deletion functionality and flag.


vadimberezniker (approved):

tylerwilliams:
V5 migration is complete.

PR #5017, opened 2023-10-12
Add active_key_version as a pebble option.

luluz66:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -197,7 +197,6 @@ type PebbleCache struct {
 	atimeUpdateThreshold time.Duration
 	atimeBufferSize      int
 	minEvictionAge       time.Duration
-	acEvictionEnabled    bool
 
 	activeKeyVersion int64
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -195,10 +197,12 @@ type PebbleCache struct {
 	atimeUpdateThreshold time.Duration
 	atimeBufferSize      int
 	minEvictionAge       time.Duration
+	acEvictionEnabled    bool
```
vadimberezniker, line 200: i think you accidentally brought this back to life

vadimberezniker (approved):

tylerwilliams (approved):
Should ActiveKeyVersion be an *int64 or int64? we have a mix in the Options struct (MaxSizeBytes vs AtimeBufferSize)

luluz66:
> Should ActiveKeyVersion be an *int64 or int64? we have a mix in the Options struct (MaxSizeBytes vs AtimeBufferSize)

We don't always explictly set it, so ActiveKeyVersion being an *int64 allows us not to always set it?

PR #5023, opened 2023-10-16
podman: support cgroup v2 for `podman.enable_stats`

bduffany:
This allows using `podman.enable_stats` (to measure task size) on Ubuntu 22.04+ (e.g. locally), since podman uses cgroup v2.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -909,23 +909,11 @@ func (c *podmanCommandContainer) readRawStats(ctx context.Context) (*repb.UsageS
 		// cgroup v2: /cpu.stat file contains a line like "usage_usec <N>" It
 		// contains other lines like user_usec, system_usec etc. but we just
 		// report the total for now.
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -896,9 +898,34 @@ func (c *podmanCommandContainer) readRawStats(ctx context.Context) (*repb.UsageS
...
+			if len(fields) != 2 || fields[0] != "usage_usec" {
+				continue
+			}
+			cpuUsec, err := strconv.ParseInt(fields[1], 10, 64)
+			if err != nil {
+				return nil, err
+			}
+			cpuNanos = cpuUsec * 1e3
+			break
+		}
```
iain-macdonald, line 928: WDYT about making this path return a NotFoundError (or whatever error you think is best) if the CPU usage isn't found, for slightly more consistency with the other path / easier debugging?

bduffany: good call, done

iain-macdonald (approved):

PR #5026, opened 2023-10-16
Fix UFFD / VBD context being cancelled too early

bduffany:
LoadSnapshot was calling `monitorVMContext` then passing the returned context to the UFFD / VBD handlers. However, this context is cancelled at the end of LoadSnapshot, so if UFFD / VBD need to do any remote chunk fetches they would fail with "context cancelled."

This PR moves the `monitorVMContext` after the UFFD / VBD setup so that their context can outlive the LoadSnapshot call.

The RemoteSnapshot test is failing because of this, but we didn't catch this because it's disabled currently.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -810,9 +810,6 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context) error {
 	c.vmCtx = vmCtx
 	c.cancelVmCtx = cancelVmCtx
 
-	ctx, cancel := c.monitorVMContext(ctx)
-	defer cancel()
-
 	var snapOpt fcclient.Opt
 	if *enableUFFD {
 		uffdType := fcclient.MemoryBackendType(fcmodels.MemoryBackendBackendTypeUffdPrivileged)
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -867,18 +864,27 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context) error {
...
+			return cause
+		}
+		return status.InternalErrorf("failed to start machine: %s", err)
 	}
 	c.machine = machine
 
 	if err := c.machine.ResumeVM(ctx); err != nil {
+		if cause := context.Cause(ctx); cause != nil {
+			return cause
+		}
```
bduffany, line 887: These aren't strictly needed but I realized while testing that these can also fail due to UFFD failures, and without the `context.Cause` check, they return a generic "context cancelled" error. This makes them return the UFFD error which caused the VM to crash.

Having to check context.Cause everywhere is really ugly though so it'd be nice to find a design where we don't have to do this.

enterprise/server/remote_execution/uffd/uffd.go
```diff
@@ -125,6 +125,9 @@ func (h *Handler) Start(ctx context.Context, socketPath string, memoryStore *cop
 
 	go func() {
 		h.handleErr = h.handle(ctx, memoryStore)
+		if h.handleErr != nil {
+			log.CtxErrorf(ctx, "UFFD handler failed: %s", h.handleErr)
+		}
```
bduffany, line 130: Added this logging back in case we forget to check `context.Cause()` somewhere where we should be.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -867,18 +864,27 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context) error {
 		return err
 	}
 
+	ctx, cancel := c.monitorVMContext(ctx)
```
maggie-lou, line 867: Could you add a comment that the context used to start the VM _should_ be different than the cancellable one used to monitor the context?

maggie-lou (approved):

PR #5027, opened 2023-10-16
Cache snapshot chunks concurrently

bduffany:
Speeds up Pause(), which is mostly relevant for unit tests. Locally, reduces the execution time of `TestFirecrackerSnapshotAndResume` from ~30s -> ~20s.

**Related issues**: N/A


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -505,43 +522,57 @@ func (l *FileCacheLoader) cacheCOW(ctx context.Context, name string, remoteInsta
 		},
 	}
 
-	dirtyChunkCount := 0
-	var dirtyBytes int64
+	eg, egCtx := errgroup.WithContext(ctx)
+	eg.SetLimit(4)
```
maggie-lou, line 526: Should we make this a flag?

PR #5039, opened 2023-10-16
Pebble Cache: delete in a queue

luluz66:
combine onRefresh and onEvict in approxlru.go

In pebble cache, onEvict will put the sample on a queue; and we delete the
entry in background. Before we delete from pebble, we will check the last access
time and only delete if it's not recently read.

**Related issues**: N/A

### DIFF
--- a/server/util/approxlru/approxlru.go
+++ b/server/util/approxlru/approxlru.go
@@ -42,18 +42,11 @@ func (s *Sample[T]) String() string {
 // OnEvict requests that the given key be evicted. The callback may return
 // skip=true to indicate that the given sample is no longer valid and should be
 // removed from the eviction pool.
-type OnEvict[T Key] func(ctx context.Context, sample *Sample[T]) (skip bool, err error)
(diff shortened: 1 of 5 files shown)


server/util/approxlru/approxlru.go
```diff
@@ -73,12 +66,12 @@ type LRU[T Key] struct {
...
 	onEvict                     OnEvict[T]
 	onSample                    OnSample[T]
-	onRefresh                   OnRefresh[T]
 	evictionResampleLatencyUsec prometheus.Observer
 	evictionEvictLatencyUsec    prometheus.Observer
 	limiter                     *rate.Limiter
 
-	samplePool []*Sample[T]
+	samplePool         []*Sample[T]
+	numEvictionWorkers int
```
tylerwilliams, line 74: is this used anywhere?

vadimberezniker (approved):

tylerwilliams (approved):

PR #5041, opened 2023-10-16
Remove unnecessary timeout causing flakes in docker test

bduffany:
We don't need the 10 second timeout - we really just want to wait for the container to create `output.txt` which we're already doing via an explicit poll on that file. The 10 second timeout will just cause the test to be flaky if the machine is under high load.

**Related issues**: N/A


vadimberezniker (approved):

vadimberezniker:
Nice branch ;)

PR #5044, opened 2023-10-17
Use vmCtx for COW cache requests

bduffany:
COW IO can happen outside of the task ctx scope, so make sure we use vmCtx for COW cache requests.

**Related issues**: N/A


maggie-lou:
Is there any way we'd be able to recreate this with a unit test? Maybe if we sleep between VM operations or something, so the VM is still running in the background?

maggie-lou (approved):

bduffany:
> Is there any way we'd be able to recreate this with a unit test? Maybe if we sleep between VM operations or something, so the VM is still running in the background?

I think so - will merge this for now so we can get dev workflows back to a good state, then work on that in a follow up

PR #5047, opened 2023-10-17
[SS] If snapshot sharing is disabled, don't compute full digests for artifacts to save computation

maggie-lou:
This behavior was accidentally removed in #4809. We should probably reinstate it before snapshot sharing is officially enabled, because it's causing greater load on the executors

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -392,9 +392,31 @@ func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotK
 		ar.OutputFiles = append(ar.OutputFiles, out)
 		eg.Go(func() error {
 			ctx := egCtx
-			d, err := fileDigest(filePath)
-			if err != nil {
-				return err
+			var d *repb.Digest
+			if *snaputil.EnableLocalSnapshotSharing || *snaputil.EnableRemoteSnapshotSharing {
+				var err error
+				d, err = fileDigest(filePath)
+				if err != nil {
+					return err
+				}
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #5047** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5047" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5047?utm_source=stack-comment).

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -392,9 +392,30 @@ func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotK
 		ar.OutputFiles = append(ar.OutputFiles, out)
 		eg.Go(func() error {
 			ctx := egCtx
-			d, err := fileDigest(filePath)
-			if err != nil {
-				return err
+			var d *repb.Digest
+			if *snaputil.EnableLocalSnapshotSharing || *snaputil.EnableRemoteSnapshotSharing {
```
bduffany, line 396: uber nit: maybe slightly more direct to check `key.RunnerId`

```suggestion
			if key.GetRunnerId() == "" {
```

bduffany (approved):

enterprise/server/remote_execution/snaploader/snaploader.go
maggie-lou: Hopefully this gets cleaned up soon anyway, but I like the flags because the intention is really obvious

PR #5049, opened 2023-10-17
Move Stdio from container.go to commandutil.go

iain-macdonald:
Makes more sense to me, LMK if you disagree

**Related issues**: N/A


bduffany (approved):
Good call, makes more sense to me too

PR #5060, opened 2023-10-18
More container logic refactoring

iain-macdonald:
This PR:
- Renames `enterprise/server/util/container` to `enterprise/server/util/container/converter` because the logic in there is currently only used forconverting container images.
- Moves `PullCredentials` out of `container.go` and into `enterprise/server/util/container/credentials` and adds a few useful utility methods.
- Makes `sociartifactstore.go` use `Credentials` instead of the Credentials proto from `registry.proto`

**Related issues**: N/A


enterprise/server/util/container/converter/converter.go
```diff
@@ -1,4 +1,4 @@
-package container
+package converter
```
bduffany, line 1: `converter` as a package name feels slightly generic, wdyt about something like `ociconv` (similar to stdlib's `strconv`) or `imageconv`? (also nest directly under `util` since we usually have a flat package structure for the util dirs)

enterprise/server/util/container/credentials/credentials.go
```diff
@@ -0,0 +1,112 @@
+package credentials
```
bduffany, line 1: similar comment here - wdyt about renaming this to `oci`, `oci_image` `ociregistry`, or `imageregistry`? And maybe move it directly under `util`, since the util dir currently has a flat structure.

So `oci.Registry` / `oci.Credentials` instead of `credentials.Registry` / `credentials.Credentials`.

enterprise/server/util/container/converter/converter.go
iain-macdonald: good call, done

enterprise/server/util/container/credentials/credentials.go
iain-macdonald: good call, done

bduffany (approved):

PR #5064, opened 2023-10-19
Fix BES ip rule handling.

vadimberezniker:
Don't return error in IP rules service if auth failed for any reason.

Add separate IP auth check in BES handler.

**Related issues**: N/A


tylerwilliams (approved):
LGTM

Could add a test for this if you want to keep it from getting broken

PR #5068, opened 2023-10-20
move per-node dashboards out of main bb dashboard

luluz66:
**Related issues**: N/A

Move per-node dashboards out of main bb dashboard.

### DIFF
--- a/tools/metrics/grafana/dashboards/buildbuddy.json
+++ b/tools/metrics/grafana/dashboards/buildbuddy.json
@@ -24,7 +24,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 1,
+  "id": 8,
   "links": [],
   "liveNow": false,
   "panels": [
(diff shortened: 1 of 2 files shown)


vadimberezniker:
These are not mac-specific graphs. I suspect we broke the graphs when moving from regional -> global prometheus. I'd be fine moving them to a different dashboard, but I'm not sure about deleting them entirely.

luluz66:
> These are not mac-specific graphs. I suspect we broke the graphs when moving from regional -> global prometheus. I'd be fine moving them to a different dashboard, but I'm not sure about deleting them entirely.

I was reading https://github.com/buildbuddy-io/buildbuddy/pull/2526 and assumed that the memory usage was only added for mac. It looks like all these metrics need to be added to https://github.com/buildbuddy-io/buildbuddy-internal/blob/178c21aef19c3b10981940e90e2cc53588d77045/enterprise/deployment/templates/prometheus-global.yml#L17

luluz66:
> These are not mac-specific graphs. I suspect we broke the graphs when moving from regional -> global prometheus. I'd be fine moving them to a different dashboard, but I'm not sure about deleting them entirely.

Done.

PR #5070, opened 2023-10-20
Move toolchain registration into WORKSPACE

sluongng:
Register our toolchains inside WORKSPACE.
Then enable automatic cc toolchain resolution for auto selection based
on platform compatibility.

This let us get rid of the manual toolchain declaration when building with
`--config=remote`. Subsequently, we won't need to specify `--cpu` and
`--host_cpu` to aid cc_toolchain_suite detection.


sluongng:
The expected outcome here is from a MacOS laptop, you should be able to just do `bazel build --config=remote //...` and it will switch to the correct toolchain, correct setup etc...

siggisim (approved):

PR #5072, opened 2023-10-20
Fix error handling in lease reconnect

bduffany:
Unclear whether this was actually happening in practice, but I noticed while re-checking this code that we are returning the wrong error when there are no more retries due to the context being cancelled.

**Related issues**: N/A


enterprise/server/scheduling/task_leaser/task_leaser.go
```diff
@@ -93,7 +94,7 @@ func (t *TaskLeaser) pingServer(ctx context.Context) ([]byte, error) {
 			r = retry.DefaultWithContext(ctx)
 		}
 		if !r.Next() {
-			return nil, err
+			return nil, originalErr
```
bduffany, line 97: This was previously returning the err on line 86, which was always nil.

vadimberezniker (approved):

PR #5076, opened 2023-10-20
[FC] Eagerly fetch chunks in the cow store

maggie-lou:
If we access a snapshot chunk, use a background queue to eagerly fetch the next X contiguous chunks, anticipating that they will also be accessed.

**Some performance results:**
The performance tests build the buildbuddy repo from a snapshot that is cached in the remote cache, but nothing is cached locally in filecache. Every testing scenario was run 6 times. 

Baseline: (45.8s - 50.6s)
Best config - 8 concurrent threads, max 1000 fetches per sec: (24.9s - 27.2s)

**Some additional sample tests for context:**

4 concurrent threads, max 1000 fetches per sec: (25.7s - 27.6s)
4 concurrent threads, max 500 fetches per sec: (28.5 - 29.5s)
8 concurrent threads, max 100 fetches per sec: (37.5s - 40.2)
-  Additional threads not that helpful, because rate limiter is so low
4 concurrent threads, max 100 fetches per sec: (38.1 - 42.3)
4 concurrent threads, max 1000 per sec, 10 contiguous chunks: (27.5 - 30)
- All other tests are with 4 contiguous chunks
- Fetching more chunks each time leads to a slowdown in performance

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -2,6 +2,7 @@ package copy_on_write
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"io"
 	"os"
@@ -34,17 +35,14 @@ const (
 	// churning through stale data
 	eagerFetchChanSize = 100
 
-	maxEagerFetchesPerSec = 100
-
-	// Max number of goroutines allowed to run concurrently when eagerly
-	// fetching chunks
-	eagerFetchConcurrency = 4
-
 	// If we access a chunk, we will queue this number of contiguous chunks
 	// to be eagerly fetched in the background, anticipating that they will
 	// also be accessed
 	numChunksToEagerFetch = 4
 )
+
+var maxEagerFetchesPerSec = flag.Int("executor.snaploader_max_eager_fetches_per_sec", 1000, "Max number of chunks snaploader can eagerly fetch in the background per second.")
+var eagerFetchConcurrency = flag.Int("executor.snaploader_eager_fetch_concurrency", 8, "Max number of goroutines allowed to run concurrently when eagerly fetching chunks.")
 
 // COWStore To enable copy-on-write support for a file, it can be split into
 // chunks of equal size. Just before a chunk is first written to, the chunk is first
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #5076** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5076" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5076?utm_source=stack-comment).

bduffany (commented):
This implementation looks great.

Before submitting, I think it'd be good to do some performance testing, possibly with different tuning values (e.g. N=0 vs N=4 / concurrency=4 vs concurrency=8, etc.). Otherwise we're a bit blind as to whether the eager fetching provides a net benefit, or how things will change if we try to adjust these tuning values.

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -473,6 +516,73 @@ func (s *COWStore) initDirtyChunk(offset int64, size int64) (ogChunk *Mmap, newC
...
+func (s *COWStore) fetchChunk(offset int64) error {
+	s.mu.RLock()
+	c := s.chunks[offset]
+	s.mu.RUnlock()
+
+	// Skip holes
+	if c == nil {
+		return nil
+	}
+
```
bduffany, line 578: nit: maybe move the rateLimiter.Wait() after this line?

bduffany:
As a rough performance test, it might be useful to try running `enterprise/tools/vmstart` now that it supports remote snapshots. Some behavior I noticed was that the first LoadSnapshot took a long time (~30s), then each subsequent one got faster

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
maggie-lou: I'm wondering if the rate limiter could also help protect over-locking on `chunks`. I'm just going to leave it as is for now - probably wouldn't be a big difference either way

maggie-lou:
> This implementation looks great.
> 
> Before submitting, I think it'd be good to do some performance testing, possibly with different tuning values (e.g. N=0 vs N=4 / concurrency=4 vs concurrency=8, etc.). Otherwise we're a bit blind as to whether the eager fetching provides a net benefit, or how things will change if we try to adjust these tuning values.

I added some performance test results to the pr body and updated the config to match the results. Also made them configurable by flag

bduffany (approved):

PR #5078, opened 2023-10-20
[SS] Try to be more clever about not trying to cache chunks that should already be cached

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -568,7 +568,8 @@ func (l *FileCacheLoader) cacheCOW(ctx context.Context, name string, remoteInsta
 		tree.Root.Files = append(tree.Root.Files, fn)
 		eg.Go(func() error {
 			ctx := egCtx
-			if cow.Dirty(c.Offset) {
+			dirty := cow.Dirty(c.Offset)
+			if dirty {
 				chunkSize, err := c.SizeBytes()
 				if err != nil {
 					return status.WrapError(err, "dirty chunk size")
@@ -583,6 +584,13 @@ func (l *FileCacheLoader) cacheCOW(ctx context.Context, name string, remoteInsta
 				}
 			}
 
+			// If the chunk was mapped but the lazy digest is not set, it could not
+			// have been fetched remotely, because we use the lazy digest to do so.
+			// This means we must have generated the chunk from a local file
+			// Note: This has to be called before we call c.Digest(), which sets
+			// the lazy digest
+			chunkFetchedRemotely := c.Mapped() && c.LazyDigest() != nil
+
 			// Get or compute the digest.
 			d, err := c.Digest()
 			if err != nil {
@@ -590,7 +598,9 @@ func (l *FileCacheLoader) cacheCOW(ctx context.Context, name string, remoteInsta
 			}
 			fn.Digest = d
 
-			if c.Mapped() {
+			// Don't cache artifacts that should already be in the cache
+			shouldCache := !chunkFetchedRemotely || dirty
+			if shouldCache {
 				path := filepath.Join(cow.DataDir(), copy_on_write.ChunkName(c.Offset, cow.Dirty(c.Offset)))
 				if err := snaputil.Cache(ctx, l.env.GetFileCache(), l.env.GetByteStreamClient(), d, remoteInstanceName, path); err != nil {
 					return status.WrapError(err, "write chunk to cache")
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #5078** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5078" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5078?utm_source=stack-comment).

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -583,14 +584,23 @@ func (l *FileCacheLoader) cacheCOW(ctx context.Context, name string, remoteInsta
...
+
 			// Get or compute the digest.
 			d, err := c.Digest()
 			if err != nil {
 				return status.WrapError(err, "compute digest")
 			}
 			fn.Digest = d
 
-			if c.Mapped() {
+			// Don't cache artifacts that should already be in the cache
```
bduffany, line 601: I had started putting together a similar change earlier, but then remembered that we might want to cache the chunks just so that we can update the access time on the remote cache object.

Maybe there's some way we can explicitly update the access time if we think it should exist, and if it doesn't exist then fall back to re-uploading to remote cache?

bduffany: Oh... I guess this should be OK because when we fetch the remote snapshot tree, that updates the access time?

tylerwilliams:
Yup 👍🏼 On Oct 20, 2023, at 3:18 PM, Brandon Duffany ***@***.***> wrote:﻿
@bduffany commented on this pull request.



In enterprise/server/remote_execution/snaploader/snaploader.go:
>  			// Get or compute the digest.
 			d, err := c.Digest()
 			if err != nil {
 				return status.WrapError(err, "compute digest")
 			}
 			fn.Digest = d
 
-			if c.Mapped() {
+			// Don't cache artifacts that should already be in the cache

Oh... I guess this should be OK because when we fetch the remote snapshot tree, that updates the access time?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -583,14 +584,23 @@ func (l *FileCacheLoader) cacheCOW(ctx context.Context, name string, remoteInsta
 				}
 			}
 
+			// If the chunk was mapped but the lazy digest is not set, it could not
+			// have been fetched remotely, because we use the lazy digest to do so.
+			// This means we must have generated the chunk from a local file
+			// Note: This has to be called before we call c.Digest(), which sets
+			// the lazy digest
+			chunkFetchedRemotely := c.Mapped() && c.LazyDigest() != nil
```
bduffany, line 592: Shouldn't the lazyDigest be non-nil if we fetched the chunk from local filecache too?

also nit: maybe clearer/more explicit to add a function to the chunk like `WasFetchedRemotely()`

maggie-lou: Hmmm yeah. I guess we could hit a situation where the chunk is in the local cache but gets evicted remotely and we'd still want it to be uploaded the remote cache. 

I'll switch to your suggestion of marking where the chunk is from. We could use this data in a metric, to validate our hypothesis that most chunks should be cached locally

enterprise/server/remote_execution/snaploader/snaploader.go
maggie-lou: We might want to update the atime in the local cache. Though I don't know if this is worth the complexity, if we can always fall back to fetching from remote cache

enterprise/server/remote_execution/snaploader/snaploader.go
maggie-lou: If we fetched the chunk from local filecache, we still shouldn't need to re-cache it right? (The only thing we might want to do is update the atime)

Either way, I like your suggestion of explicitly marking where the chunk is from. We could use this data in a metric to validate our hypothesis that most chunks should be cached locally

bduffany: I was mostly confused by the `chunkFetchedRemotely` naming here, but yeah re-caching is also somewhat of a concern. I guess we don't really need to update the atime since GetTree should handle it, and we call that unconditionally?

maggie-lou: Actually I think we should be okay. If we read it from the local cache, that will update its atime. If we read it from the remote cache, we write it to the local cache

maggie-lou:
Replaced with #5134

PR #5094, opened 2023-10-23
Expand COW concurrency tests

bduffany:
* Add a `StoreTester` which allows reading and writing random ranges concurrently and also validates byte contents (rather than just checking whether the race detector is triggered)
* Use byte-range locking to avoid concurrent writes of overlapping byte ranges. In practice, the guest's block device driver will avoid doing these types of writes, because block device files do not guarantee atomicity for `read()` or `write()`. If we can assume that overlapping byte writes will never happen concurrently, we can acquire finer-grained locks (e.g. writes can lock per-chunk instead of locking the whole store)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/BUILD
+++ b/enterprise/server/remote_execution/copy_on_write/BUILD
@@ -33,6 +33,7 @@ go_test(
     srcs = ["copy_on_write_test.go"],
     deps = [
         ":copy_on_write",
+        "//enterprise/server/remote_execution/copy_on_write/cow_cgo_testutil",
         "//enterprise/server/remote_execution/snaputil",
         "//proto:remote_execution_go_proto",
         "//server/interfaces",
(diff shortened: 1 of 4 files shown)


maggie-lou:
"In practice, the guest's block device driver will avoid doing these types of writes" - Just curious, is this something you read about firecracker? Or is this just something you can assume for all block device drivers?

maggie-lou (approved):

bduffany:
> "In practice, the guest's block device driver will avoid doing these types of writes" - Just curious, is this something you read about firecracker? Or is this just something you can assume for all block device drivers?

I misspoke, I think it's technically the filesystem driver in the guest (e.g. ext4) that provides this guarantee, not the block device driver. Block devices don't have any guarantees about atomicity when reading/writing overlapping ranges. So, the filesystem driver has to order its reads and writes to the block device to prevent data corruption.

I am basing this mostly on my understanding though, so before I merge this I am going to run some tests to verify that this is actually the case in practice.

PR #5096, opened 2023-10-23
Proto changes for git ref-based snapshot matching

bduffany:
**Related issues**: N/A

### DIFF
--- a/proto/scheduler.proto
+++ b/proto/scheduler.proto
@@ -122,6 +122,39 @@ message SchedulingMetadata {
   string executor_group_id = 5;
   // Group ID of the user that issued the Execute request.
   string task_group_id = 6;
+
+  // Firecracker remote snapshot sharing options.
+  //
+  // These allow matching tasks to remote snapshots based on properties other
+  // than Platform properties, since the Platform is hashed as part of the
+  // routing key.
+  //
+  // In particular, we would like to match tasks to snapshots based on the
+  // desired git revision of the repo in the snapshot, but we don't want the git
+  // revision info as part of the Platform (since the revision changes
+  // frequently, e.g. PR branch name).
+  //
+  // If not set, tasks will not be matched to remote snapshots at all, and
+  // instead only be matched to snapshots in the executor's local cache.
+  RemoteSnapshotOptions remote_snapshot_options = 9;
+}
+
+message RemoteSnapshotOptions {
+  // Primary git ref associated with the snapshot.
+  //
+  // When reading the snapshot from remote cache, the snapshot associated with
+  // this ref will be preferred if it exists, otherwise the fallback ref will be
+  // used.
+  //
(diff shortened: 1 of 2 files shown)


proto/scheduler.proto
```diff
@@ -122,6 +122,26 @@ message SchedulingMetadata {
   string executor_group_id = 5;
   // Group ID of the user that issued the Execute request.
   string task_group_id = 6;
+
+  // Firecracker remote snapshot sharing options. This controls which snapshot
+  // the task prefers to use, as well as which snapshot it will update in the
+  // remote cache, if any. If not set, snapshots will only be shared locally,
+  // and will use the platform hash as the routing key.
+  RemoteSnapshotOptions remote_snapshot_options = 7;
```
bduffany, line 130: I originally considered putting these in platform props but then realized that the platform hash is part of the snapshot key and also messes with the task router, so it won't quite work without some ugly hacks.

I figure it will be cleaner to have this as scheduling metadata, which we can pass to the scheduler as a serialized proto via gRPC header in our /Execute req, then ultimately access it in the executor via the ExecutionTask.

wdyt @tylerwilliams @maggie-lou ?

proto/scheduler.proto
```diff
@@ -122,6 +122,39 @@ message SchedulingMetadata {
...
+  //
+  // When writing the snapshot back to cache, this ref (not the fallback_ref)
+  // will always be used as the ref part of the key.
+  //
+  // Example: "my-cool-branch".
+  string ref = 1;
+
+  // Optional fallback ref in case the preferred one is unavailable.
+  // Example: "main"
+  string fallback_ref = 2;
```
tylerwilliams, line 157: Can we send the snapshot key to look for directly rather than relying on the client to assemble this?

bduffany: Do you mean just make these opaque strings and have the client use them directly instead of caring that it has anything to do with git? If so, that seems reasonable.

I think we would still need a notion of a "preferred" key and a "fallback" key and the client needs to know how to treat these, right? i.e. something like:

- LoadSnapshot:
  - Try pulling the "preferred" snapshot (will always be equal to the pushed branch - either "main" or a PR branch)
    - If this fails, try the "fallback" key if it's set (will be unset on push to main, and set to "main" for PRs) 
      - If this fails, start cold
- SaveSnapshot:
  - Overwrite the "preferred" key (regardless of whether we pulled from the "preferred" or "fallback" key)

tylerwilliams: sorta yeah, basically having the client be dumb here is less error prone (we don't have to worry about client assembling the exact right cache key). Also maybe these should be digests or resourcenames since they are just pointers to files in the action cache?

bduffany: We can't easily construct the full snapshot key on the app, because it depends on the software versions available on the executor, e.g. the firecracker versions will be part of the key.

tylerwilliams (approved):

PR #5097, opened 2023-10-24
Support COW 🐄 🐄 🐄

tylerwilliams:


server/util/fastcopy/fastcopy_linux.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	if destErr != nil {
+		return destErr
+	}
+	reflinkWasSuccessful = true
+	return nil
+}
+
+func FastCopy(source, destination string) error {
+	if err := reflink(source, destination); err != nil {
```
siggisim, line 58: Do you think we should have a flag to enable reflink (since I feel like most filesystems on-prem folks use won't support it). 

That way there's no overhead added to the non-reflink case?

siggisim (approved):

PR #5098, opened 2023-10-24
Add test for reflink

bduffany:
* Add a test that mounts an XFS image and tests that a reflinked file modification doesn't affect the original
* defer close `destFile` (it would have been closed on GC so this is not a big deal)

**Related issues**: N/A

### DIFF
--- a/server/util/fastcopy/BUILD
+++ b/server/util/fastcopy/BUILD
@@ -27,9 +27,17 @@ go_library(
 go_test(
     name = "fastcopy_test",
     srcs = ["fastcopy_test.go"],
+    exec_properties = {
+        "test.workload-isolation-type": "firecracker",
+        "test.recycle-runner": "true",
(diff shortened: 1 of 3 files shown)


tylerwilliams:
https://github.com/golang/go/issues/24331

Has some context on why I did it that way but maybe it's okay?

bduffany:
> https://github.com/golang/go/issues/24331

The mentioned drawbacks of setting the file to "blocking mode" don't seem to apply here since we're opening and closing the file within this func and we don't have to care about messing with the caller's incoming file descriptors. I think that's maybe why [this impl](https://github.com/KarpelesLab/reflink/blob/84acd0aec6c1b253e1c50fdec5e1c9d514c0c58a/reflink_linux.go#L13) did it this way - it probably didn't want to mess with the incoming file descriptors by calling `Fd()`.

tylerwilliams (approved):

PR #5105, opened 2023-10-25
cli: add 'bb execute' command

bduffany:
Adds a `bb execute` command as a way to run ad-hoc commands/scripts at the command line using the remote execution API.

This is preferable e.g. to running an ad-hoc action with bazel via something like `sh_test`, since (a) it's cumbersome to set up a Bazel repo, (b) Bazel will (confusingly) spawn several actions (not just the test action), and (c) this has the overhead of Bazel itself + repository fetching etc. which makes the performance numbers not as reliable.

Example usage:

```
$ bb execute \
    --remote_header=x-buildbuddy-api-key=MY_API_KEY \
    --exec_properties=workload-isolation-type=firecracker \
    --exec_properties=recycle-runner=true \
    -- \
    bash -c '[ -e ~/warm ] && echo "WARM" || echo "COLD" ; touch ~/warm'
```

The above command prints "COLD" on the first run, then "WARM" for subsequent runs.

**Related issues**: N/A


cli/execute/execute.go
```diff
@@ -0,0 +1,138 @@
+package execute
+
+import (
+	"context"
+	"os"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/cli/arg"
+	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/rexec"
```
tylerwilliams, line 10: We have a single enterprise dep in cli -- let's remove it rather than adding another.

cli/execute/execute.go
```diff
@@ -0,0 +1,138 @@
...
+Args that modify execution can be placed before '--', and the command executable
+and arguments should come afterwards.
+
+Example of running a simple bash command:
+  $ bb execute -- bash -c 'echo "Hello world!"'
+
+Example of running a bash command with runner recycling:
+  $ bb execute --exec_properties=recycle-runner=true -- bash -c 'echo "Runner uptime:" $(uptime)'
+`
+)
```
tylerwilliams, line 58: nice comment :)

cli/execute/execute.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+	ctx = metadata.NewOutgoingContext(ctx, md)
+
+	conn, err := grpc_client.DialSimple(*target)
+	if err != nil {
+		return err
+	}
+	env := real_environment.NewRealEnv(nil /*=healthchecker*/)
```
tylerwilliams, line 104: we should prolly add a NewBatchEnv() method that returns a stubbed out healthchecker for tools

enterprise/server/remote_execution/operation/operation.go
```diff
@@ -315,3 +315,105 @@ func Decode(serializedOperation string) (*longrunning.Operation, error) {
...
+	}
+	s.stream = nil
+	s.client = nil
+	s.cancel()
+	return nil
+}
+
+// Message contains an operation along with its payload conveniently
+// unmarshaled.
+type Message struct {
```
tylerwilliams, line 390: can we call this something else? it's easily confused with proto.Message

server/util/mdutil/mdutil.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"google.golang.org/grpc/metadata"
+)
+
+// Parse parses gRPC metadata from a list of strings in KEY=VALUE format.
+func Parse(kvs ...string) (metadata.MD, error) {
```
tylerwilliams, line 12: wonder if this would be better in server/util/bazel_request/bazel_request.go

server/util/flag/flag.go
```diff
@@ -16,12 +16,17 @@ var YAMLIgnore = autoflags.YAMLIgnoreTag
...
+var ErrHelp = flag.ErrHelp
 var CommandLine = flag.CommandLine
 var Args = flag.Args
 
 type Flag = flag.Flag
 type FlagSet = flag.FlagSet
 
+func New[T any](flagset *flag.FlagSet, name string, defaultValue T, usage string, tags ...autoflags.Taggable) *T {
+	return autoflags.New[T](flagset, name, defaultValue, usage, tags...)
+}
```
tylerwilliams, line 28: image

enterprise/server/util/rexec/rexec.go
```diff
@@ -0,0 +1,302 @@
...
+//	err = cmd.Start(ctx)
+//	err = cmd.Wait(ctx)
+//	res, err := cmd.Result(ctx)
+//
+// Command properties are configured by setting attributes on *Cmd:
+//
+//	cmd.Dir = "/path/to/input_root_dir"
+//	cmd.Env = []string{"FOO=BAR"}
+//	cmd.Platform = []string{"recycle-runner=true"}
+package rexec
```
tylerwilliams, line 28: this would be nice to split into another CL

tylerwilliams (changes requested):

tylerwilliams:
I mentioned in chat but will say it here too -- I think this is really cool but would really like to clean up remote bazel and consolidate it with workflows so we can make that stuff easier to use in general (and via the CLI). Will let more folks start using the firecracker-ci stuff sooner!

cli/execute/execute.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+var flags = flag.NewFlagSet("execute", flag.ContinueOnError)
+
+// Bazel-equivalent flags.
+var (
+	target        = flags.String("remote_executor", "grpcs://remote.buildbuddy.io", "Remote execution service target.")
+	instanceName  = flags.String("remote_instance_name", "", "Value to pass as an instance_name in the remote execution API.")
+	timeout       = flags.Duration("remote_timeout", 1*time.Hour, "Timeout used for the action.")
+	remoteHeaders = flag.New(flags, "remote_header", []string{}, "Header to be applied to all outgoing gRPC requests, as a `NAME=VALUE` pair. Can be specified more than once.")
```
vadimberezniker, line 29: Could we wrap the FlagSet struct and add the additional flag types there instead of exposing `New`? It would be nice if flag usage was consistent between CLI and server.

enterprise/server/util/rexec/rexec.go
bduffany: Sent https://github.com/buildbuddy-io/buildbuddy/pull/5108

server/util/mdutil/mdutil.go
bduffany: wdyt about merging `bazel_request.go` into this package, since it's not bazel-specific at this point? (I have an itch to rename that package every time I look at it :sweat_smile:)

enterprise/server/remote_execution/operation/operation.go
bduffany: renamed to ExecuteOperation in the other PR

cli/execute/execute.go
bduffany: Done - I just set it to `nil` for now since a stub impl is not needed yet.

cli/execute/execute.go
bduffany: was considering that, but got lazy. Will send out a separate PR for that shortly

bduffany: OK actually it's not so simple, because `Slice()` and `Struct()` have type parameters, and IIRC struct methods can't have per-method type params like that.

Would prefer to stick with this for now.

cli/execute/execute.go
```diff
@@ -0,0 +1,160 @@
...
+		Arguments:            cmdArgs,
+		EnvironmentVariables: environ,
+		Platform:             platform,
+	}
+	action := &repb.Action{}
+	if *timeout > 0 {
+		action.Timeout = durationpb.New(*timeout)
+	}
+	// TODO: use capabilities client and respect digest function & compressor.
+	df := repb.DigestFunction_SHA256
```
tylerwilliams, line 128: please make this a flag

server/util/mdutil/mdutil.go
tylerwilliams: that sounds fine with me if you want to do it in a separate CL

tylerwilliams (changes requested):

cli/execute/execute.go
bduffany: Done

server/remote_cache/digest/digest.go
```diff
@@ -637,3 +637,15 @@ func (g *Generator) RandomDigestBuf(sizeBytes int64) (*repb.Digest, []byte, erro
...
+// ParseFunction parses a digest function name to a proto.
+func ParseFunction(value string) (repb.DigestFunction_Value, error) {
+	switch value {
+	case "SHA256", "sha256", "":
+		return repb.DigestFunction_SHA256, nil
+	case "BLAKE3", "blake3":
+		return repb.DigestFunction_BLAKE3, nil
+	default:
+		return 0, status.InvalidArgumentErrorf("unknown digest function %q", value)
+	}
```
tylerwilliams, line 650: I think this can be:

```
		if df, ok := repb.DigestFunction_Value_value[strings.ToUpper(dfString)]; ok {
			digestFunction = repb.DigestFunction_Value(df)
		} else {
			return nil, status.InvalidArgumentErrorf("Unknown digest function: %q", dfString)
		}
```

then you can use this function on line 435 where we do this today

bduffany: Done

tylerwilliams (approved):
LGTM

PR #5108, opened 2023-10-25
Add remote execution client library

bduffany:
This PR introduces an `rexec` package, which aims to provide a standardized, idiomatic, and flexible way to use the remote execution API from golang.

The motivation is that we are accumulating more code that makes RBE requests:
* https://github.com/buildbuddy-io/buildbuddy/blob/41d6a9495288e84e4eab405c82602a97809dd811/enterprise/server/test/integration/remote_execution/rbeclient/rbeclient.go#L118
* https://github.com/buildbuddy-io/buildbuddy/blob/41d6a9495288e84e4eab405c82602a97809dd811/enterprise/server/workflow/service/service.go#L1469
* https://github.com/buildbuddy-io/buildbuddy/blob/41d6a9495288e84e4eab405c82602a97809dd811/enterprise/tools/replay_action/replay_action.go#L251
* https://github.com/buildbuddy-io/buildbuddy/blob/41d6a9495288e84e4eab405c82602a97809dd811/enterprise/server/hostedrunner/hostedrunner.go#L93
* (I'd like to add yet another location with `bb execute`, for running ad-hoc actions via the CLI)

A little bit of code duplication is normally not a bad thing in and of itself, but making an RBE request correctly is a fairly laborious and error-prone process. The steps are:
* Assemble Command proto (using verbose proto APIs)
* Upload input root directory (ideally do this in parallel with the Command upload)
* Upload Action proto, with Command / input root digests
* (For all of the above, don't forget about remote instance name, compression, and digest function)
* Begin /Execute stream - keep track of operation `name` state for reliably reconnecting the stream if it disconnects.
* If the stream disconnects, then retry with /WaitExecution
* Once a `done` operation is observed, terminate the stream
* Parse the results from Operation stream (terminating the stream when `done`), unpack protos from the stream.

The `rexec` package aims to provide a library to take care of all of this logic. It aspires to be:

* Idiomatic/intuitive: feels like `os/exec` but for remote execution.
  * Running a command is just Start() and Wait(), or equivalently `Run()`.
  * Don't try _too_ hard to adhere to `os/exec` since it may obscure important details.
  * Example: Wait() does not return an `os.ExitError` wrapping a `syscall.WaitStatus`, unlike the os/exec package. This is because syscall.WaitStatus is platform-specific, while this package supports running cross-platform executions.\
  * Another example: `cmd.Stdout` writer is not a thing, because this obscures the fact that we don't yet support stdout streaming.
* Easy to use: cache uploads are pretty mechanical, so low-level control should not be needed in most cases. Minute details of operation streaming and Action/Command proto representation can be represented using simpler APIs (e.g. `Upload()` uploads all dependencies for the Execute request)
* Flexible: should handle 90+% of things that are doable with the lower level RE API, or at least easily modifiable to do so with minimal refactoring.
  * For example, we can add an `InputRootDigest` field if people run into limitations with `Dir` only supporting a local dir (e.g. this will be needed for `replay_action`)

**Related issues**: N/A


enterprise/server/remote_execution/operation/operation.go
```diff
@@ -315,3 +315,105 @@ func Decode(serializedOperation string) (*longrunning.Operation, error) {
...
+			return op, nil
+		}
+		if !status.IsUnavailableError(err) || s.name == "" {
+			return nil, err
+		}
+		if !r.Next() {
+			return nil, s.ctx.Err()
+		}
+		req := &repb.WaitExecutionRequest{Name: s.name}
+		next, err := s.client.WaitExecution(s.ctx, req)
```
vadimberezniker, line 369: RE API client code seems a little out of place in the `operation` package. Would it make more sense to move this to the `rexec` package?

enterprise/server/util/rexec/rexec.go
vadimberezniker, line 1: We should move this out of the enterprise directory if you plan to use it in the CLI code. It seems acceptable to have the client portion of RE be part of the OSS repo.

vadimberezniker (changes requested):

enterprise/server/remote_execution/operation/operation.go
bduffany: Done

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,451 @@
...
+	return &Cmd{
+		env:       env,
+		Arguments: append([]string{executable}, args...),
+	}
+}
+
+// Close closes any streams established to the server.
+func (c *Cmd) Close() error {
+	if c.stream != nil {
+		c.stream.CloseSend()
```
tylerwilliams, line 118: should this error be returned?

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,451 @@
...
+//	err := cmd.Upload(ctx)
+//	err = cmd.Start(ctx)
+//	err = cmd.Wait(ctx)
+//	res, err := cmd.Result(ctx)
+//
+// Command properties are configured by setting attributes on *Cmd:
+//
+//	cmd.Dir = "/path/to/input_root_dir"
+//	cmd.Env = []string{"FOO=BAR"}
+//	cmd.Platform = []string{"recycle-runner=true"}
```
tylerwilliams, line 27: I think this entire comment would fit better below on Cmd?

server/util/rexec/rexec.go
bduffany: done

enterprise/server/util/rexec/rexec.go
vadimberezniker, line 1: I think you forgot to remove the enterprise version.

vadimberezniker (approved):
This is sweet.

I would suggest adding support for automatic execution retries in a follow up PR, similar to how Bazel handles execution errors. Some of our resilience logic currently assumes the client will retry errors.

enterprise/server/util/rexec/rexec.go
bduffany: woooops :face_exhaling:

bduffany:
> I would suggest adding support for automatic execution retries in a follow up PR, similar to how Bazel handles execution errors. Some of our resilience logic currently assumes the client will retry errors.

Good call

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	cmd := &repb.Command{
+		Arguments:            c.Arguments,
+		EnvironmentVariables: envVars,
+		Platform:             p,
+	}
+	var commandDigest, inputRootDigest *repb.Digest
+	eg, egctx := errgroup.WithContext(ctx)
+	eg.Go(func() error {
+		ctx := egctx
```
tylerwilliams, line 151: nit: prolly clearer if you juse use gctx directly in the calls rather than redefining ctx like this?

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+	}
+	msg.Err = gstatus.FromProto(msg.ExecuteResponse.GetStatus()).Err()
+	return msg, nil
+}
+
+// AssemblePlatform assembles a Platform proto from a list of NAME=VALUE pairs.
+// If the same name is specified more than once, the last one wins. The entries
+// are sorted by name, so that the platform is cache-friendly.
+func AssemblePlatform(pairs ...string) (*repb.Platform, error) {
```
tylerwilliams, line 408: should this be private? I searched the other CL to see where it was used but did not find it.

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+		})
+	}
+	return p, nil
+}
+
+// AssembleEnvironmentVariables assembles a list of EnvironmentVariable protos
+// from a list of NAME=VALUE pairs. If the same name is specified more than
+// once, the last one wins. The entries are sorted by name, so that the
+// environment variables are cache-friendly.
+func AssembleEnvironmentVariables(pairs ...string) ([]*repb.Command_EnvironmentVariable, error) {
```
tylerwilliams, line 433: Maybe could be package private also?

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+	m := map[string]string{}
+	for _, pair := range pairs {
+		parts := strings.SplitN(pair, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid environment variable")
```
tylerwilliams, line 438: how about

`fmt.Errorf("invalid environment variable: %q", pair)`

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+		m[parts[0]] = parts[1]
+	}
+	names := maps.Keys(m)
+	sort.Strings(names)
```
tylerwilliams, line 443: nice

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+	ExecuteOperationMetadata *repb.ExecuteOperationMetadata
+	// ExecuteResponse contains any response unpacked from the operation.
+	ExecuteResponse *repb.ExecuteResponse
+	// Err contains any error parsed from the ExecuteResponse status field.
+	Err error
+}
+
+// UnpackOperation unmarshals all expected execution-specific fields from the
+// given operationn.
+func UnpackOperation(op *longrunning.Operation) (*ExecuteOperation, error) {
```
tylerwilliams, line 387: private?

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+		s.stream = nil
+	}
+	s.client = nil
+	s.cancel()
+	return err
+}
+
+// ExecuteOperation contains an operation along with its execution-specific
+// payload.
+type ExecuteOperation struct {
```
tylerwilliams, line 373: private?

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+)
+
+type state int
+
+const (
+	initial state = iota
+	uploading
+	starting
+	waiting
+)
```
tylerwilliams, line 65: I'm a little unclear about the expected use of the API -- is it safe to call Start/Upload/Wait multiple times? What about ordering?

I think a possibly clearer way to handle this would be to return errors from each of these methods if the state is not what you'd expect from normal operation.

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,452 @@
...
+//	err := cmd.Upload(ctx)
+//	err = cmd.Start(ctx)
+//	err = cmd.Wait(ctx)
+//	res, err := cmd.Result(ctx)
+//
+// Command properties are configured by setting attributes on *Cmd:
+//
+//	cmd.Dir = "/path/to/input_root_dir"
+//	cmd.Env = []string{"FOO=BAR"}
+//	cmd.Platform = []string{"recycle-runner=true"}
```
tylerwilliams, line 27: I'm kinda meh on this bit -- maybe it's the java in me but I feel like classes should be responsible for their own state and nobody else should be reaching in and touching stuff like this. I think it'd be clearer if you disallowed this and added an option to provide a CommandProto if the caller wants to override it, otherwise you build your own.

tylerwilliams (changes requested):

server/util/rexec/rexec.go
bduffany: I went ahead and made this refactor since it feels much cleaner this way, and later if we do want a higher-level / idiomatic construct like `*Cmd` we can build something like that using these lower level primitives.

server/util/rexec/rexec.go
bduffany: after refactoring, Wait() returns one of these, so kept it public.

server/util/rexec/rexec.go
bduffany: kept it public as it's useful for packages that want to use the stream directly instead of calling `Wait()` just to get the last operation (e.g. if they want to log status updates).

server/util/rexec/rexec.go
bduffany: Done

server/util/rexec/rexec.go
bduffany: N/A after refactoring

server/util/rexec/rexec.go
bduffany: done

server/util/rexec/rexec.go
bduffany: In the latest refactor, clients will use this to assemble the Command proto.

server/util/rexec/rexec.go
bduffany: N/A after refactoring

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,282 @@
+// package rexec provides a collection of utility functions for remote execution
+// clients.
```
tylerwilliams, line 2: nit: rm "a collection of" and make one line

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,282 @@
...
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+	rspb "github.com/buildbuddy-io/buildbuddy/proto/resource"
+	gstatus "google.golang.org/grpc/status"
+)
+
+// AssembleEnv assembles a list of EnvironmentVariable protos from a list of
+// NAME=VALUE pairs. If the same name is specified more than once, the last one
+// wins. The entries are sorted by name, so that the environment variables are
+// cache-friendly.
+func AssembleEnv(pairs ...string) ([]*repb.Command_EnvironmentVariable, error) {
```
tylerwilliams, line 31: probably would call this MakeEnv or MakeEnvVars (and MakePlatform) below b/c Assemble is a little wordy

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,282 @@
...
+// are sorted by name, so that the platform is cache-friendly.
+func AssemblePlatform(pairs ...string) (*repb.Platform, error) {
+	m := map[string]string{}
+	for _, s := range pairs {
+		parts := strings.SplitN(s, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid exec property %q (expected NAME=VALUE)", s)
+		}
+		m[parts[0]] = parts[1]
+	}
```
tylerwilliams, line 63: not a huge deal but if you wanted to pull this into a helper function you could (it's the same code here and above?)

func parsePairs(pairs []string) (map[string]string, error) {
	m := map[string]string{}
	for _, s := range pairs {
		parts := strings.SplitN(s, "=", 2)
		if len(parts) != 2 {
			return nil, return status.InvalidInputError("invalid pair %q, expected key=val format")
		}
		m[parts[0]] = parts[1]
	}
    return m, nil
}

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,282 @@
...
+		})
+	}
+	return p, nil
+}
+
+// Prepare transfers the given Command and local input root directory to cache,
+// and populates the resulting digests into the given Action. An empty string
+// for input root means that an empty directory will be used as the input root.
+// A resource name pointing to the remote Action is returned.
+func Prepare(ctx context.Context, env environment.Env, instanceName string, digestFunction repb.DigestFunction_Value, action *repb.Action, cmd *repb.Command, localInputRoot string) (*rspb.ResourceName, error) {
```
tylerwilliams, line 80: nit: s/localInputRoot/inputRootDir/g ?

server/util/rexec/rexec.go
tylerwilliams: if no packages do this yet, we could start with it private

tylerwilliams (changes requested):

server/util/rexec/rexec.go
bduffany: Done

server/util/rexec/rexec.go
bduffany: Done

server/util/rexec/rexec.go
bduffany: good call, done

server/util/rexec/rexec.go
bduffany: Done

server/util/rexec/rexec.go
bduffany: done

server/util/rexec/rexec.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// parsePairs parses a list of "NAME=VALUE" pairs into a map. If the same NAME
+// appears more than once, the last one wins.
+func parsePairs(pairs []string) (map[string]string, error) {
+	m := map[string]string{}
+	for _, pair := range pairs {
+		parts := strings.SplitN(pair, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid environment variable %q (expected NAME=VALUE)", pair)
```
tylerwilliams, line 74: would prefer to use a status error here rather than a fmt.Errorf

InvalidInput fits

tylerwilliams (approved):
LGTM

Thanks for all the back and forth, this will be a nice addition!

bduffany:
> Thanks for all the back and forth, this will be a nice addition!

np, thanks for the feedback - feels like it's in a much better state now :)

server/util/rexec/rexec.go
bduffany: done

PR #5111, opened 2023-10-25
Add type-safe fixed-size byte buffer pool option to `bytebufferpool` package.

vadimberezniker:
It essentially already existed (safeSyncPool) but was private to the package.

Removed use of generics since we were ever using it with a single type.


tylerwilliams (approved):
LGTM -- might see if @bduffany has any comments

bduffany (approved):

PR #5115, opened 2023-10-26
workflows: set env vars needed for snapshot matching

bduffany:
* Removes `RemoteSnapshotOptions` - after offline discussion we decided to use env instead.
* Set `GIT_BRANCH`, `GIT_BASE_BRANCH`, `GIT_REPO_DEFAULT_BRANCH` in env, which are all useful for snapshot matching.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -502,23 +502,6 @@ func run() error {
 	}
 	ws.rootDir = rootDir
 
-	if *workflowID != "" {
-		if err := os.Setenv("CI", "true"); err != nil {
-			return status.WrapError(err, "set CI=true")
-		}
-	}
-
-	// Set convenience env vars (best-effort).
-	if *prNumber != 0 {
-		os.Setenv("GIT_PR_NUMBER", fmt.Sprint(*prNumber))
-	}
-	if *commitSHA != "" {
-		os.Setenv("GIT_COMMIT", fmt.Sprint(*commitSHA))
-	}
-	if *pushedBranch != "" {
-		os.Setenv("GIT_BRANCH", fmt.Sprint(*pushedBranch))
-	}
-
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -66,6 +67,12 @@ var (
 	enableActionMerging               = flag.Bool("remote_execution.enable_action_merging", true, "If enabled, identical actions being executed concurrently are merged into a single execution.")
 )
 
+const (
+	// gRPC header containing the serialized RemoteSnapshotOptions.
+	// Keep in sync with enterprise/server/workflow/service.go
+	snapshotOptionsHeader = "x-buildbuddy-remotesnaphotopts-bin"
```
tylerwilliams, line 73: can this go in platform / be an x-buildbuddy-platform.* header?

bduffany: That feels a bit brittle, because it could potentially affect the routing key. Platform overrides don't technically affect the routing key today (and don't get included in various platform hashes), but I feel that's a bug that we should eventually fix.

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -508,17 +508,6 @@ func run() error {
...
-	// Set convenience env vars (best-effort).
-	if *prNumber != 0 {
-		os.Setenv("GIT_PR_NUMBER", fmt.Sprint(*prNumber))
-	}
-	if *commitSHA != "" {
-		os.Setenv("GIT_COMMIT", fmt.Sprint(*commitSHA))
-	}
-	if *pushedBranch != "" {
-		os.Setenv("GIT_BRANCH", fmt.Sprint(*pushedBranch))
-	}
```
bduffany, line 520: FYI these changes haven't been released yet. I moved these into the app since it seemed simpler/cleaner to set all the Git metadata in one place.

tylerwilliams (approved):

PR #5116, opened 2023-10-26
executor: use git refs for snapshot read/write

bduffany:
:arrow_right: depends on https://github.com/buildbuddy-io/buildbuddy/pull/5115

* Use main git ref (pushed branch name) for reading and writing
* Use fallback git refs for reading only (PR base branch and repo default branch, if different from PR base branch)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -481,7 +481,7 @@ type FirecrackerContainer struct {
 	// based on the incoming task. If there is a new task, you may want to call
 	// NewContainer again rather than directly unpausing a pre-existing container,
 	// to make sure these fields are updated and the best snapshot match is used
-	snapshotKey             *fcpb.SnapshotKey
+	snapshotKeySet          *fcpb.SnapshotKeySet
 	createFromSnapshot      bool
 	supportsRemoteSnapshots bool
 
@@ -598,7 +598,7 @@ func NewContainer(ctx context.Context, env environment.Env, task *repb.Execution
 		if *snaputil.EnableLocalSnapshotSharing {
 			runnerID = ""
 		}
-		c.snapshotKey, err = snaploader.NewKey(task, cd.GetHash(), runnerID)
+		c.snapshotKeySet, err = snaploader.SnapshotKeySet(task, cd.GetHash(), runnerID)
 		if err != nil {
 			return nil, err
 		}
@@ -607,11 +607,11 @@ func NewContainer(ctx context.Context, env environment.Env, task *repb.Execution
 
 		recyclingEnabled := platform.IsTrue(platform.FindValue(task.GetCommand().GetPlatform(), platform.RecycleRunnerPropertyName))
 		if recyclingEnabled && *snaputil.EnableLocalSnapshotSharing {
-			_, err := loader.GetSnapshot(ctx, c.snapshotKey, c.supportsRemoteSnapshots)
+			_, err := loader.GetSnapshot(ctx, c.snapshotKeySet, c.supportsRemoteSnapshots)
 			c.createFromSnapshot = (err == nil)
 		}
 	} else {
-		c.snapshotKey = opts.SavedState.GetSnapshotKey()
+		c.snapshotKeySet = &fcpb.SnapshotKeySet{BranchKey: opts.SavedState.GetSnapshotKey()}
 
 		// TODO(bduffany): add version info to snapshots. For example, if a
 		// breaking change is made to the vmexec API, the executor should not
@@ -725,8 +725,8 @@ func alignToMultiple(n int64, multiple int64) int64 {
 	return n + multiple - remainder
 }
 
-func (c *FirecrackerContainer) SnapshotKey() *fcpb.SnapshotKey {
-	return proto.Clone(c.snapshotKey).(*fcpb.SnapshotKey)
+func (c *FirecrackerContainer) SnapshotKeySet() *fcpb.SnapshotKeySet {
+	return proto.Clone(c.snapshotKeySet).(*fcpb.SnapshotKeySet)
 }
 
 // State returns the container state to be persisted to disk so that this
@@ -745,7 +745,7 @@ func (c *FirecrackerContainer) State(ctx context.Context) (*rnpb.ContainerState,
 		IsolationType: string(platform.FirecrackerContainerType),
 		FirecrackerState: &rnpb.FirecrackerState{
 			VmConfiguration: c.vmConfig,
-			SnapshotKey:     c.snapshotKey,
+			SnapshotKey:     c.snapshotKeySet.GetBranchKey(),
 		},
 	}
 	return state, nil
@@ -840,7 +840,11 @@ func (c *FirecrackerContainer) saveSnapshot(ctx context.Context, snapshotDetails
 	}
 
 	snaploaderStart := time.Now()
-	if err := c.loader.CacheSnapshot(ctx, c.snapshotKey, opts); err != nil {
+	// Note: we only update the main snapshot here, i.e. the one corresponding
+	// to the pushed git branch. We do not update the fallback key(s) that we
+	// might've read from, i.e. the ones corresponding to the PR base branch or
+	// repo's default branch.
+	if err := c.loader.CacheSnapshot(ctx, c.snapshotKeySet.GetBranchKey(), opts); err != nil {
 		return status.WrapError(err, "add snapshot to cache")
 	}
 	log.CtxDebugf(ctx, "snaploader.CacheSnapshot took %s", time.Since(snaploaderStart))
@@ -935,7 +939,7 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context) error {
 	}
 	log.CtxDebugf(ctx, "Command: %v", reflect.Indirect(reflect.Indirect(reflect.ValueOf(machine)).FieldByName("cmd")).FieldByName("Args"))
 
-	snap, err := c.loader.GetSnapshot(ctx, c.snapshotKey, c.supportsRemoteSnapshots)
+	snap, err := c.loader.GetSnapshot(ctx, c.snapshotKeySet, c.supportsRemoteSnapshots)
 	if err != nil {
 		return status.WrapError(err, "failed to get snapshot")
 	}
@@ -1120,7 +1124,7 @@ func (c *FirecrackerContainer) convertToCOW(ctx context.Context, filePath, chunk
 		return nil, status.WrapError(err, "make chunk dir")
 	}
 	// Use vmCtx for the COW since IO may be done outside of the task ctx.
-	cow, err := copy_on_write.ConvertFileToCOW(c.vmCtx, c.env, filePath, cowChunkSizeBytes(), chunkDir, c.snapshotKey.InstanceName, c.supportsRemoteSnapshots)
+	cow, err := copy_on_write.ConvertFileToCOW(c.vmCtx, c.env, filePath, cowChunkSizeBytes(), chunkDir, c.snapshotKeySet.GetBranchKey().GetInstanceName(), c.supportsRemoteSnapshots)
 	if err != nil {
 		return nil, status.WrapError(err, "convert file to COW")
 	}
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -840,10 +840,10 @@ func (c *FirecrackerContainer) saveSnapshot(ctx context.Context, snapshotDetails
 	}
 
 	snaploaderStart := time.Now()
-	// Note: we only update the main snapshot here, i.e. the one corresponding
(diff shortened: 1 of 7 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -2371,3 +2357,7 @@ func appendToLog(message string) *repb.Command {
 			`},
 	}
 }
+
+func recyclingEnabledTask() *repb.ScheduledTask {
```
tylerwilliams, line 2361: eh?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -182,24 +208,45 @@ func New(env environment.Env) (*FileCacheLoader, error) {
...
+	for i, key := range allKeys {
+		manifest, err := l.getSnapshot(ctx, key)
+		if err != nil {
+			lastErr = err
+			continue
+		}
+		if i == 0 {
+			log.CtxInfof(ctx, "Found preferred snapshot ref=%q in cache", key.GetRef())
+		} else {
+			log.CtxInfof(ctx, "Found fallback snapshot ref=%q in cache", key.GetRef())
```
tylerwilliams, line 223: maybe our refs proto should just be a list of refs instead of a preferred + list of refs?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -88,10 +104,19 @@ func fileDigest(filePath string) (*repb.Digest, error) {
 
 // Snapshot holds a snapshot manifest along with the corresponding cache key.
 type Snapshot struct {
-	key      *fcpb.SnapshotKey
+	writeKey *fcpb.SnapshotKey
+	readKey  *fcpb.SnapshotKey
 	manifest *fcpb.SnapshotManifest
 }
 
+func (s *Snapshot) GetWriteKey() *fcpb.SnapshotKey {
```
tylerwilliams, line 112: I don't see this used anywhere yet, that intentional?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -42,21 +42,33 @@ const (
...
+// SnapshotKeys returns the cache keys for potential snapshot matches,
+// as well as the key that should be written to.
+//
 // TODO: include a version number in the key somehow, so that
 // if we make breaking changes e.g. to the vmexec API or firecracker
 // version etc., we can ensure that incompatible snapshots don't get reused.
-func NewKey(task *repb.ExecutionTask, configurationHash, runnerID string) (*fcpb.SnapshotKey, error) {
-	pd, err := digest.ComputeForMessage(task.GetCommand().GetPlatform(), repb.DigestFunction_SHA256)
+func SnapshotKeys(task *repb.ScheduledTask, configurationHash, runnerID string) (*fcpb.SnapshotKeys, error) {
+	pd, err := digest.ComputeForMessage(task.GetExecutionTask().GetCommand().GetPlatform(), repb.DigestFunction_SHA256)
```
tylerwilliams, line 52: kind of a nit, but when I see lots of deep references like this (especially repeated ones), I'll often pull define the common bit like:

```
func SnapshotKeys(scheduled *repb.ScheduledTask, ...

task := scheduled.GetExecutionTask()
```

then you can shorten this to `task.GetCommand().GetPlatform()` and below to `task.GetExecuteRequest().GetInstanceName()`

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -500,20 +500,20 @@ func NewContainer(ctx context.Context, env environment.Env, task *repb.Execution
...
-		recyclingEnabled := platform.IsTrue(platform.FindValue(task.GetCommand().GetPlatform(), platform.RecycleRunnerPropertyName))
+		recyclingEnabled := platform.IsTrue(platform.FindValue(task.GetExecutionTask().GetCommand().GetPlatform(), platform.RecycleRunnerPropertyName))
 		if recyclingEnabled && *snaputil.EnableLocalSnapshotSharing {
-			_, err := loader.GetSnapshot(ctx, c.snapshotKey)
+			_, err := loader.GetSnapshot(ctx, c.snapshotKeys)
 			c.createFromSnapshot = (err == nil)
 		}
 	} else {
-		c.snapshotKey = opts.SavedState.GetSnapshotKey()
+		c.snapshotKeys = &fcpb.SnapshotKeys{MainKey: opts.SavedState.GetSnapshotKey()}
```
tylerwilliams, line 516: not sure I have a better name but I found this confusing that snapshotKey**s** is a single proto rather than a slice of keys.

snapshotKeySet or SnapshotIdentifier or something? idk

tylerwilliams (changes requested):

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: I considered that, but kind of like having a separation between the keys that are read+write vs. the keys that are readonly.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: Yeah this did feel mildly awkward. I forgot to come back around and rename it. SnapshotKeySet sgtm - done

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
bduffany: woops, fixed

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: done

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: woops, reverted this readKey vs writeKey stuff. This key is only used in UnpackSnapshot so it's already readonly and it's the one associated w/ the manifest.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -737,7 +737,11 @@ func (c *FirecrackerContainer) saveSnapshot(ctx context.Context, snapshotDetails
 	}
 
 	snaploaderStart := time.Now()
-	if err := c.loader.CacheSnapshot(ctx, c.snapshotKey, opts); err != nil {
+	// Note: we only update the main snapshot here, i.e. the one corresponding
+	// to the pushed git branch. We do not update the fallback key(s) that we
+	// might've read from, i.e. the ones corresponding to the PR base branch or
+	// repo's default branch.
```
tylerwilliams, line 743: // NB: we only update the snapshot corresponding to the pushed git branch.
	// We do not update the snapshot for any fallback key(s) that we may have
	// read from, i.e. ones corresponding to the PR's base branch or the
	// repo's default branch.

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -693,7 +756,7 @@ func UnpackContainerImage(ctx context.Context, l *FileCacheLoader, imageRef, ima
 	// ChunkedFile then add it to cache.
 	// TODO(bduffany): single-flight this.
 	start := time.Now()
-	cow, err := copy_on_write.ConvertFileToCOW(ctx, l.env, imageExt4Path, chunkSize, outDir, key.InstanceName)
+	cow, err := copy_on_write.ConvertFileToCOW(ctx, l.env, imageExt4Path, chunkSize, outDir, "" /*=instanceName*/)
```
tylerwilliams, line 759: why the removal of instance name here?

tylerwilliams (changes requested):

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -42,21 +43,58 @@ const (
...
+		keys.FallbackKeys = append(keys.FallbackKeys, fallbackKey)
+	}
+	return keys, nil
+}
+
+func gitRefs(task *repb.ExecutionTask) (branchRef string, fallbackRefs []string) {
+	// NOTE: keep these names in sync with workflow service
+	branchRef = getEnv(task, "GIT_BRANCH")
+	if branchRef == "" {
+		return "", nil
```
maggie-lou, line 80: What does this case mean? Should we either return an error that the env vars weren't set correctly, or use the fallbacks if they're set?

enterprise/server/remote_execution/snaploader/snaploader_test.go
```diff
@@ -78,20 +78,82 @@ func TestPackAndUnpackChunkedFiles(t *testing.T) {
...
+	require.Equal(t, "main", forkKeys.FallbackKeys[1].Ref)
+
+	// Sanity check that the branch key is not found in cache, to make sure
+	// we're actually falling back.
+	_, err = loader.GetSnapshot(ctx, &fcpb.SnapshotKeySet{BranchKey: forkKeys.GetBranchKey()})
+	require.Error(t, err)
+
+	// Now try unpacking with the complete PR key set, including fallbacks. This
+	// should succeed.
+	mustUnpack(t, ctx, loader, forkKeys, forkWorkDir, opts)
```
maggie-lou, line 154: Could you add a test that we don't try to update any of the fallback snapshots when we save the snapshot?

proto/firecracker.proto
```diff
@@ -26,6 +26,20 @@ message VMConfiguration {
   // TODO: add container_image here?
 }
 
+message SnapshotKeySet {
+  // Snapshot key to be read from when resuming the snapshot and written to when
+  // the task is complete.
+  //
+  // For workflows, this is associated with the git branch checked out in the
+  // snapshotted git repo directory.
+  SnapshotKey branch_key = 1;
```
maggie-lou, line 35: Personally I prefer the language of something like `primary_key`. It's harder for me to conceptualize what a `branch_key` is

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: Our current "non-COW" OCI -> ext4 conversion does not have instance name in the cache key so I think it was unintentional when I introduced instance name into the cache key originally.

Do you think we should include the instance name?

enterprise/server/remote_execution/containers/firecracker/firecracker.go

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: I figured if `GIT_BRANCH` is not set, then we're not looking at a well-formed workflow action, so to be on the safe side we should ignore the fallbacks. Added a clarifying comment.

enterprise/server/remote_execution/snaploader/snaploader_test.go
bduffany: `CacheSnapshot` only accepts a single key, not a `SnapshotKeySet` so I think it's not something that can be tested in this package.

We could add a firecracker_test case but it felt a bit overkill.

proto/firecracker.proto
bduffany: haha, I originally discussed this with Tyler offline and we found `main_key` / `primary_key` to also be confusing and settled on `branch_key` :sweat_smile: 

I think this name is not super consequential so would prefer to leave as-is for now.

enterprise/server/remote_execution/snaploader/snaploader.go
tylerwilliams: This scares me a little -- are only workflows is allowed to use snapshotting? What about recyclable runners with container-isolation-type=firecracker?

enterprise/server/remote_execution/snaploader/snaploader.go
tylerwilliams: Maybe add a comment that htis is nil to match OCI->ext4 conversion?

tylerwilliams (approved):
LGTM but please make sure @maggie-lou is good before submitting

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: Those will still be allowed to use snapshotting but just won't have a git ref as part of the snapshot key.

maggie-lou: Is there a reason we don't want to include git ref for these non-workflow tasks? Seems like it would improve their snapshot match quality

maggie-lou (approved):
Had one question about including git ref for non-workflow tasks. Otherwise LGTM

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: In general, actions will not have these git ref env vars set. We are setting these specially for workflow actions.

bduffany: It's unclear to me whether git ref is the right thing there, vs something like a hash of the Action proto.

enterprise/server/remote_execution/snaploader/snaploader.go

PR #5117, opened 2023-10-26
Remove confusing PubSub log message.

vadimberezniker:
Instead, explicitely log if client disconnected.

**Related issues**: N/A


tylerwilliams (approved):
somewhat related: @siggisim and I were talking about how nice it would be if there was some way to attach a TimeoutCause to *all* cancellable contexts so that we could make it easier for customers to understand why something was stopped instead of just seeing "context cancelled"

PR #5128, opened 2023-10-30
Optimize podman IsImageCached

bduffany:
`IsImageCached` consistently adds at least 20ms of action overhead locally (10% of total execution time for a local NOP action) and around 30ms of overhead remotely. Optimize it by caching its result when returns true, since we don't expect podman images to disappear.

Also use `podman image exists` which appears to run a few ms faster than `podman image inspect` and is also more straightforward.

image

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2844

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/BUILD
+++ b/enterprise/server/remote_execution/containers/podman/BUILD
@@ -30,6 +30,7 @@ go_library(
         "//server/util/healthcheck",
         "//server/util/lockingbuffer",
         "//server/util/log",
+        "//server/util/lru",
         "//server/util/networking",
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):
LGTM

iain-macdonald (approved):
Not urgent, but do you know if we can just read this info from the local filesystem instead of having to invoke podman?

bduffany:
> Not urgent, but do you know if we can just read this info from the local filesystem instead of having to invoke podman?

I haven't looked at that, but if it's easy, that's probably preferable to this approach

PR #5131, opened 2023-10-30
Improve execution tracing

bduffany:
* Add a few spans to better understand scheduling overhead
* Have `bb execute` report time from each stage (when running with `--verbose=1`)

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2844

### DIFF
--- a/cli/execute/execute.go
+++ b/cli/execute/execute.go
@@ -133,11 +133,14 @@ func execute(cmdArgs []string) error {
 		return err
 	}
 	start := time.Now()
+	stageStart := start
 	log.Debugf("Preparing action for %s", cmd)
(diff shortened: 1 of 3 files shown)


cli/execute/execute.go
```diff
@@ -133,11 +133,14 @@ func execute(cmdArgs []string) error {
 		return err
 	}
 	start := time.Now()
+	stageStart := time.Now()
```
iain-macdonald, line 136: nite: maybe just assign from start?

iain-macdonald (approved):

vadimberezniker (approved):

cli/execute/execute.go
bduffany: done

PR #5133, opened 2023-10-30
Straighten out context timeouts and fix a npe

tylerwilliams:
- Ensure overall context is set; default it to 10 seconds.
- Raft ops need a short context, roughly 10x the RTTMilliseconds value
- Check before closing nil chan

### DIFF
--- a/enterprise/server/raft/listener/listener.go
+++ b/enterprise/server/raft/listener/listener.go
@@ -49,8 +49,6 @@ func (rl *RaftListener) AddLeaderChangeListener() (<-chan raftio.LeaderInfo, fun
 }
 
 func (rl *RaftListener) LeaderUpdated(info raftio.LeaderInfo) {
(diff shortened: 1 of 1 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -529,7 +530,7 @@ func (sm *Replica) Open(stopc <-chan struct{}) (uint64, error) {
 	if err := sm.loadReplicaState(db); err != nil {
 		return 0, err
 	}
-	go sm.processAccessTimeUpdates()
+	//go sm.processAccessTimeUpdates()
```
siggisim, line 533: intentional?

siggisim (approved):

enterprise/server/raft/replica/replica.go
tylerwilliams: no, thanks.

PR #5138, opened 2023-10-31
Enforce lease expiration on the server.

vadimberezniker:
Today we only enforce it when there's some activity on the stream.

**Related issues**: N/A


tylerwilliams (approved):
LGTM

bduffany (approved):
do you mind adding a test for this? (OK to do it as a follow up if it's easier, since I saw you were revamping scheduler_server_test)

vadimberezniker:
> do you mind adding a test for this? (OK to do it as a follow up if it's easier, since I saw you were revamping scheduler_server_test)

That was my plan. I have the test ready to go, but it depends on the test changes from the other PR.

PR #5140, opened 2023-10-31
platforms: introduce common platforms

sluongng:
Introduce a set of common platforms that we use in BuildBuddy.git on a
regular basis.
Moved the config settings under `//platforms/config`
and moved the toolchain settings under `//toolchains`.


platforms/BUILD
```diff
@@ -1,25 +1,38 @@
 package(default_visibility = ["//visibility:public"])
 
-config_setting(
-    name = "linux_x86_64",
+platform(
+    name = "linux_amd64",
```
bduffany, line 4: `s/amd64/x86_64/g` to be consistent with bazel's `cpu` constraint naming? and also so that the config_setting names match the platform names

I get that the Arch exec property doesn't match, but IMO that's more of a quirk in our RE API and we should stick with bazel's conventions

siggisim (approved):

platforms/BUILD
sluongng: I wasn't sure on what to use so I was fiddling around with both options and `amd64` was slightly shorter to write.

Will change to `x86_64`.

bduffany (approved):

PR #5142, opened 2023-10-31
Add timestamps to CI runner logs

bduffany:
This will make it a little easier to see how long things are taking, until we support proper timing metrics for the CI runner.

image

**Related issues**: N/A


maggie-lou (approved):
<3 Love this! Good idea

PR #5149, opened 2023-11-01
Make pebble cache select a key-version based on DB existence and contents if --cache.pebble.active_key_version is unset

iain-macdonald:
This PR changes the behavior of the version selector so that if `--cache.pebble.active_key_version` is negative, the active key is set to the highest version found in the database, or the highest known version if the database is newly created. If `--cache.pebble.active_key_version` is set, that version is used.

**Related issues**: [2350](https://github.com/buildbuddy-io/buildbuddy-internal/issues/2350)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -535,6 +537,12 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 		pebbleOptions.Cache = c
 	}
 
+	desc, err := cdbpebble.Peek(opts.RootDirectory, vfs.Default)
+	if err != nil {
+		return nil, err
+	}
+	created := !desc.Exists
```
maggie-lou, line 544: This variable is supposed to represent a cache that previously didn't exist, and is being created now right? Maybe rename to `newlyCreated`? `created` makes it sound like it existed in the past

Also is another way to implement this without having to import a new package checking if `databaseVersionMetadata` was not in the db? It looks like we return an empty `rfpb.VersionMetadata` in that case. Could we check for an empty lastModifyTime?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -576,10 +584,17 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
...
-	versionMetadata, err := pc.databaseVersionMetadata()
+	versionMetadata, err := pc.DatabaseVersionMetadata()
 	if err != nil {
 		return nil, err
 	}
+	if created && *opts.ActiveKeyVersion < 0 {
+		versionMetadata = pc.maxDatabaseVersionMetadata()
+	} else if created {
+		versionMetadata.MinVersion = *opts.ActiveKeyVersion
+		versionMetadata.MaxVersion = *opts.ActiveKeyVersion
```
maggie-lou, line 595: Do we have to set LastModifyUsec in this case?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -753,6 +776,9 @@ func (p *PebbleCache) maxDatabaseVersion() filestore.PebbleKeyVersion {
 }
 
 func (p *PebbleCache) activeDatabaseVersion() filestore.PebbleKeyVersion {
+	if p.activeKeyVersion < 0 {
```
maggie-lou, line 779: Should we just set activeKeyVersion = maxDBVersion at the same point where you initialize maxDBVersion above? I could see this causing a confusing bug if this method returns maxDBVersion but the field itself contains a different value (-1)

enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -2165,6 +2167,148 @@ func TestDeleteEmptyDirs(t *testing.T) {
...
+		activeKeyVersion := int64(filestore.Version1)
+		options.ActiveKeyVersion = &activeKeyVersion
+		pc := openPebbleCache(ctx, t, te, options, []string{"remote-instance-name-3"})
+		versionMetadata, err := pc.DatabaseVersionMetadata()
+		require.NoError(t, err)
+		assert.Equal(t, int64(filestore.Version1), versionMetadata.MinVersion)
+		assert.Equal(t, int64(filestore.Version3), versionMetadata.MaxVersion)
+
+		pc.Stop()
+	}
```
maggie-lou, line 2297: Don't know if it's valuable to add this, but your test cases always expand the version range. What should happen if we pass a version within the range? 

i.e. if version metadata is [1,3] and we set activeKeyVersion=2

enterprise/server/raft/filestore/filestore.go
```diff
@@ -68,6 +68,16 @@ func fileRecordSegments(r *rfpb.FileRecord) (partID string, groupID string, isol
...
 type PebbleKeyVersion int
 
+const (
+	// PebbleKeyVersion representing an unspecified key version. When this is
+	// the active key version, the database will select an active key version
+	// to use based on existing database contents and the maximum known key
+	// version.
+	//
+	// This is in a different const block so as not to affect the iota below.
+	UnspecifiedKeyVersion PebbleKeyVersion = -1
```
maggie-lou, line 78: To make this part of the pre-existing const block, you can do UnspecifiedKeyVersion PebbleKeyVersion = iota - 1. All the future consts will increment by 1

enterprise/server/raft/filestore/filestore.go
```diff
@@ -68,6 +68,16 @@ func fileRecordSegments(r *rfpb.FileRecord) (partID string, groupID string, isol
...
+)
+
 const (
 	// UndefinedKeyVersion is the version of all keys in the database
```
maggie-lou, line 82: Do we still need this? What does this represent?

maggie-lou:
Mostly LGTM  - just some small comments

enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -2165,6 +2167,148 @@ func TestDeleteEmptyDirs(t *testing.T) {
...
+	// version is written at the highest available key version.
+	activeKeyVersion := int64(filestore.UnspecifiedKeyVersion)
+	options.ActiveKeyVersion = &activeKeyVersion
+	pc := openPebbleCache(ctx, t, te, options, []string{"remote-instance-name-1"})
+	versionMetadata, err := pc.DatabaseVersionMetadata()
+	require.NoError(t, err)
+	assert.Equal(t, int64(filestore.MaxKeyVersion)-1, versionMetadata.MinVersion)
+	assert.Equal(t, int64(filestore.MaxKeyVersion)-1, versionMetadata.MaxVersion)
+
+	pc.Stop()
```
luluz66, line 2188: also check no error returned by Stop() here and below?

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: Older versions of the pebble code did not have the databaseVersionMetadata key so its non-existence does not indicate a new setup. 

Instead of importing the real pebble package, alias the Peek function in our pebble package.

iain-macdonald: Renamed to `newlyCreated` and aliased the `Peek()` call.

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: I don't think it's necessary, but it's probably slightly more correct so I went ahead and added it. Good call!

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: Yeah, I went back and forth about which approach to use. Updated to set the active version after cache creation. Let me know what you think of this approach.

enterprise/server/backends/pebble_cache/pebble_cache_test.go
iain-macdonald: Sure, sounds good.

enterprise/server/raft/filestore/filestore.go
iain-macdonald: That doesn't work because the const declaration increments the iota, changing the value of everything below it. I could add it at the bottom, and make it clear that new versions should be inserted between this const and the other versions, but thought this approach was simpler.

enterprise/server/raft/filestore/filestore.go
iain-macdonald: I think this is the version of un-migrated keys from before version 0 and could exist for on-prem deployments that have never upgraded the version.

enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -2165,6 +2167,148 @@ func TestDeleteEmptyDirs(t *testing.T) {
...
+	options := &pebble_cache.Options{RootDirectory: rootDir, MaxSizeBytes: maxSizeBytes}
+
+	// Confirm that a newly-created pebble database with an unspecified key
+	// version is written at the highest available key version.
+	activeKeyVersion := int64(filestore.UnspecifiedKeyVersion)
+	options.ActiveKeyVersion = &activeKeyVersion
+	pc := openPebbleCache(ctx, t, te, options, []string{"remote-instance-name-1"})
+	versionMetadata, err := pc.DatabaseVersionMetadata()
+	require.NoError(t, err)
+	assert.Equal(t, int64(filestore.MaxKeyVersion)-1, versionMetadata.MinVersion)
```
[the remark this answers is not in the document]
iain-macdonald: Yeah, I noticed that, but thought it'd be better to start a trend of moving to `assert` for assertions rather than being consistent with the rest of the file. If you feel strongly I can change it, though.

enterprise/server/backends/pebble_cache/pebble_cache_test.go
iain-macdonald: Whoops, the test I copied this from didn't have that. Added.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -575,10 +581,22 @@ func NewPebbleCache(env environment.Env, opts *Options) (*PebbleCache, error) {
 		includeMetadataSize:         opts.IncludeMetadataSize,
 	}
 
-	versionMetadata, err := pc.databaseVersionMetadata()
+	versionMetadata, err := pc.DatabaseVersionMetadata()
 	if err != nil {
 		return nil, err
 	}
+	if newlyCreated && *opts.ActiveKeyVersion < 0 {
```
vadimberezniker, line 588: I think opts.ActiveKeyVersion may be nil when initialized via the config?

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: `SetOptionDefaults` (called on line 514) sets a default value for `opts.ActiveKeyVersion`.

vadimberezniker, line 588: It would be a little neater if you write it like this:
```
	if newlyCreated && *opts.ActiveKeyVersion < 0 {
		activeVersion := *opts.ActiveKeyVersion
		if activeVersion < 0 {
			activeVersion = int64(filestore.MaxKeyVersion) - 1
		}
		versionMetadata.MinVersion = activeVersion
		versionMetadata.MaxVersion = activeVersion
		versionMetadata.LastModifyUsec = clock.Now().UnixMicro()
	}
```

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: Did you mean to remove this check here (p.activeKeyVersion < 0). I believe now p.activeKeyVersion can never be -1?

enterprise/server/raft/filestore/filestore.go
vadimberezniker: Maggie is right, you can manipulate the value of iota by subtracting from it so you can certainly put it in the same const block.

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: Oh yeah, missed this one. Fixed.

enterprise/server/raft/filestore/filestore.go
iain-macdonald: Ah, I misunderstood this and it wasn't working for me, but figured it out and updated. Sorry for the confusion!

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: SG, done.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -710,9 +730,18 @@ func (p *PebbleCache) databaseVersionKey() []byte {
 	return key
 }
 
+func (p *PebbleCache) maxDatabaseVersionMetadata() *rfpb.VersionMetadata {
```
vadimberezniker, line 733: you can delete this

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -751,6 +779,10 @@ func (p *PebbleCache) maxDatabaseVersion() filestore.PebbleKeyVersion {
 	return p.maxDBVersion
 }
 
+func (p *PebbleCache) setActiveDatabaseVersion(version filestore.PebbleKeyVersion) {
```
vadimberezniker, line 782: does this need to be its own func? seems like you could initialize it directly in NewPebbleCache

enterprise/server/util/pebble/pebble.go
```diff
@@ -375,6 +379,20 @@ func (idb *instrumentedDB) NewIndexedBatch() Batch {
 	return &instrumentedBatch{batch, batch, idb}
 }
 
+func Peek(dirname string, fs FS) (*DBDesc, error) {
```
vadimberezniker, line 382: I don't think we get much out of instrumenting this? I think a simple `var Peek = pebble.Peek` would be sufficient.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: Whoops, done.

enterprise/server/backends/pebble_cache/pebble_cache.go
iain-macdonald: Nope, removed.

enterprise/server/util/pebble/pebble.go
iain-macdonald: Sure, I don't feel too strongly either way, removed it.

PR #5156, opened 2023-11-01
Add temp code to detect ByteStream Read stalls.

vadimberezniker:
**Related issues**: N/A


server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -143,23 +145,46 @@ func (s *ByteStreamServer) Read(req *bspb.ReadRequest, stream bspb.ByteStream_Re
 	copyBuf := s.bufferPool.Get(bufSize)
 	defer s.bufferPool.Put(copyBuf)
 
+	detectStall := func(name string, work func()) {
```
tylerwilliams, line 148: i wrote util/canary for almost exactly this -- could you use that instead?

PR #5161, opened 2023-11-02
blobstore/aws: fix NotFound error handling

sluongng:
When configure AWS S3 blobstore, we first call HeadBucket to check
whether if the bucket exists. Upon a NotFound error, we should be
attempting to create a new bucket. For any other errors, such as
credential error or SigningError, we want to surface that logic
back to the user.

Flip the return statement between NotFound and other error types when
calling HeadBucket.


tempoz (approved):
LGTM! Thanks for doing this, Son!

PR #5163, opened 2023-11-02
Simplification of CacheProxy.Read func.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/util/cacheproxy/cacheproxy.go
+++ b/enterprise/server/util/cacheproxy/cacheproxy.go
@@ -325,12 +325,9 @@ func (c *CacheProxy) Read(req *dcpb.ReadRequest, stream dcpb.DistributedCache_Re
 	defer c.readBufPool.Put(copyBuf)
 
 	for {
-		n, err := io.ReadFull(reader, copyBuf)
+		n, err := ioutil.ReadTryFillBuffer(reader, copyBuf)
 		if err == io.EOF {
 			break
-		}
-		if err != nil && err != io.ErrUnexpectedEOF {
-			return err
 		}
 		if err := stream.Send(&dcpb.ReadResponse{Data: copyBuf[:n]}); err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


enterprise/server/util/cacheproxy/cacheproxy.go
```diff
@@ -328,17 +328,12 @@ func (c *CacheProxy) Read(req *dcpb.ReadRequest, stream dcpb.DistributedCache_Re
 		n, err := io.ReadFull(reader, copyBuf)
```
bduffany, line 328: I feel like the error handling logic is still a bit confusing. It seems like the utility of `ReadFull` is that it simplifies error handling, but we are still explicitly inspecting the errors that it returns, so we are losing out on the benefit.

wdyt about something like

```
for remaining := rn.GetDigest.GetSizeBytes(); remaining > 0; {
  b := copyBuf[:min(len(copyBuf), remaining)]
  n, err := io.ReadFull(reader, b)
  if err != nil {
    return err // all errors are unexpected
  }
  // stream.Send({ Data: b })
  remainder -= n
}
```

vadimberezniker: I don't think that's much better. I added a new ioutil func that works like io.ReadFull but does not return an error on a short read.

tylerwilliams (approved):

enterprise/server/util/cacheproxy/cacheproxy.go
bduffany: hm it seems like with this approach, we now need an additional Read() in order to discover the EOF? wondering if that would increase IO load

vadimberezniker: There's no change. Previously ErrUnexpectedEOF did not terminate the loop, it was only terminated on the next EOF.

bduffany (approved):

PR #5165, opened 2023-11-02
Remove group ID from remote snapshot key

bduffany:
The group ID makes it difficult to fetch and debug remote snapshots using the `vmstart` tool - we'd have to set up a fake authenticator and insert a fake group ID into the ctx. It's easier if we can just plug in an API key, and have the remote cache handle group ID partitioning.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -59,20 +59,41 @@ func NewKey(task *repb.ExecutionTask, configurationHash, runnerID string) (*fcpb
 	}, nil
 }
 
-// manifestKey returns the key for the snapshot manifest
+// localManifestKey returns the key for the local snapshot manifest.
 //
 // Because we always want runners to use the newest manifest/snapshot, this
 // doesn't actually create a digest of the manifest contents. It just takes a
 // hash of shared properties, for which the snapshot should be shared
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -59,20 +59,38 @@ func NewKey(task *repb.ExecutionTask, configurationHash, runnerID string) (*fcpb
...
 
-// manifestKey returns the key for the snapshot manifest
+// localManifestKey returns the key for the local snapshot manifest.
 //
 // Because we always want runners to use the newest manifest/snapshot, this
 // doesn't actually create a digest of the manifest contents. It just takes a
 // hash of shared properties, for which the snapshot should be shared
-func manifestKey(ctx context.Context, env environment.Env, s *fcpb.SnapshotKey) (*repb.Digest, error) {
+func localManifestKey(ctx context.Context, env environment.Env, s *fcpb.SnapshotKey) (*repb.Digest, error) {
 	gid, err := groupID(ctx, env)
```
maggie-lou, line 68: nit: Maybe add a comment explaining that we don't need this for the remote key, bc the cache handles group partitioning. In case someone wonders why these are different and tries to consolidate

maggie-lou (approved):

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: Done

PR #5166, opened 2023-11-02
Fix data race condition that is detected when we build with musl

tempoz:
Adds a lock around data that is accessed in two concurrent goroutines.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -1864,11 +1864,9 @@ func (c *FirecrackerContainer) SendExecRequestToGuest(ctx context.Context, cmd *
 	var lastObservedStats *repb.UsageStats
 	statsListener := func(stats *repb.UsageStats) {
 		container.Metrics.Observe(c, stats)
-		{
-			lastObservedStatsMutex.Lock()
-			defer lastObservedStatsMutex.Unlock()
-			lastObservedStats = stats
-		}
+		lastObservedStatsMutex.Lock()
+		lastObservedStats = stats
+		lastObservedStatsMutex.Unlock()
 	}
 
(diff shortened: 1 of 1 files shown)


tempoz:
* master
  * **PR #5166** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5166" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5166?utm_source=stack-comment).

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1860,10 +1860,15 @@ func (c *FirecrackerContainer) SendExecRequestToGuest(ctx context.Context, cmd *
...
+	var lastObservedStatsMutex sync.Mutex
 	var lastObservedStats *repb.UsageStats
 	statsListener := func(stats *repb.UsageStats) {
 		container.Metrics.Observe(c, stats)
-		lastObservedStats = stats
+		{
+			lastObservedStatsMutex.Lock()
+			defer lastObservedStatsMutex.Unlock()
+			lastObservedStats = stats
+		}
```
tylerwilliams, line 1871: nit: since there's no error checking that might disrupt the control flow and require the defer, this can be simplified to: 

statsListener := func(stats *repb.UsageStats) {
		container.Metrics.Observe(c, stats)
		lastObservedStatsMutex.Lock()
		lastObservedStats = stats
		lastObservedStatsMutex.Unlock()
}


(and same below)

tempoz: kk, wasn't sure of the preferred style. Done.

PR #5167, opened 2023-11-02
Enable affinity routing unconditionally, with an shutoff flag in case of emergency.

iain-macdonald:
**Related issues**: [222](https://github.com/buildbuddy-io/buildbuddy-internal/issues/222)

### DIFF
--- a/enterprise/server/scheduling/task_router/task_router.go
+++ b/enterprise/server/scheduling/task_router/task_router.go
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	affinityRoutingPermitted = flag.Bool("executor.affinity_routing_permitted", true, "If set, along with the 'affinity-routing' platform property, then (experimental) affinity routing is used. If false, then affinity routing is never used. This flag is intended to be used as an emergency shut-off for this experimental feature.")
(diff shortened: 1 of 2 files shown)


enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	affinityRoutingPermitted = flag.Bool("executor.affinity_routing_permitted", true, "If set, along with the 'affinity-routing' platform property, then (experimental) affinity routing is used. If false, then affinity routing is never used. This flag is intended to be used as an emergency shut-off for this experimental feature.")
+	affinityRoutingEnabled = flag.Bool("executor.affinity_routing_enabled", true, "Enables affinity routing, which attempts to route actions to the executor that most recently ran that action. This flag is intended for disabling this feature in an emergency.")
```
tylerwilliams, line 24: i don't think `This flag is intended for disabling this feature in an emergency.` adds much but up to you :P

tylerwilliams (approved):
I don't know which we have more of: `enable_feature_x` or `feature_x_enabled`

PR #5177, opened 2023-11-03
server/static: refactor to support FOSS build

sluongng:
Some of our FOSS code depend on sources in //enterprise to provide
the right config values. This prevents builds from our FOSS repo from
functioning correctly.

Leveraging rules_go's `tags` flag to specify a special Go build
directive `buildbuddy_foss` to selectively favor FOSS source code over
enterprise source code. Also, wrap this flag with a build config
`//platforms/configs:buildbuddy_foss` so that we can selectively declare
dependencies.

This enables us to do this

```
> rm -rf enterprise
> bazel build --config=foss server/static/...
```

Future PRs will leverage these new control knobs to fix FOSS builds
for good.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -170,6 +170,14 @@ build:macos --cxxopt=-std=c++17
 build:windows --host_cxxopt=/std:c++17
 build:windows --cxxopt=/std:c++17
 
+# Special config to help build FOSS without the //enterprise source code (WIP)
+#
+# Usage:
+#
+#   bazel build --config=foss server
+#
+build:foss --@io_bazel_rules_go//go/config:tags=buildbuddy_foss
+
 # Run Webdriver tests with --config=webdriver-debug to debug webdriver tests locally.
 # See server/testutil/webtester/webtester.go for more details.
 test:webdriver-debug --test_arg=-webdriver_debug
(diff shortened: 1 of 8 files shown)


siggisim:
I think we should find a way to either:

a) plumb these values through the environment, since that's how we handle these sorts of options i.e. `env.GetAuthenticator().SSOEnabled()` or `env.GetSecretService() != nil`. This is kind of the whole problem that the environment.go / interfaces.go is meant to solve, and this feels like it's adding yet another way to do that.

OR

b) move these [config files](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/config/config.go) that just have a single flag in them out of enterprise and into the non-enterprise directory

Adding a config setting and selects and stuff at build time is just gross and adds complexity. It should just work to run `bazel build server` without the need for extraneous config settings.

cc @tylerwilliams

sluongng:
I picked this approach because it was the easiest. Let me take a crack at the alternatives.

For reference, here is what the final state of this PR would look like https://github.com/buildbuddy-io/buildbuddy/compare/master...sluongng/foss-more-buildable

There are a few tricky parts like auditlog or githubauth that fit neither (a) nor (b). I will be sending separate PRs for those to look for comments.

PR #5181, opened 2023-11-03
Increase firecracker go sdk request timeout

bduffany:
The default timeout for `ResumeVM` is 500ms, which is not enough time when we have to resolve page faults remotely (confirmed via logging that `ResumeVM` is actually triggering page faults). This PR increases it to 30s so that we have more time.

Note, this timeout is also capped to the task ctx deadline. So things like executor shutdown will still result in proper ctx cancellation.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -210,6 +210,18 @@ var (
 	fatalErrPattern = regexp.MustCompile(`\b` + fatalInitLogPrefix + `(.*)`)
 )
 
+func init() {
+	// Configure firecracker request timeout (default: 500ms).
+	//
+	// We're increasing it from the default here since we do some remote reads
+	// during ResumeVM to handle page faults with UFFD.
+	//
+	// This 30s timeout was determined as the time it takes to load a bb repo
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -210,6 +210,18 @@ var (
...
+func init() {
+	// Configure firecracker request timeout (default: 500ms).
+	//
+	// We're increasing it from the default here since we do some remote reads
+	// during ResumeVM to handle page faults with UFFD.
+	//
+	// This 30s timeout was determined as the time it takes to load a bb repo
+	// snapshot in dev with a cold filecache (10s), times 3 to account for tail
+	// latency + larger VMs that could take longer to resume.
+	os.Setenv("FIRECRACKER_GO_SDK_REQUEST_TIMEOUT_MILLISECONDS", fmt.Sprint(30_000))
```
jdhollen, line 222: do you wanna flag this? or were you planning to roll it back after a bit? or do you think it doesnt matter much in the grand scheme of things?

bduffany: I don't think it matters too much not having this behind a flag.

jdhollen (approved):

PR #5182, opened 2023-11-03
add a db flush right after wait group finished

luluz66:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -3346,7 +3346,7 @@ func (p *PebbleCache) Stop() error {
 	if err := p.db.Flush(); err != nil {
 		return err
 	}
-	log.Infof("Pebble Cache [%s]: db flushed", p.name)
+	log.Infof("Pebble Cache [%s]: db flushed again", p.name)
 
 	return p.db.Close()
 }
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -3319,6 +3319,13 @@ func (p *PebbleCache) Stop() error {
 	}
 	log.Infof("Pebble Cache [%s]: waitgroups finished", p.name)
 
+	// Flushed db after all waitgroups finished to reduce the change of lost
```
vadimberezniker, line 3322: Can you call p.flushPartitionMetadata() as well? Would be good if we minimize any lost updates there as well.

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: Done.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -3319,6 +3319,14 @@ func (p *PebbleCache) Stop() error {
...
+	// evictions during app restarts
+	p.flushPartitionMetadata()
+	if err := p.db.Flush(); err != nil {
+		return err
+	}
+	log.Infof("Pebble Cache [%s]: db flushed", p.name)
```
vadimberezniker, line 3328: can you change this log message slightly so we can differentiate this log message from the other flush?

vadimberezniker (approved):

PR #5186, opened 2023-11-03
[SS] Skip the runner pool for snapshot-sharing-enabled runners

maggie-lou:
The motivation for the runner pool is that state is saved on the runner during an initial task, which can be reused for future workloads.

If snapshot sharing is enabled, a firecracker VM can be cloned from the cache and does not rely on previous state set on the runner, so we can circumvent the runner pool. In fact we *should* circumvent the runner pool and create a new runner with data from the incoming task, which can be used to find a better snapshot match than a runner created for a stale task

(Ex. If runner A was created for branch `feature_one` and an incoming workload is for branch `feature_two`, we should create a new runner intended for `feature_two`, rather than reuse the runner for branch `feature_one`, which would be more stale)

We need to do this because of how the Container interface is structured. We only pass data about the incoming workload when creating a new container, but not when unpausing it

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/BUILD
+++ b/enterprise/server/remote_execution/runner/BUILD
@@ -16,6 +16,7 @@ go_library(
         "//enterprise/server/remote_execution/commandutil",
         "//enterprise/server/remote_execution/container",
         "//enterprise/server/remote_execution/platform",
+        "//enterprise/server/remote_execution/snaputil",
         "//enterprise/server/remote_execution/vfs",
         "//enterprise/server/remote_execution/workspace",
         "//enterprise/server/tasksize",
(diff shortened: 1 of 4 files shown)


maggie-lou:
* master
  * **PR #5186** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5186" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5186?utm_source=stack-comment).

bduffany (approved):
nice :)

PR #5190, opened 2023-11-03
Add new context param to filecache and test it.

tylerwilliams:


enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -229,19 +248,20 @@ func (c *fileCache) FastLinkFile(node *repb.FileNode, outputPath string) (hit bo
...
 	// overwrite an existing link with different contents, all pointers
 	// to the old link would suddenly change to point to the new content,
 	// which is not good.
-	k := key(node)
+	k := groupSpecificKey(groupID, node)
 	c.l.Remove(k)
 
-	fp := c.filecachePath(node)
+	fp := c.filecachePath(k)
+	if err := disk.EnsureDirectoryExists(filepath.Dir(fp)); err != nil {
```
siggisim, line 271: Any worry in adding an additional `stat` call for each fast link for actions with tens of thousands of input files instead of doing it once up front?

enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -196,15 +202,28 @@ func (c *fileCache) scanDir() {
...
 	}
-	return node.GetDigest().GetHash() + suffix
+	return groupID + "/" + node.GetDigest().GetHash() + suffix
+}
+
+func groupIDStringFromContext(ctx context.Context) string {
+	if c, err := claims.ClaimsFromContext(ctx); err == nil {
+		return strings.ToUpper(c.GroupID)
+	}
+	return "ANON"
```
siggisim, line 217: Should we use `interfaces.AuthAnonymousUser` here too?

siggisim (approved):

enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -196,15 +202,28 @@ func (c *fileCache) scanDir() {
...
 	if node.GetIsExecutable() {
 		suffix = "." + executableSuffix
 	}
-	return node.GetDigest().GetHash() + suffix
+	return groupID + "/" + node.GetDigest().GetHash() + suffix
+}
+
+func groupIDStringFromContext(ctx context.Context) string {
+	if c, err := claims.ClaimsFromContext(ctx); err == nil {
+		return strings.ToUpper(c.GroupID)
```
vadimberezniker, line 215: nit: group IDs are always uppercase, not sure why this is needed?

vadimberezniker (approved):

enterprise/server/remote_execution/filecache/filecache.go
tylerwilliams: I did not notice an impact when testing this locally, so decided to leave this as-is for simplicity.

enterprise/server/remote_execution/filecache/filecache.go
tylerwilliams: Done, thanks!

enterprise/server/remote_execution/filecache/filecache.go
tylerwilliams: Done.

PR #5191, opened 2023-11-03
Add a little tool for running side-by-side rbe comparisons

iain-macdonald:
My bash script was getting complicated, so figured it was worth rewriting it in go for future use.

**Related issues**: N/A

### DIFF
--- a/enterprise/tools/rbesxs/rbesxs.go
+++ b/enterprise/tools/rbesxs/rbesxs.go
@@ -1,0 +1,220 @@
+// Utility for running a remote execution side-by-side test.
+//
+// Example usage:
+//
+// mkdir /tmp/sxs
+// git clone http://github.com/buildbuddy-io/buildbuddy /tmp/sxs/test
+// cp -R /tmp/sxs/test /tmp/sxs/control
+// bazel run //enterprise/tools/rbesxs -- \
+// --iterations=10 \
+// --api_key=abcdef123456 \
+// --setup \
+// --target=//server/util/status \
+// --test_repo=/tmp/sxs/test \
+// --test_flags=--enable_crazy_feature \
+// --control_repo=/tmp/sxs/control
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/commandutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/random"
+
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+)
+
+var (
+	iterations  = flag.Int("iterations", 1, "The number of side-by-side iterations to run.")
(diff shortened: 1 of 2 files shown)


enterprise/tools/rbesxs/rbesxs.go
```diff
@@ -0,0 +1,222 @@
+// Utility for running a remote execution side-by-side test.
```
bduffany, line 1: I wasn't sure what side-by-side meant initially - maybe worth adding a sentence or two more on what this tool does

enterprise/tools/rbesxs/rbesxs.go
```diff
@@ -0,0 +1,222 @@
...
+			setupCmd := newCommand(*target, *environment, *apiKey, remoteInstanceName, "", strings.Split(*setupFlags, ","))
+			log.Infof("iteration %d: running setup command: %s", i, strings.Join(setupCmd.Arguments, " "))
+			result = commandutil.Run(ctx, setupCmd, *controlRepo, nil, nil)
+			if result.ExitCode != 0 {
+				failures++
+				log.Infof("setup command failed, retrying: %s", string(result.Stderr))
+				i--
+				continue
+			}
+		}
```
bduffany, line 189: Is it intentional that the setup build runs on every iteration? I would think that you could do the setup build once per repo at the very beginning (or maybe 2-3 times per repo if you really want to make sure that every executor gets warmed up). After that, the executors should be pretty warmed up and ready for the "actual" test phase

bduffany (approved):

enterprise/tools/rbesxs/rbesxs.go
iain-macdonald: Good call, updated the comment.

enterprise/tools/rbesxs/rbesxs.go
iain-macdonald: It is intentional. Each control and test run uses a distinct remote instance names to isolate them from each other, and for the test I was running I wanted one setup build to take place with that instance name before running the test build. We could certainly make this behavior customizable in the future, but this was the behavior I wanted for my test, so it's what I added support for.

enterprise/tools/rbesxs/rbesxs.go
```diff
@@ -0,0 +1,222 @@
...
+	cmd = append(cmd, fmt.Sprintf("--remote_instance_name=%s", remoteInstanceName))
+	if tag != "" {
+		cmd = append(cmd, fmt.Sprintf("--build_metadata=TAGS=%s", tag))
+	}
+	cmd = append(cmd, "--noremote_accept_cached")
+	return &repb.Command{Arguments: cmd}
+}
+
+func bazelClean(ctx context.Context, workDir string) {
+	result := commandutil.Run(ctx, &repb.Command{Arguments: []string{"bazel", "clean"}}, workDir, nil, nil)
```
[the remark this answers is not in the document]
iain-macdonald: Oh thanks, didn't exactly understand the semantics of `--async` and thought it might return OK in cases where I wouldn't want to proceed here, but looks like that's not the case.

PR #5193, opened 2023-11-06
githubauth: move enable check to server/backends/github

sluongng:
The check basis is actually using the JwtKey on the github package.
Use it instead.


sluongng:
The goal here is to purge `enterprise` deps from `server/static/static.go` which will help unblock FOSS builds in the future.

PR #5196, opened 2023-11-06
Don't include version info when using `-c fastbuild`

bduffany:
Fixes a bug that lots of cache invalidations are happening on every commit.

Credit to @sluongng for finding this bug.

**Related issues**: N/A


maggie-lou (approved):
👏 👏

PR #5197, opened 2023-11-06
Use zerolog for GORM logging

bduffany:
Fixes the issue that GORM logs are logged to stderr unstructured.

**Related issues**: N/A

### DIFF
--- a/server/util/db/db.go
+++ b/server/util/db/db.go
@@ -16,8 +16,6 @@ import (
 	"strings"
 	"time"
 
-	golog "log"
-
 	"github.com/buildbuddy-io/buildbuddy/aws_rds_certs"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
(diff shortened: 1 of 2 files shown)


server/util/db/db.go
```diff
@@ -738,23 +724,56 @@ func ParseDatasource(ctx context.Context, datasource string, advancedConfig *Adv
...
+// parameterized queries.
 type sqlLogger struct {
-	logger.Interface
-	logLevel logger.LogLevel
+	SlowThreshold time.Duration
+	LogLevel      logger.LogLevel
+}
+
+func (l *sqlLogger) Info(ctx context.Context, format string, args ...interface{}) {
+	log.CtxInfof(ctx, format, args...)
```
vadimberezniker, line 737: I believe this will lose the original line numbers. gorm has a util function (utils.FileWithLineNum) that it uses to figure out the caller. Maybe you could use that and prepend that to the beginning of the log line?

vadimberezniker (changes requested):

server/util/db/db.go
bduffany: Done 

image

vadimberezniker (approved):

PR #5201, opened 2023-11-06
[SS] Fix race condition / bug when copying an unmapped chunk

maggie-lou:
There were two bugs related to copying unmapped chunks: 
1) There could be a race condition where thread A) is initializing a chunk in the background and setting the source field (to something like RemoteCache) and thread B) is copying an unmapped chunk to be dirtied and reading the source field (Unmapped) to copy it over to the new chunk
2) When we copy an unmapped chunk, we create a clone of the chunk (So the new chunk is also unmapped). When we try to copy over the source data using newChunk.Write(bufCopiedFromSrc) (pseudocode), it tries to initialize the new chunk from the cache because it's unmapped. We don't need to do this, because we're overwriting the chunk's data with the contents of the source chunk.

Both these problems can be fixed by just initializing the source chunk before copying it. We will always initialize the source chunk anyway, because we have to read from it to copy over the data to the dest

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -506,9 +506,14 @@ func (s *COWStore) initDirtyChunk(offset int64, size int64) (ogChunk *Mmap, newC
 	ogChunk = s.chunks[offset]
 	chunkSource := snaputil.ChunkSourceHole
 	if ogChunk != nil {
+		if !ogChunk.safeReadMapped() {
+			if err := ogChunk.initMap(); err != nil {
+				return nil, nil, err
+			}
+		}
 		ogChunk.mu.RLock()
-		defer ogChunk.mu.RUnlock()
 		chunkSource = ogChunk.source
+		ogChunk.mu.RUnlock()
 	}
 	newChunk, err = NewMmapFd(s.ctx, s.env, s.DataDir(), fd, int(size), offset, chunkSource, s.remoteInstanceName)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #5201** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5201" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5201?utm_source=stack-comment).

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -506,6 +506,8 @@ func (s *COWStore) initDirtyChunk(offset int64, size int64) (ogChunk *Mmap, newC
 	ogChunk = s.chunks[offset]
 	chunkSource := snaputil.ChunkSourceHole
 	if ogChunk != nil {
+		ogChunk.mu.RLock()
+		defer ogChunk.mu.RUnlock()
```
tylerwilliams, line 510: maybe simpler to unlock after assigning to chunkSource?

maggie-lou: Done

maggie-lou: Actually ended up changing the implementation

tylerwilliams (approved):

bduffany (approved):

PR #5205, opened 2023-11-07
Don't create decompressor if destReader returns an error.

vadimberezniker:
Otherwise, Close() will wait forever for decompression to be done.

**Related issues**: N/A


tylerwilliams (approved):
Good find :)

PR #5206, opened 2023-11-07
Improve ExecuteWorkflow logs

bduffany:
* Reduce noise (don't log "polling" msg every 1s)
* Add execution ID and invocation ID to context

**Related issues**: N/A


tylerwilliams (commented):
This reminds me of log.Every(n, msg, fmt ...)

Do we need this log message at all though? It doesn't seem like it's doing much.

bduffany:
It's letting us know whether we're still polling for an invocation. Changed it to just log once at the beginning, that should give us the same info.

maggie-lou (approved):

PR #5207, opened 2023-11-07
Set the '--transient-store' flag automatically

bduffany:
The transient-store flag significantly improves performance, so it makes sense to enable it automatically whenever the podman version supports it.

**Related issues**: N/A


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -259,6 +271,24 @@ graphroot = "/var/lib/containers/storage"
 	}, nil
 }
 
+func getPodmanVersion(ctx context.Context) (*semver.Version, error) {
+	res := runPodman(ctx, "version", nil /*=stdio*/)
```
iain-macdonald, line 275: If you add `--format '{{.Client.Version}}'` it'll print just the client version and nothing else, which will save a little bit of code in this function, and it looks like that was added pre-1.0 so should work for all except the most out-of-date podman versions.

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -985,7 +1015,7 @@ func (c *podmanCommandContainer) State(ctx context.Context) (*rnpb.ContainerStat
 
 func runPodman(ctx context.Context, subCommand string, stdio *commandutil.Stdio, args ...string) *interfaces.CommandResult {
 	command := []string{"podman"}
-	if *transientStore {
+	if podmanVersion != nil && !podmanVersion.LessThan(transientStoreMinVersion) {
 		// Use transient store to reduce contention.
 		// See https://github.com/containers/podman/issues/19824
 		command = append(command, "--transient-store")
```
iain-macdonald, line 1021: wdyt about adding a logged-once-per-executor message when the podman version is too old (or not found) for this flag?

iain-macdonald (commented):
Sweet!

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: Nice! much cleaner.

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: It should always be set, except during the initial call to `runPodman` to get the version. I switched to just use `exec.Command().CombinedOutput()` though which seems cleaner.

bduffany: oh, re. it being too old, I will add that.

bduffany: done

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -190,6 +196,16 @@ func runSociStore(ctx context.Context) {
 }
 
 func NewProvider(env environment.Env, buildRoot string) (*Provider, error) {
+	v, err := getPodmanVersion(env.GetServerContext())
+	if err != nil {
+		return nil, status.WrapError(err, "podman version")
+	}
+	podmanVersion = v
```
iain-macdonald, line 203: Any reason not to just assign this above on line 199 and get rid of `v`?

iain-macdonald (approved):

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: then I would need a `var err error` - lately I'm preferring this explicit assignment pattern to that approach, as it feels a bit too subtle that either a global or local is being assigned based on whether it's `:=` or `=`

iain-macdonald: Ah shoot, didn't notice that `podmanVersion` isn't defined here.

PR #5209, opened 2023-11-07
Don't register container providers that are not enabled

bduffany:
Before this PR, we would call `podman.NewProvider()` regardless of whether podman was enabled (same for docker, bare, and sandbox). This is causing problems for https://github.com/buildbuddy-io/buildbuddy/pull/5207, which attempts to run the `podman` command within the `NewProvider()` func. If podman is not available (e.g. in tests), we'll fail to run podman to detect the version.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/platform/platform.go
+++ b/enterprise/server/remote_execution/platform/platform.go
@@ -210,6 +210,15 @@ type ExecutorProperties struct {
 	DefaultXcodeVersion     string
 }
 
+func (p *ExecutorProperties) SupportsIsolation(c ContainerType) bool {
+	for _, s := range p.SupportedIsolationTypes {
+		if s == c {
+			return true
+		}
+	}
+	return false
+}
+
 // ParseProperties parses the client provided properties into a struct.
 // Before use, the returned platform.Properties object *must* have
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/platform/platform.go
```diff
@@ -210,6 +210,15 @@ type ExecutorProperties struct {
...
 }
 
+func (p *ExecutorProperties) SupportsIsolation(c ContainerType) bool {
+	for _, s := range p.SupportedIsolationTypes {
+		if s == c {
+			return true
+		}
+	}
+	return false
+}
```
sluongng, line 220: nit: This feels like a future refactor `p.SupportedIsolationTypes` into `map[]struct{}`, instead of a slice, for easier lookup.

I think there is a `func contains()` in this same file that does something similar.

sluongng (approved):
LGTM

enterprise/server/remote_execution/platform/platform.go
bduffany: removed contains() in favor of the new `SupportsIsolation()`

iain-macdonald (approved):

PR #5210, opened 2023-11-07
invocation_stat_service: replace config package with a new interface method

sluongng:
Move the TrendsHeatmapEnabled check into
interfaces.InvocationStatService and the corresponding flag and
implementation into invocation_stat_service.go.

This remove an `//enterprise` dependency in static.go, pave way for
fixing FOSS build.


sluongng:
After giving this a bit more reading... I think we should create a new interface (in `interfaces.go`) for front-end static configs.

That way the values could be accessed from multiple places.

sluongng:
Let's not go with this approach.

PR #5211, opened 2023-11-07
.bazelignore: add .git

sluongng:
Turn out, if you just tail Bazel java log, it just scans everything.
Removing .git directory from scanning helps speed things up slightly.


bduffany (approved):
Bazel should ignore this automatically :sweat_smile:

PR #5213, opened 2023-11-07
go_sdk: upgrade to 1.21.4

sluongng:
Fixed a few Windows path/filepath CVEs

References:
https://github.com/golang/go/issues?q=milestone%3AGo1.21.4+label%3ACherryPickApproved

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -64,6 +64,9 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_download_sdk", "go_register_toolchai
 
 go_rules_dependencies()
 
+# Keep in sync with .github/workflows/checkstyle.yaml
+GO_SDK_VERSION = "1.21.4"
+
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -72,42 +72,42 @@ go_download_sdk(
     name = "go_sdk_linux",
     goarch = "amd64",
     goos = "linux",
-    version = "1.21.3",  # Keep in sync with .github/workflows/checkstyle.yaml
+    version = "1.21.4",  # Keep in sync with .github/workflows/checkstyle.yaml
```
vadimberezniker, line 75: should we use a variable for the SDK version?

vadimberezniker (approved):

WORKSPACE
sluongng: Good idea. Should make the diff smaller.

Pushed a change for it.

vadimberezniker (approved):

PR #5215, opened 2023-11-07
[FC] Fix early chroot cleanup in firecracker.Remove()

maggie-lou:
This caused an error where we'd clean up the contents of the memory store before stopping the uffd handler, causing errors when the uffd handler tried to access the memory store.

This only happened on executor shut downs because in a typical case we'd call Pause() on the container first, which stops the uffd handler. During executor shut downs, we only call Remove(), so the uffd handler was still running when we clear chroot

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -2211,10 +2211,6 @@ func (c *FirecrackerContainer) remove(ctx context.Context) error {
 		c.rootStore = nil
 	}
 
-	if err := os.RemoveAll(filepath.Dir(c.getChroot())); err != nil {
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #5215** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5215" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5215?utm_source=stack-comment).

bduffany (approved):
great find :)

tylerwilliams (approved):
nice

PR #5222, opened 2023-11-07
[SS] Fix bug where we aren't saving snapshots

maggie-lou:
**Related issues**: N/A


maggie-lou:
* master
  * **PR #5222** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5222" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5222?utm_source=stack-comment).

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -1345,12 +1342,27 @@ func (p *pool) TryRecycle(ctx context.Context, r interfaces.Runner, finishedClea
...
+	// Clean the workspace before recycling the runner (to save on disk space).
 	if err := cr.Workspace.Clean(); err != nil {
 		log.CtxErrorf(ctx, "Failed to recycle runner %s: failed to clean workspace: %s", cr, err)
 		return
 	}
+
+	// Don't add snapshot enabled runners back to the pool because we don't need
+	// the pool logic for them. Just save the snapshot with `Container.Pause`
+	// Note: We want recycled=false here so that the container is removed in the
+	// background
```
bduffany, line 1354: nit: Pause() calls Remove(), but this logic still works because the Remove() is wrapped in a sync.Once. Doesn't matter a whole lot either way but you could remove the note here or reword it like:

```suggestion
	// Don't add snapshot enabled runners back to the pool because we don't need
	// the pool logic for them. Just save the snapshot with `Container.Pause`,
	// which also removes the container.
```

bduffany (approved):

enterprise/server/remote_execution/runner/runner.go
bduffany: as a side note, I have also gotten confused by this before - maybe Pause() just shouldn't call Remove() to match the behavior of the other container types.

maggie-lou: Ah right - modified the comment

PR #5230, opened 2023-11-08
Fix checkstyle date command on mac

bduffany:
Makes `bazel run tools/checkstyle` work at least somewhat on Mac locally. The `clang-format` binary isn't yet set up properly for Mac so that part still fails.

**Related issues**: N/A


sluongng (approved):
LGTM.

Tested on my Mac laptop and our CI should cover Linux.

PR #5240, opened 2023-11-08
Create multiple gRPC connections by default when dialing targets.

vadimberezniker:
**Related issues**: N/A


enterprise/server/cmd/executor/executor.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/disk"
 	"github.com/buildbuddy-io/buildbuddy/server/util/flag"
 	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
+	_ "github.com/buildbuddy-io/buildbuddy/server/util/grpc_server" // imported for grpc_port flag definition to avoid breaking old configs; DO NOT REMOVE.
```
tylerwilliams, line 34: nit: could we move the side-effect imports down with the other named ones?

tylerwilliams (approved):
neat

PR #5242, opened 2023-11-09
call stop() when grpc-over-http is enabled

luluz66:
**Related issues**: N/A
When grpc over http is enabled, and if during shutdown there are existing http
connections, it can cause panic because Drain() is not implemented by the
transport returned by ServeHTTP()


luluz66:
Closing this Pull Request since we decide to patch grpc package to remove Drain in https://github.com/buildbuddy-io/buildbuddy/pull/5277

PR #5254, opened 2023-11-09
Sort menu tweaks

iain-macdonald:
This PR:
* Renders the little "X" button when non-default sort parameters are selected, so the user can clear them.
* Marks the default sort options with a "(Default)" label and selects their radio buttons when they're active, even if no URL parameters are set.
* Hides the UI label and clears URL parameters when the default sort options are selected, even if other sort options were previously selected.

**Related issues**: [1954](https://github.com/buildbuddy-io/buildbuddy-internal/issues/1954)


bduffany (approved):
:tada:

PR #5256, opened 2023-11-09
Add GIT_BRANCH to workflow task routing key

bduffany:
When snapshot sharing is enabled, we want workflows to be matched to snapshots that recently ran the workflow associated with the workflow's git branch. GIT_BRANCH is part of env vars, not platform, so we need a change to add that in to the routing key.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/task_router/task_router.go
+++ b/enterprise/server/scheduling/task_router/task_router.go
@@ -228,8 +228,7 @@ func (runnerRecycler) Applies(params routingParams) bool {
 }
 
 func (runnerRecycler) preferredNodeLimit(params routingParams) int {
-	workflowID := platform.FindValue(params.cmd.GetPlatform(), platform.WorkflowIDPropertyName)
-	if workflowID != "" {
+	if isWorkflow(params.cmd) {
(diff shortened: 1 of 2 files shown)


enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -212,11 +253,17 @@ type Router interface {
...
+	// provided routing parameters.
+	//
+	// The preferredNodeLimit is the number of preferred executor nodes that
+	// should be used.
+	//
+	// The routing keys are the Redis keys where the list of preferred executor
+	// nodes are stored. When marking a task complete, only the first key is
+	// used to update the node list. When ranking nodes, keys should be treated
+	// in order of most preferred to least preferred.
+	RoutingInfo(params routingParams) (int, []string, error)
```
tylerwilliams, line 266: can we start with a single routingKey and see if there's an issue?

bduffany: Done

maggie-lou (approved):

tylerwilliams (approved):

PR #5260, opened 2023-11-09
Fix nil panic in CAS server if an empty digest is in the batch.

vadimberezniker:
**Related issues**: N/A


server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -322,6 +322,7 @@ func (s *ContentAddressableStorageServer) BatchReadBlobs(ctx context.Context, re
 				Digest: rn.GetDigest(),
 				Status: &statuspb.Status{Code: int32(codes.OK)},
 			})
+			closeTrackerData = append(closeTrackerData, downloadTrackerData{})
```
tylerwilliams, line 325: wdyt about doing this versus moving the line on 306

`		closeTrackerFuncs = append(closeTrackerFuncs, func(data downloadTrackerData) {
			downloadTracker.CloseWithBytesTransferred(int64(data.bytesReadFromCache), int64(data.bytesDownloadedToClient), data.compressor, "cas_server")
		})`

down into the if !rn.IsEmpty() block?

tylerwilliams (approved):

Issue #5248, opened 2023-11-09
[CLI] Queries break if for some common attributes


erikkerber:
**Describe the bug**

In our `.bazelrc` we had JDK 11 specified under `common`:

```
common --java_runtime_version=remotejdk_11
common --tool_java_runtime_version=remotejdk_11
common --repo_env=JAVA_HOME=../bazel_tools/jdk
```

When installing the CLI via `.bazelversion`, queries then broke with an invalid option applied:

```
ERROR: --java_runtime_version=remotejdk_11 :: Unrecognized option: --java_runtime_version=remotejdk_11
```

Changing those to `build` resolves, but it feels like a bug that the wrapper isn't a pure passthrough in this instance

**To Reproduce**
(Above)

**Expected behavior**

The wrapper is completely transparent to regular Bazel operations

bduffany:
@erikkerber the fix for this is now released in CLI version 5.0.25 !

PR #5263, opened 2023-11-09
cli: improve 'common' config parsing

bduffany:
Ensures that `common` flags apply only to commands that support them.

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy/issues/5248


cli/parser/parser.go
```diff
@@ -314,6 +314,35 @@ type CommandLineSchema struct {
 	// allow us to show plugin-specific help.
 }
 
+// CommandSupportsOpt returns whether the given opt is in the CommandOptions.
+// The opt is expected to be either "--NAME" or "-SHORTNAME", without an "=".
+func (s *CommandLineSchema) CommandSupportsOpt(opt string) bool {
+	// TODO: this func is using heuristics, since a correct impl would require
+	// us to know the schema for all bazel commands. At some point we should
+	// probably just do a one-time parse of all the commands so that we can do
+	// this properly, or see if Bazel can give us a dump of its option schema.
```
siggisim, line 323: Might be worth taking a look at https://github.com/bazelbuild/bazel/pull/19536#issuecomment-1793249509

It mentions a file called `[bazel_flags.proto](https://github.com/bazelbuild/bazel/pull/19536/files#diff-289cfc38fc9fe94347bca949249e42e12870d4d310d0c60c39cd266aba16f0b9)` that we could potentially use

siggisim (approved):

cli/parser/parser.go
bduffany: Ah looks like we can run `bazel help flags-as-proto` and it dumps a base64-encoded FlagCollection proto. Will save that for a future PR since it'll involve a pretty substantial rewrite.

sluongng:
I tried this change out today

```
@@ -40,7 +39,7 @@ build:remote-shared --verbose_failures
 build:remote-shared --platforms=//platforms:linux_x86_64
 build:remote-shared --extra_execution_platforms=//platforms:linux_x86_64

-build --incompatible_enable_cc_toolchain_resolution
+common --incompatible_enable_cc_toolchain_resolution
```

with `bb test --config=remote` and it seems like the actual invocation does not use the flag at all.

PR #5264, opened 2023-11-10
Refactor trends frontend config

sluongng:
This PR includes 2 refactoring commits:

- invocation_stat_service/config: move flag to server/static
  This enable building //server/static with one fewer //enterprise
  dependency.

- static: refactor FrontendConfig boolean condition
  Picked the more complicated configs and moved them to their own block of
  code. No logical modification were made.

**Related issues**: N/A


enterprise/server/invocation_stat_service/invocation_stat_service.go
```diff
@@ -10,10 +10,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/buildbuddy-io/buildbuddy/enterprise/server/invocation_stat_service/config"
 	"github.com/buildbuddy-io/buildbuddy/server/build_event_protocol/invocation_format"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
+	"github.com/buildbuddy-io/buildbuddy/server/static"
```
siggisim, line 16: invocation_stat_service shouldn't depend on static, feels like a recipe for future circular dependencies.

Can we just move inovcation_stat_service/config from enterprise to server like we did here: https://github.com/buildbuddy-io/buildbuddy/pull/5245

siggisim (changes requested):

enterprise/server/invocation_stat_service/invocation_stat_service.go
sluongng: I tried that first. But then I realized that we have several Trends/Heatmap flags that are scattered in different places and `static` is a convenience place to group them up.

Let me move the `trends_heatmap_enabled` flag back out again

server/static/static.go
```diff
@@ -193,6 +180,34 @@ func serveIndexTemplate(ctx context.Context, env environment.Env, tpl *template.
 		PopupAuthEnabled:                       *popupAuthEnabled,
 	}
 
+	if auth := env.GetAuthenticator(); auth != nil {
```
siggisim, line 183: Why not just leave these in place above? I find the current approach much more readable.

siggisim (changes requested):

server/static/static.go
jdhollen: i agree, even though it's repetitive.

sluongng: Getting rid of this.

siggisim (approved):

PR #5265, opened 2023-11-10
auditlog: move helper functions to server/auditlog_util

sluongng:
This allows us to build //server/buildbuddy_server with one fewer
//enterprise dependency.

### DIFF
--- a/server/interfaces/interfaces.go
+++ b/server/interfaces/interfaces.go
@@ -1214,9 +1214,9 @@ type ConfigSecretProvider interface {
 
 type AuditLogger interface {
 	Log(ctx context.Context, resource *alpb.ResourceID, action alpb.Action, request proto.Message)
-	LogWithGroup(ctx context.Context, groupID string, action alpb.Action, request proto.Message)
-	LogWithInvocation(ctx context.Context, invocationID string, action alpb.Action, request proto.Message)
-	LogWithSecret(ctx context.Context, secretName string, action alpb.Action, request proto.Message)
+	LogForGroup(ctx context.Context, groupID string, action alpb.Action, request proto.Message)
+	LogForInvocation(ctx context.Context, invocationID string, action alpb.Action, request proto.Message)
+	LogForSecret(ctx context.Context, secretName string, action alpb.Action, request proto.Message)
 	GetLogs(ctx context.Context, req *alpb.GetAuditLogsRequest) (*alpb.GetAuditLogsResponse, error)
 }
 
--- a/server/interfaces/interfaces.go
+++ b/server/interfaces/interfaces.go
@@ -1214,6 +1214,9 @@ type ConfigSecretProvider interface {
 
 type AuditLogger interface {
 	Log(ctx context.Context, resource *alpb.ResourceID, action alpb.Action, request proto.Message)
+	LogWithGroup(ctx context.Context, groupID string, action alpb.Action, request proto.Message)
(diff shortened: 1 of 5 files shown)


server/auditlog_util/auditlog_util.go
```diff
@@ -0,0 +1,26 @@
+package auditlog_util
```
siggisim, line 1: naming something `util` is always tricky, because it's hard to tell what belongs in util and what belongs in auditlog.

what if we just added 3 methods to auditlog

`al.LogWithResource`
`al.LogWithSecret`
`al.LogWithInvocation`

alternatively, we could just remove these methods altogether and either have `al.Log` take a type and a proto i.e.
```
al.Log(ctx, auditlog_util.InvocationResourceID(req.GetInvocationId()), alpb.Action_UPDATE, req)
```
becomes (which is actually less characters)
```
al.Log(ctx, alpb.ResourceType_INVOCATION, req.GetInvocationId(), alpb.Action_UPDATE, req)
```

siggisim (changes requested):

server/auditlog_util/auditlog_util.go
sluongng: Taking a type and a string ID won't work for the more elaborate resource types such as Group API Key that requires more than ID to initialize resource. For example, this is currently being used

```go
		rid := &alpb.ResourceID{
			Type: alpb.ResourceType_GROUP_API_KEY,
			Id:   req.GetApiKeyId(),
			Name: key.Label,
		}
		al.Log(ctx, rid, alpb.Action_ACCESS, req)
```

I added 1 new method `al.LogWithResourceType` instead.

server/interfaces/interfaces.go
```diff
@@ -1206,6 +1206,7 @@ type ConfigSecretProvider interface {
 
 type AuditLogger interface {
 	Log(ctx context.Context, resource *alpb.ResourceID, action alpb.Action, request proto.Message)
+	LogWithResourceType(ctx context.Context, resourceType alpb.ResourceType, id string, action alpb.Action, request proto.Message)
```
vadimberezniker, line 1209: If we're going to add a convenience method on the interface, we might as well go all the way and add type-specific functions, ie. `LogForGroup`, `LogForInvocation`, etc

vadimberezniker (changes requested):

siggisim (approved):

server/interfaces/interfaces.go
sluongng: This is done.

sluongng: @vadimberezniker PTAL 🙏

server/interfaces/interfaces.go
```diff
@@ -1214,6 +1214,9 @@ type ConfigSecretProvider interface {
 
 type AuditLogger interface {
 	Log(ctx context.Context, resource *alpb.ResourceID, action alpb.Action, request proto.Message)
+	LogWithGroup(ctx context.Context, groupID string, action alpb.Action, request proto.Message)
```
vadimberezniker, line 1217: Can you change `With` to `For`? These functions generate audit events /for/ a specific resource type, rather than /with/ a specific resource.

vadimberezniker (approved):

server/interfaces/interfaces.go
sluongng: Pushed a change for this

vadimberezniker (approved):

PR #5276, opened 2023-11-13
Update local grafana setup for multiple dashboards

bduffany:
* Makes it so that we don't have to fiddle with `GRAFANA_DASHBOARD_FILE_PATH` in order to edit the non-default dashboards. Now, we can just edit any dashboard and it will be synced back to the appropriate JSON file automatically.
* Rewrite `run.sh` in Go since the shell script was getting too arcane. We can now run it with `bazel run tools/metrics/grafana`
* Update normalization script to remove unstable `id` fields, and store the JSON filename in the `slug` field.

**Related issues**: N/A


tools/metrics/grafana/grafana.go
```diff
@@ -0,0 +1,202 @@
...
+	"syscall"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+)
+
+var (
+	kube       = flag.Bool("kube", false, "Use kubectl port-forward to point Grafana at real data.")
+	namespace  = flag.String("namespace", "monitor-dev", "k8s namespace")
+	deployment = flag.String("deployment", "prometheus-global-server", "Prometheus server resource to port-forward to.")
```
luluz66, line 28: maybe add a comment that these two flags only take effect when kube is true.

bduffany: done

luluz66 (approved):

PR #5277, opened 2023-11-13
Patch out panic from gRPC Drain when handling grpc-over-http requests.

vadimberezniker:
**Related issues**: N/A


luluz66 (approved):

tylerwilliams (approved):
do we have a long term fix here that doesn't require patching grpc forever?

vadimberezniker:
> do we have a long term fix here that doesn't require patching grpc forever?

Unfortunately I don't think there's an alternative. We would either need to convince gRPC devs to change this (which is unlikely) or stop using gRPC-over-HTTP which is not great either.

PR #5278, opened 2023-11-13
Handle already-existing symlinks without error

tylerwilliams:
If a symlink exists with the correct target; use it, and if the target is incorrect, attempt to remove it and recreate.

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -857,6 +857,23 @@ type DownloadTreeOpts struct {
 	TrackTransfers bool
 }
 
+// To be used when a symlink exists and an Exists error has been returned, but
+// the caller wants to ensure the *correct* file has been linked.
+func checkSymlink(oldName, newName string) bool {
+	pointee, err := os.Readlink(newName)
+	if err != nil {
+		return false
+	}
+	return pointee == oldName
+}
+
+func removeSymlink(name string) error {
+	if _, err := os.Lstat(name); err != nil {
+		return err
+	}
+	return os.Remove(name)
+}
+
 func DownloadTree(ctx context.Context, env environment.Env, instanceName string, digestFunction repb.DigestFunction_Value, tree *repb.Tree, rootDir string, opts *DownloadTreeOpts) (*TransferInfo, error) {
 	txInfo := &TransferInfo{}
 	startTime := time.Now()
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -928,6 +945,24 @@ func DownloadTree(ctx context.Context, env environment.Env, instanceName string,
...
+					// If the symlink exists and is correct;
+					// skip the error.
+					if checkSymlink(symlinkNode.GetTarget(), nodeAbsPath) {
+						continue
+					}
+					// If the symlink exists but was
+					// incorrect; try to remove it
+					if err := removeSymlink(nodeAbsPath); err != nil {
+						return err
+					}
```
bduffany, line 958: wdyt about replacing these two things w/ `os.RemoveAll(nodeAbsPath)`? This way even if the existing path is not a symlink we can just obliterate it and place the correct symlink on top of it.

bduffany: even if a directory already exists at the path, we ultimately want to replace it with a symlink, so I think RemoveAll() is probably the right thing to do.

Also I wasn't sure so had to double check - if the existing path is a symlink to another directory, RemoveAll() does not remove the directory that it points to. It just removes the link.

tylerwilliams: yeah same haha just double checked that. anyway this is Done.

siggisim (approved):

enterprise/server/remote_execution/dirtools/dirtools_test.go
```diff
@@ -722,6 +722,180 @@ func TestDownloadTreeEmptyDigest(t *testing.T) {
...
+				Digest: fileADigest,
+			},
+		},
+		Symlinks: []*repb.SymlinkNode{
+			&repb.SymlinkNode{
+				Name:   "fileA.symlink",
+				Target: "./fileA.txt",
+			},
+		},
+	}
```
bduffany, line 818: nit: the `childDir` makes the test slightly hard to follow - if you wanted you could have this be the top-level dir.

enterprise/server/remote_execution/dirtools/dirtools_test.go
```diff
@@ -722,6 +722,180 @@ func TestDownloadTreeEmptyDigest(t *testing.T) {
...
+	info, err := dirtools.DownloadTree(ctx, env, "", repb.DigestFunction_SHA256, directory, tmpDir, &dirtools.DownloadTreeOpts{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.NotNil(t, info, "transfers are not nil")
+	assert.Equal(t, int64(0), info.FileCount, "zero files were transferred")
+	assert.DirExists(t, filepath.Join(tmpDir, "my-directory"), "my-directory should exist")
+	assert.FileExists(t, filepath.Join(tmpDir, "my-directory/fileA.txt"), "fileA.txt should exist")
+	assert.FileExists(t, filepath.Join(tmpDir, "fileB.txt"), "fileB.txt should exist")
```
bduffany, line 896: nit: add an assertion about the new symlink target for fileA.symlink?

bduffany (approved):

enterprise/server/remote_execution/dirtools/dirtools_test.go
tylerwilliams: Done.

enterprise/server/remote_execution/dirtools/dirtools_test.go
tylerwilliams: Done.

PR #5281, opened 2023-11-13
Attempt to alloc less in GCS uploads

tylerwilliams:
tl;dr: Set GCS chunk size lower when we know the chunk size ahead of time and the chunk is smaller than the library default (16MB).

Longer version:
I noticed in profiles that this code was hot:

image

and that it copies the buffer being uploaded. The docs about `ChunkSize` say:

```
	// ChunkSize controls the maximum number of bytes of the object that the
	// Writer will attempt to send to the server in a single request. Objects
	// smaller than the size will be sent in a single request, while larger
	// objects will be split over multiple requests. The value will be rounded up
	// to the nearest multiple of 256K. The default ChunkSize is 16MiB.
	//
	// Each Writer will internally allocate a buffer of size ChunkSize. This is
	// used to buffer input data and allow for the input to be sent again if a
	// request must be retried.
	//
	// If you upload small objects (< 16MiB), you should set ChunkSize
	// to a value slightly larger than the objects' sizes to avoid memory bloat.
	// This is especially important if you are uploading many small objects
	// concurrently. See
	// https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload#size
	// for more information about performance trade-offs related to ChunkSize.
	//
	// If ChunkSize is set to zero, chunking will be disabled and the object will
	// be uploaded in a single request without the use of a buffer. This will
	// further reduce memory used during uploads, but will also prevent the writer
	// from retrying in case of a transient error from the server or resuming an
	// upload that fails midway through, since the buffer is required in order to
	// retry the failed request.
	//
	// ChunkSize must be set before the first Write call.
	ChunkSize [int](https://pkg.go.dev/builtin#int)
```

so I think it makes sense to try to set this lower when blobs are small. The metrics say that 99% of blobs are less than 1MB.

image

There is a further optimization we can do here which is to disable the internal-buffering entirely, but because that disables automatic retries, we'd need to see if it's an issue and possibly retry ourselves.


server/backends/blobstore/gcs/gcs.go
```diff
@@ -108,13 +109,27 @@ func (g *GCSBlobStore) ReadBlob(ctx context.Context, blobName string) ([]byte, e
 	return util.Decompress(b, err)
 }
 
+const gcsDefaultChunkSize = int(16e6) // 16MB
```
vadimberezniker, line 112: It looks like this constant is available as googleapi.DefaultUploadChunkSize

vadimberezniker (approved):

server/backends/blobstore/gcs/gcs.go
tylerwilliams: Your codesearch skillz are way better than mine :) Thanks!

jdhollen (approved):

PR #5290, opened 2023-11-13
podman: determine cgroup paths automatically

bduffany:
* Move all the cgroup stuff from `podman.go` to a new package `cgroup.go` since it's growing a bit, and we also want to use this for Docker.
* Instead of requiring the cgroup flags, walk the /sys/fs/cgroup tree, searching for the cid in the path, and determine mem/cpu paths from that.
  * The dir scanning approach is used because there are several different variables that determine what the paths will actually be - whether systemd is used or not, whether we're on cgroup v1 or v2, whether podman is running rootless or not... maybe other stuff (that may possibly change in the future?) I figured it was too complicated and error-prone to try and take all of these things into account.
  * In terms of performance, it takes 8ms to scan the cgroupfs tree (tested on a GCP VM). It succeeds very quickly and manages to init the cgroup paths on the first try, even when running a NOP task (/bin/true). We only do the scan once, so this overhead is incurred only on the first task executed.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/cgroup/cgroup.go
+++ b/enterprise/server/remote_execution/cgroup/cgroup.go
@@ -1,0 +1,189 @@
+package cgroup
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+)
+
+const (
+	// Standard path where cgroupfs is expected to be mounted.
+	cgroupfsPath = "/sys/fs/cgroup"
+
+	// Placeholder value representing the container ID in cgroup path templates.
+	cidPlaceholder = "{{.ContainerID}}"
+)
+
+// Paths holds cgroup path templates that map a container ID to their cgroupfs
+// file paths.
+//
+// A single instance should be shared across all containers of the same
+// isolation type, since the first call to Stats() walks the cgroupfs tree in
+// order to discover the cgroup path locations.
+type Paths struct {
+	mu sync.RWMutex
+
+	// MemoryTemplate is the path template for the cgroupfs file containing
+	// current memory usage.
+	MemoryTemplate string
+	// CPUPathTemplate is the path template for the cgroupsfs file containing
+	// cumulative CPU usage.
+	CPUTemplate string
+}
+
+func (p *Paths) Stats(ctx context.Context, cid string) (*repb.UsageStats, error) {
+	if err := p.find(ctx, cid); err != nil {
+		return nil, err
+	}
+
+	memUsagePath := strings.ReplaceAll(p.MemoryTemplate, cidPlaceholder, cid)
+	cpuUsagePath := strings.ReplaceAll(p.CPUTemplate, cidPlaceholder, cid)
+
+	memUsageBytes, err := readInt64FromFile(memUsagePath)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -0,0 +1,189 @@
...
+
+	memUsagePath := strings.ReplaceAll(p.MemoryTemplate, cidPlaceholder, cid)
+	cpuUsagePath := strings.ReplaceAll(p.CPUTemplate, cidPlaceholder, cid)
+
+	memUsageBytes, err := readInt64FromFile(memUsagePath)
+	if err != nil {
+		return nil, err
+	}
+	var cpuNanos int64
+	if p.CgroupVersion() == 1 {
```
iain-macdonald, line 61: guard against 0 here?

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -394,12 +400,19 @@ type podmanCommandContainer struct {
 	removed bool
 }
 
+// TODO: remove this and always use provider.New() instead.
 func NewPodmanCommandContainer(env environment.Env, image, buildRoot string, options *PodmanOptions) container.CommandContainer {
+	imageExistsCache, _ := newImageExistsCache()
```
iain-macdonald, line 405: I think you wanna pass this in from the Provider so it's shared across CommandContainers, right?

iain-macdonald (commented):
couple nits but overall looks good!

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: find() returns an error in that case, so I think it's not needed.

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: provider.New() doesn't call this func unfortunately - this func is only used for testing. Line 403 has a TODO to remove this func and only use the provider, even in tests

iain-macdonald: Oof, got me! Maybe one of us should get on that TODO soon 😛

enterprise/server/remote_execution/cgroup/cgroup.go
tylerwilliams: lol i agree with @iain-macdonald 

i have a small brain and seeing:

```
if 1 {
} else if 2 {
} else {
 raise error(get out)
}
```

makes me feel safe

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -0,0 +1,189 @@
...
+func readInt64FromFile(path string) (int64, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return 0, err
+	}
+	defer f.Close()
+	b, err := io.ReadAll(f)
+	if err != nil {
+		return 0, err
+	}
```
tylerwilliams, line 161: can you use `data, err := os.ReadFile(path)` here?

enterprise/server/remote_execution/containers/podman/podman.go
tylerwilliams: +1

tylerwilliams (approved):
couple nits but looks great!

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: I can do it :) had started on it in this PR, but then it started getting super big so I split it into a separate branch

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: yep done

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: okay, fair :) done

PR #5292, opened 2023-11-14
Use Profile Guided Optimization (PGO) to build auto-release artifact

tylerwilliams:
[Profile Guided Optimization](https://go.dev/doc/pgo) (PGO) allows feeding pprof files into the compiler to better optimize compilation. It is supposed to improve performance anywhere from 2-7%.

This CL add a small shell script to collect profiles and merge them, and adds a merged set of 16 prod profiles. It adds the pgo flag so when auto-release artifacts are built, PGO is used to build them.


luluz66:
The doc said the profiles need to be representative. Wondering if it will help if we collect the profiles during peak usage time in europe and in us instead of collecting profiles right before the release?

vadimberezniker (approved):

PR #5293, opened 2023-11-14
store the host_id next to the file_cache

tempoz:
It was sort of awkward to make the host_id depend on the filecache for its location. Had to add a `HostIDFilepath() string` function to the `FileCache` interface, but it works. If anyone has better ideas about how this should be structured, let me know. 

Fixes buildbuddy-io/buildbuddy-internal#2142


tempoz:
* master
  * **PR #5293** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5293?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5293?utm_source=stack-comment).

tempoz:
Could maybe use a `PersistentConfigDir` on the `FileCache` instead, and then only pass in the directory to the `GetHostID` function?

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -68,11 +68,11 @@ type Options struct {
 func NewExecutor(env environment.Env, id string, runnerPool interfaces.RunnerPool, options *Options) (*Executor, error) {
 	hostID := options.NameOverride
 	if hostID == "" {
-		if h, err := uuid.GetHostID(); err == nil {
+		if h, err := uuid.GetHostID(env.GetFileCache().HostIDFilepath()); err == nil {
```
tylerwilliams, line 71: It doesn't seem like this needs to be tied to the filecache at all, but maybe I'm missing some context. (Let me know if so!)

Instead I think this should be the "rootDirectory" configured in the executor config, and/or a different one for the app?

bduffany: I suggested putting it under the filecache dir because it was guaranteed to have the same lifecycle as the filecache - so if the filecache is persistent, the host_id will also be persistent. I think it's probably fine to put it in executor.root_directory though since that probably also has the same lifecycle as the filecache - otherwise things like hardlinking probably wouldn't work? (since different lifecycle is probably means different FS)

So +1 to putting it under executor.root_directory

tempoz: thanks, done!

server/real_environment/real_environment.go
```diff
@@ -114,7 +114,11 @@ type RealEnv struct {
 	serverIdentityService            interfaces.ClientIdentityService
 	imageCacheAuthenticator          interfaces.ImageCacheAuthenticator
 	serverNotificationService        interfaces.ServerNotificationService
+<<<<<<< HEAD
```
tylerwilliams, line 117: whoops

enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -103,14 +104,14 @@ func evictFn(v *entry, reason lru.EvictionReason) {
 // NewFileCache constructs an fileCache with maxSize that will cache files
```
tylerwilliams, line 104: Can the filecache changes can be reverted now?

server/real_environment/real_environment.go
tempoz: lol yup.

enterprise/server/remote_execution/filecache/filecache.go
tempoz: No, the file cache needs to pass the directory to `uuid.GetHostID`, since the directory has to be stored in the environment since it is coming from a flag in `enterprise`.

tempoz:
If `server/remote_cache/action_cache_server/action_cache_server.go` didn't use the host id, we wouldn't need to pipe the directory around so much, but c'est la vie

server/util/uuid/uuid.go
```diff
@@ -156,6 +152,18 @@ func GetFailsafeHostID() string {
...
+func MigrateHostID(newConfigDir string) {
+	oldConfigDir, err := oldConfigDir()
+	if err != nil {
+		return
+	}
+	oldHostIDFilepath := path.Join(oldConfigDir, hostIDFilename)
+	newHostIDFilepath := path.Join(newConfigDir, hostIDFilename)
+	// This call will move an old-style host id file (if it exists) to the new
+	// host id file location.
+	os.Rename(oldHostIDFilepath, newHostIDFilepath)
```
bduffany, line 164: nit: use `disk.MoveFile`, since the files will likely be on different filesystems and `rename` could fail.

Also I think it's slightly safer to copy the file if it exists, rather than rename, in case for some reason the build root gets wiped but the home dir does not. This is a bit messy but IMO it's better to be overly cautious here to avoid wiping peoples' caches as much as we can

server/hostid/hostid.go
```diff
@@ -0,0 +1,130 @@
...
+		dir = d
+	} else {
+		// migrate the host_id, if possible
+		// TODO (zoey): remove this after we expect most migrations to be done
+		if oldConfigDir, err := configDir(); err == nil {
+			oldHostIDFilepath := path.Join(oldConfigDir, hostIDFilename)
+			newHostIDFilepath := path.Join(dir, hostIDFilename)
+			// This call will move an old-style host id file (if it exists) to the new
+			// host id file location.
+			_ = disk.MoveFile(oldHostIDFilepath, newHostIDFilepath)
```
bduffany, line 68: sorry if my last comment was not 100% clear, I think we should make `disk.copyFileViaTempSibling` public and use that here instead. otherwise LGTM

bduffany (approved):

server/hostid/hostid.go
tempoz: kk, will do

tempoz: done.

PR #5294, opened 2023-11-14
[CLI ] Add support for MODULE.bazel files

siggisim:
This adds support for MODULE.bazel files in the CLI in a few places:
- When trying to find the workspace root when there is no WORKSPACE file present
- When auto-generating a MODULE.bazel file in Bazel 6.4 or later when bootstrapping a non-bazel repository
- When running `bb add` for example `bb add rules_go` when a MODULE.bazel file exists
- When converting from a MODULE.yaml => MODULE.bazel file

### DIFF
--- a/cli/workspace/workspace.go
+++ b/cli/workspace/workspace.go
@@ -114,7 +114,6 @@ func useModules() bool {
 	lines := strings.Split(strings.TrimSpace(string(data)), "\n")
 	version := lines[len(lines)-1]
 
-	log.Printf(version)
 	if version == "" || version == "latest" {
 		return true
 	}
--- a/cli/workspace/workspace.go
+++ b/cli/workspace/workspace.go
@@ -70,7 +70,7 @@ func CreateWorkspaceIfNotExists() (string, string, error) {
 	if err == nil {
 		return path, base, nil
 	}
-	
+
 	useModules := useModules()
 
 	fileName := "MODULE.bazel"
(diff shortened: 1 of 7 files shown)


cli/workspace/workspace.go
```diff
@@ -79,16 +89,53 @@ func CreateWorkspaceFileIfNotExists() (string, string, error) {
...
+	log.Debugf("Created %s file at %s/%s", fileName, pathVal, basename)
 	return pathVal, basename, nil
 }
 
+func useModules() bool {
+	data, err := os.ReadFile(".bazelversion")
+	if err != nil {
+		return true
+	}
+
```
bduffany, line 113: we have `bazelisk.ResolveVersion`, would that work here?

cli/workspace/workspace.go
```diff
@@ -79,16 +89,53 @@ func CreateWorkspaceFileIfNotExists() (string, string, error) {
...
+	lines := strings.Split(strings.TrimSpace(string(data)), "\n")
+	version := lines[len(lines)-1]
+
+	log.Printf(version)
```
bduffany, line 117: rm

bduffany (approved):

cli/workspace/workspace.go
siggisim: I tried that at first, but it caused a circular dependency (because `bazelisk.go` depends on `workspace.go`).

I'd like to nuke the whole WORKSPACE generation bit at some point when I'm more confident with the module side of things - so I think this is probably alright for now.

cli/workspace/workspace.go
siggisim: Done

PR #5305, opened 2023-11-15
Executor metadatadir flag

tylerwilliams:


enterprise/server/cmd/executor/executor.go
```diff
@@ -50,13 +52,14 @@ import (
...
-	localCacheSizeBytes      = flag.Int64("executor.local_cache_size_bytes", 1_000_000_000 /* 1 GB */, "The maximum size, in bytes, to use for the local on-disk cache")
-	startupWarmupMaxWaitSecs = flag.Int64("executor.startup_warmup_max_wait_secs", 0, "Maximum time to block startup while waiting for default image to be pulled. Default is no wait.")
+	appTarget                 = flag.String("executor.app_target", "grpcs://remote.buildbuddy.io", "The GRPC url of a buildbuddy app server.")
+	disableLocalCache         = flag.Bool("executor.disable_local_cache", false, "If true, a local file cache will not be used.")
+	deleteFileCacheOnStartup  = flag.Bool("executor.delete_filecache_on_startup", false, "If true, delete the file cache on startup")
+	deleteBuildRootOnStartup  = flag.Bool("executor.delete_build_root_on_startup", false, "If true, delete the build root on startup")
+	executorMedadataDirectory = flag.String("executor.metadata_directory", "", "Location where executor host_id and other metadata is stored. Defaults to executor.local_cache_directory/../")
```
bduffany, line 59: nit: if we do add other stuff here besides the host ID, things could get a bit messy - there will be a bunch of misc data files as sibling files of the cache and build_root directories, making it harder to do things like "just wipe the metadata, but keep filecache + build root" or "list all metadata files" etc.

I think it would be a bit more organized to have this default to "${local_cache_dir}/../metadata"

bduffany (approved):

enterprise/server/cmd/executor/executor.go
tylerwilliams: Done.

PR #5311, opened 2023-11-15
Separate `gplink` from `oidc.go`

siggisim:
Adds a new endpoint `/auth/gcp/link/` that sends folks through the Google Cloud scope acceptance flow (rather than hijacking the `/login/` and `/auth/` flow).

This also fixes the issue where we'd use a different gcp client for linking vs exchanging auth tokens.


tylerwilliams (approved):
nice

PR #5313, opened 2023-11-15
Add two new metrics for tracking estimated RAM/CPU usage queued or executing

iain-macdonald:
I'm planning to take a look at how this fares in dev for a little while, compared to the current autoscaler metric of queued tasks, and then explore autoscaling off this instead.

**Related issues**: N/A

### DIFF
--- a/proto/scheduler.proto
+++ b/proto/scheduler.proto
@@ -131,6 +131,15 @@ message SchedulingMetadata {
   string executor_group_id = 5;
   // Group ID of the user that issued the Execute request.
   string task_group_id = 6;
+
+  // A signal to the executor that the size of this queued task should be
+  // tracked as part of the queued-or-assigned size metrics. This is necessary
+  // because tasks may be scheduled on multiple executors, but should only
+  // contributed to this system-wide metric once, so the scheduler must inform
+  // exactly one of the executors to perform the queued task size tracking.
+  // This is for metrics purposes only and shouldn't affect the behavior of the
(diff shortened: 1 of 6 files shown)


proto/scheduler.proto
```diff
@@ -126,6 +126,15 @@ message SchedulingMetadata {
...
   string task_group_id = 6;
+
+  // A signal to the executor that the size of this queued task should be
+  // tracked as part of the queued-or-assigned size metrics. This is necessary
+  // because tasks may be scheduled on multiple executors, but should only
+  // contributed to this system-wide metric once, so the scheduler must inform
+  // exactly one of the executors to perform the queued task size tracking.
+  // This is for metrics purposes only and shouldn't affect the behavior of the
+  // scheduler or the executor.
+  bool track_queued_task_size = 9;
```
iain-macdonald, line 137: Open to other approaches for de-duping queued resource usage here!

tylerwilliams: could we flush these metrics only when the task is completed?

tylerwilliams:
reverse-ping (since it's showing up in my go/r dash)

proto/scheduler.proto
iain-macdonald: I'd like to be able to track queued requests for resources as well, so waiting for the task to complete (which could take a while) wouldn't work, unless I'm misunderstanding.

tylerwilliams (approved):

PR #5314, opened 2023-11-15
[SS] Use finer-grained locking in COWStore

maggie-lou:
In the COWStore, we need locking to prevent concurrent access of maps, and also to synchronize simultaneous actions on a chunk - in particular when we're copying a chunk. We copy chunks when they are written to. When copying, we first initialize an empty chunk, and then copy over the original data to it. We need to make sure concurrent goroutines don't try to read from the empty new chunk, or write to the outdated old chunk for example.

Our hypothesis (validated below) is that locking in this second case hurts performance, because it's wrapped around a large and slow chunk of code that includes copying the data over. This PR adds a finer grained lock in this case, so we can just lock access to that chunk and not the entire store.

Testing PR: #5334 

For an incremental build on gazelle repo: 

On master:
For store memory, total lock time is 45684 (ms)

With this PR:
For store memory, total lock time is 15 (ms)

**Related issues**: More convo in #4920

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/BUILD
+++ b/enterprise/server/remote_execution/copy_on_write/BUILD
@@ -11,6 +11,7 @@ go_library(
         "//server/environment",
         "//server/interfaces",
         "//server/remote_cache/digest",
+        "//server/util/lockmap",
         "//server/util/log",
         "//server/util/status",
         "@org_golang_x_exp//maps",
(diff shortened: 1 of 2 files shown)


maggie-lou:
* master
  * **PR #5314** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5314?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5314?utm_source=stack-comment).

bduffany:
Do you mind cherry-picking https://github.com/buildbuddy-io/buildbuddy/pull/5094 into this PR and making sure it passes? The code LGTM but would prefer to have more rigorous tests in place before merging.

maggie-lou:
> Do you mind cherry-picking #5094 into this PR and making sure it passes? The code LGTM but would prefer to have more rigorous tests in place before merging.

Passed 👍

bduffany (approved):
LGTM :)

PR #5316, opened 2023-11-16
Reduce allocations when extracting invocation ID

tylerwilliams:
just by doing a normal build I see this has a significant impact in reducing allocations because we're not re-extracting that metadata proto over and over again.

### DIFF
--- a/server/rpc/interceptors/interceptors.go
+++ b/server/rpc/interceptors/interceptors.go
@@ -139,7 +139,7 @@ func copyHeadersToContext(ctx context.Context) context.Context {
 			ctx = context.WithValue(ctx, contextKey, hdrs[0])
 		}
 	}
-	ctx = bazel_request.ExtractValuesIntoContext(ctx)
+	ctx = bazel_request.ParseRequestMetadataOnce(ctx)
 	return ctx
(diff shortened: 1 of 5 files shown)


server/rpc/interceptors/interceptors.go
```diff
@@ -279,6 +279,8 @@ func subdomainUnaryServerInterceptor() grpc.UnaryServerInterceptor {
 }
 
 func addInvocationIdToLog(ctx context.Context) context.Context {
+	ctx = bazel_request.ExtractValuesIntoContext(ctx)
```
luluz66, line 282: I wonder if it would make sense to do it in interceptors, since we are also calling GetInvocationID in propagateInvocationIDToSpanUnaryServerInterceptor

tylerwilliams: Good call; done.

tylerwilliams (commented):
I also removed some other instances where we were parsing the request metadata so there should be only one per request now.

jdhollen (approved):

bduffany (approved):

luluz66 (approved):

PR #5318, opened 2023-11-16
move scheduling/scheduler_server/config to //server

sluongng:
This should be the last enterprise dependencies in static.go.

Expected to merge this after https://github.com/buildbuddy-io/buildbuddy/pull/5264 as there will be merge conflict


sluongng:
The alternative is to include this as part of `server/remote_execution/config/config.go`.

siggisim (approved):

PR #5319, opened 2023-11-16
buildbuddy.yaml: add workflow to validate FOSS build

sluongng:
We should be able to build FOSS successfully without depending on
//enterprise.

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -10,6 +10,18 @@ actions:
           - "*"
     bazel_commands:
       - test //... --config=linux-workflows --config=race --test_tag_filters=-performance,-webdriver,-docker,-bare
+  - name: Test FOSS build
+    container_image: ubuntu-20.04
+    triggers:
+      push:
+        branches:
+          - "master"
+      pull_request:
+        branches:
+          - "*"
+    bazel_commands:
+      - run --config=linux-workflows --config=race //tools:rm-enterprise
+      - build --config=linux-workflows --config=race //server
   - name: Test with static
     container_image: ubuntu-20.04
     triggers:
--- a/tools/BUILD
+++ b/tools/BUILD
@@ -1,1 +1,13 @@
 package(default_visibility = ["//visibility:public"])
+
+genrule(
+    name = "rm-enterprise",
+    outs = ["rm-enterprise.sh"],
+    cmd = """
+    cat <<'EOF' > $@
+#!/bin/bash
+rm -rf $$BUILD_WORKING_DIRECTORY/enterprise
+EOF
+""",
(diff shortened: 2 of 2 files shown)


bduffany:
Will defer to @tylerwilliams and @siggisim on this, since I'm not sure whether it's worth the compute resources right now.

This will be cheaper if we allow configuring the instance name for workflow actions though - then we can share the snapshot cache between this workflow and other workflows.

tools/cas/BUILD
```diff
@@ -6,6 +6,7 @@ go_library(
     importpath = "github.com/buildbuddy-io/buildbuddy/tools/cas",
     visibility = ["//visibility:private"],
     deps = [
+        "//cli/storage",
```
bduffany, line 9: These changes seem to have leaked in from the other PR

sluongng: Sorry it was a bad git rebase locally that updated the wrong branch.

PTAL.

sluongng:
> This will be cheaper if we allow configuring the instance name for workflow actions though - then we can share the snapshot cache between this workflow and other workflows.

I think most of the compute here is happening on RBE and therefore, is deduplicated with other workflows on our Scheduler action merging. There is a bit of overhead in spinning up a workflow microvm and removing the dir from git checkout. But I would say that it's relatively minimal. 

The files are still existing on disk inside Git's packfiles, so reusing the workspace with other workflows should be possible as well. However, I am not sure about the analysis cache when changing the target patterns between `//server` and `//...` back and forth would affect subsequent builds' performance anyhow. (more testing needed if we ever wanna go down that path)

buildbuddy.yaml
```diff
@@ -10,6 +10,18 @@ actions:
...
+    triggers:
+      push:
+        branches:
+          - "master"
+      pull_request:
+        branches:
+          - "*"
+    bazel_commands:
+      - run --config=linux-workflows //tools:rm-enterprise
+      - build --config=linux-workflows //server
```
bduffany, line 24: to make sure the AC keys here are shared with the Test workflow, we need to add `--config=race` to both of these

sluongng: Good point. 👍 
Added.

siggisim:
Shouldn't we be able to achieve this with visibility rules?

sluongng:
> Shouldn't we be able to achieve this with visibility rules?

Historically, it has not been sufficient as folks have been setting the visibility on the target level to `public` to bypass the gazelle's package level visibility setting. 🤔

siggisim:
Let's be harsher on visibility changes in code review

sluongng:
Alright, let's defer this until later.

PR #5321, opened 2023-11-16
Reduce findmissing fanout

tylerwilliams:
We discussed over lunch the idea of reducing FindMissing fanout. This change limits the number of nodes contacted for each digest in a FindMissing call to 1 at the risk of returning false-positive responses in FindMissing (basically: saying something is missing when really we have it).

We don't believe this should cause issues because the cases where digests are not on the correct replica are rare and if they are encountered we'll possibly trigger an extra write

### DIFF
--- a/enterprise/server/backends/distributed/distributed.go
+++ b/enterprise/server/backends/distributed/distributed.go
@@ -731,8 +731,8 @@ func (c *Cache) FindMissing(ctx context.Context, resources []*rspb.ResourceName)
 	// For every digest we found, if we did not find it
 	// on the first peer in our list, we want to backfill it.
 	backfills := make([]*backfillOrder, 0)
-	for h, present := range foundMap {
-		if !present {
+	for h, foundOnPeer := range foundMap {
+		if !foundOnPeer {
 			continue
 		}
 		r := hashResources[h][0]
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/distributed/distributed.go
```diff
@@ -706,9 +706,8 @@ func (c *Cache) FindMissing(ctx context.Context, resources []*rspb.ResourceName)
 				}
 				for _, r := range resources {
 					hash := r.GetDigest().GetHash()
-					if _, ok := peerMissingHashes[hash]; !ok {
-						foundMap[hash] = struct{}{}
-					}
+					_, ok := peerMissingHashes[hash]
```
vadimberezniker, line 709: change `ok` to `missing` to make this clearer

enterprise/server/backends/distributed/distributed.go
```diff
@@ -721,16 +720,20 @@ func (c *Cache) FindMissing(ctx context.Context, resources []*rspb.ResourceName)
...
 		if len(foundMap) == len(hashResources) {
-			// If we've found everything, we can exit now.
+			// If we've tried everything, we can exit now.
 			break
 		}
 	}
 
 	// For every digest we found, if we did not find it
 	// on the first peer in our list, we want to backfill it.
 	backfills := make([]*backfillOrder, 0)
```
vadimberezniker, line 732: does backfilling have a point anymore if we're only looking at the first successful response? We can try more than one peer if a peer errors out but that doesn't mean a backfill is necessary.

vadimberezniker (approved):

enterprise/server/backends/distributed/distributed.go
tylerwilliams: Done.

enterprise/server/backends/distributed/distributed.go
tylerwilliams: I was trying to minimize the size of this change but I agree, will remove it.

PR #5324, opened 2023-11-17
tools/cas: fallback to use api key from git config

sluongng:
When use tools/cas, it's easy to leak API key to BES.

For example:
`bazel run tools/cas -- -api_key=foobar`
would send the `-api_key=foobar` over the wire to the server side for
display.

A good workaround to this is to make sure BES is not used during `bazel
run`:

```

run --bes_results_url=
run --bes_backend=
```

However it would be good to not having to set the API key manually at
all. Since our `bb` CLI is already setting the API key inside
`.git/config` in a dedicated `buildbuddy` section, let's re-use that.

### DIFF
--- a/tools/cas/cas.go
+++ b/tools/cas/cas.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/buildbuddy-io/buildbuddy/cli/storage"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
 	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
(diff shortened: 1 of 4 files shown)


tools/cas/cas.go
```diff
@@ -100,8 +101,16 @@ func main() {
...
+
+	bbApiKey := *apiKey
+	if bbApiKey == "" {
+		// If the API key is not set, try to get it from BuildBuddy section in Git config.
+		if key, err := storage.ReadRepoConfig("api-key"); err == nil {
+			bbApiKey = key
+		}
+	}
+	if bbApiKey != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-api-key", bbApiKey)
```
bduffany, line 113: I could see this approach making sense for `bb download` since it's part of the CLI, but we probably shouldn't have deps from `//tools` => `//cli`.

We probably should just redact any args from the BES that contain `api_key` in `redact.go`

sluongng: It's `//tools` depending on `//cli` here, not the reverse, which I think is ok? We do want a set of common libraries for our CLIs (tools, enterprise/tools, cli) in general right?

I could look into `redact.go` separately.

bduffany: I think we shouldn't have non-CLI packages depending on CLI-specific functionality. The "ReadRepoConfig" is a CLI-specific config right now; it's sort of "owned" by the CLI. So it's a bit weird to have non-CLI tools reading that data. This `//tools` directory is for our internal ad-hoc tooling and the shared libraries that we use for tools are all under //server today, not //cli.

This tool is sort of redundant with `bb download` and we want to remove it eventually. I think we want to deprecate this tool and switch to `bb download` for everything.

sluongng:
Let’s just skip this in favor of `bb download`

tools/cas/cas.go
sluongng: SGTM

PR #5328, opened 2023-11-17
redact: process residual in StructuredCommandLine

sluongng:


sluongng:
Im a bit iffy on this.

On one hand I think we should just wipe residual entirely instead of doing this.
On the other hand, i think we might want this for workflows with "bazel run" in it's script.

sluongng:
https://github.com/bazelbuild/bazel/discussions/20126 is also related

server/util/redact/redact_test.go
```diff
@@ -378,3 +378,51 @@ func TestRedactAPIKey(t *testing.T) {
...
+		},
+		{
+			name:     "WithAPIKey_NoSplit",
+			given:    []string{"-api_key=foobar", "-baz"},
+			expected: []string{"-baz"},
+		},
+		{
+			name:     "WithAPIKey_Split",
+			given:    []string{"-api_key", "foobar", "-baz"},
+			expected: []string{"-baz"},
```
bduffany, line 401: I'm not sure if the result here could be confusing, should the expected result be

```
"-api_key", "<REDACTED>", "-baz"
```

and for the one above,

```
"-api_key=<REDACTED>", "-baz" 
```

bduffany: also might be reasonable to redact anything matching the regex `\b(key|token|secret)\b`

that way we redact `api-key`, `api_key`, `github_token`, `client_secret` etc.

sluongng: Hmm it might need to be `(\b|_)(key|token|secret)(\b|_)` instead because `\b` does not include underscore `_`? 🤔 

Let me look into it.

bduffany: Ah yeah. This should do the trick:

```
(^|[^a-z])(key|secret|token|password)($|[^a-z])
```

sluongng: I pushed a change to make use of the new regex, let me know what you think.

bduffany (commented):
awesome, thanks for doing this :)

bduffany (approved):

PR #5332, opened 2023-11-17
Add `--registry_mirror=` which defaults to `mirror.gcr.io`

siggisim:
We already use this registry mirror when pulling images through GKE. This adds the mirror to the docker daemon that gets started inside of firecracker VMs.

Longer term, we could potentially host our own mirror here instead.

Need to do some performance testing first to make sure this actually speeds things up.


tylerwilliams:
I think we'll need to do something like [this](https://github.com/buildbuddy-io/buildbuddy/pull/5221) since we're changing the args passed to the init process.

tylerwilliams (approved):

bduffany (approved):

PR #5335, opened 2023-11-18
Commas everywhere

siggisim:
Adds comma rendering to a bunch of counts all over our UI.

### DIFF
--- a/app/format/format.tsx
+++ b/app/format/format.tsx
@@ -327,7 +327,10 @@ export function formatRole(role: string): string | null {
   return null;
 }
 
-export function formatWithCommas(num: number | Long) {
+export function formatWithCommas(num: number | Long | Number | undefined) {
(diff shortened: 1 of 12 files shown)


app/format/format.tsx
```diff
@@ -327,7 +327,10 @@ export function formatRole(role: string): string | null {
   return null;
 }
 
-export function formatWithCommas(num: number | Long) {
+export function formatWithCommas(num: number | Long | Number | undefined) {
+  if (!num) {
```
bduffany, line 331:
```suggestion
  if (num === undefined || num === null) {
```

(otherwise returns `""` when `num === 0`)

bduffany (approved):

app/format/format.tsx
siggisim: Good catch!

PR #5338, opened 2023-11-20
Make trace events with target label clickable

siggisim:
With this change, clicking on an event with a target label takes you to the target page.

Some changes on the way to the target page to make this more useful.

### DIFF
--- a/app/trace/trace_viewer.tsx
+++ b/app/trace/trace_viewer.tsx
@@ -11,6 +11,7 @@ import Panel from "./trace_viewer_panel";
 import { TraceEvent } from "./trace_events";
 import { buildTraceViewerModel, panelScrollHeight } from "./trace_viewer_model";
 import { Profile } from "./trace_events";
+import router from "../router/router";
 
 export interface TraceViewProps {
   profile: Profile;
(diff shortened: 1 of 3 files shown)


app/trace/trace_viewer.tsx
```diff
@@ -74,6 +75,7 @@ export default class TraceViewer extends React.Component<TraceViewProps, {}> {
 
     window.addEventListener("resize", this.onWindowResize);
     window.addEventListener("mousemove", this.onWindowMouseMove);
+    window.addEventListener("click", this.onWindowMouseClick);
```
bduffany, line 78: With this approach we'd need to unregister the window-level event listener in `componentWillUnmount`, since React won't do it automatically.

But I think since we only care about clicks inside the canvas, we can register it on the `canvas` element, so the listener will get cleaned up automatically:


```
<canvas onClick={(e) => onCanvasClick(e, panel)} ... />
```

the impl of `onCanvasClick` can also be simplified to

```
const target = panel.getHoveredEvent()?.args?.target;
if (target) {
  router.navigateTo(`?target=${target}#targets`);
}
```

bduffany (commented):
hm I can't seem to get this PR to work locally - when I click "Testing //:pass" it doesn't open the target page for "//:pass"

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/449b9d93-9b81-4486-95a8-e3d95066d8a0

siggisim:
> hm I can't seem to get this PR to work locally - when I click "Testing //:pass" it doesn't open the target page for "//:pass"
> 
>  rec-2023-11-20_13.39.04.mp4

You have to set the following bazel flag, which we recommend in our suggested flags.
```
 --experimental_profile_include_target_label
```

I'll add that to our .bazelrc in this PR as well.

app/trace/trace_viewer.tsx
siggisim: Done!

bduffany (approved):
this is awesome!

siggisim:
> this is awesome!

Just wait until you see what else I've got cookin'!

PR #5344, opened 2023-11-21
Add a platform property that lets the user delay execution of tasks scheduled on non-preferred nodes.

iain-macdonald:
This is a customer feature request. I took the opportunity to refactor the `enqueueTaskReservations` function in `scheduler_server.go` and would appreciate a careful review!

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/platform/platform.go
+++ b/enterprise/server/remote_execution/platform/platform.go
@@ -65,27 +65,28 @@ const (
 	// empty or unset.
 	unsetContainerImageVal = "none"
 
-	RecycleRunnerPropertyName            = "recycle-runner"
-	AffinityRoutingPropertyName          = "affinity-routing"
-	preserveWorkspacePropertyName        = "preserve-workspace"
-	nonrootWorkspacePropertyName         = "nonroot-workspace"
-	cleanWorkspaceInputsPropertyName     = "clean-workspace-inputs"
-	persistentWorkerPropertyName         = "persistent-workers"
-	persistentWorkerKeyPropertyName      = "persistentWorkerKey"
-	persistentWorkerProtocolPropertyName = "persistentWorkerProtocol"
-	WorkflowIDPropertyName               = "workflow-id"
-	workloadIsolationPropertyName        = "workload-isolation-type"
-	initDockerdPropertyName              = "init-dockerd"
-	enableDockerdTCPPropertyName         = "enable-dockerd-tcp"
-	enableVFSPropertyName                = "enable-vfs"
-	HostedBazelAffinityKeyPropertyName   = "hosted-bazel-affinity-key"
-	useSelfHostedExecutorsPropertyName   = "use-self-hosted-executors"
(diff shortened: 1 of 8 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1662,95 +1678,147 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
...
-			cancel()
-			if err != nil {
-				log.CtxWarningf(ctx, "EnqueueTaskReservation via scheduler target %q failed: %s", node.schedulerHostPort, err)
-				time.Sleep(schedulerEnqueueTaskReservationFailureSleep)
-				continue
+			enqueueRequest.Delay = nil
+		}
+		enqueued, err := s.enqueue(ctx, rankedNode.node, enqueueRequest, opts)
+		if err != nil {
+			return err
```
vadimberezniker, line 1737: I don't think you meant to return an error here. This would mean we stop attempting enqueing after the first error. 

Now that I read the code a little more I see that enqueue only returns an error if the context is done. Still this is very error prone because if someone ever modifies enqueue to return an error in other conditions, it will cause us to stop enqueing early. If you have any termination conditions you want to check, I would move them to this loop.

vadimberezniker (commented):
Overall, this looks like a very nice improvement of the existing code. I would have preferred if you separated the refactoring & delay change into separate changes, but let's not worry about that since what's done is done.

I will do another pass in a bit to make sure I haven't missed anything, but overall it looks really good.

vadimberezniker:
It would be good if @bduffany could also give it a look.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
iain-macdonald: Good catch, this is indeed bug-prone. Removed the error return from the called function and added a test that should catch the case you mentioned.

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -65,27 +65,28 @@ const (
...
-	useSelfHostedExecutorsPropertyName   = "use-self-hosted-executors"
-	disableMeasuredTaskSizePropertyName  = "debug-disable-measured-task-size"
-	disablePredictedTaskSizePropertyName = "debug-disable-predicted-task-size"
-	extraArgsPropertyName                = "extra-args"
-	EnvOverridesPropertyName             = "env-overrides"
-	EnvOverridesBase64PropertyName       = "env-overrides-base64"
-	IncludeSecretsPropertyName           = "include-secrets"
+	RecycleRunnerPropertyName             = "recycle-runner"
+	AffinityRoutingPropertyName           = "affinity-routing"
+	ColdRunnerSchedulingDelayPropertyName = "cold-runner-scheduling-delay"
```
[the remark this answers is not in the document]
iain-macdonald: Good suggestion, changed.

bduffany (approved):

PR #5346, opened 2023-11-21
Add a target scoped cache requests card to the target page

siggisim:
This makes the target page much more useful, as it links the target to the action(s) that it spawns, as well as the CAS artifacts that it uploads.

It also means that in conjunction with the clickable timing profile, we now have a way to link the timing profile, the target page, and remote cache requests / executions.

This change also makes the target card load even if the target wasn't mentioned in the BES, since it might be a dependency that is not announced that we still have cache request / timing profile entries for.

Screenshot 2023-11-20 at 4 38 25 PM


siggisim:
> nice! was thinking the other day it would be awesome if we could show executions for the target here eventually too

the nice thing is you can click on the action cache entries and see the remotely executed actions

PR #5349, opened 2023-11-21
Add timeouts and retries for CAS/AC operations.

vadimberezniker:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/2916

**Related issues**: N/A

### DIFF
--- a/server/remote_cache/cachetools/cachetools.go
+++ b/server/remote_cache/cachetools/cachetools.go
@@ -41,7 +41,6 @@ var (
 	enableUploadCompresssion = flag.Bool("cache.client.enable_upload_compression", true, "If true, enable compression of uploads to remote caches")
 	casRPCTimeout            = flag.Duration("cache.client.cas_rpc_timeout", 1*time.Minute, "Maximum time a single batch RPC or a single ByteStream chunk read can take.")
 	acRPCTimeout             = flag.Duration("cache.client.ac_rpc_timeout", 15*time.Second, "Maximum time a single Action Cache RPC can take.")
-	byteStreamWriteTimeout   = flag.Duration("cache.client.byte_stream_write_timeout", 15*time.Minute, "Maximum time a single ByteStream upload can take.")
 )
 
 func retryOptions(name string) *retry.Options {
@@ -95,6 +94,7 @@ func getBlob(ctx context.Context, bsClient bspb.ByteStreamClient, r *digest.Reso
 	}
 
 	streamMsgs := rpcutil.RecvChan[*bspb.ReadResponse](ctx, stream)
+	// XXX need to reset the timer
 	readTimer := time.NewTimer(*casRPCTimeout)
 	defer readTimer.Stop()
 
@@ -274,6 +274,10 @@ func uploadFromReader(ctx context.Context, bsClient bspb.ByteStreamClient, r *di
 
 	buf := make([]byte, uploadBufSizeBytes)
 	bytesUploaded := int64(0)
+	sender := rpcutil.SendChan[*bspb.WriteRequest](ctx, stream)
+	// XXX need to reset the timer
+	writeTimer := time.NewTimer(*casRPCTimeout)
+	defer writeTimer.Stop()
 	for {
 		n, err := rc.Read(buf)
 		if err != nil && err != io.EOF {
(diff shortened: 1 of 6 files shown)


server/remote_cache/cachetools/cachetools.go
```diff
@@ -272,6 +339,25 @@ func UploadFromReader(ctx context.Context, bsClient bspb.ByteStreamClient, r *di
...
 
+func UploadFromReader(ctx context.Context, bsClient bspb.ByteStreamClient, r *digest.ResourceName, in io.Reader) (*repb.Digest, error) {
+	// We can only retry if we can rewind the reader back to the beginning.
+	seeker, retryable := in.(io.Seeker)
+	if retryable {
+		return retry.Do(ctx, retryOptions("ByteStream.Write"), func(ctx context.Context) (*repb.Digest, error) {
+			if _, err := seeker.Seek(0, io.SeekStart); err != nil {
+				return nil, retry.NonRetryableError(err)
+			}
+			// We set a timeout for the entire write since we can't monitor
```
bduffany, line 350: hm can we do the same thing in UploadFromReader that we do in GetBlob, with the goroutine + timer? (I get that `stream.Send()` may buffer but isn't there some max number of outstanding unack'd messages?)

server/util/retry/retry.go
```diff
@@ -180,3 +186,53 @@ func (r *Retry) MaxAttempts() int {
...
+		if err != nil {
+			lastError = err
+			continue
+		}
+		if lastError != nil && !opts.DontLogFailedAttempts {
+			name := opts.Name
+			if name != "" {
+				name += " "
+			}
+			log.CtxWarningf(ctx, "%sattempt failed, but succeeded on retry: %s", name, err)
```
tylerwilliams, line 208: i'm a little fuzzy on this bit -- is it gonna log something like `GetActionResultattempt failed`?

I think it would be useful to have a String() method on the Retry that is safe to call and shows `attempt $x of $y`

server/remote_cache/cachetools/cachetools.go
```diff
@@ -86,8 +93,39 @@ func GetBlob(ctx context.Context, bsClient bspb.ByteStreamClient, r *digest.Reso
...
+		var rsp *bspb.ReadResponse
+		var err error
+		select {
+		case msg := <-streamMsgs:
+			rsp = msg.rsp
+			err = msg.err
+		case <-readTimer.C:
+			return status.DeadlineExceededError("Timed out waiting for Read response")
+		case <-ctx.Done():
+			return ctx.Err()
```
tylerwilliams, line 111: sorta dumb question -- could we use context.WithTimeout(ctx, *flagVal) if the context timeout is unset or longer, and then remove the readTimer?

vadimberezniker: Had to rework some things, but I was able to work the context timeout into this.

server/util/retry/retry.go
```diff
@@ -180,3 +186,53 @@ func (r *Retry) MaxAttempts() int {
...
+		}
+		return rsp, nil
+	}
+	return *new(T), lastError
+}
+
+// DoVoid is a convenience wrapper around Do for functions that do not return a
+// value.
+func DoVoid(ctx context.Context, opts *Options, fn func(ctx context.Context) error) error {
```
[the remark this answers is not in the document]
vadimberezniker: I heard you like a little bit of Java in your Go :)

server/util/retry/retry.go
vadimberezniker: This will log a single warning, but only if the function eventually succeeds. 

If all retries are exhausted, we will return an error so there's no need to log anything. It's up to the caller to decide if they want to log in that case. Since we don't know whether the operation will eventually succeed, I postpone logging the failed attempt until we return a non-error response.

server/remote_cache/cachetools/cachetools.go
vadimberezniker: Good idea, done. (Individual messages are not acked, but there's only so much data a sender is allowed to send before the peer has to indicate that it is willing to receive X more bytes.)

server/util/rpcutil/rpcutil.go
```diff
@@ -0,0 +1,123 @@
...
+
+// SendWithTimeoutCause attempts to send a message on the underlying stream,
+// waiting a maximum of timeout. If timeout is reached, the given cause is
+// returned as the error.
+//
+// Note that gRPC sends are asynchronous in the sense that the protocol does not
+// acknowledge individual messages. A timeout wil only occur if the sender
+// exhausts the flow-control window and the receiver does not increase it.
+func (s *Sender[T]) SendWithTimeoutCause(msg T, timeout time.Duration, cause error) error {
+	s.sendChan <- msg
```
bduffany, line 85: I think this needs a `select {}` with `<-s.ctx.Done()` here, otherwise it could get stuck?

bduffany (approved):

server/util/rpcutil/rpcutil.go
vadimberezniker: sendChan is a buffered channel and SendWithTimeoutCause does not return until either we have received an err and the channel is empty or the context is done and SendWithTimeoutCause won't be called again.

tylerwilliams (approved):

PR #5351, opened 2023-11-21
website: enable legacy OpenSSL provider

sluongng:
Make sure that `:start` and `:serve` starts the node server with the
correct parameters.

References:
- https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported

### DIFF
--- a/rules/yarn/index.bzl
+++ b/rules/yarn/index.bzl
@@ -19,7 +19,7 @@ def yarn(name, srcs, package, command = "build", deps = [], yarn = "@yarn//:yarn
     """ % (package, yarn, node, command)
 
     if executable:
-        cmd = "echo \"cd $$(dirname $(location %s)) && yarn install && yarn %s\" > $@" % (package, command)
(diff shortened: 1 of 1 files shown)


sluongng:
This is needed due to my local nodejs version is `v21.2.0`, way ahead of what we are using (`16.19.0`).

It should only affect local development of `//website` though. So I think it's a relatively harmless change.

siggisim:
Seems like the correct solution here would be to add the pinned node version to the path (similar to how we do with yarn) so it uses the nodes version specified in the WORKSPACE file rather than the one installed on your system.

bduffany:
> Seems like the correct solution here would be to add the pinned node version to the path

It looks like we are adding node to the PATH but maybe `yarn` is somehow overriding it

sluongng:
closing this in favor of https://github.com/buildbuddy-io/buildbuddy/pull/5369

PR #5356, opened 2023-11-21
Show miss reason, action mnemonic, and runner type breakdowns

siggisim:
It's basic, but it gets the job done:

Screenshot 2023-11-21 at 9 27 11 AM

### DIFF
--- a/app/invocation/invocation_cache_card.tsx
+++ b/app/invocation/invocation_cache_card.tsx
@@ -4,6 +4,8 @@ import format from "../format/format";
 import { PieChart as PieChartIcon, AlertCircle as AlertCircleIcon } from "lucide-react";
 import { ResponsiveContainer, PieChart, Pie, Cell } from "recharts";
 import capabilities from "../capabilities/capabilities";
+import { getChartColor } from "../util/color";
+import { blaze } from "../../proto/action_cache_ts_proto";
 
 interface Props {
   model: InvocationModel;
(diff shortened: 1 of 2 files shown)


app/invocation/invocation_cache_card.tsx
```diff
@@ -53,13 +55,13 @@ export default class CacheCardComponent extends React.Component<Props> {
                         <div>
                           <div className="cache-chart-label">
                             <span className="color-swatch cache-hit-color-swatch"></span>
-                            <span className="cache-stat">{format.formatWithCommas(cacheStat.actionCacheHits)} </span>
-                            hits
+                            <span className="cache-stat">{format.formatWithCommas(cacheStat.actionCacheHits)}</span>
+                            &nbsp;hits
```
bduffany, line 59: nit: consider `white-space: nowrap` instead of using non-breaking spaces here

app/invocation/invocation_cache_card.tsx
```diff
@@ -213,3 +272,69 @@ export default class CacheCardComponent extends React.Component<Props> {
     );
   }
 }
+
+let cap = 5;
```
bduffany, line 276: nit: since this is a top-level declaration, rename this to be more clear? or move it to renderBreakdown

bduffany (approved):

tylerwilliams (approved):

app/invocation/invocation_cache_card.tsx
siggisim: I had to add these because adding `white-space: nowrap` killed the spaces (and I borked the spaces before in a previous PR from formatting with prettier), so I felt like adding an explicit &nbsp; would make this less brittle (I generally avoid them, but this felt like a place where they were useful). I tried {" "} and a bunch of incantations, but nothing else worked reliably in this situation with a span next to text on a new line with nowrap enabled.

app/invocation/invocation_cache_card.tsx
siggisim: Done

PR #5357, opened 2023-11-21
Don't show `Access denied` page if the user isn't a member of any orgs

siggisim:
With this change, it will show this page instead:

Screenshot 2023-11-21 at 11 33 35 AM

### DIFF
--- a/enterprise/app/org/create_org.tsx
+++ b/enterprise/app/org/create_org.tsx
@@ -46,7 +46,13 @@ export default class CreateOrgComponent extends OrgForm<grp.CreateGroupRequest> 
         <div className="container">
           <div className="organization-page-title">Create organization</div>
           {this.props.user && !Boolean(this.props.user.groups?.length) && (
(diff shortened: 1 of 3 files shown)


enterprise/app/org/create_org.tsx
```diff
@@ -46,7 +46,13 @@ export default class CreateOrgComponent extends OrgForm<grp.CreateGroupRequest>
         <div className="container">
           <div className="organization-page-title">Create organization</div>
           {this.props.user && !Boolean(this.props.user.groups?.length) && (
-            <Banner type="info">You are logged in, but not part of any organization. Create one to continue.</Banner>
+            <Banner type="info">
+              You are logged in, but not part of any organization. Create one to continue or{" "}
+              <a className="organization-page-link" href="/logout/">
+                click here to Logout
```
bduffany, line 52: nit:

```suggestion
                click here to log out
```

bduffany (approved):

PR #5362, opened 2023-11-22
event parsing: only use 'canonical' StructuredCommandLine event

sluongng:
Bazel sends multiple command line events to the BES server:
- `original`
- `canonical`

Where the `canonical` event is the effective one while the `original`
event is algamation of command line flags and `.bazelrc` configs.
Additionally, users may specify `--experimental_tool_command_line` to
send the command line from layers of Bazel wrappers.

Let's filter for `canonical` command line event to parse since it's the
most accurate one. The other 2 events are still available via raw bep
JSON and could be inspected accordingly from our UI (if they are not
empty).

References:
- https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/runtime/CommandLineEvent.java;drc=dab6383ecef2d3996a67b2b1e4760369abc08de7

### DIFF
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -214,6 +214,9 @@ export default class InvocationModel {
       this.toolLogMap.set(log.name, new TextDecoder().decode(log.contents || new Uint8Array()));
     }
     for (let commandLine of this.structuredCommandLine) {
+      if (commandLine.commandLineLabel != "canonical") {
+        continue;
+      }
       for (let section of commandLine.sections || []) {
         for (let option of section.optionList?.option || []) {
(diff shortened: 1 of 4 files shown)


siggisim:
Just curious, what's the problem this is solving?

sluongng:
@siggisim: We have a user reported in Slack https://buildbuddy.slack.com/archives/CUY16GNK1/p1700659015942119 that their on-prem deployment of BuildBuddy FOSS would display the wrong `download_outputs_option` on the UI header.
image

I found that we are parsing in both original and canonical events. So by parsing only the final canonical event, we get a more accurate result.

bduffany (approved):

PR #5369, opened 2023-11-24
website: use Bazel's node toolchain with 'start' and 'serve'

sluongng:
An alternative to #5351.

Move the command templates out to constants to render multi-lines
correctly.

For the executable template, make sure that we set PATH before
running yarn so the correct NodeJS is picked up.


luluz66:
Would love to get this merged as I encountered problems too when trying to run website locally.

bduffany (approved):

PR #5370, opened 2023-11-27
Limit mmap memory usage

bduffany:
- Add an LRU to each COWStore that limits the number of chunks that can be mapped in memory at once. Whenever we map a chunk, the chunk is bumped to the front of the LRU. If mapping a new chunk and we're at capacity, the least recently used chunk is evicted from the LRU and unmapped (in the OnEvict callback)
- Update the task sizer to reserve extra memory for mmapped chunks. The extra memory allocated is `min(10% of file size, 1GB)`.
  - Only applies to both rootfs and memory, so at most 2GB of extra memory is reserved for each action.
  - For now we don't worry about workspacefs and just assign a default of 250MB without tracking this in the task size.

**Related issues**: N/A


enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -761,6 +763,17 @@ func ConvertFileToCOW(ctx context.Context, env environment.Env, filePath string,
 	return NewCOWStore(ctx, env, chunks, chunkSizeBytes, totalSizeBytes, dataDir, remoteInstanceName, remoteEnabled)
 }
 
+// getMmapLRU returns the shared LRU instance to be used for the mmap,
+// if applicable.
+func getMmapLRU(dataDir string) (*MmapLRU, error) {
+	// We constrain UFFD memory usage by only allowing one chunk to be mapped
+	// at once, so don't use the shared LRU for UFFD.
+	if filepath.Base(dataDir) != snaputil.MemoryFileName {
```
bduffany, line 771: This code is probably the grossest thing about this PR (depending in the file name to determine whether we use the shared LRU) but there are two reasons why I think it's OK:

* Avoids extra plumbing in snaploader / NewCOWStore / NewMmap* and keeps the PR smol
* In today's world, snaploader would just have to use the chunk file name anyway because that's how we identify the memory snapshot vs. the disk snapshot.

bduffany (commented):
Reworked based on our offline discussion about using a single LRU only for disk artifacts. (All the above comments are either obsolete or addressed in the other PR so I clicked resolve on all of them). Still need to test perf in dev. FYI @maggie-lou in case you want to try merging this with your branch and testing perf locally as well

bduffany:
The benchmark tests are looking OK (though they aren't testing parallelism etc.) so I think this PR is probably good to go for now and then we can do more perf testing / bottleneck chasing independently of this PR.

maggie-lou (approved):

server/resources/resources.go
```diff
@@ -17,6 +17,7 @@ import (
 
 var (
 	memoryBytes  = flag.Int64("executor.memory_bytes", 0, "Optional maximum memory to allocate to execution tasks (approximate). Cannot set both this option and the SYS_MEMORY_BYTES env var.")
+	mmapFraction = flag.Float64("executor.mmap_memory_fraction", 0.1, "Fraction of memory to be allocated towards mmapped files for Firecracker copy-on-write functionality. Has no effect if snapshot sharing is disabled.")
```
tylerwilliams, line 20: This would be clearer as a direct size number of bytes

tylerwilliams (approved):
LGTM

maggie-lou:
@bduffany Here are the manual performance test results on this PR branch (looks a little slower but not horrible, and should be less on the dev executors)

(remote_snapshot_sharing=true) Bazel build on a clean runner took 26m32.360026624s.
(remote_snapshot_sharing=true) Bazel build on a recycled runner (100% locally cached) took 14.527744482s.
(remote_snapshot_sharing=true) Bazel build (30% locally cached) took 26.077916022s. 70% artifacts fetched remotely.
(remote_snapshot_sharing=true) Bazel build (0% locally cached) took 50.112072259s. 100% artifacts fetched remotely.

On master the last time I ran these:
(remote_snapshot_sharing=true) Bazel build on a clean runner took 25m24.355207734s.
(remote_snapshot_sharing=true) Bazel build on a recycled runner (100% locally cached) took 12.445868634s.
(remote_snapshot_sharing=true) Bazel build (30% locally cached) took 23.409956374s. 70% artifacts fetched remotely.
(remote_snapshot_sharing=true) Bazel build (0% locally cached) took 47.780942207s. 100% artifacts fetched remotely.

PR #5372, opened 2023-11-27
Don't allow funky agg types

tylerwilliams:
Flagged during a pentest, we were swallowing these errors and breaking the sql query. Instead, just return an error.


tylerwilliams:
Fixes: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2928

PR #5376, opened 2023-11-28
[RB] Support remote snapshot sharing for remote bazel

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/hostedrunner/hostedrunner.go
+++ b/enterprise/server/hostedrunner/hostedrunner.go
@@ -174,6 +174,7 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest, 
 			// Run from the scratch disk, since the workspace disk is hot-swapped
 			// between runs, which may not be very Bazel-friendly.
 			{Name: "WORKDIR_OVERRIDE", Value: "/root/workspace"},
+			{Name: "GIT_BRANCH", Value: req.GetRepoState().GetBranch()},
 		},
 		Arguments: args,
 		Platform: &repb.Platform{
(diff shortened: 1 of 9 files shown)


enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -178,6 +179,7 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest,
 				{Name: "container-image", Value: RunnerContainerImage},
 				{Name: "recycle-runner", Value: "true"},
 				{Name: "workload-isolation-type", Value: "firecracker"},
+				{Name: platform.WorkloadTypePropertyName, Value: "remote-bazel"},
```
bduffany, line 182: I would prefer to avoid introducing another platform property since these are used as the snapshot key and will make it harder to make a change that we want to do in the future, which is sharing snapshots between remote bazel and workflows.

Can we instead change the isWorkflow logic to isCIRunner and have it check `Command.Arguments[0] == "./buildbuddy_ci_runner"`? We do have some precedent for this [here](https://github.com/buildbuddy-io/buildbuddy/blob/a1845b6e9ee69aefe809f88b84f5d2abd29ea09d/enterprise/server/remote_execution/executor/executor.go#L122-L126) if you wanted to move that to a common location (maybe `platform.go`)

maggie-lou: Good idea - changed

enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -174,6 +174,7 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest,
 			// Run from the scratch disk, since the workspace disk is hot-swapped
 			// between runs, which may not be very Bazel-friendly.
 			{Name: "WORKDIR_OVERRIDE", Value: "/root/workspace"},
+			{Name: "GIT_BRANCH", Value: req.GetRepoState().GetBranch()},
```
bduffany, line 177: In a later PR it would be great if we could also set `GIT_REPO_DEFAULT_BRANCH` by looking up the repo metadata using the GitHub app if it's installed. Otherwise I think we'll have a cold run on every new branch.

bduffany (approved):
Nice!

enterprise/server/hostedrunner/hostedrunner.go
maggie-lou: Sounds good - will add that to my list

PR #5381, opened 2023-11-28
Upgrade podman to 4.7.2

bduffany:
4.6.2 was removed from the repo we're using.

Tested in dev. Sanity-checked perf, running `podman run --rm --net=none --transient-store ubuntu:22.04 true` still takes around 170ms.

**Related issues**: N/A


iain-macdonald (approved):
+1 to building from source or something else that doesn't put us at the whim of the alvistack maintainers

PR #5382, opened 2023-11-28
Update mmap to support memory limit changes

bduffany:
* Add a concurrency-safe `Unmap` function to allow safely unmapping chunks at any time. The plan is to call this from an LRU `OnEvict` callback, which will unmap the least recently used chunk in order to save memory.
* Add a concurrency-safe `Fetch` function and call this function when eager-fetching, instead of calling `initMap`. This makes sure that we don't `mmap` unless the chunk is actually needed, since `mmap` incurs extra memory usage and causes extra churn on the LRU if the chunk is not actually used.
* Now that `Unmap` can be called at any time, fix a concurrency issue in certain mmap operations where we momentarily releasing the lock after calling `initMap`. During this time that we're not holding the lock, it is valid for `Unmap` to be called, which would result in a panic when we try to access `data`.
* When converting a file to COW, don't eagerly `mmap` it. This should save a significant amount of memory usage today, and also reduce LRU churn once the LRU is implemented.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -322,6 +322,7 @@ func (c *COWStore) readChunk(p []byte, readRelativeOffset int64, chunkStartOffse
 	copy(p[dataSize:readSize], c.zeroBuf)
 	return nil
 }
+
 func (c *COWStore) WriteAt(p []byte, off int64) (int, error) {
 	if err := checkBounds("write", c.totalSizeBytes, p, off); err != nil {
 		return 0, err
@@ -560,14 +561,12 @@ func (s *COWStore) initDirtyChunk(offset int64, size int64) (ogChunk *Mmap, newC
 	s.storeLock.RUnlock()
 	chunkSource := snaputil.ChunkSourceHole
 	if ogChunk != nil {
-		if !ogChunk.safeReadMapped() {
-			if err := ogChunk.initMap(); err != nil {
-				return nil, nil, err
-			}
+		if err := ogChunk.initMap(); err != nil {
+			return nil, nil, err
 		}
 		chunkSource = ogChunk.source
 	}
-	newChunk, err = NewMmapFd(s.ctx, s.env, s.DataDir(), fd, int(size), offset, chunkSource, s.remoteInstanceName, s.remoteEnabled)
+	newChunk, err = NewMmapFd(s.ctx, s.env, s.DataDir(), true /*=dirty*/, fd, int(size), offset, chunkSource, s.remoteInstanceName, s.remoteEnabled)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -642,11 +641,8 @@ func (s *COWStore) fetchChunk(offset int64) error {
 		return nil
 	}
 
-	var err error
-	if !c.safeReadMapped() {
-		err = c.initMap()
-	}
-	return err
+	// Fetch, but don't mmap (since it incurs extra memory usage).
+	return c.Fetch()
 }
 
 // ConvertFileToCOW reads a file sequentially, splitting it into fixed size,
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -783,11 +779,16 @@ type Mmap struct {
 
 	Offset  int64
 	dataDir string
+	dirty   bool // dirty suffix, only used to compute path
+
+	// initMu is used only to prevent concurrent attempts to mmap the data.
+	initMu sync.Mutex
```
bduffany, line 785: @maggie-lou to answer your question from the other PR about whether this extra lock is really needed -

the old code for `ReadAt` looked like this:

```
initMap() // calls Lock / defer Unlock internally
// <-- at this location, we're not holding the lock - Unmap might be called here.
RLock()
defer RUnlock()
access m.data // unsafe access, since we might've been unmapped
```

in order to fix this, we could do:

```
Lock()
defer Unlock()
initMap() // don't lock internally
access m.data
```

But this will probably hurt concurrency since we're getting an exclusive lock. To fix this, we can do RLock() / RUnlock() in ReadAt, but have `initMap` internally have its own mutex (initMu). I don't think this will result in us being sloppier around concurrency, since initMu should only ever be needed inside initMap().

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -860,43 +877,60 @@ func mmapDataFromFd(fd, size int) ([]byte, error) {
...
-	data, err := mmapDataFromPath(outputPath)
+// Fetch fetches the chunk from cache if applicable.
+func (m *Mmap) Fetch() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.fetch(m.path())
+}
+
+func (m *Mmap) fetch(path string) error {
+	if m.fetched || m.lazyDigest == nil {
```
maggie-lou, line 916: Should we return an error if m.fetched == false && m.lazyDigest == nil? Otherwise in mmapDataFromPath we'll try to open a file that doesn't exist

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -921,82 +953,74 @@ func (m *Mmap) WriteAt(p []byte, off int64) (n int, err error) {
...
 		return err
 	}
+	m.data = nil
 	metrics.COWSnapshotMemoryMappedBytes.Set(float64(
 		atomic.AddInt64(&mmappedBytes, int64(-n)),
 	))
+	// TODO: this causes issues because when we unmap then remap due to LRU
+	// eviction, we lose the remote/filecache source information, since Fetch()
+	// won't be called again. Is it safe to just delete this line?
+	// m.source = snaputil.ChunkSourceUnmapped
```
maggie-lou, line 983: Do you want this in the PR?

maggie-lou (approved):

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: I think it's fine to just have mmapDataFromPath return a NotExist error.

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: woops - removed this chunk, including the line setting m.source = Unmapped

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
maggie-lou: From a debugging standpoint it might be helpful to know if the lazyDigest is not set, because that should never be the case and would indicate a bug. Don't feel that strongly though

bduffany: Fair, done. Had to set `fetched: true` in the non-lazy constructors for this to work.

maggie-lou:
* master
  * **PR #5382** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5382?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #5384** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5382?utm_source=stack-comment).

PR #5388, opened 2023-11-28
Dial* -> Dial*WithoutPooling in a few places

jdhollen:
This seems like the right choice for these lines regardless, but we could also try to mess with when we even use the otel interceptor?

**Related issues**: N/A

### DIFF
--- a/enterprise/server/build_event_publisher/build_event_publisher.go
+++ b/enterprise/server/build_event_publisher/build_event_publisher.go
@@ -86,7 +86,7 @@ func (p *Publisher) Start(ctx context.Context) {
 // run performs a single attempt to stream any currently buffered events
 // as well as all subsequently published events to the BES backend.
 func (p *Publisher) run(ctx context.Context) error {
-	conn, err := grpc_client.DialSimple(p.besBackend)
+	conn, err := grpc_client.DialSimpleWithoutPooling(p.besBackend)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
LGTM but maybe let @vadimberezniker have a look too :)

vadimberezniker:
I think the real problem is that grpc_client.DialInternal is being called for every streaming request, not the fact that we create a pool of connections. For build_event_publisher and telemetry_client I don't see any harm in creating a pool of connections. I'd prefer to keep the package simple. I was hopeful we could eventually remove the WithoutPooling variants of the functions.

jdhollen:
> I think the real problem is that grpc_client.DialInternal is being called for every streaming request, not the fact that we create a pool of connections. For build_event_publisher and telemetry_client I don't see any harm in creating a pool of connections. I'd prefer to keep the package simple. I was hopeful we could eventually remove the WithoutPooling variants of the functions.

Yeah, I mean, if we kill all ad-hoc `Dial` calls, then it doesn't matter.  I took a stab at that in #5396, and this was just something quick I got out because it is at least a sure thing and would slow down the bleed from this issue by an order of magnitude (in case the other fix doesn't work).

This only matters if we want a backup for the other change.

jdhollen:
the extra dial calls here are theoretically also histogram leaks, just not at the scale of the bytestream code--we can just use the reuse logic for them, too.

PR #5396, opened 2023-11-29
add a cache for grpc client connection pools.

jdhollen:
flag-guarded, because this freaks me out a bit.

currently only used in bytestream code--this will prevent us from dialing on every bytestream request, etc.

there are cleaner ways for this code to be written long-term, but this reads as trivially safe so it feels like the right place to start.

**Related issues**: N/A


jdhollen:
Aaaand this has a pretty glaring bug in it.  One sec.

jdhollen:
OK, all set.

server/environment/environment.go
```diff
@@ -66,6 +66,8 @@ type Env interface {
 	GetActionCacheClient() repb.ActionCacheClient
 	GetByteStreamClient() bspb.ByteStreamClient
 	SetByteStreamClient(bspb.ByteStreamClient)
+	GetGrpcClientConnPoolCache() interfaces.GrpcClientConnPoolCache
```
vadimberezniker, line 69: Does this need to be an app wide service? It seems simpler to me to have the localized use cases instantiate their own cache as I don't expect this to be a very common pattern. At this time it's only going to be used in a single place. 

cacheproxy.go has a similar implementation but it has some quirks that I think that would make it hard to reuse the same code for both.

server/util/grpc_client/grpc_client.go
```diff
@@ -216,3 +223,64 @@ func CommonGRPCClientOptions() []grpc.DialOption {
...
+	connMutex   *sync.RWMutex
+	connPoolMap map[string]interfaces.ClosableClientConn
+}
+
+func RegisterConnPoolCache(env environment.Env) {
+	c := NewGrpcClientConnPoolCache(env)
+	env.SetGrpcClientConnPoolCache(c)
+}
+
+func (cpc *grpcClientConnPoolCache) GetGrpcClientConnPoolForURL(target string) (conn interfaces.ClosableClientConn, shouldClose bool, err error) {
```
vadimberezniker, line 238: Having to return and check this shouldClose bool is a bit unfortunate. 

Could you use an LRU to maintain the cache? you can wait for connections to go idle before closing them to avoid closing active connections.

server/util/grpc_client/grpc_client.go
```diff
@@ -216,3 +223,64 @@ func CommonGRPCClientOptions() []grpc.DialOption {
...
+	cpc.connMutex.RLock()
+	connPool, ok := cpc.connPoolMap[target]
```
vadimberezniker, line 240: i think you are overcomplicating the locking here. Dialing, by default, is a non-blocking operation. You should be okay to take a lock here for the duration of the function.

server/util/grpc_client/grpc_client.go
jdhollen: maybe i'm missing something, but i didn't use an LRU for now because if a client has just grabbed a connection pool and the LRU decides to evict / close the pool, then the caller will be out of luck--the connection won't get out of an "Idle" state unless the caller tries to do something with it, and otherwise i'm in the same weird space where i need to lock / explicitly close, right?

separately, i didn't see a really awesome way to tell if a handle was held by the caller for a long time or not here.  i guess if i make the assumption that this is only called by bytestream code, then there's a concrete "end" to the connection being used, but i think i kind of still need a real signal that everyone's "done" with connections, right?

server/util/grpc_client/grpc_client.go
jdhollen: To be clear, I just did this for the sake of avoiding making two pools for the same connection.  There's a secondary thing where I was worried about having different threads all wait in line to grab a lock just to get a connection, but Tyler showed me some benchmarks that also convinced me this probably isn't worth it.  You're just saying I should use a plain mutex for the whole call, right?

server/environment/environment.go
jdhollen: I had been kidding myself that maybe we should share this with other code that's dialing (i.e., share pools between all of that code as well), but the cacheproxy code makes me... not think so.

Anyway, this was mostly out of expedience (this approach made me not have to convert the byte stream client to an object so that it could hold a reference to the cache).  I was already planning to convert it to an object, it's just a matter of if I do it now or later.

server/util/grpc_client/grpc_client.go
vadimberezniker: > maybe i'm missing something, but i didn't use an LRU for now because if a client has just grabbed a connection pool and the LRU decides to evict / close the pool, then the caller will be out of luck--the connection won't get out of an "Idle" state unless the caller tries to do something with it, and otherwise i'm in the same weird space where i need to lock / explicitly close, right?

this is what i was thinking we could do on eviction in background, in pseudo-code:
   check if at least N seconds has passed since the pool was created to make sure client had a chance to use it, if not wait before cleaning up
   wait for pool connections to go idle
   close connections
      
Would that work?

server/util/grpc_client/grpc_client.go
vadimberezniker: > You're just saying I should use a plain mutex for the whole call, right?

Yeah, you'll be fine just grabbing the lock for the duration of the entire function. It will make the code a lot simpler.

jdhollen: Done

server/util/grpc_client/grpc_client.go
```diff
@@ -216,3 +224,55 @@ func CommonGRPCClientOptions() []grpc.DialOption {
...
+func isPermittedForDial(target string) bool {
+	u, err := url.Parse(target)
+	if err != nil {
+		return false
+	}
+	if cache_api_url.String() == "" {
+		return true
+	}
+
+	return u.Hostname() == "localhost" || (urlutil.GetDomain(u.Hostname()) == urlutil.GetDomain(cache_api_url.WithPath("").Hostname()))
```
vadimberezniker, line 248: this logic seems a bit too high level to live in a cache implementation? what do you think about moving this to the byte_stream_client code?

server/util/grpc_client/grpc_client.go
```diff
@@ -216,3 +224,55 @@ func CommonGRPCClientOptions() []grpc.DialOption {
...
+
+func (cpc *grpcClientConnPoolCache) GetGrpcClientConnPoolForURL(target string) (conn interfaces.ClosableClientConn, err error) {
+	cpc.connMutex.Lock()
+	defer cpc.connMutex.Unlock()
+	connPool, ok := cpc.connPoolMap[target]
+	if ok && connPool != nil {
+		return connPool, nil
+	}
+	if !isPermittedForDial(target) {
+		return nil, status.InvalidArgumentErrorf("Tried to download from an unpermitted domain: %s", target)
```
vadimberezniker, line 259: nit: make this error message more generic, nothing is being downloaded here

server/util/grpc_client/grpc_client.go
```diff
@@ -216,3 +224,55 @@ func CommonGRPCClientOptions() []grpc.DialOption {
...
+	}
+	cpc.connPoolMap[target] = connPool
+	log.Infof("Cached connection pool for target: %s, %d targets cached so far", target, len(cpc.connPoolMap))
+	return connPool, nil
+}
+
+func NewGrpcClientConnPoolCache(env environment.Env) *grpcClientConnPoolCache {
+	return &grpcClientConnPoolCache{
+		env:         env,
+		connMutex:   &sync.Mutex{},
```
vadimberezniker, line 275: nit: it's more idiomatic to have mutexes be non-pointer types in structs, you wouldn't need this initializing code if that was the case

vadimberezniker (changes requested):

server/libmain/libmain.go
```diff
@@ -260,7 +260,11 @@ func registerGRPCServices(grpcServer *grpc.Server, env environment.Env) {
 func registerLocalGRPCClients(env environment.Env) error {
 	// Identify ourselves as an app client in gRPC requests to other apps.
 	usageutil.SetClientType("app")
+	if grpc_client.PoolCacheEnabled() {
+		grpc_client.RegisterConnPoolCache(env)
```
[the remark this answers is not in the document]
jdhollen: Done, but threw a Maybe on the fn name because it'd drive me nuts otherwise.

server/interfaces/interfaces.go
```diff
@@ -280,6 +281,15 @@ type StoppableCache interface {
 	Stop() error
 }
 
+type ClosableClientConn interface {
+	grpc.ClientConnInterface
+	Close() error
+}
+
+type GrpcClientConnPoolCache interface {
+	GetGrpcClientConnPoolForURL(target string) (ClosableClientConn, error)
```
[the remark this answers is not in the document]
jdhollen: Done

server/util/grpc_client/grpc_client.go
jdhollen: Thanks, if there's one thing I'm not it's Idiomatic With Go

server/libmain/libmain.go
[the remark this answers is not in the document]
vadimberezniker: We use this Register pattern everywhere so using Maybe here is going to stand out.

server/remote_cache/byte_stream_client/byte_stream_client.go
```diff
@@ -169,26 +172,59 @@ func stripUser(u *url.URL) *url.URL {
...
+		return true
+	}
+
+	return u.Hostname() == "localhost" || (urlutil.GetDomain(u.Hostname()) == urlutil.GetDomain(cache_api_url.WithPath("").Hostname()))
+}
+
+func streamFromUrl(ctx context.Context, env environment.Env, url *url.URL, grpcs bool, offset int64, limit int64, writer io.Writer) (err error) {
+	target := getTargetForURL(url, grpcs)
+	var conn grpc.ClientConnInterface
+	if grpc_client.PoolCacheEnabled() {
```
vadimberezniker, line 211: use a different flag here instead of depending on pool caching being enabled?

vadimberezniker (approved):

server/libmain/libmain.go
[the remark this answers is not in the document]
jdhollen: > We use this Register pattern everywhere so using Maybe here is going to stand out.

alright, alright, i'll just leave it on the record here that i'm _not happy about this!_ fixed.

server/remote_cache/byte_stream_client/byte_stream_client.go
jdhollen: Done

tylerwilliams (approved):

PR #5399, opened 2023-11-29
[SS] Fix unmapping chunk too early when merging diff snapshot

maggie-lou:
When merging the diff snapshot, the buffer that we use to copy the data over is smaller than the chunk size, so we may write to each COWStore chunk multiple times. We want to make sure not to unmap the chunk until all those writes have finished, or we'll do a lot of unnecessary maps and unmaps

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -628,6 +628,7 @@ func MergeDiffSnapshot(ctx context.Context, baseSnapshotPath string, baseSnapsho
 	defer span.End()
 
 	var out io.WriterAt
+	var storeChunkSizeBytes int64
 	if baseSnapshotStore == nil {
 		f, err := os.OpenFile(baseSnapshotPath, os.O_WRONLY, 0644)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


maggie-lou:
* master
  * **PR #5399** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5399?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5399?utm_source=stack-comment).

bduffany (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -693,25 +693,27 @@ func MergeDiffSnapshot(ctx context.Context, baseSnapshotPath string, baseSnapsho
...
 
 				if _, err := out.WriteAt(buf[:n], offset); err != nil {
 					return err
 				}
 
-				// After the store has been updated, unmap the chunk to save memory
+				// If we've finished processing a chunk, unmap it to save memory
 				// usage on the executor
-				if baseSnapshotStore != nil {
+				nextOffsetInNextChunk := (offset%cowChunkSizeBytes())+int64(n) >= cowChunkSizeBytes()
```
bduffany, line 707: uber nit: should we call `baseSnapshotStore.ChunkSize()` so we're using the actual chunk size of the file? (in case we wind up changing the chunk size later)

PR #5404, opened 2023-11-29
Add a filter to the timing profile

siggisim:
This allows you to search the timing profile by event name, target, mnemonic, category, or output name.

Event that match will be shown in color, while all other events will be greyed out.

Example:
Screenshot 2023-11-29 at 2 07 49 PM

### DIFF
--- a/app/trace/constants.ts
+++ b/app/trace/constants.ts
@@ -27,12 +27,12 @@ export const SECTION_PADDING_BOTTOM = 1;
 
 export const TRACK_HEIGHT = 16;
 export const TRACK_VERTICAL_GAP = 1;
-export const TRACK_FILTERED_OUT_COLOR = "#ddd";
 
 export const EVENT_HORIZONTAL_GAP = 0.5;
 export const EVENT_LABEL_WIDTH_THRESHOLD = 20;
 export const EVENT_LABEL_FONT_SIZE = "11px";
 export const EVENT_LABEL_FONT_COLOR = SECTION_LABEL_FONT_COLOR;
+export const EVENT_FILTERED_OUT_COLOR = "#ddd";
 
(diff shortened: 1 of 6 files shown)


app/trace/constants.ts
```diff
@@ -27,6 +27,7 @@ export const SECTION_PADDING_BOTTOM = 1;
 
 export const TRACK_HEIGHT = 16;
 export const TRACK_VERTICAL_GAP = 1;
+export const TRACK_FILTERED_OUT_COLOR = "#ddd";
```
bduffany, line 30: uber nit: EVENT_FILTERED_OUT_COLOR and move to the group below? since it applies to individual events

app/trace/trace_viewer.tsx
```diff
@@ -84,6 +89,10 @@ export default class TraceViewer extends React.Component<TraceViewProps, {}> {
         passive: false,
       });
     }
+    if (this.filterInputRef.current) {
+      this.filterInputRef.current.addEventListener("keyup", () => this.updateFilter());
```
bduffany, line 93: would it work to write

`<FilterInput onChange={(e) => this.updateFilter(e.target.value)} ... />`

instead of using a ref + addEventListener? (I don't think it would trigger a full re-render of this TraceViewer component on each keystroke, but could be mistaken)

bduffany (approved):

app/trace/trace_viewer.tsx
siggisim: Nice, done. I initially went this way - but switched because I wasn't using state. But now that I'm driving everything off the URL param it's pretty clean.

app/trace/constants.ts
siggisim: Good catch, done.

bduffany (approved):

PR #5411, opened 2023-11-30
Don't fail buildfix on gofmt errors

maggie-lou:
Gofmt will sometimes fail on proto symlink errors. Don't fail the rest of the buildfix script in these cases.

**Related issues**: N/A

### DIFF
--- a/buildfix.sh
+++ b/buildfix.sh
@@ -40,7 +40,7 @@ echo "Building and running gofmt..."
 GO_SRCS=()
 while IFS= read -r line; do
     GO_SRCS+=("$line")
-done <<< "$(git ls-files '*.go')"
+done < <(git ls-files '*.go')
 bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w "${GO_SRCS[@]}"
 
 if which clang-format &>/dev/null; then
--- a/buildfix.sh
+++ b/buildfix.sh
@@ -37,7 +37,11 @@ echo "Formatting WORKSPACE/BUILD files..."
 buildifier -r .
 
 echo "Building and running gofmt..."
-bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w .
+gofmt_exit_code=$(bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w . || echo $?)
+
+if [ "$gofmt_exit_code" -ne 0 ]; then
+  echo "Error occurred during gofmt, but continuing..."
+fi
 
 if which clang-format &>/dev/null; then
   echo "Formatting .proto files..."
--- a/buildfix.sh
+++ b/buildfix.sh
@@ -37,11 +37,11 @@ echo "Formatting WORKSPACE/BUILD files..."
 buildifier -r .
 
 echo "Building and running gofmt..."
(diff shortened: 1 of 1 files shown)


tylerwilliams:
Is there a way to prevent the proto symlink errors from happening?

maggie-lou:
> Is there a way to prevent the proto symlink errors from happening?

I don't fully understand the proto symlink setup, if someone else has a better idea/suggestions here. But the errors I'm seeing in gofmt are of the sort: "open proto/nbd/nbd.pb.go: no such file or directory". I use the symlinked protos so that my IDE can use protos in auto-complete

buildfix.sh
```diff
@@ -37,7 +37,11 @@ echo "Formatting WORKSPACE/BUILD files..."
 buildifier -r .
 
 echo "Building and running gofmt..."
-bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w .
+gofmt_exit_code=$(bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w . || echo $?)
+
+if [ "$gofmt_exit_code" -ne 0 ]; then
+  echo "Error occurred during gofmt, but continuing..."
+fi
```
bduffany, line 44: If `bazel run` puts anything on stdout other than empty string `""` then this logic will break.

Can instead write

```
bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w . ||
    echo "Error occurred during gofmt, but continuing..."
```

buildfix.sh
```diff
@@ -37,7 +37,11 @@ echo "Formatting WORKSPACE/BUILD files..."
 buildifier -r .
 
 echo "Building and running gofmt..."
-bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w .
+gofmt_exit_code=$(bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w . || echo $?)
```
bduffany, line 40: to prevent the symlink errors, we could replace this line with

```
mapfile -t GO_SRCS < <(git ls-files '*.go')
bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w "${GO_SRCS[@]}"
```

maggie-lou: Cool! Thanks Brandon! 🙏 Macs don't have mapfile, but replaced with something equivalent.

bduffany: I think gofmt is capable of following symlinks and processing them, but the issue in your case is that there was a "dangling" symlink (nbd.proto) which we deleted, and so it no longer exists under the bazel output base. If you had run `git clean -fdx` then bazel build and re-run `sgp` to symlink protos, that would probably have fixed the issue as well.

buildfix.sh
```diff
@@ -37,7 +37,11 @@ echo "Formatting WORKSPACE/BUILD files..."
 buildifier -r .
 
 echo "Building and running gofmt..."
-bazel run "${BAZEL_QUIET_FLAGS[@]}" //:gofmt -- -w .
+GO_SRCS=()
+while IFS= read -r line; do
+    GO_SRCS+=("$line")
+done <<< "$(git ls-files '*.go')"
```
bduffany, line 43: uber nit:

```suggestion
done < <(git ls-files '*.go')
```

this results in a pipe that streams the output into the while loop instead of buffering the full output into a string first.

bduffany (approved):

PR #5412, opened 2023-11-30
Don't hold runner pool lock while unpausing

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -1058,26 +1058,42 @@ func (p *pool) String() string {
 // operation fails on a given attempt, the runner is removed from the pool.
 func (p *pool) takeWithRetry(ctx context.Context, key *rnpb.RunnerKey) *commandRunner {
 	for i := 1; i <= maxUnpauseAttempts; i++ {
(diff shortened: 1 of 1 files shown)


maggie-lou (approved):

bduffany:
> IIRC, we have the lock around Unpause to prevent something else from Removing it in the background at the same time and causing panics. Is that still a concern?

I think maybe you're thinking of the locking that's happening in TracedCommandContainer?

But also this scenario shouldn't be possible because we're transitioning the runner to `ready` state before returning it from this func (and while the lock is held), and our desired behavior for the `ready` state is that a runner can't be touched by anything other than the goroutine that's executing the current task. Separately I wanted to rename the states from `paused => available` and `ready => inUse` since that's really all the states are describing (not the actual state of the underlying container).

bduffany:
> On executor shutdown, we call Shutdown. This calls Container.Remove()

On shutdown we only call Remove() for containers that are in `paused` state - containers that are in `ready` state will hit the code here on the next recycling attempt: 

https://github.com/buildbuddy-io/buildbuddy/blob/5732aca114fb1d0b25749eb3baf89e8da3b31663/enterprise/server/remote_execution/runner/runner.go#L645-L653

PR #5414, opened 2023-11-30
Add a comment explaining how to run firecracker_test remotely.

iain-macdonald:
The only other place I could find this is in the workflow setup, but LMK if you think it belongs somewhere else!

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/BUILD
+++ b/enterprise/server/remote_execution/containers/firecracker/BUILD
@@ -81,7 +81,7 @@ genrule(
 )
 
 # To remotely execute this test, a couple of tag_filters are needed:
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/BUILD
```diff
@@ -80,6 +80,9 @@ genrule(
     ],
 )
 
+# To remotely execute this test, a couple of tag_filters are needed:
+# bazel test --config=remote --test_tag_filters=+bare --build_tag_filters=+bare \
```
bduffany, line 84: I don't think the `build_tag_filters` are required? otherwise LGTM

iain-macdonald: Yeah, looks like it works without that option. Removed.

PR #5420, opened 2023-12-01
Add /channelz page

bduffany:
Looks like this on localhost:

image

**Related issues**: N/A

### DIFF
--- a/server/util/grpc_server/grpc_server.go
+++ b/server/util/grpc_server/grpc_server.go
@@ -24,6 +24,7 @@ import (
 
 	hlpb "github.com/buildbuddy-io/buildbuddy/proto/health"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
+	channelzservice "google.golang.org/grpc/channelz/service"
 	_ "google.golang.org/grpc/encoding/gzip" // imported for side effects; DO NOT REMOVE.
 )
 
@@ -145,6 +146,8 @@ func NewGRPCServer(env environment.Env, port int, credentialOption grpc.ServerOp
 	hlpb.RegisterHealthServer(grpcServer, env.GetHealthChecker())
 
 	regServices(grpcServer, env)
+
+	channelzservice.RegisterChannelzServiceToServer(grpcServer)
 
 	go func() {
 		_ = grpcServer.Serve(lis)
--- a/server/util/monitoring/monitoring.go
+++ b/server/util/monitoring/monitoring.go
@@ -1,6 +1,7 @@
 package monitoring
 
 import (
+	"fmt"
 	"net/http"
 	"net/http/pprof"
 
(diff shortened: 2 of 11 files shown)


server/util/monitoring/monitoring.go
```diff
@@ -45,14 +49,20 @@ func RegisterMonitoringHandlers(mux *http.ServeMux) {
 
 	// Flagz page
 	handle("/flagz", http.HandlerFunc(flagz.ServeHTTP))
+
+	// Channelz page
+	handle("/", channelz.CreateHandler("/", fmt.Sprintf("%s:%d", env.GetListenAddr(), grpc_server.Port())))
```
tylerwilliams, line 54: I think /channelz should show up on the monitoring port, similar to /statusz and /healthz?

server/util/monitoring/monitoring.go
```diff
@@ -21,7 +25,7 @@ var (
 
 // Registers monitoring handlers on the provided mux. Note that using
 // StartMonitoringHandler on a monitoring-only port is preferred.
-func RegisterMonitoringHandlers(mux *http.ServeMux) {
+func RegisterMonitoringHandlers(env environment.Env, mux *http.ServeMux) {
```
tylerwilliams, line 28: See other comment but I don't expect this function to need to take an env.

server/util/grpc_server/grpc_server.go
```diff
@@ -146,6 +147,8 @@ func NewGRPCServer(env environment.Env, port int, credentialOption grpc.ServerOp
 
 	regServices(grpcServer, env)
 
+	channelzservice.RegisterChannelzServiceToServer(grpcServer)
```
tylerwilliams, line 150: does this get all grpc servers, or just one of internal/external?

tylerwilliams (changes requested):

server/util/grpc_server/grpc_server.go
bduffany: woops, this does register the channelz service to all servers but I wasn't configuring the HTTP handler correctly to actually expose the channelz page for each server. Fixed.

server/util/grpc_server/grpc_server.go
```diff
@@ -146,6 +155,8 @@ func NewGRPCServer(env environment.Env, port int, credentialOption grpc.ServerOp
...
```
vadimberezniker, line 158: Does this mean the channelz rpc service will be exposed and accessible externally?

bduffany: hm good catch. I think I misunderstood how this works - this func is just exposing the same channelz service to each server, and the /channelz page contains results across all servers.

We don't have a "monitoring gRPC server" that this could be registered to, but would it make sense to register it to the internal gRPC server?

vadimberezniker: registering to the internal gRPC server sounds good

bduffany: OK, done

vadimberezniker (approved):

tylerwilliams (approved):

PR #5427, opened 2023-12-01
Remove test-only function NewPodmanCommandContainer and initialize PodmanCommandContainers via Provider instead

iain-macdonald:
**Related issues**: N/A


enterprise/server/test/integration/podman/podman_test.go
```diff
@@ -13,12 +13,15 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/commandutil"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/container"
+	_ "github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/containers/docker"
```
bduffany, line 16: nit: can you move this to the bottom of the import section and add a comment that this is included for flags?

bduffany (approved):
nice!

enterprise/server/test/integration/podman/podman_test.go
iain-macdonald: Oh shoot, didn't realize those were supposed to be at the bottom as well. Fixed.

bduffany (approved):

PR #5428, opened 2023-12-01
[BR] Upgrade bazel remote runner to use ubuntu 20.04 image

maggie-lou:
Seeing `GLIBC_2.28' not found` build errors on the 18.04 image

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -54,7 +54,7 @@ const (
 var (
 	execOs            = flag.String("os", "linux", "If set, requests execution on a specific OS.")
 	execArch          = flag.String("arch", "amd64", "If set, requests execution on a specific CPU architecture.")
-	runnerImage       = flag.String("runner_image", "", "If set, requests execution on a specific runner image. Otherwise uses the default hosted runner version.")
+	containerImage    = flag.String("container_image", "", "If set, requests execution on a specific runner image. Otherwise uses the default hosted runner version. A `docker://` prefix is required.")
 	defaultBranchRefs = []string{"refs/heads/main", "refs/heads/master"}
 )
 
@@ -524,7 +524,7 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 		BazelCommand:       strings.Join(bazelArgs, " "),
 		Os:                 reqOS,
 		Arch:               reqArch,
-		Image:              *runnerImage,
+		ContainerImage:     *containerImage,
 	}
 
 	req.GetRepoState().Patch = append(req.GetRepoState().Patch, repoConfig.Patches...)
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -19,7 +19,6 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/cli/arg"
 	"github.com/buildbuddy-io/buildbuddy/cli/log"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/dirtools"
-	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/platform"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
 	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/cachetools"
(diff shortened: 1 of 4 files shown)


maggie-lou:
* master
  * **PR #5428** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5428?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5428?utm_source=stack-comment).

tylerwilliams (approved):
I don't think anyone is using bazel remote atm so this seems fine, assuming that's all it will affect.

Would it make sense to have remote bazel use the defaults from the buildbuddy.yaml file in the repository? (and fall back to the default buildbuddy.yaml if there is none)

maggie-lou:
> I don't think anyone is using bazel remote atm so this seems fine, assuming that's all it will affect.
> 
> Would it make sense to have remote bazel use the defaults from the buildbuddy.yaml file in the repository? (and fall back to the default buildbuddy.yaml if there is none)

People can specify a container image for each action in buildbuddy.yaml, so it might get complicated to pick a single default. I added a flag to remote bazel to specify image instead

tylerwilliams:
> > I don't think anyone is using bazel remote atm so this seems fine, assuming that's all it will affect.
> > Would it make sense to have remote bazel use the defaults from the buildbuddy.yaml file in the repository? (and fall back to the default buildbuddy.yaml if there is none)
> 
> People can specify a container image for each action in buildbuddy.yaml, so it might get complicated to pick a single default. I added a flag to remote bazel to specify image instead

Flag looks good. I mostly am trying to get you to think about a plan for upgrading the default in the future. What usually happens is:
- customers start using it, say at version=ubuntu20
- we decide we want to change the default to ubuntu22 because ubuntu20 is EOL or something
- customers are unitentionally broken because the default changed and they depended on something ubuntu20 specific

Maybe we could make the default in the CLI "latest" or something and then decide that on the server? that way nobody is misled into thinking it will always be ubuntu20?

maggie-lou:
> > > I don't think anyone is using bazel remote atm so this seems fine, assuming that's all it will affect.
> > > Would it make sense to have remote bazel use the defaults from the buildbuddy.yaml file in the repository? (and fall back to the default buildbuddy.yaml if there is none)
> > 
> > 
> > People can specify a container image for each action in buildbuddy.yaml, so it might get complicated to pick a single default. I added a flag to remote bazel to specify image instead
> 
> Flag looks good. I mostly am trying to get you to think about a plan for upgrading the default in the future. What usually happens is:
> 
> * customers start using it, say at version=ubuntu20
> * we decide we want to change the default to ubuntu22 because ubuntu20 is EOL or something
> * customers are unitentionally broken because the default changed and they depended on something ubuntu20 specific
> 
> Maybe we could make the default in the CLI "latest" or something and then decide what version that is on the server? that way nobody is misled into thinking it will always be ubuntu20?

Are you thinking if someone reads the CLI code and sees ubuntu20 and makes assumptions about that? I changed the flag default to be "", so we don't set a default runner image until hostedrunner.go. Is that what you mean?

We'd have to notify customers if we upgrade the default image in hostedrunner.go, but that feels inevitable

cli/remotebazel/remotebazel.go
```diff
@@ -54,6 +54,7 @@ const (
 var (
 	execOs            = flag.String("os", "linux", "If set, requests execution on a specific OS.")
 	execArch          = flag.String("arch", "amd64", "If set, requests execution on a specific CPU architecture.")
+	runnerImage       = flag.String("runner_image", "", "If set, requests execution on a specific runner image. Otherwise uses the default hosted runner version.")
```
bduffany, line 57: Should we call this `container_image` for better consistency with the `container-image` exec property + `buildbuddy.yaml` config field?

proto/runner.proto
```diff
@@ -83,6 +83,9 @@ message RunRequest {
 
   // If true, start the runner but do not wait for the action to be scheduled.
   bool async = 9;
+
+  // Container image to run on. Defaults to the Ubuntu 20.04 Workflows image.
+  string image = 10;
```
bduffany, line 88: `container_image` here too, to match the exec property / bb.yaml name?

bduffany: also document that a `docker://` suffix is required?

cli/remotebazel/remotebazel.go
bduffany: This was marked resolved - forgot to push?

cli/remotebazel/remotebazel.go
```diff
@@ -54,6 +54,7 @@ const (
 var (
 	execOs            = flag.String("os", "linux", "If set, requests execution on a specific OS.")
 	execArch          = flag.String("arch", "amd64", "If set, requests execution on a specific CPU architecture.")
+	containerImage    = flag.String("container_image", "", "If set, requests execution on a specific runner image. Otherwise uses the default hosted runner version.")
```
bduffany, line 57: nit: add to the help description that docker/OCI images require a `docker://` prefix

bduffany (approved):

PR #5431, opened 2023-12-01
Upgrade executors to podman 4.8.0

iain-macdonald:
The issue was a combination of: inconsistent selection of the storage-driver between our code and the podman code between the `podman version` (which has an empty storage driver) call and the first `podman image exists` call (which uses the overlayfs storage driver in dev), and a [bug](https://github.com/containers/podman/issues/20731) in the new sqlite database they're using. I expect this will be fixed in 4.8.1 or 4.8.2, but it's probably wise to explicitly specify the storage driver for all podman commands anyways.

I have aspirations to refactor our podman command-running logic and add some tests, but will save those for another PR.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/2932


bduffany (approved):

bduffany:
test failures seem somewhat legit?

PR #5440, opened 2023-12-04
Don't double-buffer GCS uploads

tylerwilliams:


server/backends/blobstore/gcs/gcs.go
```diff
@@ -116,27 +116,43 @@ func (g *GCSBlobStore) WriteBlob(ctx context.Context, blobName string, data []by
...
+	}
+
+	lastErr := status.InternalError("GCS write not run")
+	r := retry.New(ctx, &retry.Options{
+		InitialBackoff: 50 * time.Millisecond,
+		MaxBackoff:     3 * time.Second,
+		Multiplier:     2,
+		MaxRetries:     3,
+	})
+	for r.Next() {
```
vadimberezniker, line 148: I've discovered that r.Next() can return false on the first attempt if the context is cancelled which would cause this function to return the wrong result. retry.Do does not suffer from this problem so I would recommend using that instead.

vadimberezniker (changes requested):

server/backends/blobstore/gcs/gcs.go
tylerwilliams: I think this is OK because I initialize lastError to status.InternalError("GCS write not run")

vadimberezniker (approved):

PR #5448, opened 2023-12-05
Update TestMergeDiffSnapshot to test base snapshot files with holes

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
@@ -1836,6 +1836,8 @@ func testMergeDiffSnapshot(t *testing.T, cow bool) {
 		// Make a buffer to keep track of our expected bytes.
 		b := make([]byte, snapSize)
 		// Create a base snapshot with some random data pages.
+		// Regions which are not written are holes, in order to test that
+		// we handle holes in the base snapshot properly.
 		basePath := filepath.Join(tmp, fmt.Sprintf("%d.base", i))
 		{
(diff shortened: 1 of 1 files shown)


bduffany:
Marking WIP because the tests are failing.

We can go ahead with https://github.com/buildbuddy-io/buildbuddy/pull/5450 to prevent dev crashing in the meantime.

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -1833,28 +1833,39 @@ func testMergeDiffSnapshot(t *testing.T, cow bool) {
 
 		ctx := context.Background()
 		snapSize := 1e6 + rand.Int63n(1e6)
-		// Create an empty base snapshot
-		basePath := filepath.Join(tmp, fmt.Sprintf("%d.base", i))
+		// Make a buffer to keep track of our expected bytes.
 		b := make([]byte, snapSize)
-		err := os.WriteFile(basePath, b, 0644)
+		// Create a base snapshot with some random data pages.
```
maggie-lou, line 1838: nit: Maybe add a comment that the unwritten pages are holes. If you're trying to test the hole logic, it might not be immediately obvious this is testing that

maggie-lou (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
bduffany: good call, done

PR #5450, opened 2023-12-05
Add nil check in UnmapChunk

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -434,6 +434,10 @@ func (s *COWStore) UnmapChunk(offset int64) error {
 	s.storeLock.RLock()
 	c := s.chunks[chunkStartOffset]
 	s.storeLock.RUnlock()
+	if c == nil {
+		// This offset contains a hole - do nothing.
+		return nil
+	}
 	return c.Unmap()
 }
 


enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -434,6 +434,10 @@ func (s *COWStore) UnmapChunk(offset int64) error {
 	s.storeLock.RLock()
 	c := s.chunks[chunkStartOffset]
 	s.storeLock.RUnlock()
+	if c == nil {
```
maggie-lou, line 437: This is a good idea but I don't think this is the panic we saw? It was on copy_on_write.go:1101

maggie-lou (approved):

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: That line is being reached from `UnmapChunk`, and the line is:

```
m.mu.Lock()
```

The nil pointer must be coming from the `m` dereference because it's impossible for `mu` to be nil as it's not a pointer.

bduffany: Another thing to note is that Go allows calling functions on `nil` objects, so `m.Unmap()` is valid to call even if `m` is nil.

maggie-lou: Dang what the hell 😨 Did not know that

PR #5451, opened 2023-12-05
style check tooling: minor fixes

sluongng:
Upgrade our `bb` CLI used in ./buildfix.sh to latest version.

Also ensure that the `bazel run //:gazelle` does not show up as a
separate PR check in CI..

### DIFF
--- a/buildfix.sh
+++ b/buildfix.sh
@@ -66,7 +66,7 @@ fi
 
 if ((GAZELLE)); then
   echo "Fixing BUILD deps with gazelle..."
-  CLI_VERSION="5.0.25" # Update this to latest version from `git tag -l 'cli-*' --sort=committerdate | tail -n1`
+  CLI_VERSION="5.0.25" # Update this to latest version from `git tag -l 'cli-*' --sort=creatordate | tail -n1`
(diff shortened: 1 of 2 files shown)


buildfix.sh
```diff
@@ -66,7 +66,7 @@ fi
 
 if ((GAZELLE)); then
   echo "Fixing BUILD deps with gazelle..."
-  CLI_VERSION="5.0.21" # keep in sync with checkstyle.yaml
+  CLI_VERSION="5.0.25" # Update this to latest version from `git lag -l | grep cli-`
```
bduffany, line 69:
```suggestion
  CLI_VERSION="5.0.25" # Update this to latest version from `git tag -l 'cli-*' --sort=committerdate | tail -n1`
```

bduffany (approved):

buildfix.sh
bduffany: err sorry I guess maybe `creatordate` instead of `committerdate`

sluongng: Switched to creatordate. Also tried `taggerdate` but it did not give the desired result.

PR #5459, opened 2023-12-06
Short circuit distributed writes without calling FindMissing.

tylerwilliams:
When doing a new CAS write today, we're running 6 find-missing calls beforehand: 3 in bytestream (which calls cache.FindMissing) and 3 in cache_proxy, which does the same for CAS objects.

This CL removes the extra FindMissing calls happening in cacheproxy, and instead swallows the error if the server returns `ALREADY_EXISTS` for a CAS blob.

At the time I originally wrote this, I did not fully appreciate how many small writes we get, and thought preventing a large write was a good tradeoff. I no longer believe that: the extra cost of so many extra FindMissing calls is harmful, especially given the number of small writes we handle.

### DIFF
--- a/proto/distributed_cache.proto
+++ b/proto/distributed_cache.proto
@@ -29,9 +29,11 @@ message ReadResponse {
   bytes data = 1;
 }
 
+// Next Tag: 9
 message WriteRequest {
   reserved 1;
   bool finish_write = 3;
+  bool allow_already_exists = 8;
   bytes data = 4;
   resource.ResourceName resource = 7;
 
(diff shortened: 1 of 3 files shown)


proto/distributed_cache.proto
```diff
@@ -29,9 +29,11 @@ message ReadResponse {
   bytes data = 1;
 }
 
+// Next Tag: 9
 message WriteRequest {
   reserved 1;
   bool finish_write = 3;
+  bool allow_already_exists = 8;
```
vadimberezniker, line 36: allow_already_exists=True does the opposite of the name since it returns an error if something already exists. Perhaps change this to `check_already_exists` and add a comment describing what it does?

vadimberezniker (approved):

PR #5471, opened 2023-12-07
[SS] Don't recycle runners if they exceed 90% disk or memory usage

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/executor/executor.go
+++ b/enterprise/server/remote_execution/executor/executor.go
@@ -10,10 +10,8 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/auth"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/commandutil"
-	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/containers/firecracker"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/operation"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/platform"
-	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/runner"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/metrics"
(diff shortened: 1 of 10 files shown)


maggie-lou:
* master
  * **PR #5471** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5471?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5471?utm_source=stack-comment).

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -341,6 +343,34 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
...
+
+		errStr := fmt.Sprintf("%v runner exceeded 90%% of memory or disk usage, not recycling: %s", r.GetIsolationType(), maxedOutStr)
+		if cr, ok := r.(*runner.CommandRunner); ok {
+			if fc, ok := cr.Container.Delegate.(*firecracker.FirecrackerContainer); ok {
+				errStr += fmt.Sprintf("\nSnapshot debug key: %s", fc.SnapshotDebugString(ctx))
+			}
+		}
+		alert.UnexpectedEvent("runner_maxed_out", errStr)
+		log.Errorf(errStr)
+	}
```
bduffany, line 372: Can we use the existing `doNotRecycle` bool in the `commandRunner` and move this logic to `Run()` ? I think that would make the PR a lot smaller

maggie-lou: Good idea - changed

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -374,7 +375,37 @@ func (r *commandRunner) Run(ctx context.Context) *interfaces.CommandResult {
...
+	// it, because that may cause failures if it's reused
+	maxedOutStr := ""
+	for _, fsUsage := range execResult.UsageStats.GetPeakFileSystemUsage() {
+		if float64(fsUsage.UsedBytes)/float64(fsUsage.TotalBytes) >= .9 {
+			maxedOutStr += fmt.Sprintf(" %d/%d B disk used for %s", fsUsage.UsedBytes, fsUsage.TotalBytes, fsUsage.GetSource())
+		}
+	}
+	// Only firecracker containers set max memory limits on the VMs
+	if fc, ok := r.Container.Delegate.(*firecracker.FirecrackerContainer); ok {
+		usedMemoryBytes := execResult.UsageStats.MemoryBytes
```
bduffany, line 390: while testing this locally I got a nil pointer exception at this line, so I think UsageStats is not guaranteed to be set. That's sort of an issue in and of itself which I'm digging into, but for now we should use `GetMemoryBytes()` to avoid the nil dereference

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -374,7 +375,37 @@ func (r *commandRunner) Run(ctx context.Context) *interfaces.CommandResult {
...
+
+	// If the runner has exceeded 90% of memory or disk usage, don't try to recycle
+	// it, because that may cause failures if it's reused
+	maxedOutStr := ""
+	for _, fsUsage := range execResult.UsageStats.GetPeakFileSystemUsage() {
+		if float64(fsUsage.UsedBytes)/float64(fsUsage.TotalBytes) >= .9 {
+			maxedOutStr += fmt.Sprintf(" %d/%d B disk used for %s", fsUsage.UsedBytes, fsUsage.TotalBytes, fsUsage.GetSource())
+		}
+	}
+	// Only firecracker containers set max memory limits on the VMs
```
bduffany, line 388: can we do the disk thing only for firecracker as well? want to make sure we have more time to test this before applying it to all runners

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -374,7 +375,37 @@ func (r *commandRunner) Run(ctx context.Context) *interfaces.CommandResult {
...
+	}
+
+	if maxedOutStr != "" {
+		r.doNotReuse = true
+
+		errStr := fmt.Sprintf("%v runner exceeded 90%% of memory or disk usage, not recycling: %s", r.GetIsolationType(), maxedOutStr)
+		if fc, ok := r.Container.Delegate.(*firecracker.FirecrackerContainer); ok {
+			errStr += fmt.Sprintf("\nSnapshot debug key: %s", fc.SnapshotDebugString(ctx))
+		}
+		alert.UnexpectedEvent("runner_maxed_out", errStr)
```
bduffany, line 404: @tylerwilliams just double-checking that you're OK with this since you're on call this week :)

tylerwilliams (approved):

bduffany (approved):

PR #5477, opened 2023-12-07
Add a runner-recycling-max-wait to the two runner recycling tests in remote_execution_test to reduce flakiness a bit

iain-macdonald:
**Related issues**: N/A


bduffany (approved):
nice!

PR #5482, opened 2023-12-08
Use vtproto Marshal/Unmarshal functions in grpc

luluz66:
**Related issues**: N/A


server/libmain/libmain.go
```diff
@@ -342,6 +343,9 @@ func StartAndRunServices(env *real_environment.RealEnv) {
 		log.Fatalf("%v", err)
 	}
 
+	// Register the codec for all RPC servers and clients.
+	vtprotocodec.Register()
```
vadimberezniker, line 347: Do you need to call this from executor main as well?

vadimberezniker (changes requested):

server/libmain/libmain.go
luluz66: Done.

vadimberezniker (approved):

PR #5492, opened 2023-12-11
rules: introduce remote_kaniko_build

sluongng:
An experimental rule that let us build Dockerfile in RBE instead of
using local docker daemon.

**Related issues**: N/A


sluongng:
closing this for now.
This experimentation was fruitful for me to write a blog post about it.

PR #5497, opened 2023-12-11
Add spans for PullImageIfNecessary and runPodman.

vadimberezniker:
**Related issues**: N/A


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -737,6 +738,9 @@ func (c *podmanCommandContainer) State(ctx context.Context) (*rnpb.ContainerStat
 }
 
 func runPodman(ctx context.Context, subCommand string, stdio *commandutil.Stdio, args ...string) *interfaces.CommandResult {
+	ctx, span := tracing.StartSpan(ctx)
+	tracing.AddStringAttributeToCurrentSpan(ctx, "podman.subcommand", subCommand)
+	defer span.End()
```
brentleyjones, line 743: @bduffany I assume we want sub-spans in here as well, to get better visibility into what each part os doing?

brentleyjones (approved):

PR #5502, opened 2023-12-11
Proto changes for showing ExecuteResponse in the UI

bduffany:
This PR adds the proto changes needed to show `ExecuteResponse` in the action page instead of `ActionResult`. Showing `ExecuteResponse` has two advantages:
* It means we'll show the actual `ExecuteResponse` for the execution ID, rather than the latest `ActionResult` associated with the action digest. The latter is inaccurate/incorrect and often confusing.
* The `ExecuteResponse` contains a superset of info, such as:
  * The `status` field (execution error details when the task either fails to start, or terminates abnormally), which is not representable by an ActionResult.
  * An optional `server_logs` field which allows attaching arbitrary logs. We can use this for attaching firecracker VM logs.

Note, this does **NOT** take any steps towards addressing https://github.com/buildbuddy-io/buildbuddy-internal/issues/2918 which is a bit more involved as it will likely involve DB schema changes.

**Related issues**:
* https://github.com/buildbuddy-io/buildbuddy-internal/issues/2774
* https://github.com/buildbuddy-io/buildbuddy-internal/issues/1203

### DIFF
--- a/proto/execution_stats.proto
+++ b/proto/execution_stats.proto
@@ -87,6 +87,15 @@ message Execution {
   // doesn't.
   build.bazel.remote.execution.v2.Digest action_result_digest = 9;
 
+  // A digest used to retrieve the
+  // [build.bazel.remote.execution.v2.ExecuteResponse] produced by the executor
+  // for the latest attempt of this execution.
+  //
+  // This digest can be used to fetch a specially formatted ActionResult from
+  // the Action Cache. In particular, the ActionResult's stdout_raw field is set
+  // to the marshaled ExecuteResponse bytes.
+  build.bazel.remote.execution.v2.Digest execute_response_digest = 12;
+
   // The stage this execution is currently in.
   build.bazel.remote.execution.v2.ExecutionStage.Value stage = 2;
 


proto/execution_stats.proto
```diff
@@ -87,6 +87,15 @@ message Execution {
   // doesn't.
   build.bazel.remote.execution.v2.Digest action_result_digest = 9;
 
+  // A digest used to retrieve the
+  // [build.bazel.remote.execution.v2.ExecuteResponse] produced by the executor
+  // for the latest attempt of this execution.
+  //
+  // This digest can be used to fetch a specially formatted ActionResult from
```
tylerwilliams, line 94: Kind of ambivalent here, but did you consider putting this in the CAS instead as the hashed(executeResponse) minus this field?

bduffany: Yes but then we have to store it somewhere. I am hesitant to add more columns to the Executions table but if it seems clearer that way, we can do that instead.

vadimberezniker (approved):

tylerwilliams (approved):

PR #5506, opened 2023-12-11
Client side support for http streaming

siggisim:
Client side support for server-side grpc streaming over http.

This is behind a `app.streaming_http_enabled` flag.

If that flag is enabled, we'll do three things: 
- We'll change our content-type from `application/proto` to `application/grpc+proto`
- We'll send proto messages as `Length-Prefixed-Message`s (see https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)
- We'll read the response as a stream rather than just a single arraybuffer.

There is a TODO in there to support messages that come across chunks. This should be relatively straightforward to implement, but I haven't run into it in my testing - and would like to run into it before fixing.

Depends on: https://github.com/buildbuddy-io/buildbuddy/pull/5505

### DIFF
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -171,7 +171,7 @@ class RpcService {
         while (true) {
           let result = await reader?.read();
           if (result?.done) {
-            break;
+            return;
           }
 
           // Ignore the first 5 byte length-prefix, see the comment for lengthPrefixMessage for more info.
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -175,6 +175,8 @@ class RpcService {
           }
 
           // Ignore the first 5 byte length-prefix, see the comment for lengthPrefixMessage for more info.
+          // TODO(siggisim): Support messages that come across flushes, though I'm not sure if this will
+          // happen much (if at all) in practice.
           let value = result?.value.slice(5) || new Uint8Array();
           callback(null, value);
           this.events.next(method.name);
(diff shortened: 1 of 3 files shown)


bduffany (commented):
How does this work in terms of the proto API? Will this let us write streaming RPC methods now or is there some additional codegen support needed? (IIRC the protobufjs codegen doesn't generate code for streaming APIs)

siggisim:
> How does this work in terms of the proto API? Will this let us write streaming RPC methods now or is there some additional codegen support needed? (IIRC the protobufjs codegen doesn't generate code for streaming APIs?)

Here's an example:
```
    rpc_service.service.on("data", (response, method) => {
      // Todo probably check that you're dealing with the right method here
      console.log(response);
    });

    rpc_service.service.streamBuildLog(request);
```

We could probably do something fancy to make this nicer, but I don't think this is so bad.

app/service/rpc_service.ts
```diff
@@ -163,6 +163,26 @@ class RpcService {
...
     try {
+      if (capabilities.config.streamingHttpEnabled) {
+        init.headers["Content-Type"] = "application/grpc+proto";
+        init.body = lengthPrefixMessage(requestData);
+
+        const reader = (await this.fetch(url, "stream", init))?.getReader();
+        while (true) {
+          let result = await reader?.read();
+          if (result?.done) {
+            break;
```
bduffany, line 174: should this be `return` ? Seems like we'll do another `fetch` after this `break`, guessing that's not intended?

siggisim: Good catch, I am struggling with `return`s today lol.

I tested this a bunch without the fallback case at the bottom - and then added the fallback to support hiding this behind a flag, and guess I didn't notice the extra fetch.

siggisim: Fixed

bduffany (approved):

PR #5513, opened 2023-12-12
show zip file contents in code ui

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -26,6 +26,8 @@ export type FileEncoding = "gzip" | "zstd" | "";
 
 export type FetchResponseType = "arraybuffer" | "stream" | "text" | "";
 
+export type BytestreamFileOptions = { filename?: string; zip?: string };
+
 class RpcService {
   service: ExtendedBuildBuddyService;
   regionalServices = new Map<string, ExtendedBuildBuddyService>();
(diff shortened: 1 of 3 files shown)


app/service/rpc_service.ts
```diff
@@ -115,6 +117,20 @@ class RpcService {
     ) as Promise<FetchPromiseType<T>>;
   }
 
+  fetchBytestreamFileWithOptions<T extends FetchResponseType = "text">(
+    bytestreamURL: string,
+    invocationId: string,
+    options: BytestreamFileOptions,
```
bduffany, line 123: wdyt about tacking this options param onto the existing method params? maybe we can make `init` an optional field of `BytestreamFileOptions` to make it less awkward to pass other options?

app/service/rpc_service.ts
```diff
@@ -26,6 +26,8 @@ export type FileEncoding = "gzip" | "zstd" | "";
 
 export type FetchResponseType = "arraybuffer" | "stream" | "text" | "";
 
+export type BytestreamFileOptions = { filename?: string; zip?: string };
```
bduffany, line 29: nit: document what these fields mean?

bduffany (approved):

app/service/rpc_service.ts
jdhollen: Sure, done

app/service/rpc_service.ts
jdhollen: sorry, dunno why i didn't do that before, guess i didn't have my typescript brain turned on.  i tacked it on, i'll shift the args around separately so that i'm not touching a bunch of random stuff here.

bduffany (approved):

PR #5519, opened 2023-12-12
[SS] Emit a custom metric instead of an alert if a runner won't get recycled due to max resource utilization

maggie-lou:
Having an alert per event is too noisy. Add a custom metric instead. 

PR to alert on the new metric: https://github.com/buildbuddy-io/buildbuddy-internal/pull/2973

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -382,15 +382,17 @@ func (r *commandRunner) Run(ctx context.Context) *interfaces.CommandResult {
 
 	execResult := r.Container.Exec(ctx, command, &commandutil.Stdio{})
 
-	r.checkMaxResourceUtilization(ctx, execResult.UsageStats)
+	if r.hasMaxResourceUtilization(ctx, execResult.UsageStats) {
+		r.doNotReuse = true
+	}
 
 	return execResult
 }
 
-func (r *commandRunner) checkMaxResourceUtilization(ctx context.Context, usageStats *repb.UsageStats) {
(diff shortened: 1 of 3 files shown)


maggie-lou:
* master
  * **PR #5519** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5519?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5519?utm_source=stack-comment).

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -381,39 +382,72 @@ func (r *commandRunner) Run(ctx context.Context) *interfaces.CommandResult {
 
 	execResult := r.Container.Exec(ctx, command, &commandutil.Stdio{})
 
+	r.checkMaxResourceUtilization(ctx, execResult.UsageStats)
```
bduffany, line 385: This has a side effect of setting `doNotRecycle` which affects control flow a bunch, so it may be worth calling out here.

```suggestion
	// Set `doNotRecycle` if the runner is close to its resource limits.
	r.checkMaxResourceUtilization(ctx, execResult.UsageStats)
```

bduffany (approved):

enterprise/server/remote_execution/runner/runner.go
maggie-lou: Moved the side effect out of the helper function, so the behavior is not as hidden

PR #5521, opened 2023-12-12
Refactor pebble cache benchmark test

luluz66:
**Related issues**: N/A
1. add different sizes to benchmark tests
2. remove Contain benchmark test (Contain is calling FindMissing)
3. increase log levels in benchmark tests

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache_test.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache_test.go
@@ -2867,7 +2867,7 @@ func BenchmarkGetMulti(b *testing.B) {
 	pc.Start()
 	defer pc.Stop()
 
-	sizes := []int64{1024, 1024 * 1024, 10 * 1024 * 1042}
+	sizes := []int64{1024, 1024 * 1024, 10 * 1024 * 1024}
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -2914,7 +2912,10 @@ func BenchmarkFindMissing(b *testing.B) {
 	}
 }
 
-func BenchmarkContains1(b *testing.B) {
```
vadimberezniker, line 2917: I think we should keep a separate "contains" benchmark for 2 reasons:
 - It's generally good to test the entirety of the public API. Contains happens to call FindMissing today but that doesn't have to generally be the case.
 -  Contains is a special case of FindMissing with a single digest which is not covered by the FindMissing benchmark which always does checks using batches.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache_test.go
luluz66: Added back.

PR #5530, opened 2023-12-13
Clean up lockmap gc goroutines

bduffany:
**Related issues**: N/A


tylerwilliams (approved):
Anywhere else we should be calling Close()?

bduffany:
> Anywhere else we should be calling Close()?

`pebble_cache.go` could benefit from a Close() but it's not strictly necessary. I'll send a separate PR for that in case I wind up making tests fail

PR #5532, opened 2023-12-13
Fix deadlock in MmapLRU eviction

bduffany:
Concurrent calls to Mmap.Unmap() and Mmap.ReadAt/WriteAt() (for a different Mmap instance) can result in a deadlock.

Say we have mmaps m1 and m2. What can happen is:

```
m1.Unmap()
> m1.mu.Lock()
> sharedLRU.Remove()

// Some other goroutine:
m2.initMap() // via ReadAt/WriteAt
> m2.mu.Lock()
> sharedLRU.Add(m1)
>> sharedLRU.mu.Lock() // succeeds, since m1.Unmap() has not yet acquired the lock
>> sharedLRU.lru.Add(m2) // actually add m2 to the LRU

// The above Add causes an eviction of m1!
sharedLRU.OnEvict():
> m1.unmap()
>> m1.mu.Lock() // blocks, since m1.Unmap() has the lock

// Back to m1 now (resuming from the first few lines)
>> sharedLRU.mu.Lock() // blocks, since m1.initMap() holds the lock

// Deadlock!
```

The solution in this PR is to handle evictions in OnEvict in the background, instead of blocking. This should also result in better `Mmap` performance since we no longer do a synchronous `munmap` syscall in `initMap` when the LRU is full.

**Related issues**: N/A


enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -1213,8 +1223,19 @@ func NewMmapLRU() (*MmapLRU, error) {
...
+			// holding the LRU lock here).
+			//
+			// If the evictors are not keeping up, start our own goroutine to
+			// ensure the eviction still happens.
+			select {
+			case ml.evictions <- m:
+			default:
+				log.Warningf("LRU evictors are not keeping up")
+				go ml.processEviction(m)
 			}
```
tylerwilliams, line 1239: I wonder if blocking here would be better than starting another goroutine? Basically as a way to apply backpressure, rather than starting a bunch of goroutines that then might also be backed up.

tylerwilliams (approved):
LGTM

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: We can't block here because then it could lead to the deadlock scenario again.

I will add an UnexpectedEvent here instead of a Warningf since I don't think we'll hit the channel limit in practice

maggie-lou (dismissed):
Damn - these race conditions just won't quit

maggie-lou (approved):

PR #5552, opened 2023-12-14
Add trace span annotation showing result of execution_server.execute()

iain-macdonald:
**Related issues**: [2997](https://github.com/buildbuddy-io/buildbuddy-internal/issues/2997)

### DIFF
--- a/enterprise/server/remote_execution/execution_server/execution_server.go
+++ b/enterprise/server/remote_execution/execution_server/execution_server.go
@@ -663,6 +663,8 @@ func (s *ExecutionServer) execute(req *repb.ExecuteRequest, stream streamLike) e
 	executionID := ""
 	if !req.GetSkipCacheLookup() {
 		if actionResult, err := s.getActionResultFromCache(ctx, adInstanceDigest); err == nil {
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -686,6 +688,7 @@ func (s *ExecutionServer) execute(req *repb.ExecuteRequest, stream streamLike) e
 			if ee != "" {
 				ctx = log.EnrichContext(ctx, log.ExecutionIDKey, ee)
 				log.CtxInfof(ctx, "Reusing execution %q for execution request %q for invocation %q", ee, downloadString, invocationID)
+				tracing.AddStringAttributeToCurrentSpan(ctx, "task_id", "merged against "+ee)
```
brentleyjones, line 691: This can be useful when looking at trace, but not really while searching for one. Given that, I'm ambivalent on these being added.

@tylerwilliams thoughts on if these values should use this tag?

iain-macdonald: Yeah, I toyed with the idea of just adding a new tag instead that shows which path the code took through here. Just pushed that version, LMK what you think.

brentleyjones (approved):

PR #5553, opened 2023-12-14
[RB] Add flag to set remote runner target

maggie-lou:
Ex. If you want the remote runner to run in dev, you can use the command
bb --remote_runner_target=remote.buildbuddy.dev remote build //... --remote_header=x-buildbuddy-api-key=X

This also adds support for the other flags defined at the top of remotebazel.go. I don't think they were being parsed correctly before


maggie-lou:
* `master`
  * **PR #5553** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5553?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5553?utm_source=stack-comment).

cli/arg/arg.go
```diff
@@ -141,6 +156,23 @@ func GetBazelArgs(args []string) []string {
...
+	for splitIndex < len(args) {
+		if !strings.HasPrefix(args[splitIndex], "--") {
+			break
+		}
+		splitIndex++
+	}
+	cliArgs := args[:splitIndex]
+	remainingArgs := args[splitIndex:]
+	return cliArgs, remainingArgs
+}
```
bduffany, line 174: I think this logic means that the `bb` flags have to take the `--arg=value` form, which might be confusing because Bazel supports passing args like `--arg value` (without the `=`).

Go's `flag.Parse` should stop parsing as soon as it sees the first non-flag arg, which is what we want in this case I think. Would it work to just use `flag.Parse` and then parse the remaining non-flag args using our existing parsing logic?

For example:

```
var verbose = flag.Bool(...)
os.Args = []string{"bb", "--target", "remote.buildbuddy.dev", "remote", "build" "--foo=bar"}
flag.Parse()
// Should stop parsing at `remote`
flag.Args() // should return []string{"remote" "build", "--foo=bar"}
```

PR #5558, opened 2023-12-14
Support partial client-side stream reads

siggisim:
I managed to reproduce the case where the server flushes a proto response without the full proto message.

This seems to happen when a single proto message is larger than ~4kb.

This PR handles this on the client side (and completes the TODO) by:
- Reading the length-prefix at the beginning of the `Length-Prefixed-Message`
- If we get a stream read that is smaller than length of the message, create a Uint8Array that we can use to buffer the message until we've got the full thing
- Append to that buffer until we've got a full message
- Call the callback with the full buffer, and reset the buffer

### DIFF
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -186,47 +186,10 @@ class RpcService {
         init.body = lengthPrefixMessage(requestData);
 
         const reader = (await this.fetch(url, "stream", init))?.getReader();
-        let buffer = new Uint8Array();
-        let bufferedLength = 0;
-        let totalLength = 0;
-        while (true) {
-          let result = await reader?.read();
-          if (result?.done) {
-            return;
-          }
-
-          // Skip empty results so we don't have to deal with undefined values below.
-          if (!result?.value) {
-            continue;
-          }
-
-          // If we haven't seen a message yet, create a new buffer sized based on the length prefix.
-          if (totalLength == 0) {
-            totalLength = new DataView(result.value.buffer, 1, 4).getUint32(0, false /* big endian */);
-            let message = result.value.slice(5);
-            buffer = new Uint8Array(totalLength);
-            buffer.set(message);
-            bufferedLength = message.length;
-          } else if (bufferedLength < totalLength) {
-            // If we already have a buffer, just tack onto it.
-            buffer.set(result.value, bufferedLength);
-            bufferedLength += result.value.length;
-          }
-
-          // If we still haven't receieve the full message, loop again and read more until we do.
-          if (bufferedLength < totalLength) {
-            continue;
-          }
-
-          // We've got the full message, flush it to the callback so the full proto can be parsed.
-          callback(null, buffer);
+        transformLengthPrefixedReaderStreamToProtoMessageBytes(reader, (error, bytes) => {
           this.events.next(method.name);
-
-          // Now that we've flushed the buffer, reset it.
-          buffer = new Uint8Array();
-          bufferedLength = 0;
-          totalLength = 0;
-        }
+          callback(error, bytes);
+        });
       }
 
       const arrayBuffer = await this.fetch(url, "arraybuffer", init);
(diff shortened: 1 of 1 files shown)


app/service/rpc_service.ts
```diff
@@ -186,18 +186,46 @@ class RpcService {
         init.body = lengthPrefixMessage(requestData);
 
         const reader = (await this.fetch(url, "stream", init))?.getReader();
+        let buffer = new Uint8Array();
+        let bufferedLength = 0;
+        let totalLength = 0;
```
bduffany, line 191: uber nit: I got `buffered` and `total` confused while reading this, maybe `messageExpectedLength` might be more clear?

app/service/rpc_service.ts
```diff
@@ -186,18 +186,46 @@ class RpcService {
...
+          // If we haven't seen a message yet, create a new buffer sized based on the length prefix.
+          if (totalLength == 0) {
+            totalLength = new DataView(result.value.buffer, 1, 4).getUint32(0, false /* big endian */);
+            let message = result.value.slice(5);
+            buffer = new Uint8Array(totalLength);
+            buffer.set(message);
+            bufferedLength = message.length;
+          } else if (bufferedLength < totalLength) {
+            // If we already have a buffer, just tack onto it.
+            buffer.set(result.value, bufferedLength);
```
bduffany, line 212: This `buffer.set` will throw a `RangeError` if the server can return a chunk that contains the end of one message and the start of another (i.e. crossing chunk boundaries) - will that ever happen or do we always flush before sending a new chunk?

(I think `RangeError` is fine if we're not expecting it to happen but maybe worth adding a comment)

bduffany (approved):

app/service/rpc_service.ts
siggisim: Yeah, after some more testing - I managed to hit that case. Got a fix in progress.

bduffany: sweet - just as a heads up you might also hit the case where the length prefix crosses chunk boundaries - e.g. only 3 of the expected 4 bytes - I remember having to deal with that in my hackathon project haha

app/service/rpc_service.ts
siggisim: Done

app/service/rpc_service.ts
siggisim: I rewrote things a bit to make them a bit more readable, and added support for leftovers (i.e. the first case where we have the start of the next message, or even multiple full messages and a chunk) and split length prefixes.

Took a few attempts to get something that felt good, but I'm not unhappy with where it ended up. LMK what you think.

app/service/rpc_service.ts
```diff
@@ -238,6 +229,102 @@ function lengthPrefixMessage(requestData: Uint8Array) {
...
+    bufferedLength: 0,
+    messageExpectedLength: 0,
+    leftovers: new Uint8Array(),
+  };
+  while (true) {
+    // Start with any leftovers from the last turn.
+    let value: Uint8Array | undefined = streamState.leftovers;
+
+    // If we don't have any leftovers, read at least 5 bytes from the stream so we have at least a length prefix.
+    if (streamState.leftovers.length == 0) {
```
bduffany, line 254:
```suggestion
    let value: Uint8Array = streamState.leftovers;

    // If we don't have any leftovers, read at least 5 bytes from the stream so we have at least a length prefix.
    if (value.length === 0) {
```

bduffany (approved):

app/service/rpc_service.ts
bduffany: Looks good! The `readAtLeastNBytes` helps a lot

app/service/rpc_service.ts
```diff
@@ -238,6 +229,102 @@ function lengthPrefixMessage(requestData: Uint8Array) {
...
+      value = consumeLengthPrefix(streamState, value);
+    }
+
+    // Read the rest of the bytes into the buffer until it's full, saving any leftovers.
+    consumeMessageChunk(streamState, value);
+
+    // If we still haven't receive the full message, loop again and read more until we do.
+    if (streamState.bufferedLength < streamState.messageExpectedLength) {
+      continue;
+    }
```
bduffany, line 274: sorry I randomly thought about this after approving 😅  should this be

```suggestion
    // Read the rest of the bytes into the buffer until it's full, saving any leftovers.
    while (streamState.bufferedLength < streamState.messageExpectedLength) {
      consumeMessageChunk(streamState, value);
    }
```

or maybe it'd be simpler to use the `readAtLeastNBytes` func to read `bufferedLength - messageExpectedLength` bytes after consuming the length prefix. So that way it's just two calls to `readAtLeastNBytes` on each iteration

bduffany, line 274: sorry I randomly thought about this after approving - it looks like there should be a `readAtLeastNBytes()` call in here somewhere, reading at least `messageExpectedLength - value.length` bytes, before consuming the message chunk? Since otherwise it looks like we wrap around and attempt to consume a length prefix again in the case where streamState.leftovers.length === 0?

siggisim: Trying to think through this, I don't know how all three of these could be true:
- `streamState.bufferedLength < streamState.messageExpectedLength` i.e. we haven't filled up our buffer so we continue
- `streamState.messageExpectedLength == 0` so we consume another length prefix 

You mentioned this can happen if `streamState.leftovers.length == 0`, but I don't know how that can happen if `streamState.bufferedLength < streamState.messageExpectedLength`.

I totally believe that this is possible and you're right (maybe after multiple loops or something), I'm just trying to load the correct situation into my head so I can fix it properly.

siggisim: Would the case you're thinking about be handled if instead of checking that `streamState.leftovers.length == 0` we checked if we had less than 5 bytes of leftovers i.e.:
```
    // If we don't have any leftovers, read at least 5 bytes from the stream so we have at least a length prefix.
    if (streamState.leftovers.length < 5) {
      value = await readAtLeastNBytes(value, reader, 5);
    }
```

bduffany: Yeah that'd work, but I think it could be simplified because reading the length prefix and reading the message payload are both very similar - we want to read at least N bytes and then do something with the N bytes that we read. But currently it's a little bit hard to follow the control flow because we're using the nice `readAtLeastNBytes()` func for the length prefix, but not for the message payload.

I don't feel super strongly about it though since this code seems unlikely to be modified and the complexity is pretty self-contained.

siggisim: Took another crack, let me know what you think

bduffany (approved):

PR #5569, opened 2023-12-15
Refactor the cli to reduce duplicated cli command handling code

maggie-lou:
I noticed that each handler duplicates some logic. Do it upfront, which also makes it a little more clear how the command is expected to be parsed. Also add some clarifying comments


maggie-lou:
* `master`
  * **PR #5569** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5569?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5569?utm_source=stack-comment).

cli/plugin/plugin.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/git"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/creack/pty"
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
```
iain-macdonald, line 28: hmm, weird that this previously built? anyways, mind throwing a newline above so the aliased imports are separate from the unaliased ones?

iain-macdonald (approved):

cli/cmd/bb/bb.go
```diff
@@ -52,8 +52,7 @@ func run() (exitCode int, err error) {
 	// Record original arguments so we can show them in the UI.
 	originalArgs := append([]string{}, os.Args...)
 
-	// Handle global args that don't apply to any specific subcommand
-	// (--verbose, etc.)
+	// Configure verbose flag
```
bduffany, line 55: The previous comment was helpful IMO because if we want to add more CLI-global flags, this is where we should do it - otherwise the logic on line 76 (in this PR) won't necessarily work

cli/cmd/bb/bb.go
```diff
@@ -70,73 +69,62 @@ func run() (exitCode int, err error) {
...
+	case "add":
+		return add.HandleAdd(args)
+	case "analyze":
+		return analyze.HandleAnalyze(args)
+	case "wtf":
+		return ask.HandleAsk(args)
+	case "huh":
+		return ask.HandleAsk(args)
+	case "ask":
+		return ask.HandleAsk(args)
```
bduffany, line 102: uber nit: if you want you could also write this as

```
case "ask", "huh", "wtf":
  return ask.HandleAsk(args)
```

but the current code is fine too

cli/parser/parser.go
```diff
@@ -32,6 +32,25 @@ const (
...
+		"fix":      {},
+		"install":  {},
+		"login":    {},
+		"logout":   {},
+		"print":    {},
+		"remote":   {},
+		"update":   {},
+		"upload":   {},
+		"version":  {},
+	}
```
bduffany, line 52: There are starting to be quite a few places in the code that we have to update when adding a new command -

- The switch case in main.go
- This map
- The "help" map in `help.go`

We've forgotten to update the `help` section a few times before and it seems somewhat likely we can forget to update this too.

wdyt about adding a centralized command registry like

```
var Commands = []Command{
  {
    Name: "ask",
    Help: "...",
    Handler: ask.HandleAsk, // func(args []string) (exitCode int, err error)
    Aliases: []string{"wtf", "huh"},
  },
  // ...
}
```

then this `cliCommands` map, `help` output etc. can be derived from that registry.

bduffany (approved):
Thank you for this refactor, it was sorely needed :)

cli/plugin/plugin.go
maggie-lou: I think it's because my ide automatically removes unused imports and the parsing logic wasn't processing correctly that this was used, so it kept trying to remove it

cli/cmd/bb/bb.go
maggie-lou: Added a separate helper function to parse global CLI flags - will make it more clear what flags are supported and easier to add them in the future

cli/parser/parser.go
maggie-lou: Good suggestion - refactored

The one slightly messy part is that we have to handle the help command separately. Otherwise there will be a circular dependency because the commands package will import help for the help handler. And the help handler imports commands because it needs the full list of commands

bduffany (approved):

cli/commands/commands.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+type CliCommand struct {
+	Name    string
+	Help    string
+	Handler func(args []string) (exitCode int, err error)
+	Aliases []string
+}
+
+var CliCommands = []CliCommand{
```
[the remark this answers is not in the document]
maggie-lou: Renamed the package to cli_command.go because I added a package called bazel_command.go in a follow up PR

PR #5589, opened 2023-12-19
Refactor cli - extract code to run bazel commands to helper functions

maggie-lou:
**Related issues**: N/A


maggie-lou:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5589?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `master`
  * **PR #5569** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5589** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5589?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5589?utm_source=stack-comment).

cli/common/common.go
```diff
@@ -1,7 +1,12 @@
 package common
```
bduffany, line 1: A `common` package has the potential to become a dumping ground for random stuff. Maybe we can keep this in `//cli/version` but move `HandleVersion` to a separate `//cli/versioncmd` package to break the circular dep?

cli/cmd/bb/bb.go
```diff
@@ -97,149 +87,13 @@ func run() (exitCode int, err error) {
...
-		if err := p.PostBazel(outputPath); err != nil {
-			return -1, err
-		}
-	}
-
-	// TODO: Support post-run hooks?
-	// e.g. show a desktop notification once a k8s deploy has finished
-
-	return exitCode, nil
+	return bazel_command.Run(args, originalArgs)
```
bduffany, line 96: Can we just call the existing `bazelisk.Run` ? I don't think we need any of the functionality in `bazel_command` here (generating the `run` script, `watcher` etc.)

If this sounds about right, then I think we can avoid introducing the new `bazel_command` package. IMO it's kind of nice to have all of that logic in main rather than adding a layer of indirection.

bduffany (changes requested):

cli/cmd/bb/bb.go
maggie-lou: I wasn't sure if we needed the plugin and bazel pre-handler setup code for `version`, but I hear you that it feels like a little too much indirection.

I also wanted to extract the setup code to use for bazel remote, and thought the helper function could be useful in the future if we ever need to run another bazel command

maggie-lou: @bduffany I find the existing fallback behavior confusing, because it's unclear which logic is supposed to apply to bb version vs bb remote vs plain bazel commands (all of which are currently sharing the logic). I'm hoping this refactor will force us to be more explicit moving forward

For example my original goal coming in here was just to work on remote bazel, so naturally I went to the remotebazel.go file. It was confusing to realize that half of the remote bazel logic actually lived in bb.go, and that for some reason the version command was also running that code, and it wasn't clear whether it was supposed to or not

bduffany (approved):

PR #5596, opened 2023-12-19
Move image-resolving logic from sociartifactstore.go to oci.go

iain-macdonald:
This is in preparation for moving some of podman's authentication and image management into the go code so it should be faster most of the time.

**Related issues**: N/A


enterprise/server/util/oci/oci.go
```diff
@@ -110,3 +121,107 @@ func (c Credentials) String() string {
...
+			layerRef := imageRef.Context().Digest(d.String())
+			l, err := remote.Layer(layerRef, remoteOpts...)
+			if err != nil {
+				return err
+			}
+			// This issues a HEAD request for the layer.
+			_, err = l.Size()
+			return err
+		})
+	}
```
bduffany, line 218: I am not super familiar with this but just to double check, does podman do auth in the same way (checking access to all layers)? Or is it sufficient to just authorize access to the manifest?

enterprise/server/util/oci/oci.go
```diff
@@ -110,3 +121,107 @@ func (c Credentials) String() string {
 func (c Credentials) Equals(o Credentials) bool {
 	return c.Username == o.Username && c.Password == o.Password
 }
+
+func Resolve(ctx context.Context, imageName string, platform *rgpb.Platform, credentials Credentials) (v1.Image, error) {
```
bduffany, line 125: to simplify the API of this package, wdyt about having `Resolve` call `authorize()` as an implementation detail, rather than exposing `Authorized` as a separate func?

enterprise/server/util/oci/oci.go
```diff
@@ -110,3 +121,107 @@ func (c Credentials) String() string {
...
+		if err != nil {
+			return nil, status.UnknownErrorf("could not get image from descriptor: %s", err)
+		}
+		return img, nil
+	default:
+		return nil, status.UnknownErrorf("descriptor has unknown media type %q", remoteDesc.MediaType)
+	}
+}
+
+func Authorized(ctx context.Context, image v1.Image, imageName string, credentials Credentials) error {
```
bduffany, line 182: nit: call this `authorizeLayers` ?

enterprise/server/util/oci/oci.go
iain-macdonald: `podman pull`ing an image that's already been pulled just issues a GET request for the manifest. Given that, I went ahead and removed this function entirely because we don't need it for the re-authorization check we discussed and it's not really adding any value in sociartifactstore.go because the layers are fetched later on for indexing anyways.

enterprise/server/util/oci/oci.go
iain-macdonald: Removed this function per above comment.

enterprise/server/util/oci/oci.go
iain-macdonald: Removed this function per above comment.

bduffany (approved):
nice!

PR #5599, opened 2023-12-20
Add backend support for fetching pull requests from github

siggisim:
Will add caching in a follow-up since it adds complexity - and this is already a decently sized PR.

Open to any feedback / suggestions to make this simpler or less complex.

I think most of the complexity comes from doing the two searches in parallel, and then doing all of the metadata fetching in parallel.

### DIFF
--- a/enterprise/server/githubapp/githubapp.go
+++ b/enterprise/server/githubapp/githubapp.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/webhooks/webhook_data"
@@ -34,6 +35,7 @@ import (
 	"github.com/golang-jwt/jwt"
 	"github.com/google/go-github/v43/github"
 	"golang.org/x/oauth2"
+	"golang.org/x/sync/errgroup"
 
 	gh_webhooks "github.com/buildbuddy-io/buildbuddy/enterprise/server/webhooks/github"
 	ghpb "github.com/buildbuddy-io/buildbuddy/proto/github"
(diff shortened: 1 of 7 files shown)


proto/github.proto
```diff
@@ -451,3 +451,33 @@ message CreateGithubRefRequest {
...
   context.ResponseContext response_context = 1;
 }
+
+message PullRequest {
+  uint64 number = 1;
+  string title = 2;
+  string body = 3;
+  string author = 4;
+  string url = 5;
+  map<string, Review> reviews = 6;
```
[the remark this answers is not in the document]
tylerwilliams: what is it 2023

enterprise/server/githubapp/githubapp.go
```diff
@@ -1342,3 +1344,142 @@ func (a *GitHubApp) CreateGithubRef(ctx context.Context, req *ghpb.CreateGithubR
...
+func (a *GitHubApp) populatePRMetadata(ctx context.Context, client *github.Client, prIssues []*github.Issue, currentUser *github.User) (map[int]*ghpb.PullRequest, error) {
+	prsMu := &sync.Mutex{}
+	prs := make(map[int]*ghpb.PullRequest, len(prIssues))
+	metadataGroup, metadataContext := errgroup.WithContext(ctx)
+	for _, i := range prIssues {
+		i := i
+		prsMu.Lock()
+		prs[i.GetNumber()] = issueToPullRequestProto(i)
+		prsMu.Unlock()
+		urlParts := strings.Split(*i.URL, "/")
```
tylerwilliams, line 1420: nit: prolly check the length of this so you don't accidentally hit a NPE on L1423

enterprise/server/githubapp/githubapp.go
```diff
@@ -1342,3 +1344,142 @@ func (a *GitHubApp) CreateGithubRef(ctx context.Context, req *ghpb.CreateGithubR
...
+
+		metadataGroup.Go(func() error {
+			pr, _, err := client.PullRequests.Get(metadataContext, urlParts[4], urlParts[5], i.GetNumber())
+			if err != nil {
+				return err
+			}
+			prsMu.Lock()
+			p := prs[i.GetNumber()]
```
tylerwilliams, line 1428: once you've got a hold of p, you can release the lock and edit p freely below (assuming other threads don't simultaneously edit p which i don't think they do)

tylerwilliams: ugh i lied i see in your other thread you are mutating p.Reviews

enterprise/server/githubapp/githubapp.go
```diff
@@ -1342,3 +1344,142 @@ func (a *GitHubApp) CreateGithubRef(ctx context.Context, req *ghpb.CreateGithubR
...
+	resp := &ghpb.GetGithubPullRequestResponse{
+		Outgoing: []*ghpb.PullRequest{},
+		Incoming: []*ghpb.PullRequest{},
+	}
+	for _, i := range outgoing.Issues {
+		resp.Outgoing = append(resp.Outgoing, prs[i.GetNumber()])
+	}
+	for _, i := range incoming.Issues {
+		resp.Incoming = append(resp.Incoming, prs[i.GetNumber()])
+	}
```
tylerwilliams, line 1371: you can cheat since append will take a varargs, and do:
```
resp.Outgoing = append(resp.Outgoing, prs...)
resp.Incoming = append(resp.Incoming, prs...)
```

tylerwilliams: You already do this yourself on L1357 :P

enterprise/server/githubapp/githubapp.go
```diff
@@ -1342,3 +1344,142 @@ func (a *GitHubApp) CreateGithubRef(ctx context.Context, req *ghpb.CreateGithubR
...
+		if err != nil {
+			return err
+		}
+		user = u
+		return nil
+	})
+	err := searchGroup.Wait()
+	if err != nil {
+		return nil, nil, nil, err
+	}
```
tylerwilliams, line 1407: if you don't need the err you can scope it
```
	if err := searchGroup.Wait(); err != nil {
		return nil, nil, nil, err
	}
```

enterprise/server/githubapp/githubapp.go
tylerwilliams: honestly doesn't matter for shit because you're prolly fetching a few hundred things or whatever right

tylerwilliams:
prolly wanna run this once with --config=race to make sure you don't have any lock issues

enterprise/server/githubapp/githubapp.go
siggisim: It's slightly different because I'm looking up the incoming pr in the `prs` map (which is a `ghpb.PullRequest` that contains metadata rather than a `github.IssuesSearchResult` which is what `outgoing.Issues` is a list of) using the pull request number and appending that.

This comment did help me realize though that PRs across repos can have the same PR number, so I switched this to github's global [NodeId](https://docs.github.com/en/graphql/guides/using-global-node-ids) instead.

enterprise/server/githubapp/githubapp.go
```diff
@@ -1342,3 +1344,142 @@ func (a *GitHubApp) CreateGithubRef(ctx context.Context, req *ghpb.CreateGithubR
...
+		resp.Incoming = append(resp.Incoming, prs[i.GetNumber()])
+	}
+	return resp, nil
+}
+
+func (a *GitHubApp) getIncomingAndOutgoingPRs(ctx context.Context, client *github.Client) (*github.IssuesSearchResult, *github.IssuesSearchResult, *github.User, error) {
+	var incoming *github.IssuesSearchResult
+	var outgoing *github.IssuesSearchResult
+	var user *github.User
+	searchGroup, searchContext := errgroup.WithContext(ctx)
```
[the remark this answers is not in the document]
siggisim: This used to be part of a longer method that had two of these. Done.

enterprise/server/githubapp/githubapp.go
siggisim: Done

enterprise/server/githubapp/githubapp.go
```diff
@@ -1342,3 +1344,142 @@ func (a *GitHubApp) CreateGithubRef(ctx context.Context, req *ghpb.CreateGithubR
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	return incoming, outgoing, user, nil
+}
+
+func (a *GitHubApp) populatePRMetadata(ctx context.Context, client *github.Client, prIssues []*github.Issue, currentUser *github.User) (map[int]*ghpb.PullRequest, error) {
+	prsMu := &sync.Mutex{}
+	prs := make(map[int]*ghpb.PullRequest, len(prIssues))
+	metadataGroup, metadataContext := errgroup.WithContext(ctx)
```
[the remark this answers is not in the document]
siggisim: Done

enterprise/server/githubapp/githubapp.go
siggisim: I thought about it, but was like "github is never going to give me a bad URL!" lol.

You're right though, better safe than sorry - done.

enterprise/server/githubapp/githubapp.go
siggisim: Yeah, I thought about having individual mutexes - but felt it was easier to just guard it all with one since there should be max ~200 of these.

proto/github.proto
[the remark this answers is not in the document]
siggisim: hahaha I love map https://www.youtube.com/watch?v=Ld35lMnFQ10

enterprise/server/githubapp/githubapp.go
```diff
@@ -1342,3 +1344,142 @@ func (a *GitHubApp) CreateGithubRef(ctx context.Context, req *ghpb.CreateGithubR
...
+
+func issueToPullRequestProto(i *github.Issue) *ghpb.PullRequest {
+	return &ghpb.PullRequest{
+		Number:        uint64(i.GetNumber()),
+		Title:         i.GetTitle(),
+		Body:          i.GetBody(),
+		Author:        i.GetUser().GetLogin(),
+		Url:           i.GetHTMLURL(),
+		UpdatedAtUsec: i.GetUpdatedAt().UnixMicro(),
+		Reviews:       map[string]*ghpb.Review{},
```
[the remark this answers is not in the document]
siggisim: I wasn't sure (I'm still not) if the proto library would initialize this for me lol - wanted to make sure it exists before I start putting adding stuff to it (which I do all over the `populatePRMetadata` function).

PR #5602, opened 2023-12-20
Improve chunk prioritization for eager fetching

bduffany:
My hypothesis that motivated this PR was that when the `eagerFetchChan` is saturated we'll spend less time fetching newer/more important chunks and more time fetching older/less important chunks.

This PR changes the data structure for the eager fetch list from a buffered `chan` to a new `BoundedStack` data structure, so that new chunks can always be added to the list of chunks to be eagerly fetched, evicting the oldest chunk when the stack's capacity is exceeded. Under the hood, the `BoundedStack` is backed by a ring buffer, and it also exposes a channel that receives a value whenever an item is pushed (so that the eager fetch goroutines can be notified whenever it's safe to `pop`).

Did some testing, comparing `master` to this PR branch (`eager-fetch-prio`). Alternated between the two branches, clearing filecache then running vmstart to load a remote snapshot (from prod), and appended the `LoadSnapshot` duration to a file. Then generated a summary of the durations in the file.

Results (the stack-based approach reduces LoadSnapshot duration by about 25%):

```
Durations summary for eager-fetch-prio.log (n=10):
Mean: 10.915s
0.25 quartile: 10.294s
0.50 quartile (median): 10.684s
0.75 quartile: 11.424s

Durations summary for master.log (n=10):
Mean: 13.774s
0.25 quartile: 13.476s
0.50 quartile (median): 13.973s
0.75 quartile: 14.258s
```

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -37,10 +37,10 @@ const (
 	// Suffix used for dirtied chunks.
 	dirtySuffix = ".dirty"
 
-	// We don't want this to be too big, because we want to prioritize fetching
-	// more recently accessed data. If the chan is too big, we may waste time
-	// churning through stale data
-	eagerFetchChanSize = 100
+	// The capacity of the buffer of chunks that are waiting to be eagerly
+	// fetched. We don't want this to be too big, because otherwise we'd waste
+	// resources fetching chunks that are unlikely to be used.
+	eagerFetchBufferCapacity = 100
 
 	// If we access a chunk, we will queue this number of contiguous chunks
 	// to be eagerly fetched in the background, anticipating that they will
@@ -122,9 +122,12 @@ type COWStore struct {
 	// Concurrency-safe pool of buffers that can be used for copying chunks
 	copyBufPool sync.Pool
 
-	eagerFetchChan chan *eagerFetchData
-	eagerFetchEg   *errgroup.Group
-	quitChan       chan struct{}
+	// Chunk offsets to be eagerly fetched. This is using a bounded stack data
+	// structure so that if the buffer is full, a new item can still be appended,
+	// evicting the least recently added chunk.
+	eagerFetchStack *BoundedStack[int64]
+	eagerFetchEg    *errgroup.Group
+	quitChan        chan struct{}
 }
 
 // NewCOWStore creates a COWStore from the given chunks. The chunks should be
@@ -139,7 +142,10 @@ func NewCOWStore(ctx context.Context, env environment.Env, chunks []*Mmap, chunk
 	for _, c := range chunks {
 		chunkMap[c.Offset] = c
 	}
-
+	eagerFetchStack, err := NewBoundedStack[int64](eagerFetchBufferCapacity)
+	if err != nil {
+		return nil, err
+	}
 	s := &COWStore{
 		ctx:                ctx,
 		env:                env,
@@ -155,13 +161,13 @@ func NewCOWStore(ctx context.Context, env environment.Env, chunks []*Mmap, chunk
 				return &b
 			},
 		},
-		zeroBuf:        make([]byte, chunkSizeBytes),
-		chunkSizeBytes: chunkSizeBytes,
-		totalSizeBytes: totalSizeBytes,
-		ioBlockSize:    int64(stat.Sys().(*syscall.Stat_t).Blksize),
-		eagerFetchChan: make(chan *eagerFetchData, eagerFetchChanSize),
-		eagerFetchEg:   &errgroup.Group{},
-		quitChan:       make(chan struct{}),
+		zeroBuf:         make([]byte, chunkSizeBytes),
+		chunkSizeBytes:  chunkSizeBytes,
+		totalSizeBytes:  totalSizeBytes,
+		ioBlockSize:     int64(stat.Sys().(*syscall.Stat_t).Blksize),
+		eagerFetchStack: eagerFetchStack,
+		eagerFetchEg:    &errgroup.Group{},
+		quitChan:        make(chan struct{}),
 	}
 
 	s.eagerFetchEg.Go(func() error {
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -1355,3 +1353,93 @@ func checkBounds(opName string, storeSize int64, p []byte, off int64) error {
...
+	top    int
+	size   int
+}
+
+func NewBoundedStack[T any](capacity int) (*BoundedStack[T], error) {
+	if capacity <= 0 {
+		return nil, status.InvalidArgumentError("bounded stack capacity must be > 0")
+	}
+	return &BoundedStack[T]{
+		C:      make(chan struct{}, 1),
```
maggie-lou, line 1390: Is the capacity supposed to be capacity and not 1?

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -1355,3 +1353,93 @@ func checkBounds(opName string, storeSize int64, p []byte, off int64) error {
...
+// Pop returns the top element of the stack. The second return value indicates
+// whether the returned item is valid; it will be false if and only if the stack
+// was empty.
+func (s *BoundedStack[T]) Pop() (item T, ok bool) {
+	val, newSize, ok := s.pop()
+	if newSize > 0 {
+		// Re-saturate the channel if the stack still has items after popping,
+		// so that another goroutine can pop the next element.
+		select {
+		case s.C <- struct{}{}:
```
maggie-lou, line 1423: I'm kind of confused about what this is supposed to do. Are we not double adding to the channel for each element in the stack? We already added to the channel in Push()

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -1355,3 +1353,93 @@ func checkBounds(opName string, storeSize int64, p []byte, off int64) error {
 	}
 	return nil
 }
+
+// BoundedStack is a concurrency-safe stack with bounded size: when the
```
tylerwilliams, line 1357: wanna throw this in util/boundedstack

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: It should work as-is, because we replenish the buffer in Pop() whenever it still has items after popping. Setting it to `capacity` lets us remove that replenishing logic though so I removed it. This has a bit of extra memory cost but should be fine for now.

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: (See other comment)

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: done

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: Actually NVM, `struct{}` consumes 0 bytes of memory so this is just straight up better.

server/util/boundedstack/boundedstack.go
```diff
@@ -0,0 +1,100 @@
...
+	defer s.mu.Unlock()
+	s.top = (s.top + 1) % len(s.buffer)
+	s.buffer[s.top] = item
+	s.size = min(s.size+1, len(s.buffer))
+}
+
+// Pop returns the top element of the stack. The second return value indicates
+// whether the returned item is valid; it will be false if and only if the stack
+// was empty.
+func (c *BoundedStack[T]) Pop() (item T, ok bool) {
```
maggie-lou, line 83: Should we make the channel a private implementation detail of pop? That way clients don't have to worry about `IMPORTANT: after receiving a message from C, make sure to always call Pop.
// Otherwise, the stack item corresponding to the notification on C may be
// dropped unnecessarily.`

Pop could be a blocking method

tylerwilliams: +1 the API is kind of error prone as written

bduffany: I considered this but didn't want to make `Pop()` a blocking method because otherwise goroutines could get stuck indefinitely without some other cancellation / closing mechanism which also felt error prone.

I agree that the channel approach is not ideal though so I added a `Recv(ctx)` method to allow cancellation if the context is canceled, and made the channel private. wdyt? (Went with the name "recv" since it's analogous to a receive on a channel, and because we might want to reserve the name "pop" for the non-blocking pop() operation in case it makes sense to have that be public later).

maggie-lou: Should we make `Pop` private? It seems incorrect that `Push` will always add an element to the channel, but `Pop` won't remove one. You could just add a large comment that `Recv` is a blocking method?

bduffany: Made it private. I was using it for tests to check that Pop() returns nothing when the stack is empty, and it would have been awkward to do that via Recv instead.

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -615,51 +622,45 @@ func (s *COWStore) initDirtyChunk(offset int64, size int64) (ogChunk *Mmap, newC
...
 	eg := &errgroup.Group{}
 	eg.SetLimit(*eagerFetchConcurrency)
 	defer eg.Wait()
 
+	ctx, cancel := context.WithCancel(s.ctx)
+	defer cancel()
+	go func() {
+		<-s.quitChan
+		cancel()
+	}()
```
tylerwilliams, line 644: is s.ctx cancelled when quitChan is called? Seems like if it were you would no longer need this. (and maybe not even need quitChan either since you could instead wait for <-s.ctx.Done() instead)

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -615,51 +622,45 @@ func (s *COWStore) initDirtyChunk(offset int64, size int64) (ogChunk *Mmap, newC
...
-				if err != nil {
-					log.CtxWarningf(s.ctx, "COWStore eager fetch chunk failed with: %s", err)
-				}
-				return nil
-			})
+			return
 		}
+		eg.Go(func() error {
+			err := s.fetchChunk(offset)
+			if err != nil {
```
tylerwilliams, line 659: can scope this err down:

if err := s.fetchChunk(offset); err != nil {
 log da err
}

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: We don't cancel `s.ctx` in the place where `quitChan` is called. I'd prefer not to change that right now because there may be some unintended consequences there that I'm not seeing.

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
bduffany: done

maggie-lou (approved):

PR #5611, opened 2023-12-21
use server/util/proto in unit tests

luluz66:
**Related issues**: N/A


server/remote_cache/scorecard/scorecard_test.go
```diff
@@ -284,7 +284,7 @@ type fakeBlobStore struct {
 func (bs *fakeBlobStore) ReadBlob(ctx context.Context, name string) ([]byte, error) {
 	tokens := strings.Split(name, "/")
 	if attempt := tokens[1]; attempt == "1" {
-		return proto.Marshal(bs.ScoreCard)
```
jdhollen, line 287: ahahaha please tell me this doesn't actually matter

jdhollen (approved):

server/remote_cache/scorecard/scorecard_test.go
luluz66: It's test no big deal. but since i already know ScoreCard is an anomaly; so..

PR #5616, opened 2023-12-21
mac-release: fix Bazel 7 compatibility

sluongng:
It seems like the older versions are not compatible with Bazel 7.

Use of `XCODE_VERSION` instead of `DEVELOPER_DIR` for accurate
autoxcode_config result.

### DIFF
--- a/.github/workflows/release-mac.yaml
+++ b/.github/workflows/release-mac.yaml
@@ -54,9 +54,8 @@ jobs:
       - name: Build and Upload Artifacts
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          USE_BAZEL_VERSION: 6.4.0
+          XCODE_VERSION: 12.4
         run: |
-          export DEVELOPER_DIR=/Applications/Xcode_12.4.app/Contents/Developer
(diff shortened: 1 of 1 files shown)


bduffany (approved):

.github/workflows/release-mac.yaml
```diff
@@ -15,34 +15,37 @@ on:
...
-            ref="${GITHUB_REF}"
-            tag=${GITHUB_REF/refs\/tags\//}
-          else
-            # Otherwise the workflow must have been manually triggered with inputs
-            ref="${{ inputs.release_branch }}"
-            tag="${{ inputs.version_tag }}"
-          fi
-          echo "ref=$ref" >> $GITHUB_OUTPUT
-          echo "tag=$tag" >> $GITHUB_OUTPUT
+      # - name: Configure variables
```
vadimberezniker, line 27: are all these lines commented out for testing?

sluongng: yup, I am testing this PR again because I realize that this could be fixed by simply `s/DEVERLOPER_DIR/XCODE_VERSION/`.

If that's the case, we could defer the version upgrade to a different PR down the line.

sluongng: Fixed

vadimberezniker (approved):

PR #5622, opened 2023-12-21
Start sending lease_id as reconnect token.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/task_leaser/task_leaser.go
+++ b/enterprise/server/scheduling/task_leaser/task_leaser.go
@@ -29,17 +29,17 @@ const (
 )
 
 type TaskLeaser struct {
-	env            environment.Env
-	executorID     string
-	taskID         string
-	leaseID        string
-	reconnectToken string
-	quit           chan struct{}
-	mu             sync.Mutex // protects stream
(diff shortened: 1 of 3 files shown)


enterprise/server/scheduling/task_leaser/task_leaser.go
```diff
@@ -67,7 +67,9 @@ func (t *TaskLeaser) pingServer(ctx context.Context) (b []byte, err error) {
 		ExecutorId:        t.executorID,
 		TaskId:            t.taskID,
 		SupportsReconnect: *enableReconnect,
-		ReconnectToken:    t.reconnectToken,
```
bduffany, line 70: hm was deleting this line intentional? Seems like if `supportsReconnect` is false then we should keep setting it the old way? or if we don't care about rollouts, should we just instead unconditionally set this to lease ID and remove the `supportsReconnect` field?

vadimberezniker: `supportsReconnect` is true if either the new `supports_reconnect` field is set OR the old `reconnect_token` field is set which means the ReconnectToken should be populated in all cases where it was previously populated.

bduffany (approved):

PR #5623, opened 2023-12-21
Fix release workflows

maggie-lou:
I realized that GITHUB_REF is always set, even if the workflow was not triggered by a push event.

### DIFF
--- a/.github/workflows/release-m1.yaml
+++ b/.github/workflows/release-m1.yaml
@@ -36,7 +36,7 @@ jobs:
         id: vars
         run: |
           # Configure repo based on pushed tag if set
-          if [ -z "${GITHUB_REF}" ]; then
+          if [ ${{ github.event_name }} = 'push' ]; then
             ref="${GITHUB_REF}"
             tag=${GITHUB_REF/refs\/tags\//}
           else
--- a/.github/workflows/release-mac.yaml
+++ b/.github/workflows/release-mac.yaml
@@ -25,7 +25,7 @@ jobs:
         id: vars
         run: |
           # Configure repo based on pushed tag if set
-          if [ -z "${GITHUB_REF}" ]; then
+          if [ ${{ github.event_name }} = 'push' ]; then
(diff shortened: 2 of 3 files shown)


maggie-lou:
* `master`
  * **PR #5623** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5623?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5623?utm_source=stack-comment).

.github/workflows/release-m1.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
         id: vars
         run: |
           # Configure repo based on pushed tag if set
-          if [ -z "${GITHUB_REF}" ]; then
+          if [ ${{ github.event_name }} = 'push' ]; then
```
jdhollen, line 39: the docs are a little bit confusing for this stuff (to me, at least)--does event_name = 'push' actually guarantee that GITHUB_REF is set?

https://docs.github.com/en/actions/learn-github-actions/variables says `This is only set if a branch or tag is available for the event type.`

maggie-lou: The docs are so shitty 🥲 Turns out GITHUB_REF is set no matter what. I tested by both pushing some tags and manually starting the workflow and it was always set

jdhollen: fun fun. up to you based on your confidence level in that guarantee, but you could always check both.

jdhollen (approved):

.github/workflows/release-m1.yaml
maggie-lou: Want to make sure the workflow is fixed before I go on vacation so we can add that in the future if we need it

jdhollen (approved):

PR #5624, opened 2023-12-22
Add a unit test for podman.go

iain-macdonald:
**Related issues**: N/A


iain-macdonald:
@bduffany lemme know if you want me to separate the change that moves `Stdio` to interfaces.go to make this change smaller. And I'm open to feedback on how to deal with the `podman version` call.

enterprise/server/remote_execution/containers/podman/podman_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/buildbuddy-io/buildbuddy/server/testutil/testfs"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	_ "github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/containers/docker"
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+)
+
+type SlowCommandRunner struct {
+	commandsRun *atomic.Int32
```
bduffany, line 23: nit: this can be an `atomic.Int32` (non-pointer) so you don't have to construct it manually. The test can directly check `runner.commandsRun`

server/interfaces/interfaces.go
```diff
@@ -956,6 +956,21 @@ type RunnerPool interface {
 	GetBuildRoot() string
 }
 
+// CommandRunner is an interface for running command-line commands.
+type CommandRunner interface {
+	Run(ctx context.Context, command *repb.Command, workDir string, statsListener func(*repb.UsageStats), stdio *Stdio) *CommandResult
+}
```
bduffany, line 962: This naming is a bit unfortunate since we already have something called `commandRunner` in `runner.go`. We could either rename the existing `runner` to something better or call this something like `ProcessRunner` (since ultimately `Run()` here is doing low-level syscalls to create a unix/windows process and get its result, whereas the existing runner is doing higher-level stuff like managing workspaces and containers for a command).

enterprise/server/remote_execution/containers/podman/podman_test.go
```diff
@@ -0,0 +1,103 @@
...
+	props := platform.Properties{
+		ContainerImage: "docker.io/library/busybox",
+		DockerNetwork:  "off",
+	}
+	container, err := provider.New(ctx, &props, nil, nil, "")
+	require.NoError(t, err)
+
+	for i := 0; i < 5; i++ {
+		go require.NoError(t, container.PullImage(ctx, oci.Credentials{}))
+	}
```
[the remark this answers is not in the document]
iain-macdonald: Oof, I knew that, my bad!

enterprise/server/remote_execution/containers/podman/podman_test.go
iain-macdonald: That leads to a copying lock error on line 40 where the command runner is copied in the environment. I could make that a pointer, but I'd rather leave the pointer messiness in the test if that's alright with you.

server/interfaces/interfaces.go
iain-macdonald: Yeah, I saw that, I think the most correct thing to do would be to rename the `runner.go` version to `taskRunner`, as that's more consistent with the language in there and in the remote execution API. WDYT?

enterprise/server/remote_execution/commandutil/commandutil.go
```diff
@@ -116,14 +106,20 @@ func RetryIfTextFileBusy(fn func() error) error {
 	}
 }
 
+type RealCommandRunner struct{}
+
+func (_ RealCommandRunner) Run(ctx context.Context, command *repb.Command, workDir string, statsListener func(*repb.UsageStats), stdio *interfaces.Stdio) *interfaces.CommandResult {
+	return Run(ctx, command, workDir, statsListener, stdio)
+}
```
bduffany, line 113: woops I meant to mention this in the last round - we can call this `Runner` since "Real" is implied by this not being a testutil pkg and "Command" is already in the package name (commandutil)

server/interfaces/interfaces.go
bduffany: `taskRunner` SGTM

enterprise/server/remote_execution/containers/podman/podman_test.go
bduffany: You can avoid that error by changing the function signature on line 27 to be a pointer receiver

```
func (r *SlowCommandRunner) ...
```

then on line 40, declare SlowCommandRunner using a pointer:

```
commandRunner := &SlowCommandRunner{}
```

To me this is not just a workaround, but actually the most natural way to do it - for "business logic" structs which have methods and implement interfaces, I've found it's almost always desired to pass by reference and not by value. Otherwise, if you want pass-by-reference semantics in your method calls, then you have to declare all fields as pointers, which is a lot more tedious/complex than just declaring the struct itself as a pointer.

The only times I've found it makes sense to pass structs by value is for very small "tuple"-like structures consisting only of primitives, such as username/password combos, or options structs with a few int/bool config options.

iain-macdonald: Ahh, I didn't realize you could mix-and-match pointer versus non-pointer receivers for different implementations of an interface and was a bit hesitant to make the real instance receive a pointer (though I guess that doesn't really matter either), but what you suggested works so I changed it to that.

enterprise/server/remote_execution/commandutil/commandutil.go
iain-macdonald: Done

bduffany (approved):

PR #5629, opened 2023-12-22
Prevent bazel from generating MODULE.bazel file

bduffany:
Bazel keeps trying to generate `MODULE.bazel` and `MODULE.bazel.lock` when I run certain commands. This seems to fix it.

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,4 +1,4 @@
-build --noenable_bzlmod
+common --noenable_bzlmod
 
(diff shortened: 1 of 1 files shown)


sluongng (approved):
Makes sense. I avoided using "common" because BB CLI was not picking up the correct flags with it. Not sure if the parser is working now.

bduffany:
That issue should be fixed in v5.0.25 of the CLI

PR #5633, opened 2024-01-02
Add `ResourceName.address` field

bduffany:
This will be helpful for a UI refactor that aims to reduce the number of places where we manually construct bytestream:// and actioncache:// URLs (https://github.com/buildbuddy-io/buildbuddy/pull/5634)

I found myself writing a function in JS like

```
function bytestreamURL(address: string, resourceName: ResourceName): string { ... }
```

but I realized that it makes more sense for the `address` field to just be another field of `ResourceName`. A raw (string) resource name roughly follows this schema:

```
bytestream://{address}/{instance_name}/compressed-blobs/{compressor}/{digest_function}/{digest}
```

so it makes sense that `address` should be part of the structured form of this resource name string.

**Related issues**: N/A

### DIFF
--- a/proto/resource.proto
+++ b/proto/resource.proto
@@ -33,6 +33,12 @@ message ResourceName {
   // that this is only correct for the following digest types:
   // MD5, MURMUR3, SHA1, SHA256, SHA384, SHA512, VSO.
   build.bazel.remote.execution.v2.DigestFunction.Value digest_function = 5;
+
+  // The hostname or hostname:port of the cache service. If not present, the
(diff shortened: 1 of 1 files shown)


vadimberezniker:
Please get LGTM from @tylerwilliams as well.

vadimberezniker (approved):

bduffany:
Discussed with Tyler offline - adding a field to ResourceName that we're not necessarily setting everywhere could be confusing, so we'll avoid this for now and have the UI pass around the cache address explicitly instead.

PR #5634, opened 2024-01-02
frontend: tidy up CAS/AC URL code

bduffany:
* Add a new cache util package and use that for `bytestream://` URL construction instead of creating them manually everywhere.
* Tidy up a few other pieces of code in the execution card.

**Related issues**: N/A


siggisim (approved):

app/invocation/invocation_action_card.tsx
```diff
@@ -90,103 +93,65 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+        this.setState({ inputRoot, inputDirs });
       })
       .catch((e) => console.error("Failed to fetch input root:", e));
   }
 
   fetchActionResult() {
-    let digestParam = this.props.search.get("actionResultDigest");
-    if (digestParam == null) {
-      digestParam = this.props.search.get("actionDigest");
+    let digestParam = this.props.search.get("actionResultDigest") ?? this.props.search.get("actionDigest");
```
jdhollen, line 111: Should be fine, but just a reminder that this would short-circuit for `""` instead of checking `actionDigest`

app/invocation/invocation_action_card.tsx
```diff
@@ -315,14 +276,10 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
-    let digestString = node.obj.digest?.hash + "/" + node.obj.digest?.sizeBytes;
-    let dirUrl =
-      "bytestream://" +
-      this.props.model.getCacheAddress() +
-      "/blobs/" +
-      this.props.model.getDigestFunctionDir() +
-      digestString;
+    if (!node.obj?.digest) return;
 
+    let dirUrl = this.props.model.getBytestreamURL(node.obj.digest);
```
jdhollen, line 281: super nit: `const`?

app/invocation/invocation_model.tsx
```diff
@@ -388,23 +390,60 @@ export default class InvocationModel {
...
+      return build.bazel.remote.execution.v2.Compressor.Value.ZSTD;
+    }
+    return build.bazel.remote.execution.v2.Compressor.Value.IDENTITY;
+  }
+
   getDigestFunction() {
     const digestFnName = this.optionsMap.get("digest_function");
     if (!digestFnName) {
-      return undefined;
+      return build.bazel.remote.execution.v2.DigestFunction.Value.SHA256;
```
jdhollen, line 403: this seems error-prone, since we used to let the server infer it from digest length?  is it actually guaranteed that the option will be set for old-style digest functions?

jdhollen (approved):

app/invocation/invocation_action_card.tsx
bduffany: I usually use `const` for new code but since the benefits are pretty minor I prefer to stick with whatever the current code is doing to avoid adding extra noise to the diff.

app/invocation/invocation_action_card.tsx
bduffany: ack, I think it's fine as-is since it's probably a bug if the URL contains `actionDigest` set to an empty string

app/invocation/invocation_model.tsx
jdhollen: Everything you said is right--I was just wondering aloud if it's ever possible for bazel to have used SHA1 (for example) without the flag being set this way.  It's true that this has to work for blake3 URLs to work properly (i.e., if there even is a bug here, it already exists for blake3), so this is fine.

PR #5636, opened 2024-01-02
Improve lifecycle management of test redis servers

bduffany:
- If Redis crashes, proactively fail the test.
- After the current test case completes, kill the Redis server. Otherwise a bunch of Redis servers will be kept in memory until the test process exits.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/testutil/testredis/testredis.go
+++ b/enterprise/server/testutil/testredis/testredis.go
@@ -83,25 +83,33 @@ func (h *Handle) start() {
 	if err != nil {
 		assert.FailNowf(h.t, "redis binary could not be started", err.Error())
 	}
+	h.t.Cleanup(func() {
+		h.Shutdown()
+	})
 	go func() {
-		if err := cmd.Wait(); err != nil {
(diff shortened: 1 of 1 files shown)


enterprise/server/testutil/testredis/testredis.go
```diff
@@ -83,27 +83,34 @@ func (h *Handle) start() {
...
-			default:
-				log.Warningf("redis server did not exit cleanly: %v", err)
-			}
+		err := cmd.Wait()
+		close(h.done)
+		if ctx.Err() != nil {
+			// Context cancelled; test was cleaned up normally.
+			return
 		}
+		h.t.Fatalf("redis server exited unexpectedly: %s", err)
```
iain-macdonald, line 96: Flip the conditional above and put this line inside?

iain-macdonald (approved):

enterprise/server/testutil/testredis/testredis.go
bduffany: Done

PR #5637, opened 2024-01-03
Ignore MODULE.bazel because bazel 7 is auto-generating one

jdhollen:
**Related issues**: N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,10 @@ tramp
 # Bazel directories
 /bazel-*
 
+# MODULE.bazel stuff, for now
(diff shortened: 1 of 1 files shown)


siggisim (approved):

sluongng (approved):
I think c4752e8ff7 should have effectively disabled bzlmod for all Bazel commands. Before that, these files would get generated if you were running a `bazel [|a|c]query`. So I would recommend rebasing past this commit if you are seeing these files generated.

Nevertheless, there is no harm in ignoring these.

PR #5642, opened 2024-01-04
stop checking workflow_history_enabled, it's launched

jdhollen:
**Related issues**: N/A


iain-macdonald (approved):

server/static/static.go
```diff
@@ -49,7 +49,6 @@ var (
 	customerManagedEncryptionKeysEnabled   = flag.Bool("app.customer_managed_encryption_keys_enabled", false, "If set, show customer-managed encryption configuration UI.")
 	tagsUIEnabled                          = flag.Bool("app.tags_ui_enabled", false, "If set, expose tags data and let users filter by tag.")
 	timeseriesChartsInTimingProfileEnabled = flag.Bool("app.timeseries_charts_in_timing_profile_enabled", true, "If set, charts with sampled time series data (such as CPU and memory usage) will be shown")
-	workflowHistoryEnabled                 = flag.Bool("app.workflow_history_enabled", true, "If set, information about past action runs for workflows will be shown on the workflows page.")
```
sluongng, line 52: I would recommend marking the flag as deprecated instead, just in case some on-prem customer have set it in their deployment.

jdhollen: For those watching along at home, we decided to just go for it because it's reaaaally unlikely anyone ever set this flag outside of yaml.

PR #5649, opened 2024-01-04
Add userdb method to lookup user by e-mail.

vadimberezniker:
This will be used by the SCIM API to lookup an existing user corresponding to a user in the external system.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/userdb/userdb_test.go
+++ b/enterprise/server/backends/userdb/userdb_test.go
@@ -197,6 +197,35 @@ func TestInsertUser(t *testing.T) {
 	}
 }
 
+func TestGetUserByEmail(t *testing.T) {
+	env := newTestEnv(t)
+	udb := env.GetUserDB()
+	ctx := context.Background()
+
+	err := udb.InsertUser(ctx, &tables.User{
+		UserID: "US1",
+		SubID:  "SubID1",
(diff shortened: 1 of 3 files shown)


siggisim (approved):

enterprise/server/backends/userdb/userdb_test.go
```diff
@@ -197,6 +197,35 @@ func TestInsertUser(t *testing.T) {
...
+	require.NoError(t, err, "inserting multiple users should succeed")
+
+	userCtx := authUserCtx(ctx, env, t, "US1")
+	u, err := udb.GetUserByEmail(userCtx, "user1@org1.io")
+	require.NoError(t, err)
+	require.Equal(t, "US1", u.UserID)
+
+	// Should not be able to look up user in a different group.
+	_, err = udb.GetUserByEmail(userCtx, "user2@org2.io")
+	require.True(t, status.IsNotFoundError(err))
```
bduffany, line 226: nit: test "more than 1 user w/ a given email => error" ?

bduffany (approved):

enterprise/server/backends/userdb/userdb_test.go
vadimberezniker: done

PR #123, opened 2020-10-28
Add request/response protos for DeleteInvocation

bduffany:

PR #5654, opened 2024-01-05
a code review ui that may look familiar to some

jdhollen:
I don't use graphite, so I'm just blasting this out as a change that includes the server changes in #5653

I'll merge that first and bring this one up to it

It looks like this:
Screenshot 2024-01-04 at 7 24 10 PM

### DIFF
--- a/enterprise/app/review/view_pull_request.tsx
+++ b/enterprise/app/review/view_pull_request.tsx
@@ -1,0 +1,390 @@
+import Long from "long";
+import React, { ReactNode } from "react";
+import { User } from "../../../app/auth/user";
+import format from "../../../app/format/format";
+import rpc_service from "../../../app/service/rpc_service";
+import { github } from "../../../proto/github_ts_proto";
+import { Github, MessageCircle } from "lucide-react";
+
+interface ViewPullRequestComponentProps {
+  user?: User;
+  owner: string;
+  repo: string;
+  pull: number;
+  path: string;
+}
+interface State {
+  response?: github.GetGithubPullRequestDetailsResponse;
+  displayedDiffs: string[];
+}
+
+interface DiffLineInfo {
+  startLine: number;
+  lineCount: number;
+}
+
+type SourceLine = string | undefined;
+interface DiffLinePair {
+  left: SourceLine;
+  right: SourceLine;
+}
+
+interface Hunk {
+  header: string;
+  lines: DiffLinePair[];
+}
+
+export default class ViewPullRequestComponent extends React.Component<ViewPullRequestComponentProps, State> {
+  state: State = {
+    displayedDiffs: [],
+  };
+
+  componentWillMount() {
+    rpc_service.service
+      .getGithubPullRequestDetails({
+        owner: this.props.owner,
+        repo: this.props.repo,
+        pull: Long.fromInt(this.props.pull),
+      })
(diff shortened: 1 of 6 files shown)


enterprise/app/review/view_pull_request.tsx
```diff
@@ -0,0 +1,390 @@
...
+    rpc_service.service
+      .getGithubPullRequestDetails({
+        owner: this.props.owner,
+        repo: this.props.repo,
+        pull: Long.fromInt(this.props.pull),
+      })
+      .then((r) => {
+        console.log(r);
+        this.setState({ response: r });
+      });
```
bduffany, line 52: `.catch(e => error_service.handleError(e))`

bduffany (commented):
Looks awesome! :star_struck: 

A couple of UI nits (doesn't have to be in this PR)

- Maybe render `Change 123` as `Pull request #123` in the title bar
- Maybe render the pending status checks with a blue ("in progress") BG to differentiate them from the completed status checks that are grey? Or make the completed ones green so that it matches our UI status color when clicking through to the workflow invocation
- Render the repo name somewhere?

siggisim:
> Maybe render Change 123 as Pull request #123 in the title bar
I kind of like using the "Change" terminology. I've always hated "Pull request". If you're a beginner to programming - you know what a change is, but you have no idea what a pull request is.

bduffany:
I agree that "pull request" sucks and prefer Gitlab's "Merge request" terminology.
 "Change" is especially confusing in git terms, because it can refer to either a commit, or a request to merge a commit. But given the context of getting here from the "reviews" page it's probably fine.

I do think the "#" would be good to add though because that's usually how it's presented in other UIs

jdhollen:
> Looks awesome! 🤩
> 
> A couple of UI nits (doesn't have to be in this PR)
> 
> * Maybe render `Change 123` as `Pull request #123` in the title bar
> * Maybe render the pending status checks with a blue ("in progress") BG to differentiate them from the completed status checks that are grey? and make the completed ones green so that it matches our UI status color when clicking through to the workflow invocation. Currently we use grey for "disconnected" so it feels a bit confusing (I understand the "familiar" UI doesn't use these conventions :P)
> * Render the repo name somewhere?

There will be many more changes to all of these, so I'm going to punt on the bigger tweaks, but i've updated the text in a few places per your comments (and added a doc title, which i forgot to do)

jdhollen:
> I agree that "pull request" sucks and prefer Gitlab's "Merge request" terminology. "Change" is ambiguous in git terms, because it can refer to either a commit, or a request to merge a commit. But given the context of getting here from the "reviews" page it's probably fine.
> 
> I do think the "#" would be good to add though because that's usually how it's presented in other UIs

gonna let this relgious battle play out a little longer, we've currently got both on the homepage:
Screenshot 2024-01-05 at 10 22 53 AM

enterprise/app/review/view_pull_request.tsx
jdhollen: Oops, thanks

bduffany:
>  we've currently got both on the homepage:

The "PR" bit was my doing :sweat_smile: I shortened it because "Change" was taking up too much space, but maybe "#" or "ID" could also work there?

bduffany (approved):

PR #5660, opened 2024-01-05
Render workflow artifacts in the UI

bduffany:
Render any artifacts which the user wrote to `BUILDBUDDY_ARTIFACTS_DIRECTORY`, such as the gRPC log.

Also fix a small issue that we weren't setting the digest/length metadata on the file, so the digest component appeared blank.

image

**Related issues**: N/A


siggisim (approved):

tylerwilliams:
neat :P

PR #5655, opened 2024-01-05
Upgrade rules_go and gazelle

sluongng:
https://github.com/bazelbuild/rules_go/releases/tag/v0.44.2
https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.35.0

PR #5662, opened 2024-01-08
proto: handle go proto library deprecation

sluongng:
In https://github.com/bazelbuild/rules_go/pull/3761, rules_go started deprecating away
the usage of https://github.com/golang/protobuf in favor of a combination of
https://github.com/protocolbuffers/protobuf-go and https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc.

This results in `go_proto_library` requiring these 2 new compilers.
As an convinience, a `go_grpc_library` macro was introduced to wrap around `go_proto_library`
to ensure that these 2 compilers are set by default.

Effectively, in our `go_proto_library` targets, switch our usage of the old compiler to the new pair of compilers.
We are not using `go_grpc_library` yet because Gazelle has yet to support it. https://github.com/bazelbuild/bazel-gazelle/pull/1711

**Related issues**: N/A

### DIFF
--- a/proto/BUILD
+++ b/proto/BUILD
@@ -772,7 +772,8 @@ go_proto_library(
 go_proto_library(
     name = "health_go_proto",
     compilers = [
-        "@io_bazel_rules_go//proto:go_grpc",
+        "@io_bazel_rules_go//proto:go_proto",
+        "@io_bazel_rules_go//proto:go_grpc_v2",
         "//proto:vtprotobuf_compiler",
     ],
     importpath = "github.com/buildbuddy-io/buildbuddy/proto/health",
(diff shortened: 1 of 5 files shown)


sluongng:
This PR is based on https://github.com/buildbuddy-io/buildbuddy/pull/5655 and should only be merged after.

proto/BUILD
```diff
@@ -1494,8 +1512,8 @@ go_proto_library(
 go_proto_library(
     name = "worker_go_proto",
     compilers = [
```
luluz66, line 1514: vtprotobuf_compiler was acidentally deleted.

sluongng: Good call. I pushed a fix.

luluz66 (approved):

PR #5664, opened 2024-01-08
Fix workflow action result not being displayed in action page

bduffany:
* Workflow actions don't publish a `canonical` command line event. Update the logic from https://github.com/buildbuddy-io/buildbuddy/pull/5362 to prefer the canonical command line if present but fall back to the original command line otherwise. 
  * We could alternatively update workflows to send a `canonical` command line rather than an `original` one, but this approach feels more forgiving to BES clients in general.
* Fix `getCacheAddress()` incorrectly returning `address + instanceName` which results in bytestream URLs containing the instance name twice, like `grpc://remote.buildbuddy.io/instance-name/instance-name/...`

**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -215,10 +215,9 @@ export default class InvocationModel {
     for (let log of this.buildToolLogs?.log || []) {
       this.toolLogMap.set(log.name, new TextDecoder().decode(log.contents || new Uint8Array()));
     }
-    for (let commandLine of this.structuredCommandLine) {
-      if (commandLine.commandLineLabel != "canonical") {
-        continue;
-      }
+    let commandLine =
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_model.tsx
```diff
@@ -215,10 +215,9 @@ export default class InvocationModel {
     for (let log of this.buildToolLogs?.log || []) {
       this.toolLogMap.set(log.name, new TextDecoder().decode(log.contents || new Uint8Array()));
     }
-    for (let commandLine of this.structuredCommandLine) {
-      if (commandLine.commandLineLabel != "canonical") {
-        continue;
-      }
+    let commandLine =
```
sluongng, line 218: Nit: I would appreciate it if there is a small comment explaining why we fall back to "original" and first event in the case of BuildBuddy Workflow.

sluongng (approved):
I don't think there is a case where Bazel would not send canonical event, so this logic would only be applied to our Workflow event stream.

Make sense to me.

PR #5665, opened 2024-01-08
Add development script for local clickhouse

bduffany:
Add a script that starts a clickhouse server using docker, then prints the flags needed to connect to it.

Example usage: `bazel run enterprise/server -- $(tools/clickhouse)`

**Related issues**: N/A

### DIFF
--- a/tools/clickhouse
+++ b/tools/clickhouse
@@ -1,0 +1,60 @@
+#!/bin/bash
+set -euo pipefail
+
+usage='
+tools/clickhouse: start a clickhouse local server and print the app flags needed to connect to it
+
+Example:
+  $ bazel run enterprise/server -- $(tools/clickhouse)
+'
+
+# Env vars:
+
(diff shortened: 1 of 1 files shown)


tools/clickhouse
```diff
@@ -0,0 +1,60 @@
...
+'
+
+# Env vars:
+
+# Optionally set to "podman" to use podman instead
+: "${BB_CLICKHOUSE_CONTAINER_RUNNER:=docker}"
+# Port that clickhouse server will listen on
+: "${BB_CLICKHOUSE_PORT:=9000}"
+# Clickhouse image to use
+: "${BB_CLICKHOUSE_IMAGE:=docker.io/yandex/clickhouse-server:latest}"
```
luluz66, line 18: Can we use version 22.3? the same as the production.

bduffany: done

luluz66 (approved):

PR #5667, opened 2024-01-08
Improve 'Run workflow' UI

bduffany:
* Show loading spinner in the "Run results" button instead of prepending it to the workflow item
* Show results below the workflow item instead of squeezing it in to the left
* Render error messages directly instead of on hover
* Return action names in a deterministic order

BEFORE:

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/826a9560-1250-4cc3-a6b0-b8a5355aad0d

AFTER:

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/f55088c7-57a9-4aff-87ce-c57027e2d8d8

**Related issues**: N/A


siggisim (approved):
Definitely an improvement - I wonder if we should make the styling look more like invocations (i.e cards with a left colored border). 

I know we don't have all of the info needed to render the card - but might make it a little clearer what's going on and make them feel more clickable.

Either way, this is a huge improvement - so I'm not too picky and happy to get this in with or without styling changes.

bduffany:
wdyt? (I added "run results" and "history" titles to distinguish the two types of cards)

image

siggisim (approved):
Looks great!

PR #5675, opened 2024-01-08
SCIM (external user management API) user lookup & user creation.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -795,6 +795,19 @@ func (d *UserDB) GetUserByID(ctx context.Context, id string) (*tables.User, erro
 	return nil, status.NotFoundError("user not found")
 }
 
+func (d *UserDB) GetUserByIDWithoutAuthCheck(ctx context.Context, id string) (*tables.User, error) {
+	var user *tables.User
+	err := d.h.Transaction(ctx, func(tx interfaces.DB) error {
+		u, err := d.getUser(ctx, tx, id)
+		if err != nil {
+			return err
+		}
+		user = u
+		return err
+	})
+	return user, err
+}
+
 func (d *UserDB) GetUserByEmail(ctx context.Context, email string) (*tables.User, error) {
 	auth := d.env.GetAuthenticator()
 	if auth == nil {
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -784,7 +784,15 @@ func (d *UserDB) GetUserByID(ctx context.Context, id string) (*tables.User, erro
 		user = u
 		return err
 	})
-	return user, err
+	if err != nil {
+		return nil, err
+	}
+	for _, g := range user.Groups {
+		if err := perms.AuthorizeGroupAccess(ctx, d.env, g.Group.GroupID); err == nil {
+			return user, nil
+		}
+	}
+	return nil, status.NotFoundError("user not found")
 }
 
 func (d *UserDB) GetUserByEmail(ctx context.Context, email string) (*tables.User, error) {
@@ -901,6 +909,9 @@ func (d *UserDB) getUser(ctx context.Context, tx interfaces.DB, userID string) (
 	rq := tx.NewQuery(ctx, "userdb_get_user").Raw(
 		`SELECT * FROM "Users" WHERE user_id = ?`, userID)
 	if err := rq.Take(user); err != nil {
+		if db.IsRecordNotFound(err) {
+			return nil, status.NotFoundError("user not found")
+		}
 		return nil, err
 	}
 	if err := fillUserGroups(ctx, tx, user); err != nil {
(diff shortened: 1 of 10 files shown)


enterprise/server/scim/scim.go
```diff
@@ -29,68 +99,236 @@ func NewSCIMServer(env environment.Env) *SCIMServer {
...
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte("SCIM API can only be used in conjunction with SAML"))
+			return
+		}
+
+		var h handlerFunc
 		switch r.Method {
 		case "GET":
-			s.getUser(w, r)
+			if r.URL.Path == usersPath {
```
siggisim, line 144: Is your intention to remove the group path altogether? Or is that just coming later?

enterprise/server/scim/scim.go
```diff
@@ -29,68 +99,236 @@ func NewSCIMServer(env environment.Env) *SCIMServer {
...
+		return nil, err
+	}
+	entityURL := build_buddy_url.WithPath("saml/metadata?slug=" + *g.URLIdentifier)
+	u := &tables.User{
+		UserID:    pk,
+		SubID:     fmt.Sprintf("%s/%s", entityURL, ur.UserName),
+		FirstName: ur.Name.GivenName,
+		LastName:  ur.Name.FamilyName,
+		Email:     ur.UserName,
+		Groups: []*tables.GroupRole{
```
siggisim, line 317: Is there a way to specify the User's role?

siggisim (approved):

enterprise/server/oidc/oidc.go
```diff
@@ -610,6 +610,7 @@ func (a *OpenIDAuthenticator) authenticateUser(w http.ResponseWriter, r *http.Re
 	}
 
 	if authHeader := r.Header.Get(authorizationHeader); authHeader != "" {
+		log.Infof("got header: %s", authHeader)
```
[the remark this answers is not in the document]
vadimberezniker: removed.

enterprise/server/scim/scim.go
vadimberezniker: it will be back, has to be redone anyway

enterprise/server/scim/scim.go
vadimberezniker: this will need to be part of group management. theoretically you could have a user attribute that indicates the buildbuddy role but that feels gross.

siggisim: Sounds good

siggisim (approved):

enterprise/server/backends/userdb/userdb.go
```diff
@@ -775,6 +775,27 @@ func (d *UserDB) InsertUser(ctx context.Context, u *tables.User) error {
...
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+	for _, g := range user.Groups {
+		if err := perms.AuthorizeGroupAccess(ctx, d.env, g.Group.GroupID); err == nil {
+			return user, nil
+		}
+	}
```
bduffany, line 794: I'm a bit confused by what's going on here - could you add a comment?

(To me this new block seems the same as checking `len(user.Groups) > 0` - do we now need a stronger guarantee that users have to be part of at least one group?)

vadimberezniker: There's two users involved, the authenticated user and the user being looked up. This is verifying that the authenticated user is in any of the same groups as the user being looked up.

bduffany: I see. Would it make sense to also check that the authenticated user is an admin of the group that they have in common with the user being looked up? Today, only admins have userdb-level access for other users.

vadimberezniker: Done. This also required promoting SCIM API keys to Admin role (see the change in claims.go). 
I had that change in another PR because it's necessary for other group operations, but I moved it into this one since otherwise some tests fail.

server/util/claims/claims.go
```diff
@@ -141,14 +141,18 @@ func ParseClaims(token string) (*Claims, error) {
 }
 
 func APIKeyGroupClaims(akg interfaces.APIKeyGroup) *Claims {
+	keyRole := role.Default
+	// User management through SCIM requires Admin access.
+	if akg.GetCapabilities()&int32(akpb.ApiKey_SCIM_CAPABILITY) > 0 {
+		keyRole = role.Admin
+	}
```
bduffany, line 148: Instead of promoting the SCIM key to admin here, could `GetUserByID` check that the user is either an admin OR the currently authenticated API key has SCIM capability?

vadimberezniker: If we do that, we'd have to do this for every other management function we call from the SCIM code. 
I think we should avoid carving out exceptions for specific API key types. 
The SCIM use case is the first use case where an API key is used to perform administrative operations and in this case I think treating the API key as having an admin role is appropriate.

@tylerwilliams for second opinion

tylerwilliams: I think keeping it simple here, like @vadimberezniker has suggested, is probably the right move until we have a stronger reason for doing it differently. (Theres a good chance all of this scim stuff will change in the future anyway)

bduffany (approved):

PR #5678, opened 2024-01-09
Use the Xcode command-line tools instead of Xcode in Mac release workflows

brentleyjones:
As of Bazel 7.0 we aren’t using the apple_support cc_toolchain. Only `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk` is included in `builtin_include_directories`. If we have `DEVELOPER_DIR` set to something else we can see errors like following:

```
ERROR: /private/var/tmp/_bazel_brentley/7e5d3aaeb704234ad9b80179b286c374/external/com_google_absl/absl/base/BUILD.bazel:95:11: Compiling absl/base/internal/spinlock_wait.cc [for tool] failed: absolute path inclusion(s) found in rule '@@com_google_absl//absl/base:spinlock_wait':
the source file 'absl/base/internal/spinlock_wait.cc' includes the following non-builtin files with absolute paths (if these are builtin files, make sure these paths are in your toolchain):
  '/Applications/Xcode-15.1.0-Release.Candidate.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/atomic'
```

Setting `DEVELOPER_DIR` in our GitHub release workflows ensures that we use the command-line tools rather than whatever is installed on the runner. We also needed to set `--{host,}action_env=DEVELOPER_DIR` to prevent a mismatch between the version of `DEVELOPER_DIR` that Bazel itself sees and what actions/repos see. Finally, I added `BAZEL_NO_APPLE_CPP_TOOLCHAIN=1` to make 100% sure we don’t use the apple_support cc_toolchain.

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -163,8 +163,8 @@ common --nolegacy_important_outputs
 
 # Use a static PATH variable to prevent unnecessary rebuilds of dependencies like protobuf.
 common --incompatible_strict_action_env
-common --action_env=DEVELOPER_DIR
-common --host_action_env=DEVELOPER_DIR
+common:macos --action_env=DEVELOPER_DIR
+common:macos --host_action_env=DEVELOPER_DIR
 
 # rules_nodejs needs runfiles to be explicitly enabled.
 common:linux --enable_runfiles
--- a/.bazelrc
+++ b/.bazelrc
@@ -163,6 +163,8 @@ common --nolegacy_important_outputs
 
 # Use a static PATH variable to prevent unnecessary rebuilds of dependencies like protobuf.
 common --incompatible_strict_action_env
+common --action_env=DEVELOPER_DIR
+common --host_action_env=DEVELOPER_DIR
 
 # rules_nodejs needs runfiles to be explicitly enabled.
 common:linux --enable_runfiles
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -163,6 +163,8 @@ common --nolegacy_important_outputs
 
 # Use a static PATH variable to prevent unnecessary rebuilds of dependencies like protobuf.
 common --incompatible_strict_action_env
+common --action_env=DEVELOPER_DIR
+common --host_action_env=DEVELOPER_DIR
```
sluongng, line 167: I think this could be set for MacOS only to reduce the risk of cache poisoning to only MacOS builds.

I wonder if we should set the default value `/Library/Developer/CommandLineTools` here as well. To override the default, the user could set their own `--action_env=DEVELOPER_DIR=<custom_value>`.

WDYT?

brentleyjones: > I think this could be set for MacOS only to reduce the risk of cache poisoning to only MacOS builds.

Agree. I missed it during my refactoring of stuff.

> I wonder if we should set the default value /Library/Developer/CommandLineTools here as well. To override the default, the user could set their own `--action_env=DEVELOPER_DIR=<custom_value>`

It's possible to not have the command-line tools installed. We want to be able to support people that only have Xcode installed. So I think how I have it is fine.

sluongng (approved):

PR #5685, opened 2024-01-10
Update vtproto version and also generate vtproto code for grpc status

luluz66:
- Update vtprotobuf lib to use well-known types
- Generate vtproto code for grpc status.
   - create a target `grpc_status_go_proto` in `proto/BUILD` that uses `vtprotobuf_compiler.`
   - replaces the dep `@org_golang_google_genproto_googleapis_rpc//status` with `//proto:grpc_status_go_proto`
   - add build_directives to go repositories that imports `google.golang.org/genproto/googleapis/rpc/status` to avoid errors like
   
```
package imports google.golang.org/genproto/googleapis/rpc/status
          was compiled with: @//proto:grpc_status_go_proto
          but was linked with: @org_golang_google_genproto_googleapis_rpc//status:status
```

Previously, `MarshalVT()` for ScoreCard is about 50% slower than reflection based marshaling, and uses much more allocations (524.500 vs 1.0). With this change, `MarshalVT()` is 32.81% faster than reflection based marshaling. and alloc/op is reduced to 1.

This change also should improve the marshaling performance for messages like `rfpb.BatchCmdResponse` and `wfpb.ExecuteWorkflowResponse` when `google.rpc.Status` is in a repeated field.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3018


sluongng (approved):

tylerwilliams (approved):

vadimberezniker (commented):
Will gazelle properly add the correct status dependency to the BUILD files or do we need to put some kind of directive in the top level BUILD file?

luluz66:
> Will gazelle properly add the correct status dependency to the BUILD files or do we need to put some kind of directive in the top level BUILD file?

For go code, gazelle will add the correct status dependency to the BUILD file. All the BUILD files outside the proto directories in this PR were updated by bb fix.

vadimberezniker (approved):

PR #5691, opened 2024-01-11
show outgoing reviews that havent received a review yet

jdhollen:
i imagine this was done because it shows reviews with no reviewers? if so, i can try to address that separately, but this state is more annoying imo.

**Related issues**: N/A


jdhollen:
cc @iain-macdonald

siggisim (approved):
I'm cool with this, but we might want to throw a limit on the number of results we show / fetch individual info for (because I have a ton of these lol)

jdhollen:
Siggi pointed out that we could use `involves:<user>`, but this fix is small and at least fixes the most annoying thing that iain and i have been seeing, so i'm gonna drop it in.

PR #5693, opened 2024-01-11
deps.bzl: make soci_snapshotter use the new grpc compiler

sluongng:
Also formatted the file with

```bash
> buildifier -type workspace deps.bzl
```

to have a consistent ordering of repository rules attributes.


iain-macdonald (approved):
Thanks, Son!

PR #5698, opened 2024-01-11
sluongng/reapply linkopt macos

sluongng:
- Revert "macos: rollback linkopt flag (#5684)"
- testing mac release


sluongng:
This is a weird issue. I discussed with @brentleyjones about this and we found that GoLink on `exec` configuration was using a different linker than CcLink action or other GoLink action on `target` configuration.

We might need to look into using a remote toolchain on MacOS instead of the host's toolchain.
Let's close this for now.

PR #5700, opened 2024-01-11
Refactor github release workflows

maggie-lou:
Fix a couple things related to the github release workflows:

1. If an `automatic_release_tag` is set in the `action-automatic-releases` action, bugs in the action cause two releases to be created, and only one commit is included in the commit history. Revert back to triggering the action via a tag push, which solves these problems.
2. Rename workflows for better readability
3. Create a parent workflow. There are some timing dependencies between the current workflows (i.e. the release has to be created before the mac artifacts can be uploaded to it) that are now better encapsulated
4. Add a slack message if any workflow fails

### DIFF
--- a/.github/workflows/build-linux-github-release-artifacts.yaml
+++ b/.github/workflows/build-linux-github-release-artifacts.yaml
@@ -1,10 +1,18 @@
-name: "release"
+name: "Build Linux Github Release Artifacts"
 
 on:
-  push:
-    tags:
-      - "v*"
   workflow_dispatch:
+    inputs:
+      release_branch:
+        description: "Git branch to checkout."
+        required: true
+        default: "master"
+        type: string
+      version_tag:
+        description: "Version to tag release artifacts."
+        required: true
+        type: string
+  workflow_call:
     inputs:
       release_branch:
         description: "Git branch to checkout."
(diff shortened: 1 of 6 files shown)


maggie-lou:
* `master`
  * **PR #5700** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5700?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5700?utm_source=stack-comment).

.github/workflows/build-linux-github-release-artifacts.yaml
```diff
@@ -15,50 +12,26 @@ on:
         description: "Version to tag release artifacts."
         required: true
         type: string
+  workflow_call:
+    inputs:
+      release_branch:
+        description: "Git branch to checkout."
+        required: true
+        default: "master"
```
iain-macdonald, line 20: This is a kind of scary default, can we just not set a default here?

.github/workflows/build-m1-github-release-artifacts.yaml
```diff
@@ -15,11 +12,21 @@ on:
         description: "Version to tag release artifacts."
         required: true
         type: string
+  workflow_call:
+    inputs:
+      release_branch:
+        description: "Git branch to checkout."
+        required: true
+        default: "master"
```
iain-macdonald, line 20: Ditto here.

.github/workflows/build-mac-github-release-artifacts.yaml
```diff
@@ -15,31 +12,26 @@ on:
         description: "Version to tag release artifacts."
         required: true
         type: string
+  workflow_call:
+    inputs:
+      release_branch:
+        description: "Git branch to checkout."
+        required: true
+        default: "master"
```
iain-macdonald, line 20: Ditto here.

.github/workflows/create-github-release.yaml
```diff
@@ -0,0 +1,82 @@
+name: "Create Github Release"
+
+# NOTE: This workflow does not work well if not triggered by a tag push due to
```
iain-macdonald, line 3: Remove double negative: "NOTE: This workflow only works well when triggered by a tag push..."

.github/workflows/create-github-release.yaml
```diff
@@ -0,0 +1,82 @@
...
+# unexpected behavior in action-automatic-releases when manually setting an
+# automatic_release_tag.
+# To manually trigger this, you can force push a tag.
```
iain-macdonald, line 6: Worth specifying that it need to be a version tag, i.e. start with "v"?

.github/workflows/build-mac-github-release-artifacts.yaml
```diff
@@ -1,9 +1,6 @@
-name: "release-mac"
+name: "Build Mac Github Release Artifacts"
```
bduffany, line 1: uber nit, since we're renaming these files do you mind also adding the arch in the filename like `mac-m1` / `mac-intel`? I get confused between the "mac" and "m1" files sometimes since m1 is also mac :sweat_smile:

bduffany (approved):

iain-macdonald (approved):

PR #5702, opened 2024-01-11
Serialize the existence check and image pull in PullImageIfNecessary

iain-macdonald:
This should fix the serial-image-pulling bug described [here](https://github.com/buildbuddy-io/buildbuddy-internal/issues/2844#issuecomment-1868007914). This is kinda gross, but a more correct fix that simplifies locking in container implementations and moves some of the auth-checking logic into container.go is quite a bit more involved, so I figured this was a good first step.

**Related issues**: [2844](https://github.com/buildbuddy-io/buildbuddy-internal/issues/2844#issuecomment-1868007914)

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -252,8 +252,8 @@ func PullImageIfNecessary(ctx context.Context, env environment.Env, ctr CommandC
 	}
 
 	// TODO(iain): the auth/existence/pull synchronization is getting unruly.
-	unsafemu, _ := pullOperations.LoadOrStore(ctr.IsolationType()+imageRef, &sync.Mutex{})
-	mu, ok := unsafemu.(*sync.Mutex)
+	uncastmu, _ := pullOperations.LoadOrStore(ctr.IsolationType()+imageRef, &sync.Mutex{})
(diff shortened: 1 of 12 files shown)


enterprise/server/remote_execution/container/container.go
```diff
@@ -243,6 +251,16 @@ func PullImageIfNecessary(ctx context.Context, env environment.Env, ctr CommandC
 		})
 		return ctr.PullImage(ctx, creds)
 	}
+
+	// TODO(iain): the auth/existence/pull synchronization is getting unruly.
+	uncastmu, _ := pullOperations.LoadOrStore(ctr.IsolationType()+imageRef, &sync.Mutex{})
```
bduffany, line 256: The `sync.Map` usage here is a memory leak since map entries are never deleted, and the blocking on `mu.Lock` doesn't respect context cancellation / timeout.

Can we use the `singleflight` package for this instead? (example usage in `ociconv.go`)

bduffany: discussed offline, singleflight is not quite what we want because we have to also think about creds. I don't think the memory buildup will be noticeable in practice so this approach lgtm

PR #5718, opened 2024-01-12
Remove domain restriction from SCIM user creation.

vadimberezniker:
The owned domain value is currently tied to automatically adding users by domain which users may want to disable.

This means that SCIM API could be used to create a user with any e-mail, but this is fine because:
 - The user Sub ID will be tied to the SAML login for the org so the created user won't be usable outside that org
 - The user will only belong to the group being administered

**Related issues**: N/A


siggisim (approved):

vadimberezniker:
Maybe this is actually a bad idea.

PR #5719, opened 2024-01-12
Test Grid: fix a bug when computing pagination tokens

luluz66:
The pagination token should the last item in the following sql query:
```
SELECT commit_sha, max(invocation_start_time_usec)
  as latest_created_at_usec FROM "TestTargetStatuses"
 WHERE group_id = '[group_id]'
AND repo_url = '[repo_url]
AND commit_sha != ''
 GROUP BY commit_sha
ORDER BY latest_created_at_usec DESC, commit_sha asc)
```

To repo the bug locally, I run a test from 25 commits, and then open the test grid.

**Before this change,** there are only 20 commits shown, and 5 commits were missing. 
Screenshot from 2024-01-12 09-32-42
**After this change**, all 25 commits were shown.
Screenshot from 2024-01-12 10-37-17

**Related issues**: N/A
Fixes
https://github.com/buildbuddy-io/buildbuddy-internal/issues/3058

### DIFF
--- a/server/target/target.go
+++ b/server/target/target.go
@@ -649,7 +649,7 @@ func readPaginatedTargetsFromOLAPDB(ctx context.Context, env environment.Env, re
 	//      ORDER BY latest_created_at_usec DESC, commit_sha asc)
 	//    WHERE (
 	//      latest_created_at_usec < [ts_from_pagination_token] OR
-	//      (latest_crated_at_usec = [ts_from_pagination_token] AND
+	//      (latest_created_at_usec = [ts_from_pagination_token] AND
 	//      commit_sha > '[commit_sha_from_pagitation_token]'))
 	//    LIMIT [page_size]
 	//  ) AND group_id = '[group_id]'
--- a/server/target/target.go
+++ b/server/target/target.go
@@ -431,7 +431,7 @@ func fetchTargetsFromOLAPDB(ctx context.Context, env environment.Env, q *query_b
 	seenTargets := make(map[string]struct{}, 0)
 	targets := make([]*trpb.TargetMetadata, 0)
 	statuses := make(map[string][]*trpb.TargetStatus, 0)
-	var nextPageToken *tppb.PaginationToken
+	commitTimestamps := make(map[string]int64, 0)
 
 	type row struct {
 		Label                   string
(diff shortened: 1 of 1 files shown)


server/target/target.go
```diff
@@ -511,13 +508,29 @@ func fetchTargetsFromOLAPDB(ctx context.Context, env environment.Env, q *query_b
...
 
+func computeNextPaginationToken(commitTimestamps map[string]int64) *tppb.PaginationToken {
+	var nextPageToken *tppb.PaginationToken
+	for commit, lastCreatedAtUsec := range commitTimestamps {
+		if nextPageToken == nil {
+			nextPageToken = &tppb.PaginationToken{
+				InvocationEndTimeUsec: lastCreatedAtUsec,
+				CommitSha:             commit,
+			}
+		} else if lastCreatedAtUsec < nextPageToken.GetInvocationEndTimeUsec() || (lastCreatedAtUsec == nextPageToken.GetInvocationEndTimeUsec() && commit > nextPageToken.GetCommitSha()) {
```
jdhollen, line 519: InvocationEndTimeUsec is a bit of a misnomer, right?  it's actually the start time of the last invocation the user saw?

server/target/target.go
```diff
@@ -475,17 +475,13 @@ func fetchTargetsFromOLAPDB(ctx context.Context, env environment.Env, q *query_b
...
-			nextPageToken = &tppb.PaginationToken{
-				InvocationEndTimeUsec: row.InvocationStartTimeUsec,
-			}
-			if nextPageToken.GetInvocationEndTimeUsec() == row.InvocationStartTimeUsec && nextPageToken.GetCommitSha() > row.CommitSHA {
-				return nil
-			}
-			nextPageToken.CommitSha = row.CommitSHA
+		created_at_usec := commitTimestamps[row.CommitSHA]
+		if row.InvocationStartTimeUsec > created_at_usec {
+			commitTimestamps[row.CommitSHA] = row.InvocationStartTimeUsec
```
jdhollen, line 480: i'm not quite following--is the core issue that we are merging together all runs for a given commit sha into a single column in the test grid?

luluz66: The way we set the pagination token is wrong; causing some commits won't be selected.

Here is an illustration of the bug:
Let's say we have following data, and the page size is 5. So we are choosing 5 commits per page.
commit1, timestamps: 5000, 4000
commit2, timestamps: 4000, 2000
commit3, timestamps: 3000, 2000
commit4, timestamps: 2000, 1000
commit5, timestamps: 1500, 1000
commit6, timestamps: 1400, 1000
The correct behavior is that the first page will be commit1,commit2, commit3, commit4, commit5; and the next page we will show commit6. The pagination token should be (commit5, 1500)

The current logic will set the pagination token to (commit5, 1000) The following inner select query won't select commit6 because last_created_at_usec of commit6 is 1400 and it's bigger than the timestamp provided in the pagination token (1000). 
```
	//    SELECT commit_sha FROM (
	//      SELECT commit_sha, max(invocation_start_time_usec)
	//      as latest_created_at_usec FROM "TestTargetStatuses"
	//      WHERE group_id = '[group_id]'
	//      AND repo_url = '[repo_url]
	//      AND commit_sha != ''
	//      GROUP BY commit_sha
	//      ORDER BY latest_created_at_usec DESC, commit_sha asc)
	//    WHERE (
	//      latest_created_at_usec < [ts_from_pagination_token] OR
	//      (latest_crated_at_usec = [ts_from_pagination_token] AND
	//      commit_sha > '[commit_sha_from_pagitation_token]'))
	//    LIMIT [page_size]
	```

server/target/target.go
luluz66: It's a bit strange. I don't remember why it's called invocationEndTimeUsec in the pagination token anymore.

server/target/target.go
jdhollen: okay, so the core issue is that the old code picked the minimum start time for the invocations in the result set, when it should have been picking `min(max(invocation start time at commit))`? i get it now, thanks for tolerating my questions.

jdhollen (approved):

PR #5722, opened 2024-01-12
Add flag to always force cloning from filecache

siggisim:
With this flag, an executor can set `--executor.local_cache_always_clone=true` to force the executor to always use clone files from the filecache into the execution root.

### DIFF
--- a/enterprise/server/remote_execution/filecache/filecache.go
+++ b/enterprise/server/remote_execution/filecache/filecache.go
@@ -42,7 +42,7 @@ const (
 	tmpDir = "_tmp"
 )
 
-var enableAlwaysClone = flag.Bool("executor.enable_always_clone_filecache", false, "If true, files from the filecache will always be cloned instead of hardlinked")
+var enableAlwaysClone = flag.Bool("executor.local_cache_always_clone", false, "If true, files from the filecache will always be cloned instead of hardlinked")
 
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -41,6 +42,8 @@ const (
 	tmpDir = "_tmp"
 )
 
+var enableAlwaysClone = flag.Bool("executor.enable_always_clone_filecache", false, "If true, files from the filecache will always be cloned instead of hardlinked")
```
bduffany, line 45: The filecache flag names are getting wild :face_with_spiral_eyes: 
- `executor.local_cache_{directory,size_bytes}`
- `executor.enable_fastcopy_reflinking`
- `executor.enable_always_clone_filecache`

the ship has prob sailed on `enable_fastcopy_reflinking` but maybe instead of `enable_always_clone_filecache` we can call it `local_cache_always_clone`?

siggisim: Done

tylerwilliams (approved):

enterprise/server/remote_execution/filecache/filecache.go
tylerwilliams: I think we can clean this up with flag.Alias

PR #5730, opened 2024-01-16
Don't assume m is non-nil

siggisim:
**Related issues**: N/A

### DIFF
--- a/server/util/filter/filter.go
+++ b/server/util/filter/filter.go
@@ -54,7 +54,7 @@ func invocationMetricToDbField(m stat_filter.InvocationMetricType, paramPrefix s
 }
 
 func MetricToDbField(m *stat_filter.Metric, paramPrefix string) (string, error) {
-	if m.Invocation != nil {
+	if m.GetInvocation() != nil {
 		return invocationMetricToDbField(m.GetInvocation(), paramPrefix)
 	} else if m.Execution != nil {
 		return executionMetricToDbField(m.GetExecution(), paramPrefix)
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
thanks

server/util/filter/filter.go
```diff
@@ -54,7 +54,7 @@ func invocationMetricToDbField(m stat_filter.InvocationMetricType, paramPrefix s
 }
 
 func MetricToDbField(m *stat_filter.Metric, paramPrefix string) (string, error) {
-	if m.Invocation != nil {
+	if m.GetInvocation() != nil {
 		return invocationMetricToDbField(m.GetInvocation(), paramPrefix)
 	} else if m.Execution != nil {
```
jdhollen, line 59: should do this one too

vadimberezniker, line 59: what about m.Execution

jdhollen (approved):

tylerwilliams (approved):

server/util/filter/filter.go
siggisim: I did, but then go yelled at me so I took another approach that should fix both.
```
server/util/filter/filter.go:57:26: invalid operation: m.GetInvocation() != nil (mismatched types stat_filter.InvocationMetricType and untyped nil)
server/util/filter/filter.go:59:32: invalid operation: m.GetExecution() != nil (mismatched types stat_filter.ExecutionMetricType and untyped nil)
```

PR #5735, opened 2024-01-18
.github: add windows release workflow for executor

sluongng:
This should help us to start releasing Windows executor artifacts to our Github Release.

### DIFF
--- a/.github/workflows/create-github-release.yaml
+++ b/.github/workflows/create-github-release.yaml
@@ -47,6 +47,15 @@ jobs:
       version_tag: ${{ github.ref_name }}
     secrets: inherit
 
+  build-windows-artifacts:
+    uses: ./.github/workflows/build-windows-github-release-artifacts.yaml
+    if: "!contains(github.event.head_commit.message, 'release skip')"
+    needs: create-release
+    with:
+      release_branch: ${{ github.ref }}
+      version_tag: ${{ github.ref_name }}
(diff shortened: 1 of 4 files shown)


sluongng:
Question: do we want to mark the name with `-beta`?

sluongng:
Opening this PR up for early review.

Keeping the TODO part for now. Will remove it and add `gh upload` at the end before merge.

maggie-lou (approved):
LGTM just don't forget to remove the `on: pull_request` before you merge

Also do you want to call this from .create-github-release? That's what actually triggers these workflows to run

sluongng:
> Also do you want to call this from .create-github-release? That's what actually triggers these workflows to run

Good call! I added it there.

.github/workflows/create-github-release.yaml
```diff
@@ -47,6 +47,15 @@ jobs:
       version_tag: ${{ github.ref_name }}
     secrets: inherit
 
+  build-windows-artifacts:
```
bduffany, line 50: I think you also need to update `check_release_assets_uploaded.py` to make sure the release waits until these are completed

sluongng: Oh that's a sneaky script that I missed. Thanks for pointing that out.

Added.

bduffany (approved):

PR #5737, opened 2024-01-19
hit_tracker: log error to debug logs

sluongng:
This should allow us to surface the errors coming from redis if there
are any.


server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -153,10 +154,14 @@ func (s *ActionCacheServer) GetActionResult(ctx context.Context, req *repb.GetAc
...
 	if err != nil {
-		ht.TrackMiss(d)
+		htErr := ht.TrackMiss(d)
+		log.Debugf("GetActionResult: hit tracker error: %s", htErr)
 		return nil, status.NotFoundErrorf("ActionResult (%s) not found: %s", d, err)
 	}
-	defer downloadTracker.CloseWithBytesTransferred(int64(len(blob)), int64(len(blob)), repb.Compressor_IDENTITY, "ac_server")
+	defer func() {
+		dtErr := downloadTracker.CloseWithBytesTransferred(int64(len(blob)), int64(len(blob)), repb.Compressor_IDENTITY, "ac_server")
+		log.Debugf("GetActionResult: download tracker error: %s", dtErr)
```
bduffany, line 163: - Use `CtxDebug` fn (to get invocation ID annotations etc.)
- Use `err` for error name (since there's no need to distinguish it from other errors in scope)
- Only log if err is non-nil, otherwise lots of log spam will be generated.

```suggestion
		if err := downloadTracker.CloseWithBytesTransferred(int64(len(blob)), int64(len(blob)), repb.Compressor_IDENTITY, "ac_server"); err != nil {
			log.CtxDebugf(ctx, "GetActionResult: download tracker error: %s", err)
	}
```

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -101,7 +101,8 @@ func (s *ByteStreamServer) Read(req *bspb.ReadRequest, stream bspb.ByteStream_Re
 
 	ht := hit_tracker.NewHitTracker(ctx, s.env, false /*=ac*/)
 	if r.IsEmpty() {
-		ht.TrackEmptyHit()
+		htErr := ht.TrackEmptyHit()
+		log.Debugf("ByteStream Read: hit tracker TrackEmptyHit error: %s", htErr)
```
bduffany, line 105: (same comment here)

server/remote_cache/byte_stream_server/byte_stream_server.go
```diff
@@ -113,7 +114,8 @@ func (s *ByteStreamServer) Read(req *bspb.ReadRequest, stream bspb.ByteStream_Re
 	}
 	reader, err := s.cache.Reader(ctx, cacheRN.ToProto(), req.ReadOffset, req.ReadLimit)
 	if err != nil {
-		ht.TrackMiss(r.GetDigest())
+		htErr := ht.TrackMiss(r.GetDigest())
+		log.Debugf("ByteStream Read: hit tracker TrackMiss error: %s", htErr)
```
bduffany, line 118: (same comment here)

bduffany (changes requested):

server/remote_cache/action_cache_server/action_cache_server.go
sluongng: Thanks for this. What a brain fog from my part 🙈

sluongng:
added `if` clause and handle other places that called `CloseWithBytesTransferred`

server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -153,10 +154,16 @@ func (s *ActionCacheServer) GetActionResult(ctx context.Context, req *repb.GetAc
 	downloadTracker := ht.TrackDownload(d)
 	blob, err := s.cache.Get(ctx, rn.ToProto())
 	if err != nil {
-		ht.TrackMiss(d)
+		if htErr := ht.TrackMiss(d); htErr != nil {
```
bduffany, line 157: This LGTM but the prefixed errors are unconventional/weird :P `if err := ...; err != nil { ... }` is more idiomatic. If you're worried about clobbering other `err` variables in the same scope, the `err` is only scoped to the `if` block so that's not an issue.

bduffany (approved):

server/remote_cache/action_cache_server/action_cache_server.go
sluongng: The initial implementation is pretty much just me spelunking for Redis-related issues, so it was named that way to avoid clobbering.

Now that we added the `if` clause, it made more sense to use `err` like you suggested.

Pushed a change for this

PR #5738, opened 2024-01-19
Upgrade bazelisk version bundled with CI runner

bduffany:
This is needed to support patterns like `USE_BAZEL_VERSION=6.x`

**Related issues**: N/A


brentleyjones (approved):

sluongng:
We might need to upgrade bazelisk in our workflows as well.

bduffany:
urg, it's refusing to build with race detection because something in bazelisk's build graph is disabling CGO :|

might just go back to the http_file approach for now.

PR #5740, opened 2024-01-22
googleapis: upgrade to latest version

sluongng:
This includes the change in https://github.com/googleapis/googleapis/pull/831
which let us retire our custom patch.


tylerwilliams (approved):
yay less cruft :) Thanks!

PR #5741, opened 2024-01-22
Fix noisy COWStore rate limiter logs

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -649,9 +649,11 @@ func (s *COWStore) eagerFetchChunksInBackground() {
 			break // context canceled
 		}
 		if err := rateLimiter.Wait(ctx); err != nil {
-			if err != s.ctx.Err() {
-				log.CtxErrorf(s.ctx, "COWStore eager fetch rate limiter failed, stopping eager fetches: %s", err)
+			// Ignore if the ctx is done, since that's probably the root issue.
+			if ctx.Err() != nil {
+				return
 			}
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -649,7 +649,8 @@ func (s *COWStore) eagerFetchChunksInBackground() {
 			break // context canceled
 		}
 		if err := rateLimiter.Wait(ctx); err != nil {
-			if err != s.ctx.Err() {
+			// Ignore err if ctx is cancelled (that's probably the cause)
+			if ctx.Err() == nil {
```
maggie-lou, line 653: Could you explain why this fixes the problem? Having some trouble understanding the context for why errors would be set on the parent context and not the child

bduffany: On line `639` we create a child context `ctx` from the parent `s.ctx`. That context gets cancelled when we call `Close()`, but `s.ctx` is not necessarily guaranteed to be cancelled yet when that happens.

maggie-lou: That makes sense. Should we just check if ctx.Err is ContextCanceled? This condition is not very intuitive to me `if ctx.Err() == nil`

Could you also add a comment that we'd expect the child context to be cancled on shutdown due to line 643

bduffany: I inverted the logic to make it less confusing.

I don't want to check explicitly for Canceled since DeadlineExceeded is also possible and it seems unnecessary to check for both explicitly.

maggie-lou (approved):

PR #5743, opened 2024-01-22
Fix snapshot reuse when base image chunks expire from cache

bduffany:
**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3069

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/BUILD
+++ b/enterprise/server/remote_execution/containers/firecracker/BUILD
@@ -163,7 +163,7 @@ go_test(
         "--executor.firecracker_enable_merged_rootfs=true",
         "--executor.firecracker_enable_uffd=true",
         "--executor.enable_local_snapshot_sharing=true",
-        #        "--executor.enable_remote_snapshot_sharing=true", # TODO: Re-enable once performance has improved
+        "--executor.enable_remote_snapshot_sharing=true",
     ],
     exec_properties = {
         "test.Pool": "bare",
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/containers/firecracker/BUILD
```diff
@@ -163,7 +163,7 @@ go_test(
         "--executor.firecracker_enable_merged_rootfs=true",
         "--executor.firecracker_enable_uffd=true",
         "--executor.enable_local_snapshot_sharing=true",
```
bduffany, line 165: @maggie-lou  I re-enabled this since this test is marked with `tags = ["manual"]` anyway

enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -850,14 +848,84 @@ printf '%s' $ATTEMPT_NUMBER | tee ./attempts
...
+
+	// Evict all artifacts from filecache, which should expire the base image.
+	fcDir := testfs.MakeTempDir(t)
+	fc, err := filecache.NewFileCache(fcDir, fileCacheSize, false)
+	require.NoError(t, err)
+	fc.WaitForDirectoryScanToComplete()
+	env.SetFileCache(fc)
+
+	run("1") // Should start clean
+	run("2") // Should resume from previous, and increment the counter
```
bduffany, line 926: confirmed that "1" is printed before making the change :+1:

maggie-lou (approved):

PR #5746, opened 2024-01-23
sluongng/cli bisect

sluongng:
Experimental "bb bisect" command


sluongng:
I won't have spare cycles to complete this feature for the foreseeable future.
Let's close the PR for now and revisit when I find the time.

PR #5748, opened 2024-01-24
bazelisk: use Bazelisk built from source

sluongng:
Address the issue we seen in #5738 where the compilation of bazelisk
binary was blocked by `executor_windows` cross compilation.

It seems like we could not compile Go Stdlib when both race and pure go
config are enabled. The 'race' config was introduced by our flag in
bazelrc

```
common:race --@io_bazel_rules_go//go/config:race
```

while the `pure` config (no CGO usage) was trickled down from the
`executor_windows` cross compilation target. We needed pure there
because we do not have a Windows CC toolchain available while cross
compiling from Linux/MacOS.

Disable building `executor_windows` directly with `tags = [manual]`.
Instead, build it via a proxy rule `go_norace_binary`.

`go_norace_binary` is a slim down version of `go_cross_binary` in
rules_go. It applies a transition, `go_norace_transition`, on the
binary target to ensure that `race` setting is always off. This forces
Bazel to build `:executor_windows_norace` with race off, even though we
are enabling `--@io_bazel_rules_go//go/config:race` explicitly with
Bazel flag (or `.bazelrc`).


bduffany (commented):
I am not super familiar with transitions - will this cause actions to be run once for each configuration in the graph? just wondering how many actions this will add to the graph (if any)

sluongng:
That's a nerd snipe question 😜. So I did some tests

```
COMMANDS:
  bazel clean --async
  bazel test --config=remote --config=race --remote_instance_name=$RANDOM //...

# master (baseline)
INFO: 13910 processes: 2323 internal, 38 local, 11549 remote.

# With `norace` transition applied, without "manual" tag
INFO: 15431 processes: 2344 internal, 38 local, 13049 remote. (+10.93%)

# With `norace` transition applied, with "manual" tag
INFO: 13928 processes: 2341 internal, 38 local, 11549 remote. (+00.13%)

# HEAD
INFO: 13947 processes: 2341 internal, 38 local, 11568 remote. (+00.27%)
```

The quick explanation is: yes, a transition will create a new bazel `configuration` which makes all the transitive dependencies rebuild with that new `configuration`.

So if we are building both `:executor_windows` and `:executor_windows_norace` while having `config:race` enabled, then the transitive dependencies will be built twice, once with race detection enabled and once with race detection disabled. However, because we tagged `:executor_windows` with `"manual"`, it won't be built directly via `bazel test //...`, but only through `:executor_windows_norace` as one of its dependencies. 

Effectively, with this transition and the `"manual"` tag, we are replacing an existing Bazel configuration with another, so the action counts stay relatively the same. Only a few actions (37) were added, I guess these could be attributed to:

a. New `go_norace_binary` rule doing a symlink action.

b. We are cross-compiling `bazelisk` targeting Windows.

I suspect we could further reduce this by combining (1) `server/util/bazelisk` and (2) `server/testutil/testbazelisk/`. (1) is embedding the Bazelisk binary into our Executor binary (resulting in 1 big Executor binary), while (2) only includes the `bazelisk` binary as runfiles (resulting in smaller test binaries referring to a standalone `bazelisk` binary on disk). But I will not look into that for now.

bduffany (commented):
to solve the original issue would it be enough to just set `tags = ["manual"]` on the executor_windows target or is the transition also needed?

sorry for pushing back on this, I have just had some mildly bad experience with transitions (causing some hash in the file paths to change which broke our esbuild setup which depended on the paths being the same across multiple build configurations) and have heard other mixed opinions about them as well (being hard to understand, as well as bloating the build graph and slowing down builds). Even if it doesn't bloat things much now, it might add more bloat over time (perhaps unintentionally, since we don't have the greatest build graph analysis tools available yet). So I would like to give more thought to simpler alternatives first :P

sluongng:
Another note is that the 

```
    goarch = "amd64",
    goos = "windows",
    pure = "on",
```

attributes in `go_binary` are also applied via `transition` in rules_go 🙈

sluongng:
@bduffany PTAL

bduffany (approved):
LGTM

do you mind also removing the bazelisk downloads from deps.bzl?

PR #5752, opened 2024-01-25
Add codesearch protos

tylerwilliams:
Moved to main because it makes things easier. Also added request and response context.

### DIFF
--- a/proto/index.proto
+++ b/proto/index.proto
@@ -5,8 +5,8 @@ import "proto/context.proto";
 package index;
 
 message GitRepoRequest {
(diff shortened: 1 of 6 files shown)


proto/index.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+import "proto/context.proto";
+
+package index;
+
+message GitRepoRequest {
+  // A repo url like "buildbuddy-io/buildbuddy".
+  string repo_url = 1;
```
jdhollen, line 9: owner_repo

jdhollen (approved):

PR #5758, opened 2024-01-26
website: remove Bazel 6.4.0 usage in workflows

sluongng:
When we upgraded to Bazel 7.0.0 in the past, we encountered problem with
directly depending on `bazel-bin` convienience symlink created by Bazel.

In #5626, we introduced a custom Docusaurus plugin to customize how the
`bazel-bin` path is resolved and fixed this problem.

Let's make sure our workflows stay on the same Bazel version.


bduffany (approved):
thanks for reverting this :)

PR #5763, opened 2024-01-29
Upgrade go-fuse and remove patch

bduffany:
The patch is available upstream starting with v2.4.1, in https://github.com/hanwen/go-fuse/commit/b5c46276a19e02b7912f998e4ccee85330458f4d

**Related issues**: N/A


deps.bzl
```diff
@@ -2518,9 +2518,8 @@ def install_go_mod_dependencies(workspace_name = "buildbuddy"):
         importpath = "github.com/hanwen/go-fuse/v2",
         patch_args = ["-p1"],
         patch_tool = "patch",
```
sluongng, line 2520:
```suggestion
```

let's remove these too

sluongng (approved):
LGTM. I think soci depends on this as well so @iain-macdonald might want to test this out after.

iain-macdonald:
> LGTM. I think soci depends on this as well so @iain-macdonald might want to test this out after.

The parts of soci-snapshotter that are linked into the app don't rely on the go-fuse fix, which is in 2.4.2 anyways, so this PR looks good to me.

iain-macdonald (approved):

PR #5764, opened 2024-01-29
Proto changes for showing snapshot metadata in the UI

bduffany:
A few different users have run into some confusion recently about whether or not a particular VM was resumed from snapshot.

This PR adds some proto changes that will allow us to see the complete "history" of a particular snapshot in the UI. We'll be able to render the VM's ID (which is preserved across snapshots) as well as a link to the invocation + execution which created the snapshot that we resumed from.

More details in https://github.com/buildbuddy-io/buildbuddy-internal/issues/3078

**Related issues**: N/A


proto/remote_execution.proto
```diff
@@ -2266,3 +2269,22 @@ message StoredExecution {
...
+    // The ID of the invocation that created this snapshot.
+    string invocation_id = 1;
+
+    // The ID of the execution that created this snapshot.
+    string execution_id = 2;
+  }
+
+  // Who created the last snapshot that this VM was resumed from.
+  // If the VM was created from scratch, this field will be unset.
+  Creator last_snapshot_creator = 2;
```
maggie-lou, line 2289: Would naming this something like `last_snapshot_user` be more clear? Especially if we are using this field in code for example and don't have the additional comments, I could see confusion around whether this is whatever created the snapshot from scratch

tylerwilliams (approved):

proto/remote_execution.proto
bduffany: hm I think `user` is inaccurate because this is who's creating the snapshot, not who's using it.

I could sort of see the confusion with the `Create()` method but can't think of a better alternative :shrug:

bduffany: actually wdyt about `snapshot_updated_by`?

maggie-lou: How about `snapshot_last_updated_by`? That is clearer to me than `last_creator` though!

bduffany: on second thought :sweat_smile: I realized snapshots are immutable so I'm not sure I like "updated" actually. I changed it to `last_executed_task` which I think is unambiguous - it's the last task that was executed on the VM (which should be the task that took the snapshot). wdyt?

maggie-lou (approved):

PR #5768, opened 2024-01-29
Don't treat no-op group update as error.

vadimberezniker:
If you make a no-op update via the UI, it currently displays a "Group not found" error.

**Related issues**: N/A


tylerwilliams (approved):
lgtm

bduffany (approved):

PR #5769, opened 2024-01-29
Make URLIdentifier a non-pointer field.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/server/util/claims/claims.go
+++ b/server/util/claims/claims.go
@@ -202,7 +202,7 @@ func ClaimsFromSubID(ctx context.Context, env environment.Env, subID string) (*C
 
 			// If the user requested impersonation but the subdomain doesn't
 			// match the impersonation target then don't enable impersonation.
-			if sd := subdomain.Get(ctx); sd != "" && ig.URLIdentifier != nil && sd != *ig.URLIdentifier {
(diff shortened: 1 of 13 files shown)


server/util/claims/claims.go
```diff
@@ -202,7 +202,7 @@ func ClaimsFromSubID(ctx context.Context, env environment.Env, subID string) (*C
 
 			// If the user requested impersonation but the subdomain doesn't
 			// match the impersonation target then don't enable impersonation.
-			if sd := subdomain.Get(ctx); sd != "" && ig.URLIdentifier != nil && sd != *ig.URLIdentifier {
+			if sd := subdomain.Get(ctx); sd != "" && sd != ig.URLIdentifier {
```
tylerwilliams, line 205: does this need to check `ig.URLIdentifier != ""`?

tylerwilliams (approved):

server/util/claims/claims.go
vadimberezniker: No. I think the old code was incorrect. If the subdomain is set, we should ignore the impersonation if the url identifier doesn't match regardless of whether the identifier is set or not set.

PR #5770, opened 2024-01-29
[RB] Consolidate db query to fetch api key 

maggie-lou:
This will now have parity with workflow service code


maggie-lou:
* `master`
  * **PR #5770** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5770?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5770?utm_source=stack-comment).

enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -217,10 +198,13 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest,
 }
 
 func (r *runnerService) withCredentials(ctx context.Context, req *rnpb.RunRequest) (context.Context, error) {
-	apiKey, err := r.lookupAPIKey(ctx)
-	if err != nil {
-		return nil, err
+	apiKeys := metadata.ValueFromIncomingContext(ctx, authutil.APIKeyHeader)
```
vadimberezniker, line 201: What do you think about using GetAPIKeyForInternalUseOnly to have parity with the workflow code? 
I don't love the idea of looking at the API key header directly because if we make changes to the auth layer, this could potentially break since we generally don't expect non-auth code to be poking at these headers.

maggie-lou: Ack - changed

enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -217,20 +196,37 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest,
 }
 
 func (r *runnerService) withCredentials(ctx context.Context, req *rnpb.RunRequest) (context.Context, error) {
-	apiKey, err := r.lookupAPIKey(ctx)
+	gid, err := r.groupID(ctx)
+	if err != nil {
+		return nil, err
+	}
```
bduffany, line 202: nit: can use `perms.AuthenticatedUser` here instead

```
u, err := perms.AuthenticatedUser(r.env, ctx)
if err != nil {
  return nil, err
}
... GetAPIKeyForInternalUseOnly(ctx, u.GetGroupID())
```

bduffany (approved):

PR #5775, opened 2024-01-30
Export remote build duration to customers

luluz66:
**Related issues**: N/A
Customers want to monitor and/or alert on their remote build usage. 

This PR created a new metric in metrics.go to capture the linux and mac remote
execution duration; and added a new exported metric in prom.go


siggisim (approved):
Will this auto update the docs or do we have to do that separately?

luluz66:
> Will this auto update the docs or do we have to do that separately?

It will auto update once this PR is merged.

PR #5776, opened 2024-01-30
Fix target artifacts card rendering empty/0 digests

bduffany:
**Related issues**: N/A

### DIFF
--- a/app/util/cache.tsx
+++ b/app/util/cache.tsx
@@ -1,3 +1,4 @@
+import { build_event_stream } from "../../proto/build_event_stream_ts_proto";
 import { build } from "../../proto/remote_execution_ts_proto";
 import { resource } from "../../proto/resource_ts_proto";
 import Long from "long";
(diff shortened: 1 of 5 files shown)


app/util/cache.tsx
```diff
@@ -10,6 +11,8 @@ type Compressor = build.bazel.remote.execution.v2.Compressor.Value;
 const DigestFunction = build.bazel.remote.execution.v2.DigestFunction.Value;
 type DigestFunction = build.bazel.remote.execution.v2.DigestFunction.Value;
 
+const BYTESTREAM_URL_PATTERN = /^bytestream:\/\/(?<address>[^\/]+)\/((?<instanceName>.+?)\/)?(blobs|compressed-blobs\/(?<compressor>zstd))\/((?<ac>ac)\/)?((?<digestFunction>blake3)\/)?(?<hash>[0-9a-f]+)\/(?<sizeBytes>\d+)$/;
```
jdhollen, line 14: The digest function match here kinda freaks me out because it only supports the one value that we currently support.

Maybe we need to start sharing test cases between client and server or something?  Then adding support for some new field to the server wouldn't break the client?

jdhollen (approved):

PR #5778, opened 2024-01-30
use userID instead of groupID in userGroupPerm

luluz66:
**Related issues**: N/A
This PR change the GroupAuthPermissions to take in interfaces.UserInfo instead
of groupID, and renamed it to DefaultPermissions. When a personal key is used, the user_id will be the real user ID; and empty otherwise.

As a result, we will associate user_id with invocations, executions, git repositories 

Added tests in invocationdb_test.go, execution_server_test.go and build_handler_event_test.go to make sure user_id is set.

I reserve the current implementation of perms.GroupAuthPermissions and name it perms.GroupPermissionsDeprecated. This currently is used in CreateWorkflow in enterprise/server/workflow/service/service.go. Since Workflow table is to be replaced with GitRepositories, I didn't bother to change the code to use the new DefaultPermissions.

### DIFF
--- a/server/util/perms/perms.go
+++ b/server/util/perms/perms.go
@@ -61,7 +61,7 @@ func DefaultPermissions(u interfaces.UserInfo) *UserGroupPerm {
 // and Workflows table itself is deprecated.
 //
 // Please use DefaultPermissions if possible.
-func GroupPermissionsDeprecated(groupID string) *UserGroupPerm {
+func DeprecatedGroupPermissions(groupID string) *UserGroupPerm {
 	return &UserGroupPerm{
 		UserID:  groupID,
 		GroupID: groupID,
--- a/server/util/perms/perms.go
+++ b/server/util/perms/perms.go
@@ -46,7 +46,7 @@ func AnonymousUserPermissions() *UserGroupPerm {
 	}
 }
 
-func GroupAuthPermissions(u interfaces.UserInfo) *UserGroupPerm {
+func DefaultPermissions(u interfaces.UserInfo) *UserGroupPerm {
 	return &UserGroupPerm{
 		UserID:  u.GetUserID(),
 		GroupID: u.GetGroupID(),
@@ -54,7 +54,14 @@ func GroupAuthPermissions(u interfaces.UserInfo) *UserGroupPerm {
 	}
 }
 
-func GroupAuthPermissionsByGroupID(groupID string) *UserGroupPerm {
+// Deprecated.
+// We used to use this function to set group permissions prior to introducing
+// user personal keys. Currently, this is only used when we insert into Workflows
+// table (https://github.com/buildbuddy-io/buildbuddy/blob/v2.38.0/enterprise/server/workflow/service/service.go#L271)
+// and Workflows table itself is deprecated.
+//
+// Please use DefaultPermissions if possible.
+func GroupPermissionsDeprecated(groupID string) *UserGroupPerm {
 	return &UserGroupPerm{
 		UserID:  groupID,
 		GroupID: groupID,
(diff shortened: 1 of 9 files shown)


server/util/perms/perms.go
```diff
@@ -46,7 +46,15 @@ func AnonymousUserPermissions() *UserGroupPerm {
...
 }
 
-func GroupAuthPermissions(groupID string) *UserGroupPerm {
+func GroupAuthPermissions(u interfaces.UserInfo) *UserGroupPerm {
+	return &UserGroupPerm{
+		UserID:  u.GetUserID(),
+		GroupID: u.GetGroupID(),
+		Perms:   GROUP_READ | GROUP_WRITE,
+	}
+}
```
bduffany, line 55: I think we can drop `Auth` from this name since it's redundant with `Permissions`.

maybe @tylerwilliams or @vadimberezniker can weigh on this as well but I wonder if it would make sense to call this `DefaultPermissions` since these perms seem like a pretty sensible default for most things - the object is attributed to the user but we only set Perms bits for the group - which allows the user to later be kicked out of the group and then lose access to the object, while still retaining info about which user created the object.

server/util/perms/perms.go
```diff
@@ -46,7 +46,15 @@ func AnonymousUserPermissions() *UserGroupPerm {
...
+
+func GroupAuthPermissionsByGroupID(groupID string) *UserGroupPerm {
```
bduffany, line 57: maybe call this something like `GroupPermissionsOnly` since only the group is given permissions, not the user.

(Side note, I think it was a mistake to set `UserID` to `groupID` here - a group is not a user so this doesn't "typecheck" if that makes sense. Not sure if anything would break if we changed it so probably just leave as-is :shrug: )

server/util/perms/perms.go
tylerwilliams: +1 to dropping Auth from the name.

I'm also cool with DefaultPermissions if @vadimberezniker is down

server/util/perms/perms.go
tylerwilliams: +1

tylerwilliams (changes requested):

server/util/perms/perms.go
vadimberezniker: Yeah, sounds good to me.

server/util/perms/perms.go
luluz66: I renamed this function to be GroupPermissionsDeprecated. I added a comment to the function; basically this is only used when we insert into workflow table. It requires some refactoring to make it use DefaultPermissions. Since that table is going to be replaced by GitRepositories in the future, I think we can just keep the behavior unchanged.

luluz66 (commented):
Also, updated the PR descriptions, added a test in execution_server_test.go, and tested the personal user id with Executions.

server/util/perms/perms.go
```diff
@@ -46,7 +46,22 @@ func AnonymousUserPermissions() *UserGroupPerm {
...
+}
+
+// Deprecated.
+// We used to use this function to set group permissions prior to introducing
+// user personal keys. Currently, this is only used when we insert into Workflows
+// table (https://github.com/buildbuddy-io/buildbuddy/blob/v2.38.0/enterprise/server/workflow/service/service.go#L271)
+// and Workflows table itself is deprecated.
+//
+// Please use DefaultPermissions if possible.
+func GroupPermissionsDeprecated(groupID string) *UserGroupPerm {
```
tylerwilliams, line 64: should probably be called `DeprecatedGroupPermissions`

tylerwilliams (approved):
LGTM

bduffany (approved):

PR #5782, opened 2024-01-31
SCIM fixes for Azure AD.

vadimberezniker:
- Map role information in two different attributes to make both Okta and Azure AD happy. I didn't find a simple way to get both of them to play nicely with the same type of attribute.

 - Accept active attribute as a string instead of a boolean to make Azure AD happy.

 - Allow all attribute updates via PATCH.

 - Add more logging to make debugging issues easier in the future.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/scim/scim.go
+++ b/enterprise/server/scim/scim.go
@@ -18,6 +18,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
 	"github.com/buildbuddy-io/buildbuddy/server/tables"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/role"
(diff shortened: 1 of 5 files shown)


siggisim (approved):

enterprise/server/scim/scim.go
```diff
@@ -399,11 +431,24 @@ func (s *SCIMServer) createUser(ctx context.Context, r *http.Request, g *tables.
...
+// Azure AD incorrectly sends "active" field as a string instead of a native
+// boolean...
+func getBooleanValue(v any) (bool, error) {
+	if v, ok := v.(bool); ok {
+		return v, nil
+	}
+	if v, ok := v.(string); ok {
+		return strings.EqualFold(v, "true"), nil
+	}
+	return false, status.InvalidArgumentErrorf("boolean field has unexpected type %T", v)
```
tylerwilliams, line 443: do you want to print it with %v too?

tylerwilliams (approved):

enterprise/server/scim/scim.go
vadimberezniker: done

PR #5786, opened 2024-01-31
RFC: windows PR workflows

sluongng:
An alternative approach to https://github.com/buildbuddy-io/buildbuddy/pull/5748

Verify Windows build in a separate Github workflows.
This lets us remove the current cross-compilation target that uses Bazel's `transition` underneath.


.github/workflows/pr-windows.yaml
```diff
@@ -0,0 +1,21 @@
+name: "PR Windows"
```
bduffany, line 1: nit: change the workflow name and file name to match what the workflow is doing, like `name: "Build Windows executor"` / `build-executor-win.yaml`.

.github/workflows/pr-windows.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Build
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          bazelisk --output_user_root=D:/0 --windows_enable_symlinks build --config=untrusted-ci-windows --remote_header=x-buildbuddy-api-key=${{ secrets.BUILDBUDDY_ORG_API_KEY }} //enterprise/server/cmd/executor:executor
```
bduffany, line 21: nit: can you add a comment about why `--windows_enable_symlinks` is enabled?

also maybe add a comment about why we're setting `--output_user_root=D:/0`

bduffany (approved):

sluongng:
Adding @tylerwilliams @siggisim for pricing awareness.

This PR will turn on the Github-hosted Windows workflows on every PRs to validate Windows Executor build.

Reference:
- https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions
- https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

PR #5787, opened 2024-01-31
release-cli: use Github M1 runner instead of our "m1"

sluongng:
The github runner has Bazelisk preinstalled, so we could get rid of our
custom Bazelisk installation step as well.

### DIFF
--- a/.github/workflows/release-cli.yaml
+++ b/.github/workflows/release-cli.yaml
@@ -115,17 +115,10 @@ jobs:
       matrix:
         include:
           - os: ubuntu-20.04
-            shell: bash
-          - os: macos-11
-            shell: bash
-          - os: m1
-            # TODO: Can we remove this if we upgrade our m1 runner to native arm64?
-            shell: /usr/bin/arch -arch arm64e bash --noprofile --norc -eo pipefail {0}
+          - os: macos-13
+          - os: macos-13-xlarge
     runs-on: ${{ matrix.os }}
     needs: create-release
-    defaults:
-      run:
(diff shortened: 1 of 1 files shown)


sluongng:
Tested with https://github.com/buildbuddy-io/buildbuddy/actions/runs/7726152613/

bduffany (approved):
LGTM

sanity check: do the CLI release artifacts from your test run have the correct architecture? (I couldn't find it on the releases page)

.github/workflows/release-cli.yaml
```diff
@@ -115,17 +115,10 @@ jobs:
...
         include:
           - os: ubuntu-20.04
-            shell: bash
-          - os: macos-11
-            shell: bash
-          - os: m1
-            # TODO: Can we remove this if we upgrade our m1 runner to native arm64?
-            shell: /usr/bin/arch -arch arm64e bash --noprofile --norc -eo pipefail {0}
+          - os: macos-13
+          - os: macos-13-xlarge
```
bduffany, line 119: maybe add a comment here that one of these is x86_64 and one is arm64

sluongng:
Yup it does have the right architecture https://github.com/buildbuddy-io/buildbuddy/actions/runs/7726152613/job/21061866087#step:1:8

We are using the same runner for the weekly M1 release workflows.

Let me add a comment.

.github/workflows/release-cli.yaml
sluongng: done

PR #5788, opened 2024-01-31
Add option to set env vars to ExecuteWorkflow api

maggie-lou:
Customers were requesting this in the context of passing short-lived aws tokens to workflows. 

As a follow up, I plan on adding support for setting env vars to remote bazel

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -1139,6 +1139,9 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 			// NOTE: The executor is responsible for making sure this
 			// buildbuddy_ci_runner binary exists at the workspace root. It does so
 			// whenever it sees the `workflow-id` platform property.
+			//
+			// Also be cautious when adding new flags. See
+			// https://github.com/buildbuddy-io/buildbuddy-internal/issues/3101
 			"./buildbuddy_ci_runner",
 			"--invocation_id=" + invocationID,
 			"--action_name=" + workflowAction.Name,
@@ -1157,7 +1160,6 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 			"--workflow_id=" + wf.WorkflowID,
 			"--trigger_event=" + wd.EventName,
 			"--bazel_command=" + ws.ciRunnerBazelCommand(),
-			"--env_overrides=" + string(envJson),
 			"--debug=" + fmt.Sprintf("%v", ws.ciRunnerDebugMode()),
 		}, extraArgs...),
 		Platform: &repb.Platform{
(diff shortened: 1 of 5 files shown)


maggie-lou:
* `master`
  * **PR #5788** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5788?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5788?utm_source=stack-comment).

proto/api/v1/workflow.proto
```diff
@@ -32,6 +32,11 @@ message ExecuteWorkflowRequest {
   string visibility = 5;
   // If true, start the workflow but do not wait for the status to be returned.
   bool async = 6;
+  // These env vars will be applied to each action that is run. If there is a
+  // conflict between an env var set here and in buildbuddy.yaml, these
+  // overrides will take precedence. Otherwise all env vars set in
+  // buildbuddy.yaml will still apply.
+  map<string, string> env_var_overrides = 7;
```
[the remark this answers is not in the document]
maggie-lou: Done - the only one I left was in ci_runner/main.go, because I thought it may be unclear the precedence between env vars set via flag vs on the command vs in buildbuddy.yaml

enterprise/server/workflow/service/service.go
```diff
@@ -1153,6 +1157,7 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 			"--workflow_id=" + wf.WorkflowID,
 			"--trigger_event=" + wd.EventName,
 			"--bazel_command=" + ws.ciRunnerBazelCommand(),
+			"--env_overrides=" + string(envJson),
```
bduffany, line 1160: we've broken things in the past when adding new flags to the ci_runner, because Go crashes when it sees unrecognized flags. There will be older ci_runners that are assigned to this task because we distribute the CI runner with the executor rather than the app, and there are people that run self-hosted workflow executors. (This also will temporarily be broken while we do the rollout)

I think ideally we'd distribute the CI runner binary with the app (rather than distributing it w/ the executors) and upload it to CAS and make it part of the Action, so that the flags are kept in lock-step with the apps. But the last time we looked at doing that we decided it was complicated, since it makes the apps depend on both macos and linux builds of the CI runner.

For now, we can conditionally pass the `--env_overrides` flag  (i.e. if env is empty then don't set the flag). (Maybe also add a comment here for future readers if you don't mind :pray: )

maggie-lou: Done

enterprise/server/workflow/service/service.go
```diff
@@ -1181,6 +1188,11 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
 			},
 		},
 	}
+	// See https://github.com/buildbuddy-io/buildbuddy-internal/issues/3101
+	// for more info on why we don't want to pass empty flags
+	if len(string(envJson)) > 0 {
```
bduffany, line 1193: need `len(env) > 0` here instead - if `len(env) == 0` then `envJson` will be `{}` I think.

maggie-lou: Ah good catch thanks

bduffany (approved):

PR #5792, opened 2024-02-01
Enable bzlmod testing

sluongng:
Partially enable bzlmod in a dedicated BuildBuddy Workflow to incrementally migrate everything else after.

Create copies of the soci and googleapis patches so that the old patches are kept around for builds triggered from our `internal` repo.

The WORKSPACE.bzlmod file could be compared against the original WORKSPACE file with

```
> diff -u WORKSPACE WORKSPACE.bzlmod
```


.bazelrc
```diff
@@ -1,4 +1,5 @@
 common --noenable_bzlmod
+common --lockfile_mode=off
```
fmeum, line 2: Since `MODULE.bazel.lock` is in `.gitignore`, we can keep the default mode (`update`) for better local incremental build times.

MODULE.bazel
```diff
@@ -0,0 +1,332 @@
...
+    # Do not error if root user is detected
+    ignore_root_user_error = True,
+    python_version = "3.11",
+)
+
+bazel_dep(name = "zlib", version = "1.3")
+bazel_dep(name = "bazel_features", version = "1.4.1", repo_name = "io_bazel_rules_go_bazel_features")
+
+version = use_extension("@io_bazel_rules_go_bazel_features//private:extensions.bzl", "version_extension")
+use_repo(version, "bazel_features_globals", "bazel_features_version")
```
fmeum, line 22: What happens if you don't do this? If this is necessary, it sounds more like a bug in the rules_go setup. Is anything still pulling in rules_go or gazelle in WORKSPACE?

MODULE.bazel
```diff
@@ -0,0 +1,332 @@
...
+        "@@//buildpatches:gazelle.patch",
+    ],
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.35.0/bazel-gazelle-v0.35.0.tar.gz",
+        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.35.0/bazel-gazelle-v0.35.0.tar.gz",
+    ],
+)
+
+go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk")
+go_sdk.download(
```
fmeum, line 40:
```suggestion
# TODO: Remove explicit `goarch`/`goos` after the next rules_go update:
# https://github.com/bazelbuild/rules_go/commit/145b042e21a073fa3637c096bf9779cdbcac6f75
go_sdk.download(
```

MODULE.bazel
```diff
@@ -0,0 +1,332 @@
...
+    patches = [
+        "@io_bazel_rules_go//third_party:com_github_gogo_protobuf-gazelle.patch",
+    ],
+    path = "github.com/gogo/protobuf",
+)
+go_deps.gazelle_override(
+    build_file_generation = "off",
+    path = "github.com/gogo/protobuf",
+)
+go_deps.module_override(
```
fmeum, line 82: You can get `circl` via `bazel_dep(name = "circl")`, which has the patches to get the CGO build to work. If these patches here do anything more than that, we can of course keep it.

WORKSPACE.bzlmod
```diff
@@ -0,0 +1,365 @@
...
+        "https://github.com/google/googletest/archive/f53219cdcb7b084ef57414efea92ee5b71989558.tar.gz",
+    ],
+)
+
+load("@com_google_googletest//:googletest_deps.bzl", "googletest_deps")
+
+googletest_deps()
+
+http_archive(
+    name = "com_google_protobuf",
```
fmeum, line 107: Can we get this via `bazel_dep(name  = "protobuf", repo_name = "com_google_protobuf")` instead?

fmeum (approved):

MODULE.bazel
sluongng: It seems like removing direct deps on `bazel_features` worked for now.

I don't recall what issue I was facing last time, but since then, we have done a rules_go and bazel upgrade which might have fixed things.

Removing this next round.

MODULE.bazel
sluongng: Nice that it was fixed. I was hoping that I could get to it once I merged this PR 👀 

Will add a note next round

MODULE.bazel
sluongng: I will defer this to a later PR.

We are building with both bzlmod turned on and off right now so I don't want to have inconsistency between the 2 modes.

WORKSPACE.bzlmod
sluongng: Similar to circl, I will defer this to a later PR.

Protobuf on BCR does not have the latest version that we use and I am not too sure how we could archive_override while reusing the patches on BCR.

Downgrading to the BCR version "23.1" works, but it's a set back. Perhaps we could contribute 23.1 to BCR after this PR and before the next PR.

bduffany (approved):
I think this could probably be a separate PR (since bzlmod is not enabled everywhere yet?), but do we need to update the auto-sync workflow that copies deps from OSS => Internal?

sluongng:
@bduffany It's still to be determined based on how we enable bzlmod for the `internal` repo.
Right now, it's not yet clear to me if we could do that now, or if we have to wait for some change in Bazel 7.2.0 (the visibility thing).

In the happy case, once we switch to bzlmod, the Go deps should be managed solely via `go.mod` and `MODULE.bazel` files. So we could probably slim down `deps.bzl` one by one until it's gone completely.

fmeum (approved):

PR #5797, opened 2024-02-02
Add workaround for VMs freezing after resuming from snapshot on AMD CPUs

bduffany:
This PR sets the kernel boot option `lapic=notscdeadline` on AMD host CPUs only. I don't fully understand what this does but it allows `TestFirecrackerStressIO` to pass on my AMD machine if I set `maxRunsPerVM = 20`.

I read that LAPIC TSC deadline is meant to improve performance, so disabling it could potentially be bad - but I haven't noticed any performance degradation e.g. when trying a Python app that does 4 hot-loops in separate CPU cores using `multiprocessing`. But anyway, we're only disabling it on AMD which for now only affects local development. I'd rather have the tests pass with some slight performance degradation, than have it be flaky on our dev machines.

I left a TODO for us to remove this flag if the Firecracker folks sort out the root cause of the freezing.

**Related issues**: N/A


tylerwilliams (approved):

maggie-lou:
Nice! Good sign that they're narrowing in on a fix

PR #5798, opened 2024-02-02
buildpatches: minor fixes to patch files

sluongng:
These fixes are backport from #5792. Expect no-ops.

soci_snapshotter should not need the MODULE.bazel and WORKSPACE.bzlmod
file to build under bzlmod. Removing the hunks to simplify the patch.

Fix space identation in vtprotobuf patch.


luluz66 (approved):
vtprotobuf.patch looks good to me.

tempoz (approved):

PR #5800, opened 2024-02-02
bazelisk: upgrade to 1.19.0

sluongng:
Backported from #5792.

Because of https://github.com/bazelbuild/bazelisk/pull/462,
the way `.bazeliskrc` and related environment variables are layered over
each others have changed.

New package `config` was introduce for this reason, and functions that
depend on Bazelisk configs is expected to propagate config.Config struct
around.


sluongng:
Tbh, I am not quite sure about my fix for this as I am not that familiar with our `cli/bazelisk`. 👀

There could be call sites that are trying to `os.GetEnv` directly without going through the config package.

bduffany:
The fix looks good, but could you expand on what you mean by this?

> There could be call sites that are trying to os.GetEnv directly without going through the config package.

sluongng:
Im gona need to look at this a bit closer because we have calls like this

```bash
> rg os.Getenv cli/bazelisk
cli/bazelisk/bazelisk.go
105:    return filepath.Base(os.Args[0]) == "bazelisk" || os.Getenv("BAZELISK_SKIP_WRAPPER") == "true"
178:    if v := os.Getenv("BB_USE_BAZEL_VERSION"); v != "" {
181:    return os.Getenv("USE_BAZEL_VERSION")
```

specifically the last call, I want to make sure that we support `.bazeliskrc` as well as `.bazelversion` 🤔 

Or we could just ignore it and say `.bazeliskrc` has never been supported and we could addressed it in a separate PR.

The remaining call site that could get affected is our debug logs here

```bash
cli/cmd/bb/bb.go
57:     log.Debugf("env: BAZEL_REAL=%s", os.Getenv("BAZEL_REAL"))
58:     log.Debugf("env: BAZELISK_SKIP_WRAPPER=%s", os.Getenv("BAZELISK_SKIP_WRAPPER"))
59:     log.Debugf("env: USE_BAZEL_VERSION=%s", os.Getenv("USE_BAZEL_VERSION"))
60:     log.Debugf("env: BB_USE_BAZEL_VERSION=%s", os.Getenv("BB_USE_BAZEL_VERSION"))
```

Im not sure how valuable these are just yet 🤔

sluongng:
Ok given that we have this comment

```
	// TODO: Handle the cases where we were invoked via .bazeliskrc
	// or USE_BAZEL_FALLBACK_VERSION (less common).
```

I will defer supporting `.bazeliskrc` to later.

With the new bazelisk's `config` package supporting multiple layers of configuration overlaying each other, I think it should be easier to do if we have a use case for it.

bduffany (approved):
Yeah, we don't have unit tests for .bazeliskrc, so we may as well call it unsupported for now. LGTM

PR #5803, opened 2024-02-02
googleapis: build status_go_proto with vtprotobuf compiler

sluongng:
Instead of maintaining our own status_go_proto, let's patch the
existing target inside googleapis and use it.

This let us remove custom Gazelle directives in our go_repository.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -116,6 +116,10 @@ gazelle_dependencies()
 
 http_archive(
     name = "googleapis",
+    patch_args = ["-p1"],
+    patches = [
+        "//buildpatches:googleapis.patch",
+    ],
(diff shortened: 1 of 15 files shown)


sluongng:
This was backported from https://github.com/buildbuddy-io/buildbuddy/pull/5792

The motivation is explained https://github.com/bazelbuild/bazel/issues/19301#issuecomment-1923527294 where bzlmod applied some new visibility setting that prevents targets in `go_repository` repos depending on our `//proto:grpc_status_go_proto` target. A workaround is available by using the undocumented target syntax `@@//proto:grpc_status_go_proto`, but I think this change is overall cleaner for us to maintain.

luluz66 (approved):

PR #5806, opened 2024-02-02
Copy `jvm.out` to the workflow artifacts directory when Bazel OOMs or internal errors

siggisim:

PR #5810, opened 2024-02-03
Fix jvm.out error logging

bduffany:
my suggestion [here](https://github.com/buildbuddy-io/buildbuddy/pull/5806#discussion_r1476859302) was wrong lol

(Also, removed the logging in the success case, because we already have a log message for each file that's uploaded, which displays at the end of the workflow run)

**Related issues**: N/A


siggisim (approved):
Haha! I was just about to send you this
https://github.com/buildbuddy-io/buildbuddy/compare/master...siggi-dev-branch-20240202-172541?body=&expand=1

siggisim:
Were you able to get a successful upload? I managed to get a workflow to oom in dev by adding the startup option `--host_jvm_args=-Xmx5m`, and I saw the message print - but I didn't get an artifact uploaded: https://app.buildbuddy.dev/invocation/e87050dd-55d3-4758-8fcf-b856eb1ee165

Was going to debug some more next week.

bduffany:
@siggisim the only thing I've tested so far was commenting out the conditional check (exit code) and then running the `ci_runner_test` case that checks for artifacts being uploaded. With the conditional commented out, the test fails because `jvm.out` got uploaded unexpectedly (which is what we'd want in that case).

PR #5811, opened 2024-02-03
go.mod: add dependencies for go tooling

sluongng:
This is a change backported from #5792.

Under bzlmod, there will be no `deps.bzl` file.
The `go_repository` repos are automatically generated with gazelle's
`go_deps` extension parsing `go.mod` and `go.sum` files.
This let us consolidate go dependency management to a single source of
truth that is consistent with our editor experience.

The downside of this approach is that we should not be adding
`go_repository` declarations manually, but instead, adding them to
`go.mod` file by calling `go mod tidy -e`.
Create a dummy Go source file that imports all the Go tools dependencies
and guard it with a Go build directive that will not be used.

### DIFF
--- a/tools/deps/deps.go
+++ b/tools/deps/deps.go
@@ -1,0 +1,34 @@
+//go:build tools
+
+// This tools/deps package keep track of Go tooling dependency that we use
+// but do not depend directly in code.
+//
+// Importing them here helps nudge "go mod tidy" to include them inside
+// "go.mod" and "go.sum" files, which are used as source of truth for rules_go
+// and gazelle under bzlmod.
+//
+// The package is shielded with a "go:build tools" directive.
+// Without a specific flag, gazelle should NOT generate go targets for this
+// package.
+package deps
+
+import (
+	// Device manager
+	_ "gitlab.com/arm-research/smarter/smarter-device-manager"
+
+	// Gazelle deps for 'bb fix'
+	_ "github.com/pmezard/go-difflib/difflib"
+
+	// rules_webtesting deps
+	_ "github.com/gorilla/mux"
+
+	// static analyzers
+	_ "github.com/nishanths/exhaustive"
+	_ "golang.org/x/tools/go/analysis"
+	_ "honnef.co/go/tools/staticcheck"
+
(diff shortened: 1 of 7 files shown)


tools/deps/deps.go
```diff
@@ -0,0 +1,40 @@
+//go:build tools
+
+// This tools/deps package keep track of Go tooling dependency that we use
+// but do not depend directly in code.
```
bduffany, line 4: what sorts of tool deps need to be included here? i.e. how did you build this list of deps below? (Is there a straightforward way of automating this?)

sluongng: There are 3 groups added here:

- Go deps we depend on directly in our BUILD files but not our Go files. I.e. `gitlab.com/arm-research/smarter/smarter-device-manager`

- Transitive Go deps that our `http_archive` repo use. For example, in WORKSPACE mode, we used to load some Go deps with
```python
load("@com_github_sluongng_nogo_analyzer//staticcheck:deps.bzl", "staticcheck")

staticcheck()
```

This is no longer possible because of the new way Gazelle operate in Bzlmod

```
ERROR: /Users/sluongng/work/buildbuddy/buildbuddy/BUILD:79:5: //:vet_actual depends on @@org_golang_x_tools//go/analysis/passes/unusedresult:unusedresult in repository @@org_golang_x_tools which failed to fetch. no such package '@@org_golang_x_tools//go/analysis/passes/unusedresult': Unable to load package for @@bazel_gazelle_go_repository_cache//:go.env: The repository '@@bazel_gazelle_go_repository_cache' could not be resolved: Repository '@@bazel_gazelle_go_repository_cache' is not defined
```

- Our transitive proto/grpc deps that we used to add to `server/libmain/libmain.go` to help with `go mod tidy -e` not adding them from `proto/**` generated code.

---

These were pretty much manually picked from the errors I encountered #5792.
So maintaining this list would have to be a manual effort for now.

The UX in Bzlmod is documented here https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/bzlmod.md#depending-on-tools.

bduffany (approved):

PR #5812, opened 2024-02-04
runfiles refactoring

sluongng:
This PR is structured in multiple sets of commits for easier review.
All of these changes are refactoring and thus, there is no functional change expected.

* Commit 1: `patch rules_go`
  As part of working on this PR, an issue was found with rules_go's go_test not including sufficient `data` targets while compiling the `external` test package. A patch was created and merged in rules_go to fix this. However, because these has been no new release, export the patch from rules_go to use it in our repo.

* Commit 2: `s/bazel(|go).Runfile/runfiles.Rlocation/`
  Refactor usage of the deprecated Runfile API with the newer Rlocation calls.
  Runfiles paths are updated to add the needed repo prefix `buildbuddy/`.

* Commit 3-15: use `$(rlocationpath ...)`
  This series of patches moved the hardcoded `buildbuddy/...` runfile paths to variables.
  The variables are set at build time via rules_go's `x_defs` attribute.
  The values are generated from `$(rlocationpath <target>)` of Bazel.
  Included minor cleanups to ensure `data` is declared in the right package.

This unblocks the bzlmod migration in #5792

### DIFF
--- a/cli/testutil/testcli/BUILD
+++ b/cli/testutil/testcli/BUILD
@@ -7,6 +7,9 @@ go_library(
     data = ["//cli/cmd/bb"],
     importpath = "github.com/buildbuddy-io/buildbuddy/cli/testutil/testcli",
     visibility = ["//visibility:public"],
+    x_defs = {
+        "bbRunfilePath": "$(rlocationpath //cli/cmd/bb)",
+    },
     deps = [
         "//cli/storage",
         "//server/testutil/testbazel",
--- a/enterprise/config/BUILD
+++ b/enterprise/config/BUILD
@@ -10,7 +10,7 @@ filegroup(
             "buildbuddy.release.yaml",
             "executor.release.yaml",
         ],
-        "//conditions:default": glob(["**"]),
+        "//conditions:default": glob(["*.yaml"]),
     }),
 )
 
--- a/enterprise/server/testutil/buildbuddy_enterprise/buildbuddy_enterprise.go
+++ b/enterprise/server/testutil/buildbuddy_enterprise/buildbuddy_enterprise.go
@@ -28,9 +28,11 @@ var (
 	remoteSSOSlug     = flag.String("remote_sso_slug", "", "For remote targets, the SSO slug to login. Self-auth must be enabled for these targets.")
 )
 
-const (
-	DefaultConfig = "buildbuddy/enterprise/config/test/buildbuddy.selfauth.yaml"
-	NoAuthConfig  = "buildbuddy/enterprise/config/test/buildbuddy.noauth.yaml"
+var (
+	// set by x_defs in BUILD file
+	DefaultConfig         string
+	NoAuthConfig          string
+	buildbuddyRunfilePath string
 )
 
 func Run(t *testing.T, args ...string) *app.App {
@@ -48,7 +50,7 @@ func RunWithConfig(t *testing.T, appConfig *app.App, configPath string, args ...
 	return app.RunWithApp(
 		t,
 		appConfig,
-		/* commandPath= */ "buildbuddy/enterprise/server/cmd/server/buildbuddy_/buildbuddy",
+		/* commandPath= */ buildbuddyRunfilePath,
 		commandArgs,
 		/* configPath= */ configPath,
 	)
--- a/enterprise/server/testutil/buildbuddy_enterprise/buildbuddy_enterprise.go
+++ b/enterprise/server/testutil/buildbuddy_enterprise/buildbuddy_enterprise.go
@@ -29,8 +29,8 @@ var (
 )
 
 const (
-	DefaultConfig = "enterprise/config/test/buildbuddy.selfauth.yaml"
-	NoAuthConfig  = "enterprise/config/test/buildbuddy.noauth.yaml"
+	DefaultConfig = "buildbuddy/enterprise/config/test/buildbuddy.selfauth.yaml"
+	NoAuthConfig  = "buildbuddy/enterprise/config/test/buildbuddy.noauth.yaml"
 )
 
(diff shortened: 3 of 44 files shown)


cli/testutil/testcli/BUILD
```diff
@@ -7,6 +7,9 @@ go_library(
     data = ["//cli/cmd/bb"],
     importpath = "github.com/buildbuddy-io/buildbuddy/cli/testutil/testcli",
     visibility = ["//visibility:public"],
+    x_defs = {
+        "BBRunfilePath": "$(rlocationpath //cli/cmd/bb)",
```
bduffany, line 11: nit: lowercase `BB` to make this private?

`bbRunfilePath`

(here and in the .go file)

enterprise/server/test/integration/remote_cache/remote_cache_test.go
```diff
@@ -239,7 +239,7 @@ func TestBuild_RemoteCacheFlags_Compression_SecondBuildIsCached(t *testing.T) {
 
 func TestBuild_RemoteCache_ScoreCard(t *testing.T) {
 	app := buildbuddy_enterprise.RunWithConfig(
-		t, buildbuddy_enterprise.DefaultAppConfig(t), buildbuddy_enterprise.NoAuthConfig,
+		t, buildbuddy_enterprise.DefaultAppConfig(t), buildbuddy_enterprise.NoAuthConfigRunfilePath,
```
bduffany, line 242: imo we should keep these callsites the same, since `RunfilePath` seems like more of an implementation detail of the package.

i.e. change `NoAuthConfigRunfilePath` => `NoAuthConfig`, and change the current `NoAuthConfigName` => `noAuthConfigName` (make it private). (Same for the default config)

enterprise/server/testutil/buildbuddy_enterprise/buildbuddy_enterprise.go
```diff
@@ -29,12 +30,36 @@ var (
...
+	for _, p := range strings.Split(configFilesRunfilePaths, " ") {
+		if strings.HasSuffix(p, DefaultConfigName) {
+			DefaultConfigRunfilePath = p
+			continue
+		}
+		if strings.HasSuffix(p, NoAuthConfigName) {
+			NoAuthConfigRunfilePath = p
+			continue
+		}
+	}
```
bduffany, line 58: To simplify this a little, could we export the test config files as individual BUILD targets, rather than including them in the `config_files` glob? The `config_files` target gets distributed with our release binaries so it'd be nice to remove the testonly configs from that glob pattern anyway.

bduffany (commented):
This feels like a nice change overall - feels like it gives a little more safety. If we change the runfiles paths then it will break at build time now, rather than at run time.

Just a couple nits but otherwise looks good :+1:

enterprise/server/testutil/buildbuddy_enterprise/buildbuddy_enterprise.go
sluongng: I didn't like this pattern as well. I worked on https://github.com/bazelbuild/rules_go/pull/3852 as an alternative solutions instead. I think in the future we gonna have use cases for `rlocationpaths` on multiple files so it would be nice to have it ready.

I don't mind adding a `BUILD` file inside `enterprise/config/test`. Let me try that.

sluongng:
@bduffany PTAL 🙏

enterprise/config/BUILD
```diff
@@ -10,7 +10,10 @@ filegroup(
             "buildbuddy.release.yaml",
             "executor.release.yaml",
         ],
-        "//conditions:default": glob(["**"]),
+        "//conditions:default": glob(["*.yaml"]) + [
+            "//enterprise/config/test:buildbuddy.noauth.yaml",
+            "//enterprise/config/test:buildbuddy.selfauth.yaml",
+        ],
```
bduffany, line 16: nit: I think we should probably have the tests which need these configs explicitly depend on them, instead of including them in the config_files target. It feels a little weird to include test configs in a filegroup target which has other non-test files as well.

bduffany (approved):

enterprise/server/testutil/buildbuddy_enterprise/buildbuddy_enterprise.go
```diff
@@ -29,8 +29,15 @@ var (
 )
 
 const (
-	DefaultConfig = "enterprise/config/test/buildbuddy.selfauth.yaml"
-	NoAuthConfig  = "enterprise/config/test/buildbuddy.noauth.yaml"
+	defaultConfigName = "test/buildbuddy.selfauth.yaml"
+	noAuthConfigName  = "test/buildbuddy.noauth.yaml"
+)
```
bduffany, line 34: we can delete these now right?

sluongng: yup. Will clean up.

enterprise/config/BUILD
sluongng: tbh, I don't fully understand the BUILD file logic in this `enterprise/config` dir.

It seems like we could remove `:bundle` entirely since, as well as change the `filegroup `into `exports_files` and force other targets to explicitly depend on the right target.

So let's do this instead: I will remove the test config files from the filegroup in this PR. And I will investigate removing the filegroup and `:bundle` in a separate PR. WDYT?

bduffany: Sounds good, although I don't think removing the filegroup or bundle target is needed so just removing the test config files seems good enough for now.

sluongng: SGTM

PR #5814, opened 2024-02-05
Fix jvm.out upload

bduffany:
Make sure we upload artifacts before returning from the function due to the bazel error.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -984,6 +984,11 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


bduffany:
@siggisim FYI `jvm.out` only had a single line in this case

`Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "main"`

idk how useful this is? or maybe it'll contain more useful stuff in a less contrived example where we aren't just setting a 5m memory limit?

siggisim (approved):
Ah nice! Yeah, I think we probably only want jvm.out for Bazel Internal errors and heap dump for OOMs.

bduffany:
k, I can send a separate PR to also get the heap dump.

PR #5816, opened 2024-02-05
[SS] Add tool to upload local snapshot to remote cache

maggie-lou:
This will make it easier for us to debug local snapshot failures, because we can then use vmstart to start the VM from the remote cache


maggie-lou:
* `master`
  * **PR #5816** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5816?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5816?utm_source=stack-comment).

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func getToolEnv() *real_environment.RealEnv {
+	healthChecker := healthcheck.NewHealthChecker("upload-local-snapshot")
+	re := real_environment.NewRealEnv(healthChecker)
+
+	// Set a large file cache size so this tool doesn't evict anything from
+	// the executor's filecache
+	// This is set to match the prod file cache size
+	fcSize := int64(1_250_000_000_000)
```
bduffany, line 111: let's set it much larger to avoid evicting even if we change the prod cache size

```suggestion
	fcSize := int64(1e18)
```

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
```diff
@@ -0,0 +1,274 @@
...
+	if *apiKey != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-api-key", *apiKey)
+	}
+	env := getToolEnv()
+	loader, err := snaploader.New(env)
+	if err != nil {
+		log.Fatalf("Failed to init snaploader: %s", err)
+	}
+
+	allSnapshotDigests := getAllDigestsFromSnapshotManifest(ctx, env, snapshotInstanceName, snapshotGroupID, key)
```
bduffany, line 81: we should probably return the manifest here so that `uploadManifest` uploads the same manifest that we read here (since the manifest stored at the FC key might change while we're uploading)

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
```diff
@@ -0,0 +1,274 @@
...
+		Hash:      hashStrings(kd.GetHash(), ".manifest"),
+		SizeBytes: 1, /*arbitrary size*/
+	}
+	acDigest := digest.NewResourceName(remoteManifestKey, remoteInstanceName, rspb.CacheType_AC, repb.DigestFunction_BLAKE3)
+	if err := cachetools.UploadActionResult(ctx, env.GetActionCacheClient(), acDigest, localManifestACResult); err != nil {
+		log.Fatalf("Error uploading manifest to remote cache: %s", err)
+	}
+}
+
+func readLocalManifestACResult(ctx context.Context, env environment.Env, key *fcpb.SnapshotKey, groupID string) *repb.ActionResult {
```
bduffany, line 256: nit: can we call the func from snaploader instead? just want to avoid the logic going out of sync (these tools tend to break a lot, just want to minimize the amount of maintenance we have to do)

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
```diff
@@ -0,0 +1,274 @@
...
+	localManifestACResult := readLocalManifestACResult(ctx, env, key, groupID)
+
+	kd, err := digest.ComputeForMessage(key, repb.DigestFunction_SHA256)
+	if err != nil {
+		log.Fatalf("Error generating digest for snapshot key: %s", err)
+	}
+	remoteManifestKey := &repb.Digest{
+		Hash:      hashStrings(kd.GetHash(), ".manifest"),
+		SizeBytes: 1, /*arbitrary size*/
+	}
```
bduffany, line 249: is there some func in snaploader we can call directly here instead?

bduffany (approved):
LGTM just nits

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
maggie-lou: Good call - done

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
maggie-lou: Done

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
maggie-lou: Made the RemoteManifestKey function public and used that here

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
maggie-lou: Done. Did some refactoring to reuse more logic

enterprise/tools/upload_local_snapshot/upload_local_snapshot.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func uploadManifestRemoteCache(ctx context.Context, env environment.Env, key *fcpb.SnapshotKey, localACResult *repb.ActionResult, remoteInstanceName string) {
+	remoteManifestKey, err := snaploader.RemoteManifestKey(key)
+	if err != nil {
+		log.Fatalf("Error generating digest for snapshot key: %s", err)
+	}
+	acDigest := digest.NewResourceName(remoteManifestKey, remoteInstanceName, rspb.CacheType_AC, repb.DigestFunction_BLAKE3)
+	if err := cachetools.UploadActionResult(ctx, env.GetActionCacheClient(), acDigest, localACResult); err != nil {
+		log.Fatalf("Error uploading manifest to remote cache: %s", err)
```
bduffany, line 230: (no need to change this PR, just a small note for the future) log.Fatal prevents deferred cleanup funcs from being run (e.g. the temp dir cleanup) since it causes the process to immediately exit. As a workaround I like to use this boilerplate for Go binaries:

```
func main() {
  if err := run(); err != nil {
    log.Fatal(err.Error())
  }
}

func run() error {
  // ...
  if err != nil {
    return err
  }
  // ...
}
```

bduffany (approved):
LGTM

PR #5818, opened 2024-02-05
Add .bazelrc config for deflaking tests

bduffany:
I have this config in my `user.bazelrc` but it will be helpful to have a common config that we can mention in "flaky test" bug reports instead of having to list out all the bazel flags we're using.

e.g. "I can no longer repro the flakiness with `bazel test //target --config=deflake-go` so I'm closing this bug"

**Related issues**: N/A


sluongng (approved):
TIL test.failfast.

LGTM

PR #5819, opened 2024-02-05
Proto changes for stepped workflows

bduffany:
## Context

Here is the workflow UX that we are aiming for (roughly - we discussed moving the buttons into the 3-dot menu, and we'll also probably want some way to search the logs across steps. The search part in particular is challenging but I think we can add it later without needing to change the protocol introduced in this PR):

image

The key aspects of this UX that we'd like to support are:
* Workflow logs are split into discrete "steps"
* Each step has some metadata attached to it
  * name
  * status (not started, in-progress, success, or error)
  * duration
* Workflow steps no longer have to be bazel commands, and can be arbitrary bash scripts.
  * we'll automatically discover any "child" invocations created in each step, and render nice invocation cards in the step's logs

## Stepped logs

This PR adds a `step_id` field to the ProgressEventId proto. This will be used to write the output from each workflow step to separate logs in blobstore.

For now, there is a caveat that the server only needs to support one step at a time and can return an error if the client tries to tack on more logs to a previous step. I figured this should help simplify the implementation for now, so that we can avoid keeping track of multiple log streams on the server (i.e. we can avoid introducing a stream ID => log map with N entries). Later we can relax this restriction if we need to for whatever reason, but it seems unlikely that it will  be needed.

## Workflow step metadata/results and child invocation discovery

Add protos for workflow step started / completed. In the new "steps" world, we'll just have plain old steps instead of bazel commands, so the existing ChildInvocationsConfigured approach won't totally work (we won't know the invocations ahead of time).

In order to discover child invocations, the CI runner can generate a script which appears as `bazel` in `$PATH`. The script will look like this (pseudo-code):

```
- Generate an invocation ID
- Ping the parent CI runner (over unix socket), telling it that a new child invocation has started.
  The CI runner will publish a ChildInvocationsConfigured event
- exec /path/to/real/bazel <args ...> --invocation_id=<generated_invocation_id>
```

Or a much simpler approach would be to just scan the logs for "Streaming build logs to <url>". TBH, we should probably start with this simpler approach and only switch to something more complicated if this has issues.

**Related issues**: N/A

### DIFF
--- a/proto/build_event_stream.proto
+++ b/proto/build_event_stream.proto
@@ -53,6 +53,18 @@ message BuildEventId {
     // assigned; the only meaningful operation on this field is test for
     // equality.
     int32 opaque_count = 1;
+
+    // BUILDBUDDY-SPECIFIC FIELDS BELOW.
+
+    // The invocation step ID. This can be used to group progress streams into
+    // discrete "steps". At most one step's stream may be open at once - the
+    // server may finalize a stream when transitioning from one stream ID to
+    // another, and may return an error if attempting to resume a previously
+    // finalized stream.
+    //
+    // BuildBuddy will use this ID to render each step's output for a workflow
+    // invocation.
+    int32 step_id = 1000;
   }
 
   // Identifier of an event indicating the beginning of a build; this will
@@ -274,7 +286,14 @@ message BuildEventId {
 
   // Identifier of an event providing a list of commands that are
   // run as children of this invocation.
-  message ChildInvocationsConfiguredId {}
+  message ChildInvocationsConfiguredId {
+    // The workflow step number in which the child invocation(s) occurred.
+    int32 step_number = 1;
+
+    // A unique ID, required if multiple child invocations appear in the same
+    // step.
+    int32 opaque_count = 2;
+  }
 
   // Identifier of an event providing the status of a completed child
   // invocation.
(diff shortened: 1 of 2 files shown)


proto/build_event_stream.proto
```diff
@@ -1464,6 +1514,9 @@ message ChildInvocationsConfigured {
 
   // Child Bazel invocations.
   repeated InvocationMetadata invocation = 1;
+
+  // The workflow step in which these child invocations appeared.
+  int32 step_id = 2;
```
maggie-lou, line 1519: inconsistent with step_number in ChildInvocationsConfiguredId

proto/build_event_stream.proto
```diff
@@ -53,6 +53,18 @@ message BuildEventId {
...
+
+    // The invocation step ID. This can be used to group progress streams into
+    // discrete "steps". At most one step's stream may be open at once - the
+    // server may finalize a stream when transitioning from one stream ID to
+    // another, and may return an error if attempting to resume a previously
+    // finalized stream.
+    //
+    // BuildBuddy will use this ID to render each step's output for a workflow
+    // invocation.
+    int32 step_id = 1000;
```
maggie-lou, line 67: Is there a reason why this is an id while the other messages use step_number? It might be easier to string messages together if they are consistent

bduffany:
Will reopen later once I resume work on this.

PR #5822, opened 2024-02-06
WORKSPACE: rename workspace

sluongng:
This an alternative to https://github.com/buildbuddy-io/buildbuddy-internal/pull/3102.

In preparation for bzlmod migration, we will need a consistent workspace
name for 'internal' usage.  This affects the runfile path look up that
will be updated in #5812.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,5 @@
 workspace(
-    name = "buildbuddy",
+    name = "com_github_buildbuddy_io_buildbuddy",
 )
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
--- a/deps.bzl
+++ b/deps.bzl
@@ -2,7 +2,7 @@ load("@bazel_gazelle//:deps.bzl", "go_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
 
 # bazelisk run //:gazelle -- update-repos -from_file=go.mod -to_macro=deps.bzl%install_go_mod_dependencies
-def install_go_mod_dependencies(workspace_name = "buildbuddy"):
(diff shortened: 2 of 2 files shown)


bduffany (commented):
This change feels somewhat likely to break stuff - can you try building the internal docker image etc. and make sure everything is fine?

siggisim:
Not a huge fan of these ugly `com_github_buildbuddy_io_buildbuddy` style names

bduffany:
to keep the `buildbuddy` name while also addressing the concern from the other concern about not breaking existing usages of `--override_repository=com_github_buildbuddy_io_buildbuddy`, is it possible to define some sort of "repository alias" so that those usages will still work?

siggisim:
If this is a bzlmod thing, then you can set a different `repo_name` i.e.:
```
bazel_dep(name = "rules_apple", version = "3.2.1", repo_name = "build_bazel_rules_apple")
```

sluongng:
This PR was motivated to support https://github.com/buildbuddy-io/buildbuddy/pull/5812 where we adopt the new runfile look-up approach from Bazel + rules_go with this.

```diff
-	return testfs.RunfilePath(t, "cli/cmd/bb/bb_/bb")
+	return testfs.RunfilePath(t, "buildbuddy/cli/cmd/bb/bb_/bb")
```

the problematic part was the `buildbuddy/` prefix was determined by the repo's name.

Because we change our repo name between `@buildbuddy` and `@com_github_buildbuddy_io_buildbuddy` depending on the build context (internal or external repo), we can't maintain this `buildbuddy/` prefix inside `*.go` code.

I recently learned that we could generate the string `"buildbuddy/cli/cmd/bb/bb_/bb"` or `"com_github_buildbuddy_io_buildbuddy/cli/cmd/bb/bb_/bb"` dynamically at build time using `$(rlocationpath //cli/cmd/bb)` in BUILD file. This allows us from keeping the workspace name as is.

I have sent a set of patches to #5812 to do just that.

PR #5824, opened 2024-02-06
Revert "bazelisk: use Bazelisk built from source (#5748)"

bduffany:
Getting this error when running workflows in firecracker locally:

/root/workspace/bazelisk: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /root/workspace/bazelisk)

The error goes away with the revert applied.

The bazelisk we were previously using was statically linked. Based on the error, it looks like it's no longer getting statically linked when building from source, so we run into glibc versioning issues :/

**Related issues**: N/A

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -6737,6 +6737,24 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
         urls = ["https://github.com/bazelbuild/bazel/releases/download/6.0.0/bazel-6.0.0-linux-x86_64"],
     )
     http_file(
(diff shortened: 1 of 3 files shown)


sluongng (approved):

sluongng:
From https://github.com/bazelbuild/bazelisk/blob/0c6620a7a02939e6bdd0c6af420d4c6179de6373/BUILD#L69-L101

It looks like the release binary is always built with `pure = "on",`, which applies a transition on the binary target to force a no CGO compilation.

We could use that in our `select` instead 🤔 

Do you know if we could test for this in our "baremetal" test setup somehow?

PR #5840, opened 2024-02-07
impersonation: redirect to subdomain from invocations

vadimberezniker:
This will allow us to scope the impersonation cookie to the subdomain.

Part of addressing https://github.com/buildbuddy-io/buildbuddy-internal/issues/3079

**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation_not_found.tsx
+++ b/app/invocation/invocation_not_found.tsx
@@ -13,20 +13,51 @@ interface Props {
   user?: User;
 }
 
-export default class InvocationNotFoundComponent extends React.Component<Props> {
-  handleLoginClicked() {
-    authService.login();
+interface State {
+  impersonationGroupID?: string;
+}
+
+export default class InvocationNotFoundComponent extends React.Component<Props, State> {
+  state: State = {};
+
+  isPermissionDenied: boolean;
+  canImpersonate: boolean;
+
+  constructor(props: Props) {
+    super(props);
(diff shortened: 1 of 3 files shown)


siggisim (approved):

app/invocation/invocation_not_found.tsx
```diff
@@ -13,22 +13,53 @@ interface Props {
...
+export default class InvocationNotFoundComponent extends React.Component<Props, State> {
+  state: State = {};
+
+  isPermissionDenied: boolean;
+  canImpersonate: boolean;
+
+  constructor(props: Props) {
+    super(props);
+    this.isPermissionDenied = props.error?.code === "PermissionDenied";
+    this.canImpersonate = props.user?.canCall("getInvocationOwner") || false;
```
bduffany, line 29: nit: it's generally not desired to have component-level state that is derived from props, because react components are expected to re-render if their props change (and the state would have to be manually kept in sync). Instead of having state - for `this.canImpersonate` we can replace that with `this.props.user?.canImpersonate()` and for `isPermissionDenied` we can have a method like `isPermissionDenied()` (or check the error code inline)

Not a huge deal in this case since this component's props are unlikely to change, but sticking with this convention should help minimize bugs/surprises later.

bduffany (approved):

app/invocation/invocation_not_found.tsx
vadimberezniker: done

bduffany (approved):

PR #5841, opened 2024-02-07
rules_go: upgrade to v0.45.1

sluongng:
Also upgrade Go SDK to 1.21.7.
Ensure that the versions of gazelle and rules_go inside go.mod is
consistent with the versions we are using in WORKSPACE.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/buildbuddy-io/buildbuddy
 
-go 1.21
+go 1.21.7
```
sluongng, line 3: Without this, I would get

```
diff --git a/go.mod b/go.mod
index 7b4b50bc5..744723379 100644
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/buildbuddy-io/buildbuddy

-go 1.21.7
+go 1.21.4
+
+toolchain go1.21.7

 replace (
        github.com/awslabs/soci-snapshotter => github.com/buildbuddy-io/soci-snapshotter v0.0.8 // keep in sync with buildpatches/com_github_awslabs_soci_snapshotter.patch
```

when running `bazel run //:go -- mod tidy -e`. So I pin it to the specific patch version.

bduffany:
(friendly reminder to update the internal repo dep too)

sluongng:
> (friendly reminder to update the internal repo dep too)

good call. Sent a PR for that.

vadimberezniker (approved):

PR #5843, opened 2024-02-07
prometheus/client_model: move _go_proto target to the external repo

sluongng:
Instead of maintaining a custom target in our repo, simply nudge
Gazelle to generate the _go_proto target with the right compilers
in the `prometheus/client_model` repo.

This let us get rid of additional build directives needed for other
dependencies inside `deps.bzl` and make our setup friendlier to the
future bzlmod migration.


sluongng:
This is a bad idea as the result will require us adding
```
gazelle:resolve go github.com/prometheus/client_model/go @com_github_prometheus_client_model//io/prometheus/client:go
```

everywhere. Closing this for an alternative solution.

PR #5844, opened 2024-02-07
Add new roles to role proto

bduffany:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3091

### DIFF
--- a/enterprise/app/org/org_members.tsx
+++ b/enterprise/app/org/org_members.tsx
@@ -39,11 +39,16 @@ type State = {
   isRemoveLoading?: boolean;
 };
 
-const ROLE_LABELS: Record<grp.Group.Role, string> = {
-  [grp.Group.Role.UNKNOWN_ROLE]: "",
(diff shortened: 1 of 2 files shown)


enterprise/app/org/org_members.tsx
```diff
@@ -39,11 +39,16 @@ type State = {
   isRemoveLoading?: boolean;
 };
 
-const ROLE_LABELS: Record<grp.Group.Role, string> = {
-  [grp.Group.Role.UNKNOWN_ROLE]: "",
-  [grp.Group.Role.ADMIN_ROLE]: "Admin",
-  [grp.Group.Role.DEVELOPER_ROLE]: "Developer",
-};
+function getRoleLabel(role: grp.Group.Role): string {
```
bduffany, line 42: the typechecker didn't appreciate the missing entries for the new role values so I updated this to make the typechecker not break when we add new roles.

siggisim (approved):

PR #5851, opened 2024-02-07
fix a pebble race condition

luluz66:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3098
In FindMissing, we locked the keys after creating the iterator. This can cause
race condition and is also manifested in the flakiness TestMigrateVersions.

In this PR, I locked all the keys present in the request prior to creating the iterator.  When CDC is enabled, I created a new iterator after I locked all the keys in the chunked metadata.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1601,43 +1601,81 @@ func (p *PebbleCache) FindMissing(ctx context.Context, resources []*rspb.Resourc
 	}
 	defer db.Close()
 
+	lockedResources, err := p.lockResourceKeys(ctx, resources)
+	if err != nil {
+		return nil, err
+	}
 	iter := db.NewIter(nil /*default iterOptions*/)
 	defer iter.Close()
 
 	var missing []*repb.Digest
-	for _, r := range resources {
-		err = p.findMissing(ctx, iter, r)
+	for _, lockedResource := range lockedResources {
+		err = p.findMissing(ctx, db, iter, lockedResource.key)
 		if err != nil {
-			missing = append(missing, r.GetDigest())
+			missing = append(missing, lockedResource.resource.GetDigest())
 		}
+		lockedResource.unlockFn()
 	}
 	return missing, nil
 }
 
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1601,44 +1601,82 @@ func (p *PebbleCache) FindMissing(ctx context.Context, resources []*rspb.Resourc
...
-	if err != nil {
-		return err
+type resourceLockInfo struct {
+	key      filestore.PebbleKey
+	unlockFn func()
+	resource *rspb.ResourceName
+}
+
+func (p *PebbleCache) lockResourceKeys(ctx context.Context, resources []*rspb.ResourceName) ([]*resourceLockInfo, error) {
+	res := make([]*resourceLockInfo, 0, len(resources))
```
vadimberezniker, line 1629: the keys would have to be sorted to avoid deadlocks
you could have 2 requests, one for `a, b` and one for `b, a`. if both locked their first digest first, neither would be able to make progress.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1601,44 +1601,82 @@ func (p *PebbleCache) FindMissing(ctx context.Context, resources []*rspb.Resourc
 	}
 	defer db.Close()
 
+	lockedResources, err := p.lockResourceKeys(ctx, resources)
+	if err != nil {
+		return nil, err
+	}
```
vadimberezniker, line 1607: I don't recall but do we have data on how much memory/cpu we save by creating a single iterator per FindMissing call? We could avoid this complexity if we didn't need to have a single iterator for the lifetime of the FindMissing call.

tylerwilliams: Could we use https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/util/pebble/pebble.go#L575 instead?

luluz66: See this PR: https://github.com/buildbuddy-io/buildbuddy/pull/5877

luluz66:
Close this PR and see https://github.com/buildbuddy-io/buildbuddy/pull/5877 for an alternative solution.

PR #5853, opened 2024-02-08
SCIM: refactor user operations to operate by SubID instead of relying on group auth.

vadimberezniker:
This allows users to be synced via SCIM w/o having to add the users to a group first.

The users can then be added separately to different groups via group SCIM ops.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/userdb/userdb.go
+++ b/enterprise/server/backends/userdb/userdb.go
@@ -807,37 +807,21 @@ func (d *UserDB) GetUserByIDWithoutAuthCheck(ctx context.Context, id string) (*t
 	return user, err
 }
 
-func (d *UserDB) GetUserByEmail(ctx context.Context, email string) (*tables.User, error) {
-	auth := d.env.GetAuthenticator()
-	if auth == nil {
-		return nil, status.InternalError("No auth configured on this BuildBuddy instance")
-	}
-	u, err := d.env.GetAuthenticator().AuthenticatedUser(ctx)
-	if err != nil {
-		return nil, err
-	}
-	rq := d.h.NewQuery(ctx, "userdb_get_user_by_email").Raw(`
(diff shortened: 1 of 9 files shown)


enterprise/server/backends/userdb/userdb.go
```diff
@@ -931,12 +915,60 @@ func (d *UserDB) GetImpersonatedUser(ctx context.Context) (*tables.User, error)
...
+		byID[u.UserID] = u
+	}
+
+	for len(users) > 0 {
+		count := min(len(users), 200)
+		userIDs := []string{}
+		for _, u := range users[:count] {
+			userIDs = append(userIDs, u.UserID)
+		}
+		rq := d.h.NewQuery(ctx, "userdb_get_users_by_sub_id_prefix").Raw(`
```
bduffany, line 943: nit: do you want to use a different query name for this one?

```suggestion
		rq := d.h.NewQuery(ctx, "userdb_get_user_groups_by_sub_id_prefix").Raw(`
```

enterprise/server/backends/userdb/userdb.go
```diff
@@ -931,12 +915,60 @@ func (d *UserDB) GetImpersonatedUser(ctx context.Context) (*tables.User, error)
...
+	users, err := db.ScanAll(rq, &tables.User{})
 	if err != nil {
-		return err
+		return nil, err
 	}
+
+	byID := map[string]*tables.User{}
+	for _, u := range users {
+		byID[u.UserID] = u
+	}
```
bduffany, line 935: nit: the map makes the ordering non-deterministic which might make this harder to debug (we've been saved once before by a certain query being deterministic). One option is changing the loop to

```
for users := users; len(users) > 0; {
```

bduffany (approved):

PR #5854, opened 2024-02-08
Proto changes for new role-based user API key capabilities

bduffany:
Currently, the "Personal API keys" UI allows admin users to assign arbitrary capabilities to their API keys, and developers always get CAS_WRITE keys (they cannot customize the capabilities). This logic is hard-coded in the frontend code, and the server validates the capabilities requested via the API.

However, now that we've defined a concrete role => capabilities mapping, the server can send up the explicit list of capabilities that the user is allowed to assign. This will make it so that the frontend does not have to duplicate the role => capabilities mapping that we already have on the server, which will make the code simpler and easier to maintain now that we're adding new roles.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3091

### DIFF
--- a/proto/grp.proto
+++ b/proto/grp.proto
@@ -96,7 +96,9 @@ message Group {
   bool external_user_management = 17;
 
   // Maximum set of allowed capabilities that the user can assign to a user API
-  // key created within this group.
+  // key created within this group. Note that these are specific to the
+  // currently authenticated user, not to
+  // the group as a whole.
   repeated api_key.ApiKey.Capability allowed_user_api_key_capabilities = 18;
 }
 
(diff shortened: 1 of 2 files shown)


proto/grp.proto
```diff
@@ -94,6 +94,10 @@ message Group {
   // Whether users are being managed by an external system using the SCIM API.
   // Disables in-app user management.
   bool external_user_management = 17;
+
+  // Maximum set of allowed capabilities that the user can assign to a user API
```
vadimberezniker, line 98: Can you clarify this comment to make it more obvious that this restriction is specific to the currently authenticated user rather than something applicable to the group as a whole?

vadimberezniker (approved):

proto/grp.proto
bduffany: Done

PR #5858, opened 2024-02-08
[SS] Don't recycle runners if the command exit code is non-zero

maggie-lou:
Here's the [definition of CommandResult](https://github.com/buildbuddy-io/buildbuddy/blob/master/server/interfaces/interfaces.go#L985).

Examples of instances this was relevant: 
1) With bazel remote, I tried to apply a bad patch which corrupted my git environment. The snapshot was still saved, and the next time I used it, the git env had to be deleted and recreated which wasted time

2) I passed an invalid bes_backend to the ci runner. The build failed and the bazel process hung for a bit while it tried to upload the build events. It was still hanging when the snapshot was saved. When we tried to unmount VBD, we got the error `failed to sync workspace: failed to unmount workspace: unmount failed: device or resource busy`, I'm assuming because that bazel process was still running.

### DIFF
--- a/enterprise/server/remote_execution/executor/executor.go
+++ b/enterprise/server/remote_execution/executor/executor.go
@@ -341,7 +341,7 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
 		log.CtxErrorf(ctx, "Failed to publish ExecuteResponse: %s", err)
 		return finishWithErrFn(err)
 	}
-	if cmdResult.Error == nil {
+	if cmdResult.Error == nil && cmdResult.ExitCode == 0 {
 		log.CtxInfof(ctx, "Task finished cleanly.")
 		finishedCleanly = true
 	}
--- a/server/interfaces/interfaces.go
+++ b/server/interfaces/interfaces.go
@@ -982,7 +982,7 @@ type Stdio struct {
 
 // CommandResult captures the output and details of an executed command.
 type CommandResult struct {
-	// Error is populated only if the command was unable to be started, or if it was
+	// InitError is populated only if the command was unable to be started, or if it was
 	// started but never completed.
 	//
 	// In particular, if the command runs and returns a non-zero exit code (such as 1),
@@ -1000,7 +1000,7 @@ type CommandResult struct {
 	// reason, such as DEADLINE_EXCEEDED (if the command times out), UNAVAILABLE (if
 	// there is a transient error that can be retried), or RESOURCE_EXHAUSTED (if the
 	// command ran out of memory while executing).
-	Error error
+	InitError error
 	// CommandDebugString indicates the command that was run, for debugging purposes only.
 	CommandDebugString string
 	// Stdout from the command. This may contain data even if there was an Error.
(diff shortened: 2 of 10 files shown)


maggie-lou:
* `master`
  * **PR #5858** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5858?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5858?utm_source=stack-comment).

server/interfaces/interfaces.go
```diff
@@ -982,7 +982,7 @@ type Stdio struct {
 
 // CommandResult captures the output and details of an executed command.
 type CommandResult struct {
-	// Error is populated only if the command was unable to be started, or if it was
+	// InitError is populated only if the command was unable to be started, or if it was
```
bduffany, line 985: I think this name is more confusing than the current name, because this error could also be set if the command is SIGKILL'd or times out (both happen after initialization)

Maybe we could call it NonExitError or something?

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -341,7 +341,7 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
 		log.CtxErrorf(ctx, "Failed to publish ExecuteResponse: %s", err)
 		return finishWithErrFn(err)
 	}
-	if cmdResult.Error == nil {
+	if cmdResult.Error == nil && cmdResult.ExitCode == 0 {
```
bduffany, line 344: This will cause a persistent worker process to be killed if it tries to compile a source file with an error, which is perfectly OK and should not kill the persistent worker.

It's also not clear whether this is the right approach for workflow runs and remote bazel runs, because let's say we spend a lot of time building the full action graph but then there's a small compile error - it means we'll throw away a bazel cache which is mostly good, and so the next compile that just fixes the small build error will have to start from scratch.

bduffany (changes requested):
I think we should restrict this to Firecracker only - for non-Firecracker actions (e.g. persistent worker) it means that a warm runner will get evicted from the pool.

bduffany (commented):
Examples of instances this was relevant:

> With bazel remote, I tried to apply a bad patch which corrupted my git environment. The snapshot was still saved, and the next time I used it, the git env had to be deleted and recreated which wasted time

How was your git environment corrupted? Can we detect bad patches and avoid applying them?

> I passed an invalid bes_backend to the ci runner. The build failed and the bazel process hung for a bit while it tried to upload the build events. It was still hanging when the snapshot was saved. When we tried to unmount VBD, we got the error failed to sync workspace: failed to unmount workspace: unmount failed: device or resource busy, I'm assuming because that bazel process was still running.

There's a separate fix for this that I want to do, which is merging the root FS with the workspace FS

enterprise/server/remote_execution/executor/executor.go
maggie-lou: Ack. I wonder if we should filter on certain error codes. It kind of seems like if there's a build error we might want to recycle it, but if there's a setup error it's probably not worth it (i.e. anything related to git)

Re your other comment: Yes we can avoid applying bad patches, but there are other cases - i.e. If there's a merge conflict between two branches, if there's a network error and we can't fetch a branch etc.

PR #5859, opened 2024-02-08
create iterator after lock in pebble cache

luluz66:
**Related issues**: N/A
In Metadata function(), we currently create the iterator before we lock, which
can cause race condition described in https://github.com/buildbuddy-io/buildbuddy/pull/4460

In deleteMetadataOnly(), I added a comment stating the the key should be locked prior to calling this function. 

For FindMissing, I addressed the issue in https://github.com/buildbuddy-io/buildbuddy/pull/5851

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -1758,7 +1758,7 @@ func (p *PebbleCache) deleteMetadataOnly(ctx context.Context, key filestore.Pebb
 	}
 	defer db.Close()
 
-	unlockFn := p.locker.RLock(key.lockID())
+	unlockFn := p.locker.RLock(key.LockID())
 
 	iter := db.NewIter(nil /*default iterOptions*/)
 	defer iter.Close()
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -1758,13 +1758,17 @@ func (p *PebbleCache) deleteMetadataOnly(ctx context.Context, key filestore.Pebb
 	}
 	defer db.Close()
 
+	unlockFn := p.locker.RLock(key.LockID())
```
vadimberezniker, line 1761: deleteMetadataOnly is called from handleMetadataMismatch and both places that call handleMetadataMismatch have the lock already.

vadimberezniker (changes requested):

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: You could add a comment to the function mentioning that a key lock should already be held before calling the function.

vadimberezniker (approved):

PR #5860, opened 2024-02-08
Add new roles to UI (behind flag)

bduffany:
* Add new roles to dropdown (behind `--app.reader_writer_roles_enabled`)
* Attempt to communicate the permissions differences a little better using a table (see screen recording)

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/12874b08-c873-41c8-a868-10b9153cc327

**Related issues**: N/A

### DIFF
--- a/enterprise/app/org/BUILD
+++ b/enterprise/app/org/BUILD
@@ -92,6 +92,7 @@ ts_library(
     deps = [
         "//app/alert:alert_service",
         "//app/auth:auth_service",
+        "//app/capabilities",
         "//app/components/banner",
         "//app/components/button",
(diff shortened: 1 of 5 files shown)


siggisim (approved):
Should we add some whitespace above the table and maybe make it full width?

Also wondering if we should link to some role documentation here.

bduffany:
> Should we add some whitespace above the table and maybe make it full width?

Done - it feels slightly harder to read the read/write perms columns but it's not too bad, and the component feels overall more balanced.

image

> Also wondering if we should link to some role documentation here.

Yeah, I will add docs soon and link to them. I don't think we have any docs yet about user management.

PR #5862, opened 2024-02-08
remove transactions surrounding exclusively non-locking reads

tempoz:
Non-locking reads have no effects to rollback and do not lock rows during the transaction, so there is no need to wrap them in a transaction.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -358,25 +358,30 @@ func (d *AuthDB) GetAPIKeyGroupFromAPIKey(ctx context.Context, apiKey string) (i
 	}
 
 	akg := &apiKeyGroup{}
-	err := d.h.TransactionWithOptions(ctx, db.Opts().WithStaleReads(), func(tx interfaces.DB) error {
-		qb := d.newAPIKeyGroupQuery(sd, true /*=allowUserOwnedKeys*/)
-		keyClauses := query_builder.OrClauses{}
-		if !*encryptOldKeys {
-			keyClauses.AddOr("ak.value = ?", apiKey)
+	qb := d.newAPIKeyGroupQuery(sd, true /*=allowUserOwnedKeys*/)
+	keyClauses := query_builder.OrClauses{}
+	if !*encryptOldKeys {
+		keyClauses.AddOr("ak.value = ?", apiKey)
+	}
+	if d.apiKeyEncryptionKey != nil {
+		encryptedAPIKey, err := d.encryptAPIKey(apiKey)
+		if err != nil {
+			return nil, err
 		}
-		if d.apiKeyEncryptionKey != nil {
-			encryptedAPIKey, err := d.encryptAPIKey(apiKey)
-			if err != nil {
-				return err
-			}
-			keyClauses.AddOr("ak.encrypted_value = ?", encryptedAPIKey)
-		}
-		keyQuery, keyArgs := keyClauses.Build()
-		qb.AddWhereClause(keyQuery, keyArgs...)
-		q, args := qb.Build()
-		rq := tx.NewQuery(ctx, "authdb_get_api_key_group_by_key").Raw(q, args...)
-		return rq.Take(akg)
-	})
+		keyClauses.AddOr("ak.encrypted_value = ?", encryptedAPIKey)
+	}
+	keyQuery, keyArgs := keyClauses.Build()
+	qb.AddWhereClause(keyQuery, keyArgs...)
+	q, args := qb.Build()
+
+	err := d.h.NewQueryWithOpts(
+		ctx,
+		"authdb_get_api_key_group_by_key",
+		db.Opts().WithStaleReads(),
+	).Raw(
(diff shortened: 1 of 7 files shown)


tempoz:
* `master`
  * **PR #5862** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5862?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #5863** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5864** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5865** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5865?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5862?utm_source=stack-comment).

server/build_event_protocol/target_tracker/target_tracker.go
```diff
@@ -467,14 +467,14 @@ func isTestCommand(command string) bool {
 	return command == "test" || command == "coverage"
 }
 
-func readRepoTargetsWithTx(ctx context.Context, env environment.Env, repoURL string, tx interfaces.DB) ([]*tables.Target, error) {
+func readRepoTargetsWithTx(ctx context.Context, env environment.Env, repoURL string, h interfaces.DB) ([]*tables.Target, error) {
```
tylerwilliams, line 470: Can we move the logic of this function into readRepoTargets instead?

(Since it's only called from one place (I think?) and it's not using a transaction any longer, WithTx is misleading)

enterprise/server/backends/authdb/authdb.go
```diff
@@ -435,27 +446,35 @@ func (d *AuthDB) GetAPIKeyGroupFromAPIKeyID(ctx context.Context, apiKeyID string
...
-		}
-		user.Groups = gs
-		return nil
-	})
+		`,
+		int32(grpb.GroupMembershipStatus_MEMBER),
+		user.UserID,
+	)
+
+	user.Groups, err = db.ScanAll(rq, &tables.GroupRole{})
```
tylerwilliams, line 477: It seems to me that queries like this, with  two select statements run, the second depending on the first, should be using REPEATABLE READ isolation level to avoid possible race conditions.

What do you think?

tempoz: Oh interesting, I wasn't aware of REPEATABLE READ. You're definitely right; we could get inconsistency that doesn't currently exist if we don't keep this in a transaction. Not that I'm looking at doing this right now, but would it make sense if we moved some of these that don't have meaningful application logic in the middle into `JOIN`s?

tylerwilliams: yeah I think if we removed the two-dependent-select statements from this CL, then sent another CL to make those JOINs (or leave them in a REPEATABLE READ txn) we'd be solid

tempoz: Sounds good!

server/build_event_protocol/target_tracker/target_tracker.go
tempoz: Done!

tylerwilliams (approved):
LGTM but please let @vadimberezniker have a look before submitting. Thanks!

enterprise/server/backends/userdb/userdb.go
```diff
@@ -796,11 +794,8 @@ func (d *UserDB) InsertUser(ctx context.Context, u *tables.User) error {
 func (d *UserDB) GetUserByID(ctx context.Context, id string) (*tables.User, error) {
 	var user *tables.User
 	err := d.h.Transaction(ctx, func(tx interfaces.DB) error {
-		u, err := d.getUser(ctx, tx, id)
-		if err != nil {
-			return err
-		}
-		user = u
+		var err error
```
vadimberezniker, line 797: Do we need  the txn here?

enterprise/server/backends/userdb/userdb.go
```diff
@@ -822,11 +817,8 @@ func (d *UserDB) GetUserByID(ctx context.Context, id string) (*tables.User, erro
 func (d *UserDB) GetUserByIDWithoutAuthCheck(ctx context.Context, id string) (*tables.User, error) {
 	var user *tables.User
 	err := d.h.Transaction(ctx, func(tx interfaces.DB) error {
-		u, err := d.getUser(ctx, tx, id)
-		if err != nil {
-			return err
-		}
-		user = u
+		var err error
```
vadimberezniker, line 820: Do we need  the txn here?

vadimberezniker (approved):

PR #5866, opened 2024-02-08
Make action-merging more resilient to dying apps/executors

iain-macdonald:
The current implementation of action merging:
1. Checks Redis for a pending execution ID for a given action digest when an Execute request is received, reusing the pending execution ID if one is found, or
2. Writes two entries in Redis bidirectionally mapping action IDs to/from execution IDs with an 8 hour TTL for scheduled executions, and finally
3. Deletes this bidirectional map when the execution completes.

This is not resilient to unexpected failures in the app during execution, or cases where all of the executors simultaneously exit unhealthily. If either of these things happen, the action-merging entries written in step 2 can persist, because step 3 never happens. These "stuck" action merging entries block subsequent execution requests for that action, as alluded to in the linked bug.

This PR changes the implementation to make action merging recover from these failures more quickly by:
1. Recording the action merging information in Redis when the execution is queued, but with a 10 minute TTL instead of an 8 hour TTL.
2. Bumping the TTL of the action-merging Redis entries by 20 seconds every time a LeaseTask request bumping an executor's lease on an execution is healthily processed in the streaming LeaseTask RPC server. This should happen about every 10 seconds during normal operation.
3. Removing both of these action-merging Redis entries when an execution is unclaimed in the LeastTask RPC handler.

These changes should help address the hung action-merging bugs we've seen over the past few months. The window for stuck actions decreases from 8 hours to 10 minutes in the event queued executions are lost, and decreases it further to 20 seconds in the event live executions are lost. This should make manual intervention to clean up stuck actions unnecessary, although the 10 minute window is still a bit painful.

I also added a test for the dead-app case which necessitated some changes in the testing code to support removing apps. This new test fails at head, but passes at this PR. I also added another test for the TTL bumping. And finally, I extracted the logic for working with action-merging entries in Redis into its own package.

**Related issues**: [2860](https://github.com/buildbuddy-io/buildbuddy-internal/issues/2860)

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -1255,8 +1255,6 @@ func (s *SchedulerServer) unclaimTask(ctx context.Context, taskID, leaseID, reco
 	}
 	log.CtxDebugf(ctx, "Released task claim in Redis (reconnecting=%t)", reconnectToken != "")
 
-	action_merger.DeletePendingExecution(ctx, s.rdb, taskID)
-
 	return nil
 }
 
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/action_merger"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/platform"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/tasksize"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
@@ -1220,6 +1221,9 @@ func (s *SchedulerServer) deleteClaimedTask(ctx context.Context, taskID string) 
 	if c, ok := r.(int64); !ok || c != 1 {
 		return status.NotFoundErrorf("unable to delete claimed task %s", taskID)
 	}
+
+	action_merger.DeletePendingExecution(ctx, s.rdb, taskID)
+
 	return nil
 }
 
@@ -1240,6 +1244,9 @@ func (s *SchedulerServer) unclaimTask(ctx context.Context, taskID, leaseID, reco
 		return status.NotFoundErrorf("unable to release task claim for task %s", taskID)
 	}
 	log.CtxDebugf(ctx, "Released task claim in Redis (reconnecting=%t)", reconnectToken != "")
+
+	action_merger.DeletePendingExecution(ctx, s.rdb, taskID)
+
 	return nil
 }
 
@@ -1282,6 +1289,8 @@ func (s *SchedulerServer) claimTask(ctx context.Context, taskID, reconnectToken 
 		log.CtxErrorf(ctx, "claimTask %q error: unknown error code: %d", taskID, c)
 		return "", status.UnknownErrorf("unknown error %d", c)
 	}
+
+	action_merger.RecordClaimedExecution(ctx, s.rdb, taskID)
 
 	return leaseId, nil
 }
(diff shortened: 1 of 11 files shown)


enterprise/server/test/integration/remote_execution/remote_execution_test.go
```diff
@@ -1691,21 +1710,86 @@ func TestActionMerging(t *testing.T) {
...
+	cmd := &repb.Command{
+		Arguments: []string{"sh", "-c", "sleep 60"},
+		Platform:  platform,
+	}
+	cmd1 := rbe.Execute(cmd, &rbetest.ExecuteOpts{CheckCache: true, InvocationID: "invocation1"})
+	op1 := cmd1.WaitAccepted()
+	WaitForPendingExecution(rbe.GetRedisClient(), op1)
+
+	// Wait long enough for the first action-merging state to have expired.
+	time.Sleep(21 * time.Second)
```
iain-macdonald, line 1750: Let me know if you don't like these `sleep`s (here and on line 1784). An alternative would be to support overriding the TTL extension for testing.

siggisim:
Am I reading the description correctly that if we get two Execute requests at roughly the same time for the same action that are both queued for 100ms that before they'd be merged, but in the new world they wouldn't be merged?

Or put another way, if we're under heavy load and all actions are queued for 30s+ and the same integration test comes in 4 times in that 30s span - we'll run the integration test 4 times once we churn through the queue instead of once?

iain-macdonald:
> Am I reading the description correctly that if we get two Execute requests at roughly the same time for the same action that are both queued for 100ms that before they'd be merged, but in the new world they wouldn't be merged?
> 
> Or put another way, if we're under heavy load and all actions are queued for 30s+ and the same integration test comes in 4 times in that 30s span - we'll run the integration test 4 times once we churn through the queue instead of once?

Yes, that's correct. Without a better mechanism for rescheduling queued tasks that are lost when all executors disappear unhealthily, or at least being able to detect that an somehow un-merge actions, this is the simplest way we could come up with for preventing subsequent actions merging against these lost actions, but it comes at the expense of some more duplicate work than was possible in the previous implementation.

iain-macdonald:
Updated PR based on offline discussion with Siggi (and earlier feedback from Vadim). Now, it supports merging queued actions as well, but the TTL is only 10 minutes instead of 8 hours, so actions that are enqueued and then orphaned will only leave a 10 minute window during which they'll blackhole other, identical actions. This feels like a reasonable compromise to me, but I'm open to feedback, either for pursuing a different approach, or tweaking the values I've chosen.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1230,6 +1234,9 @@ func (s *SchedulerServer) unclaimTask(ctx context.Context, taskID, leaseID, reco
 		return status.NotFoundErrorf("unable to release task claim for task %s", taskID)
 	}
 	log.CtxDebugf(ctx, "Released task claim in Redis (reconnecting=%t)", reconnectToken != "")
+
+	action_merger.DeletePendingExecution(ctx, s.rdb, taskID)
```
vadimberezniker, line 1238: If a task is unclaimed, it doesn't mean it's done. It can still be retried on a different executor. Do we need to remove the link here?

enterprise/server/test/integration/remote_execution/remote_execution_test.go
vadimberezniker: Yeah, this is too much sleep for a test. Make it configurable.

vadimberezniker (commented):
Looks good overall, but let's reduce the sleep time in tests.

enterprise/server/test/integration/remote_execution/remote_execution_test.go
iain-macdonald: Yeah, fair enough. Made it sleep less and test more.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
iain-macdonald: Yeah, I think that makes sense. I had it in there because I was worried the task could be unclaimed and not reclaimed, and then get stuck again, but with only a 20 second blackhole period I think it's fine to let the entry sit around on unclaim.

vadimberezniker (approved):

PR #5872, opened 2024-02-09
Suggest `--digest_function=blake3`

brentleyjones:
**Related issues**: N/A


bduffany:
We might want to wait on this until we have more data on whether it speeds up builds or not. I know we've suggested to a few customers that they experiment with setting this flag, but I'm not aware of any customers that have reported back to us on whether they measured any improvement.

brentleyjones:
Related: https://github.com/bazelbuild/bazel/discussions/22011

PR #5875, opened 2024-02-09
remove transaction from single call to `addUserToGroup`

tempoz:
`addUserToGroup` is a series of non-locking reads followed by a single terminating write, and thus does not need to be wrapped in a transaction unless the caller is doing additional work in that transaction.

**Related issues**: N/A


bduffany (commented):
I think the intent of this transaction was to make `addUserToGroup` atomic, since the `role` assigned in the `UPDATE` case depends on the result of the `COUNT`. Will it still be atomic if we remove the wrapping transaction? (or if it's not actually atomic today, any idea what we need to change to fix it? maybe SERIALIZABLE isolation or something?)

PR #5879, opened 2024-02-09
[RB] Add unit test for importing user.bazelrc and print an update so it doesn't seem like cli is hanging when unpausing a snapshot

maggie-lou:
Fixes buildbuddy-io/buildbuddy-internal#2979

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -55,19 +55,17 @@ const (
 	gitConfigSection           = "buildbuddy"
 	gitConfigRemoteBazelRemote = "remote-bazel-remote-name"
 	defaultRemoteExecutionURL  = "remote.buildbuddy.io"
-	defaultRemoteResultsURL    = "https://app.buildbuddy.io/invocation/"
 )
 
 var (
 	remoteFlagset = flag.NewFlagSet("remote", flag.ContinueOnError)
 
-	execOs           = remoteFlagset.String("os", "linux", "If set, requests execution on a specific OS.")
-	execArch         = remoteFlagset.String("arch", "amd64", "If set, requests execution on a specific CPU architecture.")
-	containerImage   = remoteFlagset.String("container_image", "", "If set, requests execution on a specific runner image. Otherwise uses the default hosted runner version. A `docker://` prefix is required.")
-	envInput         = bbflag.New(remoteFlagset, "env", []string{}, "Environment variables to set in the runner environment. Key-value pairs can either be separated by '=' (Ex. --env=k1=val1), or if only a key is specified, the value will be taken from the invocation environment (Ex. --env=k2). To apply multiple env vars, pass the env flag multiple times (Ex. --env=k1=v1 --env=k2). If the same key is given twice, the latest will apply.")
-	remoteRunner     = remoteFlagset.String("remote_runner", defaultRemoteExecutionURL, "The Buildbuddy grpc target the remote runner should run on.")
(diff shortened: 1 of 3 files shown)


maggie-lou:
* `master`
  * **PR #5879** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5879?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5879?utm_source=stack-comment).

cli/remotebazel/remotebazel.go
```diff
@@ -561,13 +559,14 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
...
 
+	log.Printf("\nWaiting for available remote runner...\n")
 	rsp, err := bbClient.Run(ctx, req)
 	if err != nil {
 		return 0, status.UnknownErrorf("error running bazel: %s", err)
 	}
 
 	iid := rsp.GetInvocationId()
-	log.Debugf("Invocation ID: %s", iid)
+	log.Printf("\nStreaming remote bazel logs to: %s/invocation/%s", *remoteRunner, iid)
```
bduffany, line 569: IIRC the remote_runner flag is a gRPC target? if so, this won't necessarily be a valid http URL that people can click. We could add a secondary flag like `remote_runner_results_url` which defaults to `https://app.buildbuddy.io/invocation/` - similar to Bazel's `bes_results_url` flag except with a default value set to match `remote_runner`.

cli/remotebazel/remotebazel.go
```diff
@@ -561,13 +559,14 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 		req.Timeout = timeout.String()
 	}
 
+	log.Printf("\nWaiting for available remote runner...\n")
```
bduffany, line 562: Doesn't have to be in this PR but we might want to wait until we've waited 2-3s without receiving the invocation ID before we print this message. Feels like this will be pretty noisy in the case where the invocation starts immediately (ideally we want it to feel ~instant and we wouldn't print this message ever)

cli/remotebazel/remotebazel.go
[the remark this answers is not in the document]
maggie-lou: Here's what it looks like with the preceding new line: https://share.zight.com/GGudWgNQ

I think it looks a little more readable with some extra indentation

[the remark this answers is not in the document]
bduffany: I think for things like logs, we should avoid trying to add visual "padding" in the form of extra newlines - our current convention is that the log lines are densely packed and we should probably stick to that for now. If we start adding whitespace in some areas then we should also do it in other areas, but then I think we'd want to put a little more thought into what should constitute a "section" and what shouldn't.

cli/remotebazel/remotebazel.go
bduffany: sounds good :+1:

cli/remotebazel/remotebazel.go
bduffany: Since the `ci_runner` knows the source of truth, should we make the CI runner's first line of log output be something like "Streaming (workflow |remote bazel) logs to <url>"?

maggie-lou: Oh good idea - changed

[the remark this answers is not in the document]
maggie-lou: Not relevant anymore

bduffany (approved):

PR #5883, opened 2024-02-12
Update user API keys UI for new roles

bduffany:
If the new roles are enabled (via flag):
- "Reader" defaults to "Read-only" and can't select other options
- "Developer" role can now set _either_ "Read-only" or "CAS-only" (could only set CAS-only before)
- "Writer" role can select any option

**Related issues**: N/A


vadimberezniker (approved):
I think it would be nice if the UI explained why certain capabilities are not selectable. I don't know what the best way to do this, maybe a little blurb between the selectable and non-selectable options that says "the following capabilities are not allowed by your role:"?

bduffany:
@vadimberezniker I agree that'd be helpful - it will take a bit of refactoring to do it in a clean way, so will do this in a followup

PR #5885, opened 2024-02-12
Paint it black

siggisim:
[Paint it Black](https://www.youtube.com/watch?v=O4irXQhgMqg)
https://www.youtube.com/watch?v=O4irXQhgMqg

This PR removes blue-gray as our primary color and replaces it with off-black / mono-chrome (the Grey scale from https://materialui.co/colors/).

I find this makes the app feel cleaner / more modern, and really makes the colors / content pop (i.e. the timing profile or the build status color).

For pages where there is now no color (like the empty state pages), I will send some follow-up PRs to bring a pop of color to the button on the page that we think the user should click.

Before:
Screenshot 2024-02-12 at 11 47 12 AM

After:
Screenshot 2024-02-12 at 11 46 53 AM


bduffany (approved):

sluongng:
https://youtu.be/3iZ9JRVmJ5o?si=I0leFzh2ZyUBoezE I prefer this version 😋

PR #5886, opened 2024-02-12
SCIM: support user deletion via DELETE as well

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/scim/scim_test.go
+++ b/enterprise/server/scim/scim_test.go
@@ -99,6 +99,10 @@ func (tc *testClient) Put(url string, body []byte) (int, []byte) {
 	return tc.do(http.MethodPut, url, body)
 }
 
+func (tc *testClient) Delete(url string) (int, []byte) {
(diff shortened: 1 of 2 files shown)


enterprise/server/scim/scim_test.go
```diff
@@ -511,6 +521,16 @@ func TestDeleteUser(t *testing.T) {
 		require.True(t, status.IsNotFoundError(err))
 	}
 
+	// Delete user US104 using a DELETE request.
+	{
+		code, body := tc.Delete(baseURL + "/scim/Users/US104")
+		require.Equal(tc.t, http.StatusNoContent, code, "body: %s", string(body))
+		require.NoError(t, err)
+		_, err = udb.GetUserByID(userCtx, "US104")
+		require.Error(t, err)
```
bduffany, line 530: nit: for the sake of illustration, maybe also assert that the user exists before doing `tc.Delete` ?

bduffany (approved):

PR #5849, opened 2024-02-07
Add user config files to .gitignore

maggie-lou:
I find myself wasting a lot of time setting the same config options when doing local testing. Add a user-configs dir to .gitignore, so I can maintain a set of configs that I like to use.

PR #5892, opened 2024-02-12
Support using server configs from nested directories

maggie-lou:
Users may have specific configs they like to apply when running apps/executors locally. (See this [PR](https://github.com/buildbuddy-io/buildbuddy/pull/5849))

Edit to [this PR](https://github.com/buildbuddy-io/buildbuddy/pull/5812/files#diff-7ef0c2365b35c8781111728f4d4d9ad7b2a0f0a199b1fb9d9f52036cc5d56833)

### DIFF
--- a/enterprise/config/BUILD
+++ b/enterprise/config/BUILD
@@ -10,7 +10,7 @@ filegroup(
             "buildbuddy.release.yaml",
             "executor.release.yaml",
         ],
-        "//conditions:default": glob(["*.yaml"]),
+        "//conditions:default": glob(["**/*.yaml"]),
     }),
 )
 


maggie-lou:
* `master`
  * **PR #5892** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5892?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5892?utm_source=stack-comment).

enterprise/config/BUILD
```diff
@@ -10,7 +10,7 @@ filegroup(
             "buildbuddy.release.yaml",
             "executor.release.yaml",
         ],
-        "//conditions:default": glob(["*.yaml"]),
+        "//conditions:default": glob(["**/*.yaml"]),
```
bduffany, line 13: If you run `git clean` then it'll delete your user configs - wdyt about putting configs in a parent dir like `--config_file=../my_config.yaml` ?

vadimberezniker:
Why does the config need to be packaged for local use? --config_file can read it from anywhere.

vadimberezniker:
> > this PR
> 
> @vadimberezniker I don't know if you're asking philosophically or practically, but from my testing the config file has to be included in here:
> 
> https://github.com/buildbuddy-io/buildbuddy/blob/47eec2275d6f37a56e20c851bb25a6dbf82c7a29/enterprise/server/cmd/server/BUILD#L95

Are you using absolute or relative paths when specifying --config_file? 

If you're using relative paths then I can see how you could run into problems if the config file is not referenced by the app binary since `bazel run` changes the working directory. You can get around this if you specify --config_file using absolute paths or using $PWD/config/path if you use bazel run from the root of the directory.

sluongng:
I'm not against this change, but given the comments above, I will leave the decision to Brandon and Vadim.

I usually edit the local config file and keep it either in a branch or a git stash instead of keeping a separate set of config files.

This makes me wonder, do you have some common usage patterns, or best practices in your files that we could port back to the repo? Long-lived local forks of the config could be a sign that our default config files are doing an inadequate job of aiding local development. 🤔

bduffany:
This isn't the first time we've hit some confusion due to the app starting from the runfiles dir - I was confused by this not working at one point and other people have asked me about it as well.

I wonder if our logic for `--config_file` should also try resolve paths relative to `$BUILD_WORKSPACE_DIRECTORY` if this env var is set and the initial path lookup fails. A simpler solution could be to have the app just cd to that directory in main() but I'm not sure if that will break runfiles path resolution, maybe worth just trying it out.

maggie-lou:
> > > this PR
> > 
> > 
> > @vadimberezniker I don't know if you're asking philosophically or practically, but from my testing the config file has to be included in here:
> > https://github.com/buildbuddy-io/buildbuddy/blob/47eec2275d6f37a56e20c851bb25a6dbf82c7a29/enterprise/server/cmd/server/BUILD#L95
> 
> Are you using absolute or relative paths when specifying --config_file?
> 
> If you're using relative paths then I can see how you could run into problems if the config file is not referenced by the app binary since `bazel run` changes the working directory. You can get around this if you specify --config_file using absolute paths or using $PWD/config/path if you use bazel run from the root of the directory.

Ah I see. I was using a relative path. Using absolute path fixed it

maggie-lou:
> I'm not against this change, but given the comments above, I will leave the decision to Brandon and Vadim.
> 
> I usually edit the local config file and keep it either in a branch or a git stash instead of keeping a separate set of config files.
> 
> This makes me wonder, do you have some common usage patterns, or best practices in your files that we could port back to the repo? Long-lived local forks of the config could be a sign that our default config files are doing an inadequate job of aiding local development. 🤔

Using absolute path is sufficient. I'll close this. 

Re the other options: I find there to be some annoying overhead with always cherry-picking my local config changes and making sure they don't get committed. And some of my common usage patterns are specific to my work - i.e. I'm frequently testing workflows, so I need the git client config and firecracker support, but most people on our team don't need those

PR #5899, opened 2024-02-13
Remove perms.AuthenticatedUser

bduffany:
Background: we have two packages `perms` and `authutil` and it's somewhat confusing what should go in which package, as both seem to contain broadly useful auth utilities. Ideally, `perms` would only have utils relating to our unix-ish Perms that we store in the DB (i.e. read/write/execute bit-fiddling stuff), and `authutil` would contain more broadly useful auth utils.

This PR kicks off this effort by removing the `perms.AuthenticatedUser` func. This utility function is not really needed - all it does is check whether `env.GetAuthenticator()` is `nil`, and if it's not, it calls `GetAuthenticatedUser()` on it. `GetAuthenticator()` should never return `nil` though, because we always either bind the real authenticator, or a non-`nil` instance of `nullauth.NullAuthenticator`. So instead, just replace all existing call sites with `env.GetAuthenticator().AuthenticatedUser(ctx)`.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -567,7 +567,7 @@ func newAPIKeyToken() (string, error) {
 }
 
 func (d *AuthDB) authorizeGroupAdminRole(ctx context.Context, groupID string) error {
-	u, err := perms.AuthenticatedUser(ctx, d.env)
+	u, err := d.env.GetAuthenticator().AuthenticatedUser(ctx)
 	if err != nil {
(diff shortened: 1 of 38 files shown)


vadimberezniker (approved):
There are still some places where we check if env.GetAuthenticator() is nil. Should we clean those up as well?

bduffany:
Yep, I will send a follow-up to remove those nil checks

tylerwilliams (approved):
nice!

PR #5903, opened 2024-02-13
Clean up nil checks on `env.GetAuthenticator()`

bduffany:
Follow-up from https://github.com/buildbuddy-io/buildbuddy/pull/5899#pullrequestreview-1878738508

The app binds a `NullAuthenticator` instance if auth is disabled, otherwise it binds a "real" authenticator, so `env.GetAuthenticator()` will never return `nil`. As a result, we can simplify any code that checks whether the authenticator is nil.

Also, some of our existing code was already not doing the nil checks, so if anything, this PR makes the nil checking consistent by removing it everywhere.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/api/api_server.go
+++ b/enterprise/server/api/api_server.go
@@ -61,10 +61,6 @@ func NewAPIServer(env environment.Env) *APIServer {
 	}
 }
 
-func (s *APIServer) checkPreconditions(ctx context.Context) (interfaces.UserInfo, error) {
-	return s.env.GetAuthenticator().AuthenticatedUser(ctx)
-}
-
 func (s *APIServer) authorizeWrites(ctx context.Context) error {
 	canWrite, err := capabilities.IsGranted(ctx, s.env, akpb.ApiKey_CACHE_WRITE_CAPABILITY)
 	if err != nil {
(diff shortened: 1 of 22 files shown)


enterprise/server/api/api_server.go
```diff
@@ -62,10 +62,6 @@ func NewAPIServer(env environment.Env) *APIServer {
 }
 
 func (s *APIServer) checkPreconditions(ctx context.Context) (interfaces.UserInfo, error) {
```
tempoz, line 64: Do we still need this function if we've removed all the precondition checks?

At least it seems like we should rename it.

server/telemetry/telemetry_client.go
```diff
@@ -203,13 +204,12 @@ func getHostname() string {
 
 func getFeatures(env environment.Env) *telpb.TelemetryFeature {
 	cache := env.GetCache()
-	authenticator := env.GetAuthenticator()
 	api := env.GetAPIService()
-
+	_, isNullAuth := env.GetAuthenticator().(*nullauth.NullAuthenticator)
```
tempoz, line 208: Nit: instead of casting here, we could add `Enabled()` to the `Authenticator` interface, which would remove the direct dependency on `nullauth` and make this kind of check less awkward to do in general.

tempoz (approved):
Thanks for doing this! It makes code around auth that much less error-prone and really cleans up the existing code

enterprise/server/api/api_server.go
bduffany: Done (inlined the AuthenticatedUser checks instead of calling this func)

server/telemetry/telemetry_client.go
bduffany: I would slightly prefer to leave as-is to avoid expanding the API surface of the authenticator, and also because a proper API method might encourage this type of checking.

vadimberezniker (approved):

PR #5904, opened 2024-02-13
Add comments for exported methods in RangeMap, and update some tests

luluz66:
**Related issues**: N/A
In TestAddOrdering, check the order of range.Start instead of range.Val, since we are ordering by the interval instead of value. 
In TestGet, add test cases where only one of start and end matches.

### DIFF
--- a/server/util/rangemap/rangemap.go
+++ b/server/util/rangemap/rangemap.go
@@ -14,7 +14,8 @@ var (
 	RangeDoesNotExistError = errors.New("Range does not exist")
 )
 
-// Ranges are [inclusive,exclusive)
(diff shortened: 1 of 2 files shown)


server/util/rangemap/rangemap.go
```diff
@@ -122,6 +131,8 @@ func (rm *RangeMap) Get(start, end []byte) *Range {
 	return nil
 }
 
+// GetOverlapping returns a list of ranges overlapped witht the specific start
```
tylerwilliams, line 134: s/witht/with/g

tylerwilliams (approved):

PR #5906, opened 2024-02-13
[SS] Support caching old snapshot manifests

maggie-lou:
Because we overwrite the master snapshot key every time a snapshot is saved, it can be difficult to debug snapshot failures because a previous, broken snapshot key might get overwritten. Add a unique identifier to each snapshot run and cache snapshot manifests for each run.

Now if we have an invocation link for a failed workflow run, we can grab the snapshot ID from the action details page for the workflow action and search for it in the logs. We can then find the snapshot key associated with that snapshot ID, and use vmstart to run that snapshot.


maggie-lou:
* `master`
  * **PR #5906** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5906?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5906?utm_source=stack-comment).

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1017,6 +1021,17 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context) error {
...
+	// Set unique per-run identifier on the vm metadata so this exact snapshot
+	// run can be identified
+	if snap.GetVMMetadata() == nil {
+		md := &repb.VMMetadata{
+			VmId:       c.id,
+			SnapshotId: c.id,
+		}
+		snap.SetVMMetadata(md)
+	}
+	snap.GetVMMetadata().SnapshotId = c.id
```
bduffany, line 1034: hm I am a little confused why we are setting the snapshot ID when unpacking the snapshot - I would think that since we've just unpacked the snapshot, the ID should stay the same until just before we're about to commit a new version to AC?

proto/remote_execution.proto
```diff
@@ -2294,9 +2294,17 @@ message VMMetadata {
     // ExecuteResponse. The digest function used to compute the digest should
     // match the one associated with the invocation.
     Digest execute_response_digest = 4;
+
+    // A unique UUID assigned each time a snapshot is used.
+    // For example, even if the same execution is retried multiple times using
+    // the same snapshot key, each run will have a unqiue snapshot_id.
+    string snapshot_id = 5;
```
bduffany, line 2301: I am a bit unclear on why a task has a snapshot ID associated with it - the snapshot ID is a property of the VM, not the task right?

maggie-lou: Let's say you have an invocation link for a workflow. This workflow (VM B) was forked from VM A. In my mental model, if you grab snapshot ID B from that invocation page, you're starting from the snapshot VM B saved. If you want to start from the snapshot VM A saved, you should follow the VM metadata link to the previous workflow link and grab snapshot ID A from there.

Under that model, my intention is that this snapshot_id (set for `last_executed_task`) represents the ID of the snapshot the VM was resumed from (snapshot-id-A). 

The snapshot ID set directly on the VMMetadata object (line 2308) represents the ID of the current snapshot run (before and during task execution - snapshot-id-B). 

I think we can make this a little more clear by adding some more context to the UI. "VM resumed from snapshot ID X" "VM saved to snapshot ID Y". Wdyt?

app/invocation/invocation_action_card.tsx
```diff
@@ -557,6 +557,14 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
                                   {this.state.actionResult.executionMetadata.vmMetadata.vmId}
                                 </div>
+                                {this.state.actionResult.executionMetadata.vmMetadata.snapshotId && (
+                                  <>
+                                    <div className="metadata-title">Snapshot ID</div>
+                                    <div className="metadata-detail">
+                                      {this.state.actionResult.executionMetadata.vmMetadata.snapshotId}
+                                    </div>
+                                  </>
+                                )}
```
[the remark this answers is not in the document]
maggie-lou: Done

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1017,6 +1025,17 @@ func (c *FirecrackerContainer) LoadSnapshot(ctx context.Context) error {
 	if err != nil {
 		return status.WrapError(err, "failed to get snapshot")
 	}
+
+	// Set unique per-run identifier on the vm metadata so this exact snapshot
+	// run can be identified
+	if snap.GetVMMetadata() == nil {
+		md := &repb.VMMetadata{
+			VmId:       c.id,
+			SnapshotId: c.id,
```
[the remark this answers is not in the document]
maggie-lou: Done. In hindsight we should've generated a unique field for VM ID instead. Because c.id is unique per firecracker run, which is more aligned to what snapshotID is than vmID.

proto/remote_execution.proto
maggie-lou: Ack - will do in a follow up

enterprise/server/remote_execution/containers/firecracker/firecracker.go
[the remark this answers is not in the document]
bduffany: Yeah, agreed

PR #5913, opened 2024-02-14
Remove global admin group logic

bduffany:
This "admin" group is no longer used since we added proper impersonation. Checked in the DB, and no users are a member of the "admin" group ID according to the UserGroups table.

**Related issues**: N/A


vadimberezniker (approved):
nice

tylerwilliams (approved):
🔥

PR #5918, opened 2024-02-14
Roll forward schduler_server refactor

iain-macdonald:
This is a revert of [5639](https://github.com/buildbuddy-io/buildbuddy/pull/5639) with a fix for [the bug](https://buildbuddy-corp.slack.com/archives/C01D5GHRJ59/p1704318477269539) mentioned in there. I put a bit of effort into writing a test for the bug and was able to tickle it with a simple test and throwing a `time.Sleep()` into the production code, because it relies on a specific sequence of events that are difficult to trigger organically. That's obviously not submittable, but I've included it in the git history of this branch so you can take a look or play around with it if you'd like.

To fix the bug, I got rid of the node ring and replaced it with a slice from which nodes are removed after enqueue attempts are made, and refreshed that slice of nodes each time it's empty.

Here's the diff versus the previous iteration (useful for reviewing scheduler_server): https://github.com/buildbuddy-io/buildbuddy/compare/4331112728bacb670e8ad3955396072a6706fd7a...4d102420fa00e3dd0933d44fbe4919bf775324ce

**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -530,6 +530,21 @@ func toNodeInterfaces(nodes []*executionNode) []interfaces.ExecutionNode {
 	return out
 }
 
+// If the debug-executor-id platform property is set, filters the given nodes
+// to the nodes with that ID. The returned list may be empty.
+func filterToDebugExecutorID(nodes []*executionNode, task *repb.ExecutionTask) []*executionNode {
+	id := platform.FindEffectiveValue(task, "debug-executor-id")
+	if id == "" {
+		return nodes
+	}
+	for _, n := range nodes {
+		if n.executorID == id {
+			return []*executionNode{n}
+		}
+	}
+	return nil
+}
+
 type nodePoolKey struct {
 	groupID string
 	os      string
@@ -1719,7 +1734,7 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 					enqueueRequest.GetTaskSize().GetEstimatedMemoryBytes())
 			}
 			candidateNodes = filterToDebugExecutorID(candidateNodes, task)
-			if len(nodes) == 0 {
+			if len(candidateNodes) == 0 {
 				return status.UnavailableErrorf("requested executor ID not found")
 			}
 			rankedNodes = s.taskRouter.RankNodes(ctx, cmd, remoteInstanceName, toNodeInterfaces(candidateNodes))
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -1706,7 +1706,8 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 			break
 		}
 		enqueueStart := time.Now()
-		if s.enqueue(ctx, preferredNode, enqueueRequest, opts) {
+		enqueued, _ := s.enqueue(ctx, preferredNode, enqueueRequest, opts)
+		if enqueued {
 			scheduledOnPreferredNode = true
 			successfulReservations = append(successfulReservations, successfulReservation(preferredNode, enqueueStart))
 		}
@@ -1766,7 +1767,11 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 		} else if delayable {
 			enqueueRequest.Delay = nil
 		}
-		if s.enqueue(ctx, rankedNode.GetExecutionNode().(*executionNode), enqueueRequest, opts) {
+		enqueued, rpcErr := s.enqueue(ctx, rankedNode.GetExecutionNode().(*executionNode), enqueueRequest, opts)
+		if rpcErr != nil {
+			time.Sleep(schedulerEnqueueTaskReservationFailureSleep)
+		}
+		if enqueued {
 			if rankedNode.IsPreferred() {
 				scheduledOnPreferredNode = true
 			}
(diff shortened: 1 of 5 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1802,6 +1759,52 @@ func getNonPreferredSchedulingDelay(cmd *repb.Command) time.Duration {
...
+}
+
+func enqueueOnConnectedExecutor(ctx context.Context, node *executionNode, request *scpb.EnqueueTaskReservationRequest) bool {
+	if node.handle == nil {
+		log.CtxErrorf(ctx, "nil handle for a local executor %q", node.GetExecutorID())
+		return false
+	}
+	_, err := node.handle.EnqueueTaskReservation(ctx, request)
+	if err != nil {
+		log.Infof("failed to enqueue task on connected executor: %s", err)
```
bduffany, line 1781:
```suggestion
		log.CtxInfof(ctx, "Failed to enqueue task on connected executor: %s", err)
```

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1802,6 +1759,52 @@ func getNonPreferredSchedulingDelay(cmd *repb.Command) time.Duration {
...
+	}
+
+	schedulerClient, err := s.schedulerClientCache.get(node.schedulerHostPort)
+	if err != nil {
+		log.CtxWarningf(ctx, "Could not get SchedulerClient for %q: %s", node.schedulerHostPort, err)
+		return false
+	}
+	rpcCtx, cancel := context.WithTimeout(ctx, schedulerEnqueueTaskReservationTimeout)
+	_, err = schedulerClient.EnqueueTaskReservation(rpcCtx, request)
+	cancel()
```
bduffany, line 1799: nit: we can `defer cancel()` now that this is in its own function

```suggestion
	defer cancel()
	_, err = schedulerClient.EnqueueTaskReservation(rpcCtx, request)
```

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1802,6 +1759,52 @@ func getNonPreferredSchedulingDelay(cmd *repb.Command) time.Duration {
...
+	if err != nil {
+		log.CtxWarningf(ctx, "EnqueueTaskReservation via scheduler target %q failed: %s", node.schedulerHostPort, err)
+		time.Sleep(schedulerEnqueueTaskReservationFailureSleep)
```
bduffany, line 1802: Should we move this sleep outside the func and sleep whenever it returns false? (I think it would help us avoid getting into a hot CPU loop if something is borked.)

Alternatively we could return `(bool, error)` and abort the scheduling loop if a "fatal" error is returned. (non-fatal errors would return `(false, nil)`)

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1671,110 +1662,76 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
...
 	attempts := 0
-	var rankedNodes []*rankedExecutionNode
-	sampleIndex := 0
+	var rankedNodes []interfaces.RankedExecutionNode
 	nonPreferredDelay := getNonPreferredSchedulingDelay(cmd)
-	delaySpecified := enqueueRequest.GetDelay() != nil
-	for probesSent < probeCount {
-		attempts++
+	delayable := enqueueRequest.GetDelay() == nil
+	for len(successfulReservations) < probeCount {
```
bduffany, line 1683: Since `successfulReservations` is a list of "debug strings" it feels slightly unsafe to be using it for control flow here.

I would prefer to bring back the `probesSent` counter and then do the control flow based on that, then keep `successfulReservations` purely for logging/debugging purposes

enterprise/server/scheduling/scheduler_server/scheduler_server.go
iain-macdonald: That seems like a reasonable goal. I went ahead and did the `(bool, error)` approach because that let me preserve the existing behavior as-is, always sleeping on false would change the behavior by introducing a new sleep in a few cases which I'd like to avoid in this PR. LMK what you think.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
iain-macdonald: I'd kind of like to not have a variable that's just keeping track of the length of this slice, because keeping track of 2 variables separately is more bug-prone than just 1. If you'd prefer, I could introduce a little struct for tracking the successful reservations with a `.add()` function, but I'd prefer to just rely on the length of the slice.

iain-macdonald:
Whoops, forgot to re-request review, but Brandon if you could take another look some time (not urgent), I'd appreciate it. Not planning to merge this in this week's release as I'd rather give it a little time in dev.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1818,6 +1780,54 @@ func getNonPreferredSchedulingDelay(cmd *repb.Command) time.Duration {
...
+func enqueueOnConnectedExecutor(ctx context.Context, node *executionNode, request *scpb.EnqueueTaskReservationRequest) bool {
+	if node.handle == nil {
+		log.CtxErrorf(ctx, "nil handle for a local executor %q", node.GetExecutorID())
+		return false // no sleep
+	}
+	_, err := node.handle.EnqueueTaskReservation(ctx, request)
+	if err != nil {
+		log.CtxInfof(ctx, "Failed to enqueue task on connected executor: %s", err)
+	}
+	return err == nil // no sleep
```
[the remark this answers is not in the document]
iain-macdonald: sg, done

PR #5920, opened 2024-02-14
Install podman from pre-downloaded deb archives

bduffany:
This PR removes our dependency on the third-party apt repo during deployment.

- Add a script to pre-package podman deps in a ZIP archive.
- All dependencies from the third-party apt repo (alvistack) are included as `.deb` archives within the ZIP.
- All other deps (i.e. available in the standard apt repos) are just listed in deps.txt within the ZIP.
- The ZIP file also contains an installer that first installs prereqs from deps.txt, then installs the deb packages.
- The Dockerfile now just downloads the ZIP from GCS and just runs the install script, instead of trying to install from apt.

This script is pretty hacky and there is probably a better way to do this, but this should at least hold us over until we can build from source, or until we're prepared to switch the executor to a different Linux distro, or until there are official deb repos available.

**Related issues**: N/A

### DIFF
--- a/dockerfiles/executor_image/Dockerfile
+++ b/dockerfiles/executor_image/Dockerfile
@@ -16,10 +16,22 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - && \
 # Install iproute2 ("ip" command) to configure networking on host.
 RUN apt-get update && apt-get install -y umoci iproute2 amazon-ecr-credential-helper
 
-# Install podman and skopeo
-RUN echo 'deb https://downloadcontent.opensuse.org/repositories/home:/alvistack/Debian_11/ /' >> /etc/apt/sources.list.d/home:alvistack.list && \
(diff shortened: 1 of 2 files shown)


luluz66:
Are there any size differences before and after in terms of how much data executor need to download to install podman?

bduffany:
Somehow the executor image is 400MB smaller after this PR. This has me a bit skeptical. `podman version` works but I haven't yet tested running a container. Will try pushing dev later today

bduffany:
After using [`dive`](https://github.com/wagoodman/dive) to inspect the layer contents, I believe our current layer which installs podman is bloated because of the `apt-get -y upgrade` command. This winds up installing upgraded versions of lots of binaries. If I remove that command from our current Dockerfile and rebuild it, then the image size diff is much smaller (only a few MB)

luluz66 (approved):

iain-macdonald (approved):

PR #5928, opened 2024-02-14
Enable golang static check analyzer S1037: 'Elaborate way of sleeping'

iain-macdonald:
Turns out we're only sleeping un-elaborately rn.

**Related issues**: N/A


tylerwilliams (approved):

iain-macdonald:
I have received multiple pings asking what the "elaborate" way of sleeping is, so here ya go, from the golang staticcheck docs: "Using a select statement with a single case receiving from the result of time.After is a very elaborate way of sleeping that can much simpler be expressed with a simple call to time.Sleep."

PR #5931, opened 2024-02-15
Revert "runfiles refactoring (#5812)"

sluongng:
On MacOS, we deploy our executor as a standalone Go binary instead of a
container image. This causes our runfiles to not get included in the
deployment.

Traditionally this is fine as we embedded the files we need inside the
Go binary itself and switch between the embedded files and runfiles with
fileresolver package. However in #5812, the embedded file path was
replaced with a runfile path. This works when the runfiles are available
but will fail when the runfiles are not (on our MacOS Workflows executor).

Reverts commit 5e9fdd02f8d6f5a392070843e8168ea410ff0db2.


enterprise/server/remote_execution/workspace/workspace.go
```diff
@@ -187,10 +183,14 @@ func (ws *Workspace) AddCIRunner(ctx context.Context) error {
 	if exists {
 		return nil
 	}
+
+	// Make sure that we load the ci runner from embedded bundle instead of runfiles
+	// because we do not deploy MacOS Executor with runfiles.
```
bduffany, line 188: I think this comment is inaccurate since we haven't yet made the change to only load the CI runner from `embed.FS`.

Currently, `ci_runner_bundle` will load from either runfiles or `embed.FS` depending on what's available.

```suggestion
```

bduffany (approved):
(LGTM except for the nit above)

enterprise/server/remote_execution/workspace/workspace.go
sluongng: I would say that it's not a surefire way to achieve this. However, because we are exporting `ci_runner_bundle.BundlePrefix` and using the exported value, I expect `fileresolver.go` to always look up the file inside the embedded bundle.

A better way to achieve this would be to export the embedded FS and use it directly without going through fileresolver. WDYT if we take that route instead?

sluongng: After tinkering with that route, it seems like we have call sites that expect this bundle to act as either embedded fs or runfiles fs. Refactoring this bundle into 2 separate FS could be time-consuming so let's get this merged for now.

bduffany: Looks like the test is failing to find the CI runner binary: https://buildbuddy.buildbuddy.io/invocation/8a482fc8-52c7-438f-9574-33864a71babb?target=%2F%2Fenterprise%2Fserver%2Ftest%2Fintegration%2Fworkflow%3Aworkflow_test&targetStatus=6

can you double check that this PR is a clean revert?

PR #5932, opened 2024-02-15
[SS] Ignore uffd errors due to mapping the same page multiple times

maggie-lou:
If there is a page fault in one process on a multi-core machine, that process will be paused while the page fault is handled, but the other processes will not (https://stackoverflow.com/questions/15886938/can-other-processes-run-during-memory-paging). So if multiple processes on the machine access the same memory location at the same time, they can trigger duplicate page fault notifications for the same address.

(In debug logging I noticed that there were multiple page fault notifications for the exact same GuestFaultingAddr. So it was not an issue related to address math. The page fault handler was literally trying to allocate the same memory address twice.)

Also remove debug logging if we map the same chunk twice, because there are valid cases for this if Firecracker assigns multiple GuestRegionUFFDMappings to the same chunk.

Ex. Chunk 0 has size 100 and contains guest VM address [50-100] [250-300].

**Related issues**: Fixes buildbuddy-io/buildbuddy-internal#3092

### DIFF
--- a/enterprise/server/remote_execution/uffd/uffd.go
+++ b/enterprise/server/remote_execution/uffd/uffd.go
@@ -347,20 +347,21 @@ func (h *Handler) handle(ctx context.Context, memoryStore *copy_on_write.COWStor
 			InvalidBytesAtChunkStart: int64(invalidBytesAtChunkStart),
 			InvalidBytesAtChunkEnd:   invalidBytesAtChunkEnd,
 		}
-		if alreadyMappedRanges, alreadyMapped := h.mappedPageFaults[int64(chunkStartOffset)]; alreadyMapped {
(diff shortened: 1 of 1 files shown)


maggie-lou:
* `master`
  * **PR #5932** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5932?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5932?utm_source=stack-comment).

tylerwilliams (approved):

bduffany (approved):

PR #5934, opened 2024-02-15
[RB] Support passing remote header overrides that apply to the remote runner

maggie-lou:
This will make it easier to test different configurations for the remote runner.

Sample command to run the runner in a custom pool:
`bb remote --remote_runner="grpcs://app.buildbuddy.dev" --runner_exec_properties=Pool=custom build //... --remote_header=x-buildbuddy-api-key=XXX`

**Related issues**: N/A

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -55,17 +55,26 @@ const (
 	gitConfigSection           = "buildbuddy"
 	gitConfigRemoteBazelRemote = "remote-bazel-remote-name"
 	defaultRemoteExecutionURL  = "remote.buildbuddy.io"
+
+	// overrideHeaderPrefix is a prefix used to override platform props via
+	// remote headers. The property name immediately follows the prefix in the
+	// header key, and the header value is used as the property value.
+	//
+	// Example header:
+	//     x-buildbuddy-platform.key: value
+	overrideHeaderPrefix = "x-buildbuddy-platform."
 )
 
 var (
 	remoteFlagset = flag.NewFlagSet("remote", flag.ContinueOnError)
 
-	execOs         = remoteFlagset.String("os", "linux", "If set, requests execution on a specific OS.")
-	execArch       = remoteFlagset.String("arch", "amd64", "If set, requests execution on a specific CPU architecture.")
-	containerImage = remoteFlagset.String("container_image", "", "If set, requests execution on a specific runner image. Otherwise uses the default hosted runner version. A `docker://` prefix is required.")
-	envInput       = bbflag.New(remoteFlagset, "env", []string{}, "Environment variables to set in the runner environment. Key-value pairs can either be separated by '=' (Ex. --env=k1=val1), or if only a key is specified, the value will be taken from the invocation environment (Ex. --env=k2). To apply multiple env vars, pass the env flag multiple times (Ex. --env=k1=v1 --env=k2). If the same key is given twice, the latest will apply.")
-	remoteRunner   = remoteFlagset.String("remote_runner", defaultRemoteExecutionURL, "The Buildbuddy grpc target the remote runner should run on.")
-	timeout        = remoteFlagset.Duration("timeout", 0, "If set, requests that have exceeded this timeout will be canceled automatically. (Ex. --timeout=15m; --timeout=2h)")
+	execOs                = remoteFlagset.String("os", "linux", "If set, requests execution on a specific OS.")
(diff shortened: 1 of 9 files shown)


maggie-lou:
* `master`
  * **PR #5934** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5934?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5934?utm_source=stack-comment).

bduffany:
Since we control the runner action end-to-end, we don't need to use the remote header override mechanism, and can instead set platform props directly.

* The runner can accept a flag like `--runner_exec_properties=NAME=VALUE` (let's call it exec_properties to match bazel, even though they're called platform props in the remote APIs)
* The `RunRequest` gRPC proto can accept an extra `exec_properties` field to directly attach platform properties to the action, instead of using gRPC headers

maggie-lou:
> Since we control the runner action end-to-end, we don't need to use the remote header override mechanism, and can instead set platform props directly.
> 
> * The runner can accept a repeated flag like `--runner_exec_properties=NAME=VALUE` (let's call it exec_properties to match bazel's terminology, even though they're called platform props in the remote APIs)
> * The `RunRequest` gRPC proto can accept an extra `exec_properties` field to directly append platform properties to the `Command`, instead of using gRPC headers

Ack

cli/remotebazel/remotebazel.go
```diff
@@ -538,6 +539,19 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
...
+		keyValArr := strings.SplitN(prop, "=", 2)
+		if len(keyValArr) != 2 {
+			log.Warnf("Invalid exec property %s. Key value pairs must be separated by '='.", prop)
+			continue
+		}
+		key := keyValArr[0]
+		val := keyValArr[1]
+
+		execProps[key] = val
+	}
```
bduffany, line 553: nit: can replace this with `rexec.MakePlatform(*execPropsFlag...)` from `server/util/rexec`

enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -195,6 +195,13 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest,
 		})
 	}
 
+	for k, v := range req.GetExecProperties() {
```
bduffany, line 198: since Go maps are non-deterministic, we might want to sort the platform properties by name after this for-loop, so that the snapshot cache key is deterministic (I think we blindly hash the Platform proto currently and the hash will be affected by the ordering of props)

(we could also directly use the Platform proto for the exec_properties field but I think that will probably make it less user-friendly, since Bazel uses a map for exec_properties)

maggie-lou: Done

What do you think about sorting and deduping the platform propreties in `snaploader` when we generate the key? It may invalidate some snapshots we have now, but will make the system more stable long term. (We could also have a short term migration period where we save snapshots under 2 keys - the existing key and a new one with the props sorted and deduped, and in a week we switch over to exclusively using the new key)

proto/runner.proto
```diff
@@ -94,6 +94,10 @@ message RunRequest {
 
   // Max time before run should be canceled automatically.
   string timeout = 12;
+
+  // Exec properties to apply to the run request.
+  repeated build.bazel.remote.execution.v2.Platform.Property exec_properties =
```
bduffany, line 99: I kind of preferred having this as a `map<string, string>` that you had before (I think?) because it matches how exec properties are specified in bazel.

bduffany: err, nevermind, this is an internal API so it doesn't matter too much.

server/util/rexec/rexec.go
```diff
@@ -63,6 +64,25 @@ func MakePlatform(pairs ...string) (*repb.Platform, error) {
...
+			// are duplicates
+			return j < i
+		}
+		return props[i].Name < props[j].Name
+	})
+	props = slices.CompactFunc(props, func(i, j *repb.Platform_Property) bool {
+		return i.Name == j.Name
+	})
+	return props
+}
```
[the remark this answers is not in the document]
maggie-lou: Ack. Added a TODO and will do in a followup. For now moved the helper function to hostedrunner.go where it is used

bduffany (approved):

PR #5937, opened 2024-02-16
ci_runner/bundle: stop using runfiles

sluongng:
In the past, we used to maintain a separate local Executor build
where `buildbuddy_ci_runner` binary is accessed via runfiles. However,
in production, we ship our Executor as a standalone binary with and
embedded FS inside the binary.

With the upcoming changes to how runfiles are accessed in bzlmod, it
would be simpler for us to retire the local build with runfiles in this
code path.

Make sure that we always use the embedsrcs to access
`buildbuddy_ci_runner` binary.


enterprise/server/cmd/ci_runner/run_local.sh
```diff
@@ -50,20 +50,14 @@ TEMPDIR=$(mktemp --dry-run | xargs dirname)
...
-mkdir -p "$RUNNER_DATA_DIR"
-rm -f "$RUNNER_DATA_DIR/ci_runner"
-cp "$RUNNER_PATH" "$RUNNER_DATA_DIR/ci_runner"
-echo "Copied ci_runner to $RUNNER_DATA_DIR/ci_runner"
-
 mkdir -p "$CI_RUNNER_BAZEL_CACHE_DIR"
 
+# Build the ci_runner_image locally an load it into Docker daemon
+bazel run enterprise/server/cmd/ci_runner:ci_runner_image
+
```
sluongng, line 57: this script is a bit old so Im not too sure on how to update it accordingly 🤔

bduffany: This script isn't very useful anymore and isn't actively maintained so feel free to delete it. These days it's more convenient to use `bb remote` or `ExecuteWorkflow` (either via API or UI) to test the CI runner locally.

enterprise/server/cmd/ci_runner/bundle/bundle.go
```diff
@@ -3,15 +3,19 @@ package bundle
...
-	"github.com/buildbuddy-io/buildbuddy/server/util/fileresolver"
 )
 
+const RunnerName = "buildbuddy_ci_runner"
+
 // NB: Include everything in bazel `embedsrcs` with `*`.
 //
-//go:embed *
+//go:embed buildbuddy_ci_runner
 var all embed.FS
```
bduffany, line 13: we can simplify by changing this to

`var RunnerBytes []byte`

instead of

`var all embed.FS`

Then we wouldn't need to deal with errors anywhere since the error check would effectively be moved to compile time. In `workspace.go` instead of copying from one file to another, we can do a single `os.WriteFile` call.

sluongng: It has been a while but I'm back!

could you please take another look?

bduffany (approved):
LGTM

and welcome back!

PR #5939, opened 2024-02-16
Log action resource name in bb execute debug logs

bduffany:
This is useful to be able to debug the action afterwards (e.g. fetch the full ActionResult from AC using `tools/cas`).

**Related issues**: N/A

### DIFF
--- a/cli/execute/execute.go
+++ b/cli/execute/execute.go
@@ -149,6 +149,8 @@ func execute(cmdArgs []string) error {
 		return err
 	}
 	log.Debugf("Uploaded inputs in %s", time.Since(stageStart))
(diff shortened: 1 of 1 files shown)


cli/execute/execute.go
```diff
@@ -149,6 +149,8 @@ func execute(cmdArgs []string) error {
 		return err
 	}
 	log.Debugf("Uploaded inputs in %s", time.Since(stageStart))
+	actionStr, _ := digest.ResourceNameFromProto(arn).DownloadString()
```
iain-macdonald, line 152: Probably want to check this error, just in case, no?

iain-macdonald (approved):

cli/execute/execute.go
bduffany: Done

PR #5947, opened 2024-02-20
cli: fix starlark flags not expanding from common configs

bduffany:
**Related issues**: fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3128


siggisim (approved):

maggie-lou (approved):
LGTM - that ticket said we should look for "@" too. Is that something we should add?

bduffany:
I realized that checking for "@" is not needed since those flags will also contain "/".

PR #5950, opened 2024-02-20
[raft] retry when one replica is unavailable

luluz66:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3125


enterprise/server/raft/sender/sender.go
```diff
@@ -188,6 +188,10 @@ func (s *Sender) tryReplicas(ctx context.Context, rd *rfpb.RangeDescriptor, fn r
 				break
 			}
 		}
+		if status.IsUnavailableError(err) {
+			//try a different replica if the current replica is not available.
```
tylerwilliams, line 192: ubernit: // try a different replica if the current replica is not available.

tylerwilliams (approved):

PR #5954, opened 2024-02-20
Move nodehost ownership into the store.

tylerwilliams:
This "flattens" the hierarchy a little bit by making the store directly responsible for a lot more of the units it needs. And this means the outer raft_cache.go layer can be simplified a bit.


enterprise/server/raft/store/store.go
```diff
@@ -97,7 +102,52 @@ type Store struct {
 	egCancel context.CancelFunc
 }
 
-func New(env environment.Env, rootDir string, nodeHost *dragonboat.NodeHost, gossipManager interfaces.GossipService, sender *sender.Sender, registry registry.NodeRegistry, listener *listener.RaftListener, apiClient *client.APIClient, grpcAddress string, partitions []disk.Partition) (*Store, error) {
+// We need to provide a factory method that creates the DynamicNodeRegistry, and
```
luluz66, line 105: registryHolder implements NodeRegistryFactory. ....When nodeHost is created, it will call Create() to create the registry ....

enterprise/server/raft/bringup/bringup.go
```diff
@@ -81,28 +78,11 @@ func New(nodeHost *dragonboat.NodeHost, grpcAddr string, createStateMachineFn db
 
 func (cs *ClusterStarter) markBringupComplete() {
 	cs.doneOnce.Do(func() {
-		cs.log.Infof("Bringup is complete on %s", cs.nodeHost.ID())
+		cs.log.Infof("Bringup is complete!")
```
luluz66, line 81: Info

luluz66 (approved):

PR #5955, opened 2024-02-20
Address two nits (thanks autosubmit! you are so good)

tylerwilliams:


enterprise/server/raft/store/store.go
```diff
@@ -102,10 +102,8 @@ type Store struct {
 	egCancel context.CancelFunc
 }
 
-// We need to provide a factory method that creates the DynamicNodeRegistry, and
-// hand this to the raft library when we set things up. When nodeHost is created
-// it will call this method to create the registry and use it until nodehost
-// close.
+// registryHolder implements NodeRegistryFactory. WWhen nodeHost is created, it
```
luluz66, line 105: WWhen -> When

luluz66 (approved):

PR #5959, opened 2024-02-21
Fix invalid usage of uuid in firecracker code

maggie-lou:
I realized that u.String() will always return the same value, when I intended for it to generate a new value.

I will refactor vmID and snapshotID in a follow up, so it's more clear that snapshotID is a unique per-run value, and vmID should only be generated at the beginning of a VM's lifecycle.


maggie-lou:
* **#5959** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5959?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5959?utm_source=stack-comment).

PR #5960, opened 2024-02-21
Add linux-arm64 release binaries

bduffany:
Test build (tested everything except the `gh release upload` command): https://github.com/buildbuddy-io/buildbuddy/actions/runs/7993652037/job/21829927623?pr=5960.

I ran `which gh` to sanity check that `gh` was installed. Looks like it is not yet installed by default on the arm64 runners, so I added an installation step for that.

We probably also should add docker image builds - can do that in a separate PR.

**Related issues**: N/A

### DIFF
--- a/.github/workflows/build-linux-amd64-github-release-artifacts.yaml
+++ b/.github/workflows/build-linux-amd64-github-release-artifacts.yaml
@@ -1,4 +1,4 @@
-name: "Build Linux Github Release Artifacts"
+name: "Build Linux amd64 Github Release Artifacts"
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


maggie-lou (approved):

tylerwilliams (approved):

tylerwilliams:
you want to file a bug on that repo (from the email) letting them know the installed packages are not the same?

tylerwilliams:
Also thanks for doing this! will be nice to have :)

bduffany:
(Waiting to submit this until I have time to do a test build with the arm64 binaries.)

PR #5971, opened 2024-02-22
Implement delete in code editor

siggisim:
**Related issues**: N/A


enterprise/app/code/code.tsx
```diff
@@ -45,7 +45,7 @@ interface State {
   fullPathToModelMap: Map<string, monaco.editor.ITextModel>;
   fullPathToDiffModelMap: Map<string, monaco.editor.IDiffEditorModel>;
   originalFileContents: Map<string, string>;
-  changes: Map<string, string>;
+  changes: Map<string, string | null>;
```
bduffany, line 48: nit: maybe add a comment here like

```suggestion
  /** Map of file path to patch contents, or null if the file is being deleted. */
  changes: Map<string, string | null>;
```

(hopefully I understood this right lol)

bduffany (approved):

PR #5974, opened 2024-02-22
Clean up pull request flow

siggisim:
- Reset review request state in case of a backend error, and print the error message
- Reset update pr state in case of a backend error, and print the error message
- Update fetch latest content after merging a PR
- Only count merge conflicts if file contents are different than editor file contents
- If on a PR branch and you update the default branch and have no changes - switch back to the default branch (this likely means the PR was merged outside of the code editor UI)
- If on a PR branch, show the PR branch in the breadcrumbs
- Fix branch ref when updating PR branch

### DIFF
--- a/enterprise/app/code/code.tsx
+++ b/enterprise/app/code/code.tsx
@@ -540,6 +540,7 @@ export default class CodeComponent extends React.Component<Props, State> {
 
   handleDeleteClicked(fullPath: string) {
     this.state.changes.set(fullPath, null);
+    this.state.pathToIncludeChanges.set(fullPath, true);
     this.updateState({ changes: this.state.changes });
   }
 
--- a/enterprise/app/code/code.tsx
+++ b/enterprise/app/code/code.tsx
@@ -28,6 +28,7 @@ import Long from "long";
 import ModuleSidekick from "../sidekick/module/module";
 import BazelVersionSidekick from "../sidekick/bazelversion/bazelversion";
 import BazelrcSidekick from "../sidekick/bazelrc/bazelrc";
+import error_service from "../../../app/errors/error_service";
 
 interface Props {
   user: User;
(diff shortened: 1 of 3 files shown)


enterprise/app/code/code.tsx
```diff
@@ -626,6 +639,7 @@ export default class CodeComponent extends React.Component<Props, State> {
       .then(() => {
         window.open(this.state.prLink, "_blank");
         this.handleClearPRClicked();
+        this.handleUpdateCommitSha(() => {});
```
bduffany, line 642:
```suggestion
        this.handleUpdateCommitSha();
```

enterprise/app/code/code.tsx
```diff
@@ -647,7 +661,7 @@ export default class CodeComponent extends React.Component<Props, State> {
           head: this.state.repoResponse?.defaultBranch,
         })
       )
-      .then((response) => {
+      .then(async (response) => {
         console.log(response);
         let newCommits = response.aheadBy;
         let newSha = response.commits.pop()?.sha;
```
bduffany, line 667: not an issue with this PR specifically but FYI I think `newCommits == new Long(0)` will always return false on line 668 below - probably want either `Number(newCommits) === 0` or `Long.ZERO.equals(newCommits)`

bduffany (approved):

enterprise/app/code/code.tsx
siggisim: Done

enterprise/app/code/code.tsx
siggisim: There's a comment on `handleUpdateCommitSha` that says:
```
  // If a callback is set, alert messages will not be shown.
```

And I don't want to show a `You're already up to date!` toast in this case.

PR #5977, opened 2024-02-22
Proto changes to fetch less usage data

bduffany:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3139

### DIFF
--- a/proto/usage.proto
+++ b/proto/usage.proto
@@ -7,21 +7,25 @@ import "proto/context.proto";
 message GetUsageRequest {
   // Request context.
   context.RequestContext request_context = 1;
+
+  // The usage period to be fetched, as a UTC month in "YYYY-MM" format.
+  // If empty, the current usage period will be returned.
+  string usage_period = 2;
 }
 
 message GetUsageResponse {
   // Response context.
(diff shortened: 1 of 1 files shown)


proto/usage.proto
```diff
@@ -7,21 +7,25 @@ import "proto/context.proto";
...
   //
-  // If historical data exists, the server may return historical data from some
-  // recent number of months (clients should handle any number of months).
+  // TODO(bduffany): make this non-repeated once the client and server are
+  // updated.
   repeated Usage usage = 2;
+
+  // The available usage periods that may be specified in a subsequent
+  // GetUsageRequest.
+  repeated string available_usage_periods = 3;
```
jdhollen, line 28: How will you do this without still doing a full scan?

jdhollen: Oh, I guess you can just do a separate query for the min entry, nevermind.

jdhollen (approved):

proto/usage.proto
bduffany: I was just going to have the server return the last 6 months, and if there's no data for a particular period then we return 0. I think it's equivalent to what we have today

PR #5987, opened 2024-02-23
[RB] Remove session affinity key because it prevents snapshot sharing

maggie-lou:
I noticed that in a github workflow using remote bazel, the snapshots were never recycled. There is no guarantee that workflows will be run on stable machines with stable node IDs and dir structure. The SessionAffinityKey is set as a platform prop, which is hashed for the snapshot key, so the changing value was causing snapshot misses.

**Related issues**: buildbuddy-io/buildbuddy-internal#3140

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -3,7 +3,6 @@ package remotebazel
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
 	"flag"
 	"fmt"
 	"io"
@@ -35,7 +34,6 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/google/uuid"
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc/metadata"
 
@@ -501,10 +499,6 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 
 	ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-api-key", opts.APIKey)
 
-	instanceHash := sha256.New()
-	instanceHash.Write(uuid.NodeID())
-	instanceHash.Write([]byte(repoConfig.Root))
-
 	reqOS := runtime.GOOS
 	if *execOs != "" {
 		reqOS = *execOs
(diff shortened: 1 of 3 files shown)


maggie-lou:
* **#5987** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5987?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5987?utm_source=stack-comment).

cli/remotebazel/remotebazel.go
```diff
@@ -501,10 +499,6 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 
 	ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-api-key", opts.APIKey)
 
-	instanceHash := sha256.New()
-	instanceHash.Write(uuid.NodeID())
-	instanceHash.Write([]byte(repoConfig.Root))
```
bduffany, line 506: Should we keep this around for local machines only, so that users won't get assigned to other users' snapshots unexpectedly? On GitHub, the env var `CI=true` will be set, so maybe if CI=true then we can set `instanceHash = ""` ?

bduffany: it seems like it might also be useful to keep `repoConfig.Root` around as the instance hash, regardless of whether CI=true or not

maggie-lou: Yeah I was going back and forth if it would be more beneficial or annoying to use others' snapshots, but I think it would be more beneficial in the longterm. It would result in faster build times for everyone. It would be like a global analysis cache. If you want to use a user-specific snapshot, you can always set something like --runner_exec_properties=user=maggie

I logged `repoConfig.Root` and it includes the full directory path. I think that the repo path on github runners can change. [Here](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/hostedrunner/hostedrunner.go#L157)we default the affinity key to the URL, which seems sufficient to me

bduffany: If you're running `bb remote` for the first time, then it will be nice to be matched up to a warm analysis cache, but when running `bb remote` multiple times in succession (i.e. interactively), if you have a lot of users doing that, then Bazel will constantly be switching Git branches and there will be lots of cache evictions. I think that's what the original intent of the "session" key was - we'd have per-user sessions that would not get clobbered by other users' sessions, so when you are running `bb` iteratively, you keep getting matched up the same VM and it feels like you're just developing locally.

maggie-lou: We do include the branch name in the snapshot key, so this feels like a reasonable trade off to me? I'd assume most users will be on a branch

The dream is that remote bazel would be even better than developing locally. It would be awesome if you always hit a warm bazel instance, even if your local machine would not have

bduffany (approved):

cli/remotebazel/remotebazel.go
vadimberezniker: I have the same workflow as Brandon (working off master). Doesn't hurt to experiment though.

vadimberezniker (approved):

cli/remotebazel/remotebazel.go
maggie-lou: Oh interesting. An idea for the future is we could see if a snapshot exists for a user session and only if it doesn't do we use a more general snapshot. Similar idea to what we do with fallback github branch keys now

PR #5990, opened 2024-02-23
don't call raft API when the deadline has approached

luluz66:
**Related issues**: N/A


enterprise/server/raft/client/client.go
```diff
@@ -110,7 +114,12 @@ func RunNodehostFn(ctx context.Context, nhf func(ctx context.Context) error) err
 			break
 		}
 
-		opCtx, cancel := context.WithTimeout(ctx, singleOpTimeout(ctx))
+		timeout := singleOpTimeout(ctx)
+		if timeout <= 0 {
+			// The deadline has already approached.
```
tylerwilliams, line 119: s/approached/passed/g?

tylerwilliams (approved):

PR #5994, opened 2024-02-26
Add filecache metrics to IOStats proto

bduffany:
This should provide a little more insight into input fetch times in the case where 0 files are downloaded from remote cache.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3142

### DIFF
--- a/proto/remote_execution.proto
+++ b/proto/remote_execution.proto
@@ -2177,7 +2177,9 @@ message IOStats {
   // than downloading from the remote cache.
   int64 local_cache_hits = 7;
 
-  // Total time spent linking inputs from local cache, across all threads.
+  // Time spent linking inputs from local cache, totalled across all threads.
+  // For example, if 8 concurrent threads each spend 2ms of execution time
+  // hardlinking files, this will be 16ms.
   google.protobuf.Duration local_cache_duration = 8;
 
   // The number of files uploaded in this tree.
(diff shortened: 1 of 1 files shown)


proto/remote_execution.proto
```diff
@@ -2173,6 +2173,13 @@ message IOStats {
   // The time taken to download the tree.
   int64 file_download_duration_usec = 3;
 
+  // Total number of inputs that were provisioned from the local cache rather
+  // than downloading from the remote cache.
+  int64 local_cache_hits = 7;
+
+  // Total time spent linking inputs from local cache, across all threads.
+  google.protobuf.Duration local_cache_duration = 8;
```
tylerwilliams, line 2181: The comment is unclear to me -- is this a wall time or sum(all thread times)?

I think we want walltime.

bduffany: Updated the comment to clarify - it's the sum of times across all threads.

I think wall time could be defined as something like `end_timestamp_of_last_hardlink - start_timestamp_of_first_hardlink` but I'm not sure how useful that would be, because today we have remote fetches interleaved with hardlinks. If there is at least 1 remote fetch in the action, then I think the wall time becomes hard to interpret?

The sum of all thread times is also kind of hard to interpret, but I think it more directly correlates to the amount of "work" that we had to do in order to set up the input root.

tylerwilliams: The thing that's confusing about this is that there are other metrics in this proto, like `file_download_duration_usec` which are walltime measures.

The value of the walltime measure (I think?) is that it is directly measuring how long the user is waiting for something. It's also easy to interpret, where sum(all thread times) is hard to interpret especially because you don't have number of threads.

tylerwilliams (approved):

PR #5995, opened 2024-02-26
Render IOStats in action page

bduffany:
Renders the existing fields as well as the new filecache fields added in https://github.com/buildbuddy-io/buildbuddy/pull/5994. The existing fields were shown in the executions page but not when drilling down to the action page.

---

Screenshot:

image

---

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3142


siggisim (approved):
LGTM (heads up you're not merging into master)

PR #6000, opened 2024-02-26
Add rusage to UsageStats proto

bduffany:
The IO stats (`inblock` / `outblock`) seem to be somewhat reasonable:

```
---
Command: yes | head -c $(( 512 * 1000 )) > /tmp/yes.txt; sync /tmp/yes.txt
Inblock: 496
Oublock: 1000
---
Command: echo 3 > /proc/sys/vm/drop_caches; cat <<<'' 
Inblock: 2312
Oublock: 0
---
Command: echo 3 > /proc/sys/vm/drop_caches; cat /tmp/yes.txt | cat
Inblock: 4104
Oublock: 0
```

**Related issues**: N/A

### DIFF
--- a/proto/execution_stats.proto
+++ b/proto/execution_stats.proto
@@ -11,49 +11,6 @@ import "proto/stat_filter.proto";
 
 package execution_stats;
 
-// Proto representation of the rusage struct used in the Linux/BSD getrusage()
-// system call.
-message Rusage {
-  // ru_utime: Amount of CPU time in seconds spent in userspace.
-  int64 user_cpu_time_usec = 1;
-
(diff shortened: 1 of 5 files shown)


proto/execution_stats.proto
```diff
@@ -11,49 +11,6 @@ import "proto/stat_filter.proto";
 
 package execution_stats;
```
bduffany, line 13: moved this to `remote_execution.proto` to break a circular dep (this file depends on remote_execution.proto)

bduffany:
ugh, turns out this doesn't work on macOS for some reason, only Linux. inblock/oublock always report 0. Will find another way to report IO stats on macOS.

PR #6011, opened 2024-02-27
Update replay_action to allow overriding the command

bduffany:
This is useful to replay an action targeting the same VM snapshot but with a different command so that we can run debug commands within the VM.

**Related issues**: N/A


maggie-lou (approved):
Nice!

PR #6010, opened 2024-02-27
Removes unnecessary select and fixes potential time-of-check, time-of-use problem

tempoz:
Rather than first performing a select to determine if we should insert or update, we can attempt an insert which does nothing on conflict and then update if the query executes without error but affects no rows.

**Related issues**: N/A

PR #6002, opened 2024-02-26
Only perform the select to determine if there is no related execution row after attempting an update

tempoz:
This makes our most-common codepath only use one query instead of two

**Related issues**: N/A

PR #6012, opened 2024-02-27
Uses a single query instead of two queries to retrieve a user by subid

tempoz:
Obviates the need for a transaction to maintain READ REPEATABLE isolation and also reduces the queries required for the operation.

**Related issues**: N/A


tempoz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6012?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#6012** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6010** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6010?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6003** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6003?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5865** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5865?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#5875](https://github.com/buildbuddy-io/buildbuddy/pull/5875) <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/5875?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6002** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6002?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6012?utm_source=stack-comment).

bduffany (approved):
LGTM (friendly reminder, need a second LGTM since it's [auth code](https://www.buildbuddy.dev/code-reviews#auth-code))

enterprise/server/backends/authdb/authdb.go
```diff
@@ -429,32 +430,46 @@ func (d *AuthDB) GetAPIKeyGroupFromAPIKeyID(ctx context.Context, apiKeyID string
 }
 
 func (d *AuthDB) LookupUserFromSubID(ctx context.Context, subID string) (*tables.User, error) {
-	user := &tables.User{}
-	// TODO(zoey): switch this to JOIN
-	err := d.h.TransactionWithOptions(ctx, db.Opts().WithStaleReads(), func(tx interfaces.DB) error {
-		rq := tx.NewQuery(ctx, "authdb_lookup_user_from_subid").Raw(
```
vadimberezniker, line 435: Is there a practical situation where having these two selects outside of a single transaction would cause problems? Seems that would be simpler than trying to combine everything into a single complex query.

tempoz: Is this that much more complicated? The second query is already doing an implicit inner join of `Groups` and `UserGroups`; this is just switching that to a left join of the column of `Users` that matches the `sub_id`, `UserGroups`, and `Groups`. It should be about as fast and be only one query as opposed to multiple (which reduces our overall qps, though admittedly not by that much as this is by far not the bulk of the queries we make).

I agree that removing the transaction here without merging the queries is likely safe, given that the user returned by the initial query is unlikely to change between the two queries, if it exists, and if it doesn't exist the second query will never run. I think the one potential weird failure mode is if a user was deleted between the two queries, we would get back a user that was a member of no groups, which would be a weird, inconsistent state of that user that never should have existed, but it would be unlikely to actually cause any problems, since it would just mean that user would not authenticate as a member of any groups.

All of that said, this is hardly a critical change, it just felt like low-hanging fruit to consolidate queries that didn't need to be a series of separate queries. If you think it's an issue, I can just close the PR.

vadimberezniker: I think it wouldn't look as complicated if we made the index on sub_id unique at which point it would become 3 plain joins. There's no reason for there to be multiple users with the same sub_id. 

The other complication is that you need additional logic to differentiate the case of "no user" and "user and no groups" due to having the data returned in a single query. 

This is not a performance sensitive code path so I don't see the incentive to try to force everything into a single query.

tempoz: Works for me, thanks for your input!

I might open an issue on unique-ifying the sub_id

vadimberezniker (approved):

tylerwilliams (approved):
LGTM, Approval

PR #6022, opened 2024-02-28
Fix GH actions auth

bduffany:
Set an API key on GH actions runner executions.

This worked without an API key when I was testing locally because I was running the executor with `--debug_enable_anonymous_runner_recycling`, but in dev I got an error because runner recycling requires auth.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/githubapp/githubapp.go
+++ b/enterprise/server/githubapp/githubapp.go
@@ -46,6 +46,7 @@ import (
 	"github.com/google/go-github/v59/github"
 	"golang.org/x/oauth2"
 	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	gh_webhooks "github.com/buildbuddy-io/buildbuddy/enterprise/server/webhooks/github"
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):

enterprise/server/githubapp/githubapp.go
```diff
@@ -273,6 +274,29 @@ func (a *GitHubApp) startGitHubActionsRunnerTask(ctx context.Context, event *git
 		return status.FailedPreconditionError("workflow job cannot be nil")
 	}
 
+	log.CtxInfof(ctx, "Starting ephemeral GitHub Actions runner execution for %s/%s", event.GetRepo().GetOwner().GetLogin(), event.GetRepo().GetName())
+
+	// Get an org API key for the installation. The user must have linked the
+	// installation to an org via the BuildBuddy UI, otherwise this will return
+	// NotFound.
+	installation := &tables.GitHubAppInstallation{}
+	err := a.env.GetDBHandle().NewQuery(ctx, "github_app_get_installation_for_workflow_job").Raw(`
```
maggie-lou, line 283: Sanity check - the installation ID is some sort of secret right? i.e. There'd be no way for people to hack into authenticating as another group with this?

bduffany: (resolved offline)

maggie-lou (approved):

PR #6039, opened 2024-02-29
[Raft] flush the batch more frequently when recovering from snapshot

luluz66:
**Related issues**:
https://github.com/buildbuddy-io/buildbuddy-internal/issues/3153


enterprise/server/raft/replica/replica.go
```diff
@@ -1849,11 +1864,18 @@ func (sm *Replica) ApplySnapshotFromReader(r io.Reader, db ReplicaWriter) error
 		if err := wb.Set(kv.Key, kv.Value, nil); err != nil {
 			return err
 		}
+		batchSize += n
```
tylerwilliams, line 1867: Can we use https://pkg.go.dev/github.com/cockroachdb/pebble#Batch.Len instead?

tylerwilliams (approved):
LGTM tho

PR #6053, opened 2024-03-01
change exported metrics from gauge to counter

luluz66:
Change the exported metric types from gauge to counter; and also add a label "pod_name" to exported metrics.

Currently, the exported metric is the sum of all counters from the all the apps. During the rollout, the metrics from one app will reset to zero one by one. As a result, the aggregation never reset to zero.

### DIFF
--- a/enterprise/server/backends/prom/prom.go
+++ b/enterprise/server/backends/prom/prom.go
@@ -269,6 +269,10 @@ func (q *promQuerier) fetchMetrics(ctx context.Context, groupID string) (map[str
 			queryParams = append(queryParams, &promQueryParams{
 				metricName: config.sourceMetricName, sumByFields: config.LabelNames,
 			})
+		case dto.MetricType_COUNTER:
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):

enterprise/server/backends/prom/prom.go
```diff
@@ -39,7 +39,7 @@ var (
 			ExportedFamily: &dto.MetricFamily{
 				Name: proto.String("exported_buildbuddy_remote_execution_queue_length"),
 				Help: proto.String("Number of actions currently waiting in the executor queue."),
-				Type: dto.MetricType_GAUGE.Enum(),
+				Type: dto.MetricType_COUNTER.Enum(),
```
bduffany, line 42: should this one be a gauge still? the queue length metric we have for the executors today is a gauge

luluz66: Done

tylerwilliams (approved):

PR #6063, opened 2024-03-04
Support passing a commit sha in the ExecuteWorkflow API

maggie-lou:
This PR adds support for specifying a commit sha to checkout in the ExecuteWorkflow API.

It explicitly differentiates (1) the git ref (can be a commit sha or a branch) the CI runner should checkout, from (2) reporting metadata for the workflow (i.e. used for reporting Github statuses, populating populate UI details)

For (1), it uses a new flag `pushed_ref` that has been added to the ci_runner binary. For (2), it uses the existing flags `pushed_branch` and `commit_sha`, which will now be used for reporting only.

Because we previously conflated the two, there were cases when we would not let users specify a commit sha to checkout, because the behavior was confusing / not consistent when they wanted to specify both a branch and commit sha.

[This will also let customers report Github statuses for different commits than the ones being tested. This may be helpful if they have custom merge behavior in their CI pipelines](https://buildbuddy-corp.slack.com/archives/C057TAUAQ7P/p1701995207224059)


maggie-lou:
* **#6063** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6063?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

docs/enterprise-api.md
```diff
@@ -797,6 +798,17 @@ curl -d '{
 -H "x-buildbuddy-api-key: YOUR_BUILDBUDDY_API_KEY" \
 -H 'Content-Type: application/json' \
 https://app.buildbuddy.io/api/v1/ExecuteWorkflow
+
+# Execute workflow on commit ABC on branch cool-feature
+curl -d '{
+  "repo_url": "https://github.com/buildbuddy-io/buildbuddy-ci-playground",
+  "ref": "ABC",
```
bduffany, line 805: uber nit: make it look a bit more like a commit sha (slightly longer and lowercase)

```suggestion
# Execute workflow on commit abc123 on branch cool-feature
curl -d '{
  "repo_url": "https://github.com/buildbuddy-io/buildbuddy-ci-playground",
  "ref": "abc123",
```

docs/enterprise-api.md
```diff
@@ -835,6 +848,17 @@ message ExecuteWorkflowRequest {
   // overrides will take precedence. Otherwise all env vars set in
   // buildbuddy.yaml will still apply.
   map<string, string> env = 7;
+
+  message ReportingMetadata {
+    string branch = 1;
```
bduffany, line 853: For branch name, can we determine the branch name automatically? I think ideally we would be able to make it "just work" if the user passes a commit SHA which exactly corresponds to an existing remote branch tip.

For commit_sha, what's the use case that you have in mind?

enterprise/server/api/api_server.go
```diff
@@ -526,14 +526,18 @@ func (s *APIServer) ExecuteWorkflow(ctx context.Context, req *apipb.ExecuteWorkf
 	requestCtx := requestcontext.ProtoRequestContextFromContext(ctx)
 
 	wfID := wfs.GetLegacyWorkflowIDForGitRepository(user.GetGroupID(), req.GetRepoUrl())
+	branchName := req.GetReportingMetadata().GetBranch()
+	if branchName == "" {
+		branchName = req.GetRef()
```
bduffany, line 531: It seems like the branch name could wind up being a commit SHA here if the user forgets to pass `branch` in the metadata - will that break anything on the CI runner?

docs/enterprise-api.md
```diff
@@ -835,6 +848,17 @@ message ExecuteWorkflowRequest {
...
+    string commit_sha = 2;
+  }
+  // Github statuses should be reported with these fields.
```
bduffany, line 856: I think it's worth clarifying what kinds of reporting we're referring to - there is info reported both to BB (build metadata) and GH (commit status)

docs/enterprise-api.md
maggie-lou: Here's an example of a situation where we won't be able to determine the branch name automatically:
At the beginning of the CI run, we emit a WorkflowConfigured event with a PushedBranch field, that's used to populate the UI. In order to go from commit sha -> branch name, we need a local clone of the repo (using command `git name-rev --name-only <commit sha>` - I couldn't find a remote command for this). We don't want to delay emitting WorkflowConfigured until after we clone the repo, because that can be slow.

maggie-lou: Actually I think I may have found a solution here

maggie-lou:
This PR was getting really big so split into several smaller PRs. Bulk of logic moved here: https://github.com/buildbuddy-io/buildbuddy/pull/6476/commits/0e4eba784e81a2c80f4b780bbb21e8c65a31707d#

docs/enterprise-api.md
maggie-lou: Fixed in the other PR

enterprise/server/api/api_server.go
maggie-lou: That will just mean the metadata in the UI will show that the branch name is the commit SHA, which is not ideal and may be confusing, but shouldn't break anything else

PR #6068, opened 2024-03-04
Fix issue where "resumed from" link breaks inside the same invocation

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -76,7 +76,6 @@ export default class InvocationActionCardComponent extends React.Component<Props
       }
     }
     if (prevProps.search.get("executeResponseDigest") != this.props.search.get("executeResponseDigest")) {
-      this.setState({ executeResponse: undefined });
       this.fetchExecuteResponse();
     }
   }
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_action_card.tsx
```diff
@@ -68,6 +68,19 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
 
+  componentDidUpdate(prevProps: Readonly<Props>, prevState: Readonly<State>, snapshot?: any): void {
+    if (prevProps.search.get("actionDigest") != this.props.search.get("actionDigest")) {
+      this.fetchAction();
+      if (!this.props.search.has("executeResponseDigest")) {
+        this.fetchActionResult();
+      }
+    }
+    if (prevProps.search.get("executeResponseDigest") != this.props.search.get("executeResponseDigest")) {
+      this.setState({ executeResponse: undefined });
```
bduffany, line 79: nit: should we do this as the first line of `fetchExecuteResponse()` ? (since callers would always want to clear out the existing state I think?)

bduffany (approved):

app/invocation/invocation_action_card.tsx
jdhollen: Sure, done

PR #6071, opened 2024-03-04
[Raft] use sender.SyncRead directly instead of sender.Run

luluz66:
**Related issues**: N/A
sender.Run -> sender.run since apart from the usage in raft/cache, we don't use 
sender.Run outside sender any more, and instead we shall use sender.SyncRead or
sender.SyncPropose.


tylerwilliams (approved):
nice

PR #6076, opened 2024-03-05
Add more logging to vmexec server

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/vmexec/vmexec.go
+++ b/enterprise/server/vmexec/vmexec.go
@@ -1,6 +1,7 @@
 package vmexec
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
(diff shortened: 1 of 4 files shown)


enterprise/server/vmexec/vmexec.go
```diff
@@ -322,11 +361,45 @@ func newCommand(start *vmxpb.ExecRequest) (*command, error) {
...
 
+func (c *command) logDebugInfo(err error) {
+	log.Warningf("Command terminated abnormally: %s", err)
+	// Check integrity of the executable
+	if len(c.cmd.Args) > 0 {
+		executable := c.cmd.Args[0]
+
+		realpath, err := exec.LookPath(executable)
+		if err != nil {
+			log.Warningf("Failed to look up executable in $PATH: %s", err)
```
maggie-lou, line 372: Should we exit early here?

enterprise/server/vmexec/vmexec.go
```diff
@@ -322,11 +361,45 @@ func newCommand(start *vmxpb.ExecRequest) (*command, error) {
 	}, nil
 }
 
+func (c *command) logDebugInfo(err error) {
```
maggie-lou, line 364: Should we do some error filtering here? There can be lots of uninteresting reasons a command fails

enterprise/server/vmexec/vmexec.go
bduffany: Done

enterprise/server/vmexec/vmexec.go
bduffany: Done, filtering to "exec format error" for now

jdhollen (approved):

PR #6080, opened 2024-03-05
Ignore invalid flags in the ci_runner script

maggie-lou:
This will make it easier to add flags to this script in the future.

Right now, if you pass an undefined flag to this script, the script will immediately fail when `flag.Parse` encounters the unexpected flag. This will make it challenging to ever iterate on this script and add new flags, especially because customers with self-hosted executors may rarely update their executors to pull in new changes to this script. The workflow service / our schedulers don't know what version of this script will be on an executor, so it won't be able to tell whether it's safe to send a new flag.

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -131,39 +131,41 @@ const (
 )
 
 var (
-	besBackend         = flag.String("bes_backend", "", "gRPC endpoint for BuildBuddy's BES backend.")
-	cacheBackend       = flag.String("cache_backend", "", "gRPC endpoint for BuildBuddy Cache.")
-	rbeBackend         = flag.String("rbe_backend", "", "gRPC endpoint for BuildBuddy RBE.")
-	besResultsURL      = flag.String("bes_results_url", "", "URL prefix for BuildBuddy invocation URLs.")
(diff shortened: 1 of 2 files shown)


maggie-lou:
* **#6080** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6080?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -656,6 +660,27 @@ func run() error {
 	return nil
 }
 
+// parseFlags should not fail when parsing an undefined flag.
+// This lets us add new flags to this script without breaking older executors
+// (on self-hosted executors, for example) that aren't expecting them when the
+// app server tries to send them.
+func parseFlags() error {
+	// Silence noisy output when parsing an unknown flag
```
bduffany, line 668: nit: this would let us avoid having to use flagset above

```suggestion
	// Ignore errors when parsing an unknown flag
	flagset := flag.CommandLine
	flagset.Init(os.Args[0], flag.ContinueOnError)
```

bduffany (approved):

enterprise/server/cmd/ci_runner/main.go
maggie-lou: Done

PR #6085, opened 2024-03-06
[RB] Try to set default branch env var for better snapshot matching

maggie-lou:
The snapshot key is based off the git branch. If we do not have a snapshot for a branch, we support falling back to snapshots for default branches, as specified in the `GIT_REPO_DEFAULT_BRANCH` and `GIT_BASE_BRANCH` env vars.
This change will enable more snapshot hits.

**Related issues**: N/A

### DIFF
--- a/.github/workflows/remote-bazel-build.yaml
+++ b/.github/workflows/remote-bazel-build.yaml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Test
         run: |
-          bb remote test //... \
+          bb remote --env=GIT_REPO_DEFAULT_BRANCH=master test //... \
             --config=linux-workflows --config=race \
             --remote_executor=grpcs://buildbuddy.buildbuddy.io \
             --test_tag_filters=-performance,-webdriver,-docker,-bare \
--- a/enterprise/server/hostedrunner/hostedrunner.go
+++ b/enterprise/server/hostedrunner/hostedrunner.go
@@ -2,7 +2,6 @@ package hostedrunner
 
 import (
 	"context"
-	"encoding/json"
 	"io"
 	"io/fs"
 	"path/filepath"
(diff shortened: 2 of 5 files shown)


enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -197,6 +188,13 @@ func (r *runnerService) createAction(ctx context.Context, req *rnpb.RunRequest,
 		})
 	}
 
+	for k, v := range req.GetEnv() {
```
maggie-lou, line 191: When generating fallback snapshot keys, we only look at cmd.EnvironmentVariables, and not at the --env_overrides flag

.github/workflows/remote-bazel-build.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       - name: Test
         run: |
-          bb remote test //... \
+          bb remote --env=GIT_REPO_DEFAULT_BRANCH=master test //... \
```
bduffany, line 25: revert before merging?

enterprise/server/hostedrunner/hostedrunner.go
bduffany: Makes sense - do we need `env_overrides` anymore after this change or can we remove it?

bduffany (approved):

enterprise/server/hostedrunner/hostedrunner.go
maggie-lou: No we don't. Cleaned up in workflow service too. Don't know why I didn't implement this way to begin with...

.github/workflows/remote-bazel-build.yaml
maggie-lou: The remote bazel code determines the default branch from the local git state. The github runner only checks out one commit, so we'd have to do that hacky thing where we briefly checkout master in order to take advantage of that. It's easier to just explicitly pass the flag

PR #6090, opened 2024-03-07
Get TreeCache and DirectoryWithDigest from VTPool

luluz66:
Add pooling for TreeCache and DirectoryWithDigest

Add benchmark test
Change TestGetTree depth and branchingFactor to parameters that caused nil
pointer in development.

**Related issues**: N/A


luluz66:
This approach seems very error-prone to bugs.

PR #6091, opened 2024-03-07
newhash readonly

tylerwilliams:
- Add new hash and readonly params to distributed cache.
- Buildfix.


siggisim (approved):

enterprise/server/backends/distributed/distributed.go
```diff
@@ -35,11 +35,17 @@ var (
 	redisTarget                  = flag.String("cache.distributed_cache.redis_target", "", "A redis target for improved Caching/RBE performance. Target can be provided as either a redis connection URI or a host:port pair. URI schemas supported: redis[s]://[[USER][:PASSWORD]@][HOST][:PORT][/DATABASE] or unix://[[USER][:PASSWORD]@]SOCKET_PATH[?db=DATABASE] ** Enterprise only **", flag.Secret)
 	groupName                    = flag.String("cache.distributed_cache.group_name", "", "A unique name for this distributed cache group. ** Enterprise only **")
 	nodes                        = flag.Slice("cache.distributed_cache.nodes", []string{}, "The hardcoded list of peer distributed cache nodes. If this is set, redis_target will be ignored. ** Enterprise only **")
+	consistentHashFunction       = flag.String("cache.distributed_cache.consistent_hash_function", "CRC32", "A consistent hash function to use when hashing data. CRC32 or SHA256")
+	consistentHashReplicas       = flag.Int("cache.distributed_cache.consistent_hash_replicas", 100, "The number of copies of each peer on the consistent hash ring")
```
bduffany, line 39: is it too late to rename "replicas" to "virtual nodes"? I found the "replicas" terminology confusing as I was trying to grok `consistent_hash.go`, because we use "replicas" to refer to both the host names (`GetAllReplicas`) as well as the extra "copies" of each server that we place on the consistent hash ring. When reading about consistent hashes online, it seems like the standard terminology for these extra copies is "virtual nodes".

vadimberezniker (approved):

enterprise/server/backends/distributed/distributed.go
tylerwilliams: Works for me.

PR #6092, opened 2024-03-07
Fix dev QA script

maggie-lou:
* Disable bzlmod on builds (and remove `bazel` repo build because it's been fully migrated to bzlmod, and we haven't done enough testing to guarantee our toolchain work well with bzlmod)
* Fail script if a QA build fails
* Add API key to builds
* Update to the latest buildbuddy toolchain

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3167

### DIFF
--- a/tools/dev_qa.py
+++ b/tools/dev_qa.py
@@ -87,30 +87,6 @@ REPO_CONFIGS = [
                 --remote_header=x-buildbuddy-api-key={}
         """.format(API_KEY),
     },
-    # NOTE: We are pinning to this commit sha for the bazel repo because it's
-    # before they fully migrated to bzlmod. Our toolchain has not been fully
-    # migrated to supporting bzlmod yet.
-    {
-        "name": "bazel",
-        "repo_url": "https://github.com/bazelbuild/bazel",
-        "commit_sha": "9df8c97f2fc532c966b612559430a5a248886d1b",
-        "command": """
-            bazel build //src/main/java/com/google/devtools/build/lib/... \
-                --remote_executor=remote.buildbuddy.dev \
-                --remote_cache=remote.buildbuddy.dev \
-                --bes_backend=remote.buildbuddy.dev \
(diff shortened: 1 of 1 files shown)


tools/dev_qa.py
```diff
@@ -24,67 +24,64 @@
...
                 --host_platform=@buildbuddy_toolchain//:platform \
                 --platforms=@buildbuddy_toolchain//:platform \
                 --crosstool_top=@buildbuddy_toolchain//:toolchain \
-                --build_metadata=TAGS=dev-qa
-        """,
+                --remote_header=x-buildbuddy-api-key={}
+        """.format(API_KEY),
     },
     {
-        "name": "bazel",
```
bduffany, line 70: nit: instead of removing `bazel` we could also pin at a known good commit and bazel version. I think in general we probably want to pin all of these to a known good commit (plus a bazel version if a `.bazelversion` is missing from the repo), so that if these repos break then we don't have to block the release

bduffany (approved):
LGTM (we can maybe pin to a commit another time, since that seems like it might be slightly tricky when using the github checkout action)

tools/dev_qa.py
maggie-lou: Done - pinned all the commit shas to current HEAD. For the bazel repo, pinned to the parent commit before they cleared their WORKSPACE file. Also pinned bazel version to 7.0.2

Was going back and forth on whether we should pin the bazel version, because we might want to know if builds are breaking on a later bazel version. But I think pinning it will make things more stable, and we can find that out other ways

bduffany (approved):

tools/dev_qa.py
maggie-lou: I'm actually having trouble getting `bazel` repo to build on an older commit. I want to make sure these fixes get in before the next release, so just going to remove bazel for now. We'll still have a go, c++, and python test build

PR #6113, opened 2024-03-11
blow away previous state when clicking links on workflows

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/components/link/link.tsx
+++ b/app/components/link/link.tsx
@@ -2,7 +2,8 @@ import React from "react";
 import router from "../../router/router";
 
 export type LinkProps = {
-  ignorePreviousState?: boolean;
+  /** Prevents existing filter state from being preserved when navigating. */
+  resetFilters?: boolean;
 } & React.AnchorHTMLAttributes<HTMLAnchorElement>;
 
 /**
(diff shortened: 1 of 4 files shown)


app/components/link/link.tsx
```diff
@@ -1,7 +1,9 @@
 import React from "react";
 import router from "../../router/router";
 
-export type LinkProps = React.AnchorHTMLAttributes<HTMLAnchorElement>;
+export type LinkProps = {
+  ignorePreviousState?: boolean;
```
bduffany, line 5: I think "previous state" might be ambiguous because there are various other pieces of state in the URL, wdyt about `resetFilter` or something like that?

(also nit add a comment)

```suggestion
  /** Prevents existing filter state from being preserved when navigating. */
  resetFilter?: boolean;
```

bduffany (approved):

app/components/link/link.tsx
jdhollen: I like `resetFilters` better than `resetFilter`, but otherwise i updated this here and in `router.tsx`

bduffany (approved):

app/components/link/link.tsx
bduffany: I like that better too :+1:

PR #6126, opened 2024-03-12
Fix executable diagnostics

bduffany:
Also log the first 1K bytes of the executable (I realized that if there is corruption, just knowing the digest of the corrupted contents will not be super helpful)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/vmexec/vmexec.go
+++ b/enterprise/server/vmexec/vmexec.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"sort"
 	"strings"
 	"syscall"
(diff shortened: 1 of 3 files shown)


enterprise/server/vmexec/vmexec.go
```diff
@@ -360,11 +361,19 @@ func newCommand(start *vmxpb.ExecRequest) (*command, error) {
...
-	path, err := exec.LookPath(executable)
-	if err != nil {
-		log.Warningf("Failed to look up executable in $PATH: %s", err)
-		return
+func logExecutableInfo(workDir string, executable string) {
+	var path string
+	// Relative paths like "tools/foo" or "./foo" are resolved relative to the
+	// working directory. Unqualified executable names like "bash" are resolved
+	// using $PATH lookups.
+	if strings.Contains(executable, "/") {
```
jdhollen, line 369: nit, because i don't think we actually have a case where we need to care about it, but this should also exclude executables that start with `/`, right?

jdhollen (approved):

enterprise/server/vmexec/vmexec.go
bduffany: good call, done

PR #6127, opened 2024-03-12
Add metrics for distributed cache peer lookups

bduffany:
Add a histogram metric to record how many peers we're having to consult for each digest, for each distributed cache operation.

For now, just adds metrics for `Reader` and `FindMissing` operations - can later add operations for other operations if we think they'd be useful.

**Related issues**: N/A


jdhollen (approved):

enterprise/server/backends/distributed/distributed.go
```diff
@@ -751,6 +760,21 @@ func (c *Cache) FindMissing(ctx context.Context, resources []*rspb.ResourceName)
...
+		if _, ok := foundMap[r.GetDigest().GetHash()]; ok {
+			continue
+		}
+		metrics.DistributedCachePeerLookups.With(prometheus.Labels{
+			metrics.DistributedCacheOperation: "FindMissing",
+			metrics.CacheHitMissStatus:        "miss",
+		}).Observe(float64(lookups))
+		count++
+	}
+	log.Infof("Misses: %d", count)
```
vadimberezniker, line 776: Forgot to remove.

vadimberezniker (approved):

enterprise/server/backends/distributed/distributed.go
bduffany: Woops, thanks

PR #6153, opened 2024-03-14
Use executor host ID for task routing

bduffany:
Depends on:
- https://github.com/buildbuddy-io/buildbuddy/pull/6156

When executors restart, all task routing info is effectively lost because we route by `executor_id`, which is set to a new UUID when the executor restarts. This PR makes it so we route by `executor_host_id` instead, which is preserved across restarts.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -479,6 +479,7 @@ func (h *executorHandle) startTaskReservationStreamer() {
 
 type executionNode struct {
 	executorID            string
+	executorHostID        string
 	assignableMemoryBytes int64
 	assignableMilliCpu    int64
 	// Optional host:port of the scheduler to which the executor is connected. Only set for executors connecting using
@@ -507,6 +508,10 @@ func (en *executionNode) String() string {
 
 func (en *executionNode) GetExecutorID() string {
 	return en.executorID
+}
+
+func (en *executionNode) GetExecutorHostID() string {
+	return en.executorHostID
 }
 
 func nodesThatFit(nodes []*executionNode, taskSize *scpb.TaskSize) []*executionNode {
@@ -611,6 +616,7 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode, 
 
 		executors = append(executors, &executionNode{
 			executorID:            id,
+			executorHostID:        node.GetRegistration().GetExecutorHostId(),
 			schedulerHostPort:     node.GetSchedulerHostPort(),
 			assignableMemoryBytes: node.GetRegistration().GetAssignableMemoryBytes(),
 			assignableMilliCpu:    node.GetRegistration().GetAssignableMilliCpu(),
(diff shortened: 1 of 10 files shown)


bduffany:
Some tests are failing, moving to draft while I fix.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -479,6 +479,7 @@ func (h *executorHandle) startTaskReservationStreamer() {
 
 type executionNode struct {
 	executorID            string
+	executorHostID        string
```
maggie-lou, line 482: nit: Can you add a comment explaining the difference with these IDs

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -124,35 +124,41 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
 		return nonePreferred(nodes)
 	}
 
-	nodeByID := map[string]interfaces.ExecutionNode{}
+	nodeByExecutorID := map[string]interfaces.ExecutionNode{}
```
maggie-lou, line 127: Do we need this?

maggie-lou (approved):

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -124,35 +124,41 @@ func (tr *taskRouter) RankNodes(ctx context.Context, cmd *repb.Command, remoteIn
...
+	nodeByHostID := map[string]interfaces.ExecutionNode{}
```
iain-macdonald, line 128: Doesn't this assume there's only one executor per host? I think this should be a multi-map (or map to a slice) and then the logic below should probably select a random executor if there are multiple on a preferred host. Mind adding a test for that case too?

bduffany: I don't think random routing is the right behavior, because we don't support multiple executors per host unless they are using different filecache directories, in which case picking a random executor might route the task to the wrong one (the filecache might not have the task's inputs).

Ideally, I think the behavior would be "route to the executor that last executed the task, or if it doesn't exist, route to the executor that now lives on that host." Implementing this behavior would complicate how we store data in Redis, though. Since we don't really recommend or support running multiple executors per host (except in unit tests, where we fake the host ID to force different filecache dirs), I would prefer to keep this as-is for now.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
bduffany: Done

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Looks like this is still here, maybe forgot to push?

enterprise/server/scheduling/task_router/task_router.go
iain-macdonald: Yeah, I think the behaviour you describe is what we'd like, but I suppose it's not necessary at the moment. Do you think you could add a comment in the code elaborating?

iain-macdonald (approved):

enterprise/server/scheduling/task_router/task_router.go
bduffany: Done

PR #6163, opened 2024-03-15
[RB] Fix duplicate logs when the console does not support ANSI escape sequences

maggie-lou:
We run the ci_runner in interactive mode so that it prints real-time updates for git and bazel commands.
However the Github Actions shell does not support ANSI escape sequences (https://github.com/actions/runner/issues/241). When running remote bazel or the ExecuteWorkflow API in a Github Action runner, this causes output lines to be printed several times (because the real-time printing tries to delete and over-write a previous line with an updated progress message, but deletion isn't supported). 

Add a configurable option to disable interactive mode
 

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3143

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -67,6 +67,7 @@ var (
 	remoteRunner   = remoteFlagset.String("remote_runner", defaultRemoteExecutionURL, "The Buildbuddy grpc target the remote runner should run on.")
 	timeout        = remoteFlagset.Duration("timeout", 0, "If set, requests that have exceeded this timeout will be canceled automatically. (Ex. --timeout=15m; --timeout=2h)")
 	execPropsFlag  = bbflag.New(remoteFlagset, "runner_exec_properties", []string{}, "Exec properties that will apply to the *ci runner execution*. Key-value pairs should be separated by '=' (Ex. --runner_exec_properties=NAME=VALUE). Can be specified more than once. NOTE: If you want to apply an exec property to the bazel command that's run on the runner, just pass at the end of the command (Ex. bb remote build //... --remote_default_exec_properties=OSFamily=linux).")
+	interactive    = remoteFlagset.Bool("interactive", true, "Whether to run the script in interactive mode (will print real-time progress updates using ANSI escape sequences). This may cause duplicate logs to be printed if the console does not support ANSI escape sequences.")
 
 	defaultBranchRefs = []string{"refs/heads/main", "refs/heads/master"}
 )
@@ -615,6 +616,7 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 		ContainerImage: *containerImage,
 		Env:            envVars,
 		ExecProperties: platform.Properties,
+		Interactive:    *interactive,
 	}
 	req.GetRepoState().Patch = append(req.GetRepoState().Patch, repoConfig.Patches...)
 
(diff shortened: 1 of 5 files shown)


maggie-lou:
* **#6163** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6163?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

cli/remotebazel/remotebazel.go
```diff
@@ -67,6 +67,7 @@ var (
 	remoteRunner   = remoteFlagset.String("remote_runner", defaultRemoteExecutionURL, "The Buildbuddy grpc target the remote runner should run on.")
 	timeout        = remoteFlagset.Duration("timeout", 0, "If set, requests that have exceeded this timeout will be canceled automatically. (Ex. --timeout=15m; --timeout=2h)")
 	execPropsFlag  = bbflag.New(remoteFlagset, "runner_exec_properties", []string{}, "Exec properties that will apply to the *ci runner execution*. Key-value pairs should be separated by '=' (Ex. --runner_exec_properties=NAME=VALUE). Can be specified more than once. NOTE: If you want to apply an exec property to the bazel command that's run on the runner, just pass at the end of the command (Ex. bb remote build //... --remote_default_exec_properties=OSFamily=linux).")
+	interactive    = remoteFlagset.Bool("interactive", true, "Whether to run the script in interactive mode (will print real-time progress updates using ANSI escape sequences). This may cause duplicate logs to be printed if the console does not support ANSI escape sequences.")
```
bduffany, line 70: Instead of making this an explicit flag, I think it would be better to check whether the command's output is a terminal or not. You can use `terminal.IsTTY(os.Stderr) && terminal.IsTTY(os.Stdout)` to check (from the `cli/terminal` package). When running locally in a terminal, this should return true. If you send the output to a pipe, like `bb remote ... | cat`, it'll return false. On GH actions, I think it will return false, since the actions runner is not running actions inside an interactive terminal.

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1929,6 +1930,12 @@ func runCommand(ctx context.Context, executable string, args []string, env map[s
 	if dir != "" {
 		cmd.Dir = dir
 	}
+
+	if !*interactive {
```
bduffany, line 1934: Probably should be `!*interactive  && *workflowID == ""` so we continue to use a PTY for workflows, since BB workflows supports ANSI sequences.

bduffany (commented):
Just as a heads up, I don't think this will fully fix the linked issue, because we're still using ANSI escape sequences to redraw the log tail. I think we will need some server-side changes to allow configuring an invocation in "non-streaming" mode, which might be a bit involved

maggie-lou:
> Just as a heads up, I don't think this will fully fix the linked issue, because we're still using ANSI escape sequences to erase and redraw the live log tail. I think we will need some server-side changes to allow configuring an invocation in "non-streaming" mode, which might be a bit more involved. But this PR is needed either way so it's a good starting point 👍

Simplified the PR by only checking for interactive support in cli/remotebazel

cli/remotebazel/remotebazel.go
maggie-lou: Done

cli/remotebazel/remotebazel.go
```diff
@@ -415,6 +418,44 @@ func streamLogs(ctx context.Context, bbClient bbspb.BuildBuddyServiceClient, inv
...
+		})
+		if err != nil {
+			return status.UnknownErrorf("error streaming logs: %s", err)
+		}
+
+		if l.GetLive() {
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		drawChunk(l)
```
bduffany, line 449: I think this could be

```
os.Stdout.Write(chunkGetBuffer())
```

and then we could get rid of drawChunk

bduffany (approved):

PR #6165, opened 2024-03-15
Upgrade protobuf deps

sluongng:
Upgrade the related protobuf dependencies.

Resolve https://pkg.go.dev/vuln/GO-2024-2611

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -2151,8 +2151,8 @@ def install_go_mod_dependencies(workspace_name = "buildbuddy"):
     go_repository(
         name = "com_github_golang_protobuf",
         importpath = "github.com/golang/protobuf",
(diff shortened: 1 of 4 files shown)


deps.bzl
```diff
@@ -6400,9 +6400,10 @@ def install_go_mod_dependencies(workspace_name = "buildbuddy"):
     )
     go_repository(
         name = "org_golang_google_protobuf",
+        build_file_proto_mode = "disable_global",
```
iain-macdonald, line 6403: mind throwing a comment in here explaining what we discussed offthread?

iain-macdonald (approved):

deps.bzl
sluongng: Added a comment clarifying why this was needed and when we could remove it.

PR #6173, opened 2024-03-15
Fix approvals on changes when the user has no draft comments

jdhollen:
For those of us who just smash approve

**Related issues**: N/A

### DIFF
--- a/proto/github.proto
+++ b/proto/github.proto
@@ -635,7 +635,14 @@ message GetGithubPullRequestDetailsResponse {
 
 message SendGithubPullRequestReviewRequest {
   context.RequestContext request_context = 1;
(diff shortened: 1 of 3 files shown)


proto/github.proto
```diff
@@ -635,7 +635,14 @@ message GetGithubPullRequestDetailsResponse {
 
 message SendGithubPullRequestReviewRequest {
   context.RequestContext request_context = 1;
+  // The unique idenitifer of the review.  If this is not set, pull_request_id
```
tylerwilliams, line 638:
```suggestion
  // The unique identified of the review.  If this is not set, pull_request_id
```

tylerwilliams (approved):
LGTM

PR #6177, opened 2024-03-18
Show error for workflows that fail before invocation is created

bduffany:
Today, when a workflow execution fails, the UI just shows "Invocation is waiting for an available worker". With this PR, we show the execution error instead.

image

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3027


enterprise/server/execution_service/execution_service.go
```diff
@@ -32,9 +35,16 @@ func checkPreconditions(req *espb.GetExecutionRequest) error {
 	return status.FailedPreconditionError("An execution lookup with invocation_id must be provided")
 }
 
-func (es *ExecutionService) queryExecutions(ctx context.Context, baseQuery *query_builder.Query) ([]*tables.Execution, error) {
+func (es *ExecutionService) getInvocationExecutions(ctx context.Context, invocationID string) ([]*tables.Execution, error) {
```
bduffany, line 38: The diff here looks a little weird, but it's getting rid of `queryExecutions` and moving that logic into `getInvocationExecutions` since it wasn't called anywhere else, and changing the invocations JOIN to be a LEFT JOIN instead.

app/invocation/invocation.tsx
```diff
@@ -86,7 +97,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
     this.logsSubscription = this.logsModel.onChange.subscribe({
       next: () => this.forceUpdate(),
     });
-    if (!this.isQueued()) {
+    if (!this.isQueued() && !this.props.search.get("runnerFailed")) {
```
siggisim, line 100: Would it make sense to show a retry button if this `runnerFailed` prop is set?

siggisim (approved):

app/invocation/invocation.tsx
bduffany: It will be a bit involved because the current retry button uses BES metadata in order to retry the workflow. We could potentially change the implementation to allow passing in an `Action` to be retried (we would just re-fetch the Github token).

PR #6179, opened 2024-03-18
Fix nil pointer ref in vbd

maggie-lou:
Error: https://buildbuddy-corp.slack.com/files/USLACKBOT/F06QSAARCAU/new_error_in_flame-build_executor_executor-54df5bd945-dkck9

We unmount vbd in the background, which [sets f.server=nil when it's complete](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/vbd/vbd.go#L146). We also may [return early from the Unmount call if the context is cancelled](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/vbd/vbd.go#L137), and we don't wait for the background unmount process to complete. If we don't set the vbd=nil, when we're cleaning up the VM we might try to unmount again, which will cause a nil pointer ref

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -1273,10 +1273,11 @@ func (c *FirecrackerContainer) hotSwapWorkspace(ctx context.Context, execClient 
 			return status.InternalErrorf("failed to stub out workspace drive: %s", err)
 		}
 		if c.workspaceVBD != nil {
-			if err := c.workspaceVBD.Unmount(ctx); err != nil {
+			err := c.workspaceVBD.Unmount(ctx)
+			c.workspaceVBD = nil
+			if err != nil {
 				return status.WrapError(err, "unmount workspace vbd")
 			}
-			c.workspaceVBD = nil
 		}
 		if c.workspaceStore != nil {
 			c.workspaceStore.Close()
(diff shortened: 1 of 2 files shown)


maggie-lou:
* **#6179** <a href="https://app.graphite.dev/github/pr/buildbuddy-io/buildbuddy/6179?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

tylerwilliams (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1273,10 +1273,11 @@ func (c *FirecrackerContainer) hotSwapWorkspace(ctx context.Context, execClient
 			return status.InternalErrorf("failed to stub out workspace drive: %s", err)
 		}
 		if c.workspaceVBD != nil {
-			if err := c.workspaceVBD.Unmount(ctx); err != nil {
+			err := c.workspaceVBD.Unmount(ctx)
```
bduffany, line 1276: nice catch! do you mind doing this for the other unmounts as well?

maggie-lou: unmountAllVBDs doesn't exit early if there's an error, so we should be all good https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/containers/firecracker/firecracker.go#L2399

bduffany: ah cool, thanks for checking :+1:

PR #6191, opened 2024-03-19
cli/fix: support bzlmod

sluongng:
Ensure that bzlmod is correctly detected.

In case bzlmod is used, we need to query for the location of

```
  @@gazelle~override~go_deps~bazel_gazelle_go_repository_config//:WORKSPACE
```

so that we could feed it to Gazelle to generate BUILD files with correct
import.

### DIFF
--- a/cli/bazelisk/bazelisk.go
+++ b/cli/bazelisk/bazelisk.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -14,7 +15,7 @@ import (
 	"github.com/bazelbuild/bazelisk/core"
 	"github.com/bazelbuild/bazelisk/repositories"
 	"github.com/buildbuddy-io/buildbuddy/cli/arg"
-	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	bbLog "github.com/buildbuddy-io/buildbuddy/cli/log"
 	"github.com/buildbuddy-io/buildbuddy/cli/workspace"
 )
 
@@ -71,6 +72,11 @@ func BazelInfo(requestInfos []string) (map[string]string, error) {
 		Stdout: stdout,
 		Stderr: stderr,
 	}
+	// Prevent Bazelisk `log.Printf` call to write directly to stderr
+	oldWriter := log.Writer()
+	log.SetOutput(stdout)
+	defer log.SetOutput(oldWriter)
+
 	_, err := Run(bazelArgs, opts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to run `bazel info`: %w %s", err, stderr.String())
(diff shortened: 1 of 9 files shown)


sluongng:
There is a dependency between `cli/bazelisk -> cli/workspace`

this made it's not possible to use `bazelisk` inside `workspace` to call `bazel info`.

sluongng:
I want to trickle `useModules` down `PathAndBasename()` as well. However, it will lead to a bigger refactoring and potentially another cyclic deps.

I think we could make do with an assumption with `PathAndBasename` for now. If we need to fix it to be super accurate, we could do it after.

cli/bazelisk/bazelisk.go
```diff
@@ -13,7 +15,7 @@ import (
 	"github.com/bazelbuild/bazelisk/core"
 	"github.com/bazelbuild/bazelisk/repositories"
 	"github.com/buildbuddy-io/buildbuddy/cli/arg"
-	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	bbLog "github.com/buildbuddy-io/buildbuddy/cli/log"
```
bduffany, line 18: nit: we want to prioritize using our custom logging package so it would be better IMO to alias the stdlib package instead. e.g.

```
golog "log"
```

cli/bazelisk/bazelisk.go
```diff
@@ -62,6 +64,39 @@ func Run(args []string, opts *RunOpts) (exitCode int, err error) {
...
+	stderr := &bytes.Buffer{}
+	opts := &RunOpts{
+		Stdout: stdout,
+		Stderr: stderr,
+	}
+	// Prevent Bazelisk `log.Printf` call to write directly to stderr
+	oldWriter := log.Writer()
+	log.SetOutput(stderr)
+	defer log.SetOutput(oldWriter)
+
```
bduffany, line 79: it seems OK to me to put this logic in `Run()` since to me the expected behavior of `RunOpts` would be that all stderr/stdout from bazelisk, including go logs, goes to the stdout/stderr writers passed in. wdyt?

cli/bzlmod/bzlmod.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/buildbuddy-io/buildbuddy/cli/bazelisk"
+)
+
+var (
+	once       sync.Once
+	useModules bool
+	infoErr    error
+)
+
+func UseModules() (bool, error) {
```
bduffany, line 18: nit: this can be simplified using the new `sync.OnceValues`

```
var UseModules = sync.OnceValues(func() (bool, error) {
  // return directly instead of setting globals
})
```

bduffany, line 18: nit: would it make sense to call this `Enabled()`? so the callsite is `if bzlmod.Enabled() { ... }`

bduffany (approved):

cli/bazelisk/bazelisk.go
sluongng: done

cli/bazelisk/bazelisk.go
sluongng: hmm I tried fitting it in if RunOpts.stderr is set 🤔

cli/bzlmod/bzlmod.go
sluongng: Good call. Didn't know about this 😅

sluongng: It's a lot nicer with the new name ☺️

bduffany (approved):

PR #6203, opened 2024-03-20
Add a lookaside cache in distributed.go for small CAS values

tylerwilliams:
<<<All credit to @vadimberezniker for originally suggesting this and @jdhollen for instrumenting things to the point where we think it's a good idea>>>

This CL introduces a configurable size LRU in distributed.go to cache response data at read time when it's read from any remote node. If that same data is re-requested, it can be served directly from the read-through cache and skip the work of hitting a remote peer.

Doesn't cache values over 100K, and doesn't hold on to any value longer than 15 minutes (configurable).


enterprise/server/backends/distributed/distributed.go
```diff
@@ -48,13 +52,19 @@ var (
...
 )
 
 const (
 	// Each hinted handoff is a digest (~64 bytes), prefix, and peer
 	// (40 bytes). So keeping around 100000 of these means an extra 10MB
 	// per peer.
 	maxHintedHandoffsPerPeer = 100000
+
+	// Max size of an entry that can be cached in the lookaside cache.
+	maxLookasideEntrySize = 100_000
```
jdhollen, line 67: might be worth adding a flag for this for if we're flushing it way too fast

enterprise/server/backends/distributed/distributed.go
```diff
@@ -484,14 +609,54 @@ func (c *Cache) remoteGetMulti(ctx context.Context, peer string, isolation *dcpb
...
+
+	results, err := c.cacheProxy.RemoteGetMulti(ctx, peer, isolation, stillMissing)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, r := range stillMissing {
+		buf, ok := results[r.GetDigest()]
+		if ok {
+			c.addLookasideEntry(r, buf)
```
jdhollen, line 634: missing a max size check here, i think?

enterprise/server/backends/distributed/distributed.go
```diff
@@ -266,6 +301,96 @@ func (c *Cache) Check(ctx context.Context) error {
...
+		return nil, status.NotFoundError("lookaside cache disabled")
+	}
+	k, err := digest.ResourceNameFromProto(r).DownloadString()
+	if err != nil {
+		c.log.Debugf("Not getting lookaside entry: %s", err)
+		return nil, err
+	}
+	c.lookasideMu.RLock()
+	entry, ok := c.lookaside.Get(k)
+	c.lookasideMu.RUnlock()
```
bduffany, line 358: `Get` does a mutate of the LRU so this might need to be Lock? (maybe worth also adding a concurrency test since the race detector didn't trigger)

enterprise/server/backends/distributed/distributed.go
tylerwilliams: Whoops, fixed.

enterprise/server/backends/distributed/distributed.go
tylerwilliams: Done.

enterprise/server/backends/distributed/distributed.go
tylerwilliams: Fixed.

vadimberezniker (approved):

PR #6213, opened 2024-03-21
split out failed and successful builds in trends chart

jdhollen:
It looks like this:
Screenshot 2024-03-21 at 11 36 58 AM

- add an "other" stat too, but don't render it for now because it's always empty.
- keep clickthrough behavior without filters--i don't think landing on a page with only successful builds when you click the green part of a bar is intuitive, so keeping things as-is seems better.

**Related issues**: N/A

### DIFF
--- a/enterprise/app/trends/trends.tsx
+++ b/enterprise/app/trends/trends.tsx
@@ -309,20 +309,20 @@ export default class TrendsComponent extends React.Component<Props, State> {
                 data={this.state.timeKeys}
                 dataSeries={[
                   {
-                    name: "successful builds",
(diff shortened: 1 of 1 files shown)


jdhollen:
> lgtm!
> 
> Do you think putting the bit on the bottom and adding a 1px white line between the two looks any better?

moved failures to the bottom because it's easier to see failures going up and down that way, which is what most users presumably care about.

getting a stroke for a gap on bars in recharts without messing with axis rendering is seemingly not super straightforward (seems like the mostly suggest it for area charts), so punting on that.

PR #6217, opened 2024-03-22
go_sdk: upgrade to 1.22.1

sluongng:
Change log: https://tip.golang.org/doc/go1.22

Since musl toolchain in --config=static is not compatible with the
latest Go version, disable static workflows for now.

Issue is reported in https://github.com/uber/hermetic_cc_toolchain/issues/171


buildbuddy.yaml
```diff
@@ -10,17 +10,6 @@ actions:
           - "*"
     bazel_commands:
       - test //... --config=linux-workflows --config=race --test_tag_filters=-performance,-webdriver,-docker,-bare
-  - name: Test with static
```
iain-macdonald, line 13: Could you file a bug to re-enable this workflow with a reference to the upstream issues, so that we don't lose track?

iain-macdonald (approved):

buildbuddy.yaml
sluongng: Tracking it here https://github.com/buildbuddy-io/buildbuddy-internal/issues/3245

PR #6218, opened 2024-03-22
nogo: add missing go vet checks

sluongng:
Remove usage of `vet = True` in nogo definition. It's simply a shortcut
attribute to appends some additional analyzers to the nogo binary.

Instead, run `go tool vet help` and copy all listed analyzers in the doc
to our analyzer. This list is more up-to-date than the one in current
rules_go.

Fix 2 new vet issues found with new analyzers added:

- Misuse of unbuffered os.Signal channel as argument to signal.Notify
  The channel should be buffered with minimal size of 1.

- time.Since should not be used in defer statement
  See https://github.com/golang/go/issues/60048 for more information

Also re-enabled `nilness` shows a range of issues in our code base that compares `nil == nil`, `non-nil != nil`. Fixed those issues in a separate commit for easier review.


sluongng:
based on https://github.com/buildbuddy-io/buildbuddy/pull/6217

enterprise/server/scheduling/task_leaser/task_leaser.go
```diff
@@ -183,11 +183,10 @@ func (t *TaskLeaser) Close(ctx context.Context, taskErr error, retry bool) {
 	if taskErr == nil || !retry {
 		req.Finalize = true
 	} else {
+		// taskErr != nil && retry
```
maggie-lou, line 186: leftover comment

server/util/ioutil/ioutil.go
```diff
@@ -71,22 +73,20 @@ func (c *CustomCommitWriteCloser) Commit() error {
 }
 
 func (c *CustomCommitWriteCloser) Close() error {
-	var firstErr error
```
maggie-lou, line 74: Why can't we continue to use this as a local variable?

maggie-lou (approved):
A couple small things but mostly looks good

enterprise/server/scheduling/task_leaser/task_leaser.go
sluongng: I intentionally left it there as the inverted logic was not obvious to me at first.

Can remove.

server/util/ioutil/ioutil.go
sluongng: I think I could have misunderstood the implementation here.

Let me fix it

server/util/clickhouse/clickhouse.go
```diff
@@ -109,9 +109,6 @@ func (r *rawQuery) IterateRaw(fn func(ctx context.Context, row *sql.Rows) error)
 			return err
 		}
 	}
-	if err != nil {
```
vadimberezniker, line 112: i think this is supposed to check rows.Err()

server/util/db/db.go
```diff
@@ -1018,9 +1018,6 @@ func (r *rawQuery) IterateRaw(fn func(ctx context.Context, row *sql.Rows) error)
 			return err
 		}
 	}
-	if err != nil {
```
vadimberezniker, line 1021: i think this is supposed to check rows.Err()

tools/metrics/grafana/grafana.go
```diff
@@ -174,9 +174,13 @@ func exportNormalizedDashboard(d *Dashboard) error {
 		return err
 	}
 	outPath := filepath.Join(dashboardsDir, fileName)
+	outFile, err := os.ReadFile(outPath)
+	if err != nil {
```
bduffany, line 178: nit: the original intent was to write the file if it did not exist or if it contains different bytes

```suggestion
	if err != nil && !os.IsNotExist(err) {
```

bduffany (approved):

tools/metrics/grafana/grafana.go
sluongng: In case the file does not exist, the byte slice content return will be nil and after the `bytes.Equal` result 🤔 

I guess we want to write when the previous file was not there (i.e. on the first time exporting)?

sluongng: I sent a fix that would handle the nil content case.

vadimberezniker (approved):

PR #6223, opened 2024-03-25
buildpatches: replace module references with alias

sluongng:
Because of the visibility issue in bzlmod (1), repos that are
generated from a module extension cannot depend on build targets in
first-level modules directly.

The workaround is to create an alias for these first-level module's
targets in the main repository, then use the special alias `@@` to refer
to the main repository.

We expect this issue to be fixed in Bazel 7.2.

(1): https://github.com/bazelbuild/bazel/issues/19301#issuecomment-1923527294

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -7,6 +7,12 @@ load("//rules/go:index.bzl", "go_sdk_tool")
 
 package(default_visibility = ["//visibility:public"])
 
+alias(
+    name = "zlib",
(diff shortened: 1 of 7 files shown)


sluongng:
closing in favor of creating a dedicated bzlmod version of the patches.

Folded this into https://github.com/buildbuddy-io/buildbuddy/pull/5792

PR #6227, opened 2024-03-25
Don't recycle VM if workspace device is still busy after unmounting

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -2020,7 +2020,7 @@ func (c *FirecrackerContainer) create(ctx context.Context) error {
 	return nil
 }
 
-func (c *FirecrackerContainer) SendExecRequestToGuest(ctx context.Context, conn *grpc.ClientConn, cmd *repb.Command, workDir string, stdio *interfaces.Stdio) *interfaces.CommandResult {
+func (c *FirecrackerContainer) SendExecRequestToGuest(ctx context.Context, conn *grpc.ClientConn, cmd *repb.Command, workDir string, stdio *interfaces.Stdio) (_ *interfaces.CommandResult, healthy bool) {
 	ctx, span := tracing.StartSpan(ctx)
 	defer span.End()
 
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -2257,30 +2250,61 @@ func (c *FirecrackerContainer) Exec(ctx context.Context, cmd *repb.Command, stdi
...
-		// Command was successful, let's unpack the files back to our
-		// workspace directory now.
+		// Unmount the workspace and pause the VM before syncing to ensure that
+		// the guest's FS cache is flushed to the backing file on the host and
+		// that no concurrent operations can occur while we're reading the
+		// workspace disk.
+		//
+		// TODO(bduffany): after notifying users of these warnings below,
+		// make it a hard failure if we fail to unmount or if the workspace disk
+		// is still busy after unmounting.
```
bduffany, line 2275: The logic here is a bit ugly but the idea is to avoid breaking actions which are working today just because this new Unmount logic is failing. Once this is rolled out I would like to check and see how often these logs are happening in practice and then make the unmount failures real errors instead of warnings.

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -2257,30 +2250,61 @@ func (c *FirecrackerContainer) Exec(ctx context.Context, cmd *repb.Command, stdi
...
 			result.Error = status.WrapError(err, "failed to copy action outputs from VM workspace")
 			return result
 		}
-
+		// Resume the VM and remount the workspace now that we're done copying.
 		if err := c.machine.ResumeVM(ctx); err != nil {
 			result.Error = status.InternalErrorf("error resuming VM after copying workspace outputs: %s", err)
 			return result
 		}
+		if unmounted {
```
jdhollen, line 2302: curious if you're worried about this double-mounting or what? feels to me like there are cases where the error in the unmount() call above doesn't _necessarily_ mean that the guest didn't unmount the drive (even though farfetched).

i guess i'm asking if it would be more or less safe in your opinion to drop this `if unmounted` check.

jdhollen: (as an alternative, we could make MountWorkspace a no-op if the workspace is already mounted)

bduffany: hm I think I just forgot to set `DoNotRecycle = true` if the unmount failed (done). That way we don't have to worry about double-mounting because the VM won't be reused.

jdhollen (approved):

tylerwilliams (approved):

PR #6232, opened 2024-03-25
Add executor option to blackhole traffic to private ranges.

vadimberezniker:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/podman.go
+++ b/enterprise/server/remote_execution/containers/podman/podman.go
@@ -836,14 +836,7 @@ func (c *podmanCommandContainer) removeImage(ctx context.Context, imageName stri
 	return status.UnknownErrorf("podman rmi failed: %s", string(result.Stderr))
 }
 
-// Configure the secondary network for podman so that traffic from podman will be routed through
-// the secondary network interface instead of the primary network.
-func ConfigureSecondaryNetwork(ctx context.Context) error {
-	if !networking.IsSecondaryNetworkEnabled() {
-		// No need to configure secondary network for podman.
-		return nil
-	}
-
+func configureSecondaryNetwork(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -874,6 +867,25 @@ func ConfigureSecondaryNetwork(ctx context.Context) error {
 	if err := networking.AddRouteIfNotPresent(ctx, route); err != nil {
 		return err
 	}
+
+	return nil
+}
+
+// Configure the secondary network for podman so that traffic from podman will be routed through
```
luluz66, line 874: nit: ConfigureIsolation configures...also maybe mention the blackholing

tylerwilliams (approved):

server/util/networking/networking.go
```diff
@@ -496,7 +515,32 @@ func routingTableContainsTable(tableEntry string) (bool, error) {
 	return false, nil
 }
 
-// ConfigurePolicyBasedRoutingForNetworkWIthRoutePrefix configures policy routing for secondary
+// ConfigureRoutingForIsolation sets up a routing table for handling network
+// isolation via either a secondary network interface or blackholing.
+func ConfigureRoutingForIsolation(ctx context.Context) error {
+	if !IsSecondaryNetworkEnabled() && !*blackholePrivateRanges {
```
luluz66, line 521: nit: use IsPrivateRangeBlackholingEnabled()

enterprise/server/remote_execution/containers/podman/podman.go
vadimberezniker: done

server/util/networking/networking.go
vadimberezniker: done

PR #6233, opened 2024-03-25
Make more firecracker errors retryable

bduffany:
**Related issues**: N/A


jdhollen (approved):
i can both recognize that you are using UnavailableError properly and also claim that it is the worst-named status in grpc

PR #6240, opened 2024-03-26
Add invocations by Bazel versions graph to Grafana dashboard

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/tools/metrics/grafana/dashboards/buildbuddy.json
+++ b/tools/metrics/grafana/dashboards/buildbuddy.json
@@ -80,7 +80,7 @@
             "h": 7,
             "w": 6,
             "x": 0,
-            "y": 57
+            "y": 65
           },
           "hiddenSeries": false,
           "id": 21,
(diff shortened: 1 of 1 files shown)


tools/metrics/grafana/dashboards/buildbuddy.json
```diff
@@ -680,414 +680,510 @@
...
       "targets": [
         {
           "datasource": {
             "type": "prometheus",
             "uid": "prom"
           },
+          "editorMode": "code",
+          "expr": "sum by (bazel_version) (rate(buildbuddy_invocation_bazel_version{region=\"${region}\"}[${window}]))",
+          "interval": "",
+          "legendFormat": "{{invocation_status}}",
```
vadimberezniker, line 1149: can you double check the legend is showing up correctly? this doesn't seem right

vadimberezniker (approved):

tools/metrics/grafana/dashboards/buildbuddy.json
iain-macdonald: Changing this to `bazel_version` didn't make any visible difference, so I'll just set it to that.

PR #6241, opened 2024-03-26
make tasks running chart only show tasks for the pool being viewed

jdhollen:
it's currently showing all tasks for all executor types.

**Related issues**: N/A


vadimberezniker (approved):
"executing by stage" is also missing the filter

jdhollen:
> "executing by stage" is also missing the filter

Looks like it's there to me--I'll figure out what I'm missing tomorrow :P

PR #6246, opened 2024-03-27
Rework flag tags to allow for easier extensibility

tempoz:
Make tags extensible without needing to special-case each new tag type that needs to modify methods.

Go `flag.Value`s need to be able to produce accurate string forms of their zero values even when the instance is nil because of how they are used in `flagset.Usage`:

https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/flag/flag.go;l=642-644

This means if we are wrapping a value, we need to capture the type of the value in a type parameter of the wrapping type. Unfortunately, that means that if we want the `Tag` function of a tag we pass in to the `New` or `Var` functions would need to have a type parameter as well if we needed them to wrap the flag value themselves, resulting in the awkward and repetitive:

```go
var foo = autoflags.New(flag.CommandLine, "foo", Foo{}, flagutil.SecretTag[Foo]{})
```

where `SecretTag[T any]` would have a `Tag` method that would wrap the flag in question in a `SecretFlag[T]`.

To avoid this, I had been custom-recognizing the tag types that need to wrap flags inside of `Var[T]` and just ignoring their tag methods and doing the wrapping there. However, ideally creating a new tag doesn't require this weird special-casing. To that end, I have created `TagFlag[T]` which can be created by the `Tag[T]` function to wrap flag values and then be passed to the `Tag` method of a tag as a `Tagged`. The method can then freely modify the functions by using `Tagged`'s interface to set functions without needing to parameterize the type of the tag or create `flag.Value`s that cannot satisfy the requirement of producing valid strings when nil.

I did all this when I was adding a new `Hidden` flag tag for buildbuddy-io/buildbuddy-internal#2804 and then realized I was going to have to add yet another special case to the tag for loop in `Var[T]`.

This also prevents duplication of code and boilerplate for every tag-flag-value-type, as can be seen with how it allows `Secret` and `Deprecate` to both be defined using TagFlag. It should, generally, also make it easier to define new tags without touching `flagutil` at all.

**Related issues**: N/A

### DIFF
--- a/server/util/flagutil/flagutil.go
+++ b/server/util/flagutil/flagutil.go
@@ -1,6 +1,8 @@
 package flagutil
 
 import (
+	"flag"
+
 	"github.com/buildbuddy-io/buildbuddy/server/util/flagutil/common"
 )
 
@@ -33,4 +35,10 @@ func ResetFlags() error {
 // a given flag name.
 func GetDereferencedValue[T any](name string) (T, error) {
 	return common.GetDereferencedValue[T](common.DefaultFlagSet, name)
+}
+
+// Expand updates the flag value to replace any placeholders in format ${FOO}
+// with the content of calling the mapper function with the placeholder name.
+func Expand(v flag.Value, mapper func(string) (string, error)) error {
+	return common.Expand(v, mapper)
 }
--- a/server/util/flagutil/types/autoflags/autoflags.go
+++ b/server/util/flagutil/types/autoflags/autoflags.go
@@ -10,23 +10,7 @@ import (
 
 	flagtypes "github.com/buildbuddy-io/buildbuddy/server/util/flagutil/types"
 	flagtags "github.com/buildbuddy-io/buildbuddy/server/util/flagutil/types/autoflags/tags"
-	flagyaml "github.com/buildbuddy-io/buildbuddy/server/util/flagutil/yaml"
 )
-
-type Taggable flagtags.Taggable
-
-var SecretTag = flagtags.SecretTag
-
-var DeprecatedTag = flagtags.DeprecatedTag
-
-type yamlIgnoreTag struct{}
-
-func (_ *yamlIgnoreTag) Tag(flagset *flag.FlagSet, name string, f flagtags.Tagged) flag.Value {
-	flagyaml.IgnoreFlagForYAML(name)
-	return nil
-}
-
-var YAMLIgnoreTag = &yamlIgnoreTag{}
 
 // New declares a new flag named `name` with the specified value `defaultValue`
 // of type `T` and the help text `usage`. It returns a pointer to where the
@@ -34,7 +18,7 @@ var YAMLIgnoreTag = &yamlIgnoreTag{}
 // `SecretTag` to mark a flag that contains a secret that should be redacted in
 // output, or use `DeprecatedTag(migrationPlan)` to mark a flag that has been
 // deprecated and provide its migration plan.
-func New[T any](flagset *flag.FlagSet, name string, defaultValue T, usage string, tags ...Taggable) *T {
+func New[T any](flagset *flag.FlagSet, name string, defaultValue T, usage string, tags ...flagtags.Taggable) *T {
 	value := reflect.New(reflect.TypeOf((*T)(nil)).Elem()).Interface().(*T)
 	Var(flagset, value, name, defaultValue, usage, tags...)
 	return value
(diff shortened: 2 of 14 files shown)


bduffany (commented):
mind updating the PR description with a bit more context / motivation as well as a high-level summary of the changes? There is a lot of code being moved around and I'm not familiar enough with this part of the code to grasp what's happening in this PR after a first pass :)

tempoz:
> mind updating the PR description with a bit more context / motivation as well as a high-level summary of the changes? There is a lot of code being moved around and I'm not familiar enough with this part of the code to grasp what's happening in this PR after a first pass :)

Of, course, no problem!

server/util/flagutil/types/autoflags/tags/tags.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// from YAMLSetValueHooked
+	YAMLSetValueHookFunc(func())
+}
+
+type Taggable interface {
+	Tag(flagset *flag.FlagSet, name string, f Tagged)
+}
+
+type TagFlag[T any] struct {
```
bduffany, line 39: could you add a comment explaining what this is? The name doesn't feel very intuitive to me. Would `TaggedFlag` be more accurate?

server/util/flagutil/types/autoflags/tags/tags.go
```diff
@@ -0,0 +1,199 @@
...
+	ExpandFunc(func(mapping func(string) (string, error)) error)
+
+	// from Secretable
+	IsSecretFunc(func() bool)
+
+	// from SetValueForFlagNameHooked
+	SetValueForFlagNameHookFunc(func())
+
+	// from YAMLSetValueHooked
+	YAMLSetValueHookFunc(func())
```
bduffany, line 32: The methods here seem to be "func variants" of the corresponding interfaces that are mentioned, so I guess that is why we can't embed the interfaces directly here. Could you add a comment explaining more about why the func variants are needed?

server/util/flagutil/types/autoflags/tags/tags.go
```diff
@@ -0,0 +1,199 @@
...
+func (t *TagFlag[T]) Set(value string) error {
+	if t.setFunc != nil {
+		return t.setFunc(value)
+	}
+	return t.Value().Set(value)
+}
+
+func (t *TagFlag[T]) StringFunc(stringFunc func() string) {
+	t.stringFunc = stringFunc
+}
```
bduffany, line 72: maybe prefix the setters with `Set`? I realize `SetSetFunc` is awkward but I think it's better overall for the setters to be named more intuitively. Another option is to replace `*Func` with `-er`, like `ValueSetter`, `Expander`, etc.

server/util/flagutil/flagutil.go
```diff
@@ -34,3 +36,9 @@ func ResetFlags() error {
 func GetDereferencedValue[T any](name string) (T, error) {
 	return common.GetDereferencedValue[T](common.DefaultFlagSet, name)
 }
+
+// Expand updates the flag value to replace any placeholders in format ${FOO}
+// with the content of calling the mapper function with the placeholder name.
+func Expand(v flag.Value, mapper func(string) (string, error)) error {
+	return common.Expand(v, mapper)
+}
```
bduffany, line 44: nit: could we call `common.Expand` directly instead of using this func?

server/util/flagutil/types/autoflags/autoflags.go
```diff
@@ -2,52 +2,26 @@ package autoflags
...
-			"be called for 'deprecatedTag'; tagging a flag as deprecated is instead "+
-			"handled by 'New' or 'Var'.",
-		name,
-	)
-}
-
-func DeprecatedTag(migrationPlan string) Taggable {
-	return &deprecatedTag{migrationPlan: migrationPlan}
-}
+var DeprecatedTag = flagtags.DeprecatedTag
```
bduffany, line 20: to remove one level of indirection, could we reference the `flagtags.*` members in the `flag` package and remove these references?

bduffany (approved):

server/util/flagutil/types/autoflags/tags/tags.go
tempoz: Done!

server/util/flagutil/flagutil.go
tempoz: The indirect function is not required, and I've changed this to an assignment, but `common` is intentionally not accessible outside of `flagutil` and its subpackages, so we do need to basically "export" it here.

PR #6248, opened 2024-03-27
Add warnings to the UI when workflow VM is low on disk/memory

bduffany:
image

**Related issues**: N/A


app/invocation/invocation_suggestion_card.tsx
```diff
@@ -156,6 +160,85 @@ const matchers: SuggestionMatcher[] = [
...
+    let memoryUsage = 0.0;
+    if (memoryTotalBytes !== 0) {
+      memoryUsage = memoryUsedBytes / memoryTotalBytes;
+    }
+
+    const lowResources: string[] = [];
+    const yamlSuggestions: string[] = [];
+    if (rootDiskUsage > 0.98) {
+      lowResources.push(`disk (${formatBytes(rootDiskBytesUsed)} used of ${formatBytes(rootDiskBytesTotal)} total)`);
+      // Conservatively suggest adding 5GB for now.
```
jdhollen, line 196: Eh, I don't know what a "good" default here is, but this nearly triples the default, right?

jdhollen (approved):

app/invocation/invocation_suggestion_card.tsx
bduffany: workflows default to 20GB of disk

bduffany: Oh, sorry in the screenshot I was explicitly requesting 2GB of disk in buildbuddy.yaml, overriding the default, if that's what you're referring to

jdhollen: i thiiiink I just was remembering the 2GB workspace dir slush size and combined that with your screenshot to think i knew something :P

PR #6250, opened 2024-03-27
Publish an error status when buildbuddy.yaml is invalid

bduffany:
Example (the details link just goes to https://buildbuddy.io/docs/workflows-config for now. Later we could do something fancier, like link to a YAML editor in the Code UI, with some type of schema validation)

image

Also fix logging to use `CtxErrorf` when we fail to start the workflow.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3115


siggisim (approved):
Could we use the execution error display for this? Cool with doing that later.

bduffany:
> Could we use the execution error display for this? Cool with doing that later.

In this case we haven't started an execution, so it would be a bit more involved.

siggisim:
> > Could we use the execution error display for this? Cool with doing that later.
> 
> In this case we haven't started an execution, so it would be a bit more involved.

Ah ok, this seems fine for now.

maggie-lou (approved):

PR #6256, opened 2024-03-28
nogo: enable SA1024, SA4010, SA4009, SA4011, SA5001, SA5011

sluongng:
I found that these analyzers produced useful lint errors and not much noise.

Among these, the commit that enabled `SA4011` should be reviewed carefully
as it attempts to fix some potential bugs that were not checked previously.

```
commit f24c418084bc170e812fe40e36c985f52ec7351b
Author: Son Luong Ngoc <sluongng@gmail.com>
Date:   Thu Mar 28 10:23:36 2024 +0100

    nogo: enable SA1024

    This helps us identify `strings.TrimLeft` and `strings.TrimRight` miss
    uses.

    Reference: https://staticcheck.io/docs/checks/#SA1024

commit 387f2668e31a94c2bc8c7d82ae2333879203a27a
Author: Son Luong Ngoc <sluongng@gmail.com>
Date:   Thu Mar 28 10:31:22 2024 +0100

    nogo: enable SA4010

    this result of append is never used, except maybe in other appends

commit 241792be1d7a840977527fc4ef11a519648e4768
Author: Son Luong Ngoc <sluongng@gmail.com>
Date:   Thu Mar 28 10:34:26 2024 +0100

    nogo: enable SA5001

    should check returned error before deferring f.Close()

commit ed3087b5f091f010cced1b7c662284e75696c55c
Author: Son Luong Ngoc <sluongng@gmail.com>
Date:   Thu Mar 28 10:49:55 2024 +0100

    nogo: enable SA4011

    ```
    ineffective break statement. Did you mean to break out of the outer loop?
    ```

    Most of the cases, `break` statement is not needed as it simply breakout
    of the `select` clause.

    However, there are some cases where `break` was misinterpret as breaking
    out of the outer loop, or exit the function. Fixed those cases to use
    either `break {label}` or `return` instead.

commit 64215b7ab1783ce521ec43eed1a240a748704b97
Author: Son Luong Ngoc <sluongng@gmail.com>
Date:   Thu Mar 28 11:47:34 2024 +0100

    nogo: enable SA4009

    In filepath.WalkDir, the WalkFunc has an err argument that we forgot to
    handle.

    From https://pkg.go.dev/path/filepath#WalkFunc

    ```
    The err argument reports an error related to path, signaling that Walk
    will not walk into that directory. The function can decide how to handle
    that error; as described earlier, returning the error will cause Walk to
    stop walking the entire tree.

    Walk calls the function with a non-nil err argument in two cases.

    First, if an os.Lstat on the root directory or any directory or file in
    the tree fails, Walk calls the function with path set to that directory
    or file's path, info set to nil, and err set to the error from os.Lstat.

    Second, if a directory's Readdirnames method fails, Walk calls the
    function with path set to the directory's path, info, set to an
    fs.FileInfo describing the directory, and err set to the error from
    Readdirnames.
    ```

    Since we are not doing any partial walk to make use of filepath.SkipDir
    or filepath.SkipAll, returning the error directly to stop the entire
    walk.

commit e4be01f2e841fc5ad7c68f1f561f94de61358257
Author: Son Luong Ngoc <sluongng@gmail.com>
Date:   Thu Mar 28 12:01:21 2024 +0100

    nogo: enable SA5011

    Possible nil pointer dereference

```

Reference: 
- https://staticcheck.io/docs/checks/#SA1024
- https://staticcheck.io/docs/checks/#SA4009
- https://staticcheck.io/docs/checks/#SA4010
- https://staticcheck.io/docs/checks/#SA4011
- https://staticcheck.io/docs/checks/#SA5001
- https://staticcheck.io/docs/checks/#SA5011


enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -879,11 +879,12 @@ func ConvertFileToCOW(ctx context.Context, env environment.Env, filePath string,
 	}
 
 	var chunksMu sync.Mutex
+out:
 	for chunkStartOffset := int64(0); chunkStartOffset < totalSizeBytes; chunkStartOffset += chunkSizeBytes {
 		select {
 		case <-egCtx.Done():
 			// One goroutine failed - exit the for loop
-			break
+			break out
```
iain-macdonald, line 887: Do you mind either factoring this out to a separate function or adding a control variable and moving the `default` logic out of the `select` statement so we don't need to use the labelled loop here?

tools/cacheload/cacheload.go
```diff
@@ -270,10 +270,11 @@ func main() {
 			writeQPSCounter.Inc()
 
 			if *readQPS > 0 {
+			outer:
 				for i := 0; i < readsPerWrite; i++ {
 					select {
 					case writtenDigests <- d:
-						break
+						break outer
```
iain-macdonald, line 277: Ditto here, do you mind breaking out of the outer loop using a control variable instead of a labelled-for-loop?

enterprise/server/backends/distributed/distributed.go
```diff
@@ -1106,12 +1106,6 @@ func (c *Cache) GetMulti(ctx context.Context, resources []*rspb.ResourceName) (m
 			peerRequests[peer] = append(peerRequests[peer], perHashResources[0])
 		}
 		if len(peerRequests) == 0 {
-			stillMissing := make([]string, 0)
-			for h := range hashResources {
-				if _, ok := gotMap[h]; !ok {
-					stillMissing = append(stillMissing, h)
-				}
-			}
```
tylerwilliams, line 1114: deleting this is kind of funky -- I assume you remove it because stillMissing is not used. Is that a bug?

tylerwilliams:
Sorry to be a pain but could you split this CL up and do one analyzer at a time? There's a lot of critical code changed in this CL and I'm not confident it will get the review it deserves in one big change like this.

sluongng:
Splitting this PR into 

https://github.com/buildbuddy-io/buildbuddy/pull/6264
https://github.com/buildbuddy-io/buildbuddy/pull/6265
https://github.com/buildbuddy-io/buildbuddy/pull/6266
https://github.com/buildbuddy-io/buildbuddy/pull/6267
https://github.com/buildbuddy-io/buildbuddy/pull/6268
https://github.com/buildbuddy-io/buildbuddy/pull/6269

PR #6257, opened 2024-03-28
event_parser: set username on Gitlab CI

sluongng:
For gitlab customers, it would be ideal to use GITLAB_USER_NAME instead
of USER to identify the user.


sluongng:
Context for this https://buildbuddy-corp.slack.com/archives/C065U1KU8KE/p1706542779444159

siggisim (changes requested):
If we did this, we would probably want to do it for Github, BuildKite, etc as well.

You'd also need to update the docs and bunch more references. See this PR as an example: https://github.com/buildbuddy-io/buildbuddy/pull/5988/files

siggisim:
Can you add these to the list in the "Environment variable redacting" section in the docs too?

siggisim (changes requested):
Here: https://github.com/buildbuddy-io/buildbuddy/blob/ae192fa688c26d3a0e6c226995d5888ea458d6d8/docs/guide-metadata.md?plain=1#L241

sluongng:
@siggisim I added them to the redact doc.

I think these variables are exposed to the frontend to display in the "Invocation details" card. Do you want to include that in this PR as well? 🤔

siggisim (approved):
Cool either way on that one.

Lgtm!

PR #6260, opened 2024-03-29
Fix creating repos under non-organizations

siggisim:
We used to implicitly rely on the presence of the `organization` string to determine if a repo should live under an organization or not (because this is how the golang github library we're using did it).

This means we didn't have owner information when creating non-organization repos, causing workflow linking to fail.

This PR fixes that by making the API explicit about whether or not the repo is being created under an organization, and always taking an `owner` field instead of an `organization` field.

### DIFF
--- a/enterprise/app/repo/repo.tsx
+++ b/enterprise/app/repo/repo.tsx
@@ -254,6 +254,7 @@ export default class RepoComponent extends React.Component<RepoComponentProps, R
     if (selectedInstallation) {
       r.owner = selectedInstallation.login;
       r.installationId = selectedInstallation.id;
+      r.installationTargetType = selectedInstallation.targetType;
     }
 
     if (update) {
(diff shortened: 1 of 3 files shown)


enterprise/app/repo/repo.tsx
```diff
@@ -250,15 +250,20 @@ export default class RepoComponent extends React.Component<RepoComponentProps, R
...
+      r.installationId = selectedInstallation.id;
+    }
+
     if (update) {
       r.skipRepo = true;
       r.skipLink = true;
       r.templateVariables = Object.fromEntries(this.state.vars.entries());
     }
 
     if (selectedInstallation?.targetType == "Organization") {
```
tylerwilliams, line 265: Does it make sense to just pass up installation target type directly instead of converting it to a bool?

tylerwilliams (approved):

enterprise/app/repo/repo.tsx
siggisim: Good idea, done!

PR #6263, opened 2024-03-29
ci_main: fix cred helper handle 'store'

sluongng:
In a previous change, we managed to detect when the credential helper
was invoked by `git` with the `get` operation and provide a special
espcape hatch for it.

However, `get` is not the only operation that `git` will call us with.
The other operations include `store` and `erase` in the current
git-credentials spec.

Adjust the credential helper detection when parsing flags to make sure
we don't run into infinite loop with `store` and `erase` operations
(or future new operations).

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -673,12 +673,6 @@ func parseFlags() error {
 
 	unparsedArgs := os.Args[1:]
 	for len(unparsedArgs) > 0 {
-		// Handle special case for credential helper using a subcommand
-		//   ./ci_runner --credential_helper get
-		if len(unparsedArgs) == 1 && unparsedArgs[0] == "get" {
-			break
-		}
-
 		err := flagset.Parse(unparsedArgs)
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -666,11 +666,15 @@ func parseFlags() error {
 	flagset.Init(os.Args[0], flag.ContinueOnError)
 	flagset.Usage = func() {}
 
+	isCredHelper := false
```
bduffany, line 669: I think we could get rid of the special case handling in this func and instead do something like this in the place where we call parseFlags() today

```
if slices.Contains(os.Args, "--credential_helper") {
  flag.Parse()
} else {
  // Use special flag parsing for CI runner to handle flags that aren't defined yet.
  parseFlags()
}
```

sluongng: that's a lot cleaner!

applied.

bduffany: could you also revert the `parseFlags` changes from the previous PR ? Those should no longer be needed

bduffany (approved):

PR #6264, opened 2024-03-29
nogo: enable SA4010

sluongng:
this result of append is never used, except maybe in other appends

### DIFF
--- a/enterprise/server/backends/distributed/distributed.go
+++ b/enterprise/server/backends/distributed/distributed.go
@@ -1118,12 +1118,6 @@ func (c *Cache) GetMulti(ctx context.Context, resources []*rspb.ResourceName) (m
 			peerRequests[peer] = append(peerRequests[peer], perHashResources[0])
 		}
 		if len(peerRequests) == 0 {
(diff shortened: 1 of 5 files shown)


enterprise/server/backends/distributed/distributed.go
```diff
@@ -1118,12 +1118,6 @@ func (c *Cache) GetMulti(ctx context.Context, resources []*rspb.ResourceName) (m
 			peerRequests[peer] = append(peerRequests[peer], perHashResources[0])
 		}
 		if len(peerRequests) == 0 {
-			stillMissing := make([]string, 0)
-			for h := range hashResources {
-				if _, ok := gotMap[h]; !ok {
-					stillMissing = append(stillMissing, h)
-				}
-			}
```
sluongng, line 1126: The last time this was used is in a Debug log.
The log was removed in f7feb5cdb2d6216c8887eab62f0f4317f6515b97 according to 

```
git log -p -L:\ GetMulti\(:enterprise/server/backends/distributed/distributed.go -W
```

vadimberezniker (approved):

PR #6269, opened 2024-03-29
nogo: enable SA1024

sluongng:
This helps us identify `strings.TrimLeft` and `strings.TrimRight` miss
uses.

Reference: https://staticcheck.io/docs/checks/#SA1024


iain-macdonald (approved):
Dang TrimLeft...

PR #6271, opened 2024-03-29
Add linked github repos to code editor empty state

siggisim:
**Related issues**: N/A

### DIFF
--- a/enterprise/app/code/code_empty.tsx
+++ b/enterprise/app/code/code_empty.tsx
@@ -4,13 +4,15 @@ import { invocation } from "../../../proto/invocation_ts_proto";
 import format from "../../../app/format/format";
 
 interface State {
-  repoStats: invocation.InvocationStat[];
+  reposWithStats: string[];
+  linkedRepos: string[];
   loading: boolean;
 }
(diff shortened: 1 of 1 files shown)


enterprise/app/code/code_empty.tsx
```diff
@@ -4,13 +4,15 @@ import { invocation } from "../../../proto/invocation_ts_proto";
 import format from "../../../app/format/format";
 
 interface State {
-  repoStats: invocation.InvocationStat[];
+  reposWithStats: string[];
+  linkedRepos: string[];
   loading: boolean;
 }
 
 export default class CodeEmptyStateComponent extends React.Component {
```
bduffany, line 12:
```suggestion
export default class CodeEmptyStateComponent extends React.Component<{}, State> {
```

(this will probably surface some type errors :stuck_out_tongue: )

bduffany (approved):

enterprise/app/code/code_empty.tsx
siggisim: Good catch, fixed

PR #6272, opened 2024-03-29
Code editor build button tweaks

siggisim:
Some tweaks to the code editor build button based on my experience using it:
1) Fix z-indexes that clash with monaco
2) Move the command state out of the button and into the editor component - this makes it so if you trigger a build through other means (like a sidekick), the command still gets added to your command history so it's easy to run again
3) Adds a button to edit the current command, instead of always having to create a new one
4) Adds the option to set a default config (this makes things like the sidekick more useful when you need to always add a config like --config=remote)

Included a couple of small unrelated changes that I'm happy to break out if you'd like:
1) Add a 8px more padding to the bottom of the build logs card to match the top
2) Fix new file / new folder when not clicking on a node

### DIFF
--- a/enterprise/app/code/code_build_button.tsx
+++ b/enterprise/app/code/code_build_button.tsx
@@ -6,70 +6,55 @@ import Popup, { PopupContainer } from "../../../app/components/popup/popup";
 import Spinner from "../../../app/components/spinner/spinner";
 import { ChevronDown, PlayCircle } from "lucide-react";
 
-const LOCAL_STORAGE_STATE_KEY = "code-button-commands";
-const MAX_COMMANDS = 10;
-
 export interface CodeBuildButtonProps {
   onCommandClicked: (args: string) => void;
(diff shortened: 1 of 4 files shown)


enterprise/app/code/code_build_button.tsx
```diff
@@ -6,92 +6,85 @@ import Popup, { PopupContainer } from "../../../app/components/popup/popup";
...
-  private handleCustomClicked() {
+  private handleCustomClicked(defaultValue: string) {
     this.onCloseMenu();
-    let customArgs = prompt("bazel");
+    let customArgs = prompt("bazel", defaultValue);
     if (!customArgs) {
       return;
     }
+    if (customArgs.startsWith("bazel ")) {
+      customArgs.replace("bazel ", "");
```
bduffany, line 46:
```suggestion
      customArgs = customArgs.replace("bazel ", "");
```

bduffany (approved):

PR #6281, opened 2024-03-30
Fix port race condition in distributed_test.FindMissing

vadimberezniker:
The issue affects only local runs. There's no guarantee that a port allocated by testport.FindFree() is still free by the time we try to listen on it and currently we silently swallow the listen error which leads to unpredictable results.

If you are happy with this approach, I will extend it to the other test cases as well.

Part of addressing https://github.com/buildbuddy-io/buildbuddy-internal/issues/3114

**Related issues**: N/A


tylerwilliams (commented):
Wdyt about making listener required?

PR #6283, opened 2024-04-01
workflows/remote-bazel: fix default branch not found

sluongng:
Listing all local branches to find the default branch could yield unreliable results.
For example, in Github Action, there is no local `refs/heads/{master,main}` written by the `actions/checkout` plugin.
It's also interesting how `refs/remote/origin/HEAD` symbolic ref is not written, so trying to use the symbolic ref might not work.
`refs/remote/origin/master` is written conditionally based on whether `fetch-depth: 0` is configured in the `actions/checkout` plugin or not.

Employ 2 strategies to detect the default branch: `local` and `remote`.
For `local` strategy, we attempt to read `refs/heads/{master,main}`, each, from 2 places: in the `.git/refs` directory and the `.git/packed-refs` file.
For `remote` strategy, run `git ls-remote origin` and see if either `refs/heads/{master,main}` exists.

### DIFF
--- a/.github/workflows/remote-bazel-build.yaml
+++ b/.github/workflows/remote-bazel-build.yaml
@@ -16,15 +16,13 @@ jobs:
         with:
           token: ${{ secrets.BUILDBUDDY_GITHUB_USER_TOKEN }}
 
-      - name: Build and install bb cli
+      - name: Download bb cli
         run: |
-          bazelisk build --remote_header=x-buildbuddy-api-key=${{ secrets.BUILDBUDDY_ORG_API_KEY }} --config=cache cli &&\
-          cp -f $(bazel info bazel-bin|xargs)/cli/cmd/bb/bb_/bb /usr/local/bin/ &&\
-          which bb
+          cli/install.sh
 
       - name: Test
         run: |
-          BB_VERBOSE=1 bb remote --env=GIT_REPO_DEFAULT_BRANCH=master test //... \
+          bb remote --env=GIT_REPO_DEFAULT_BRANCH=master test //... \
             --config=linux-workflows --config=race \
             --remote_executor=grpcs://buildbuddy.buildbuddy.io \
             --test_tag_filters=-performance,-webdriver,-docker,-bare \
--- a/.github/workflows/remote-bazel-build.yaml
+++ b/.github/workflows/remote-bazel-build.yaml
@@ -16,13 +16,15 @@ jobs:
         with:
           token: ${{ secrets.BUILDBUDDY_GITHUB_USER_TOKEN }}
 
-      - name: Download bb cli
+      - name: Build and install bb cli
         run: |
-          cli/install.sh
+          bazelisk build --remote_header=x-buildbuddy-api-key=${{ secrets.BUILDBUDDY_ORG_API_KEY }} --config=cache cli &&\
+          cp -f $(bazel info bazel-bin|xargs)/cli/cmd/bb/bb_/bb /usr/local/bin/ &&\
+          which bb
 
       - name: Test
         run: |
-          bb remote --env=GIT_REPO_DEFAULT_BRANCH=master test //... \
+          BB_VERBOSE=1 bb remote --env=GIT_REPO_DEFAULT_BRANCH=master test //... \
             --config=linux-workflows --config=race \
             --remote_executor=grpcs://buildbuddy.buildbuddy.io \
             --test_tag_filters=-performance,-webdriver,-docker,-bare \
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -3,6 +3,7 @@ package remotebazel
 import (
 	"bytes"
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -15,6 +16,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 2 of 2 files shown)


cli/remotebazel/remotebazel.go
```diff
@@ -167,27 +169,67 @@ func determineRemote(repo *git.Repository) (*git.Remote, error) {
...
+			localErr = errors.Join(localErr, err)
+			continue
+		}
+		return defaultRefName, nil
+	}
+
+	return "", status.NotFoundErrorf("could not determine local default branch: %s", localErr)
+}
+
+func remoteDefaultRef(repo *git.Repository) (string, error) {
```
bduffany, line 202: I think we should do something like https://stackoverflow.com/a/50056710 so that we aren't depending on the branches being named "main" or "master"

Also, it looks like `getBaseBranchAndCommit` is already doing a `git ls-remote origin` call, unconditionally. Maybe we can instead parse the output from `git remote show`, so that we don't have to do 2 remote fetches. (These have relatively high latency, like 400 ms for each call, so we should try to just do one if possible)

cli/remotebazel/remotebazel.go
```diff
@@ -167,27 +169,67 @@ func determineRemote(repo *git.Repository) (*git.Remote, error) {
 }
 
 func determineDefaultBranch(repo *git.Repository) (string, error) {
-	branches, err := repo.Branches()
+	f := sync.OnceValues(func() (string, error) {
```
bduffany, line 172: this `OnceValues` won't work as intended, because it's being redefined on each invocation of `determineDefaultBranch`.

cli/remotebazel/remotebazel.go
sluongng: We can try to read the symbolic ref to determine the default branch as part of the `local` strategy.

```
git symbolic-ref refs/remotes/origin/HEAD
```

However, as I noted in the PR summary, in GitHub Actions `actions/checkout` plugin does not write `refs/remotes/origin/HEAD` symbolic ref. So in such a case, we would still need to fall back onto a remote-based option (with `git ls-remote`).

Note that `git remote show` is a higher level command that does a combination of reading the remotes/origin/HEAD ref and calling `git ls-remote`. `git remote show -n` would skip the `ls-remote` call, but the underlying file-parsing methodology still applies.

.github/workflows/remote-bazel-build.yaml
```diff
@@ -16,13 +16,15 @@ jobs:
         with:
           token: ${{ secrets.BUILDBUDDY_GITHUB_USER_TOKEN }}
 
-      - name: Download bb cli
```
maggie-lou, line 19: Just a reminder to revert this before you merge

cli/remotebazel/remotebazel.go
sluongng: I made some changes to attempt to find the local `refs/remotes/origin/*` refs before finally reaching out to `git ls-remote` as the last resort.

Also added a warning log there just to make sure.

Let me know what you think.

cli/remotebazel/remotebazel.go
sluongng: I removed the OnceValues usage entirely.

I guess it's not "safe" to assume `repo` will not change between func calls? 🤔

.github/workflows/remote-bazel-build.yaml
sluongng: side question: why don't we test the CLI from the current HEAD?

This currently make the "remote bazel" failures in each PR not very actionable 🤔

maggie-lou: Hmm that's a good point. It's slow to build the cli on each run, so maybe we can detect if cli-specific files have changed before building. I can add that separately

cli/remotebazel/remotebazel.go
```diff
@@ -167,27 +169,71 @@ func determineRemote(repo *git.Repository) (*git.Remote, error) {
...
+	for _, defaultRefName := range []plumbing.ReferenceName{plumbing.Master, plumbing.Main} {
+		if _, err := repo.Reference(defaultRefName, false /*resolved*/); err != nil {
+			localErr = errors.Join(localErr, err)
+			continue
+		}
+		return defaultRefName.String(), nil
+	}
+
+	// Check for existence of
+	//   refs/remotes/origin/HEAD
```
maggie-lou, line 200: Should `refs/remotes/origin/HEAD` be included in this list?

cli/remotebazel/remotebazel.go
maggie-lou: It probably is safe to assume `repo` won't change, because it refers to the local repo state. Seems unlikely you'd change the default branch while simultaneously running remote bazel on the same machine

Looks like we call `determineDefaultBranch` twice - you could just save the values to vars and pass them into the functions that use them so we don't have to do the potentially expensive remote calls twice

cli/remotebazel/remotebazel.go
maggie-lou: Because the github checkout action doesn't set the local state, remote bazel will probably end up making the remote call pretty frequently. I think Brandon's point is that now we'll make duplicate calls to `ls-remote` in `getBaseBranchAndCommit` and here `originRemote.Fetch(&git.FetchOptions{}); refs, err := originRemote.List(&git.ListOptions{})` Could we make the call once and pass into each function?

cli/remotebazel/remotebazel.go
sluongng: Yeah. Usually, that's a "symbolic ref" that points to the actual default branch ref of that remote. On my local machine it's something like this

```
> cat .git/refs/remotes/origin/HEAD
ref: refs/remotes/origin/master
```

So the expectation is `repo.Reference("refs/remotes/origin/HEAD", resolved=true)` would yield the `refs/remotes/origin/master` ref after the symbolic ref is resolved. So it would be more accurate than "guessing" for repositories with both `master` and `main` branches available.

I still haven't looked into why this symbolic ref is not created in GitHub Workflows though.

cli/remotebazel/remotebazel.go
sluongng: Let me look into this.
Trickle-down the refs should be fine.

But I feel like we could use a dedicated `//cli/git` package with our own `repo` struct that caches all these results to re-use within an invocation. 🤔 
Something for a future PR.

cli/remotebazel/remotebazel.go
```diff
@@ -167,27 +170,66 @@ func determineRemote(repo *git.Repository) (*git.Remote, error) {
...
+
+	return "", status.NotFoundErrorf("could not determine default branch: %s", errors.Join(localErr, remoteErr))
+}
+
+func localDefaultRef(repo *git.Repository) (string, error) {
+	var localErr error
+	// Check for existence of
+	//   refs/heads/main
+	//   refs/heads/master
+	for _, defaultRefName := range []plumbing.ReferenceName{plumbing.Master, plumbing.Main} {
```
bduffany, line 192: did you want to use `defaultRefNames` here instead of `[]plumbing.ReferenceName{...}`?

bduffany (approved):

cli/remotebazel/remotebazel.go
```diff
@@ -167,27 +170,66 @@ func determineRemote(repo *git.Repository) (*git.Remote, error) {
...
+	if localErr == nil {
+		return branch, nil
 	}
 
-	allBranches := make(map[string]struct{})
-	err = branches.ForEach(func(branch *plumbing.Reference) error {
-		allBranches[string(branch.Name())] = struct{}{}
-		return nil
-	})
+	log.Warnf("Could not determine default git branch locally. Finding the default branch on the 'origin' remote.\n")
```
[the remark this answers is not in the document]
sluongng: agree

PR #6284, opened 2024-04-01
cli/storage: fix login with git-worktree

### LINKED ISSUE #6282: [CLI] `bb login` fails in `git-worktree` linked worktrees.
The `bb login` command fails when using a [git worktree](https://git-scm.com/docs/git-worktree),
because `git worktree` replaces the `.git` directory with a file-based pointer to the worktree root in linked worktrees.

```
$ bazel login
Enter your API key from https://app.buildbuddy.io/settings/cli-login: $SEKRAT
failed to write API key to local .git/config: failed to determine git repo root: "/absolute/path/to/worktree_checkout/.git" is not a directory

$ bazel version
Bazelisk version: v1.19.0
bb 5.0.25
Build label: 5.4.0
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 15 16:14:25 2022 (1671120865)
Build timestamp: 1671120865
Build timestamp as int: 1671120865

$ cat .git
gitdir: /absolute/path/to/main/.git/worktrees/worktree_checkout

$ git rev-parse --git-dir
 /absolute/path/to/main/.git/worktrees/worktree_checkout
```

sluongng:
Instead of detecting for WORKSPACE and check for `.git` directory,
leverage `git` to detect the root of the repo.
This location should be ideal to read/write git config to local config
file.

Closes #6282.

### DIFF
--- a/cli/storage/storage.go
+++ b/cli/storage/storage.go
@@ -7,8 +7,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
-
-	"github.com/buildbuddy-io/buildbuddy/cli/workspace"
+	"sync"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cli/storage/storage.go
```diff
@@ -52,23 +51,13 @@ func CacheDir() (string, error) {
...
+		return "", fmt.Errorf("failed to run git rev-parse: %s", err)
 	}
-	dotgit, err := os.Stat(filepath.Join(ws, ".git"))
-	if err != nil {
-		if os.IsNotExist(err) {
-			return "", fmt.Errorf("not inside a git repository")
-		}
-		return "", fmt.Errorf("failed to determine git repo root: %s", err)
-	}
-	if dotgit.Mode().IsDir() {
```
bduffany, line 67: Will `.git` be a symlink in this case? wondering if we can just check if it's a dir or a symlink.

Where possible I would like to avoid invoking `git` because all of these subprocess invocations add up and lead to more latency on every command invocation compared to bazel.

sluongng:
```
> git worktree add ../bb-wt    Preparing worktree (new branch 'bb-wt')                                                   HEAD is now at 70811560f fix store case                                                   > cd ../bb-wt

> cat .git                                                   gitdir: /home/nb/work/buildbuddy/buildbuddy/.git/worktrees/bb-wt                          
                                                                              > ls -al .git
.rw-rw-r-- 65 nb  2 Apr 15:25 .git                                                        

> readlink .git
exit 1
```

It's a file with content pointing to the worktree dir inside   the actual .git dir. https://git-scm.com/docs/git-worktree#_details

Interestingly, it's possible for each worktree to have a separate git-config value 🤔

bduffany: hm ok, I guess we could parse that file but maybe not worth it for now.

bduffany (approved):

PR #6286, opened 2024-04-02
Add docs for running shell scripts as bazel commands

bduffany:
A frequently asked question is "how do I run bash scripts in workflows?" This PR adds a docs section explaining how to do this using Bazel commands.

**Related issues**: N/A


sluongng (approved):
Made good use of that code-block title. Very nice!

maggie-lou (approved):
Awesome - thanks for adding-!

brentleyjones (approved):

PR #6288, opened 2024-04-02
Confirm that we don't trivially leak scorecards in any kind of finalize

tempoz:
Modified tests to additionally confirm that we don't leak scorecards just because of a disconnect or retry.

**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler_test.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/backends/chunkstore"
+	"github.com/buildbuddy-io/buildbuddy/server/backends/memory_metrics_collector"
 	"github.com/buildbuddy-io/buildbuddy/server/build_event_protocol/build_event_handler"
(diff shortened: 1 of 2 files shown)


server/build_event_protocol/build_event_handler/build_event_handler_test.go
```diff
@@ -509,6 +513,10 @@ func TestHandleEventWithWorkspaceStatusBeforeStarted(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, "abc123", invocation.CommitSha)
 	assert.Equal(t, inspb.InvocationStatus_COMPLETE_INVOCATION_STATUS, invocation.InvocationStatus)
+
+	metrics, err := te.GetMetricsCollector().ReadCounts(ctx, "hit_tracker/"+testInvocationID)
```
bduffany, line 517: I think since we are creating "fake" builds by maually streaming BES events, we won't have any cache requests associated with the build, so this assertion would pass even if we didn't flush correctly (because this list will be empty)

The testenv should have a cache client available as `te.GetPooledByteStreamClient()`, so it might work to create a dedicated test that opens a BES stream, uploads a blob (`cachetools.UploadBlob`) then downloads the blob (with the invocation ID set in RequestMetadata, so that it gets associated with the BES stream), then making assertions on the counts before/after calling finalize

PR #6289, opened 2024-04-02
Run RBE tests on linux-arm64

bduffany:
Adds a GitHub workflow that runs applicable RBE tests on linux-arm64 runners.

**Related issues**: N/A

### DIFF
--- a/.github/workflows/test-executor-linux-arm64.yaml
+++ b/.github/workflows/test-executor-linux-arm64.yaml
@@ -41,7 +41,6 @@ jobs:
           [engine]
           cgroup_manager = "cgroupfs"
           '
-          mkdir -p ~/.local
           BUILD_FLAGS=(
             --config=cache
             --bes_backend=remote.buildbuddy.io
--- a/.github/workflows/test-executor-linux-arm64.yaml
+++ b/.github/workflows/test-executor-linux-arm64.yaml
@@ -1,0 +1,63 @@
+name: "Test linux/arm64 executor"
+
+on:
+  push:
+    branches:
+      - "master"
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04-16cpu-arm64
+    if: "!contains(github.event.head_commit.message, 'ci skip')"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      # GitHub's arm64 runners do not have the same pre-installed software as
+      # the amd64 runners so we have to install a few things here.
+
+      - name: Install bazelisk
+        run: |
+          curl -L --output /tmp/bazelisk https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-linux-arm64
+          chmod +x /tmp/bazelisk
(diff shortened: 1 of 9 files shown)


tylerwilliams (approved):

iain-macdonald (approved):

.github/workflows/test-executor-linux-arm64.yaml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          curl -L --output /tmp/bazelisk https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-linux-arm64
+          chmod +x /tmp/bazelisk
+          sudo mv /tmp/bazelisk /usr/bin/bazelisk
+          sudo ln -s /usr/bin/bazelisk /usr/bin/bazel
+
+      - name: Install apt packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gcc g++ podman
```
sluongng, line 33: Could we run these in an ARM64 container? 🤔 
https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

It could simplify this setup by quite a bit.

.github/workflows/test-executor-linux-arm64.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Run tests
+        run: |
+          # Have podman use cgroupfs to prevent warnings like
+          # "The cgroupv2 manager is set to systemd but there is no systemd user session available"
+          # TODO: figure out why these warnings are happening
+          echo >/tmp/containers.conf '
+          [engine]
+          cgroup_manager = "cgroupfs"
+          '
+          mkdir -p ~/.local
```
sluongng, line 44: dumb question: why is `~/.local` needed to be created manually here?

enterprise/server/test/integration/podman/podman_test.go
```diff
@@ -380,9 +385,13 @@ func TestForceRoot(t *testing.T) {
 }
 
 func TestUser(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		// TODO: build podman ourselves, and remove this
+		t.Skipf("--passwd arg is not yet supported by podman version available on GitHub actions runner")
```
sluongng, line 390: Nit: specify the expected vs actual version number so we can flip this in the future when newer versions become available.

sluongng (approved):
A few nits and questions, but LGTM.

Very excited to see this being added!

.github/workflows/test-executor-linux-arm64.yaml
bduffany: Container jobs aren't supported in the arm runners beta yet - idk if it's mentioned in the public docs but they mentioned that in their email.

.github/workflows/test-executor-linux-arm64.yaml
bduffany: Don't remember why I added this but it works without it - removed.

enterprise/server/test/integration/podman/podman_test.go
bduffany: Done

PR #6292, opened 2024-04-03
Fix min version for podman --transient-store flag

bduffany:
I hit this on my Raspberry Pi - the version installed from apt is 4.3.1 which doesn't have --transient-store. Testing with `podman run --rm --privileged mgoltsche/podman:4.3.1 podman run --transient-store echo hello` confirms this as well.

**Related issues**: N/A


sluongng (approved):
Consistent with https://github.com/containers/podman/commit/4a8d953425d029dce24b77912549e3d2ef380100

LGTM

PR #6298, opened 2024-04-04
xcode: fix TrimLeft to TrimPrefix

sluongng:
The previous usage of TrimLeft yielded an unexpected result, so to
compensate for it, we added a "/" prefix to the result developerDirPath.

Change this to TrimPrefix requires us from removing that forward-slash
compenstation.

Provide a test to demonstrate the issue.


sluongng:
Funny to me how MacOS thinks that `//////////Applications/Xcode.app` is a valid path.

PR #6299, opened 2024-04-04
Update release.py to support arch-specific executor images

bduffany:
This PR adds support for (1) decoupling the docker push from release.py, and (2) tagging the executor docker image with a suffix based on CPU arch. These changes are all behind flags, so it should not affect our release workflow until we are ready to add the arm64 executor image.

With this support, we can refactor the `release.py` job into a job graph like:
- `job1` just bump the version tag
- `job2` `matrix=[arm64, amd64]` update docker images, just the executor image if on arm64
- `job3` create a multi-platform docker manifest for the executor, which references the platform-specific images we pushed earlier. This can just be a few bash commands in a `run` script, like `docker manifest create ...`, `docker manifest annotate ...`, `docker manifest push ...`.

**Related issues**: N/A


release.py
```diff
@@ -153,19 +153,21 @@ def push_image_with_bazel(bazel_target, image_tag):
     ).format(image_tag=image_tag, target=bazel_target)
     run_or_die(command)
 
-def update_docker_images(version_tag, skip_update_latest_tag):
+def update_docker_images(version_tag, skip_update_latest_tag, update_executor_image_only, arch_specific_executor_tag):
```
iain-macdonald, line 156: Change the update_executor_image_only to update_apps or something and flip the negation below.

iain-macdonald (approved):
LGTM, Approval

release.py
bduffany: resolved offline to keep as-is because it's unclear if we added more binaries whether we'd want to built those for arm64 as well or not

bduffany: Actually I misunderstood the original intention - added individual flags for `--update_app_image`, `--update_enterprise_app_image`, and `--update_executor_image`.

iain-macdonald (approved):

PR #6304, opened 2024-04-04
[raft] create a worker to handle update tags tasks and skip update tags when necessary

luluz66:
**Related issues**: N/A
Fix https://github.com/buildbuddy-io/buildbuddy-internal/issues/3266. Right now,
everytime we call `AddRange` we launch a goroutine to update tags. Each updateTags
call will grab `gm.mu`. This causes `gossipManager.AddListener` to be blocked in
`NewWithArgs`, because `AddListener` needs to wait for `gm.mu` to be unlocked. 

In this PR, I created a worker to handle updateTags tasks, and will skip
updateTags if the task createdAt is before the lastExecutedAt. When I tested in
dev, we skipped update tags 813 times and only ran updateTags for 17 times.
`NewWithArgs` finished in 11 seconds.

### DIFF
--- a/enterprise/server/raft/store/store.go
+++ b/enterprise/server/raft/store/store.go
@@ -220,7 +220,7 @@ func NewWithArgs(env environment.Env, rootDir string, nodeHost *dragonboat.NodeH
 	// StartOnDiskReplica can be blocked when the the buffered leaderChangeListener channel is full.
 	s.leaseKeeper.Start()
 
-	go s.updateTagsWorker.Start()
+	s.updateTagsWorker.Start()
 
 	// rejoin configured clusters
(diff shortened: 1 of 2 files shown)


enterprise/server/raft/store/store.go
```diff
@@ -1122,7 +1132,57 @@ func (s *Store) Usage() *rfpb.StoreUsage {
...
+		break
+	default:
+		alert.UnexpectedEvent(
+			"update_tags_channel_buffer_full",
+			"Failed to update tags: update tags task buffer is full")
+	}
+}
+
+func (w *updateTagsWorker) Start() {
+	for task := range w.tasks {
```
tylerwilliams, line 1168: do you need to close w.tasks so this goroutine doesn't leak?

tylerwilliams (approved):

PR #6308, opened 2024-04-05
Add a button to invalidate all workflows for a repo

maggie-lou:
This [PR](https://github.com/buildbuddy-io/buildbuddy/pull/6303) changes the behavior of the "rerun from clean runner" button to only invalidate the relevant workflow snapshot. Previously, the button would invalidate all workflows for that repo.

Add a new button to the workflows page to invalidate all workflows for a repo. We may need this if we make a breaking change that requires a universal snapshot upgrade, like a firecracker version increase, for example.

Also remove usages of ExecuteWorkflowRequest.Clean in preparation of deprecating that field. It's strange behavior that we could invalidate snapshots of unrelated workflows when invoking ExecuteWorkflow. We should use the new targeted API for this case.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -1025,3 +1025,13 @@ func UnpackContainerImage(ctx context.Context, l *FileCacheLoader, instanceName,
 	log.CtxDebugf(ctx, "Converted containerfs to COW in %s", time.Since(start))
 	return cow, nil
 }
+
+func ConvertSnapshotKeyForVMMetadata(k *fcpb.SnapshotKey) *repb.VMMetadata_SnapshotKey {
+	return &repb.VMMetadata_SnapshotKey{
+		InstanceName:      k.InstanceName,
+		PlatformHash:      k.PlatformHash,
+		ConfigurationHash: k.ConfigurationHash,
+		Ref:               k.Ref,
+		SnapshotId:        k.SnapshotId,
+	}
+}
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -1025,13 +1025,3 @@ func UnpackContainerImage(ctx context.Context, l *FileCacheLoader, instanceName,
 	log.CtxDebugf(ctx, "Converted containerfs to COW in %s", time.Since(start))
 	return cow, nil
 }
-
-func ConvertSnapshotKeyForVMMetadata(k *fcpb.SnapshotKey) *repb.VMMetadata_SnapshotKey {
-	return &repb.VMMetadata_SnapshotKey{
-		InstanceName:      k.InstanceName,
-		PlatformHash:      k.PlatformHash,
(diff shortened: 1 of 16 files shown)


go.work
```diff
@@ -0,0 +1,4 @@
+go 1.20
+use ./bazel-gazelle
+use ./bazel-gazelle/tests/bcr
+use ./.
```
bduffany, line 4: accidentally checked in?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1345,6 +1347,25 @@ func (s *BuildBuddyServer) InvalidateSnapshot(ctx context.Context, request *wfpb
 	return nil, status.UnimplementedError("Not implemented")
 }
 
+func (s *BuildBuddyServer) InvalidateAllSnapshotsForRepo(ctx context.Context, req *wfpb.InvalidateAllSnapshotsForRepoRequest) (*wfpb.InvalidateAllSnapshotsForRepoResponse, error) {
+	if wfs := s.env.GetWorkflowService(); wfs != nil {
+		if err := s.checkInvalidateSnapshotPerms(ctx); err != nil {
+			return nil, status.UnauthenticatedError(err.Error())
+		}
+
+		groupID := req.GetRequestContext().GetGroupId()
```
bduffany, line 1356: nit: the preferred way to get the group ID is `u, err := env.GetAuthenticator().GetAuthenticatedUser()` then call `u.GetGroupID()`. The auth layer takes care of determining the group ID based on request context + JWT for web requests, or API key owner for API requests.

(also missed this in the RPC added above).

It'd also be good to avoid passing the raw `groupID` to `InvalidateAllSnapshotForRepo` and get the group ID inside that func instead - generally we want to avoid passing around raw group IDs as much as possible.

maggie-lou: Done

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -88,10 +88,14 @@ func (l *FileCacheLoader) currentSnapshotVersion(ctx context.Context, key *fcpb.
...
-		// in this case.
-		return "", nil
+		// Version metadata might not exist in the cache if:
+		// * The snapshot version has never been set (Ex. if you've never invalidated
+		//   a snapshot with this key)
+		// * The version metadata has expired from the cache
+		// In the latter case, we want to be careful to not fallback to an older,
+		// invalid snapshot. So here we generate a new version ID to guarantee
+		// we start from a clean snapshot.
+		return l.env.GetSnapshotService().InvalidateSnapshot(ctx, key)
```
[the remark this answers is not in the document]
maggie-lou: InvalidateSnapshot saves the version metadata in the cache. We want to make sure it's set, or every VM will generate a new version ID.

Ex. Let's say the version metadata expires from the cache.
T0: A VM starts up. There's no version metadata in the cache, so it generates a new version. It completes and saves a snapshot (saving a snapshot does not update the version metadata)
T1: Another VM starts up.  If we hadn't saved the version metadata at T0, this will also generate a new version

bduffany (approved):

PR #6315, opened 2024-04-08
Disable Firecracker and soci-snapshotter on arm64 for now

bduffany:
We'll need arm64 builds of soci-snapshotter as well as vmlinux and firecracker/jailer before we can enable these. Would also be good to have CI tests for these to avoid breakage. Since this seems like significant work, disable these for now and avoid building / shipping them as part of the go binary / docker image.

**Related issues**: N/A


iain-macdonald (approved):
LGTM, Approval

PR #6316, opened 2024-04-08
cli/printlog: handle compact exec log

sluongng:
Add spawn.proto from Bazel.git, which includes ExecLogEntry that is used
to write the new `--experimental_execution_log_compact_file`.

Also add a new flag to `bb print` that would parse the compact log file,
reconstruct the SpawnExec and pretty print the SpawnExec to stdout.
The parsing was recreated from the Java implementation in (1).

Add a bazelrc flag to enable this flag by default so that the file is
always sent to our BES sever for future analysis.

(1): https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/exec/SpawnLogReconstructor.java;l=161;drc=3f9d80c35d88a280c2d53682f3d201b4733a3fff


.bazelrc
```diff
@@ -266,6 +266,8 @@ common --experimental_profile_include_target_label
 common --experimental_profile_include_primary_output
 # Don't merge timing profile actions.
 common --noslim_profile
+# Include compact execution log
+common --experimental_execution_log_compact_file=compact_exec.log
```
sluongng, line 270: It would be nice if we could have a flag to always write this file to the output base by default and upload it to BES after (similar to how timing profile works)

cli/printlog/printlog.go
```diff
@@ -46,10 +50,244 @@ func HandlePrint(args []string) (int, error) {
...
+		files:    make(map[int32]*spb.File),
+		dirs:     make(map[int32]*reconstructedDir),
+		symlinks: make(map[int32]*spb.File),
+		sets:     make(map[int32]*spb.ExecLogEntry_InputSet),
+	}
+}
+
+func (slr *SpawnLogReconstructor) GetSpawnExec() (*spb.SpawnExec, error) {
+	entry := &spb.ExecLogEntry{}
+	br := bufio.NewReader(slr.input)
```
bduffany, line 122: since this buffers `slr.input` I think we might need to store `br` as a struct member and reuse the same reader for each function call

cli/printlog/printlog.go
```diff
@@ -46,10 +50,244 @@ func HandlePrint(args []string) (int, error) {
...
+		case *spb.ExecLogEntry_Directory_:
+			slr.dirs[entry.GetId()] = reconDir(e.Directory)
+		case *spb.ExecLogEntry_UnresolvedSymlink_:
+			slr.symlinks[entry.GetId()] = reconSymlink(e.UnresolvedSymlink)
+		case *spb.ExecLogEntry_InputSet_:
+			slr.sets[entry.GetId()] = e.InputSet
+		case *spb.ExecLogEntry_Spawn_:
+			return slr.reconSpawn(e.Spawn)
+		default:
+			return nil, fmt.Errorf("unknown entry: %s", entry)
```
bduffany, line 146: for forwards-compatibility should we just log a warning here?

cli/printlog/printlog.go
```diff
@@ -46,10 +50,244 @@ func HandlePrint(args []string) (int, error) {
...
+		}
+	}
+}
+
+func (slr *SpawnLogReconstructor) reconSpawn(s *spb.ExecLogEntry_Spawn) (*spb.SpawnExec, error) {
```
bduffany, line 151: nit: `s/recon/reconstruct/g`

.bazelrc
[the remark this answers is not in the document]
sluongng: I will make it write for `/tmp` for now and add a TODO for the future.

cli/printlog/printlog.go
sluongng: ah I missed that. Good call!

cli/printlog/printlog.go
sluongng: done

.bazelrc
sluongng: added as TODO

cli/printlog/printlog.go
sluongng: Oh good call. I did not think of that.

Switched to `log.Warnf`

cli/printlog/printlog.go
```diff
@@ -46,10 +50,236 @@ func HandlePrint(args []string) (int, error) {
...
+	return &reconstructedDir{
+		path:  d.GetPath(),
+		files: filesInDir,
+	}
+}
+
+func reconstructFile(parentDir *spb.ExecLogEntry_Directory, file *spb.ExecLogEntry_File) *spb.File {
+	f := &spb.File{Digest: file.GetDigest()}
+	if parentDir != nil {
+		f.Path = filepath.Join(parentDir.GetPath(), file.GetPath())
```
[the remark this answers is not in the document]
sluongng: good call. Will fix

proto/spawn.proto
```diff
@@ -0,0 +1,352 @@
+// Copyright 2018 The Bazel Authors. All rights reserved.
```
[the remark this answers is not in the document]
sluongng: Right now the way we are "vendoring" these proto files is by copying them as-is with as little modification as possible. So I am hesitating to add additional comments to them right now.

If needed, let's do it in a separate PR for all other proto files for consistency.

PR #6320, opened 2024-04-08
Add a bit more logging to release.py

bduffany:
**Related issues**: N/A

### DIFF
--- a/release.py
+++ b/release.py
@@ -126,6 +126,8 @@ def push_image_for_project(project, version_tag, bazel_target, skip_update_lates
     version_image = get_image(project, version_tag)
     if version_image is None:
         push_image_with_bazel(bazel_target, version_tag)
+    else:
+        print(f'Image gcr.io/{project}:{version_tag} already exists, skipping bazel build.')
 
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -126,6 +126,8 @@ def push_image_for_project(project, version_tag, bazel_target, skip_update_lates
     version_image = get_image(project, version_tag)
     if version_image is None:
         push_image_with_bazel(bazel_target, version_tag)
+    else:
+        print(f'Image gcr.io/{project}:{version_tag} already exists, skipping bazel build.')
```
bduffany, line 130: @maggie-lou is this behavior intentional? I noticed when running release.py for a version I had already pushed, that it was not rebuilt, even when running the script with `--force`

maggie-lou: Originally I think I had added this to make the script more idempotent (so we could rerun the auto release workflows without them changing anything). The thinking was that if you wanted to push a different image, you should bump the version number. But I could be convinced to change the behavior. It's a little confusing what the script will or won't do in various cases

maggie-lou (approved):

PR #6325, opened 2024-04-08
Don't use `sqlite3://:memory:`

bduffany:
`sqlite3://:memory:` has some gotchas which I think were causing the issues with flaky webdriver tests in which tables were suddenly disappearing in the middle of the test.

I randomly stumbled across this HN comment which mentioned the same symptom we were seeing (tables randomly disappearing): https://news.ycombinator.com/item?id=39956165

Also see the linked FAQ here: https://github.com/mattn/go-sqlite3?tab=readme-ov-file#faq

Specifically

> Why I'm getting no such table error?

> Why is it racy if I use a sql.Open("sqlite3", ":memory:") database?

> Each connection to ":memory:" opens a brand new in-memory sql database, so if the stdlib's sql engine happens to open another connection and you've only specified ":memory:", that connection will see a brand new database.

**Related issues**: N/A

### DIFF
--- a/server/testutil/app/app.go
+++ b/server/testutil/app/app.go
@@ -54,7 +54,7 @@ func RunWithApp(t *testing.T, app *App, commandPath string, commandArgs []string
 		"--static_directory=static",
 		"--app_directory=/app",
 		fmt.Sprintf("--app.build_buddy_url=http://localhost:%d", app.HttpPort),
-		"--database.data_source=sqlite3://:memory:",
+		fmt.Sprintf("--database.data_source=sqlite3://%s", filepath.Join(dataDir, "buildbuddy.db")),
 		fmt.Sprintf("--storage.disk.root_directory=%s", filepath.Join(dataDir, "storage")),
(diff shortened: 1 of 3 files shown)


server/testutil/app/app.go
```diff
@@ -54,7 +54,7 @@ func RunWithApp(t *testing.T, app *App, commandPath string, commandArgs []string
 		"--static_directory=static",
 		"--app_directory=/app",
 		fmt.Sprintf("--app.build_buddy_url=http://localhost:%d", app.HttpPort),
-		"--database.data_source=sqlite3://:memory:",
```
tylerwilliams, line 57: Any reason you don't want to use the workaround suggested in the linked thread?

bduffany: The workaround involves 3 different DB flags, which seemed less clean. I don't feel particularly strongly one way or the other but in my experience the on-disk DB is very fast so I figured this would be OK.

tylerwilliams: sgtm

tylerwilliams (approved):
(and good find!)

vadimberezniker (approved):

PR #6332, opened 2024-04-09
Add skopeo back to executor image

bduffany:
I think I accidentally removed this while resolving a merge conflict. This is causing the baremetal tests to fail in dev, because it is unable to convert OCI images to EXT4 which requires skopeo.

**Related issues**: N/A


sluongng (approved):
LGTM

PR #6333, opened 2024-04-09
Add a button to the action details page to invalidate a recycled runner snapshot

maggie-lou:
Screenshot 2024-04-09 at 2 37 11 PM
Screenshot 2024-04-09 at 2 37 18 PM
Screenshot 2024-04-09 at 2 37 24 PM

**Related issues**: N/A


jdhollen:
first of all, hell yeah

second, what do you think of just making this a button right next to the "saved from" and/or "resumed from" headers in the execution metadata? it's kind of far away from the thing that it affects right now, and we're not short on whitespace in the metadata section :P

app/invocation/invocation_action_card.tsx
```diff
@@ -553,7 +596,57 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+                          isOpen={this.state.showInvalidateSnapshotModal}
+                          onRequestClose={() =>
+                            this.setState({ showInvalidateSnapshotModal: false, isMenuOpen: false })
+                          }>
+                          <Dialog>
+                            <DialogHeader>
+                              <DialogTitle>Confirm invalidate VM snapshot</DialogTitle>
+                            </DialogHeader>
+                            <DialogBody>
+                              <p>Are you sure you want to invalidate the recycled VM snapshot used for this action?</p>
```
bduffany, line 626:
```suggestion
                              <p>Are you sure you want to invalidate the VM snapshot used for this action?</p>
```

app/invocation/invocation_action_card.tsx
```diff
@@ -553,7 +596,57 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+                              <p>
+                                That will prevent this recycled runner from being reused. This will make future actions
+                                slower, so this flag is not encouraged.
+                              </p>
```
bduffany, line 630: I don't think we need this warning just yet since these are normal actions rather than VM actions, but if you think it's worth keeping I think it could be clarified a little

```suggestion
                              <p>
                                A new VM will be booted for the next run of this action, which may result in longer execution time.
                              </p>
```

bduffany (approved):
This LGTM as a first iteration - I agree w/ Jim's comment about possibly putting it closer to the snapshot info. Maybe something like this hacked up HTML could work?

image


or moving the 3-dot to the same row also seems good.

PR #6335, opened 2024-04-09
Lower ttl for cache stats

tempoz:
We only allow retries of invocations for four hours, so 24 hours is an excessive TTL for cache stats. Lower it to four.

**Related issues**: N/A


bduffany (changes requested):
These counts are buffered while invocations are in progress and it's possible that an invocation could be running for several hours.

How much are the counts contributing to memory usage vs. the scorecard protos etc? If it's not that much then I'd be inclined to leave this expiration as-is

PR #6337, opened 2024-04-09
Use podman-static version at HEAD in integration tests

bduffany:
- Add a rule that repackages podman-static into a format where we can consume it easily from both integration tests and in the executor image.
- Update integration tests to install podman-static from the archive instead of using a pre-installed version in the container image

**Related issues**: N/A


deps.bzl
```diff
@@ -6935,3 +6936,15 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
         sha256 = "85b1c2591274422234955e906aee39e6f793a88a74f3efc49a1852a0646ce08f",
         urls = ["https://storage.googleapis.com/buildbuddy-tools/binaries/clang-format/clang-format-14_linux-x86_64"],
     )
+
+    http_file(
+        name = "com_github_buildbuddy_io_podman_static_podman-linux-amd64",
+        urls = ["https://github.com/buildbuddy-io/podman-static/releases/download/buildbuddy-{podman_version}/podman-linux-amd64.tar.gz".format(podman_version = PODMAN_VERSION)],
+        sha256 = "ac14152d2ef5cb25abbd615893cd56355f951f075a7ae663f0b2c1bc5d3fb77e",
```
iain-macdonald, line 6943: Won't the hash here and below change with the version? And if so, should it be defined in the same place as `PODMAN_VERSION` to reduce confusion?

enterprise/server/remote_execution/containers/podman/BUILD
```diff
@@ -53,3 +53,39 @@ go_test(
...
+        PODMAN_ARCHIVE=$${SRCS[0]}
+        REGISTRIES_CONF=$${SRCS[1]}
+        tar --extract --file="$$PODMAN_ARCHIVE" --strip-components=1 --directory="$$TMPDIR"
+
+        # Remove README.md (not needed)
+        rm "$$TMPDIR/README.md"
+
+        # Replace /etc/containers/registries.conf with our custom config.
+        rm "$$TMPDIR/etc/containers/registries.conf"
+        cp "$$REGISTRIES_CONF" "$$TMPDIR/etc/containers/registries.conf"
```
sluongng, line 87: Simplify

```suggestion
        # Replace /etc/containers/registries.conf with our custom config.
        cp -f "$$REGISTRIES_CONF" "$$TMPDIR/etc/containers/registries.conf"
```

enterprise/server/remote_execution/containers/podman/BUILD
```diff
@@ -53,3 +53,39 @@ go_test(
...
+    cmd_bash = """
+        TMPDIR=$$(mktemp -d)
+        trap 'rm -rf "$$TMPDIR"' EXIT
+        SRCS=($(SRCS))
+        PODMAN_ARCHIVE=$${SRCS[0]}
+        REGISTRIES_CONF=$${SRCS[1]}
+        tar --extract --file="$$PODMAN_ARCHIVE" --strip-components=1 --directory="$$TMPDIR"
+
+        # Remove README.md (not needed)
+        rm "$$TMPDIR/README.md"
```
sluongng, line 83: Nit:

```suggestion
        # Remove README.md (not needed)
        rm -f "$$TMPDIR/README.md"
```

enterprise/server/remote_execution/containers/podman/BUILD
```diff
@@ -53,3 +53,39 @@ go_test(
...
 )
+
+# Repackages the podman-static distribution to make it more easily consumable
+# both by the container_layer rule and by tests:
+#
+# - Replaces the registry config with our custom one. This way, we don't have to
+#   override it separately.
+# - Strips the top-level podman-static directory. This is needed because the
+#   'tars' attribute of container_layer does not seem to support stripping
+#   components, currently.
```
sluongng, line 64: Nit: since we are unpacking this later to `/` root dir of the Executor container, it would be nice to have a comment here stating what's the expected file layout inside the output tarball.

sluongng (approved):
Only small nits from me.

Good improvement overall. 🙇

enterprise/server/remote_execution/containers/podman/BUILD
bduffany: I prefer the current implementation because it fails if /etc/containers/registries.conf does not exist in the original archive. This is something we'd want to know about because it hints that the registries configuration has possibly moved to a different location.

enterprise/server/remote_execution/containers/podman/BUILD
bduffany: Same deal here, if the README no longer exists then we would want this to fail so that we could clean this up.

enterprise/server/remote_execution/containers/podman/BUILD
bduffany: Done

platforms/podman/version.bzl
```diff
@@ -0,0 +1,13 @@
+"""The podman version used in tests and distributed with the buildbuddy executor image.
```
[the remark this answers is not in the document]
bduffany: Fair enough, done

deps.bzl
```diff
@@ -6935,3 +6943,15 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
...
+        name = "com_github_buildbuddy_io_podman_static_podman-linux-amd64",
+        urls = ["https://github.com/buildbuddy-io/podman-static/releases/download/buildbuddy-{podman_version}/podman-linux-amd64.tar.gz".format(podman_version = PODMAN_VERSION)],
+        sha256 = PODMAN_STATIC_SHA256_AMD64,
+    )
+
+    http_file(
+        name = "com_github_buildbuddy_io_podman_static_podman-linux-arm64",
+        urls = ["https://github.com/buildbuddy-io/podman-static/releases/download/buildbuddy-{podman_version}/podman-linux-arm64.tar.gz".format(podman_version = PODMAN_VERSION)],
+        sha256 = PODMAN_STATIC_SHA256_ARM64,
+    )
```
sluongng, line 6957: Question: I wonder if it would be better if we use `http_archive` with a simple inline `build_file_content` here?

That would let us unpack the archive at the repository phase and use only the static Go binary during the build phase. In such a case, we would not have to worry about the README or `registries.conf` in the existing archive at all.

sluongng: I guess we still do care whether the registries.conf exists in the right path eh? 🤔

sluongng (approved):

deps.bzl
bduffany: yeah registries.conf comes from our repro so IDK if build_file_content would work

iain-macdonald (approved):

PR #6341, opened 2024-04-09
Move VMMetadata proto to a more fitting home

maggie-lou:
The VMMetadata proto feels a little too firecracker-specific to live in remote_execution.proto. I've hit some problems with import cycles (firecracker.proto needs to import remote_execution.proto for generic RBE concepts like `files`, but now there is an implicit import loop because remote_execution.proto contains the definition for this firecracker-specific field).

This PR sets VMMetadata in the more generic `repeated google.protobuf.Any auxiliary_metadata` field of `ExecutedActionMetadata`. That allows us to move the proto definition into firecracker.proto

**Related issues**: Based on feedback/challenges from https://github.com/buildbuddy-io/buildbuddy/pull/6303


app/invocation/invocation_action_card.tsx
```diff
@@ -393,8 +394,29 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
 
+  // For firecracker actions, VM metadata is stored in the auxiliary metadata field
+  // of the execution metadata. Try to decode it into an object if it exists.
+  private getFirecrackerVMMetadata(): firecracker.VMMetadata | null | undefined {
+    const auxiliaryMetadata = this.state.actionResult?.executionMetadata?.auxiliaryMetadata;
+    if (!auxiliaryMetadata) {
+      return this.getDeprecatedFirecrackerVMMetadata();
+    }
+    for (const metadata of auxiliaryMetadata) {
+      if (metadata.typeUrl.includes("firecracker.VMMetadata")) {
```
bduffany, line 405: nit: can we do an equality check on the `typeUrl` instead of `.includes()` ? or does the URL include some contents we aren't expecting?

bduffany (approved):
nice, thanks for making this change!

app/invocation/invocation_action_card.tsx
maggie-lou: Changed. I thought the includes might be slightly less brittle, but don't feel strongly

PR #6342, opened 2024-04-09
split out data model for code review into immutable objects

jdhollen:
just a little Light Refactoring™

Review, Thread, and Comment are now immutable objects, modulo a few things that are less urgent to do this for--i've left TODOs for those fields.

This will make it much easier to build a coherent model where the user can move between different pages (viewing a single file, replying, viewing all files, ...) without worrying too much about maintaining state properly.  Shoulda done this sooner.

**Related issues**: N/A


siggisim (approved):
LGTM, Approval

PR #6345, opened 2024-04-10
Have ExtendContextForFinalization respect the parent context's cancelation

bduffany:
If the parent context has the default action timeout (7 days) and gets canceled (either before or after calling ExtendContextForFinalization), then the context returned by ExtendContextForFinalization will have a deadline that is 7 days + 15s in the future, rather than 15s starting from the time of cancellation like we would expect.

This PR changes the implementation to match the expected behavior in this case, which is to apply a deadline that is `<whenever the parent context gets canceled, either via deadline or manual cancel>` + 15s.

**Related issues**: N/A

### DIFF
--- a/server/util/background/background.go
+++ b/server/util/background/background.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
-	"github.com/buildbuddy-io/buildbuddy/server/util/timeutil"
 )
 
 type disconnectedContext struct {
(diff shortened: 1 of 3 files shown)


server/util/background/background.go
```diff
@@ -35,14 +72,23 @@ func ToBackground(ctx context.Context) context.Context {
...
+		// period, cancel the extended context.
+		select {
+		case <-ctx.Done():
+			return // cancelled manually
+		case <-parent.Done():
 		}
-	}
-	return context.WithTimeout(ctx, timeout)
+		t := time.NewTimer(timeout)
+		defer timeutil.StopAndDrainTimer(t)
```
vadimberezniker, line 85: I don't think you need this. Draining the timer only matters if you plan on reusing the timer.

vadimberezniker (approved):

server/util/background/background.go
bduffany: ah right, changed to just `t.Stop()`

PR #6347, opened 2024-04-10
Clear replica states after we apply remote snapshot to recover

luluz66:
**Related issues**: N/A

Also added a test to verify.

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -723,7 +723,7 @@ func (sm *Replica) clearInMemoryReplicaState() {
 }
 
 // clearInMemoryReplicaState clears in-memory and on-disk replica state.
-func (sm *Replica) clearReplicaState() error {
+func (sm *Replica) clearReplicaState(db ReplicaWriter) error {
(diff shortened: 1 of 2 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -705,6 +705,51 @@ func (sm *Replica) loadRangeLease(db ReplicaReader) {
 	sm.setRangeLease(constants.LocalRangeLeaseKey, buf)
 }
 
+// clearInMemoryReplicaState clears in-memory replica state.
+func (sm *Replica) clearInMemoryReplicaState() {
+	sm.rangeMu.Lock()
+	sm.rangeDescriptor = nil
```
tylerwilliams, line 711: if a replica is opened and then raft calls RestoreFromSnapshot -- will clearReplicaState call store.Remove() before the replica is rewritten from the snapshot?

luluz66: I added store.RemoveRange() in ClearReplicaState if sm.rangeDescriptor is not nil

tylerwilliams (approved):

PR #6350, opened 2024-04-11
cli/printlog: improve compact exec log

sluongng:
Fixed an issue in `reconstructDir()` where the filesInDir slice was
initiated with nil entries.

Rename the compact log file to have the right extension.

### DIFF
--- a/cli/printlog/printlog.go
+++ b/cli/printlog/printlog.go
@@ -91,6 +91,8 @@ func printCompactExecLog(path string) error {
 	}
 }
 
+// SpawnLogReconstructor reconstructs "compact execution log" format back to the original format.
+// As of Bazel 7.1, this is the recommended way to consume the new compact format.
 type SpawnLogReconstructor struct {
 	input *bufio.Reader
 
@@ -203,6 +205,7 @@ func (slr *SpawnLogReconstructor) reconstructSpawn(s *spb.ExecLogEntry_Spawn) (*
 	}
 	se.ListedOutputs = listedOutputs
 	se.ActualOutputs = actualOutputs
+
 	return se, nil
 }
 
@@ -253,7 +256,7 @@ func (slr *SpawnLogReconstructor) reconstructInputs(setID int32) map[string]*spb
 }
 
 func reconstructDir(d *spb.ExecLogEntry_Directory) *reconstructedDir {
-	filesInDir := make([]*spb.File, len(d.GetFiles()))
+	filesInDir := make([]*spb.File, 0, len(d.GetFiles()))
 	for _, file := range d.GetFiles() {
 		filesInDir = append(filesInDir, reconstructFile(d, file))
 	}
--- a/.bazelrc
+++ b/.bazelrc
@@ -268,8 +268,8 @@ common --experimental_profile_include_primary_output
 common --noslim_profile
 # Include compact execution log
 # TODO(sluongng): make Bazel writes this to output_base automatically
-common:linux --experimental_execution_log_compact_file=/tmp/bazel_compact_exec_log.bin
-common:macos --experimental_execution_log_compact_file=/tmp/bazel_compact_exec_log.bin
+common:linux --experimental_execution_log_compact_file=/tmp/bazel_compact_exec_log.binpb.zstd
+common:macos --experimental_execution_log_compact_file=/tmp/bazel_compact_exec_log.binpb.zstd
 
 # Try importing a user specific .bazelrc
 # You can create your own by copying and editing the template-user.bazelrc template:


cli/printlog/compact/compact.go
```diff
@@ -0,0 +1,267 @@
...
+		Cacheable:       s.GetCacheable(),
+		RemoteCacheable: s.GetRemoteCacheable(),
+		TimeoutMillis:   s.GetTimeoutMillis(),
+		Metrics:         s.GetMetrics(),
+		Digest:          s.GetDigest(),
+	}
+	// Handle env vars
+	envVar := s.GetEnvVars()
+	if slr.sort {
+		slices.SortFunc(envVar, compMapEntries)
```
fmeum, line 99: The environment variables are already sorted: https://github.com/bazelbuild/bazel/blob/62f302b4f1205bf1769bb7fbb5eff1e62f51dca2/src/main/java/com/google/devtools/build/lib/exec/SpawnLogContext.java#L85

cli/printlog/compact/compact.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	se.EnvironmentVariables = envVar
+
+	// Handle platform
+	platform := s.GetPlatform()
+	if slr.sort && platform != nil {
+		slices.SortFunc(platform.Properties, compMapEntries)
```
fmeum, line 106: The order of properties is deterministic and may be relevant. I would lean towards not sorting them.

cli/printlog/compact/compact.go
```diff
@@ -0,0 +1,267 @@
...
+	inputs := slr.reconstructInputs(s.GetInputSetId())
+	toolInputs := slr.reconstructInputs(s.GetToolSetId())
+	var spawnInputs []*spb.File
+	for path, file := range inputs {
+		if _, ok := toolInputs[path]; ok {
+			file.IsTool = true
+		}
+		spawnInputs = append(spawnInputs, file)
+	}
+	if slr.sort {
```
fmeum, line 120: The old exec format didn't sort the individual fields of the spawn proto as those are already deterministically ordered (based on `depset` order). Instead, to make the log more comparable overall, we would need to sort the protos with respect to each other. 

Bazel topologically sorts by consumer-producer relationship and then by output path, see https://github.com/bazelbuild/bazel/blob/5d4feefed7e39b20a6c5deb5f74394abbf622a52/src/main/java/com/google/devtools/build/lib/exec/StableSort.java#L55. I think that this is what we should do if `-sort` is specified.

fmeum (changes requested):

cli/printlog/compact/compact.go
sluongng: Im a bit afraid of implementing this.

Currently, we are converting from the new compact format to the old format, which is big and could take a long time to sort.
But perhaps it's fine because we are not processing the log inside the invocation. 🤔 

Let me give it a try.

cli/printlog/compact/compact.go
sluongng: will remove this

cli/printlog/compact/compact.go
sluongng: will remove this

cli/printlog/BUILD
```diff
@@ -7,8 +7,8 @@ go_library(
     deps = [
         "//cli/arg",
         "//cli/log",
+        "//cli/printlog/compact",
```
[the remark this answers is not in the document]
sluongng: Will move it back in the next iteration.

cli/printlog/compact/compact.go
fmeum: I don't think it's possible to sort the new compact format directly as consumer/producer relationships can only be discovered after you have "flattened" the input/output file sets. If you want, we could take a look at this together at some point and see whether we can do better.

sluongng: I have removed the --sort for now.

This PR is now a lot smaller to fix obvious bugs and add some minor doc improvements.

PTAL.

fmeum (approved):

cli/printlog/printlog.go
```diff
@@ -203,6 +205,7 @@ func (slr *SpawnLogReconstructor) reconstructSpawn(s *spb.ExecLogEntry_Spawn) (*
 	}
 	se.ListedOutputs = listedOutputs
 	se.ActualOutputs = actualOutputs
+
```
sluongng, line 208:
```suggestion
	se.ActualOutputs = actualOutputs
```

Remove empty line

PR #6351, opened 2024-04-11
Update DB schema for workflows non-root migration

bduffany:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3286

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -589,7 +589,7 @@ type GitRepository struct {
 	// DefaultNonRootRunner decides whether workflow and remote bazel runs
 	// within this repository should run as a non-root user by default.
 	// TODO(http://go/b/3286): Remove this field after completing migration.
-	DefaultNonRootRunner bool `gorm:"not null;default:false"`
+	DefaultNonRootRunner bool `gorm:"not null;default:0"`
 }
 
 func (g *GitRepository) TableName() string {
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -585,6 +585,11 @@ type GitRepository struct {
 	// BuildBuddy-initiated invocations within this repo, such as workflows or
 	// remote Bazel.
 	InstanceNameSuffix string `gorm:"not null;default:''"`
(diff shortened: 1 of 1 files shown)


server/tables/tables.go
```diff
@@ -585,6 +585,11 @@ type GitRepository struct {
 	// BuildBuddy-initiated invocations within this repo, such as workflows or
 	// remote Bazel.
 	InstanceNameSuffix string `gorm:"not null;default:''"`
+
+	// DefaultNonRootRunner decides whether workflow and remote bazel runs
+	// within this repository should run as a non-root user by default.
+	// TODO(http://go/b/3286): Remove this field after completing migration.
+	DefaultNonRootRunner bool `gorm:"not null;default:false"`
```
maggie-lou, line 592: Can you change to `default:0`

Bools are stored as tiny ints. There's a bug in the gorm library where it thinks a migration is occurring where the default value is changing from false -> 0

server/tables/tables.go
```diff
@@ -585,6 +585,11 @@ type GitRepository struct {
 	// BuildBuddy-initiated invocations within this repo, such as workflows or
 	// remote Bazel.
 	InstanceNameSuffix string `gorm:"not null;default:''"`
+
+	// DefaultNonRootRunner decides whether workflow and remote bazel runs
+	// within this repository should run as a non-root user by default.
+	// TODO(http://go/b/3286): Remove this field after completing migration.
+	DefaultNonRootRunner bool `gorm:"not null;default:0"`
```
maggie-lou, line 592: nit: Would this be more readable inverted as `DefaultRootRunner`

maggie-lou (approved):

server/tables/tables.go
bduffany: slightly more readable but right now I don't want to spend the effort trying to figure out whether we'd also have to update our current creation queries to explicitly set this to true, so will just keep this as is. (This will go away soon anyway)

PR #6353, opened 2024-04-11
Disable debug logs in disconnectedContext.Deadline()

bduffany:
These debug logs are turning out too noisy and making the logs less usable.

**Related issues**: N/A

### DIFF
--- a/server/util/background/BUILD
+++ b/server/util/background/BUILD
@@ -5,7 +5,6 @@ go_library(
     srcs = ["background.go"],
     importpath = "github.com/buildbuddy-io/buildbuddy/server/util/background",
     visibility = ["//visibility:public"],
-    deps = ["//server/util/log"],
 )
 
 go_test(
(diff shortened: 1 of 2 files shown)


maggie-lou:
This is still really spammy if you're using `ExtendContextForFinalization`. I'm testing running a workflow locally, and it uses ExtendContextForFinalization to cleanup the runner. The logs are pretty unusable because this is logged hundreds (thousands?) of times. How would you feel about just deleting the log and adding a comment to ExtendContextForFinalization? We're already manually manipulating the context at that point, it feels fair enough to me to have a disclaimer that the deadline is not valid

bduffany:
@maggie-lou this log hints at an actual bug so I would like to fix these instances if you're seeing the debug log - do you mind locally adding `debug.PrintStack()` (from `runtime/debug`) just before the debug log in this file and sending me the stack trace?

vadimberezniker (approved):

bduffany:
I don't see a quick fix for the codepath that Maggie found so I went ahead and just disabled this completely for now. If this becomes more of an issue we can deal with it case-by-case

PR #6356, opened 2024-04-11
Make snaploader code more rollout safe before my proto change

maggie-lou:
If we have to rollback https://github.com/buildbuddy-io/buildbuddy/pull/6341/files, any snapshots written in the meantime will break, because there will be two auxiliary metadata entries (the vm config + the new vm metadata). Add this first to make rollouts safer.

I'll cherry pick and deploy this before #6341 goes out.

**Related issues**: N/A


bduffany (approved):
good catch

PR #6357, opened 2024-04-11
windows: don't depend on Bash

sluongng:
On Windows, `bash` is available through WSL. However WSL is only
available on host machines with Hyper-V and Virtualization enabled.

Instead of relying on `bash` to be available for our Windows build,
simply use PowerShell wherever possible.


sluongng:
closing this, we can assume `bash.exe` is available thanks to the MSYS2 dependency of Bazel.

PR #6362, opened 2024-04-12
Revert "workflows/remote-bazel: fix default branch not found"

maggie-lou:
Reverts buildbuddy-io/buildbuddy#6283

This PR makes a [remote git fetch](https://github.com/buildbuddy-io/buildbuddy/blob/master/cli/remotebazel/remotebazel.go#L340). We are currently only [initializing access to the local git repo](https://github.com/buildbuddy-io/buildbuddy/blob/master/cli/remotebazel/remotebazel.go#L263) in the cli, so this remote git fetch will fail for private repos ([Thread](https://buildbuddy-corp.slack.com/archives/C04S4M5U3BK/p1712877290612589?thread_ts=1710888921.160789&cid=C04S4M5U3BK))

If we want the cli to have remote git access, we need to either [grant it access to use the local SSH agent](https://stackoverflow.com/questions/75510923/in-go-git-how-to-authenticate-using-the-ssh-agent) or use some other mechanism

Given that fetching the default branch is just an optimization right now, revert for now while we discuss how we want to implement this

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -3,7 +3,6 @@ package remotebazel
 import (
 	"bytes"
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -16,7 +15,6 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


siggisim (approved):

sluongng (approved):
Technically we don't do a `git-fetch`, we should only do a `git ls-remote` which is lighter.
But for a private repository, even `ls-remote` is blocked.

I was not aware of the go-git issue with `ssh://` remote. Would shelling out to `git` binary work?

bduffany:
re working around this issue - would it work to use `origin/HEAD` as the default branch, but as an opaque string instead of actually trying to resolve it on the client side?

sluongng:
Since we did not merge this, it might be worth looking into https://github.com/buildbuddy-io/buildbuddy/pull/6369 as a possible alternative fix.

maggie-lou:
> re working around this issue - would it work to use `origin/HEAD` as the default branch, but as an opaque string instead of actually trying to resolve that ref to a branch name on the client side? The `ci_runner` should do a `git fetch` and `.git/refs/remotes/origin/HEAD` should contain the default branch name once the fetch has completed, IIUC

I think that should work. We don't really need the name of the default branch, we just need to be able to check it out, which we can directly do with the existing command in ci_runner `git checkout --force -B local_branch_name origin/HEAD`.

The other place we use the default branch name is as a fallback branch for snapshot routing. As long as we're consistent with using origin/HEAD, that should work okay too

maggie-lou:
> Since we did not merge this, it might be worth looking into #6369 as a possible alternative fix.

I prefer Brandon's approach, because it is faster and sufficient for our needs now. Going to merge this revert for now, and here's a ticket for implementing Brandon's approach: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3296

sluongng:
@bduffany i have not look closely into the server-side code, but at least on the client-side, we still need to know the default branch name in order to find the merge-base commit? 

We cannot assume origin/HEAD is available on the client side because evidently github actions does not create that symbolic ref 🤔

bduffany:
I don't think the client necessarily needs to compute the commit SHA of the default branch - that can be done in the CI runner no?

sluongng:
> I don't think the client necessarily needs to compute the commit SHA of the default branch - that can be done in the CI runner no?

Hmm that's true. We already make the default branch configurable on the client side via an environment variable.

PR #6364, opened 2024-04-12
Extend finalization deadline for pausing firecracker VMs

maggie-lou:
#6345 fixed a bug where in many cases, ExtendContextForFinalization would extend the context deadline ~indefinitely (7 days, to be specific). Now that it's fixed, certain code paths are failing because the finalization deadlines we'd set are too short

P99 to pause a firecracker VM takes 1.5min. Setting the deadline to 3min to be safe (P95 frequently reaches 1.5min as well)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -91,7 +91,7 @@ const (
 	// How long to spend waiting for a runner to be removed before giving up.
 	runnerCleanupTimeout = 30 * time.Second
 	// Allowed time to spend trying to pause a runner and add it to the pool.
-	runnerRecycleTimeout = 30 * time.Second
+	runnerRecycleTimeout = 3 * time.Minute
 	// How long to spend waiting for a persistent worker process to terminate
 	// after we send the shutdown signal before giving up.
 	persistentWorkerShutdownTimeout = 10 * time.Second


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -200,7 +200,7 @@ const (
 	guestVFSMountDir = "/vfs"
 
 	// How long to allow for the VM to be finalized (paused, outputs copied, etc.)
-	finalizationTimeout = 10 * time.Second
+	finalizationTimeout = 3 * time.Minute
```
bduffany, line 203: This is used for `remove()` and for syncing the workspace drive so I think we should keep the previous value

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -89,9 +89,9 @@ const (
 	removed
 
 	// How long to spend waiting for a runner to be removed before giving up.
-	runnerCleanupTimeout = 30 * time.Second
+	runnerCleanupTimeout = 3 * time.Minute
```
bduffany, line 92: I think this is unrelated to the pause timeouts so we might want to keep as-is unless we find that the current timeout is too short

bduffany (approved):
thanks for fixing this :pray:

PR #6369, opened 2024-04-15
remotebazel: replacing go-git with git shell

sluongng:
When calling to the remote repository that was setup with ssh protocol,
go-git would require additional auth configuration which might be harder
to obtain.

Replace go-git with plain "git" solves the problem as user should
already have authentication configured for git setup locally.

This also enables the usage of `ls-remote --symref`, which resolves the
symbolic refs on the remote origin. Since the only symbolic ref exists on
the remote side is HEAD, we can assume that this flag provide us the
correct default branch of the repo on the remote side.

Replaces #6362

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -176,7 +176,7 @@ func determineDefaultBranch(repo *git.Repository) (string, error) {
 	}
 
 	log.Debugf("Could not determine default git branch locally. Finding the default branch on the 'origin' remote.\n")
-	branch, remoteErr := remoteDefaultRef(repo)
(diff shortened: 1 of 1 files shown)


bduffany (commented):
What do you think about https://github.com/buildbuddy-io/buildbuddy/pull/6362#issuecomment-2052509035 as an alternative to trying to fetch the default branch from the remote at all?

Fundamentally, fetching the remote default branch on the client side is going to either be (a) slow or (b) possibly inconsistent with the remote. If possible it would be nice to determine the default branch on the ci_runner because it's already doing a fetch, so the `origin/HEAD` ref should always be up to date there.

sluongng:
Closing this for now since the alternative was merged.

PR #6370, opened 2024-04-15
proto: update remote_asset.proto

sluongng:
Including changes from:

- https://github.com/bazelbuild/remote-apis/commit/55153ba61dcf6277849562a30bca9fa3906ad9a0
- https://github.com/bazelbuild/remote-apis/commit/1f36c310b28d762b258ea577ed08e8203274efae

### DIFF
--- a/proto/remote_asset.proto
+++ b/proto/remote_asset.proto
@@ -205,6 +205,11 @@ message FetchBlobRequest {
   //
   // Specified qualifier names *MUST* be unique.
   repeated Qualifier qualifiers = 5;
+
+  // The digest function the server must use to compute the digest.
+  //
+  // If unset, the server SHOULD default to SHA256.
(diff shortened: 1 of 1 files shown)


fmeum (approved):
LGTM, but could we pull in these files from the pinned source of truth instead?

sluongng:
@fmeum we are currently using an elaborate setup in //proto. Such a change would require us to refactor the import paths of some packages which is inherently noisy. It would also affect workflows of folks who rely on the generated code to be under //proto for gopls completion. 

Im not saying it's impossible, but I suspect the pains currently out weight the gains. Let me know if you have ideas around this.

fmeum:
Yeah, if gopls support relies on this being an actual file, then I agree it's better to leave as is. Otherwise we could pull this in via `http_file` and then "move" into `//proto` repo via Skylib's `copy_file`, which maintains the path but of course only produces a non-source file.

PR #6372, opened 2024-04-15
Decrease default task timeout and maximum allowed timeout

bduffany:
Decrease default timeout to 8h and maximum allowed user-configurable timeout to 24h. Both are configurable as an executor flag.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/executor/executor.go
+++ b/enterprise/server/remote_execution/executor/executor.go
@@ -42,15 +42,12 @@ var (
 	// just repeat the last state change message after every
 	// execProgressCallbackPeriod. If this is set to 0, it is disabled.
 	execProgressCallbackPeriod = flag.Duration("executor.task_progress_publish_interval", 60*time.Second, "How often tasks should publish progress updates to the app.")
-
-	slowTaskThreshold = flag.Duration("executor.slow_task_threshold", 1*time.Hour, "Warn about tasks that take longer than this threshold.")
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/executor/executor.go
```diff
@@ -243,16 +238,12 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
 		return true, err
 	}
 	md.ExecutionStartTimestamp = timestamppb.Now()
-	maxDuration := infiniteDuration
-	if currentDeadline, ok := ctx.Deadline(); ok {
-		maxDuration = time.Until(currentDeadline)
-	}
```
bduffany, line 249: This block was not needed because the `WithTimeout` on line 255 below should already be respecting any existing deadline in the context.

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -92,16 +90,13 @@ func timevalDuration(tv syscall.Timeval) time.Duration {
 	return time.Duration(tv.Sec)*time.Second + time.Duration(tv.Usec)*time.Microsecond
 }
 
-func parseTimeout(timeout *durationpb.Duration, maxDuration time.Duration) (time.Duration, error) {
+func parseTimeout(timeout *durationpb.Duration) (time.Duration, error) {
 	if timeout == nil {
-		if maxDuration == 0 {
-			return infiniteDuration, nil
-		}
```
bduffany, line 99: I think this check was not needed because it could only be true if `time.Until(currentDeadline)` on line 248 happened to return exactly 0, and even in that case, the context will already be done so it doesn't matter what timeout we return from this func.

PR #6375, opened 2024-04-15
Only return EOF in cache proxy if a more descriptive error is not available

maggie-lou:
If we enter this if block, sendErr will always be `EOF` (due to the block at L629). There may be a more descriptive error in `err` which we should return instead

**Related issues**: Came upon this debugging https://github.com/buildbuddy-io/buildbuddy-internal/issues/3263

### DIFF
--- a/enterprise/server/util/cacheproxy/cacheproxy.go
+++ b/enterprise/server/util/cacheproxy/cacheproxy.go
@@ -633,7 +633,7 @@ func (wc *streamWriteCloser) Commit() error {
 	if status.IsAlreadyExistsError(err) {
 		return nil
 	}
-	if sendErr != nil {
+	if err == nil && sendErr != nil {
 		return sendErr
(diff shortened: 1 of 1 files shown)


siggisim (approved):

tylerwilliams (approved):

vadimberezniker (approved):

enterprise/server/util/cacheproxy/cacheproxy.go
```diff
@@ -633,7 +633,7 @@ func (wc *streamWriteCloser) Commit() error {
 	if status.IsAlreadyExistsError(err) {
 		return nil
 	}
-	if sendErr != nil {
+	if err == nil && sendErr != nil {
```
jdhollen, line 636: This could just be `if err != nil { return err }`, right? (And return `senderr` below)

PR #6376, opened 2024-04-16
Use block-based size for filecache entry size

bduffany:
When calling `AddFile`, stat the file to get the number of blocks consumed by the file, and compute the size based on the number of blocks consumed, rather than the content stored in the blocks. The extra `stat()` syscall does not significantly affect performance according to benchmarks:

```
goos: linux
goarch: amd64
cpu: AMD Ryzen 9 7950X 16-Core Processor            
                                   │ /tmp/bench_false.txt │        /tmp/bench_true.txt         │
                                   │        sec/op        │   sec/op     vs base               │
FilecacheLink/100%Link/0%Add/1K-32            15.52m ± 1%   15.57m ± 1%       ~ (p=0.631 n=10)
FilecacheLink/95%Link/5%Add/1K-32             14.77m ± 2%   14.75m ± 3%       ~ (p=0.436 n=10)
geomean                                       15.14m        15.15m       +0.07%
```

A future improvement could be to also account for the file metadata, not just the file contents. But just computing the block-based size should be good enough for now.

**Related issues**:
* Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3300
* Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3248

### DIFF
--- a/enterprise/server/remote_execution/filecache/filecache.go
+++ b/enterprise/server/remote_execution/filecache/filecache.go
@@ -258,14 +258,14 @@ func (c *fileCache) addFileToGroup(groupID string, node *repb.FileNode, existing
 	// which is not good.
 	k := groupSpecificKey(groupID, node)
 
-	sizeOnDisk := node.GetDigest().GetSizeBytes()
 	info, err := os.Stat(existingFilePath)
 	if err != nil {
 		return status.WrapError(err, "stat")
 	}
-	// Stat always returns number of 512 byte blocks, regardless of the FS
-	// settings.
-	sizeOnDisk = info.Sys().(*syscall.Stat_t).Blocks * 512
+	sizeOnDisk, err := disk.EstimatedFileDiskUsage(info)
+	if err != nil {
+		return status.WrapError(err, "estimate disk usage")
+	}
 
 	c.lock.Lock()
 	defer c.lock.Unlock()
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -261,6 +258,15 @@ func (c *fileCache) addFileToGroup(groupID string, node *repb.FileNode, existing
...
 	k := groupSpecificKey(groupID, node)
 
+	sizeOnDisk := node.GetDigest().GetSizeBytes()
+	info, err := os.Stat(existingFilePath)
+	if err != nil {
+		return status.WrapError(err, "stat")
+	}
+	// Stat always returns number of 512 byte blocks, regardless of the FS
+	// settings.
+	sizeOnDisk = info.Sys().(*syscall.Stat_t).Blocks * 512
```
vadimberezniker, line 268: I'm guessing this doesn't work on Windows. I'm not sure what the status of Windows executors are, but we may need to special case this calculation for this to compile on Windows @sluongng

sluongng: Windows executor should be functional on HEAD. If you guys need to test something, please ping me on Slack.

PR #6381, opened 2024-04-16
[RB] Automatically authorize remote bazel runs if a workflow for the same repo has been configured

maggie-lou:
**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3293


enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -220,16 +223,62 @@ func (r *runnerService) withCredentials(ctx context.Context, req *rnpb.RunReques
...
 		{Name: "BUILDBUDDY_API_KEY", Value: apiKey.Value},
 		{Name: "REPO_USER", Value: req.GetGitRepo().GetUsername()},
-		{Name: "REPO_TOKEN", Value: req.GetGitRepo().GetAccessToken()},
+		{Name: "REPO_TOKEN", Value: accessToken},
 	}
 	ctx = withEnvOverrides(ctx, envOverrides)
 	return ctx, nil
 }
 
+func (r *runnerService) getGitToken(ctx context.Context, groupID string, repoURL string) (string, error) {
```
bduffany, line 256: nit: avoid passing in the group ID here (best to avoid plumbing around group IDs as raw strings as much as we can, since strings don't inherently encode any auth info)

if you call `u.GetGroupID()` inside this func then you can also skip calling AuthorizeGroupAccess here because `GetGroupID` always returns an authorized group ID

enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -220,16 +223,62 @@ func (r *runnerService) withCredentials(ctx context.Context, req *rnpb.RunReques
...
+	// If no access token is provided explicitly, try fetching the token.
+	accessToken := req.GetGitRepo().GetAccessToken()
+	if accessToken == "" {
+		repoURL, err := git.NormalizeRepoURL(req.GetGitRepo().GetRepoUrl())
+		if err != nil {
+			return nil, status.WrapError(err, "normalize git repo url")
+		}
+		isGitRepo := repoURL.Hostname() == "github.com"
+
+		if isGitRepo {
```
bduffany, line 235: is this check needed?

maggie-lou: I guess not - removed

bduffany (approved):

PR #6382, opened 2024-04-16
fetch_server: support digest function and blake3

sluongng:
Implement support for digest function in FetchBlobRequest as well as
usage of Blake3 as a digest func.

### DIFF
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -313,7 +313,7 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 		if err != nil {
 			return nil, status.UnavailableErrorf("failed to compute checksum digest: %s", err)
 		}
-		if expectedChecksum != "" && checksumDigestRN.GetDigest().Hash != expectedChecksum {
+		if expectedChecksum != "" && checksumDigestRN.GetDigest().GetHash() != expectedChecksum {
 			return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, checksumDigestRN.GetDigest().Hash, expectedChecksum)
 		}
 	}
@@ -321,7 +321,7 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 	if err != nil {
 		return nil, status.UnavailableErrorf("failed to add object to cache: %s", err)
 	}
-	// If the requsted digestFunc is supplied is the same with the checksum sri,
+	// If the requested digestFunc is supplied is the same with the checksum sri,
 	// verify the expected checksum of the downloaded file after storing it in our cache.
 	if checksumFunc == storageFunc && expectedChecksum != "" && blobDigest.Hash != expectedChecksum {
 		return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, blobDigest.Hash, expectedChecksum)
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -231,7 +231,7 @@ func (p *FetchServer) findBlobInCache(ctx context.Context, instanceName string, 
 		SizeBytes: 1,
 	}
 	cacheRN := digest.NewResourceName(blobDigest, instanceName, rspb.CacheType_CAS, checksumFunc)
-	log.CtxInfof(ctx, "Looking up %s in cache", blobDigest.Hash)
+	log.CtxDebugf(ctx, "Looking up %s in cache", blobDigest.Hash)
 
 	// Lookup metadata to get the correct digest size to be returned to
 	// the client.
@@ -257,7 +257,7 @@ func (p *FetchServer) findBlobInCache(ctx context.Context, instanceName string, 
 		return nil
 	}
 
-	log.CtxInfof(ctx, "FetchServer found %s in cache", digest.String(blobDigest))
+	log.CtxDebugf(ctx, "FetchServer found %s in cache", digest.String(blobDigest))
 	return blobDigest
 }
 
@@ -266,7 +266,7 @@ func (p *FetchServer) findBlobInCache(ctx context.Context, instanceName string, 
 // expectedChecksum (if non-empty), and if there is a mismatch then an error is
 // returned.
 func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteInstanceName string, httpClient *http.Client, uri string, storageFunc repb.DigestFunction_Value, checksumFunc repb.DigestFunction_Value, expectedChecksum string) (*repb.Digest, error) {
-	log.CtxInfof(ctx, "Fetching %s", uri)
+	log.CtxDebugf(ctx, "Fetching %s", uri)
 	rsp, err := httpClient.Get(uri)
 	if err != nil {
 		return nil, status.UnavailableErrorf("failed to fetch %q: HTTP GET failed: %s", uri, err)
@@ -326,7 +326,7 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 	if checksumFunc == storageFunc && expectedChecksum != "" && blobDigest.Hash != expectedChecksum {
 		return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, blobDigest.Hash, expectedChecksum)
 	}
-	log.CtxInfof(ctx, "Mirrored %s to cache (digest: %s)", uri, digest.String(blobDigest))
+	log.CtxDebugf(ctx, "Mirrored %s to cache (digest: %s)", uri, digest.String(blobDigest))
 	return blobDigest, nil
 }
 
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -31,8 +31,9 @@ import (
 )
 
 const (
-	checksumQualifier = "checksum.sri"
+	ChecksumQualifier = "checksum.sri"
 	sha256Prefix      = "sha256-"
+	blake3Prefix      = "blake3-"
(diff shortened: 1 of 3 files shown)


server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -203,11 +186,70 @@ func (p *FetchServer) FetchDirectory(ctx context.Context, req *rapb.FetchDirecto
...
+	return blobDigest
+}
+
 // mirrorToCache uploads the contents at the given URI to the given cache,
 // returning the digest. The fetched contents are checked against the given
-// expectedSHA256 (if non-empty), and if there is a mismatch then an error is
+// expectedHash (if non-empty), and if there is a mismatch then an error is
 // returned.
-func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteInstanceName string, httpClient *http.Client, uri, expectedSHA256 string) (*repb.Digest, error) {
+func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteInstanceName string, httpClient *http.Client, uri string, storageFunc repb.DigestFunction_Value, checksumFunc repb.DigestFunction_Value, checksum string) (*repb.Digest, error) {
```
bduffany, line 252: nit: rename `checksum` to `expectedHash` to match the function docs. (Or `expectedChecksum` if you like "checksum" better - I think the "expected" part is useful because this is not the verified checksum, it's the unverified request checksum)

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -246,12 +288,27 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 			log.Errorf("Failed to remove temp file: %s", err)
 		}
 	}()
-	blobDigest, err := cachetools.UploadFile(ctx, bsClient, remoteInstanceName, repb.DigestFunction_SHA256, tmpFilePath)
+
+	// If the requsted digestFunc is supplied and differ from the checksum sri,
```
bduffany, line 292:
```suggestion
	// If the requested digestFunc is supplied and differs from the checksum sri,
```

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -109,52 +108,36 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
 		return nil, err
 	}
 
-	var expectedSHA256 string
-
+	storageFunc := req.GetDigestFunction()
+	if storageFunc == repb.DigestFunction_UNKNOWN {
+		storageFunc = repb.DigestFunction_SHA256
+	}
+	var checksumFunc repb.DigestFunction_Value
```
bduffany, line 115: it seems inconvenient for servers that the storage digest can differ from the checksum digest. Is it too late to change this design? If not, I wonder if it'd make sense to enforce that they are equal (by ignoring `--digest_function` just for remote asset fetches and instead just using whatever digest has been supplied in the fetch target)

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -203,11 +186,70 @@ func (p *FetchServer) FetchDirectory(ctx context.Context, req *rapb.FetchDirecto
...
+	if err != nil {
+		log.CtxErrorf(ctx, "Failed to renew %s: %s", digest.String(blobDigest), err)
+		return nil
+	}
+	if !exists {
+		log.CtxInfof(ctx, "Blob %s expired before we could renew it", digest.String(blobDigest))
+		return nil
+	}
+
+	// If the digestFunc is supplied and differ from the checksum sri,
```
bduffany, line 226: I feel like it's a bit non-intuitive to be doing an upload in `findBlobInCache` - maybe split this into a separate func?

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -203,11 +186,70 @@ func (p *FetchServer) FetchDirectory(ctx context.Context, req *rapb.FetchDirecto
...
+	// after looking up the cached blob using checksum sri, re-upload
+	// that blob using the requested digestFunc.
+	if checksumFunc != storageFunc {
+		b, err := cache.Get(ctx, cacheRN.ToProto())
```
bduffany, line 230: `cache.Get()` will buffer the full repository in memory and these can be several GB in size. We should instead use a streaming approach, calling `cache.Reader` then passing that to `UploadBlob`

bduffany (changes requested):

server/remote_asset/fetch_server/fetch_server.go
sluongng: We could add comments into the remote-apis spec with something like "MUST be equal" or something similar?

Realistically, I have implemented it on the Bazel side so that the CRI checksum will be the same as the request's digest function. So we should only expect misuse from non-Bazel clients.

server/remote_asset/fetch_server/fetch_server.go
sluongng: This is what I tried initially, but UploadBlob requires an `io.ReadSeeker` and cache.Reader was missing the "Seeker" interface 🤔 

Let me see if there is a nicer way.

server/remote_asset/fetch_server/fetch_server.go
sluongng: Fixed

server/remote_asset/fetch_server/fetch_server.go
sluongng: Renamed all to `expectedChecksum` to be consistent with the qualifier name.

server/remote_asset/fetch_server/fetch_server.go
sluongng: Moved to a separate func.

server/remote_asset/fetch_server/fetch_server.go
sluongng: I could be wrong but it seems like we can't seek on the blob content because it's in pebble cache and could be inline 🤔

For disk cache, there could be a way around it but that would require working around the Cache interface (or adding a new method).

For now, I took a similar approach as the HTTP download code path: write to a temp file and then copy its content to the cache.

bduffany: You could use `cachetools.UploadFromReader` instead

server/remote_asset/fetch_server/fetch_server.go
bduffany: I think adding the "MUST be equal" comment would be ideal. And if they're not equal we could return an error

sluongng: Raised https://github.com/bazelbuild/remote-apis/pull/294

It might take a while to get merged though. I think we should proceed with the current implementation and remove the redundant code once the API spec is updated. WDYT?

bduffany: sgtm

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -203,11 +192,80 @@ func (p *FetchServer) FetchDirectory(ctx context.Context, req *rapb.FetchDirecto
...
+	blobDigest := &repb.Digest{
+		Hash: expectedChecksum,
+		// The digest size is unknown since the client only sends up
+		// the hash. We can look up the size using the Metadata API,
+		// which looks up only using the hash, so the size we pass here
+		// doesn't matter.
+		SizeBytes: 1,
+	}
+	cacheRN := digest.NewResourceName(blobDigest, instanceName, rspb.CacheType_CAS, checksumFunc)
+	log.CtxInfof(ctx, "Looking up %s in cache", blobDigest.Hash)
```
bduffany, line 234: wdyt about removing these log.Info lines in the non-error cases? I think they might be a tad excessive

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -246,12 +304,27 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
...
+
+	// If the requested digestFunc is supplied and differ from the checksum sri,
+	// verify the downloaded file with the checksum sri before storing it to
+	// our cache.
+	if checksumFunc != storageFunc {
+		checksumDigestRN, err := cachetools.ComputeFileDigest(tmpFilePath, remoteInstanceName, checksumFunc)
+		if err != nil {
+			return nil, status.UnavailableErrorf("failed to compute checksum digest: %s", err)
+		}
+		if expectedChecksum != "" && checksumDigestRN.GetDigest().Hash != expectedChecksum {
```
bduffany, line 316:
```suggestion
		if expectedChecksum != "" && checksumDigestRN.GetDigest().GetHash() != expectedChecksum {
```

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -246,12 +304,27 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
...
+			return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, checksumDigestRN.GetDigest().Hash, expectedChecksum)
+		}
+	}
+	blobDigest, err := cachetools.UploadFile(ctx, bsClient, remoteInstanceName, storageFunc, tmpFilePath)
 	if err != nil {
 		return nil, status.UnavailableErrorf("failed to add object to cache: %s", err)
 	}
-	if expectedSHA256 != "" && blobDigest.Hash != expectedSHA256 {
-		return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, blobDigest.Hash, expectedSHA256)
+	// If the requsted digestFunc is supplied is the same with the checksum sri,
```
bduffany, line 324:
```suggestion
	// If the requested digestFunc is supplied is the same with the checksum sri,
```

bduffany (approved):

server/remote_asset/fetch_server/fetch_server.go
sluongng: Good suggestion.

I will switch these to Debug level instead of removing them.
These could be fairly useful for development purposes.

PR #6383, opened 2024-04-16
Clean up cache stats on channel close

tempoz:
- Added a function in channel close to ensure cache stats are cleaned up exactly once no matter what.

- Added some logging to make this issue easier to track if it happens again, since I can't repro it and it only seems to happen in prod, so I can't be sure this is an actual fix.

- Did a small refactor to not attempt to process an empty channel.

- Removed an error case that cannot occur (there is no way that we can return `AlreadyExistsError` from `HandleEvent`).

**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -162,7 +162,21 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
 	b.cancelFnsByInvID.Store(iid, cancel)
 
 	b.openChannels.Add(1)
+
+	enqueueStatsOnce := &sync.Once{}
 	onClose := func() {
+		enqueueStatsOnce.Do(func() {
+			persist := &PersistArtifacts{}
+			if !*disablePersistArtifacts {
+				testOutputURIs := buildEventAccumulator.TestOutputURIs()
+				persist.URIs = make([]*url.URL, 0, len(testOutputURIs))
+				persist.URIs = append(persist.URIs, buildEventAccumulator.BuildToolLogURIs()...)
+				persist.URIs = append(persist.URIs, testOutputURIs...)
+			}
+
+			b.statsRecorder.Enqueue(ctx, invocation, persist)
+			log.CtxInfo(ctx, "Enqueued for stats recording in close.")
+		})
 		b.openChannels.Done()
 		b.cancelFnsByInvID.Delete(iid)
 	}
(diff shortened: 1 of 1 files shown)


server/build_event_protocol/build_event_handler/build_event_handler.go
```diff
@@ -162,7 +162,21 @@ func (b *BuildEventHandler) OpenChannel(ctx context.Context, iid string) interfa
...
+			persist := &PersistArtifacts{}
+			if !*disablePersistArtifacts {
+				testOutputURIs := buildEventAccumulator.TestOutputURIs()
+				persist.URIs = make([]*url.URL, 0, len(testOutputURIs))
+				persist.URIs = append(persist.URIs, buildEventAccumulator.BuildToolLogURIs()...)
+				persist.URIs = append(persist.URIs, testOutputURIs...)
+			}
+
+			b.statsRecorder.Enqueue(ctx, invocation, persist)
+			log.CtxInfo(ctx, "Enqueued for stats recording in close.")
```
bduffany, line 178: Looking through the logic in `build_event_server.go` it looks like we always call `FinalizeInvocation` except in 2 cases:
1. If `HandleEvent` returns an `AlreadyExists` error [here](https://github.com/buildbuddy-io/buildbuddy/blob/8fc5a88f4189fc20a8dfd21c6c4fdf504d54095f/server/build_event_protocol/build_event_server/build_event_server.go#L167-L170): in this case I guess we don't want to run this logic since another invocation would already have been responsible for this, I think?
2. If we see an unexpected sequence number [here](https://github.com/buildbuddy-io/buildbuddy/blob/8fc5a88f4189fc20a8dfd21c6c4fdf504d54095f/server/build_event_protocol/build_event_server/build_event_server.go#L203-L206) - this is pretty rare, so I think we don't need to worry about it too much?

We also log for both of these cases in case you want to see how often this is happening today

Another thing is that `FinalizeInvocation` can potentially short-circuit if there's an error, but we log those too

Here is a link with all the relevant logs: https://console.cloud.google.com/logs/query;query=resource.labels.namespace_name%3D%22buildbuddy-prod%22%0AsourceLocation.file%3D%22build_event_server.go%22%0A%22Missing%20ack%22%20OR%20%22AlreadyExistsError%20handling%20event%22%20OR%20%22Error%20finalizing%20invocation%22;cursorTimestamp=2024-04-16T13:35:05.707808731Z;duration=PT12H?project=flame-build

tempoz: Note that HandleEvent can't return `AlreadyExists`, as I said in the PR description, but in light of the fact that it looks like the scorecards are leftover from cache-only invocations, I don't think that the clean up in close is necessary any longer and I'll remove it.

PR #6384, opened 2024-04-16
[Raft] Add a direct delete request

luluz66:
**Related issues**: N/A
DirectDelete is used to delete keys in meta range. The current Delete method
only allows us to delete file records. We need this methods to maintain
transaction records.

### DIFF
--- a/enterprise/server/raft/replica/replica.go
+++ b/enterprise/server/raft/replica/replica.go
@@ -803,6 +803,15 @@ func (sm *Replica) directWrite(wb pebble.Batch, req *rfpb.DirectWriteRequest) (*
 	return &rfpb.DirectWriteResponse{}, err
 }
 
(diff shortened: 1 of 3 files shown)


enterprise/server/raft/replica/replica.go
```diff
@@ -803,6 +803,15 @@ func (sm *Replica) directWrite(wb pebble.Batch, req *rfpb.DirectWriteRequest) (*
 	return &rfpb.DirectWriteResponse{}, err
 }
 
+func (sm *Replica) directDelete(wb pebble.Batch, req *rfpb.DirectDeleteRequest) (*rfpb.DirectDeleteResponse, error) {
+	if !isLocalKey(req.GetKey()) {
+		return nil, status.InvalidArgumentErrorf("cannot direct delete non-local key")
```
tylerwilliams, line 808: nit: "cannot direct delete non-local key; use Delete instead"

tylerwilliams (approved):

PR #6385, opened 2024-04-16
Temporarily make VM logging opt-in only

bduffany:
**Related issues**: N/A


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -449,6 +449,7 @@ func (p *Provider) New(ctx context.Context, props *platform.Properties, task *re
 			NumCpus:           numCPUs,
 			MemSizeMb:         int64(math.Max(1.0, float64(sizeEstimate.GetEstimatedMemoryBytes())/1e6)),
 			ScratchDiskSizeMb: int64(float64(sizeEstimate.GetEstimatedFreeDiskBytes()) / 1e6),
+			EnableLogging:     platform.IsTrue(platform.FindEffectiveValue(task.GetExecutionTask(), "debug-enable-vm-logs")),
```
maggie-lou, line 452: nit: I'm okay with you merging this now so it makes it into the release, but it would be nice if we pulled this from `props` so there's a more standardized way of propagating platform data

maggie-lou (approved):

PR #6393, opened 2024-04-17
Small clean-up of build event handler logic

tempoz:
Small refactor primarily to try to improve readability.

- Added some logging.

- Did a small refactor to not attempt to process an empty channel.

- Removed an error case that cannot occur (there is no way that we can return `AlreadyExistsError` from `HandleEvent`).

**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/build_event_server/build_event_server.go
+++ b/server/build_event_protocol/build_event_server/build_event_server.go
@@ -159,15 +159,12 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 				streamID = in.OrderedBuildEvent.StreamId
 				ctx = log.EnrichContext(ctx, log.InvocationIDKey, streamID.InvocationId)
 				channel = s.env.GetBuildEventHandler().OpenChannel(ctx, streamID.InvocationId)
+				log.CtxInfof(ctx, "Opened a channel for stream ID %s.", streamID)
 				channelDone = channel.Context().Done()
 				defer channel.Close()
 			}
 
 			if err := channel.HandleEvent(in); err != nil {
-				if status.IsAlreadyExistsError(err) {
-					log.CtxWarningf(ctx, "AlreadyExistsError handling event; this means the invocation already exists and may not be retried: %s", err)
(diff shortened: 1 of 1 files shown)


server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -159,15 +159,12 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 				streamID = in.OrderedBuildEvent.StreamId
 				ctx = log.EnrichContext(ctx, log.InvocationIDKey, streamID.InvocationId)
 				channel = s.env.GetBuildEventHandler().OpenChannel(ctx, streamID.InvocationId)
+				log.CtxInfof(ctx, "Opened a channel for stream ID %s.", streamID)
```
bduffany, line 162: nit: I think streamID is a struct - since the format string has `%s` idk if you're expecting it to be a string?

The context is enriched with the invocation ID so something like this might be fine:

```suggestion
				log.CtxInfo(ctx, "Opened invocation channel")
```

server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -184,7 +181,12 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 }
 
 func postProcessStream(ctx context.Context, channel interfaces.BuildEventChannel, streamID *bepb.StreamId, acks []int, stream pepb.PublishBuildEvent_PublishBuildToolEventStreamServer) error {
-	if channel != nil && channel.GetNumDroppedEvents() > 0 {
+	if channel == nil {
+		log.CtxInfo(ctx, "Closing empty channel.")
```
bduffany, line 185: nit: maybe move this check outside of `postProcessStream` and change the info message to "No events received on stream"

bduffany (approved):

server/build_event_protocol/build_event_server/build_event_server.go
tempoz: Done!

PR #6403, opened 2024-04-17
Display `testSuite` time when parsing `test.xml` if populated

siggisim:
If not populated, fallback to the overall test time like we currently do.

### DIFF
--- a/app/target/target_test_cases_card.tsx
+++ b/app/target/target_test_cases_card.tsx
@@ -66,12 +66,14 @@ export default class TargetTestCasesCardComponent extends React.Component<Props>
             <div className="title">{this.props.testSuite.getAttribute("name")}</div>
             <div className="test-subtitle">
               {testCases.length} {testCases.length == 1 ? "test" : "tests"} {this.getStatusTitle()} in{" "}
-              {format.durationMillis(
-                durationToMillisWithFallback(
-                  this.props.buildEvent?.testResult?.testAttemptDuration,
-                  this.props.buildEvent?.testResult?.testAttemptDurationMillis || 0
-                )
-              )}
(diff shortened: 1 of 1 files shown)


app/target/target_test_cases_card.tsx
```diff
@@ -66,12 +66,14 @@ export default class TargetTestCasesCardComponent extends React.Component<Props>
...
             <div className="test-subtitle">
               {testCases.length} {testCases.length == 1 ? "test" : "tests"} {this.getStatusTitle()} in{" "}
-              {format.durationMillis(
-                durationToMillisWithFallback(
-                  this.props.buildEvent?.testResult?.testAttemptDuration,
-                  this.props.buildEvent?.testResult?.testAttemptDurationMillis || 0
-                )
-              )}
+              {this.props.testSuite.getAttribute("time")
+                ? `${this.props.testSuite.getAttribute("time")} s`
```
jdhollen, line 70: could you do `durationSec` here so that the formatting matches the other case?

siggisim: It's kinda tricky because this `time` attribute is a string like "0.7" that comes straight from the xml. 

This is the same as how we handle it below at the test case level on line 88/90 https://github.com/buildbuddy-io/buildbuddy/pull/6403/files#diff-34df2b08917df5b5945ad24cb3a3c3c8e4e8e2d38c0eb42b2c80ada22bc88055R90

I could parse the string into a number and then format it both here - I guess I'm mostly just worried about someone putting some non-parseable stuff into that time="" xml field (like 0.7s or something).

Lmk if you have strong opinions, and I could do the string parsing here and below.

Issue #6404, opened 2024-04-18
Show analysis phase errors explicitly


evis:
**Is your feature request related to a problem? Please describe.**

Case:
1. Run `bazel build --keep-going` on CI to see all problems in build
2. It has only one problem, during analysis phase

What happens:
1. Buildbuddy doesn't show this problem explicitly
2. It's quite hard to find this error in log. Log example: https://gist.github.com/evis/791843039e9655975a818182b66940e4 — the failure is
```
[2024-04-17 16:33:14] ERROR: /place/sandbox-data/tasks/5/8/2283762585/mounted_arcadia/classifieds/verticals-backend/etc-mono/services/anubis/cli/BUILD:55:11: in deps attribute of scala_test rule //etc-mono/services/anubis/cli:tests: rule '//etc-mono/services/anubis/cli:test_java_proto' does not exist. Since this rule was created by the macro 'scala_test', the error might have been caused by the macro implementation
[2024-04-17 16:33:14] ERROR: /place/sandbox-data/tasks/5/8/2283762585/mounted_arcadia/classifieds/verticals-backend/etc-mono/services/anubis/cli/BUILD:55:11: Analysis of target '//etc-mono/services/anubis/cli:tests' failed
```

It's not at the end of my build, it's somewhere between it's begin and end.

**Describe the solution you'd like**
Show analysis phase errors in buildbuddy UI explicitly (maybe, on Targets tab).

**Describe alternatives you've considered**
1. Run `bazel build` without `--keep-going` option. But then, we won't see all failures.
2. Run `bazel build --keep-going --no-build`, and then `bazel build --keep-going`. But then, the build will be slower.

evis:
Wow, that's amazing! Thank you so much, will be waiting for the new release!

PR #6407, opened 2024-04-18
Rework invocation error card and render more types of errors

bduffany:
* Structure errors as a list rather than trying to pick out a single error from the build
* Support rendering more types of errors
* Fetch failed action stdout (not just stderr) - e.g. TypeScript actions put their errors on stdout. So now we show typecheck errors directly in this card

---

Before:
image

After:
image

---

Before:
image

After:
image

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy/issues/6404

### DIFF
--- a/app/invocation/invocation_error_card.tsx
+++ b/app/invocation/invocation_error_card.tsx
@@ -3,75 +3,174 @@ import React from "react";
 import InvocationModel from "./invocation_model";
 import { failure_details } from "../../proto/failure_details_ts_proto";
 import TerminalComponent from "../terminal/terminal";
-import rpc_service from "../service/rpc_service";
+import rpc_service, { CancelablePromise } from "../service/rpc_service";
+import { exitCode } from "../util/exit_codes";
+import { build_event_stream } from "../../proto/build_event_stream_ts_proto";
 
 const debugMessage =
   "Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging";
+
 interface Props {
   model: InvocationModel;
 }
 
 interface State {
-  stdErr: string;
+  model?: CardModel;
 }
+
+type CardModel = {
+  errors: ErrorData[];
+};
+
+type ErrorData = {
+  action?: build_event_stream.ActionExecuted;
+  actionStderr?: string;
+  actionStdout?: string;
+
+  aborted?: build_event_stream.BuildEvent;
+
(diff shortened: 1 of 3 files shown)


app/invocation/invocation_error_card.tsx
```diff
@@ -3,75 +3,174 @@ import React from "react";
...
+    const model = this.getModel(this.props);
+    if (!modelsEqual(this.getModel(this.props), this.getModel(prevProps))) {
+      this.fetchModelData(model);
     }
   }
 
-  prettyPrintedStdErr() {
-    if (!this.state.stdErr) {
-      return "";
+  getModel(props = this.props): CardModel {
```
jdhollen, line 52: small nit, i feel like defaulting is kinda error-prone here, and given that you only use it this way in `componentDidMount`, you might as well make this a standalone function

app/invocation/invocation_error_card.tsx
```diff
@@ -3,75 +3,174 @@ import React from "react";
...
-    description = description.replaceAll(debugMessage, `\x1b[90m${debugMessage}\x1b[0m\n \n`);
+    this.inFlightFetch = new CancelablePromise(Promise.all(promises));
+    this.inFlightFetch.then(() => this.setState({ model }));
+  }
+
+  getTitle(model: CardModel) {
+    // If there was a failed action with a non-zero exit code, use that as the title.
+    for (const error of model.errors) {
+      if ((error.action?.exitCode ?? 0) !== 0) {
+        return `${this.props.model.failedAction?.action?.type} action failed with exit code ${this.props.model.failedAction?.action?.exitCode}`;
```
jdhollen, line 103: any better value to put here for `type` ? i guess this is fine, it just drives me nuts that nullishness isn't explicitly linked between the two.

app/invocation/invocation_error_card.tsx
```diff
@@ -3,75 +3,174 @@ import React from "react";
...
+      }
+    }
+    // If the finished event exists and there's a non-zero exit code, use that as the card title.
+    for (const error of model.errors) {
+      if ((error.finished?.exitCode ?? 0) !== 0) {
+        return exitCode(error.finished?.exitCode?.name ?? "");
```
jdhollen, line 109: i think all of the wildcards on the return here are not needed and `name` is defined, right?

jdhollen (approved):

PR #6412, opened 2024-04-19
Use monaco to show PR file diffs

jdhollen:
UI is still a bit scratchy, but i'm trying to vaguely control the size of this PR.  Definitely probably works.

**Related issues**: N/A


siggisim (approved):

jdhollen:
@bduffany siggi wants you to look at my janky zones and portals and see if you know a cooler way to do all of this

siggisim:
> @bduffany siggi wants you to look at my janky zones and portals and see if you know a cooler way to do all of this

haha i'm cool with the portals, just curious if you had any react tricks that would generally make this interaction between monaco and react nicer

bduffany:
> @bduffany siggi wants you to look at my janky zones and portals and see if you know a cooler way to do all of this

I don't know much about Monaco extensions etc. but from the comment on `AutoZone` it sounded like zones are the way to render stuff in the editor, but can't contain arbitrary HTML? If so I think the current approach with portals sounds fine and if we find it too painful we can always try to figure out something better later on

jdhollen:
> > @bduffany siggi wants you to look at my janky zones and portals and see if you know a cooler way to do all of this
> 
> I don't know much about Monaco extensions etc. but from the comment on `AutoZone` it sounded like zones are the way to render stuff in the editor, but they can't contain arbitrary HTML? If so I think the current approach with portals sounds fine and if we run into limitations then we can always try to figure out something better later on

We can put arbitrary HTML in zones, but Monaco swallows the mouse / keyboard events, so it's no good for interactive stuff.

jdhollen:
The github folks told me there's nothing to be done about the weird post-merge state that this PR wound up in other than closing it, so I'm closing it.

bduffany:
> We can put arbitrary HTML in zones, but Monaco swallows the mouse / keyboard events, so it's no good for interactive stuff.

Did you already see https://stackoverflow.com/a/59145089 ?

jdhollen:
> > We can put arbitrary HTML in zones, but Monaco swallows the mouse / keyboard events, so it's no good for interactive stuff.
> 
> Did you already see https://stackoverflow.com/a/59145089 ?

Yes--sorry, replying too early.

- As the comment down at the bottom of the top answer says, zones still get some events swallowed even if you try to play tricks with their html.  Not using any of them right now, but it just didn't seem worth it.
- You still have to do a bit of song and dance to size the zone properly for its content, so there's no real layout benefit.

PR #6413, opened 2024-04-19
export metrics: fix mac executor metrics

luluz66:
for mac executors pod_name is not set, fallback to inventory_hostname instead

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3317

### DIFF
--- a/enterprise/server/backends/prom/prom.go
+++ b/enterprise/server/backends/prom/prom.go
@@ -28,7 +28,8 @@ import (
 )
 
 const (
-	podNameLabel = "pod_name"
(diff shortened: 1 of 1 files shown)


tylerwilliams:
I wish we could fix this in a way that didn't involve adding a fallback like this. WDYT?

luluz66:
> I wish we could fix this in a way that didn't involve adding a fallback like this. WDYT?

see https://github.com/buildbuddy-io/buildbuddy-internal/pull/3321

PR #6414, opened 2024-04-19
Pull `GitRepo` and `RepoState` into a separate `git.proto`

siggisim:
This removes a duplicate proto definition between `runner.proto` and `workflow.proto`.

This also paves the way to use these protos to define a "Workspace".


tylerwilliams (approved):
LGTM

PR #6416, opened 2024-04-19
Fix incorrect mediaType in arm64 executor image layers

bduffany:
The `container_push` rule from `rules_docker` seems to have a bug where it builds some (but not all) image layers with mediaType `application/vnd.oci.image.layer.v1.tar` rather than `application/vnd.docker.image.rootfs.diff.tar.gzip`, even though we're setting `format = "Docker"`. (This only seems to happen on arm64 for some reason, but it could possibly just be because the linux arm64 runners are on ubuntu 22.04 and might have a newer version of dockerd). containerd and podman do not seem to support the `oci` format, or at least they don't support mixing both types of image layers in the same manifest.

I didn't see a straightforward way to customize the mediaType in any of the `rules_docker` rules, so the fix in this PR is to build the image via the `container_image` target but then push the image with `docker tag && docker push` instead of running the `container_push` target. Tested this by running the workflow manually and confirming that the arm64 image no longer has OCI-format layers.

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3316

### DIFF
--- a/deployment/BUILD
+++ b/deployment/BUILD
@@ -1,8 +1,6 @@
 load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
-load("@k8s_deploy//:defaults.bzl", "k8s_deploy")
 load("@io_bazel_rules_k8s//k8s:object.bzl", "k8s_object")
 load("@io_bazel_rules_k8s//k8s:objects.bzl", "k8s_objects")
-load("@io_bazel_rules_docker//container:push.bzl", "container_push")
 load("//rules/flags:index.bzl", "write_flag_to_file")
 
 # TODO(tylerw): break depoloyment.yaml apart.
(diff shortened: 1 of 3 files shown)


iain-macdonald (approved):
Good find! Do you think it'd make sense to try using the rules_oci version of container_push to see if it has the bug as well?

bduffany:
> Good find! Do you think it'd make sense to try using the rules_oci version of container_push to see if it has the bug as well?

If we're going to try rules_oci I think we would want to migrate all of our image rules to rules_oci so we aren't in a messy state of using both rulesets at the same time. It might make sense if the benefits are worth it, but I don't know of any other benefits atm (besides fixing this one bug which we have a reasonable workaround for)

PR #6424, opened 2024-04-23
Add timeout to guest mount/unmount RPCs

bduffany:
Mount/unmount should be relatively fast, so if these RPCs take longer than 1 minute then the vmexec server is likely stuck and the mount/unmount will likely never complete. Return an error in this case rather than allowing the RPC to hang.

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #6439, opened 2024-04-24
Fix release.py running docker images instead of just building

bduffany:
https://github.com/buildbuddy-io/buildbuddy/pull/6416 changed the behavior of `release.py` to build the docker images with `bazel run` but then push them with the `docker` CLI, rather than using the `container_push` targets.

When calling `bazel run` on the BuildBuddy server `go_image` targets though, it actually runs the server image after building it. This differs from `container_image` which just builds and tags the image. This behavior causes `release.py` to get stuck running the server.

To fix this issue, this PR changes the `buildbuddy_image` server targets (OSS and enterprise) to be `container_image` targets rather than `go_image` targets.

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #6440, opened 2024-04-24
Use non-aliased target for image push

bduffany:
When we derive the image tag from the target name (as rules_docker does), we need to use the non-aliased target name since rules_docker doesn't know about our aliasing.

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #6442, opened 2024-04-24
Clean up code flow for gRPC server setup and server registration

iain-macdonald:
no more closures

**Related issues**: N/A

### DIFF
--- a/server/libmain/libmain.go
+++ b/server/libmain/libmain.go
@@ -219,7 +219,7 @@ func GetConfiguredEnvironmentOrDie(healthChecker *healthcheck.HealthChecker, app
 }
 
 func registerInternalGRPCServices(env *real_environment.RealEnv) error {
-	b, err := grpc_server.NewBuilder(env, grpc_server.InternalGRPCPort(), false)
+	b, err := grpc_server.NewBuilder(env, grpc_server.InternalGRPCPort(), false /*=ssl*/)
 	if err != nil {
 		return err
 	}
@@ -229,7 +229,7 @@ func registerInternalGRPCServices(env *real_environment.RealEnv) error {
 	}
 	env.SetInternalGRPCServer(b.GetServer())
 
-	sb, err := grpc_server.NewBuilder(env, grpc_server.InternalGRPCSPort(), true)
+	sb, err := grpc_server.NewBuilder(env, grpc_server.InternalGRPCSPort(), true /*=ssl*/)
 	if err != nil {
 		return err
 	}
@@ -251,7 +251,7 @@ func registerInternalServices(env *real_environment.RealEnv, grpcServer *grpc.Se
 }
 
 func registerGRPCServices(env *real_environment.RealEnv) error {
-	b, err := grpc_server.NewBuilder(env, grpc_server.GRPCPort(), false)
+	b, err := grpc_server.NewBuilder(env, grpc_server.GRPCPort(), false /*=ssl*/)
 	if err != nil {
 		return err
 	}
@@ -262,7 +262,7 @@ func registerGRPCServices(env *real_environment.RealEnv) error {
 	env.SetGRPCServer(b.GetServer())
 	grpc_server.EnableGRPCOverHTTP(env, b.GetServer())
 
-	sb, err := grpc_server.NewBuilder(env, grpc_server.GRPCSPort(), true)
+	sb, err := grpc_server.NewBuilder(env, grpc_server.GRPCSPort(), true /*=ssl*/)
 	if err != nil {
 		return err
 	}
--- a/server/libmain/libmain.go
+++ b/server/libmain/libmain.go
@@ -238,7 +238,7 @@ func registerInternalGRPCServices(env *real_environment.RealEnv) error {
 		if err = sb.Start(); err != nil {
 			return err
 		}
-		env.SetInternalGRPCSServer(b.GetServer())
+		env.SetInternalGRPCSServer(sb.GetServer())
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


server/libmain/libmain.go
```diff
@@ -218,14 +218,65 @@ func GetConfiguredEnvironmentOrDie(healthChecker *healthcheck.HealthChecker, app
...
+	sb, err := grpc_server.NewBuilder(env, grpc_server.InternalGRPCSPort(), true /*=ssl*/)
+	if err != nil {
+		return err
+	}
+	if sb != nil {
+		registerInternalServices(env, sb.GetServer())
+		if err = sb.Start(); err != nil {
+			return err
+		}
+		env.SetInternalGRPCSServer(b.GetServer())
```
tylerwilliams, line 241: should this be sb.GetServer()?

server/libmain/libmain.go
```diff
@@ -218,14 +218,65 @@ func GetConfiguredEnvironmentOrDie(healthChecker *healthcheck.HealthChecker, app
...
+	sb, err := grpc_server.NewBuilder(env, grpc_server.GRPCSPort(), true /*=ssl*/)
+	if err != nil {
+		return err
+	}
+	if sb != nil {
+		registerServices(env, sb.GetServer())
+		if err = sb.Start(); err != nil {
+			return err
+		}
+		env.SetGRPCSServer(b.GetServer())
```
tylerwilliams, line 274: should this be sb.GetGetServer()

server/util/grpc_server/grpc_server.go
```diff
@@ -68,112 +68,81 @@ func MaxRecvMsgSizeBytes() int {
...
+	return b.server
 }
 
-func RegisterInternalGRPCSServer(env *real_environment.RealEnv, regServices RegisterServices) error {
-	if *internalGRPCSPort == 0 {
-		return nil
+func NewBuilder(env environment.Env, port int, ssl bool) (*Builder, error) {
+	b := &Builder{env: env, port: port, ssl: ssl}
+	if ssl && !env.GetSSLService().IsEnabled() {
+		return nil, nil
```
tylerwilliams, line 87: returning (nil, nil) is kind of unusual -- why is that the right thing here?

tylerwilliams (changes requested):

tylerwilliams:
Added @vadimberezniker too so this gets a second set of eyes

server/libmain/libmain.go
iain-macdonald: oof, yeah, fixed

server/libmain/libmain.go
iain-macdonald: oof, yeah, fixed

server/util/grpc_server/grpc_server.go
iain-macdonald: I was going to defend this decision, but then I realized I can just check whether the SSL service is enabled before constructing this server in libmain.go, so I did that instead.

server/util/grpc_server/grpc_server.go
```diff
@@ -68,112 +68,81 @@ func MaxRecvMsgSizeBytes() int {
...
-	env.SetInternalGRPCServer(grpcServer)
-	return nil
+func (b *Builder) GetServer() *grpc.Server {
+	return b.server
 }
 
-func RegisterInternalGRPCSServer(env *real_environment.RealEnv, regServices RegisterServices) error {
-	if *internalGRPCSPort == 0 {
-		return nil
+func NewBuilder(env environment.Env, port int, ssl bool) (*Builder, error) {
```
vadimberezniker, line 84: I was a little thrown off by the naming as this is not a "builder" in the conventional sense. What do you think instead of returning a `*Builder` you return `(*grpc.Server server, func() error, error)` and get rid of the `*Builder` struct altogether? You could also rename `NewBuilder` to `New` so at the call site the call becomes `s, start, err := grpc_server.New(...)`.

server/libmain/libmain.go
```diff
@@ -218,14 +218,65 @@ func GetConfiguredEnvironmentOrDie(healthChecker *healthcheck.HealthChecker, app
 	return realEnv
 }
 
-func registerInternalGRPCServices(grpcServer *grpc.Server, env *real_environment.RealEnv) {
+func registerInternalGRPCServices(env *real_environment.RealEnv) error {
```
vadimberezniker, line 221: I would call these methods something along the lines of `startInternalGRPCServers` rather than `register...` as you are not only registering gRPC services here but also starting the actual gRPC server.

vadimberezniker (changes requested):

server/util/grpc_server/grpc_server.go
iain-macdonald: Yeah, not a great name. In a previous iteration it was more builder-y. I'd prefer to keep the struct because in the future I think there'll be more configuration options added and I think it's a bit more clear that passing callbacks around, so I renamed this to GRPCServer.

server/libmain/libmain.go
iain-macdonald: SG, done

server/util/grpc_server/grpc_server.go
vadimberezniker: Do you have a specific example in mind of what you think will be added to the struct in the future? I'm not a big fan of making speculative changes for something that may not ever come to be. The struct can always be added in the future, as needed if more complexity is required. 

As far as returning a function, it's not that uncommon to do in Go code. I don't see a big difference between asking the user to call Start on the struct or calling the returned start function.

iain-macdonald: I'm going to add a way to add additional RPC interceptors for cache proxy auth.

I know it's not uncommon to return callbacks in go and it's effectively the same thing, but I always find passing functions around more confusing than working with an object. I can change it if you'd really like but part of my motivation with this change was to remove the closures.

vadimberezniker (approved):

tylerwilliams (approved):

PR #6447, opened 2024-04-25
Fetch invocation history from clickhouse, merge in unfinished builds from mysql

jdhollen:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/invocation_search_service/invocation_search_service.go
+++ b/enterprise/server/invocation_search_service/invocation_search_service.go
@@ -417,6 +417,9 @@ func getFilterValue(inv *inpb.Invocation, sortField inpb.InvocationSort_SortFiel
 }
 
 func getExtraFilterForBlendedQuery(firstResult *inpb.Invocation, lastResult *inpb.Invocation, sort *inpb.InvocationSort) *sfpb.StatFilter {
+	if firstResult == nil && lastResult == nil {
+		return nil
+	}
 	f := &sfpb.StatFilter{}
 	switch sort.SortField {
 	case inpb.InvocationSort_UPDATED_AT_USEC_SORT_FIELD:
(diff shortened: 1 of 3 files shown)


jdhollen:
Fully intending to write tests, but hoping to have you all take a look and decide if this is worth it or not.

This only improves performance for queries where the user is either:

1. Not really filtering a ton + using the default sort order (because the range we query from mysql is small), or 
2. Filtering out pending builds (because we won't go to mysql at all)

Short of doing something that lets us skip MySQL entirely, you need some version of this hack.  

If we don't add an index for overall status, then queries for other sort orders will stay slow--they'll require a full scan of all 30 days or whatever the user is looking at (i don't think it's reasonable to add an index for every sort order).

If we add an index and it's still not enough for some orgs, we could make performance better by marking long-disconnected and long-pending builds as "completed" with a janitor task (and flush to clickhouse)--then we could rely the status index to make the MySQL queries for "actually-pending" builds fast, since there shouldn't be *that* many per org at any point in time.

tylerwilliams (approved):

enterprise/server/invocation_search_service/invocation_search_service.go
```diff
@@ -34,7 +37,8 @@ const (
 )
 
 var (
-	olapInvocationSearchEnabled = flag.Bool("app.olap_invocation_search_enabled", true, "If true, InvocationSearchService will query clickhouse for some queries.")
+	blendedInvocationSearchEnabled = flag.Bool("app.blended_invocation_search_enabled", false, "If true, InvocationSearchService will query clickhouse *and* the regular db.")
+	olapInvocationSearchEnabled    = flag.Bool("app.olap_invocation_search_enabled", true, "If true, InvocationSearchService will query clickhouse for some queries.")
```
bduffany, line 41: nit: I couldn't figure out the difference between these 2 flags by reading the help strings, maybe the second flag could use more detail (or just have the 1 "blended" flag?)

bduffany (approved):

enterprise/server/invocation_search_service/invocation_search_service.go
jdhollen: Did what I can here to make this clearer (if wordier)

jdhollen (commented):
Alright, gonna add tests and submit.

Issue #6446, opened 2024-04-25
Feature Request: Show execution log in artifacts tab


brentleyjones:
**Is your feature request related to a problem? Please describe.**

When using the new `--experimental_execution_log_compact_file` flag (and currently `--remote_build_event_upload=all`) a compact execution log is uploaded to the remote cache. It is referenced in the `BuildToolLogs` event under the `"execution.log"` `name`. Currently this file doesn't show up in the Artifacts tab (or anywhere else).

**Describe the solution you'd like**

A link somewhere to easily download this file.

sluongng:
We started to support compact execution log in our CLI with `bb print --compact_execution_log=log.binpb.zstd [--sort=true]`.

I think we could consider providing a dedicated UI for it once it hits a stable status. Right now it's still marked as unstable on the Bazel side with a few bugs reported.

In the meantime, I think we could render `bytestream://` URL in the Raw event tab as clickable. That way users could simply still these files themselves 🤔 

WDYT @siggisim ?

brentleyjones:
> In the meantime, I think we could render bytestream:// URL in the Raw event tab as clickable. That way users could simply still these files themselves

I would like this as well, since what I did was manually take a URL from the artifacts page and replaced the digest with the one I cared about from the raw event.

PR #6449, opened 2024-04-25
Add some basic rendering for the execution log in the UI

siggisim:
Renders the compact execution log on the `Executions` tab if the following Bazel flags are set:
```
--experimental_execution_log_compact_file
--remote_build_event_upload=all
```

Also adds a download button to the card title so users can easily download the file for their own analysis.

Kept the rendering basic to start. Currently renders most of the spawn details we get in a list, and each links to its corresponding action details card.

Added filtering by mnemonic and runner type, sorting by total duration asc / desc, and text filtering by target label, args, and digest.

I explicitly didn't touch any input / output file rendering / diffing / linking / graph creation - even though I think that's probably the most valuable thing we can do with this information. Just want to get something simple in to start.

Fixes https://github.com/buildbuddy-io/buildbuddy/issues/6446

cc @brentleyjones

### DIFF
--- a/app/invocation/invocation_exec_log_card.tsx
+++ b/app/invocation/invocation_exec_log_card.tsx
@@ -214,7 +214,7 @@ export default class InvocationExecLogCardComponent extends React.Component<Prop
           !l.spawn?.args.join(" ").toLowerCase().includes(this.props.filter.toLowerCase()) &&
           !l.spawn?.digest?.hash.toLowerCase().includes(this.props.filter.toLowerCase())
         ) {
-          return;
+          return false;
         }
 
         return true;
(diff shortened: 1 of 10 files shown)


app/invocation/invocation_exec_log_card.tsx
```diff
@@ -0,0 +1,321 @@
...
+          return false;
+        }
+        if (
+          this.props.filter != "" &&
+          !l.spawn?.targetLabel.toLowerCase().includes(this.props.filter.toLowerCase()) &&
+          !l.spawn?.mnemonic.toLowerCase().includes(this.props.filter.toLowerCase()) &&
+          !l.spawn?.args.join(" ").toLowerCase().includes(this.props.filter.toLowerCase()) &&
+          !l.spawn?.digest?.hash.toLowerCase().includes(this.props.filter.toLowerCase())
+        ) {
+          return;
```
jdhollen, line 217: surprised this didn't whine at you, but i guess this should be `return false`, too

jdhollen (approved):

app/invocation/invocation_exec_log_card.tsx
siggisim: Good catch, done!

PR #6455, opened 2024-04-26
Add goroutine to clean up orphan scorecards

tempoz:
- **set up channel to track scorecard/counter expiries**
- **added expiryList, fixed imports**
- **now using lookuplist**
- **buildfix**

Adds `HitTrackerManager`, which starts a goroutine that listens on a channel for invocation ids. When it receives an id, it assigns it an expiration time and tacks it onto the end of its expiryList. If that id already exists in the list, the old value is removed. It wakes up when the d at the front of the list reaches its expiration time and then pops it off and checks for a related invocation in the invocationdb. If none is found, it cleans up any old scorecards and counters and then waits for the next expiration time or message.

**Related issues**: N/A


server/remote_cache/hit_tracker/hit_tracker.go
```diff
@@ -31,6 +35,7 @@ import (
 
 var (
 	detailedStatsEnabled = flag.Bool("cache.detailed_stats_enabled", false, "Whether to enable detailed stats recording for all cache requests.")
+	orphanScorecardTTL   = flag.Duration("cache.orphan_scorecard_ttl", time.Minute*30, "How long to hold on to scorecards after the last cache request if there is no related entry in the invocation database.")
```
bduffany, line 38: would it be equivalent to do something like this?

* When inserting cache scorecard entries into redis, set the expiration to `max(current_expiration, 30 minutes)`
* When receiving the first event in the build event handler, set the expiration to something longer (i.e. whatever it is now)

This way we wouldn't need to keep local state on the server (i.e. the lookup list) or do any DB lookups

bduffany: err I guess the issue is that redis is not actively evicting keys that have expired, so this wouldn't work. But I still wonder if there is a simpler option that only stores data in Redis rather than keeping local state in the apps as well as doing DB lookups, wdyt?

tempoz: > which might not work if the server restarts (?) and also avoids DB lookups

It absolutely won't, at which point they'll get kicked out by the TTL. I talked to Tyler and he said that was fine, since this is just to reduce cache usage and there's nothing critical about making sure these get removed.

> But I still wonder if there is an approach that only involves reading/writing some data in Redis rather than keeping local state in the apps as well as doing DB lookups, wdyt?

This should only be one database read per invocation, and I could reduce it further by doing a cache lookup first to see if the data even needs to be cleaned up (since invocations that use us as more than just a cache won't need cleaned up). This feels like a pretty light touch to me already, and aside from adding the data structure to `util`, it's fairly simple, no?

Other options I can think of:

- Have invocations leave a marker in redis that a particular invocation id is "owned" by an event stream, and thus won't need its state cleaned out. (no DB lookups, but we still use a lookup list locally)

- Implement a lookup list in redis and then have apps periodically check it for things that need to be cleaned out (no local state, still hit the database; disadvantage is that this is, I think, quite a bit more complex)

- Do both

Thoughts?

bduffany: > at which point they'll get kicked out by the TTL. I talked to Tyler and he said that was fine, since this is just to reduce cache usage and there's nothing critical about making sure these get removed.

hmm for some reason I thought the issue we were seeing was that Redis TTL eviction was not being aggressive enough :thinking: I might be totally misremembering though and maybe eviction is working fine, and the issue was just that we have too many entries that are waiting to be evicted. Let's assume for the sake of discussion that Redis TTL is working fine.

> This should only be one database read per invocation

In the current implementation I think since each app has its own HitTrackerManager, this means we are doing up to `N` DB lookups per invocation where `N` is the number of apps deployed (assuming the invocation does a number of cache requests significantly greater than N)? My worry is that scaling up our app deployment would increase load on the DB which doesn't seem great - and also N is high enough today where the added load seems like it'd be significant?

> Other options [...] Thoughts?

I think a combination of both of those approaches sounds good. Redis' sorted set data structure might be a useful tool here, if each set item is ranked by its expiration time. We could maybe have one sorted set which stores the cache-based expiration time for each invocation ID (the rank would be set to +30min on each cache request) and another sorted set which stores an "override" expiration time for each invocation ID, which gets set to +24h on the first BES event, or something like that. The one thing I'd mention is that for the cache-based redis requests, we would want to batch those up using a pipeline to avoid excessive load on Redis. Maybe the hit tracker manager could be the thing that does that batching

PR #6461, opened 2024-04-29
Fix VM network clone IP calculation and consolidate cleanup

bduffany:
Instead of using `vmIdx` for the clone IP, we want to use the `netIdx` which we have locked. The `netIdx` is different from the `vmIdx` in the case where the `vmIdx` is already in use.

This also restructures the cleanup work that needs to be done after calling `SetupVethPair`. Because the clone IP depends on the `netIdx` and not the `vmIdx`, some of the current cleanup work that is done in `firecracker.go` would now need to know about the `netIdx`. But this feels a bit like implementation details leaking out. So, in order to avoid exposing the `netIdx` and having to plumb it around, this PR makes it so that the cleanup func already returned by `SetupVethPair` now also does this cleanup work. This makes the networking API a bit simpler since more of the cleanup work is done automatically.

**Related issues**: N/A

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -10,7 +10,10 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"time"
 
+	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
+	"github.com/buildbuddy-io/buildbuddy/server/util/background"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/random"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
@@ -36,6 +39,11 @@ const (
 	netNamespacePrefix = "bb-executor-"
 	// Total number of available host IP ranges that can be allocated to VMs.
 	numAssignableNetworks = 1000
+	// Minimum time to allow for networking cleanup. We intentionally use a
+	// long-ish timeout here because if cleanup fails then we might leave the
+	// network in a bad state, preventing a host IP from being
+	// usable.
+	networkingCleanupTimeout = 1 * time.Minute
 )
 
 // runCommand runs the provided command, prepending sudo if the calling user is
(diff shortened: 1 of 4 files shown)


server/util/networking/networking.go
```diff
@@ -36,6 +39,11 @@ const (
 	netNamespacePrefix = "bb-executor-"
 	// Total number of available host IP ranges that can be allocated to VMs.
 	numAssignableNetworks = 1000
+	// Minimum time to allow for networking cleanup. We intentionally use a
```
maggie-lou, line 42: * Maximum

server/util/networking/networking.go
```diff
@@ -346,6 +336,35 @@ func (a *HostNetAllocator) unlock(netIdx int) {
 //	# add a route in the root namespace so that traffic to 192.168.0.3 hits 10.0.0.2, the veth0 end of the pair
 //	$ sudo ip route add 192.168.0.3 via 10.0.0.2
 func SetupVethPair(ctx context.Context, netNamespace, vmIP string, vmIdx int) (_ func(context.Context) error, err error) {
+	// Keep a list of cleanup work to be done. For each resource we create in
+	// this func that needs to be explicitly cleaned up, we append a cleanup
+	// task to this list. Cleanup work is done in the reverse order in which
+	// it's added to this list (i.e., it's a stack).
+	var cleanupStack []func(ctx context.Context) error
```
maggie-lou, line 343: I find this logic a bit complex to follow. For example it takes some mental work to figure out the order the cleanup tasks are executing (i.e. "if I want to cleanup the route before I unlock the net idx, when should I add it to the slice?") I would personally find it a bit more readable to use booleans

Ex. 
if shouldCleanupRoute {}
if shouldCleanupRule {}

server/util/networking/networking.go
```diff
@@ -346,6 +336,35 @@ func (a *HostNetAllocator) unlock(netIdx int) {
...
+				alert.UnexpectedEvent("vm_network_cleanup_failed", "VM networking cleanup failed. This might keep the VM host IP locked, so if there are a lot of these errors, VM networking may eventually stop functioning. Error: %s", err)
+				return err
+			}
+		}
+		return nil
+	}
+	// If we return an error from this func then we need to clean up any
+	// resources that were created before returning.
+	defer func() {
+		if err != nil {
```
maggie-lou, line 363: When `err` is shadowed - i.e. `hostEndpointNet, err := hostNetAllocator.Get(vmIdx)` - does this still work correctly?

bduffany: Yeah - in that example, `err` is not being declared (shadowed), it's being re-assigned. The declaration happens in the named return value.

I was on the fence about whether to do this vs using a wrapper func but I think ultimately the wrapper func is more error prone because we have to remember to `return cleanup, err` instead of `return nil, err` everywhere in this func.

maggie-lou: Ah okay. I wasn't sure if a new `err` variable was being created, but it seems like it won't, and only the value will get reassigned (https://go.dev/ref/spec#Short_variable_declarations)

maggie-lou (approved):

PR #6465, opened 2024-04-30
timing_card: get profile from structured flag

sluongng:
Currently we are assuming that the BuildToolLogs event will have
relatively small amount of files that are uploaded via bytestream API.
Using that assumption, we guessed that the build timing profile must be
one of the 2 files referenced in this event, and differentiate them
based on their file name.

Detect the actual timing profile name by searching through canonical
command line for a `--profile=` flag. If the user does not specify a
custom profile name, assume the default `command.profile.gz` name.

### DIFF
--- a/app/invocation/invocation_timing_card.tsx
+++ b/app/invocation/invocation_timing_card.tsx
@@ -74,11 +74,10 @@ export default class InvocationTimingCardComponent extends React.Component<Props
   }
 
   getProfileFile(): build_event_stream.File | undefined {
-    const profileOption = this.props.model
-      .structuredCommandLine!.find((scl) => scl.commandLineLabel == "canonical")!
-      .sections.find((s) => s.sectionLabel == "command options")!
-      .optionList!.option.find((o) => o.optionName == "profile");
-    const profileName = profileOption ? profileOption.optionValue : "command.profile.gz";
+    const profileName = this.props.model
+      .structuredCommandLine?.find((scl) => scl.commandLineLabel == "canonical")?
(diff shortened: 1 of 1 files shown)


jdhollen (approved):
I wrote this exact change (more or less) at some point, no clue why I never sent it (execution.log isn't the only other build tool log obviously, so the logic at HEAD breaks if you use any of the other output-generating flags)

app/invocation/invocation_timing_card.tsx
```diff
@@ -74,8 +74,14 @@ export default class InvocationTimingCardComponent extends React.Component<Props
   }
 
   getProfileFile(): build_event_stream.File | undefined {
+    const profileOption = this.props.model
+      .structuredCommandLine!.find((scl) => scl.commandLineLabel == "canonical")!
+      .sections.find((s) => s.sectionLabel == "command options")!
+      .optionList!.option.find((o) => o.optionName == "profile");
+    const profileName = profileOption ? profileOption.optionValue : "command.profile.gz";
```
bduffany, line 81: nit: avoid using non-null assertions

```suggestion
    const profileName = this.props.model
      .structuredCommandLine?.find((scl) => scl.commandLineLabel == "canonical")?
      .sections?.find((s) => s.sectionLabel == "command options")?
      .optionList?.option?.find((o) => o.optionName == "profile")?.optionValue ?? "command.profile.gz";
```

sluongng: Ohhh.. I do like this!

PR #6467, opened 2024-04-30
exec_log_card: download execution log with ext

sluongng:
Let users download the compact execution log with the correct file
extension. This way users could immediately tell how to consume the
downloaded file.

Also support the new file name that was proposed in (1).

(1): https://github.com/bazelbuild/bazel/pull/22172

### DIFF
--- a/app/invocation/invocation_exec_log_card.tsx
+++ b/app/invocation/invocation_exec_log_card.tsx
@@ -63,7 +63,9 @@ export default class InvocationExecLogCardComponent extends React.Component<Prop
   getExecutionLogFile(): build_event_stream.File | undefined {
     return this.props.model.buildToolLogs?.log.find(
       (log: build_event_stream.File) =>
-        log.name == "execution.log" && log.uri && Boolean(log.uri.startsWith("bytestream://"))
+        (log.name == "execution.log" || log.name == "execution_log.binpb.zstd") &&
+        log.uri &&
+        Boolean(log.uri.startsWith("bytestream://"))
     );
   }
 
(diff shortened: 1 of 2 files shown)


app/invocation/invocation_exec_log_card.tsx
```diff
@@ -112,7 +114,11 @@ export default class InvocationExecLogCardComponent extends React.Component<Prop
     }
 
     try {
-      rpcService.downloadBytestreamFile("execution.log", profileFile.uri, this.props.model.getInvocationId());
+      rpcService.downloadBytestreamFile(
+        "execution_log.binpb.zstd",
```
jdhollen, line 118: doesn't this depend on which of the execution_log flags the user chose to use (json, binary, compact?)

sluongng: Currently only the compact format will get published via BES
```
sluongng/exec-log-extension ~/work/bazelbuild/bazel> rg shouldPublish -A3 src/main/java/com/google/devtools/build/lib/exec/
src/main/java/com/google/devtools/build/lib/exec/CompactSpawnLogContext.java
190:  public boolean shouldPublish() {
191-    // The compact log is small enough to be uploaded to a remote store.
192-    return true;
193-  }

src/main/java/com/google/devtools/build/lib/exec/ExpandedSpawnLogContext.java
138:  public boolean shouldPublish() {
139-    // The expanded log tends to be too large to be uploaded to a remote store.
140-    return false;
141-  }

src/main/java/com/google/devtools/build/lib/exec/SpawnLogContext.java
78:  public abstract boolean shouldPublish();
79-
80-  /** Computes the environment variables. */
81-  protected ImmutableList<EnvironmentVariable> getEnvironmentVariables(Spawn spawn) {
```

The other formats are usually quite chunky in size. (could goes up to hundred of gigs each build in some setup)

jdhollen (approved):

PR #6470, opened 2024-04-30
workflows: allow customizing git fetch

bduffany:
* Instead of `git clone`, do a `git init` which doesn't fetch anything.
* Allow customizing the `git fetch` via two new workflow config options `git_fetch_filters` and `git_fetch_depth`. `git_fetch_filters` defaults to `["blob:none"]` to match what we have today, and `git_fetch_depth` defaults to `0` (equivalent to not being set).

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -1648,11 +1648,6 @@ func (ws *workspace) sync(ctx context.Context) error {
 		return err
 	}
 	merge := action.GetTriggers().GetPullRequestTrigger().GetMergeWithBase()
-	if *gitFetchDepth > 0 {
-		// If not fetching the full commit history, then we might not have the
-		// merge-base commit available, so we can't reliably merge.
-		merge = false
-	}
 	// If enabled by the config, merge the target branch (if different from the
 	// pushed branch) so that the workflow can pick up any changes not yet
 	// incorporated into the pushed branch.
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -972,7 +972,6 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 		artifactsDir := artifactsPathForCommand(ws, i)
 		namedSetID := filepath.Base(artifactsDir)
 
-		// Note: we don't use ar.action.Env here because it's already
 		runErr := runCommand(ctx, *bazelCommand, expandEnv(args), nil, action.BazelWorkspaceDir, ar.reporter)
 		exitCode := getExitCode(runErr)
 		ar.reporter.Publish(&bespb.BuildEvent{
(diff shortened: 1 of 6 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -970,7 +972,8 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 		artifactsDir := artifactsPathForCommand(ws, i)
 		namedSetID := filepath.Base(artifactsDir)
 
-		runErr := runCommand(ctx, *bazelCommand, expandEnv(args), action.Env, action.BazelWorkspaceDir, ar.reporter)
+		// Note: we don't use ar.action.Env here because it's already
```
vadimberezniker, line 975: "it's already" what?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1646,6 +1649,11 @@ func (ws *workspace) sync(ctx context.Context) error {
 		return err
 	}
 	merge := action.GetTriggers().GetPullRequestTrigger().GetMergeWithBase()
+	if *gitFetchDepth > 0 {
+		// If not fetching the full commit history, then we might not have the
+		// merge-base commit available, so we can't reliably merge.
+		merge = false
```
vadimberezniker, line 1655: Instead of silently turning off merging, is it possible to error out here?

vadimberezniker (approved):

enterprise/server/cmd/ci_runner/main.go
bduffany: lol, not sure how this creeped into this PR. Removed

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -151,6 +151,8 @@ var (
 	patchURIs          = flag.Slice("patch_uri", []string{}, "URIs of patches to apply to the repo after checkout. Can be specified multiple times to apply multiple patches.")
 	recordRunMetadata  = flag.Bool("record_run_metadata", false, "Instead of running a target, extract metadata about it and report it in the build event stream.")
 	gitCleanExclude    = flag.Slice("git_clean_exclude", []string{}, "Directories to exclude from `git clean` while setting up the repo.")
+	gitFetchFilters    = flag.Slice("git_fetch_filters", []string{}, "Filters to apply to `git fetch` commands.")
```
sluongng, line 154: Nit: I think we could get away with supporting a single filter currently thanks to https://git-scm.com/docs/git-rev-list#Documentation/git-rev-list.txt---filterltfilter-specgt

Then perhaps if there is a valid use case, we could add support for multiple filters in the future.

sluongng (approved):
1 small nit, feel free to ignore.

LGTM.

enterprise/server/cmd/ci_runner/main.go
bduffany: I considered this, but using multiple filters with the `combine` syntax requires %-encoding each of the individual filters which doesn't feel very user friendly to me. Using a slice here doesn't add much complexity so I think it's fine to start with a slice

PR #6475, opened 2024-04-30
Refactor git logic in ci runner for readability

maggie-lou:
Pull logic into helper functions for more readability, and simplify the logic a bit (Note that I didn't directly copy and paste the code into helper functions - I changed it slightly)

**Related issues**: N/A


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -154,6 +147,15 @@ var (
 	gitFetchFilters    = flag.Slice("git_fetch_filters", []string{}, "Filters to apply to `git fetch` commands.")
 	gitFetchDepth      = flag.Int("git_fetch_depth", 0, "Depth to use for `git fetch` commands.")
 
+	// Flags to configure setting up git repo
```
bduffany, line 150: nit: should the `git_*` flags and `patch_uri` flag also be moved to this section? since those also affect the git repo state

enterprise/server/test/integration/ci_runner/ci_runner_test.go
```diff
@@ -748,7 +748,7 @@ actions:
 	assert.Contains(t, result.Output, "args: {{ Switcheroo! }}")
 }
 
-func TestRunAction_TargetRepoOnly(t *testing.T) {
+func TestRunAction_PushedRepoOnly(t *testing.T) {
```
bduffany, line 751: The workflow service currently sets both pushed and target on push events, so we should probably set both flags here to make sure we're testing the actual usage?

(If the runner service only sets pushed repo URL, then we could make a copy of this test to make sure it works in both cases)

```suggestion
func TestRunAction_PushEvent(t *testing.T) {
```

enterprise/server/test/integration/ci_runner/ci_runner_test.go
```diff
@@ -770,8 +770,8 @@ func TestRunAction_TargetRepoOnly(t *testing.T) {
 		"--workflow_id=test-workflow",
 		"--action_name=Print args",
 		"--trigger_event=push",
-		"--target_repo_url=file://" + repoPath,
-		"--target_branch=master",
+		"--pushed_repo_url=file://" + repoPath,
+		"--pushed_branch=master",
```
bduffany, line 774: (following the previous comment)

```suggestion
		"--target_repo_url=file://" + repoPath,
		"--target_branch=master",
		"--pushed_repo_url=file://" + repoPath,
		"--pushed_branch=master",
```

bduffany (approved):

enterprise/server/test/integration/ci_runner/ci_runner_test.go
maggie-lou: Added a new test

PR #6481, opened 2024-05-01
Add NHID to replica descriptor

luluz66:
**Related issues**: N/A
Store the nhid in ReplicaDesciptor. This enables us to tell which store has
a particular range by looking at the rangeDescriptor which contains a list of 
ReplicaDescriptors.


tylerwilliams (approved):
LGTM

maybe even less invasive would be to add a NHID() method to IStore in replica.go and then pull out the nhid in replica.New()

PR #6482, opened 2024-05-01
Add service for generating short-lived user certificates.

vadimberezniker:
The initial use case is for SSH certificates.

The user will present an OIDC token and a public key. After validating the token, the server will sign the public key and return the signed key back to the user.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/certgenerator/certgenerator.go
+++ b/enterprise/server/certgenerator/certgenerator.go
@@ -1,0 +1,170 @@
+package main
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/auth"
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/configsecrets"
+	"github.com/buildbuddy-io/buildbuddy/server/config"
+	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
+	"github.com/buildbuddy-io/buildbuddy/server/ssl"
+	"github.com/buildbuddy-io/buildbuddy/server/util/flag"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_server"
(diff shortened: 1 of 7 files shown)


enterprise/server/certgenerator/certgenerator.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, err
+	}
+	g := &generator{
+		verifier:  provider.Verifier(&oidc.Config{ClientID: *tokenClientID}),
+		sshSigner: s,
+	}
+	return g, nil
+}
+
+func main() {
```
bduffany, line 120: I think it's probably worth adding a basic test with a valid token and invalid token, if it's not too difficult (could maybe use selfauth as the oidc provider?)

bduffany (approved):
LGTM (but would be good to have a test in place before deploying to prod)

proto/certgenerator.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message GenerateResponse {
+  // Signed SSH certificate, in OpenSSH format.
+  string ssh_cert = 1;
+}
+
+// Services responsible for generating short-lived user certificates.
+service CertGenerator {
+  rpc Generate(GenerateRequest) returns (GenerateResponse);
+}
```
tylerwilliams, line 20: newline!

server/http/interceptors/interceptors.go
```diff
@@ -273,18 +273,18 @@ func (w *instrumentedResponseWriter) Flush() {
 	}
 }
 
-func alertOnPanic() {
+func alertOnPanic(err any) {
```
tylerwilliams, line 276: 👀

tylerwilliams (approved):

enterprise/server/certgenerator/certgenerator.go
vadimberezniker: Didn't seem that straightforward, but I'll take another look in a bit.

server/http/interceptors/interceptors.go
vadimberezniker: Hope this was self-explanatory. We were previously not logging the panic message.

PR #6485, opened 2024-05-01
Add command line tool for obtaining a SSH cert.

vadimberezniker:
Obtains an OIDC token from gcloud, sends it to the remote server along with the SSH public key to be signed.

Once the cert is obtained, it's written to a location that ssh expects to find it.

**Related issues**: N/A

### DIFF
--- a/tools/bbcert/bbcert.go
+++ b/tools/bbcert/bbcert.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"os/exec"
+	"path"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/flag"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+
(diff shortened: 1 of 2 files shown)


bduffany (approved):

tools/bbcert/bbcert.go
```diff
@@ -0,0 +1,80 @@
...
+
+var (
+	server = flag.String("server", "", "gRPC target for the certificate server")
+)
+
+const (
+	sshDir = ".ssh"
+	// For now, assume user is using an RSA key.
+	defaultRSAPublicKeyFile   = "id_rsa.pub"
+	defaultRSACertificateFile = "id_rsa-cert.pub"
```
tylerwilliams, line 25: might be nice to put something in this filename that indicates it's ours. Would help folks distinguish this key from other keys they might have

vadimberezniker: The certificate is tied to the user's key pair and it has to be named <key>-cert.pub for ssh to be able to find it. 
For convenience, I'm using the key that the user should already have on their system.

If we want to name it something else, we'd also have to have the tool generate a separate key pair. 
We could do that but I don't see a real benefit in doing so.

tylerwilliams (approved):

PR #6487, opened 2024-05-01
Run a kythe workflow if codesearch is enabled

tylerwilliams:
If enabled, automatically insert a custom workflow which will effectively run:
```bash
blaze build --override_repository kythe_release=$KYTHE_DIR --bazelrc=$KYTHE_DIR/extractors.bazelrc //...
```
only when the default branch is Pushed.

### DIFF
--- a/enterprise/server/workflow/config/config.go
+++ b/enterprise/server/workflow/config/config.go
@@ -18,6 +18,10 @@ const (
 	// FilePath is the path where we can expect to locate the BuildBuddyConfig
 	// YAML contents, relative to the repository root.
 	FilePath = "buildbuddy.yaml"
+
+	// KytheActionName is the name used for actions automatically
+	// run by us if code search is enabled.
+	KytheActionName = "AutoKythe"
 )
 
 type BuildBuddyConfig struct {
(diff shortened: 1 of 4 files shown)


enterprise/server/workflow/config/config.go
```diff
@@ -18,6 +18,10 @@ const (
 	// FilePath is the path where we can expect to locate the BuildBuddyConfig
 	// YAML contents, relative to the repository root.
 	FilePath = "buildbuddy.yaml"
+
+	// KytheActionName is the name used for actions automatically
+	// run by us if code search is enabled.
+	KytheActionName = "AutoKythe"
```
bduffany, line 24: this will show up in invocation history - idk if you wanted to call it something more obvious to users like "Build source code indexes" or something

(yeah, kind of ugly that the UI label doubles as the action's "ID", but ... :shrug: )

enterprise/server/workflow/config/config.go
```diff
@@ -144,6 +148,21 @@ func NewConfig(r io.Reader) (*BuildBuddyConfig, error) {
...
+	if targetRepoDefaultBranch != "" {
+		pushTriggerBranches = append(pushTriggerBranches, targetRepoDefaultBranch)
+	}
+	return &Action{
+		Name: KytheActionName,
+		Triggers: &Triggers{
+			Push: &PushTrigger{Branches: pushTriggerBranches},
+		},
+		// Note: default Bazel flags are written by the runner to ~/.bazelrc
+		BazelCommands: []string{"build --override_repository kythe_release=$KYTHE_DIR //..."},
```
bduffany, line 162: do you want to add --remote_cache here too? (bes_backend is enabled by default via our generated bazelrc, but remote_cache isn't)

bduffany (approved):

enterprise/server/workflow/config/config.go
tylerwilliams: Done.

enterprise/server/workflow/config/config.go
tylerwilliams: Done.

bduffany (approved):

PR #6492, opened 2024-05-02
Use cache compression; allocate 100G for kythe stuff

tylerwilliams:


enterprise/server/workflow/config/config.go
```diff
@@ -165,7 +165,9 @@ func KytheIndexingAction(targetRepoDefaultBranch string) *Action {
 			Push: &PushTrigger{Branches: pushTriggerBranches},
 		},
 		ContainerImage: `ubuntu-20.04`,
-		// Note: default Bazel flags are written by the runner to ~/.bazelrc
+		ResourceRequests: ResourceRequests{
+			Disk: "100GB",
+		},
```
bduffany, line 170: wdyt about also doubling the CPU request, at least while we're testing this out (to get a little faster iteration time?) the builds seem pretty CPU-constrained

```suggestion
		ResourceRequests: ResourceRequests{
			CPU: "6",
			Disk: "100GB",
		},
```

bduffany (approved):

PR #6496, opened 2024-05-02
hacked up webworker support for monaco.

jdhollen:
**Related issues**: N/A


enterprise/app/BUILD.bazel
```diff
@@ -105,6 +108,14 @@ ts_library(
     ],
 )
 
+js_library(
```
bduffany, line 111: not a big deal since it's such a small amount of code but does this have to be JS or could we use TS for this too?

bduffany (approved):

enterprise/app/BUILD.bazel
jdhollen: heh, wasn't planning on mailing this just yet, but I did this because of typescript whining about the import from monaco (maybe there's a types package i missed?)

PR #6497, opened 2024-05-03
timing_card: roll forward profile detection

sluongng:
- **Reapply "timing_card: get profile from structured flag" (#6495)**
- **timing_card: fix filename detection**


bduffany (approved):

jdhollen (approved):
Hah, that's a little simpler

PR #6499, opened 2024-05-03
Fix some broken links in platforms docs

bduffany:
Confusingly, using `./name` as the link target will refer to the subpath under the current URL, instead of under the current directory.

**Related issues**: N/A


sluongng (approved):
Thanks for fixing this.

PR #6503, opened 2024-05-04
action_card: make timeline offset base on start time

sluongng:
Make the action timeline look more like a trace view.

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -335,6 +335,7 @@ export default class InvocationActionCardComponent extends React.Component<Props
     if (filteredEvents.length == 0) {
       return null;
     }
+    const startTimestamp = filteredEvents[0].timestamp;
 
     const totalDuration = durationSeconds(
       filteredEvents[0].timestamp!,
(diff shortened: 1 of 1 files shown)


sluongng:
image

CSS is hard

sluongng:
image

Ok CSS is still hard.

bduffany:
Having to hover to get the timing info feels like a step backward in terms of usability. I also feel like the bars look a bit awkward because they are sort of floating in space, whereas the current gray-ish background provides a nice visual guide.

wdyt about keeping the current chart design but just sliding over the filled regions to correspond with their start time, so that it looks more like a timeline, instead of having them left-aligned like they are today?

sluongng:
I think you are right. There is a downgrade of glanceability now that I have the time to step away and come back to look at this. 🤔 

Let me try the old design with the bars shifted like you suggested.

sluongng:
image

Much simpler change as well 😆

sluongng:
@bduffany what do you think about adding the total duration of the action to the timeline though? We don't have that information from the page, so the user would have to add up the phases manually, or back out to the "Executions" tab to see the total duration there.

bduffany:
> what do you think about adding the total duration of the action to the timeline

sgtm

bduffany (approved):

app/invocation/invocation_action_card.tsx
```diff
@@ -335,41 +335,61 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+            style={{ flex: `1 0 0`, backgroundColor: "green" }}></div>
+        </div>
         {filteredEvents.map((event, i) => {
           // Don't render the end marker.
           if (i == filteredEvents.length - 1) return null;
 
           const next = filteredEvents[i + 1];
           const duration = durationSeconds(event.timestamp!, next.timestamp!);
           const weight = duration / totalDuration;
+          offset += weight;
```
bduffany, line 368: sorry I realized I forgot to double-check this math :face_with_spiral_eyes: should the `offset += weight` be happening after we render the element? seems to me like the rough pseudo-code should be

```
weight = duration / totalDuration;
bar1Weight = offset;
bar2Weight = weight;
bar3Weight = 1 - weight - offset;
// Offset should apply to *next* bar that is rendered
offset += weight;
// now render bars with bar1Weight, bar2Weight, bar3Weight
```

sluongng: Well spot, I thought of adding it after as well, but that would create a larger diff.

So right now I simply add `weight` into `offset` before the loop. Then inside the loop, I subtract out the `weight` value from each bar, then simplified the calculation. Which resulted in the same value.

This is what it currently looks like on my end 
image

PR #6506, opened 2024-05-06
Fix non-hermeticity with rules_nodejs' directory inputs

fmeum:
Fixes warnings such as:
```
WARNING: /private/var/tmp/_bazel_fmeum/412888b82b4f18156bc415025cb8faa1/external/npm_fast-glob-3.2.7/BUILD.bazel:6:10: input 'extract_tmp/package' to @@npm_fast-glob-3.2.7//:_npm_fast-glob-3.2.7 is a directory; dependency checking of directories is unsound
```

**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -261,6 +261,9 @@ common --combined_report=lcov
 
 # Use BLAKE3 digest function.
 startup --digest_function=BLAKE3
+# Bazel doesn't track the contents of source directories by default, which can result
+# in non-hermeticity.
+startup --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1
 
 # Include target names in timing profile so it's clickable.
 common --experimental_profile_include_target_label


sluongng:
We are not using rules_js, so im not quite sure what's the scope of impact here?

fmeum:
> We are not using rules_js, so im not quite sure what's the scope of impact here?

Sorry, I meant rules_nodejs. I updated the PR title and description with an example.

sluongng (approved):
Do you think we could/should move this jvm property to a Bazel startup/common flag? I think https://cs.opensource.google/bazel/bazel/+/c64421bc35214f0414e4f4226cc953e8c55fa0d2 might have overdue on the "short-live" promise.

LGTM

fmeum:
We should get rid of this flag and enable it by default :-) But anything touching this flag is viewed with a lot of (rightful) skepticism by Bazel folks, so we would need to do this right.

PR #6507, opened 2024-05-06
Fix race condition in TaskLeaser.Close

bduffany:
**Related issues**: N/A


vadimberezniker (approved):

bduffany:
hmm I think the test failures here are legit as they don't seem to be happening on the master branch. Looking into it

bduffany:
nvm, there is actually no race condition because we are grabbing a mutex both in pingServer and in Close.

PR #6510, opened 2024-05-06
Remove GH actions test workflow

bduffany:
This workflow doesn't seem super useful now that the `Test` BB workflow is the one that's required for merging.

**Related issues**: N/A


siggisim:
I kinda like having these so we can compare durations and statuses with the workflow ones.

PR #6513, opened 2024-05-07
Use redis-style expire commands

tempoz:
Description pending

**Related issues**: N/A


tempoz:
Too complicated for what it gains us; closing.

PR #6514, opened 2024-05-07
Avoid sending task keepalive ping after stream is closed

bduffany:
This is a follow-up from https://github.com/buildbuddy-io/buildbuddy/pull/6507 which was closed.

There was no race condition in that original PR because we were holding a mutex before sending anything on the stream. However, there is a separate issue where `pingServer` could potentially try to send a message on the stream after `Close()` was called. In practice, this likely just adds noise to logs, but I figured it would be nice to fix.

Also fix a small bug that the short-circuiting check in `Close()` was not actually short-circuiting, though in practice we never actually hit this codepath because we only Close once.

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #6515, opened 2024-05-07
Implement placement driver to add and replace dead replicas

luluz66:
Implements Driver Queue that handles cases to up-replicate and replace dead replicas.
Implements a StoreMap, which is used by the DriverQueue to figure out which store is alive or dead. 

Tested locally for the following scenarios:
- (replaceDeadReplica) run 4 raft nodes; kill one of the app that has one of the replicas; and see that eventually after the deadTimeout, the replica was added to another app.
- (addReplica) change the min replicas for a replica from 3 to 4; and see ranges got
  added to the 4th app.

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #6519, opened 2024-05-07
[ci_runner] Support only passing a commit sha and not a branch

maggie-lou:
Also support not passing a target_repo_url

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -813,12 +813,13 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 	if *prNumber != 0 {
 		buildMetadata.Metadata["PULL_REQUEST_NUMBER"] = fmt.Sprintf("%d", *prNumber)
 	}
-	if *targetRepoURL != "" {
-		buildMetadata.Metadata["REPO_URL"] = *targetRepoURL
-	}
-	if *pushedRepoURL != *targetRepoURL {
+	baseURL := *pushedRepoURL
+	isFork := *targetRepoURL != "" && *pushedRepoURL != *targetRepoURL
+	if isFork {
+		baseURL = *targetRepoURL
 		buildMetadata.Metadata["FORK_REPO_URL"] = *pushedRepoURL
 	}
+	buildMetadata.Metadata["REPO_URL"] = baseURL
 	if *visibility != "" {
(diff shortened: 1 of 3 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1699,27 +1699,38 @@ func (ws *workspace) fetchRefs(ctx context.Context) error {
...
-	isPushedBranchInFork := *targetRepoURL != "" && *pushedRepoURL != *targetRepoURL
-	if isPushedBranchInFork {
-		forkBranches = append(forkBranches, *pushedBranch)
-	} else if *pushedBranch != *targetBranch {
-		baseRefs = append(baseRefs, *pushedBranch)
+	isPushedRefInFork := *targetRepoURL != "" && *pushedRepoURL != *targetRepoURL
+	if isPushedRefInFork {
+		// Try to fetch a branch if possible, because fetching
+		// commits that aren't at the tip of a branch requires
+		// users to set an additional config option (uploadpack.allowAnySHA1InWant)
```
bduffany, line 1711: If the user force pushes the branch to a different commit before we get to this point, I think the commit could no longer exist and the checkout might fail?

I think we sort of already have this problem today though since we're currently just fetching by branch name so I'd be fine to keep as-is

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2084,14 +2110,19 @@ func getStructuredCommandLine() *clpb.CommandLine {
 }
 
 func configureGlobalCredentialHelper(ctx context.Context) error {
-	if !strings.HasPrefix(*targetRepoURL, "https://") {
+	baseURL := *pushedRepoURL
+	isFork := *targetRepoURL != "" && *pushedRepoURL != *targetRepoURL
+	if isFork {
+		baseURL = *targetRepoURL
+	}
```
bduffany, line 2117: nit: this logic is duplicated in a few places, might be worth extracting a function

bduffany (approved):

enterprise/server/cmd/ci_runner/main.go
maggie-lou: I think it's okay to ignore that case. If the user force pushes a branch, it feels valid that they should not be able to use an older commit that they manually removed from git history

bduffany: sgtm

PR #6522, opened 2024-05-07
Allow moving mouse outside heatmap while making selection

bduffany:
A small itch I've wanted to scratch for a while. I find this a bit easier to use because it means I don't have to keep my mouse perfectly within the chart bounds while making a selection (this gets more difficult when the selection is close to the edge of the chart), but lmk if you think the old behavior is better and I'd be fine closing this

Screenshot (shows current behavior first, then new behavior)

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/1a733863-62bd-4430-9440-ad746f0eb83d

**Related issues**: N/A

### DIFF
--- a/enterprise/app/trends/heatmap.tsx
+++ b/enterprise/app/trends/heatmap.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import moment from "moment";
 import Long from "long";
-
+import { clamp } from "../../../app/util/math";
 import { stats } from "../../../proto/stats_ts_proto";
 import { ScaleBand, scaleBand } from "d3-scale";
 import { withResizeDetector } from "react-resize-detector";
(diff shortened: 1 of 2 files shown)


enterprise/app/trends/heatmap.tsx
```diff
@@ -220,12 +227,13 @@ class HeatmapComponentInternal extends React.Component<HeatmapProps, State> {
...
       return;
     }
+    document.addEventListener("mousemove", this.onMouseMove);
     document.addEventListener("mouseup", (e) => this.onMouseUp(e), { once: true });
     this.pendingClick = [data, data];
     this.setState({ selectionToRender: this.pendingClick });
   }
 
-  onMouseMove(e: React.MouseEvent<SVGSVGElement, MouseEvent>) {
+  onMouseMove = (e: MouseEvent) => {
```
jdhollen, line 236: heh, this syntax kinda grosses me out a bit even though i understand it's perfectly valid.  there's part of me that would almost prefer defining the field as `(e) => this.onMouseMove(e)` and keeping the function defined as-is, but i realize this is just a quirk of mine.  not sure why i'm still typing.

jdhollen (approved):
thanks for this, not sure why i went the other way when implementing this tbh.

PR #6525, opened 2024-05-08
invocation_action_card: render output symlinks

sluongng:
Also fixed Command.environmentVariables and platform properties to
render as `None` when emptied.


sluongng:
Preview:

image

bduffany (approved):
nit: the text alignment feels a bit off - wdyt about this? (Changes are in the `symlinks-css-tweaks` branch if you want to apply them)

image

PR #6529, opened 2024-05-08
Publish workflow invocation workspace status event sooner

bduffany:
WorkspaceStatus is one of the events that prevents us from writing the Invocation row to the DB (due to the code [here](https://github.com/buildbuddy-io/buildbuddy/blob/a5f5c0be34c89dbaa82a913977f5dec8feee0898/server/build_event_protocol/build_event_handler/build_event_handler.go#L977-L978) and [here](https://github.com/buildbuddy-io/buildbuddy/blob/a5f5c0be34c89dbaa82a913977f5dec8feee0898/server/build_event_protocol/build_event_handler/build_event_handler.go#L1168-L1175)), meaning the UI will be stuck in "Invocation is in progress..." state while we are cloning the git repo. This PR moves the WorkspaceStatus event to be published before attempting to sync the git repo, so that we can get a nicer invocation page sooner even if the git repo sync is taking longer than usual.

Verified by adding an artificial 10s sleep to `ws.setup()` locally, and checking that this results in "Invocation in progress..." state being displayed for at least 10s, while after the fix, we can see the full invocation page while the sleep is happening.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -855,6 +855,25 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 			return nil
 		}
 	}
+	workspaceStatusEvent := &bespb.BuildEvent{
+		Id: &bespb.BuildEventId{Id: &bespb.BuildEventId_WorkspaceStatus{WorkspaceStatus: &bespb.BuildEventId_WorkspaceStatusId{}}},
+		Payload: &bespb.BuildEvent_WorkspaceStatus{WorkspaceStatus: &bespb.WorkspaceStatus{
+			Item: []*bespb.WorkspaceStatus_Item{
+				{Key: "BUILD_USER", Value: ar.username},
+				{Key: "BUILD_HOST", Value: ar.hostname},
+				{Key: "GIT_BRANCH", Value: *pushedBranch},
+				{Key: "GIT_TREE_STATUS", Value: "Clean"},
+				// Note: COMMIT_SHA may not actually reflect the current state of the
+				// repo since we merge the target branch before running the workflow;
+				// we set this for the purpose of reporting statuses to GitHub.
+				{Key: "COMMIT_SHA", Value: *commitSHA},
+				{Key: "REPO_URL", Value: baseURL},
+			},
+		}},
+	}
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -855,6 +855,25 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
...
+		Payload: &bespb.BuildEvent_WorkspaceStatus{WorkspaceStatus: &bespb.WorkspaceStatus{
+			Item: []*bespb.WorkspaceStatus_Item{
+				{Key: "BUILD_USER", Value: ar.username},
+				{Key: "BUILD_HOST", Value: ar.hostname},
+				{Key: "GIT_BRANCH", Value: *pushedBranch},
+				{Key: "GIT_TREE_STATUS", Value: "Clean"},
+				// Note: COMMIT_SHA may not actually reflect the current state of the
+				// repo since we merge the target branch before running the workflow;
+				// we set this for the purpose of reporting statuses to GitHub.
+				{Key: "COMMIT_SHA", Value: *commitSHA},
```
maggie-lou, line 869: One thing to note is that we sometimes set commitSha in sync() if it was not passed in as a flag (i.e. only a branch was passed). For that case, now the UI won't contain the commit sha

maggie-lou (approved):

enterprise/server/cmd/ci_runner/main.go
bduffany: If this becomes a major issue for people, we could write to the DB twice, once when seeing WorkflowConfigured and once when seeing WorkspaceStatus

bduffany: hmm actually I think this might affect any users who are currently using the ExecuteWorkflow API and only passing in a branch. For those people, we won't set COMMIT_SHA which would break GH status publishing.

@tylerwilliams wdyt about allowing workflows to do 1 extra invocation DB write - once when getting WorkflowConfigured, and once after checking out the git repo? The problem it'd solve is that today, we don't always know the commit_sha until checking out the repo, but if the git fetch is slow then the user is stuck in this "Invocation in progress..." page until it's done, rather than being able to see the status of the fetch

bduffany:
Will close in favor of saving the invocation row when we get WorkflowConfigured.

PR #6532, opened 2024-05-08
CLI fix fixes

siggisim:
- Support searching for `bazel_deps` that contain repo_name (like ours do) ex:
```
bazel_dep(name = "rules_go", version = "0.47.1", repo_name = "io_bazel_rules_go")
```

- Add repo_root flag to gazelle invocations with bzlmod enabled (because gazelle has a hard time locating this repo root)

---

Still trying to understand / figure out a better solution for this gnarliness https://github.com/buildbuddy-io/buildbuddy/blob/master/cli/fix/fix.go#L76 which doesn't seem to work for me. But these two changes are nice to have regardless.

### DIFF
--- a/cli/add/add.go
+++ b/cli/add/add.go
@@ -30,7 +30,7 @@ Adds the given dependency to your WORKSPACE file.
 	footerTemplate = "###### End auto-generated section for %s ######"
 
 	headerRegex = regexp.MustCompile(`##### Begin auto-generated section for \[https://registry\.build/(.+?)@(.+?)\]`)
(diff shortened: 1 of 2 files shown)


bduffany (approved):

fmeum:
The canonical repo name is considered an implementation detail and changed between 7.0.0 and 7.1.0. 

Do we need to hardcode it? Instead, couldn't we use `bazel run` with a "friendly" (aka apparent) repository name to run a `gazelle` target instead? That would have the benefit of ensuring that the repo config file is up-to-date (especially, that it exists in the first place) and we wouldn't have to duplicate the logic.

fmeum:
@sluongng Could you report that crash as a Bazel issue?

PR #6533, opened 2024-05-08
Remove support for creating WORKSPACE files

siggisim:
I don't think this was ever widely used, and it only worked in very limited circumstances.

I'd rather focus our efforts on making MODULE.bazel file generation work well.

This also allows us to remove this grossness where we call `bazel info release starlark-semantics`.


sluongng (approved):
Much cleaner!

LGTM

bduffany (approved):

PR #6534, opened 2024-05-08
Remove bazel query from cli-initiated gazelle run

siggisim:
I'm not totally sure what this code was initially trying to achieve, but I don't think it's working as intended.

It introduces a `bazel query` before we call bazel which really slows things down (and makes errors difficult to parse since the output is hidden).

When I try to run this (either on the BuildBuddy repo or on a fresh repo), I get:
```
ERROR: no such package '@@gazelle~override~go_deps~bazel_gazelle_go_repository_config//': The repository '@@gazelle~override~go_deps~bazel_gazelle_go_repository_config' could not be resolved: Repository '@@gazelle~override~go_deps~bazel_gazelle_go_repository_config' is not defined
```

We probably need to do this some other way if it's truly needed, but removing it for now since it seems broken.


bduffany (approved):

sluongng:
Here is how the file looks on disk

```bash
> head -20 $(bazel info --enable_bzlmod output_base)/external/bazel_gazelle_go_repository_config/WORKSPACE
# Code generated by generate_repo_config.go; DO NOT EDIT.

go_repository(
    name = "cat_dario_mergo",
    importpath = "dario.cat/mergo",
)

go_repository(
    name = "cc_mvdan_gofumpt",
    importpath = "mvdan.cc/gofumpt",
)

go_repository(
    name = "cc_mvdan_interfacer",
    importpath = "mvdan.cc/interfacer",
)

go_repository(
    name = "cc_mvdan_lint",
    importpath = "mvdan.cc/lint",
...
```

siggisim:
I will focus on fixing this next, but going to remove for now since the current code isn't working anymore.

PR #6535, opened 2024-05-09
prefetch first 3 files in a review.

jdhollen:
making slow but steady progress on optimizing ui lag on load, so this is still slow, but it's at least better than it was before.

**Related issues**: N/A


enterprise/app/review/view_pull_request.tsx
```diff
@@ -691,3 +694,24 @@ export default class ViewPullRequestComponent extends React.Component<ViewPullRe
     );
   }
 }
+
+const FILES_TO_PREFETCH = 3;
```
iain-macdonald, line 698: nit: should this go at the top of the file?

jdhollen: sure, done.

jdhollen (commented):
See comments

PR #6536, opened 2024-05-09
dirtools: fix handling symlink inside output dir

sluongng:
Switch from `dirHelper.IsOutputPath` to `dirHelper.ShouldUploadFile` to detect cases where the symlink was not explicitly declared as a command output, but a parent directory was.

Move this check outside of the `if` clause as we need to do this regardless of REAPI protocol version.

Use only the symlink file name instead of the full path as when we add the symlink to its parent directory.

Closes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3347


sluongng:
Will add some tests to better catch these cases in the future.

bduffany (approved):

PR #6537, opened 2024-05-09
Restrict group_id label cardinality

bduffany:
The "Action stage durations" chart cardinality is too high and makes it slow to load. A single query loads over 300K individual timeseries from the DB.

Restricting to just prober group IDs should fix this (we want to split out prober latency to keep the prober latency chart working). If we need to query by other group IDs, we can use Clickhouse instead.

**Related issues**: N/A

### DIFF
--- a/server/util/metricsutil/metricsutil.go
+++ b/server/util/metricsutil/metricsutil.go
@@ -10,9 +10,9 @@ var (
 	observedGroupIDs = flag.Slice("metrics.observed_group_ids", []string{}, "Group IDs allowed in group_id metrics labels. Groups not in this list will be labeled as 'ANON' if unauthenticated or 'other' if authenticated.")
 )
 
-// GroupIDLabel returns a value suitable for use with the GroupID label. It
-// restricts the observed group IDs to the allowed values configured by flag.
-func GroupIDLabel(groupID string) string {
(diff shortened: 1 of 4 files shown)


bduffany:
This would break the prober latency graph, need to think of how to keep that working.

server/util/metricsutil/metricsutil.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/buildbuddy-io/buildbuddy/server/util/flag"
+)
+
+var (
+	observedGroupIDs = flag.Slice("metrics.observed_group_ids", []string{}, "Group IDs allowed in group_id metrics labels. Groups not in this list will be labeled as 'ANON' if unauthenticated or 'other' if authenticated.")
+)
+
+// GroupIDLabel returns a value suitable for use with the GroupID label. It
+// restricts the observed group IDs to the allowed values configured by flag.
+func GroupIDLabel(groupID string) string {
```
vadimberezniker, line 15: Can we call this FilteredGroupIDLabel? There will still be cases where we will want unfiltered group IDs on metrics.

vadimberezniker (approved):

server/util/metricsutil/metricsutil.go
bduffany: Done

PR #6540, opened 2024-05-09
Fix kythe workflow action

maggie-lou:
Currently, the kythe workflow action is only running for the ExecuteWorkflow API and not when a webhook triggers the workflow service. My understanding is the opposite should be happening

Consolidate the code where we match workflow actions to prevent this from happening in the future.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -557,24 +557,24 @@ func (ws *workflowService) ExecuteWorkflow(ctx context.Context, req *wfpb.Execut
 				return
 			}
 			invocationID = invocationUUID.String()
-			ctx = log.EnrichContext(ctx, log.InvocationIDKey, invocationID)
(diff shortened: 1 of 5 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -685,15 +681,22 @@ func (ws *workflowService) InvalidateAllSnapshotsForRepo(ctx context.Context, re
 	return err
 }
 
-func (ws *workflowService) enableExtraKytheIndexingAction(ctx context.Context) (bool, error) {
+func (ws *workflowService) addDefaultConfigActions(ctx context.Context, c *config.BuildBuddyConfig, workflow *tables.Workflow, wd *interfaces.WebhookData) error {
```
bduffany, line 684: uber nit: maybe `addKytheActionIfEnabled` since "DefaultConfigActions" feels a bit ambiguous (could refer to the default "Test" action that we create when bb.yaml doesn't exist)

bduffany (approved):
nice!

PR #6542, opened 2024-05-09
some performance improvements to diff rendering--not perfect but way better

jdhollen:
- Don't re-render when size hasn't changed.
- Debounce resize events instead of relying on Monaco to do it.
- Dispose editor when component is unmounted (preventing attempts to re-render)
- Corresponding hacks and nonsense to make things work clean.
- A random bug fix where zones were being accounted to the wrong editor.

**Related issues**: N/A


jdhollen:
Whoops, forgot to rename a few things.

jdhollen:
alright, there we go

siggisim (approved):

PR #6544, opened 2024-05-09
Add initial execution progress streaming impl

bduffany:
* Adds a `Publisher` interface that lets us publish the new progress updates concurrently with other progress updates. We need this because there's a loop in `ExecuteTaskAndStreamResults` which periodically publishes progress updates while the task is executing, but we also want to publish fine-grained progress events while this loop is ongoing.
* Publish the basic progress updates: pulling image, downloading inputs, executing, uploading outputs.
* Update `bb execute` to log debug messages with these progress updates.
* Some small refactoring in the `rexec` / `operation` packages.

TODO in future PRs:
* Publish progress for VM boot / snapshot resume.
* Don't publish "pulling image" status if we're not actually pulling an image.

Screenshot showing it in action (with `bb execute` for now as a simple client):

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/5ca844f9-9628-4be5-ba6f-ac8aa2923f3a

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/operation/operation.go
+++ b/enterprise/server/remote_execution/operation/operation.go
@@ -51,11 +51,7 @@ func (p *Publisher) Context() context.Context {
 	return p.stream.Context()
 }
 
-// Send publishes a message asynchronously. It is safe for concurrent use.
-//
-// If an error happened in a previous call, it may be surfaced by a later call
-// to this function. This delayed error surfacing is compatible with the stream
-// API, since Send() does not wait for the recipient to ack.
+// Send publishes a message on the stream. It is safe for concurrent use.
 func (p *Publisher) Send(op *longrunning.Operation) error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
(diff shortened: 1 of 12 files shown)


enterprise/server/remote_execution/operation/operation.go
```diff
@@ -167,31 +240,38 @@ func (c *retryingClient) CloseAndRecv() (*repb.PublishOperationResponse, error)
...
+		if err != nil {
+			return nil, err
+		}
+		op.Metadata = mdAny
 	}
-	operation.Result = &longrunning.Operation_Response{Response: result}
-
-	if stage == repb.ExecutionStage_COMPLETED {
-		operation.Done = true
+	if er != nil {
```
maggie-lou, line 267: uber-nit: Could you rename `er`? It looks like a typo of `err`

maggie-lou (approved):

enterprise/server/remote_execution/operation/operation.go
bduffany: haha it does :laughing: done

PR #6545, opened 2024-05-10
action_card: display output dirs as trees

sluongng:
Render the output directories as expandable trees so that users could
explore their contents.

This PR is based on #6525 so we should only merge this PR after.

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -568,7 +568,7 @@ export default class InvocationActionCardComponent extends React.Component<Props
         {symlinks.length ? (
           <div className="action-list">
             {symlinks.map((symlink) => (
-              <div>
+              <div className="tree-node-symlink">
                 <span>
                   <FileSymlink className="icon symlink-icon" />
                 </span>{" "}
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -416,24 +416,38 @@ export default class InvocationActionCardComponent extends React.Component<Props
       rpcService.downloadBytestreamFile(node.obj.name, dirUrl, this.props.model.getInvocationId());
       return;
     }
+
     rpcService
       .fetchBytestreamFile(dirUrl, this.props.model.getInvocationId(), "arraybuffer")
-      .then((buffer) => {
-        let dir = build.bazel.remote.execution.v2.Directory.decode(new Uint8Array(buffer));
+      .then((buffer: ArrayBuffer) => new Uint8Array(buffer))
+      .then((array: Uint8Array) =>
+        node.type == "tree"
+          ? build.bazel.remote.execution.v2.Tree.decode(array).root
+          : build.bazel.remote.execution.v2.Directory.decode(array)
+      )
+      .then((dir: build.bazel.remote.execution.v2.Directory | null | undefined) => {
+        if (!dir) {
+          return;
+        }
+
         this.state.treeShaToExpanded.set(digestString, true);
-        let directories: TreeNode[] = dir.directories.map((child) => ({
-          obj: child,
-          type: "dir",
-        }));
-        let files: TreeNode[] = dir.files.map((child) => ({
-          obj: child,
-          type: "file",
-        }));
-        let symlinks: TreeNode[] = dir.symlinks.map((child) => ({
-          obj: child,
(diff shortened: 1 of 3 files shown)


app/invocation/invocation_action_card.tsx
```diff
@@ -517,6 +531,62 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+        <div className="action-property-title">Output directories</div>
+        {actionsResult.outputDirectories.length ? (
+          <div className="action-list">
+            {actionsResult.outputDirectories.map((dir) => (
+              <TreeNodeComponent
+                node={{
+                  obj: new build.bazel.remote.execution.v2.DirectoryNode({ name: dir.path, digest: dir.treeDigest }),
+                  type: "dir",
+                }}
+                treeShaToExpanded={this.state.treeShaToExpanded}
```
sluongng, line 546: Unlike input, which only has one tree, there could be multiple output directories in ActionResult, which is currently translated to multiple trees. This has aside effects:

The current implementation shares the same `treeShaToExpanded` state among all input/output trees. This could cause expanding 1 tree in input to also expand that tree in output. In practice, I think it's unlikely that these directories are duplicated. We could fix this later by providing a separate state for each tree.

sluongng:
A preview

image

app/invocation/invocation_action_card.tsx
sluongng: A second problem with this approach is that if there are 2 or more output dirs, the prefix `bazel-out/<target-platform>/bin` prefix is not duplicated between the rendered trees.

I briefly considered deduplicating them, but when we do that, it would not be clear to the user which directories in the merged tree were the output directory specified in the ActionResult.

I plan to add `Expected output paths` to render the Command data in a future PR. Once that's done, I think we could reconsider merging the output trees in our UI.

siggisim:
> A preview
> 
> image

Just some style nits:

It looks like the two new icons need to be sized the same as the folder / download icons - and I suspect some  `display:flex; align-items: center;` is needed on that whole new row. And maybe a little padding around the arrow

Also, is there a digest for these we could render like for the other files?

sluongng:
Here is the latest preview which includes @bduffany CSS fix from the previous PR

image

> Also, is there a digest for these we could render like for the other files?

The symlink could point to an existing file in the tree, or it could point to a file on the execution host or a non-existing file, in which case it would be a dangling symlink.

We could consider adding a digest when the file it points to does exist in the tree, but I would prefer attempting that in a separate PR.

siggisim:
> Here is the latest preview which includes @bduffany CSS fix from the previous PR
> 
> image

The icons look a little darker than the download / file icons. And I feel like the arrow icon should potentially be a bit smaller.

app/invocation/invocation_action_tree_node.tsx
```diff
@@ -20,7 +20,7 @@ type SymlinkNode = build.bazel.remote.execution.v2.SymlinkNode;
 
 export interface TreeNode {
   obj: FileNode | DirectoryNode | SymlinkNode;
-  type: "file" | "dir" | "symlink";
+  type: "file" | "dir" | "symlink" | "tree";
 }
```
bduffany, line 24: I missed this in the earlier PR, but this looks like an ideal use case for the [discriminated union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) pattern.

```typescript
export type TreeNode = {
  obj: FileNode;
  type: "file";
} | {
  obj: DirectoryNode;
  type: "directory";
} | {
  obj: SymlinkNode;
  type: "symlink";
} | {
  obj: DirectoryNode;
  type: "tree";
}
```

If you defined the type like that, then you could get some nicer type narrowing:

```typescript
if (node.type === "symlink") {
  // Within this `if` block, TS narrows the type of `node` to `{ obj: SymlinkNode; type: "symlink" }`

  doSomethingWith(obj.digest); // This doesn't compile, because symlink doesn't have a digest field
}
```

sluongng:
image

Changed the opacity of the symlink icons to be more consistent with the rest of the icons in the tree.

app/invocation/invocation_action_tree_node.tsx
sluongng: Great suggestion! 🎉 

I changed it to union type.

PR #6546, opened 2024-05-10
change default container image to 20.04

sluongng:
This is a follow up from
https://github.com/buildbuddy-io/buildbuddy-toolchain/pull/25


bduffany (changes requested):
I think this will break people who aren't setting a container-image explicitly in their actions, no?

bduffany:
I think it'd be good to add some logging first to see if there are any customers who aren't explicitly setting the container image, and proactively warn them about this upcoming change

siggisim:
Yeah, I think we should avoid breaking changes like these. We should recommend folks upgrade their toolchain version if they'd like a newer base image, and should generally recommend against using the empty value here.

sluongng:
Alright, let's close this for now and think about what to do with empty `container-image` `exec_properties` value 🤔

PR #6554, opened 2024-05-12
invocation_action_card: render expected outputs

sluongng:
Render the command's expected outputs so that users can compare them
against the actual outputs in the ActionResult.

Base on https://github.com/buildbuddy-io/buildbuddy/pull/6545

### DIFF
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -1062,7 +1062,6 @@
 .action-list .missing-output {
   display: flex;
   align-items: center;
-  gap: 4px;
 
   .icon {
     height: 18px;
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -1067,7 +1067,6 @@
     height: 18px;
     width: 18px;
     opacity: 0.7;
-    stroke: red;
   }
 
   .expected-output-label {
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -1054,11 +1054,20 @@
     opacity: 0.7;
   }
 
-  .icon.check-icon {
-    stroke: green;
+  .expected-output-label {
+    color: #444;
   }
+}
 
-  .icon.unlink {
+.action-list .missing-output {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+
+  .icon {
+    height: 18px;
+    width: 18px;
+    opacity: 0.7;
     stroke: red;
   }
 
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -165,7 +165,9 @@
   color: #757575;
 }
 
-.invocation-action-card .icon.file-icon {
+.invocation-action-card .icon.file-icon,
+.invocation-action-card .icon.folder-icon,
+.invocation-action-card .icon.file-question-icon {
   margin-right: 4px;
   width: 18px;
   height: 18px;
@@ -1049,8 +1051,6 @@
   gap: 4px;
 
   .icon {
-    height: 18px;
-    width: 18px;
     opacity: 0.7;
   }
 
--- a/app/invocation/invocation.css
+++ b/app/invocation/invocation.css
@@ -1043,6 +1043,28 @@
   color: #aaa;
 }
 
+.action-list .expected-output-name {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+
+  .icon {
+    height: 18px;
+    width: 18px;
+    opacity: 0.7;
+  }
+  .icon.check-icon {
+    stroke: green;
+  }
+  .icon.unlink {
+    stroke: red;
+  }
+}
+
+.expected-output-name .expected-output-label {
+  color: #444;
+}
+
 .state-page button.impersonate-button,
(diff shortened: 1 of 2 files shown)


sluongng:
Preview 
image

sluongng:
In the case of `output_files` and `output_directories`, we can tell whether the client expected a file or a dir. However, these are deprecated in favor of `output_paths`, which is not possible to tell whether they are files or dirs.

sluongng:
A potentially nice addition to this would be if the ActionResult is available, we could 

(a) check if the expected output was created and display a ✅ or ❓ to indicate output available/missing.

(b) display the file type (directory, file, symlink) of the expected output

sluongng:
Hmm, I am not implementing either (a) or (b) because... I can't come up with a design that looks right. 😆 

image
image

sluongng:
Hmm, it looks a bit better?
image

sluongng:
Preview of the latest changes

image

sluongng:
@siggisim Oh I think changing this to "Missing Outputs" is a great idea!

But in the case when the Action Result is not available, we should still display the Expected Outputs?

bduffany:
I kind of like having this flat list of outputs - the outputs tree is a bit clunkier because you have to keep clicking on each directory which has some latency.

sluongng:
@bduffany currently we only render output directories as trees. Output files and output symlinks are rendered as flat list still. And for the directory, we do 1 tree per dir. So no clicking is required to see all the actual outputs. Only when you want to explore the contents of the output dirs, you can now click for more. 

This PR is more about the Expected Outputs. For example, we do render Actions which dont have attached Action Result. In such case, user might still be interested in reviewing what the build rules was expecting the Command to produce. 

If Action Result does exist, then there are some duplicated information between the Expected Outputs and actual (File, Dir, Symlink) Outputs like Siggi suggested.

siggisim:
> @bduffany currently we only render output directories as trees. Output files and output symlinks are rendered as flat list still. And for the directory, we do 1 tree per dir. So no clicking is required to see all the actual outputs. Only when you want to explore the contents of the output dirs, you can now click for more.
> 
> This PR is more about the Expected Outputs. For example, we do render Actions which dont have attached Action Result. In such case, user might still be interested in reviewing what the build rules was expecting the Command to produce.
> 
> If Action Result does exist, then there are some duplicated information between the Expected Outputs and actual (File, Dir, Symlink) Outputs like Siggi suggested.

I think we should only show expected outputs in two cases:
1) We don't have the list of actual outputs (i.e. the Action Result isn't present)
2) The actual outputs and the expected outputs differ (i.e. there is a missing output)

If we show it in other cases, it's a little confusing why they're not clickable, why there are two sets, etc.

sluongng:
Preview

- Expected Result:
image


image

bduffany:
Can you fix the text alignment? The icon alignment looks OK but the alignment of filenames vs. dir names looks off.

Also can we show directories first, then files? Most file UIs that I've used will show directories first, then files.

image

sluongng:
Latest previews:

1. Expected outputs

image

3. Missing outputs

image

app/invocation/invocation.css
```diff
@@ -1030,6 +1032,36 @@
...
+
+.action-list .missing-output {
+  display: flex;
+  align-items: center;
+
+  .icon {
+    height: 18px;
+    width: 18px;
+    opacity: 0.7;
+    stroke: red;
```
bduffany, line 1057: nit: add the `icon-red` class to the icon instead, to get the material red color that we use elsewhere in the UI

app/invocation/invocation_action_card.tsx
```diff
@@ -587,6 +587,126 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+            <span>
+              <Folder className="icon file-question-icon" />
+            </span>
+            <span className="missing-output-label">{missingDir}</span>
+          </div>
+        ))}
+        {missingFiles.map((missingFile) => (
+          <div className="missing-output">
+            <span>
+              <FileQuestion className="icon file-question-icon" />
```
bduffany, line 669: (per other commment)

```suggestion
              <FileQuestion className="icon file-question-icon icon-red" />
```

app/invocation/invocation_action_card.tsx
```diff
@@ -587,6 +587,126 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+        ))}
+      </>
+    );
+
+    const renderMissingOutputFilesAndDirs = (missingFiles: Array<string>, missingDirs: Array<string>) => (
+      <>
+        {missingDirs.map((missingDir) => (
+          <div className="missing-output">
+            <span>
+              <Folder className="icon file-question-icon" />
```
bduffany, line 661:
```suggestion
              <Folder className="icon file-question-icon icon-red" />
```

bduffany (approved):

app/invocation/invocation.css
bduffany: err it's actually just `red` not `icon-red`. so `className="icon red"`

PR #6556, opened 2024-05-13
invocation_timing_card: add windows alt separator

sluongng:
Windows support both backward and forward slash as valid separator.
Fix detecting the wrong name for the timing profile.

### DIFF
--- a/app/invocation/invocation_timing_card.tsx
+++ b/app/invocation/invocation_timing_card.tsx
@@ -79,8 +79,17 @@ export default class InvocationTimingCardComponent extends React.Component<Props
         ?.find((scl) => scl.commandLineLabel == "canonical")
         ?.sections?.find((s) => s.sectionLabel == "command options")
         ?.optionList?.option?.find((o) => o.optionName == "profile")?.optionValue ?? "command.profile.gz";
-    const separator = profilePath.includes("\\") ? "\\" : "/";
-    const profileName = profilePath.substring(profilePath.lastIndexOf(separator) + 1);
+
+    let profileName: string;
+    if (profilePath.includes("\\")) {
+      // Windows supports both forward slash and backward slash as separator.
+      const separator = "\\";
+      const altSeparator = "/";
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_timing_card.tsx
```diff
@@ -79,8 +79,17 @@ export default class InvocationTimingCardComponent extends React.Component<Props
         ?.find((scl) => scl.commandLineLabel == "canonical")
         ?.sections?.find((s) => s.sectionLabel == "command options")
         ?.optionList?.option?.find((o) => o.optionName == "profile")?.optionValue ?? "command.profile.gz";
-    const separator = profilePath.includes("\\") ? "\\" : "/";
-    const profileName = profilePath.substring(profilePath.lastIndexOf(separator) + 1);
+
+    let profileName: string;
+    if (profilePath.includes("\\")) {
```
fmeum, line 84: Couldn't we just replace all `\\` with `/` and then search? Would be more readable and I doubt the extra allocation matters.

sluongng: It would only make the code a bit shorter, but not clearer (at least to me). 

My concern is when a Unix build has a `\` slash in its log file name but is escaped. In that case, we will wrongly detect it as a Windows path.

I wonder if there is some information in BEP that lets you tell for certain what is the host OS. That way, we could use that to switch between Windows and Unix logic.

fmeum: I agree that it's not ideal to handle `\` as a separator on non-Windows, but doesn't the current version also not handle escaped backslashes?

We could probably identify Windows vs. Unix based on the client env, for example `PATH`. That would be better.

sluongng:
Discussed with Fabian on Slack: that it might be easier in the future if we just make the profile name in the BuildToolLogs event a constant instead.

bduffany (approved):

fmeum (approved):

PR #6560, opened 2024-05-13
Update RPC service to support server-streaming RPCs

bduffany:
Allows defining server-streaming RPC methods, and calling them like this:

```typescript
const stream = rpc_service.service.waitExecution(
  // First argument: provide the request message object or fields (same as calling a unary RPC)
  { executionId },

  // Second argument (new): provide stream handlers.
  // All handlers are required
  {
    // Handle response message
    next: (response) => { /* handle response */ },

    // Handle error
    // In most cases, we should check for network errors here and retry the stream w/ exponential backoff,
    // in case of app restarts or a flaky network connection. In a later PR we can probably add some utils
    // to make this easier.
    error: (e) => { /* ... */ },

    // Handle EOF (no more data on stream).
    complete: () => { /* ... */ },
  },
);

// ... later, e.g. componentWillUnmount():

// Cancels the underlying fetch. The error/complete callbacks will not be invoked.
stream.cancel();
```

This API is defined by https://github.com/buildbuddy-io/protoc-gen-protobufjs so if there's anything about this API you don't like, LMK - it's pretty easy to change it upstream since nothing is using this yet. The naming of `next` / `error` / `complete` is inspired by `rxjs` (but the generated code does not use rxjs, just vanilla JS + the existing protobufjs deps).

I tested this E2E in a WIP branch, calling `WaitExecution` and streaming real-time execution state to the UI.

**Related issues**: N/A

### DIFF
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -462,8 +462,10 @@ type CancelableService<Service extends protobufjs.rpc.Service> = protobufjs.rpc.
       infer Request,
       infer Response
     >
-      ? UnaryRpcMethod<Request, Response>
-      : Service[MethodName];
+      ? /* Unary RPC: transform the generated method's return type from Promise to CancelablePromise. */
+        UnaryRpcMethod<Request, Response>
+      : /* Server-streaming RPC: keep the original method as-is. */
+        Service[MethodName];
   };
 
 type FetchPromiseType<T extends FetchResponseType> = T extends ""
--- a/app/service/rpc_service.ts
+++ b/app/service/rpc_service.ts
@@ -1,11 +1,21 @@
 import { Subject } from "rxjs";
-import { buildbuddy } from "../../proto/buildbuddy_service_ts_proto";
+import { buildbuddy, $stream } from "../../proto/buildbuddy_service_ts_proto";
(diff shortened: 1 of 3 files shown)


jdhollen:
Is it actually worth making the error + complete handlers optional? it seems like every language is structured differently for this so i guess the callback structure is kiiinda just whatever we want (golang returns `(resp, err)`, java has a blocking api with exceptions, etc.), so i like this in general, i think personally i'd rather just enforce these fns existing (even if empty) instead of letting folks accidentally forget them.

bduffany:
> Is it actually worth making the error + complete handlers optional?

I was on the fence about making the error handler optional - seems like a good practice to always define one, so I'll go ahead and make that change.

I'm not sure about the `complete` handler though - seems like in a good number of cases you don't necessarily care what happens after you're done reading from the stream. I'd be fine making it required for now though and then later if we find we're always defining it as `() => {}` we can maybe clean that up.

bduffany:
Upgraded the codegen to a new version which makes `error` and `complete` required, and also fixes the build errors.

jdhollen (approved):

siggisim (approved):

PR #6563, opened 2024-05-13
Add flags to make metrics collector expiries configurable

tempoz:
The current default expiration times for the metrics collector are extremely generous, and can result in us getting alerts about redis cache usage. This happens especially when users are running a lot of invocations that only use us as the cache without sending us build events, as we can't manage the lifetime of those scorecards and counts in any way other than by allowing them to expire when they hit their ttl, since we have no idea when the related invocation is over.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
+++ b/enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
@@ -2,7 +2,6 @@ package redis_metrics_collector
 
 import (
 	"context"
-	"flag"
 	"strconv"
 	"time"
 
@@ -10,12 +9,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
 	"github.com/go-redis/redis/v8"
 
-	_ "github.com/buildbuddy-io/buildbuddy/server/remote_cache/hit_tracker"
-	flagtypes "github.com/buildbuddy-io/buildbuddy/server/util/flagutil/types"
-)
-
-var (
-	countExpiration = flagtypes.Alias[time.Duration](flag.CommandLine, "cache.count_ttl")
+	cache_config "github.com/buildbuddy-io/buildbuddy/server/cache/config"
 )
 
 type collector struct {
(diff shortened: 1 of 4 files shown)


enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
```diff
@@ -2,17 +2,20 @@ package redis_metrics_collector
...
+
+	_ "github.com/buildbuddy-io/buildbuddy/server/remote_cache/hit_tracker"
+	flagtypes "github.com/buildbuddy-io/buildbuddy/server/util/flagutil/types"
 )
 
-const (
-	// How long until counts expire from Redis.
-	countExpiration = 1 * 24 * time.Hour
+var (
+	countExpiration = flagtypes.Alias[time.Duration](flag.CommandLine, "cache.count_expiration")
```
bduffany, line 18: nit: the other package doesn't use the flag - might be better to move that flag here rather than using an alias

bduffany (approved):

server/remote_cache/hit_tracker/hit_tracker.go
```diff
@@ -30,7 +30,9 @@ import (
 )
 
 var (
-	detailedStatsEnabled = flag.Bool("cache.detailed_stats_enabled", false, "Whether to enable detailed stats recording for all cache requests.")
+	detailedStatsEnabled       = flag.Bool("cache.detailed_stats_enabled", false, "Whether to enable detailed stats recording for all cache requests.")
+	scorecardResultsExpiration = flag.Duration("cache.detailed_stats_expiration", 3*time.Hour, "How long to go without receiving any cache requests for an invocation before deleting the invocation's detailed results from the metrics collector. Has no effect if cache.detailed_stats_enabled is not set.")
+	countExpiration            = flag.Duration("cache.count_expiration", 24*time.Hour, "How long to go without receiving any cache requests for an invocation before deleting the invocation's counts from the metrics collector.")
```
bduffany, line 35: nit: maybe `s/_expiration/_ttl/` since it seems we use _ttl for other flags

enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
tempoz: I thought about that, but if we implement expiring stuff in other metrics collectors (like memory_metrics_collector) we'd want to have it here. I can move it if you still think we should, though.

server/remote_cache/hit_tracker/hit_tracker.go
tempoz: sgtm!

enterprise/server/backends/redis_metrics_collector/redis_metrics_collector.go
bduffany: I thought that might have been the motivation - maybe `server/config/cache/cache.go` would be a better place then? The alias feels slightly less clean since it's matching up the two flags based on a loosely typed string

tempoz: works for me; thanks!

PR #6565, opened 2024-05-13
Stream execution progress to the action page

bduffany:
Starting with the action page mostly as a proof of concept because it's easier to test than workflows and is maybe somewhat useful for larger actions. It will be more interesting to show this info on the queued invocation page (for workflow / remote bazel invocations).

Screen recording of an execution with an artificial 8s delay at the beginning of ExecuteTaskAndStreamResults + 750ms delay when transitioning between each execution state (otherwise the updates go by too quickly...) (also the "Unknown" execution status should probably be "Queued", need to fix that separately)

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/d76dd269-d966-48a9-b51e-e30696f142be

**Related issues**: N/A

### DIFF
--- a/app/invocation/execution_status.tsx
+++ b/app/invocation/execution_status.tsx
@@ -45,14 +45,16 @@ export function executionStatusLabel(op: ExecuteOperation) {
     case ExecutionStage.QUEUED:
       return "Queued";
     case ExecutionStage.EXECUTING:
-      // Return fine-grained progress using BB-specific enum
-      break;
+      // Return fine-grained progress using BB-specific metadata
+      return executionProgressLabel(op);
     case ExecutionStage.COMPLETED:
       return "Completed";
     default:
       return "Unknown";
   }
+}
 
+function executionProgressLabel(op: ExecuteOperation) {
   switch (op.progress?.executionState) {
     case ExecutionState.PULLING_CONTAINER_IMAGE:
       return "Pulling container image";
--- a/app/invocation/execution_status.tsx
+++ b/app/invocation/execution_status.tsx
@@ -1,0 +1,104 @@
+import rpc_service, { ServerStreamHandler, ServerStream } from "../service/rpc_service";
+import { execution_stats } from "../../proto/execution_stats_ts_proto";
+import { build } from "../../proto/remote_execution_ts_proto";
+
+const ExecutionStage = build.bazel.remote.execution.v2.ExecutionStage.Value;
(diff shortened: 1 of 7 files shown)


app/invocation/execution_status.tsx
```diff
@@ -0,0 +1,104 @@
...
+    return "Completed";
+  }
+
+  switch (op.metadata?.stage) {
+    case ExecutionStage.CACHE_CHECK:
+      return "Checking cache";
+    case ExecutionStage.QUEUED:
+      return "Queued";
+    case ExecutionStage.EXECUTING:
+      // Return fine-grained progress using BB-specific enum
```
jdhollen, line 48: nit: for readability, might be better to just break out the `executionState`-based text into its own function and call it directly here--the fall-through is a bit awkward.

app/invocation/invocation.tsx
```diff
@@ -197,6 +197,12 @@ export default class InvocationComponent extends React.Component<Props, State> {
       .catch((error: any) => {
         console.error(error);
         this.setState({ error: BuildBuddyError.parse(error) });
+
+        // To make manual testing easier, support a special 'retry' param that
+        // auto-reloads the page if there is an error.
```
jdhollen, line 202: why this one rpc + why this particular behavior?

app/invocation/invocation_execution_card.tsx
```diff
@@ -74,6 +76,12 @@ export default class ExecutionCardComponent extends React.Component<Props, State
         this.fetchUpdatedProgress();
       }
 
+      // To help with manual testing, support a special "openFirst" URL param
+      // which automatically navigates to the first execution that we fetched.
```
jdhollen, line 80: Probably need to add a list of these somewhere (whether by an enum in code or just a list in a markdown file somewhere)--the way that they're placed right now means me and you will be the only people that know about them.

jdhollen:
One extra thing--might be worth putting a tiny little spinner next to the status or something (until it's completed, obviously)?  As it is it's not obvious that the text will potentially update.

I don't _particularly_ like using a spinner, just think there should be something.

jdhollen (approved):

bduffany:
> One extra thing--might be worth putting a tiny little spinner next to the status or something (until it's completed, obviously)? As it is it's not obvious that the text will potentially update.

> I don't particularly like using a spinner, just think there should be something.

I agree, I was thinking of maybe a pulsing blue dot but something tells me it's going to look a little weird. Will play around with this in a separate PR

PR #6570, opened 2024-05-14
Use structured error types in frontend RPC service

bduffany:
(This is all behind the `streaming_http_enabled` flag, which is only enabled in dev currently.)

This PR is preparing for making server-streaming RPCs retryable on the client. Ideally we'd always retry these, otherwise when the apps roll out, the live-streaming would suddenly stop.

We currently throw a somewhat haphazard mixture of `string` and `TypeError` from the RPC service methods when a fetch fails. This PR updates things to be more structured, so that we can more easily distinguish between gRPC status errors, other HTTP errors (with status code + body, but missing gRPC error message), or generic network errors (e.g. can't reach the app, app restarted, etc.).

**Related issues**: N/A


app/service/rpc_service.ts
```diff
@@ -164,25 +171,40 @@ class RpcService {
...
       // Read error message from response body
-      let message = "";
+      let body = "";
       try {
-        message = await response.text();
+        body = await response.text();
       } catch (e) {
-        message = `unknown (failed to read response body: ${e})`;
+        if (structuredErrors) {
+          // If we failed to read the response body then ignore the status code
```
jdhollen, line 183: Not sure how I feel about this precedence--I guess we still have the network tab to tell the difference, I just figure if a 500 error yielded an unparseable response then saying "500" is more useful than "failed to read".

app/util/errors.ts
```diff
@@ -1,30 +1,156 @@
-export type ErrorCode = "Unknown" | "NotFound" | "AlreadyExists" | "PermissionDenied" | "Unauthenticated";
+import { google as google_status } from "../../proto/grpc_status_ts_proto";
+import { google as google_code } from "../../proto/grpc_code_ts_proto";
 
+export type ErrorCode = Omit<GRPCCodeName, "OK">;
+
+/**
+ * High-level error class representing any error from the BuildBuddy server,
+ * either GRPC or HTTP.
```
jdhollen, line 8: The class hierarchy is a little weird here based on the documentation--do you actually intend to flip the code on its head at some point (HTTPStatusError extends BBError) or more cleanly compose things so that HTTPStatusError/FetchError etc are held by BBError?

bduffany: Yeah, thanks for pointing this out - I think what I would like to do is:
- Getting rid of `BuildBuddyError` entirely - it's sort of poorly defined
- In the places where we're doing stuff like `catch (e) { error_service.handleError(BuildBuddyError.parse(e)); }` we would just pass `e` instead of trying to parse the error. It's kind of ugly that we "parse" errors at all and I would rather have the errors be nicely structured at a lower level to begin with.
- The error_service could then do nicer stuff depending on the type of error that it's rendering, rather than receiving a generic `BuildBuddyError` which is more "watered down". e.g. for `FetchError` we could show a nice "network broken" icon or something like that (if we really wanted to, haha)

app/service/rpc_service.ts
bduffany: I was thinking that this approach would play more nicely with retry middleware (on the frontend) if we return a generic network error here. That way if we return an error status but fail to read the response body, the client can at least retry for a chance to get a more informative error message that could maybe help them fix the error. I guess if the client got permanently disconnected though, they will be stuck with the "failed to read" error, but I think that's not too much worse than "500: <failed to read error message>"

bduffany:
Gah I accidentally merged #6578 into this PR. Fixing up the commits now...

app/service/rpc_service.ts
jdhollen: Yeah alright, i'm sold

app/util/errors.ts
jdhollen: Sure, that works, how about plopping something in the comment here about what the plan is?

jdhollen (approved):

PR #6574, opened 2024-05-15
[raft] record last replica id and added timestamp in range descriptor

luluz66:
**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #6575, opened 2024-05-15
[raft] Allow us to query whether a replica is current or behind

luluz66:
**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #6577, opened 2024-05-15
Add target to Execution

sluongng:
Each Execution is often corresponding to a Bazel's Action, but end users
don't often think of Execution or Action because Bazel's build rules
system often abstract that away from them.

Instead, folks often think of Bazel build targets.
Each targets uses a build rule, in combination with a build
configuration, to create multiple actions. Usually each of these actions
are created with a unique mnemonic to help identify them a bit easier
(note that this is not always the case).

Capture the target label from Executor's RequestMetadata when record the
Execution into OLAPDB. We will evaluate adding mnemonic and
configuration at a later stage.

### DIFF
--- a/enterprise/server/remote_execution/execution_server/execution_server.go
+++ b/enterprise/server/remote_execution/execution_server/execution_server.go
@@ -358,9 +358,12 @@ func (s *ExecutionServer) recordExecution(ctx context.Context, executionID strin
 	if err != nil {
 		return status.InternalErrorf("failed to get invocations for execution %q: %s", executionID, err)
 	}
-
+	rmd := bazel_request.GetRequestMetadata(ctx)
 	for _, link := range links {
 		executionProto := execution.TableExecToProto(&executionPrimaryDB, link)
+		if rmd != nil {
+			executionProto.TargetLabel = rmd.GetTargetId()
+		}
 		inv, err := s.env.GetExecutionCollector().GetInvocation(ctx, link.GetInvocationId())
 		if err != nil {
 			log.CtxErrorf(ctx, "failed to get invocation %q from ExecutionCollector: %s", link.GetInvocationId(), err)
--- a/server/util/clickhouse/schema/schema.go
+++ b/server/util/clickhouse/schema/schema.go
@@ -159,6 +159,9 @@ type Execution struct {
 
 	Stage int64
 
+	// RequestMetadata
+	TargetLabel string
+
 	// IOStats
 	FileDownloadCount        int64
 	FileDownloadSizeBytes    int64
@@ -244,6 +247,7 @@ func (e *Execution) AdditionalFields() []string {
 		"InvocationLinkType",
 		"Tags",
 		"OutputPath",
+		"TargetLabel",
 	}
 }
 
--- a/proto/remote_execution.proto
+++ b/proto/remote_execution.proto
@@ -2391,4 +2391,7 @@ message StoredExecution {
 
   string output_path = 31;
   string status_message = 32;
+
+  // TODO(sluongng): add mnemonic and configuration.
+  string target_label = 33;
 }
--- a/server/util/clickhouse/clickhouse.go
+++ b/server/util/clickhouse/clickhouse.go
@@ -234,6 +234,7 @@ func buildExecution(in *repb.StoredExecution, inv *sipb.StoredInvocation) *schem
 		Success:                            inv.GetSuccess(),
 		InvocationStatus:                   inv.GetInvocationStatus(),
 		Tags:                               invocation_format.ConvertDBTagsToOLAP(inv.GetTags()),
+		TargetLabel:                        in.GetTargetLabel(),
 	}
 }
 


sluongng:
The configuration id is interesting because we do have access to the `Configuration Mnemonic` via BEP. For example:

```
[-] Build event 8,460 - configuration
{
    "id": {
        "configuration": {
            "id": "20e01b5335974a27b7b60bd112c74f02267b2fae707284965510ae588c61d6da"
        }
    },
    "configuration": {
        "mnemonic": "linux_x86_64-fastbuild-ST-92d146075906",
        "platformName": "k8",
        "cpu": "k8",
        "makeVariable": {
            "GENDIR": "bazel-out/linux_x86_64-fastbuild-ST-92d146075906/bin",
            "BINDIR": "bazel-out/linux_x86_64-fastbuild-ST-92d146075906/bin",
            "COMPILATION_MODE": "fastbuild",
            "TARGET_CPU": "k8"
        }
    }
}
[-] Build event 8,462 - configuration
{
    "id": {
        "configuration": {
            "id": "b833a96e5c43bc2f3e8e8ea6fc4560405cff455dfc6bde5b5fb9daa4c090f488"
        }
    },
    "configuration": {
        "mnemonic": "linux_x86_64-opt-exec",
        "platformName": "k8",
        "cpu": "k8",
        "makeVariable": {
            "GENDIR": "bazel-out/linux_x86_64-opt-exec/bin",
            "BINDIR": "bazel-out/linux_x86_64-opt-exec/bin",
            "COMPILATION_MODE": "opt",
            "TARGET_CPU": "k8"
        },
        "isTool": true
    }
}
```

So if we have a go_binary being built multiple times, once for each configuration, then we could (a) tell the action execution apart using this config ID and (b) provide the config hash for user to diff them locally with `bazel config` command.

proto/execution_stats.proto
```diff
@@ -127,6 +126,13 @@ message Execution {
   // retried internally, this Execution will reflect the latest execution
   // attempt.
   string execution_id = 14;
+
+  // Information about Bazel build target which created this Action's Execution.
+  // Note that these fields could be empty if the client does not send a
+  // RequestMetadata with these fields populated.
+  string target_id = 15;
```
siggisim, line 133: I generally like the idea of this, but I think the reason we've avoided doing this in the past is that the executions table is already huge (both in number of rows and overall size), and this would add 3 large variable length strings which would further exacerbate the problem.

This feels like something better suited either for Clickhouse or if we ever end up storing the executions for an invocation in a GCS blob like we've talked about.

jdhollen: so i agree that this is a lot of data to jam into mysql when we would probably never do a lookup based on it, but just to be clear, we basically never do lookups unless we're keying off of execution id or invocation id, right?  is there some case i'm forgetting about where having very-wide rows in the executions table actually slows things down?  does it matter for insert speed?

siggisim: Was mostly just thinking for size, and us moving away from using mysql for execution metadata. Our MySQL DB is over 5TB, and I think over half of that is the executions table (though I may be wrong).

Don't get me wrong, I would LOVE to have this data more easily available. If @tylerwilliams is cool with it, I am.

luluz66: wrt to the MySQL execution tables sizes: we probably could look into the execution janitor a bit more and see if it's keeping up with the current rate of executions.

tylerwilliams: I think even with our janitor approach, there is fragmentation happening and we're not recouping space on disk.

I'd prefer not writing more data to the mysql executions table, but I think putting it in clickhouse would be fine?

sluongng: I pushed a change that include the following:

a. Write the metadata to Redis instead of MySQL
b. Flush the data from Redis to ClickHouse when the execution completes
c. When GetExecution is called, fetch the metadata from Redis for incompleted actions and Clickhouse for completed actions

server/util/clickhouse/schema/schema.go
```diff
@@ -157,6 +157,10 @@ type Execution struct {
 
 	Stage int64
 
+	TargetID        string
```
jdhollen, line 160: i know this is what the remote execution proto calls it, but ugh, i really don't like calling this an ID.  i guess this is non-actionable, just felt like i had to say it.

jdhollen (commented):
this all looks good to me, i feel like it's not my zone to approve in given everyone else commenting :P

server/util/clickhouse/schema/schema.go
sluongng: Hmm that's a good point.

What do you think about `TargetName` or `ParentTarget`? I know that we do have a `Targets` table lurking around for the Target Tracker, so renaming this makes sense to me.

jdhollen: The corresponding column in `Targets` is `Label`, so `TargetLabel` might be good if you want a suffix.

server/util/clickhouse/schema/schema.go
```diff
@@ -157,6 +157,10 @@ type Execution struct {
...
+	ActionMnemonic  string
+	ConfigurationID string
```
jdhollen, line 162: Sorry, just thinking about this a bit more, I know this is also what the protocol provides, but we currently have this awkward split between linux + mac RBE minutes, and this string doesn't make it terribly easy to filter for the difference--it's specific to our cloud deployment, i guess, but maybe we could do something to distinguish these two since they bill differently?

sluongng: Discussed this with Jim on Slack and we concluded that the Mac vs Linux RBE minutes information should originate from BuildBuddy Executors, while these fields are populated from Bazel clients and thus, unrelated.

sluongng:
After talking to @bduffany and thinking a bit more about this problem: We could get this link from the Compact Exec Log (if it's enabled). At least the Target Label + mnemonic are available with a spawn digest (which I think is the same as the action digest? Need to verify). For the Configuration, we could probably scan the expected outputs path and see which Config's mnemonic was being used as the prefix.

The downside of relying on this log is that it would only be available for completed builds. Builds that are in progress or disrupted midway will not have that log file available. But perhaps that's ok for now?

Made me I wonder if there is a way for us to offload more of these invocation data into object storage as JSON blobs.

siggisim:
> After talking to @bduffany and thinking a bit more about this problem: We could get this link from the Compact Exec Log (if it's enabled). At least the Target Label + mnemonic are available with a spawn digest (which I think is the same as the action digest? Need to verify).

We won't have the compact exec log for for every invocation, what's the downside of the current approach?

jdhollen:
@siggisim @tylerwilliams @bduffany any thoughts on landing this at least for the `target` field?  seems like everyone agrees we want that specific data to exist, and we could always switch to compact exec log when it's available (and even consider using that to record info about local executions at some point)

bduffany:
I think I'm fine with this overall - I'm a little worried about storing more long string fields in the Executions table, but if it turns out to cause problems, we can revert.

bduffany:
Actually sorry, I took another look through the implementation - I'm not a fan of having to plumb all of this information through redis, could we maybe store the original `RequestMetadata` in the `Execution` proto and send it back up to the executor as part of `ExecutedActionMetadata`?

sluongng:
I think the downside of that approach is that in-progress executions will not have a label attached to them, which will make it hard to filter them on our UI? 🤔

enterprise/server/execution_service/execution_service.go
```diff
@@ -97,16 +97,60 @@ func (es *ExecutionService) GetExecution(ctx context.Context, req *espb.GetExecu
...
+				}
+			}
+		}
+	}
+
+	// Batch query for completed executions metadata from ClickHouse
+	oh := es.env.GetOLAPDBHandle()
+	ess := es.env.GetExecutionSearchService()
+	if oh != nil && ess != nil {
+		execIDToMetadata, err := ess.FetchExecutionRequestMetadata(ctx, completedExecs)
```
jdhollen, line 140: if the data is expected to be in clickhouse at this point, shouldn't the metadata be in clickhouse as well?  we're only doing one write.

jdhollen (commented):
Sorry, to be clear, I also don't want to store the new data in MySQL--i think the consensus is that it should only go into clickhouse (which is why Son took the redis approach).

I don't think there's a great way to pull `ExecutionTask` data otherwise, because we're currently flushing them when the invocation is done based on what wound up in mysql, right?  I assume this'll become easier if we actually wind up supporting updates in clickhouse.

I guess my only other concern would be about redis client traffic since this effectively adds another request for every execution?  do we have all the data necessary to write this at the time that we're initially inserting execution data into redis instead (and avoiding the extra key insert)?  or does that get cleaned up too soon?

bduffany:
> I think the downside of that approach is that in-progress executions will not have a label attached to them, which will make it hard to filter them on our UI?

This is a problem that I think we want to solve separately - we have been discussing how to store in-progress invocations in ClickHouse. I think we want to do something more general than just storing the `RequestMetadata` in Redis. We probably want to buffer the whole in-progress Execution somehow, including the other fields such as output_path etc.

> Sorry, to be clear, I also don't want to store the new data in MySQL--i think the consensus is that it should only go into clickhouse (which is why Son took the redis approach).

Yeah sorry, I had to refresh myself on this PR, I thought we were storing in MySQL for some reason.

> I don't think there's a great way to pull ExecutionTask data otherwise, because we're currently flushing them when the invocation is done based on what wound up in mysql, right? I assume this'll become easier if we actually wind up supporting updates in clickhouse.

We could plumb the RequestMetadata back to the server as part of `ExecutedActionMetadata.auxiliary_metadata`, similar to what we do for platform_overrides.

sluongng:
I rebased this on top of the latest origin/master today and resolved the conflicts.
So to confirm, the direction we wana go with this PR is:
- Remove mnemonic + config data.
- Remove redis code path (in-progress executions will not have a label attached)
- Route the metadata through Executor and back when UpdateActionResult is called(?)

Let me look into this.

sluongng:
It seems like most of the relevant metadata has already been added to ExecutionTask.RequestMetadata. Our Executor would pick it up and attach it back to `publish.Operation` rpc
https://github.com/buildbuddy-io/buildbuddy/blob/c7c2a7681c6f2424f75bc8c1e43cc55ceaba4e15/enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go#L336-L341

So the only thing left to do is to read the metadata during `recordExecution()` and add it to ClickHouse.

I will need to test this flow manually still 🤔

sluongng:
```
998c681b70a2 :) select target_label from buildbuddy_local.Executions;

SELECT target_label
FROM buildbuddy_local.Executions

Query id: d0ae7edd-abd6-42e8-9fc0-2e316b30abaf

   ┌─target_label─┐
1. │ //:a         │
   └──────────────┘
   ┌─target_label─┐
2. │ //:a         │
   └──────────────┘
   ┌─target_label─┐
3. │ //:a         │
   └──────────────┘

3 rows in set. Elapsed: 0.006 sec.
```

Tested and it worked.

@jdhollen @bduffany PTAL 🙇

sluongng:
Local config to test
```
--- a/enterprise/config/buildbuddy.local.yaml
+++ b/enterprise/config/buildbuddy.local.yaml
@@ -8,10 +8,11 @@ app:
   user_owned_keys_enabled: true
   streaming_http_enabled: true
   invocation_log_streaming_enabled: true
+  enable_write_executions_to_olap_db: true
 database:
   data_source: "sqlite3:///tmp/${USER}-buildbuddy-enterprise.db"
-#olap_database:
-#data_source: "clickhouse://default:@127.0.0.1:9000/buildbuddy_local"
+olap_database:
+  data_source: "clickhouse://default:@127.0.0.1:9000/buildbuddy_local"
 storage:
   ttl_seconds: 86400 # One day in seconds.
   disk:
```

ClickHouse setup
```
docker run --rm -d -p 8123:8123 -p9000:9000 --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server:23.8

clichouse client -n <<-EOSQL
    CREATE DATABASE IF NOT EXISTS buildbuddy_local;
EOSQL
```

Note that we need to pin to 23.8, something is failing with latest version.

sluongng:
Added a small UI change to render the target label in the Executions tab as well.
The position is consistent with the current label in the Spawns tab.

image

enterprise/server/execution_search_service/execution_search_service.go
```diff
@@ -53,6 +54,42 @@ type ExecutionWithInvocationId struct {
 	invocationID string
 }
 
+func (s *ExecutionSearchService) FetchExecutionRequestMetadata(ctx context.Context, execIDs []string) (map[string]*repb.RequestMetadata, error) {
```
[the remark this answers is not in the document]
jdhollen: i don't think it's necessary to store this in mysql, so i think i agree with (1), but more generally, can we just separate the actual recording of the value out from the creation of this service? the change to write the value out to clickhouse is actually quite small, right?

as far as ways to make this service work without being slow, we might be able to play some games with the invocation and execution's known start + end times to send a request that we are certain will fetch the right execution?  but i think we can at least get the data written while we're figuring that out.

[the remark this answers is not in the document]
bduffany: Agreed - just populating the data in ClickHouse seems like a good starting point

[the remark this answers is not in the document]
sluongng: I trimmed down the PR to only the "write to OLAP" code path plus some minor cleanups.
Yanked out the read + UI code paths completely.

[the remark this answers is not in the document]
sluongng: Let's evaluate how to read this properly in a future PR.

I think my read query could add updated_at_usec and invocation_uuid if needed 🤔 
But perhaps there could be a cleaner solution out there.

enterprise/server/execution_service/execution_service.go
```diff
@@ -97,7 +97,7 @@ func (es *ExecutionService) GetExecution(ctx context.Context, req *espb.GetExecu
 	}
 	// Sort the executions by start time.
 	sort.Slice(executions, func(i, j int) bool {
-		return executions[i].Model.CreatedAtUsec < executions[j].Model.CreatedAtUsec
+		return executions[i].CreatedAtUsec < executions[j].CreatedAtUsec
```
bduffany, line 100: nit: revert

```suggestion
		return executions[i].Model.CreatedAtUsec < executions[j].Model.CreatedAtUsec
```

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -337,52 +337,53 @@ func (s *ExecutionServer) updateExecution(ctx context.Context, executionID strin
...
 		return status.InternalErrorf("failed to get invocations for execution %q: %s", executionID, err)
 	}
-
+	rmd := bazel_request.GetRequestMetadata(ctx)
 	for _, link := range links {
 		executionProto := execution.TableExecToProto(&executionPrimaryDB, link)
-		inv, err := s.env.GetExecutionCollector().GetInvocation(ctx, link.GetInvocationId())
+		if rmd != nil {
+			executionProto.TargetLabel = rmd.GetTargetId()
+		}
```
bduffany, line 365: seems like these 2 lines are the only changes needed? (maybe revert the other changes?)

server/util/clickhouse/schema/schema.go
```diff
@@ -159,6 +159,9 @@ type Execution struct {
 
 	Stage int64
 
+	// ExecutionMetadata
```
bduffany, line 162:
```suggestion
	// RequestMetadata
```

enterprise/server/util/execution/execution.go
```diff
@@ -23,7 +23,7 @@ func TableExecToProto(in *tables.Execution, invLink *sipb.StoredInvocationLink)
 		GroupId:                            in.GroupID,
 		UpdatedAtUsec:                      in.UpdatedAtUsec,
 		ExecutionId:                        in.ExecutionID,
-		InvocationUuid:                     strings.Replace(invLink.GetInvocationId(), "-", "", -1),
+		InvocationUuid:                     strings.ReplaceAll(invLink.GetInvocationId(), "-", ""),
```
bduffany, line 26: revert this change?

bduffany (approved):
LGTM, seems there are just a few noisy diffs that could be reverted

enterprise/server/remote_execution/execution_server/execution_server.go
sluongng: Yeah, I thought I would leave the cleanup code in as they are a bit nicer to look at.

I reverted them all. Let's keep the scope of change small.

bduffany (approved):

jdhollen (approved):
we do have an existing test case for `execution_server`, so if you wanted to add some verification there, you could.

sluongng:
The existing test is more like a unit test running on top of sqlite.

To properly test this, I would need to turn it into a firecracker test (need firecracker+dockerd) and implement the read paths inside the test.

I think we could save it for the subsequent PR(s) that implement the read code path. wdyt @jdhollen?

jdhollen:
Yeah sure yes, sorry

sluongng:
Thanks everyone 🤞

PR #6561, opened 2024-05-13
Enable streaming HTTP everywhere

bduffany:
It's been enabled in dev for a while and I haven't noticed any issues with it.

**Related issues**: N/A

PR #6578, opened 2024-05-15
Allow sending gRPC status on server stream after headers are already flushed

bduffany:
The problem this PR is trying to solve is that we don't have a way to communicate gRPC errors on the stream after we've written at least one message on the stream, because at that point the header is already finalized. This means that we don't have a good way to look at the error returned by the server and decide whether it's retryable. More specifically, when we do return an error, the client just sees that the stream is "done" and can't tell whether an error has occurred or not.

According to the protocol we are currently following ([here](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)) the "right" way to send the gRPC status code is by using [HTTP Trailers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer). But unfortunately, [trailers are not supported by the Fetch API](https://github.com/whatwg/fetch/issues/981).

So instead, this PR introduces an extension to the protocol where we instead encode the gRPC status proto as a serialized message on the stream. The special length-prefix header `0x00FFFFFFFF` (i.e. `compressed=0, length=Uint32Max`) is used to denote that a status proto is about to appear. The status proto then appears as a `Length-Prefixed-Message` as defined in the original protocol.

If this feels like too much like a violation of the protocol, we could maybe require that the client explicitly sets a header asking for this extension to be enabled, and if the flag is not set then we use the more standard Trailers approach. I opted not to do that for now, since the Web app is the only client of these streaming RPCs currently and we could always add it later.

**Related issues**: N/A


siggisim:
There was some trailers parsing logic in the library that I based some of this stuff off of. I didn't implement it because I didn't think we'd need it, but - maybe we could use that?

Client side:
- https://github.com/improbable-eng/grpc-web/blob/1d9bbb09a0990bdaff0e37499570dbc7d6e58ce8/client/grpc-web/src/ChunkParser.ts#L34
- https://github.com/improbable-eng/grpc-web/blob/1d9bbb09a0990bdaff0e37499570dbc7d6e58ce8/client/grpc-web/src/client.ts#L130

Server side:
- https://github.com/improbable-eng/grpc-web/blob/1d9bbb09a0990bdaff0e37499570dbc7d6e58ce8/go/grpcweb/grpc_web_response.go#L97

bduffany:
@siggisim updated the PR to use the same format that gRPC-Web uses (more details in the PR desc). Also tested locally - PTAL

package.json
```diff
@@ -63,5 +63,8 @@
   },
   "resolutions": {
     "@types/react": "^16.8.17"
+  },
+  "dependencies": {
+    "browser-headers": "^0.4.1"
```
siggisim, line 68: I'm not sure if what we're doing is correct or not, but currently all of our deps go up in dev dependencies lol

siggisim (approved):
This stuff is a little finicky, so let's let it bake for a bit in dev before we submit https://github.com/buildbuddy-io/buildbuddy/pull/6561

package.json
bduffany: woops, moved this up. IIUC it doesn't really matter which deps section we use in our case because we aren't publishing an NPM package

bduffany:
> This stuff is a little finicky, so let's let it bake for a bit in dev before we submit https://github.com/buildbuddy-io/buildbuddy/pull/6561

will mark that PR as a draft so I don't accidentally submit :sweat_smile:

PR #6581, opened 2024-05-16
cache requests card: separate button for downloading vs. navigating

bduffany:
image

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3381

**Related issues**: N/A

### DIFF
--- a/app/invocation/cache_requests_card.tsx
+++ b/app/invocation/cache_requests_card.tsx
@@ -1,7 +1,17 @@
 import React from "react";
 import router from "../router/router";
 import InvocationModel from "./invocation_model";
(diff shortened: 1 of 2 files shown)


jdhollen (commented):
Isn't this supposed to happen already for in-app urls?

jdhollen:
Bah, replied too soon

bduffany:
> Isn't this supposed to happen already for in-app urls?

I changed the approach here (separate download button instead of nested clickable elements) but just to answer this question, we call `preventDefault()` to prevent browser navigation (not `stopPropagation()`)

siggisim (approved):

PR #6584, opened 2024-05-16
Update deps for dependabot

tempoz:
Updates deps so that we can use a secure version of `semver`

**Related issues**: https://github.com/buildbuddy-io/buildbuddy/security/dependabot/127


bduffany (approved):
nice, thanks for taking care of this :)

PR #6589, opened 2024-05-16
Accumulate kzip urls

tylerwilliams:
If kzip (kythe) URIs are present in the BES, pull them out for later processing.

### DIFF
--- a/server/build_event_protocol/accumulator/accumulator.go
+++ b/server/build_event_protocol/accumulator/accumulator.go
@@ -3,6 +3,7 @@ package accumulator
 import (
 	"context"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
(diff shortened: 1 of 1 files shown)


server/build_event_protocol/accumulator/accumulator.go
```diff
@@ -195,6 +213,10 @@ func (v *BEValues) TestOutputURIs() []*url.URL {
 	return v.testOutputURIs
 }
 
+func (v *BEValues) KZIPURIs() []*url.URL {
```
siggisim, line 216: why are you yelling?

siggisim (approved):

server/build_event_protocol/accumulator/accumulator.go
tylerwilliams: lol u like Kzip?

siggisim: no, i just think go is funny

bduffany (approved):

PR #6592, opened 2024-05-16
Add support for diffing bytestream files in editor

siggisim:
This will allow us to easily show diffs input files that changed in the UI.

### DIFF
--- a/enterprise/app/code/code.tsx
+++ b/enterprise/app/code/code.tsx
@@ -285,13 +285,12 @@ export default class CodeComponent extends React.Component<Props, State> {
 
     const bytestreamURL = this.props.search.get("bytestream_url") || "";
     const invocationID = this.props.search.get("invocation_id") || "";
-    const zip = this.props.search.get("z") || undefined;
-    let filename = this.props.search.get("filename");
-    if (this.isSingleFile() && bytestreamURL) {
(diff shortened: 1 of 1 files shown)


enterprise/app/code/code.tsx
```diff
@@ -466,6 +465,54 @@ export default class CodeComponent extends React.Component<Props, State> {
...
+    let compareFilename = this.props.search.get("compare_filename") || "";
+
+    let modelA = this.modelForBytestreamUrl(bytestreamURL, invocationID, filename, zip);
+    let modelB = this.modelForBytestreamUrl(compareBytestreamURL, compareInvocationID, "diff-" + compareFilename, zip);
+
+    Promise.all([modelA, modelB]).then((m) => {
+      if (!this.diffEditor) {
+        this.diffEditor = monaco.editor.createDiffEditor(this.diffViewer.current!);
+      }
+      let diffModel = { original: m[0], modified: m[1] };
```
bduffany, line 496: idk if this is clearer or not but you could avoid `Promise.all` and `m[0]` `m[1]`  by using await here:

```suggestion
      let diffModel = { original: await modelA, modified: await modelB };
```

bduffany (approved):

enterprise/app/code/code.tsx
siggisim: Good call, done!

PR #6594, opened 2024-05-17
action card timing: ensure the min css value

sluongng:
In #6576, we made sure that all duration in the Timeline phases are
non-negative when time drift happens.

However, because the total duration is still being calculated by taking
the difference between the first event and last event, which is
inaccurate because the first event timestamp could be after the second
event timestamp due to drift. As the total duration is shorter than it
should be, it's possible for the `offset% + event%` value to be
higher than the actual total duration (100%). This causes the last bar of
the timeline, which is calculated with `100% - (offset% + event%)`, to
have a negative value. This prevents `flex` from not being rendered in
the div's style due to invalid initial value, causing strange offset.

Apply several fixes to mitigate the issue:

- Fix total duration calculation by taking the difference between
  the smallest timestamp and the longest timestamp.

- Refactor the code to only add to offset after the flex initial values
  are calculated.

- Ensure that flex initial values cannot goes below zero.

- Fix "Timeline" still being rendered when no event is available.

Closes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3377

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -319,63 +319,81 @@ export default class InvocationActionCardComponent extends React.Component<Props
     );
   }
 
-  private renderTimelines() {
-    const metadata = this.state.actionResult?.executionMetadata;
-
+  private renderTimelines(metadata: build.bazel.remote.execution.v2.ExecutedActionMetadata) {
     type TimelineEvent = {
       name: string;
       color: string;
-      timestamp?: Timestamp | null;
+      timestamp: Timestamp | null | undefined;
     };
     const events: TimelineEvent[] = [
       {
         name: "Queued",
         color: "#3F51B5",
-        timestamp: metadata?.queuedTimestamp,
+        timestamp: metadata.queuedTimestamp,
       },
       {
         name: "Initializing",
         color: "#673AB7",
-        timestamp: metadata?.workerStartTimestamp,
+        timestamp: metadata.workerStartTimestamp,
       },
       {
         name: "Downloading inputs",
         color: "#FF6F00",
-        timestamp: metadata?.inputFetchStartTimestamp,
+        timestamp: metadata.inputFetchStartTimestamp,
       },
       {
         name: "Preparing runner",
         color: "#673AB7",
-        timestamp: metadata?.inputFetchCompletedTimestamp,
+        timestamp: metadata.inputFetchCompletedTimestamp,
       },
       {
         name: "Executing",
         color: "#1E88E5",
-        timestamp: metadata?.executionStartTimestamp,
+        timestamp: metadata.executionStartTimestamp,
       },
       {
         name: "Preparing for upload",
         color: "#673AB7",
-        timestamp: metadata?.executionCompletedTimestamp,
+        timestamp: metadata.executionCompletedTimestamp,
       },
       {
         name: "Uploading outputs",
         color: "#FF6F00",
-        timestamp: metadata?.outputUploadStartTimestamp,
+        timestamp: metadata.outputUploadStartTimestamp,
       },
       // End marker -- not actually rendered.
       {
         name: "Upload complete",
         color: "",
-        timestamp: metadata?.outputUploadCompletedTimestamp,
+        timestamp: metadata.outputUploadCompletedTimestamp,
       },
     ];
 
-    const filteredEvents = events.filter((event) => event && event.timestamp);
+    type FilteredTimelineEvent = {
+      name: string;
+      color: string;
+      timestamp: Timestamp;
+    };
+    let filteredEvents: FilteredTimelineEvent[] = events
+      .filter((event): event is FilteredTimelineEvent => event.timestamp != null)
+      .map((event, i, events) => {
+        if (i == events.length - 1) return event;
+        const next = events[i + 1];
+        // Queued event is recorded by the scheduler (not the workers)
+        // which may result in confusing timestamps due to clock drift.
+        // To reduce confusion, apply an adjustment here to enforce
+        // that all timestamps appear in monotonically increasing order
+        // when rendered as a timeline.
+        event.timestamp =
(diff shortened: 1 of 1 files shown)


bduffany (commented):
If I understand this change correctly, I think the timeline will still wind up rendering with the queued event's end timestamp potentially rendering after the worker start timestamp. I think this could still end up being confusing to users.

I think maybe a better fix would be to adjust the timestamps to make them appear as though they're monotonically increasing? i.e. something like this

```typescript
// Queued event is recorded by the scheduler (not the workers)
// which may result in confusing timestamps due to clock drift.
// To reduce confusion, apply an adjustment here to enforce
// that all timestamps appear in monotonically increasing order.
for (let i = 0; i < filteredEvents.length - 1; i++) {
  if (isAfter(filteredEvents[i].timestamp, filteredEvents[i+1].timestamp)) {
    events[i].timestamp = events[i+1].timestamp;
  }
}
```

sluongng:
Oh nice call. I was mostly focusing on the duration part of the chart and forgot that we also render out timestamp as text. And you are right, when clock drift happen, the timestamp text of Queued event will be slightly confusing because it happens after the next event.

I applied a fix.

With mock data from the invocation reported by @siggisim, the timeline would look like this

image

sluongng:
Hmm let me fix the timestamp in `Total` 🙈

sluongng:
image

Ok it looks better now.

app/invocation/invocation_action_card.tsx
```diff
@@ -319,122 +319,161 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+    type FilteredTimelineEvent = {
+      name: string;
+      color: string;
+      timestamp: Timestamp;
+    };
+    let filteredEvents: FilteredTimelineEvent[] = events.filter(
+      (event): event is FilteredTimelineEvent => event.timestamp != null
+    );
+
+    if (filteredEvents.length <= 1) {
```
bduffany, line 381: nit, why the change from `== 0` to `<= 1`? seems like if we have 1 event we could render it without any errors

app/invocation/invocation_action_card.tsx
```diff
@@ -319,122 +319,161 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+          ? next.timestamp
+          : event.timestamp;
+
+      return event;
+    });
+
+    const allUnixTimestamp = filteredEvents.map((e) => timestampToUnixSeconds(e.timestamp));
+    const startUnixTimestamp = Math.min(...allUnixTimestamp);
+    const endUnixTimestamp = Math.max(...allUnixTimestamp);
+    const totalDuration = Math.max(0, endUnixTimestamp - startUnixTimestamp);
```
bduffany, line 404: I think this could be reverted to the old logic now that the timestamps are guaranteed to be monotonically increasing

app/invocation/invocation_action_card.tsx
```diff
@@ -319,122 +319,161 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
-                  style={{ flex: `${weight} 0 0`, backgroundColor: event.color }}></div>
-                <div
-                  className="timeline-event-gray"
-                  title={`${event.name} (${format.durationSec(duration)}, ${(weight * 100).toFixed(2)}%)`}
-                  style={{ flex: `${1 - offset} 0 0`, backgroundColor: `rgba(0, 0, 0, .1)` }}></div>
+      <>
+        <div className="metadata-title">Timeline</div>
+        <div>
+          <div className="metadata-detail">
+            <span className="label">Total @ {format.formatTimestamp(filteredEvents[0].timestamp)}</span>
```
bduffany, line 412: nit: The `Total @ <timestamp>` feels a bit awkward, maybe just make the label empty. We show the start timestamp for the queued event so I think this is redundant anyway

```suggestion
            <span className="label"></span>
```

app/invocation/invocation_action_card.tsx
```diff
@@ -319,122 +319,161 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
+            // Actual flex initial value.
+            //
+            // Time drift between app and executor could cause negative duration,
+            // which is not a valid flex initial value and causes broken render.
+            // Make sure to enforce a minimum value of zero for "flex" style so
+            // it is always included in the final render result.
+            const leftBar = Math.max(0, offset);
+            const middleBar = Math.max(0, weight);
+            const rightBar = Math.max(0, 1 - leftBar - middleBar);
+
```
bduffany, line 446: I think all of these changes can be reverted now?

bduffany (approved):

app/invocation/invocation_action_card.tsx
sluongng: Make sense to revert the top hunk here.

But I will leave the leftBar -> rightBar hunk in so that these are calculated before adding up the offset, which is a bit easier to read.

app/invocation/invocation_action_card.tsx
sluongng: I will remove the timestamp for now. It's a bit weird to just see a duration on the right column without a label on the left.

app/invocation/invocation_action_card.tsx
sluongng: Will revert to old logic

app/invocation/invocation_action_card.tsx
sluongng: I don't think I did it intentionally. Probably was testing out something with the mock data. Will revert.

sluongng:
Latest preview 
image

bduffany (approved):

app/invocation/invocation_action_card.tsx
bduffany: Ah yeah sorry, that's what I meant :D

sluongng:
Pushed a minor style fix. Tested locally with mock data and nothing changed.

PR #6597, opened 2024-05-17
Don't publish UNKNOWN ExecutionStage

bduffany:
In practice, we were publishing UNKNOWN as the initial status instead of QUEUED. Publish QUEUED instead.

**Related issues**: N/A


iain-macdonald (approved):

siggisim (approved):
No reticulating splines!?

siggisim:
> > No reticulating splines!?
> 
> 
> 
> That's a UI concern, separation of concerns and all that :P

Haha yeah I was just worried if we never published Unknown that we'd never have the opportunity to show it. 

I'm cool with this though, mostly just joking.

bduffany:
ah true, all of our statuses are too unambiguous now. no fun :P

PR #6598, opened 2024-05-17
list out all months since group creation when on usage page.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/usage_service/usage_service.go
+++ b/enterprise/server/usage_service/usage_service.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
+	"github.com/buildbuddy-io/buildbuddy/server/tables"
 	"github.com/buildbuddy-io/buildbuddy/server/util/authutil"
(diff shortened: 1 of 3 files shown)


enterprise/server/usage_service/usage_service.go
```diff
@@ -106,6 +98,20 @@ func (s *usageService) GetUsage(ctx context.Context, req *usagepb.GetUsageReques
 	return rsp, nil
 }
 
+func (s *usageService) GetUsage(ctx context.Context, req *usagepb.GetUsageRequest) (*usagepb.GetUsageResponse, error) {
+	groupID := req.GetRequestContext().GetGroupId()
+	if err := authutil.AuthorizeGroupAccess(ctx, s.env, groupID); err != nil {
+		return nil, err
+	}
```
bduffany, line 105: nit: the auth middleware is aware of the preferred group ID in the request context, so you could instead write `env.GetAuthenticator().GetAuthenticatedUser()` then `u.GetGroupID()`

bduffany (approved):

enterprise/server/usage_service/usage_service.go
jdhollen: sure sure--this was just a move of what was there before, but happy to update it

PR #6600, opened 2024-05-17
Refactor container provider args

bduffany:
Refactor container provider args so that we can add new optional params without every container implementation having to handle the new params. (I want to add `*operation.Publisher` to these args, but only Firecracker will be using it). 

These args are all truly optional so a struct seems better than using required positional args here (which force callers to consider the new args). For example, the Bare container implementation doesn't look at any of these struct fields.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/container/BUILD
+++ b/enterprise/server/remote_execution/container/BUILD
@@ -7,6 +7,7 @@ go_library(
     srcs = ["container.go"],
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/container",
     deps = [
+        "//enterprise/server/remote_execution/operation",
         "//enterprise/server/remote_execution/platform",
         "//enterprise/server/util/oci",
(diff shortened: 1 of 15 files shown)


enterprise/server/remote_execution/operation/operation.go
```diff
@@ -68,6 +68,8 @@ func (p *Publisher) Send(op *longrunning.Operation) error {
 
 // Ping re-publishes the current execution progress state.
 func (p *Publisher) Ping() error {
+	time.Sleep(750 * time.Millisecond)
```
iain-macdonald, line 71: did you mean to include this change in this pr?

iain-macdonald (commented):
I wrote this review in the BuildBuddy code review UI alpha!

enterprise/server/remote_execution/operation/operation.go
bduffany: Woops, removed

iain-macdonald (approved):

PR #6603, opened 2024-05-20
Add Task struct to hold operation publisher and make containers more easily extensible

bduffany:
In https://github.com/buildbuddy-io/buildbuddy/pull/6600 I forgot that `provider.New()` is only called when we initially create the container. So actually, we need the `*operation.Publisher` to be passed to `Run`, `Create`, `Exec`, and `Unpause`.

I felt like adding `*operation.Publisher` as a "required" argument to every container implementation seemed a bit overkill, since we really only want to use the `Publisher` for Firecracker. To make this type of change less laborious in the future, I think it would be nice to have something like a "task env" that applies just to the task, and contains various utilities / metadata associated with a single task.

I have been thinking about doing something like this for a while, for other reasons:
* Testing is becoming more of a chore because whenever we add new optional params to each container impl, we have to update every test. This happened with `oci.Credentials` previously as well as the `container.Provider` refactor. All tests today have to pass around empty `oci.Credentials` which makes the tests more noisy and more work to maintain.
* We are currently getting access to the parsed `platform.Props` in sort of an ugly way - `provider.New()` takes in the props and then is expected to store them internally and then use them to decide e.g. which container image to pull. I think it would be clearer if we instead passed in the `platform.Props` to `Run()` / `Pull()` etc. so that `provider.New()` could just be a minimal implementation that just returns a struct and does nothing else.
* It's a bit awkward that some container implementations just store the `workDir` in `Create()` and do nothing else with the work dir until `Exec()` is called. With this new proposed interface, each method could just inspect `task.WorkDir` as needed.

Sending this PR out as a draft to get some initial feedback on this idea before actually updating all the container implementations.

**Related issues**: N/A


enterprise/server/remote_execution/container/container.go
```diff
@@ -208,35 +239,39 @@ type CommandContainer interface {
...
+	Exec(ctx context.Context, task *Task, stdio *interfaces.Stdio) *interfaces.CommandResult
+
 	// Unpause un-freezes a container so that it can be used to execute commands.
-	Unpause(ctx context.Context) error
+	Unpause(ctx context.Context, task *Task) error
+
 	// Pause freezes a container so that it no longer consumes CPU resources.
-	Pause(ctx context.Context) error
+	Pause(ctx context.Context, task *Task) error
+
```
tylerwilliams, line 274: it doesn't seem like Task should be a required argument here.

bduffany: Removed

enterprise/server/remote_execution/container/container.go
```diff
@@ -84,7 +84,7 @@ type Provider interface {
 // It's not strictly necessary to look at all of these params in order to create
 // a working container implementation, but they might be useful.
 type Init struct {
-	// WorkDir is the working directory for the initially assigned task.
+	// WorkDir is the working directory for the *initially* assigned task.
 	WorkDir string
```
iain-macdonald, line 88: Can we eliminate this parameter in favor ofTask.WorkDir?

enterprise/server/remote_execution/container/container.go
```diff
@@ -208,35 +239,39 @@ type CommandContainer interface {
...
+
 	// Exec runs a command inside a container, with the same working dir set when
 	// creating the container.
 	//
 	// If stdin is non-nil, the contents of stdin reader will be piped to the
 	// stdin of the executed process. If stdout is non-nil, the stdout of the
 	// executed process will be written to the stdout writer rather than being
 	// written to the command result's stdout field (same for stderr).
-	Exec(ctx context.Context, command *repb.Command, stdio *interfaces.Stdio) *interfaces.CommandResult
+	Exec(ctx context.Context, task *Task, stdio *interfaces.Stdio) *interfaces.CommandResult
```
iain-macdonald, line 267: Move stdio to Task?

iain-macdonald:
I think this is a nice simplification and I agree with the points you outlined in the PR description, just have a couple comments on fields in Init and Task. But I think it's worth pursuing this once we get those ironed out!

enterprise/server/remote_execution/container/container.go
maggie-lou: In general - could we remove these fields that apply "to the initially assigned task" and rely on the Task struct passed into each specific function? It feels confusing and error prone to persist this "initial state"

bduffany: Yeah, I meant to add a TODO here to remove `Init` in favor of having `Run()` / `Create()` just look at the task struct - Done. At the end of the day, `provider.New()` should just be returning an empty struct, and the lifecycle methods should actually be responsible for initialization.

enterprise/server/remote_execution/container/container.go
bduffany: I thought it might be a bit confusing to move it to `Task` because `Run` doesn't respect stdio currently (`stdio` was originally added to support persistent workers).

I do think it'd be a cleaner interface if both `Run` and `Exec` respected stdio though - it would let us simplify our implementation of `--debug_stream_command_outputs` as well as make it simpler to implement live-streaming of stdout/stderr to the UI. But I am leaning toward waiting to move this into `Task` until we actually make that refactor.

PR #6608, opened 2024-05-20
server-side code for a generic dimension (worker, invocation id, etc.) filter.

jdhollen:
i will use this to do drilldowns by invocation id and worker id + it'll make it easier to add others in the future.

**Related issues**: N/A


server/util/filter/filter.go
```diff
@@ -55,6 +55,24 @@ func invocationMetricToDbField(m stat_filter.InvocationMetricType, paramPrefix s
 	}
 }
 
+func executionDimensionToDbField(m stat_filter.ExecutionDimensionType, paramPrefix string) (string, error) {
```
bduffany, line 58: nit: maybe `s/paramPrefix/tablePrefix/` ?

bduffany (approved):

server/util/filter/filter.go
jdhollen: Sure yeah, done

PR #6613, opened 2024-05-20
Live-stream invocation logs

bduffany:
- Publish a message on a pubsub channel (in Redis) whenever logs are updated
- Have clients read logs using the same 3s polling logic as today, except the polling happens on the server-side, and the server also subscribes to the pubsub channel (if available) to eagerly re-fetch the tail as soon as it is updated. The pubsub channel should be available except on OSS (for now) and possibly also during Redis restarts.

Demo (local terminal on right, BuildBuddy invocation page on left - note how the UI updates ~immediately)

https://github.com/buildbuddy-io/buildbuddy/assets/2414826/85056461-dcfb-43e0-b7c9-344531227076

**Related issues**: N/A

### DIFF
--- a/server/buildbuddy_server/buildbuddy_server.go
+++ b/server/buildbuddy_server/buildbuddy_server.go
@@ -13,6 +13,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/backends/chunkstore"
 	"github.com/buildbuddy-io/buildbuddy/server/build_event_protocol/build_event_handler"
@@ -40,6 +41,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/role"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/subdomain"
+	"golang.org/x/time/rate"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/encoding/protojson"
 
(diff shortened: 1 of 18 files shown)


server/eventlog/eventlog.go
```diff
@@ -341,6 +350,9 @@ func (w *EventLogWriter) writeChunkToKeyValStore(ctx context.Context, writeReque
 		curChunk,
 	)
 	w.lastChunk = curChunk
+	if w.pubsub != nil {
+		w.pubsub.Publish(ctx, w.pubsubChannel, "")
```
jdhollen, line 354: is my understanding right that this dangles the pubsub stream value as an empty string for 8 hours? if so, is there a way to just kill the stream here instead?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1158,6 +1160,63 @@ func (s *BuildBuddyServer) GetEventLogChunk(ctx context.Context, req *elpb.GetEv
 	return resp, err
 }
 
+func (s *BuildBuddyServer) GetEventLog(req *elpb.GetEventLogChunkRequest, stream bbspb.BuildBuddyService_GetEventLogServer) error {
+	ctx := stream.Context()
+	// Fetch the event log once as soon as we get the request, once whenever
+	// we see an update from Redis, and once every 3s (as a fallback).
```
jdhollen, line 1166: if we're still willing to do every 3s, is it really worth tracking in redis?  if we don't track anything in redis, we're only going to do extra work when someone is actually looking at the invocation vs. the pubsub approach, which writes to redis no matter what, right?

bduffany: Not sure I fully got your question, but these Redis pubsub writes/reads should be extremely cheap (relative to remote execution, and also relative to the existing invocation log tail that is stored in Redis)

The 3s poll is just a fallback in case Redis is unavailable for any reason (e.g. it restarts, or for OSS instances). It'll also let us get rid of the other log streaming endpoint in favor of just this one RPC.

server/eventlog/eventlog.go
bduffany: We don't want to kill the stream here, since this is where the log is being updated - the empty string is just a signal for clients to poll the tail of the log. The empty string will dangle for 8h but I think it won't take up enough memory to be worth worrying about

siggisim (approved):

server/buildbuddy_server/buildbuddy_server.go
jdhollen: Sorry, all I was really asking here was why bother doing pubsub if you're going to do a 3s fallback anyway?

jdhollen (approved):

server/buildbuddy_server/buildbuddy_server.go
bduffany: I think I answered your question but just in case :sweat_smile: 
* The "happy" path is that this is a BB cloud instance and Redis is 100% available - in this case we'll only be reading from the pubsub chan in practice, since bazel sends updates much more frequently than once per 3s
* The "sad" path is either this is a self-hosted BB instance, or this is a cloud instance and redis restarted or something. In this case we do the 3s fallback so that we're updating the logs at least as frequently as we do today

PR #6616, opened 2024-05-21
cachetools: add err check for Seek reset

sluongng:
also removed the interface type cast and use `rsc` directly as the constraint is only enforced by downstream function's params.


sluongng:
too out of date

PR #6618, opened 2024-05-21
dirtools: refactor UploadTree to use filepath.WalkDir instead of recursion

sluongng:
Replace the complicated recursion setup with a more straightforward filepath.WalkDir implementation. Files are now eagerly queued for upload as the traversal happens, while directories and trees are queued after the tree traversal to ensure the Merkle tree is computed correctly.

This comes with a side effect: the ordering of "repeated" fields in remote execution protos, such as Tree.Children or Directory.Files, could change. The spec recommends that we should sort these fields using lexical ordering wherever possible, or a stable ordering in other cases. Effectively, we see that the Tree object in TestUploadTree/LotsOfNesting needed to be updated because of the ordering in Tree.Root.Directory and Tree.Children were updated. For this reason, added stable sorting on the relevant slices to make sure that future updates are backward compatible.


sluongng:
way outdated

PR #6620, opened 2024-05-21
Add usage column for memory usage

bduffany:
Went with GB-usec as the units which will let us represent 2.5 million GB-hours for each usage row, which seems plenty.

**Related issues**: N/A


tylerwilliams:
Could we match what gcp does for these units?

bduffany:
> Could we match what gcp does for these units?

Looks like GCP uses GB-hours as their units: https://cloud.google.com/compute/all-pricing

We are using GB for the memory part of the metric, but need to use a more granular unit like usec or nanos for the duration part, because the usage table only supports int64 values currently (not float64). Regardless of what units we use here, we can always change how we present them to the user if needed.

tylerwilliams (approved):

PR #6621, opened 2024-05-21
Add a very simple workflow that the probers can use to test workflow health

maggie-lou:
**Related issues**: Used in this script: https://github.com/buildbuddy-io/buildbuddy/pull/6628

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -110,6 +110,10 @@ actions:
           - "*"
     bazel_commands:
       - test //... --host_platform=//:many_layered_image --config=linux-workflows --config=race --test_tag_filters=-performance,-webdriver,-docker,-bare
+  - name: Prober test
+    container_image: ubuntu-20.04
+    bazel_commands:
+      - version
 
 plugins:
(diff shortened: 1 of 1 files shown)


bduffany (commented):
Could you put this in a separate private repo like `buildbuddy-io/buildbuddy-workflow-probers` and register that repo to the probers org(s) in dev/prod? Just would like to avoid adding noise to the GitHub statuses published to commits in the OSS repo, as well as our build history in the BB org.

maggie-lou:
@bduffany Re-opening because we can only link a github app to 1 BB org. I considered using remote bazel in the prober instead, but we'd have to clone a git repo, which feels too heavy weight for a prober.

bduffany:
I just added `buildbuddy-io/probers` to the repos allowed by the app installation and linked it to the BB org, so you could try testing that out?

maggie-lou:
> I just added `buildbuddy-io/probers` to the repos allowed by the app installation and linked it to the BB org, so you could try testing that out?

Oh duh you're right. That works

PR #6629, opened 2024-05-22
event_index: drop Fetch event

sluongng:
Folks who are interested in Fetch information don't simply want a binary
success/fail status, but they also want to check the execution of the
related Bazel repository rules as a whole.

Since that information is often available via the Timing profile with
great resolution, displaying this event in RAW tab is a waste of space.

### DIFF
--- a/server/build_event_protocol/event_index/event_index.go
+++ b/server/build_event_protocol/event_index/event_index.go
@@ -171,6 +171,9 @@ func (idx *Index) Add(event *inpb.InvocationEvent) {
 	case *bespb.BuildEvent_Progress:
 		// Drop progress events
 		return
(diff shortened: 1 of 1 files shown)


siggisim:
We have a fetch UI that shows if you click on the fetch count in the header of an invocation. 

If we want to remove these events, we'd need to move that to a server side endpoint.

siggisim (changes requested):
See comment

sluongng:
Oh wow, I have never clicked that button before 🙈

I imagine we could move the Fetch count metric into the Invocation message, but we probably gonna lose out on the details of all the fetched URLs. Closing this for now.

PR #6630, opened 2024-05-22
Add a prometheus metric tracking time saved due to action merging.

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/action_merger/action_merger.go
+++ b/enterprise/server/remote_execution/action_merger/action_merger.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
@@ -22,6 +23,11 @@ const (
 	// set to twice the length of `remote_execution.lease_duration` to give a
 	// short grace period in the event of missed leases.
 	DefaultClaimedExecutionTTL = 20 * time.Second
+
+	// TTL for redis entries tracking the time at which an execution began.
+	// This is used to calculate the amount of time saved due to action merging
+	// for metrics, and is not critical to action-merging functionality.
+	executionStartTimeTTL = 8 * time.Hour
 )
 
 var (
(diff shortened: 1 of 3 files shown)


tylerwilliams:
This looks OK to me but I'm going to let @vadimberezniker provide the final stamp.

bduffany (commented):
I feel like the wall time metric can be misleading because wall time goes up when the system is under higher load. So under higher load it will appear as though we're saving more time due to action merging.

Maybe we could use CPU time instead?

iain-macdonald:
> I think wall time can be misleading because wall time goes up when the system is under higher load. So under higher load it will appear as though we're saving more time due to action merging when in reality we're actually just taking longer to run actions.
> 
> Maybe we could use CPU time instead? We have access to this in the ExecutedActionMetadata so I wonder if when we get the response on the execution_server, we could record a metric observation equal to the number of merged actions * CPU nanos used?

I'd like a metric tracking the amount of time saved from the user's perspective because in the case you described (long queue times + heavy load), successfully merged actions are saving the queue and execution time prior to the second action's submission from the user's perspective. That said, this implementation isn't 100% correct as it measures the first action's queue time and assumes we would save that much queue time for the second action. I figured that was a reasonable approximation to make given that it's simpler to implement.

We could add another metric tracking CPU, or other resources saved as well.

bduffany:
> That said, this implementation isn't 100% correct as it measures the first action's queue time and assumes we would save that much queue time for the second action. I figured that was a reasonable approximation to make given that it's simpler to implement.

I think there are a lot of scenarios that would negatively affect the accuracy of this approximation - for example, the second action could be routed to a new executor that is more lightly loaded than the first, and wind up executing more quickly. So we're actually costing time instead of saving time.

Thinking a bit more, CPU savings is actually problematic too, because it assumes that each action execution is going to use roughly the same amount of CPU, when in reality, CPU usage increases under load due to context switching.

Given these downsides I'm not actually sure how reliable these metrics would be :thinking: but maybe others have thoughts/opinions.

PR #6633, opened 2024-05-22
hacked up dimensions client code

jdhollen:
**Related issues**: N/A

### DIFF
--- a/enterprise/app/filter/filter.tsx
+++ b/enterprise/app/filter/filter.tsx
@@ -26,6 +26,7 @@ import Radio from "../../../app/components/radio/radio";
 import { compactDurationSec, formatDateRange } from "../../../app/format/format";
 import router from "../../../app/router/router";
 import {
+  DIMENSION_PARAM_NAME,
   START_DATE_PARAM_NAME,
   END_DATE_PARAM_NAME,
   ROLE_PARAM_NAME,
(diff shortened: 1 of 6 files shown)


enterprise/app/filter/filter.tsx
```diff
@@ -456,6 +463,11 @@ export default class FilterComponent extends React.Component<FilterProps, State>
                 {compactDurationSec(Number(maximumDurationValue))}
               </span>
             )}
+            {dimensions.map((v) => (
+              <span className="advanced-badge" title={getDimensionName(v.dimension ?? undefined)}>
+                {getDimensionIcon(v.dimension ?? undefined)} {v.value}
+              </span>
```
bduffany, line 469: nit: would it make sense to do `v.dimension && <span ...>` instead of the `?? undefined` stuff?

enterprise/app/filter/filter_util.tsx
```diff
@@ -77,6 +79,60 @@ function splitAndTrimTags(param: string | null): string[] {
...
+    });
+  } else if (type == "i" && Object.values(stat_filter.InvocationDimensionType).find((v) => v === enumValue)) {
+    return new stat_filter.Dimension({
+      invocation: enumValue,
+    });
+  }
+  return undefined;
+}
+
+export function getFiltersFromDimensionParam(dimensionParamValue: string): stat_filter.DimensionFilter[] {
```
bduffany, line 103: nit: mind adding a comment with an example input string that this is parsing and what it returns?

bduffany (approved):

enterprise/app/filter/filter_util.tsx
jdhollen: yup, done

enterprise/app/filter/filter.tsx
jdhollen: sure yeah, done

PR #6634, opened 2024-05-23
Dedupe large (bytestreamed) in-flight executor downloads

tylerwilliams:


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -769,41 +810,51 @@ func (ff *BatchFileFetcher) bytestreamReadFiles(ctx context.Context, instanceNam
...
+		if err := f.Close(); err != nil {
+			return nil, err
+		}
+		fileCache := ff.env.GetFileCache()
+		if fileCache != nil {
+			if err := fileCache.AddFile(ff.ctx, fp0.FileNode, fp0.FullPath); err != nil {
+				log.Warningf("Error adding file to filecache: %s", err)
+			}
 		}
+		return fp0, nil
```
jdhollen, line 845: the `groups` map grows unbounded, right? is it possible to clean it up here given that subsequent callers will hit the filecache?  this is a genuine question, i imagine there's some extra squirrelly crap around whether or not some code has already missed the filecache and is about to use the singleflight object?

tylerwilliams: I simplified things a little so it's now just one entry per GroupID, but yes, it does.

Given our current usage this should not be a substantial use of memory pretty much ever. I thought a little about cleaning that per-groupID entry up, but it's not really simple: we'd need to keep some tracking info around that duplicates a lot of the state inside the singleflight library, and then somehow delete entries when there are no active singleflight requests, and it ends up being a lot of synchronization.

I did also look into whether or not the singleflight's internal map will grow without bound (one entry per digest), but entries are [deleted](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.7.0:singleflight/singleflight.go;l=160) after the Do() func returns, so I don't think this should be an issue. (If it is we'll see it in dev immediately)

jdhollen: ah cool, this definitely seems like plenty

jdhollen (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -769,41 +817,54 @@ func (ff *BatchFileFetcher) bytestreamReadFiles(ctx context.Context, instanceNam
...
+		}
 
-	ff.statsMu.Lock()
-	ff.stats.FileDownloadSizeBytes += d.GetSizeBytes()
-	ff.stats.FileDownloadCount += 1
-	ff.statsMu.Unlock()
+		ff.statsMu.Lock()
+		ff.stats.FileDownloadSizeBytes += d.GetSizeBytes()
+		ff.stats.FileDownloadCount += 1
+		ff.statsMu.Unlock()
```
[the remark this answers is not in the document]
bduffany: hmm I think since this is being done inside the `group.Do()`, it will only be attributed to the first file fetcher that initiated the request?

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -769,41 +817,54 @@ func (ff *BatchFileFetcher) bytestreamReadFiles(ctx context.Context, instanceNam
...
+		}
+		f, err := os.OpenFile(fp0.FullPath, os.O_RDWR|os.O_CREATE, mode)
+		if err != nil {
+			return nil, err
+		}
+		resourceName := digest.NewResourceName(fp0.FileNode.Digest, instanceName, rspb.CacheType_CAS, ff.digestFunction)
+		if ff.supportsCompression() {
+			resourceName.SetCompressor(repb.Compressor_ZSTD)
+		}
+		if err := cachetools.GetBlob(ctx, bsClient, resourceName, f); err != nil {
```
bduffany, line 834: any idea what would happen to the other actions if the first action to reach the group here gets canceled?

enterprise/server/remote_execution/dirtools/dirtools.go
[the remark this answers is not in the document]
sluongng: You are right. The test added verified for FileCount == 1 as well.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -769,41 +817,54 @@ func (ff *BatchFileFetcher) bytestreamReadFiles(ctx context.Context, instanceNam
...
-		return err
-	}
-	resourceName := digest.NewResourceName(fp.FileNode.Digest, instanceName, rspb.CacheType_CAS, ff.digestFunction)
-	if ff.supportsCompression() {
-		resourceName.SetCompressor(repb.Compressor_ZSTD)
-	}
-	if err := cachetools.GetBlob(ctx, bsClient, resourceName, f); err != nil {
-		return err
-	}
+	fpInterface, err, _ := ff.group.Do(d.GetHash(), func() (interface{}, error) {
```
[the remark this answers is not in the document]
tylerwilliams: sure, good call.

enterprise/server/remote_execution/dirtools/dirtools.go
tylerwilliams: Sure, done.

bduffany (approved):

PR #6638, opened 2024-05-23
Use bazelisk in all GitHub workflows

fmeum:
**Related issues**: N/A


sluongng (approved):
This LGTM, but since some of these workflows don't run on PR, I would recommend playing around with the `on: pull_request:`, as well as commenting out any actual release uploads, to verify that the commands are still working as intended before merging.

PR #6648, opened 2024-05-24
Remove podman overhead when rapidly recycling runners

bduffany:
(Context: https://buildbuddy-corp.slack.com/archives/C057TAUAQ7P/p1716579465393109?thread_ts=1716243012.902439&cid=C057TAUAQ7P)

When runner recycling is enabled, we currently have significant overhead introduced by podman due to running `podman unpause`, `podman exec`, and `podman pause` for each execution. `pause` and `unpause` each add about 20ms of overhead, and `exec` adds 60-100ms of overhead.

This PR adds 2 optimizations to eliminate the overhead from these podman commands, particularly in the case where the same runner is being reused at a high frequency:

* **Delayed pause**: instead of immediately pausing after execution is done, wait until the runner has idled for a few seconds (the idle duration is configurable via flag). This avoids unnecessary pause/unpause cycles in the case where several tasks are queued up and are eligible to reuse the same runner. In practice, runners likely aren't using significant CPU while idle, so we'll likely be burning very few CPU cycles by keeping the runners alive while they are not assigned a task. However, a future improvement could be to base the pause delay on how much CPU the runner is trying to use while idle (well-behaved runners should be using ~0 CPU while idle, and so they don't really need to be paused at all).
* **Exec server**: instead of using `podman exec`, run a gRPC server inside the container and send commands to it over a socket (we can use the `vmexec` server implementation which we already use to send commands to Firecracker VMs).

Results:
* **Benchmark: sequential NOP actions**: With sequential NOP tasks run locally with `recycle-runner=true`, this optimization increases throughput from 7.5 actions/s => 79.3 actions/s (about 10X increase in throughput).
* **Benchmark: build //server**: When building `//server` with `recycle-runner=true` set on all actions and bypassing cache, this reduces average (N=24) build time from 62s to 53s (~15% reduction) and p90 from 63s to 54s. In practice, not all actions have `recycle-runner=true` but we have some customers that are heavily relying on it - so we should be able to execute these actions at a higher rate, reducing the need for autoscaling and burning less CPU on `podman` commands.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/runner_test.go
+++ b/enterprise/server/remote_execution/runner/runner_test.go
@@ -703,6 +703,10 @@ func TestRunnerPool_DelayedPause_PausedInBackground(t *testing.T) {
 
 	// Should be paused after a short delay.
 	<-pauseCh
+
+	// Make sure we can get the same runner again, unpausing it.
+	r2 := mustGetPausedRunner(t, ctx, pool, newTask())
+	assert.Same(t, r1, r2, "should get the same pooled runner")
 }
 
 func newPersistentRunnerTask(t *testing.T, key, arg, protocol string, resp *wkpb.WorkResponse) *repb.ScheduledTask {
--- a/enterprise/server/remote_execution/execserver/execserver.go
+++ b/enterprise/server/remote_execution/execserver/execserver.go
@@ -1,3 +1,11 @@
+// execserver is a little program that can be run inside a container to accept
+// commands from the host. Its purpose is to provide an alternative to things
+// like 'docker exec' or 'podman exec' which have significant overhead.
+//
+// It listens on a unix socket, which should be mounted into the container from
+// the host. It exposes the Exec service, which can be called using the
+// vmexec_client package.
+
 package main
 
 import (
@@ -14,7 +22,7 @@ import (
 )
 
 var (
-	socket = flag.String("socket", "", "Listen path")
+	socket = flag.String("socket", "", "Unix socket path for serving the exec service")
 )
 
 func main() {


enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -125,6 +143,15 @@ type Provider struct {
 }
 
 func NewProvider(env environment.Env, buildRoot string) (*Provider, error) {
+	// Unpack embedded execserver to buildRoot.
```
maggie-lou, line 146: Gate behind the flag?

enterprise/server/remote_execution/containers/podman/podman.go
```diff
@@ -62,6 +69,7 @@ var (
 	podmanRuntime       = flag.String("executor.podman.runtime", "", "Enables running podman with other runtimes, like gVisor (runsc).")
 	podmanStorageDriver = flag.String("executor.podman.storage_driver", "overlay", "The podman storage driver to use.")
 	podmanEnableStats   = flag.Bool("executor.podman.enable_stats", true, "Whether to enable cgroup-based podman stats.")
+	enableExecServer    = flag.Bool("executor.podman.enable_exec_server", false, "Whether to use an execution server inside each podman container for executing commands, rather than using 'podman exec'. Applies only to actions with runner recycling enabled.")
```
maggie-lou, line 72: From the code it seems like this applies to all actions and not just those with runner recycling enabled?

Also why does it matter if runner recycling is enabled? If this is faster, should we always use it?

enterprise/server/remote_execution/runner/runner_test.go
```diff
@@ -643,6 +662,49 @@ func TestRunnerPool_TaskSize(t *testing.T) {
...
+	pool := newRunnerPool(t, env, &opts)
+	ctx := withAuthenticatedUser(t, context.Background(), env, "US1")
+
+	r1 := mustGetNewRunner(t, ctx, pool, newTask())
+	pauseCh := r1.Container.Delegate.(*fakeContainer).Paused
+
+	mustAddWithoutEviction(t, ctx, pool, r1)
+
+	// Should be paused after a short delay.
+	<-pauseCh
```
maggie-lou, line 705: Add `r2 := mustGetPausedRunner(t, ctx, pool, newTask())` to test that even if the runner is paused in the background, we can still successfully unpause it?

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -721,9 +745,68 @@ func (p *pool) add(ctx context.Context, r *taskRunner) *labeledError {
 	metrics.RunnerPoolMemoryUsageBytes.Add(float64(r.memoryUsageBytes))
 	metrics.RunnerPoolCount.Inc()
 
-	// Officially mark this runner paused and ready for reuse.
+	// Mark this runner ready for reuse.
 	r.state = paused
 
+	// If the pause is delayed, schedule that now.
+	if delay := r.pauseDelay(); delay > 0 {
```
maggie-lou, line 752: Why do we pause the runner (if there's no delay) / initiate the pause delay at different times? Is the timing important?

maggie-lou (approved):
Looks good to me but I'd have @iain-macdonald also approve because he's more familiar with podman behavior

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: In the recycling case, we are replacing the `podman exec` command with a gRPC call to the exec server, which is faster. In the non-recycling case, we are using `podman run` to create the container and execute the command in a single podman command. I wouldn't expect the exec server to speed up `podman run`, since most of the overhead in `podman run` is in creating and starting the container.

enterprise/server/remote_execution/containers/podman/podman.go
bduffany: I can add the flag guard later - for now this makes it easier to test, since we can turn on this optimization by setting the `debug-use-exec-server` prop even if the flag is not enabled for all actions.

enterprise/server/remote_execution/runner/runner.go
bduffany: We need to have the mutex held here because this is mutating `r.unpause`. The current call to `Pause()` doesn't have the mutex held. The timing doesn't really matter as long as the mutex is held - I just chose to put the logic here because this is close to where we're doing other runner state mutations (`r.state = paused`).

enterprise/server/remote_execution/runner/runner_test.go
bduffany: good call - done

iain-macdonald (changes requested):
This is pretty clever! That said, I'm not too excited to add this degree of complexity to our podman implementation and have been wondering if we could alleviate this performance overhead by investing in just running crun (or, performance permitting, runsc) directly. What would you think about looking into that instead? Happy to help, of course!

bduffany:
@iain-macdonald  I can try to put together a small prototype and get some perf numbers for `crun pause`, `crun unpause`, `crun exec`. A practical implementation of `crun` is going to be some effort - if you want to help scope it out and make a list of the parts that we'd need to implement, that would be very useful :) 

I think we need to figure out
* rootfs (with and without soci-snapshotter)
* networking (though we could add this later, and only allow using this impl if `dockerNetwork=none`)
* namespaces
* cgroups
* seccomp
* apparmor
* capabilities
* `podman logs` equivalent (for debugging) - probably just a log file
* Other stuff?

PR #6654, opened 2024-05-28
use regional service to stream execution data.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -280,7 +280,11 @@ export default class InvocationComponent extends React.Component<Props, State> {
     const runnerExecution = this.state.runnerExecution;
     if (!runnerExecution?.executionId) return;
 
-    this.runnerExecutionStream = waitExecution(runnerExecution.executionId, {
+    const service = rpcService.getRegionalServiceOrDefault(
+      this.state.model?.stringCommandLineOption("remote_cache") ?? ""
+    );
+
+    this.runnerExecutionStream = waitExecution(service, runnerExecution.executionId, {
       next: (op) => {
         this.setState({ runnerLastExecuteOperation: op });
         if (op.response) {
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -280,8 +280,12 @@ export default class InvocationComponent extends React.Component<Props, State> {
     const runnerExecution = this.state.runnerExecution;
     if (!runnerExecution?.executionId) return;
 
+    // We technically know that we won't have the invocation at this point,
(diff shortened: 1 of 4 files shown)


app/invocation/invocation.tsx
```diff
@@ -280,7 +280,11 @@ export default class InvocationComponent extends React.Component<Props, State> {
     const runnerExecution = this.state.runnerExecution;
     if (!runnerExecution?.executionId) return;
 
-    this.runnerExecutionStream = waitExecution(runnerExecution.executionId, {
+    const service = rpcService.getRegionalServiceOrDefault(
+      this.state.model?.stringCommandLineOption("remote_cache") ?? ""
```
bduffany, line 284: We won't have an invocation model at this point unfortunately, because the runner execution itself is responsible for creating the invocation and we stream the execution status just so that we can see the runner execution progress before the invocation gets created. Once we have `this.state.model` we render a nicer page.

In practice we can always use the default service because workflows are only enabled in us-west1 (for now)

bduffany (approved):

app/invocation/invocation.tsx
bduffany: (I think it's fine to leave this code actually - probably just need to change remote_cache to remote_executor though)

app/invocation/invocation_action_card.tsx
```diff
@@ -138,7 +138,10 @@ export default class InvocationActionCardComponent extends React.Component<Props
     this.setState({ lastOperation: undefined });
     const executionId = this.props.search.get("executionId");
     if (!executionId) return;
-    this.operationStream = waitExecution(executionId, {
+
+    const service = rpcService.getRegionalServiceOrDefault(this.props.model.stringCommandLineOption("remote_cache"));
```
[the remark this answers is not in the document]
jdhollen: thanks, bad but normally functional copy-paste on my part.

app/invocation/invocation.tsx
jdhollen: i'm kinda caught between doing a nearly-useless change and at least keeping the code flagged as something that needs to care about regions.  i guess i'm gonna just go for it here--that way if we ever do have a bit more data in state we can update here.

My understanding is right that there's nothing dangling on the execution right now that would otherwise identify the region (i.e., we just have the digest and don't know what server that lives on?)

bduffany: yeah we don't have any region identifiers AFAIK - we could in theory add region to the Executions table though.

bduffany (approved):

PR #6655, opened 2024-05-28
Fix files not being cloned when initially adding to filecache

bduffany:
Context: https://buildbuddy-corp.slack.com/archives/C05GZM1EGM8/p1705093049089379

Also add a little tool to make it easier to test with XFS locally.

**Related issues**: N/A

### DIFF
--- a/enterprise/tools/mount_temp_xfs.sh
+++ b/enterprise/tools/mount_temp_xfs.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -euo pipefail
+
+# Mounts an empty, temporary XFS filesystem to the given path.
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):

enterprise/tools/mount_temp_xfs.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
tylerwilliams, line 1: I also wrote https://gist.github.com/tylerwilliams/58800b42524c31b9036dd3bbe03b855c

to test performance of this stuff.

PR #6656, opened 2024-05-28
Show history from all branches if a workflow never runs on main/master

jdhollen:
Before:
Screenshot 2024-05-28 at 3 02 58 PM

After:
Screenshot 2024-05-28 at 3 02 32 PM

**Related issues**: N/A


maggie-lou (approved):
Nice! Thanks for improving this

PR #6657, opened 2024-05-28
[ci_runner] Default to fetch depth=1 if possible to improve fetch performance

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -1634,9 +1634,6 @@ func (ws *workspace) sync(ctx context.Context) error {
 	if err := ws.fetchPushedRef(ctx); err != nil {
 		return status.WrapError(err, "fetch pushed ref")
 	}
-	if err := ws.fetchTargetRef(ctx); err != nil {
-		return status.WrapError(err, "fetch target ref")
-	}
 
 	// Clean up in case a previous workflow made a mess.
 	cleanArgs := []string{
@@ -1674,8 +1671,10 @@ func (ws *workspace) sync(ctx context.Context) error {
 	// If enabled, merge the target branch (if different from the
 	// pushed branch) so that the workflow can pick up any changes not yet
 	// incorporated into the pushed branch.
-	mergeEnabled := action.GetTriggers().GetPullRequestTrigger().GetMergeWithBase()
-	if mergeEnabled && ws.shouldMergeBranches() {
+	if ws.shouldMergeBranches(action.GetTriggers()) {
+		if err := ws.fetchTargetRef(ctx); err != nil {
+			return status.WrapError(err, "fetch target ref")
+		}
 		targetRef := fmt.Sprintf("%s/%s", gitRemoteName(*targetRepoURL), *targetBranch)
 		if _, err := git(ctx, ws.log, "merge", "--no-edit", targetRef); err != nil && !isAlreadyUpToDate(err) {
 			errMsg := err.Output
@@ -1707,8 +1706,10 @@ func (ws *workspace) sync(ctx context.Context) error {
 	return nil
 }
 
-func (ws *workspace) shouldMergeBranches() bool {
-	return *targetRepoURL != "" &&
+func (ws *workspace) shouldMergeBranches(actionTriggers *config.Triggers) bool {
+	return actionTriggers.GetPullRequestTrigger() != nil &&
+		actionTriggers.GetPullRequestTrigger().GetMergeWithBase() &&
+		*targetRepoURL != "" &&
 		*targetBranch != "" &&
 		(*pushedRepoURL != *targetRepoURL || *pushedBranch != *targetBranch)
 }
@@ -1749,9 +1750,6 @@ func (ws *workspace) fetchPushedRef(ctx context.Context) error {
 }
 
 func (ws *workspace) fetchTargetRef(ctx context.Context) error {
-	if !ws.shouldMergeBranches() {
-		return nil
-	}
 	// Fetch with --depth=0 to ensure the merge base commit is fetched
 	fetchDepth := 0
 	return ws.fetch(ctx, *targetRepoURL, []string{*targetBranch}, fetchDepth)
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -91,6 +91,10 @@ const (
 	defaultGitRemoteName = "origin"
 	forkGitRemoteName    = "fork"
 
+	// If smart fetch depth is enabled, the runner will try to fetch the minimum
+	// depth required.
+	smartFetchDepth = -1
+
 	// Env vars set by workflow runner
 	// NOTE: These env vars are not populated for non-private repos.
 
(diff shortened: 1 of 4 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1627,8 +1631,11 @@ func (ws *workspace) sync(ctx context.Context) error {
 		return status.InvalidArgumentError("expected at least one of `pushed_branch` or `commit_sha` to be set")
 	}
 
-	if err := ws.fetchRefs(ctx); err != nil {
-		return status.WrapError(err, "fetch refs")
+	if err := ws.fetchPushedRef(ctx); err != nil {
+		return status.WrapError(err, "fetch pushed ref")
+	}
+	if err := ws.fetchTargetRef(ctx); err != nil {
+		return status.WrapError(err, "fetch target ref")
```
bduffany, line 1638: not 100% sure but in the case where the pushed and target refs are both in the same repo, I think `fetchRefs` would fetch both refs in a single git fetch command. I'd think that might be faster than doing 2 separate commands, since the client and server would have to re-negotiate which blobs need to be fetched?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1700,51 +1707,56 @@ func (ws *workspace) sync(ctx context.Context) error {
...
-			pushedRefToFetch = *commitSHA
-		}
-		forkRefs = append(forkRefs, pushedRefToFetch)
-	} else {
-		if *pushedBranch == "" {
-			baseRefs = append(baseRefs, *commitSHA)
-		} else if *pushedBranch != *targetBranch {
-			baseRefs = append(baseRefs, *pushedBranch)
-		}
+func (ws *workspace) shouldMergeBranches() bool {
```
bduffany, line 1710: IMO this name is a bit misleading since it doesn't take `GetMergeWithBase()` into account. Not sure what a better name would be though - maybe "isMultiBranch()"? like, the workflow "involves" multiple branches (which may be fetched but not necessarily merged)

bduffany (commented):
Hm it looks like I actually forgot to plumb through `GitFetchDepth` from `buildbuddy.yaml` in the CI runner args [here](https://github.com/buildbuddy-io/buildbuddy/pull/6470/files#diff-c84f2b557cbccb24c92decc8f2c7f2164084794e999079a41831b3dd83bfda6b) :skull: do you mind also adding that in this PR?

(We might need to make `GitFetchDepth` a pointer so that we can tell whether it was set in the YAML, and if it was not set then we shouldn't pass the flag to the CI runner at all, so that it properly defaults to -1)

maggie-lou:
> Hm it looks like I actually forgot to plumb through `GitFetchDepth` from `buildbuddy.yaml` in the CI runner args [here](https://github.com/buildbuddy-io/buildbuddy/pull/6470/files#diff-c84f2b557cbccb24c92decc8f2c7f2164084794e999079a41831b3dd83bfda6b) 💀 do you mind also adding that in this PR?

Ah I was wondering why things hadn't broken. My understanding is that the merge behavior will break if we fetch the target and fork with --depth=1, because there is no merge base. So I think we'd have to rollout the backend changes first, and then plumb through the flags (and make sure people on self-hosted executors have both changes)

maggie-lou:
> > Hm it looks like I actually forgot to plumb through `GitFetchDepth` from `buildbuddy.yaml` in the CI runner args [here](https://github.com/buildbuddy-io/buildbuddy/pull/6470/files#diff-c84f2b557cbccb24c92decc8f2c7f2164084794e999079a41831b3dd83bfda6b) 💀 do you mind also adding that in this PR?
> 
> Ah I was wondering why things hadn't broken. My understanding is that the merge behavior will break if we fetch the target and fork with --depth=1, because there is no merge base. So I think we'd have to rollout the backend changes first, and then plumb through the flags (and make sure people on self-hosted executors have both changes)



> > Hm it looks like I actually forgot to plumb through `GitFetchDepth` from `buildbuddy.yaml` in the CI runner args [here](https://github.com/buildbuddy-io/buildbuddy/pull/6470/files#diff-c84f2b557cbccb24c92decc8f2c7f2164084794e999079a41831b3dd83bfda6b) 💀 do you mind also adding that in this PR?
> 
> Ah I was wondering why things hadn't broken. My understanding is that the merge behavior will break if we fetch the target and fork with --depth=1, because there is no merge base. So I think we'd have to rollout the backend changes first, and then plumb through the flags (and make sure people on self-hosted executors have both changes)

@bduffany It actually feels like we shouldn't have customers use this flag at all, especially if they want to merge with base

enterprise/server/cmd/ci_runner/main.go
maggie-lou: I did a test, and it's about the same (Tested on a couple different branches as a sanity check, and while the results differ, on average the times look similar)

I'd prefer not to pass --merge_with_base as a flag, because we already have so many and it's messy how we have to read the buildbuddy.yaml in both the workflow service and the ci_runner, and the workflow service passes some of the data as a flag, but the ci_runner reads the other half of the data directly from the config

`time git fetch --force origin rb_local_commit master` 
=> 12.99s user 1.81s system 46% cpu 31.689 total
-------
`time git fetch --force origin rb_local_commit ` (30.419s total)
=> 13.16s user 1.87s system 52% cpu 28.750 total
`time git fetch --force origin master `
=> 0.26s user 0.07s system 20% cpu 1.669 total

bduffany: > I'd prefer not to pass --merge_with_base as a flag, because we already have so many and it's messy how we have to read the buildbuddy.yaml in both the workflow service and the ci_runner, and the workflow service passes some of the data as a flag, but the ci_runner reads the other half of the data directly from the config

Sort of a side note (we don't really have to do this now) but another benefit of having this info is that we can skip fetching the `target_branch` when `merge_with_base` is false. I also agree that adding more flags is ugly. Tyler recently added a flag to pass the `Action` as base64, which we could use instead. It's currently only used for Kythe actions, but I think it would be fine to pass it unconditionally

bduffany: D'oh, nevermind - I see that we're fetching the pushed ref first, then reading the config, then deciding whether to fetch the target branch based on that. LGTM

maggie-lou:
@bduffany Sorry if things are confusing while I'm changing things / working across multiple PRs. Here's a summary of what I've been thinking

1) I moved the fetch of the target branch until after we checkout the pushed branch and can read merge_from_base from the buildbuddy config. This way we don't fetch it if merge_with_base is disabled, and we don't need to add another flag for merge_with_base.

I also did some testing and it doesn't seem like the fetch is much faster if we fetch multiple branches at the same time (in a comment above)

2) [I have another PR to make git fetch filters more generic.](https://github.com/buildbuddy-io/buildbuddy/pull/6667) Given that we were not passing `git_fetch_depth` to the ci_runner, it appears as if no one is using that feature :P To simplify, I'm thinking we just remove `git_fetch_depth`, and users can pass `--depth=0` in `git_fetch_filters` if they need. That way we encourage people to use the default behavior. I could do that in a follow-up PR, but for now won't add the flag to the workflow service

3) In a follow up I can look into using the serialized action flag, as opposed to re-reading the buildbuddy.yaml in ci_runner

bduffany (approved):

PR #6667, opened 2024-05-29
[ci_runner] Support passing arbitrary fetch options

maggie-lou:
This will make the flag more flexible. For example, a customer wants to pass custom fetch options to the ci_runner like `--no-show-forced-updates`

**Related issues**: N/A


maggie-lou:
Added https://github.com/buildbuddy-io/buildbuddy/pull/6674 instead
Not having custom filters will make it easier for us to iterate on the default behavior

PR #6669, opened 2024-05-29
Delete runner pool persisted state code

bduffany:
The persisted state mechanism is no longer needed now that we have COW snapshots for VMs and match actions to snapshots via the cache.

**Related issues**: N/A


tylerwilliams (approved):

enterprise/server/remote_execution/containers/firecracker/containeropts.go
```diff
@@ -12,19 +11,19 @@ type ContainerOpts struct {
...
-	// This is not required if SavedState is specified, since the configuration
-	// is stored along with the state.
+	// This is not required if SnapshotKey is specified, since the configuration
+	// is stored in the snapshot.
 	VMConfiguration *fcpb.VMConfiguration
 
+	// SnapshotKey is an optional snapshot key to resume from. If it is not
+	// found in cache, then an error will be returned when attempting to create
+	// the VM.
+	SnapshotKey *fcpb.SnapshotKey
```
maggie-lou, line 21: nit: This is only used for vmstart right? Should we rename this to something like `DebugOverrideSnapshotKey` to make it clear that even if this is not set, we can and will still successfully start from snapshots

bduffany: That name feels a little too verbose / overly qualified to me - I think the comment explaining that it's optional is sufficient

PR #6670, opened 2024-05-29
Populate `use_repo` in MODULE.bazel with `bb fix`

siggisim:
Right now only populates it if it doesn't exist in the MODULE.bazel file already. In the future we can improve this to merge with existing `use_repo` rules.

### DIFF
--- a/cli/fix/fix.go
+++ b/cli/fix/fix.go
@@ -53,9 +53,7 @@ func HandleFix(args []string) (exitCode int, err error) {
 		return 1, err
 	}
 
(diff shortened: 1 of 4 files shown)


bduffany (approved):

fmeum:
Not entirely sure how well it can be integrated into `bb`, but `bazel mod tidy` in Bazel 7.1.2 and later automatically manages `use_repo` for `go_deps` and other extensions. It collects `buildozer` commands and uses a hermetic buildozer binary to apply them locally.

PR #6675, opened 2024-05-30
fix index creation: CREATE INDEX doesn't support specifying both ALGORITHM and LOCK

jdhollen:
**Related issues**: N/A


sluongng (approved):
LGTM

vadimberezniker (approved):

PR #6676, opened 2024-05-30
Use pointer instead of struct for left joined tables

tempoz:
@bduffany pointed out that this was a better way to do `LEFT JOIN`s. Should make it more clear what's happening in the query and remove the possibility of actual empty fields being mistaken for missing rows (even though that's very unlikely here).

**Related issues**: N/A


tempoz:
This is auth code, and so requires two approvals

bduffany (approved):

vadimberezniker (approved):

PR #6678, opened 2024-05-30
[ci_runner] Fix using incorrect url to report github statuses for

maggie-lou:
The "target repo" terminology doesn't make sense for the ExecuteWorkflow API or remote bazel, so I removed it from the ExecuteWorkflow API. I forgot to update these references to use the pushed repo URL instead

**Related issues**: N/A

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -1574,16 +1574,16 @@ func (ws *workflowService) createQueuedStatus(ctx context.Context, wf *tables.Wo
 	}
 	invocationURL += "?queued=true"
 	status := github.NewGithubStatusPayload(actionName, invocationURL, "Queued...", github.PendingState)
-	baseURL := getBaseURL(wd)
-	provider, err := ws.providerForRepo(baseURL)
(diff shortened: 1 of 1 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -1575,11 +1574,28 @@ func (ws *workflowService) createQueuedStatus(ctx context.Context, wf *tables.Wo
...
+	provider, err := ws.providerForRepo(baseURL)
 	if err != nil {
 		return err
 	}
-	return provider.CreateStatus(ctx, wf.AccessToken, wd.TargetRepoURL, wd.SHA, status)
+	return provider.CreateStatus(ctx, wf.AccessToken, baseURL, wd.SHA, status)
+}
+
+// getBaseURL returns the primary URL the workflow should be attributed to.
+func getBaseURL(wd *interfaces.WebhookData) string {
```
bduffany, line 1586: nit: should we call this 'statusReportingURL' ? I could see 'baseURL' getting confused with TargetRepoURL

bduffany (approved):

PR #6680, opened 2024-05-30
[raft]  Add a session proto message for idempotency support

luluz66:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3452

### DIFF
--- a/proto/raft.proto
+++ b/proto/raft.proto
@@ -306,7 +306,7 @@ message Session {
   uint64 index = 2;
   // A timestamp when the session will be cleaned up; and when the state machine
   // can safely clean up the entry.
-  int64 expired_at = 3;
+  int64 expiry_usec = 3;
 
   // Optional. The response associated with the command associated with the
   // index.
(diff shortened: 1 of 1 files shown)


proto/raft.proto
```diff
@@ -311,6 +330,9 @@ message BatchCmdRequest {
 
   // Any post-commit hooks that should be run after a batch is *committed*.
   repeated PostCommitHook post_commit_hooks = 5;
+
+  // Session is used to guarantee idempotency for writes on the state machine.
+  optional Session session = 6;
```
tylerwilliams, line 335: Should this be in a header instead?

proto/raft.proto
```diff
@@ -294,6 +294,25 @@ enum FinalizeOperation {
...
+  // An unique identifier of the client session.
+  // Example: 018fcb2f-cfab-718b-974e-64907df56881
+  bytes id = 1;
+  // Each BatchCmdRequest has its unique <id, index>. The index is
+  // monotonically increasing.
+  // Example: 1717100891000000000
+  uint64 index = 2;
+  // A timestamp when the session will be cleaned up; and when the state machine
+  // can safely clean up the entry.
+  int64 expired_at = 3;
```
tylerwilliams, line 309: s/expired_at/expiry/g

And can we specify units in the comment?

proto/raft.proto
luluz66: Done.

proto/raft.proto
luluz66: I renamed to expiry_usec.

tylerwilliams (approved):

PR #6684, opened 2024-05-31
[ci_runner] Use merge commit sha from webhook event

maggie-lou:
To support merge-with-base behavior on pull requests, we've been fetching the target branch with --depth=0 (to ensure the merge-base is fetched) and manually calling `git merge` on the ci_runner. This can be slow. GitHub automatically generates the mergeCommitSHA on pull requests, so use that from the webhook data instead if it has been set

**Related issues**: N/A


server/interfaces/interfaces.go
```diff
@@ -754,13 +753,20 @@ type WebhookData struct {
...
 	TargetBranch string
 
+	// MergeCommitSHA is the SHA generated by merging the pushed branch into the
+	// target branch. This behavior is used to guarantee that CI runs on up-to-date
+	// refs. It will be empty for non-pull-request events.The merge commit SHA is
+	// generated by a background job, so can be empty for pull-request events if
+	// the background job has not completed, or if the webhook provider does not
+	// provide a generated merge commit. In this case, it should be manually
+	// generated by merging the pushed and target branches.
+	MergeCommitSHA string
```
bduffany, line 768: any idea how reliable this merge_commit_sha is? if we're unsure, can we put it behind a flag and test in dev for a while?

bduffany: err I guess if it's non-empty, then it should exist remotely, otherwise they wouldn't know the SHA?

bduffany (approved):

server/interfaces/interfaces.go
maggie-lou: If it exists on the webhook data, we should be able to reliably fetch it. I'll merge this now, so we'll have until Thursday to see how this does in dev

PR #6688, opened 2024-05-31
Allow fetching ExecuteResponse for invocation actions when only actionDigest is known

bduffany:
If we only have an `actionDigest` in the URL (i.e. when clicking on an action in the cache requests card), look up the Execution from the executions table using the invocation ID + action digest, and fetch the ExecuteResponse from the resulting execution ID. (We store the ExecuteResponse in AC at the key given by `hash(executionID)`)

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3454


app/invocation/invocation_action_card.tsx
```diff
@@ -241,23 +304,33 @@ export default class InvocationActionCardComponent extends React.Component<Props
...
-          this.fetchStdout(actionResult);
-          this.fetchStderr(actionResult);
-          this.fetchServerLogs(executeResponse);
-        }
+        return executeResponse;
+      });
+  }
+
+  fetchExecuteResponseByActionDigest(actionDigest: build.bazel.remote.execution.v2.Digest) {
+    return rpc_service.service
```
jdhollen, line 312: this should be using the "regional" rpc service via `getRegionalServiceOrDefault`, right?

server/buildbuddy_server/buildbuddy_server.go
jdhollen, line 1: this is just an accidental carry-over from the other pr, right?

bduffany: woops, rebased

app/invocation/invocation_action_card.tsx
bduffany: yup, done

jdhollen (approved):

PR #6689, opened 2024-05-31
[ci_runner] Explicitly pass --unshallow flag when attempting to fetch full git history

maggie-lou:
The `--depth` option is sticky across fetching the same branch. So if I fetch `master` in one snapshot run with `--depth=1`, in a future snapshot run I'll have to explicitly set `--unshallow` if I want it to fetch the whole git history. *shakes fist at git* 

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -1858,8 +1858,18 @@ func (ws *workspace) fetch(ctx context.Context, remoteURL string, refs []string,
 	for _, filter := range *gitFetchFilters {
 		fetchArgs = append(fetchArgs, "--filter="+filter)
 	}
+	// The --depth option is sticky when fetching the same branch. If --depth
+	// was set in a previous snapshot run, make sure to unset it if needed
 	if fetchDepth == 0 {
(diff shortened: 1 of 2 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1858,7 +1858,9 @@ func (ws *workspace) fetch(ctx context.Context, remoteURL string, refs []string,
 	for _, filter := range *gitFetchFilters {
 		fetchArgs = append(fetchArgs, "--filter="+filter)
 	}
-	if fetchDepth > 0 {
+	if fetchDepth == 0 {
```
bduffany, line 1861: just sanity checking, do we handle -1 ('smart' depth) correctly here?

bduffany (approved):

enterprise/server/cmd/ci_runner/main.go
maggie-lou: Yeah - we never directly pass the flag value, so we'll never pass -1. (Though I have considered how it's not ideal that we might accidentally pass -1. I wish it was a pointer and not an int)

PR #6693, opened 2024-06-03
Annotate explicitly requested resources in task size proto

bduffany:
Allow differentiating between user-reserved memory/CPU vs. resources that we determined using our automatic task sizer.

**Related issues**: N/A

### DIFF
--- a/proto/scheduler.proto
+++ b/proto/scheduler.proto
@@ -102,8 +102,8 @@ message SchedulingMetadata {
   // Task size used for scheduling purposes, when the scheduler is deciding
   // which executors (if any) may execute a task, and also when an executor is
   // deciding which task to dequeue. Executors may see a different value of this
-  // field than what the scheduler sees, depending on measured_task_size. See
-  // documentation of that field for more info.
+  // field than what the scheduler sees, depending on measured_task_size or
+  // predicted_task_size. See documentation of those fields for more info.
   TaskSize task_size = 1;
 
   // Task size measured from a previous task execution of a similar task, if
@@ -125,7 +125,9 @@ message SchedulingMetadata {
   // The resources explicitly requested by the user. These will be unset if the
   // user did not explicitly request resources.
   //
-  // These should NOT be used for scheduling purposes.
+  // These should NOT be used for scheduling purposes. See comments on other
+  // TaskSize fields in this message to understand how tasks are sized for
+  // scheduling.
   TaskSize requested_task_size = 10;
 
   string os = 2;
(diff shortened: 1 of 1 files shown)


proto/scheduler.proto
```diff
@@ -95,6 +95,14 @@ message TaskSize {
 
   // The task's estimated disk space requirement (beyond task inputs).
   int64 estimated_free_disk_bytes = 3;
+
+  // Whether the memory estimate comes from an explicit request parameter as
+  // opposed to an automatic estimate.
+  bool explicit_memory = 4;
```
siggisim, line 101: Thoughts on using separate `estimated_` and `requested_` fields here instead of the booleans?

I think either way is fine, just curious about the tradeoff.

bduffany: I thought that recording the raw resource requests as integer fields might get confusing, because we apply caps to certain resources, and IMO it'd be unclear whether these fields are pre- or post- cap. Also, having explicit values would open up more opportunity to use those requested values in places that we shouldn't be.

tylerwilliams (commented):
See comments

bduffany:
@tylerwilliams what do you think about my reasoning in  https://github.com/buildbuddy-io/buildbuddy/pull/6693#discussion_r1624727131 ? Sorry I edited a little bit so if you saw my reply originally it might've been a bit stale

tylerwilliams:
> @tylerwilliams what do you think about my reasoning in [#6693 (comment)](https://github.com/buildbuddy-io/buildbuddy/pull/6693#discussion_r1624722285) ? Sorry I edited a little bit so if you saw my reply originally it might've been a bit stale

I think you could just add a comment if it's unclear. If you really cared you could add requested, capped, used, etc. Ultimately you want to keep these fields separate because it's unambiguous and you might want to show them separately later.

bduffany:
OK @tylerwilliams I added a `requested_task_size` field to SchedulerMetadata. We don't send this up as part of the ExecutedActionMetadata today, so my plan is to add `SchedulingMetadata` to the `auxiliary_metadata` (list of `any` protos)

proto/scheduler.proto
```diff
@@ -122,6 +122,12 @@ message SchedulingMetadata {
   // exceed the executor's limits.
   TaskSize predicted_task_size = 8;
 
+  // The resources explicitly requested by the user. These will be unset if the
+  // user did not explicitly request resources.
+  //
+  // These should NOT be used for scheduling purposes.
```
tylerwilliams, line 128: nit: point to what should be used instead?

tylerwilliams (approved):
LGTM

thanks for the changes

proto/scheduler.proto
bduffany: Done

PR #6697, opened 2024-06-03
Add boilerplate for OCI runtime

bduffany:
Done (modulo bugs)
- [x] Minimal `Create()` (generate OCI bundle dir w/ `config.json` - only busybox is supported as the rootfs for now)
- [x] Minimal `Exec()` (run a command within the container)
- [x] `Delete()`
- [x] seccomp profile

Big TODOs:
- [ ] rootfs (container images)
  - [ ] Implement `PullImage` / `IsImageCached` (maybe use google/go-containerregistry?)
  - [ ] Deduping parallel image pulls
  - [ ] Creating the overlayfs for each image
  - [ ] (nice to have) Image streaming?
- [ ] Figure out stdio (stdin, stdout, stderr). Wiring these up to the exec.Command doesn't seem to work. We can maybe look at how containers/conmon deals with this
  - [ ] Test remote persistent workers - these rely on stdin 
- [ ] cgroups / implement Stats()
- [ ] Networking
- [ ] Security features (rlimits, capabilities, apparmor)

Smaller tasks:
- [ ] Implement Run() (`crun run`)
- [ ] Devices (/dev/shm etc.) - see TODOs in `createSpec`
- [ ] Debug story (i.e. equivalent of `podman logs` to debug stuck tasks)
- [ ] Misc TODOs sprinkled as comments throughout the file
- [ ] Support more platform props (`dockerUser` etc.)
- [ ] Support more executor-level flags (devices, mounts, etc.)
- [ ] Add more tests
- [ ] Rootless mode (nice to have, to make local testing easier)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/commandutil"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/container"
-	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/platform"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/oci"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
@@ -36,9 +35,9 @@ var (
 )
 
 const (
-	// Root dir where container runtime info will live.
-	// TODO: pick something better here
-	ociVersion   = "1.1.0-rc.3" // matches podman
+	ociVersion = "1.1.0-rc.3" // matches podman
+
+	// Execution root directory path relative to the container rootfs directory.
 	execrootPath = "buildbuddy-execroot"
 )
 
@@ -181,7 +180,7 @@ func (c *ociContainer) createBundle(ctx context.Context, cmd *repb.Command) erro
 }
 
 func (c *ociContainer) IsolationType() string {
-	return string(platform.OCIContainerType)
+	return "oci" // TODO: make const in platform.go
 }
 
 func (c *ociContainer) IsImageCached(ctx context.Context) (bool, error) {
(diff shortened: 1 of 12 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -0,0 +1,595 @@
...
+
+	log.CtxDebugf(ctx, "Running %v", runtimeArgs)
+
+	cmd := exec.Command(runtimeArgs[0], runtimeArgs[1:]...)
+	cmd.Dir = wd
+	// TODO: figure out process I/O. crun's process model does not seem to be
+	// compatible with just setting stdout/stderr to an io.Writer like we would
+	// expect - if we set stdout to a bytes.Buffer here for example, then
+	// cmd.Run() will hang. One observation is that crun seems to re-exec
+	// itself, which might have something to do with it.
```
fmeum, line 541: Just adding something that I learned the hard way in the past: If `cmd.Stdout` is set to a writer that isn't an `os.File`, Go will spawn a goroutine to copy the output and wait for it in `cmd.Wait()`. This goroutine will not complete until the stdout of the launched process is closed, even if the process itself exited (but e.g. launched a child process that keeps stdout open).

bduffany: Yeah, that behavior has caused problems for me before as well and thought that might have been the issue here too, but I think something different is happening in this case that I don't understand yet. I tried using `os.Pipe()`which returns `os.File` for the read/write ends of the pipe,  and even tried using the lower-level `syscall.ForkExec` to pass the file descriptors for the write-ends of the pipes, but no luck. (Turns out this is effectively what `exec.Command` does under the hood anyway, so I guess it's not too surprising that this didn't work)

I think `crun` does fork a child process, because the `ForkExec` call returns a pid that is different than the `crun` process that I see in `ps aux` while it's hanging. I guess maybe the child process is blocked while trying to write to the pipe, for some reason.

I was briefly looking at https://github.com/containers/conmon which I believe is how podman is spawning / managing the runtime but it will take a bit of time for me to grok.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -0,0 +1,595 @@
...
+		}
+		return fmt.Errorf("exit code %d", res.ExitCode)
+	}
+	return nil
+}
+
+func (c *ociContainer) invokeRuntime(ctx context.Context, command *repb.Command, args ...string) *interfaces.CommandResult {
+	start := time.Now()
+	defer func() {
+		// DO NOT SUBMIT
```
iain-macdonald, line 510: ?

iain-macdonald (commented):
🤯

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: woops - I actually like having these logs while we're developing, so changed to a TODO to remove it later.

iain-macdonald (approved):
Nice work! This is awesome!

PR #6698, opened 2024-06-03
[ci_runner] Rename commit field to ref to be more generic

maggie-lou:
Sorry I'd written a TODO in my notes to do this and forgot to add it in the last PR.

I figured this might be helpful, because bitbucket exposes a merge branch, but not a merge commit SHA.

**Related issues**: N/A


bduffany (approved):
I think this is fine but wonder if we should move towards an approach of trying to resolve refs to SHAs as soon as we get the workflow execution request (e.g. using the git provider's API), just to reduce the possibility of race conditions (though it won't be perfect)?

maggie-lou:
> I think this is fine but wonder if we should move towards an approach of trying to resolve refs to SHAs as soon as we get the workflow execution request (e.g. using the git provider's API), just to reduce the possibility of race conditions (though it won't be perfect)?

@bduffany At this point the previous code has already been deployed (to dev), so I'd need to make this PR backwards compatible anyway. I'm leaning towards just punting on this for now. You're right that we lose accuracy guarantees without a sha, which isn't a good solution

PR #6701, opened 2024-06-04
[raft] store createdAt instead of expiredAt in Session

luluz66:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3452
Store CreatedAt instead of ExpiredAt in Session. This allows more flexibility when we want to change the TTL of the session.

Also use *sessionLifetime directly instead of pass it into client session to make creation methods simpler.


tylerwilliams (approved):
Nice

PR #6706, opened 2024-06-04
Don't show error in the UI when execution stream is not found

bduffany:
If an execution stream no longer exists (e.g. if the execution happened a long time ago), then `WaitExecution` returns a pubsub error saying that the stream is not found. Check for this error in the UI and just ignore it, displaying an "Unknown" execution status instead (or the status from the cached ExecuteResponse if it hasn't expired from cache yet).

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3473

### DIFF
--- a/app/invocation/BUILD
+++ b/app/invocation/BUILD
@@ -135,6 +135,7 @@ ts_library(
         "//app/service:rpc_service",
         "//app/terminal",
         "//app/util:cache",
+        "//app/util:rpc",
         "//proto:firecracker_ts_proto",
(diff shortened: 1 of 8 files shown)


vadimberezniker (approved):
Looks like this change is already included in https://github.com/buildbuddy-io/buildbuddy/pull/6704?

bduffany:
> Looks like this change is already included in #6704?

ah I accidentally pushed that PR branch to this commit. Will fix

PR #6711, opened 2024-06-05
add rpcs to fetch stats on flaky targets and example flakes.

jdhollen:
**Related issues**: N/A


server/target/target.go
```diff
@@ -740,6 +741,173 @@ func readPaginatedTargetsFromPrimaryDB(ctx context.Context, env environment.Env,
...
 
+func GetTargetStats(ctx context.Context, env environment.Env, req *trpb.GetTargetStatsRequest) (*trpb.GetTargetStatsResponse, error) {
+	u, err := env.GetAuthenticator().AuthenticatedUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !isReadFromOLAPDBEnabled(env) {
+		return nil, status.UnimplementedError("Target stats requires an OLAP DB.")
+	}
+	sevenDaysAgo := time.Now().Add(-7 * 24 * time.Hour).UnixMicro()
```
iain-macdonald, line 752: Sub?

iain-macdonald (approved):
LGTM, Approval

server/target/target.go
jdhollen: unfortunately, `Sub` takes a time and returns the duration between the two times, whereas `Add` returns a new time where the specified duration has been added.  makes total sense, right?

PR #6714, opened 2024-06-05
Fix bb remote on macos

bduffany:
This will allow using `bb remote --os=darwin` to target mac workflow executors. Today we have to run `bb remote --os=darwin --runner_exec_properties=workload-isolation-type=none --runner_exec_properties=container-image=` and it fails with "exec format error" since we're provisioning the CI runner from the Linux build bundled with the app.

* Provision the ci_runner from the executor instead of the Linux build bundled with the app, so that we get the correct architecture. This matches what we do for workflows.
* When setting `bb remote --os=darwin`, don't set container-image and set workload-isolation-type=none (instead of firecracker).

**Related issues**: N/A

### DIFF
--- a/enterprise/server/hostedrunner/BUILD
+++ b/enterprise/server/hostedrunner/BUILD
@@ -7,7 +7,6 @@ go_library(
     srcs = ["hostedrunner.go"],
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/hostedrunner",
     deps = [
-        "//enterprise/server/cmd/ci_runner/bundle",
         "//enterprise/server/remote_execution/operation",
         "//enterprise/server/remote_execution/platform",
         "//proto:remote_execution_go_proto",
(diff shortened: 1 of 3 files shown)


maggie-lou:
Because we aren't using any isolation type, this would break if multiple people try to use remote bazel on the same executor, right?

bduffany:
@maggie-lou it should generally work - Mac workflows support running multiple concurrent runs. Your comment made me realize though that we shouldn't be setting WORKDIR_OVERRIDE for macos and instead run workflows within the workspace directory (done).

maggie-lou:
I forgot that we have mac workflows. Nice though - the working directory change should address my original concern

This PR reminds me of how we wish the ci_runner binary was bundled with the apps instead. Should we 
1) Only use the ci_runner from the executor for Macs? 
2) Modify workflows to use the remote bazel approach, where the app uploads the binary to the cache and the executors download it?

bduffany:
@maggie that seems reasonable, mind if we do that as a followup?

PR #6718, opened 2024-06-05
[ci_runner] Turn off merge commit sha usage by default

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/workflow/config/config.go
+++ b/enterprise/server/workflow/config/config.go
@@ -85,10 +85,21 @@ type PullRequestTrigger struct {
 	Branches []string `yaml:"branches"`
 	// NOTE: If nil, defaults to true.
 	MergeWithBase *bool `yaml:"merge_with_base"`
+	// If MergeWithBase is enabled, determines whether the CI runner should manually
+	// merge the pushed and target branches. If this is disabled, the runner
+	// will try to use the merge commit SHA provided by the CI provider as a
+	// performance optimization.
+	// If MergeWithBase is disabled, this does nothing.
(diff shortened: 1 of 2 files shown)


enterprise/server/workflow/config/config.go
```diff
@@ -85,12 +85,23 @@ type PullRequestTrigger struct {
 	Branches []string `yaml:"branches"`
 	// NOTE: If nil, defaults to true.
 	MergeWithBase *bool `yaml:"merge_with_base"`
+	// If MergeWithBase is enabled, determines whether the CI runner should manually
+	// merge the pushed and target branches. If this is disabled, the runner
+	// will try to use the merge commit SHA provided by the CI provider as a
+	// performance optimization.
+	// If MergeWithBase is disabled, this does nothing.
+	ForceManualMergeWithBase *bool `yaml:"force_manual_merge_with_base"`
```
bduffany, line 93: The manual merge feels somewhat like an implementation detail. I wonder if we should call this something like `fetch_branch_history`? (Since really the issue is that the remote merge is also forcing `fetch_depth: 1` to be enabled?)

jdhollen (approved):

enterprise/server/workflow/config/config.go
bduffany: Another option is to always use the remote merge commit but then let the user control the fetch depth via the existing `fetch_depth` option. That way there aren't two controls for the same thing

bduffany: Discussed offline - we can check this in as an experimental option just to avoid breaking tomorrow's release, then later think more about what this config should look like (or if we want a config for this at all)

bduffany (approved):

PR #6721, opened 2024-06-05
[ci_runner] Enable merge commit usage in our workflows

maggie-lou:
DNM until after https://github.com/buildbuddy-io/buildbuddy/pull/6718 goes out

This will let us test the behavior

**Related issues**: N/A

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -8,6 +8,7 @@ actions:
       pull_request:
         branches:
           - "*"
+        force_manual_merge_with_base: false
     bazel_commands:
       - test //... --config=linux-workflows --config=race --test_tag_filters=-performance,-webdriver,-docker,-bare
   - name: Test with BzlMod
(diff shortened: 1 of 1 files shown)


bduffany (approved):

buildbuddy.yaml
```diff
@@ -30,6 +32,7 @@ actions:
       pull_request:
         branches:
           - "*"
+        force_manual_merge_with_base: false
```
bduffany, line 35: while you're here do you mind also setting `git_fetch_depth: 0` on the action?

More explanation in the thread starting [here](https://github.com/buildbuddy-io/buildbuddy-internal/issues/3482#issuecomment-2151017215) but I think we cannot determine changed files unless either (a) we do a non-shallow fetch, or (b) if GitHub's generated merge commits are generated using `--squash`. (I'm not sure whether that's the case yet)

bduffany: err, actually it looks like GitHub's generated merge commit is always a merge commit even if commit history is linear. So we can set `git_fetch_depth: 2` instead of `git_fetch_depth: 0`, and we should be able to use `HEAD^1` as the diff base.

PR #6723, opened 2024-06-06
Update to Bazel 7.2.0

fmeum:
If there are no regressions, this will become 7.2.0 on Monday.

**Related issues**: N/A


fmeum:
There is a failure on Windows that I will bisect.

sluongng:
could we/should we go to 7.2.0 with this PR?
Or is there a 7.2.1 in the pipeline worth waiting for?

sluongng (approved):

PR #6724, opened 2024-06-06
counting tree uploads

sluongng:
Start counting Tree objects toward the upload statistic of each execution.

Also improved the test setup for the "LotsOfNesting" case to lay out the tree shape.
Expect this to help us troubleshoot issues with Directory and Tree content easier in the future.


bduffany (commented):
do you mind if we hold off on this? I am working on an approach that counts transfers at a lower-level (`BatchCASUploader`) which I think would be better than the current approach of sprinkling this accounting logic in a bunch of different places.

sluongng:
discussed with @bduffany in Slack about this.
We won't be able to skip the dirs in between. Gona close this for now. Some hunks here could be cherry-picked into a future PR if needed.

PR #6725, opened 2024-06-06
Fix prettier.sh when using merge commit

bduffany:
**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy/pull/6725


maggie-lou:
Unfortunately I don't think this will work if a PR has multiple commits. We would want the diff to include files changed from all commits, not just the last one.

Here's an example merge commit sha (you can fetch and check it out locally): a6cd17fa569b4d80d4940f6048f07cb707e06100
From PR: https://github.com/buildbuddy-io/buildbuddy/pull/6702

bduffany:
`HEAD^1` is not pointing to the previous commit in the PR, it's pointing to the merge base commit. This is because git's generated merge commit seems to be created using `git merge --no-ff` from what I can tell. As a result, there will always be an explicit merge commit created, which has two parents. Parent 1 (HEAD^1) is the base branch, e.g. `master`. Parent 2 (HEAD^2) is the PR branch.

Note that `HEAD^1` is not the same as `HEAD~1` (you might already know this, but calling this out since I didn't know this before, in case this is the source of confusion)

The ASCII diagrams in this PR also try to illustrate this (maybe poorly)

bduffany:
For that commit you linked:

```
#!/usr/bin/env bash
set -e
cd "$(mktemp -d)"
git init .
git remote add origin https://github.com/buildbuddy-io/buildbuddy
git fetch origin a6cd17fa569b4d80d4940f6048f07cb707e06100 --depth=2
git checkout a6cd17fa569b4d80d4940f6048f07cb707e06100
git log --all --oneline --graph
```

It shows this graph:

```
*   a6cd17f (HEAD) Merge 817a18f34da851060a98d0d691da068353631f76 into 2e3a33b292a29c2d7a867d9b190c5b9a716d1256
|\  
| * 817a18f (grafted) ./buildfix.sh -a
* 2e3a33b (grafted) store createdAt instead of expiredAt
```

`HEAD^1` is

maggie-lou:
TIL -Okay I did not know that HEAD~1 and HEAD^1 were different 😅 

Still however, in that example, HEAD^1 points to 2e3a33b `store createdAt instead of expiredAt`. That's the first commit of the[ feature branch](https://github.com/buildbuddy-io/buildbuddy/pull/6702/commits), not the the tip of the master branch

bduffany:
> That's the first commit of the[ feature branch](https://github.com/buildbuddy-io/buildbuddy/pull/6702/commits), not the the tip of the master branch

I think that PR was not initially based on master (i.e. it was a stacked PR). I think it was stacked on top of this PR: https://github.com/buildbuddy-io/buildbuddy/pull/6701

So once the base branch changed to `master`, there should be a more recent merge commit at which the workflow was run. But I think by the time it was rebased, we disabled the merge commit stuff? So I can't find an invocation in history.

TLDR this all seems working as intended?

bduffany:
we chatted more about this offline - turns out that GitHub's merge refs that it creates automatically are not automatically created unless the user takes certain actions in the UI, so will close this out.

PR #6727, opened 2024-06-06
Move output_paths migration boilerplate to NewDirHelper, instead of its caller

bduffany:
While looking at `NewDirHelper` constructor, it took me a while to figure out that the `outputDirectories` and `outputPaths` parameters are not just direct passthroughs from `command.GetOutputDirectories()` and `command.GetOutputPaths()` respectively.

Instead, every caller of `NewDirHelper` is expected to write some boilerplate code to adapt the old-style `output_directories`/`output_files` into a new-style `output_paths` input.

It took me a while to realize this, and I spent a lot of time confused about why we only needed to call `IsOutputPath` in order to determine whether a directory needed to be uploaded, rather than `IsOutputPath || IsOutputDirectory || IsOutputFile`.

So, instead of having the caller adapt these fields, do it internally. To me, this is a lot more clear, since it keeps all the details of the `output_paths` migration in one place. It also helps simplify the calling code and reduce the possibility for errors.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -83,7 +83,14 @@ type DirHelper struct {
 	dirPerms fs.FileMode
 }
 
-func NewDirHelper(rootDir string, outputDirectories, outputPaths []string, dirPerms fs.FileMode) *DirHelper {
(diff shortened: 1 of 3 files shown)


sluongng (approved):
LGTM

PR #6728, opened 2024-06-06
Rename some variables and add comments in UploadTree to help with readability

bduffany:
This PR attempts to make some improvements that I think would help when trying to grok this code:

* Rename `treesToUpload` to `outputDirectoryPaths`, since it's a list of output paths that correspond to `ActionResult.output_directories`, not a list of `Trees`.
* Use more consistent naming for all variables that refer to paths: always start with an unambiguous qualifier like `parent` or `child`, and consistently use `FullPath` to refer to an absolute path rather than less obvious alternatives like `fqfn` (it's an acronym that not everyone may get) and `dir` (can be confused with `Directory` proto)
* Rename `uploadFileFn` and `uploadDirFn` to `visitFile` and `visitDir` since these don't upload anything; they just collect things to upload.
* Rename `trees` to `outputDirectoryTrees` to make it more clear that trees are directly related to output directories.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -382,16 +382,17 @@ func handleSymlink(dirHelper *DirHelper, rootDir string, cmd *repb.Command, acti
 func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper, instanceName string, digestFunction repb.DigestFunction_Value, rootDir string, cmd *repb.Command, actionResult *repb.ActionResult) (*TransferInfo, error) {
 	txInfo := &TransferInfo{}
 	startTime := time.Now()
-	treesToUpload := make([]string, 0)
+	outputDirectoryPaths := make([]string, 0)
 	filesToUpload := make([]*fileToUpload, 0)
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -401,15 +402,15 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
 		return uploadableFile.FileNode(), nil
 	}
 
-	var uploadDirFn func(parentDir, dirName string) (*repb.DirectoryNode, error)
-	uploadDirFn = func(parentDir, dirName string) (*repb.DirectoryNode, error) {
+	var visitDir func(parentFullPath, dirName string) (*repb.DirectoryNode, error)
```
iain-macdonald: supernit: inline this? unless that doesn't compile?

iain-macdonald (approved):
Nice, thanks for doing this!

enterprise/server/remote_execution/dirtools/dirtools.go
bduffany: yeah, doesn't compile unfortunately

PR #6730, opened 2024-06-06
Remove Bazel 6.0.0 binaries from deps.bzl

iain-macdonald:
These were use for old probers that I deleted a while ago.

**Related issues**: N/A

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -6894,24 +6894,6 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
         urls = ["https://github.com/bazelbuild/bazel/releases/download/5.3.2/bazel-5.3.2-linux-arm64"],
     )
     http_file(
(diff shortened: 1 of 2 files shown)


iain-macdonald:
Brandon, I see you added the arm64 binaries in https://github.com/buildbuddy-io/buildbuddy/pull/6289/files#diff-80b2ade301026edea37e1df9b9bdc7e58d93ecb5e8c1293cab97734e5c438738 but I was wondering if they're needed or if Bazelisk will fetch Bazel for us for integration tests? I think it's the latter but wanted to confirm.

bduffany (approved):
Most integration tests are explicitly depending on the bazel version they want to use, to avoid network usage if possible. So if the build doesn't break when removing the 6.0.0 build I think this should be OK

iain-macdonald:
Sounds good, removed the 6.5.0-arm64 variant too as tests pass without it.

PR #6731, opened 2024-06-06
Fix prettier.sh, take 2

bduffany:
When running on master, make sure we diff against `HEAD~1`. We previously were diffing against master itself which would result in an empty diff.

Also fetch with `depth: 0`, otherwise we won't be able to resolve `HEAD~1` for push events.

**Related issues**: N/A


maggie-lou:
What's the point of running this workflow on push: master?

bduffany:
> What's the point of running this workflow on push: master?

In case a PR hasn't merged with master in a while, it might be possible that once it's merged with master, the resulting code doesn't pass a style check. Running it on master will check for that

maggie-lou (approved):

PR #6732, opened 2024-06-06
Run required checks when release branch is pushed

bduffany:
**Related issues**: N/A


maggie-lou:
We can do this as well, but we need this for -internal repo right?

bduffany:
Yep, will add it for the internal repo too

iain-macdonald (approved):

PR #6735, opened 2024-06-06
[raft] pass one clock

luluz66:
**Related issues**: N/A
Put the clock in testutil.StoreFactory; so that when we create the store, we can
use the same clock. This also allow us to pass the clock when we call
StartShard.

Also pass the session to the leasekeeper from the store; so that it can be
created from the same clock.

This is needed to make the test for cleaning up expired sessions. Otherwise,
some sessions will be created with fake clock and some are created with real
clock, which can make test and debug very difficult.


tylerwilliams (approved):
LGTM, Approval

PR #6736, opened 2024-06-06
[raft] clean up expired sessions

luluz66:
**Related issues**: buildbuddy-io/buildbuddy-internal#3452
Initiate cleanup sessions in raft/store.

Use the current goroutine that scans the leased replica, and queue tasks to
delete sessions.

Use a rate limiter to make sure we are not delete sessions for all replicas at
once.

Record the time when we last delete the sessions and skip if we deleted the
sessions within SessionLifetime().


tylerwilliams (approved):
LGTM, Approval

PR #6741, opened 2024-06-07
[ociruntime] Fix stderr/stdout/stdin handling

fmeum:
Wire up `interfaces.Stdio` with the option of discarding all output, which is prevents hanging in `create` as `sleep` never closes its inherited stdout/stderr.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -1,6 +1,7 @@
 package ociruntime
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"encoding/json"
@@ -175,6 +176,7 @@ func (c *ociContainer) createBundle(ctx context.Context, cmd *repb.Command) erro
 		return err
 	}
 	os.Stderr.Write(b)
+	os.Stderr.WriteString("\n")
 
 	return nil
 }
@@ -208,12 +210,13 @@ func (c *ociContainer) Create(ctx context.Context, workDir string) error {
 	if err := c.createBundle(ctx, pid1); err != nil {
 		return status.UnavailableErrorf("create OCI bundle: %s", err)
 	}
-	// Create container
-	if err := c.invokeRuntimeSimple(ctx, "create", "--bundle="+c.bundlePath(), cid); err != nil {
+	// Create container (discard output as pid 1 is sleep, which inherits and never closes
+	// stdout/stderr)
+	if err := c.invokeRuntimeSimple(ctx, nil, "create", "--bundle="+c.bundlePath(), c.cid); err != nil {
 		return status.UnavailableErrorf("create container: %s", err)
 	}
 	// Start container
-	if err := c.invokeRuntimeSimple(ctx, "start", c.cid); err != nil {
+	if err := c.invokeRuntimeSimple(ctx, &interfaces.Stdio{}, "start", c.cid); err != nil {
 		return status.UnavailableErrorf("start container: %s", err)
 	}
 	return nil
@@ -221,7 +224,7 @@ func (c *ociContainer) Create(ctx context.Context, workDir string) error {
 
 func (c *ociContainer) Exec(ctx context.Context, cmd *repb.Command, stdio *interfaces.Stdio) *interfaces.CommandResult {
 	// TODO: wire up stdio
-	return c.invokeRuntime(ctx, cmd, "exec", "--cwd="+execrootPath, c.cid)
+	return c.invokeRuntime(ctx, cmd, stdio, "exec", "--cwd="+execrootPath, "--detach", c.cid)
 }
 
 func (c *ociContainer) Pause(ctx context.Context) error {
@@ -240,7 +243,7 @@ func (c *ociContainer) Remove(ctx context.Context) error {
 
 	var firstErr error
 
-	if err := c.invokeRuntimeSimple(ctx, "delete", "--force", c.cid); err != nil {
+	if err := c.invokeRuntimeSimple(ctx, &interfaces.Stdio{}, "delete", "--force", c.cid); err != nil {
 		firstErr = status.UnavailableErrorf("failed to delete container: %s", err)
 	}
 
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -208,20 +210,21 @@ func (c *ociContainer) Create(ctx context.Context, workDir string) error {
...
 	// Start container
-	if err := c.invokeRuntimeSimple(ctx, "start", c.cid); err != nil {
+	if err := c.invokeRuntimeSimple(ctx, &interfaces.Stdio{}, "start", c.cid); err != nil {
 		return status.UnavailableErrorf("start container: %s", err)
 	}
 	return nil
 }
 
 func (c *ociContainer) Exec(ctx context.Context, cmd *repb.Command, stdio *interfaces.Stdio) *interfaces.CommandResult {
 	// TODO: wire up stdio
```
bduffany, line 226:
```suggestion
```

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -208,20 +210,21 @@ func (c *ociContainer) Create(ctx context.Context, workDir string) error {
...
-	return c.invokeRuntime(ctx, cmd, "exec", "--cwd="+execrootPath, c.cid)
+	return c.invokeRuntime(ctx, cmd, stdio, "exec", "--cwd="+execrootPath, "--detach", c.cid)
```
bduffany, line 227: out of curiosity, why do we need `--detach` here? (maybe worth adding a comment for future readers too)

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -208,20 +210,21 @@ func (c *ociContainer) Create(ctx context.Context, workDir string) error {
 	if err := c.createBundle(ctx, pid1); err != nil {
 		return status.UnavailableErrorf("create OCI bundle: %s", err)
 	}
-	// Create container
-	if err := c.invokeRuntimeSimple(ctx, "create", "--bundle="+c.bundlePath(), cid); err != nil {
+	// Create container (discard output as pid 1 is sleep, which inherits and never closes
+	// stdout/stderr)
+	if err := c.invokeRuntimeSimple(ctx, nil, "create", "--bundle="+c.bundlePath(), c.cid); err != nil {
```
fmeum, line 215: This comes with the caveat that we will not see error messages emitted by `create`. We could instead redirect stdout/stderr to actual files for this command only.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -208,20 +210,21 @@ func (c *ociContainer) Create(ctx context.Context, workDir string) error {
 	if err := c.createBundle(ctx, pid1); err != nil {
 		return status.UnavailableErrorf("create OCI bundle: %s", err)
 	}
-	// Create container
-	if err := c.invokeRuntimeSimple(ctx, "create", "--bundle="+c.bundlePath(), cid); err != nil {
+	// Create container (discard output as pid 1 is sleep, which inherits and never closes
+	// stdout/stderr)
```
bduffany, line 214: Hmm any idea why `create` is the thing that hangs, rather than `start`? I would think that the sleep process isn't created until we call `start`, but I'm not sure. At least if I do `podman create` it looks like that doesn't create the process

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -487,8 +490,8 @@ func (c *ociContainer) createSpec(cmd *repb.Command) (*specs.Spec, error) {
 	return &spec, nil
 }
 
-func (c *ociContainer) invokeRuntimeSimple(ctx context.Context, args ...string) error {
-	res := c.invokeRuntime(ctx, &repb.Command{}, args...)
+func (c *ociContainer) invokeRuntimeSimple(ctx context.Context, stdio *interfaces.Stdio, args ...string) error {
```
bduffany, line 493: could we instead use the non-simple func for the callers that want to discard stdio? The (badly named) `Simple` variant was intended to always buffer stderr, and if the command fails, return the buffered stderr in the `error`.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany, line 215: by discarding stdio here, I think that means that we won't get the error message from `crun` itself if the `create` fails? (the previous code was sending the error message to `os.Stderr` which is not a good long-term solution, but might be helpful during development while we figure out a better solution)

Also, we don't have to figure this out now - but in the future I think it'd be nice if we could separate the stdout, stderr, and exit code from the runtime, vs the user's executed command. This was one of the pain points with both podman and docker - we have to do heuristics on the exit code and `stderr` contents in order to determine whether there was a podman error that can be retried (`UnavailableError`), vs. a normal action failure (ActionResult with exit_code != 0)

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
fmeum: I investigated this more and, if I'm not mistaken, `crun` starts the entrypoint command in `create`, not `start`. Since I also find this very surprising, I linked the call chain in a comment. I don't know whether `runc` or `podman` behave differently, we may need to extend the new handling to `start`.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
fmeum: I implemented a different approach that allows us to get the `create` error message if the call itself fails. It is a bit hacky, but avoids reimplementing `exec.Cmd.Wait()` in terms of `Cmd.Process`.

Separating runtime from container failures is something I could look into, but I don't know how to do this yet in non-`terminal` mode.

bduffany (approved):

PR #6743, opened 2024-06-07
Fill out the JWT field in AuthenticateResponse

iain-macdonald:
For the period where we're just running the cache-proxy within our own GCP deployment or other deployments that we're comfortable sharing the JWT key with, we can use this to support remotely authenticating. At some point we will probably want to sign these differently, probably using RSA public/private keys, but I'll address that later.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/auth_service/auth_service.go
+++ b/enterprise/server/auth_service/auth_service.go
@@ -3,18 +3,31 @@ package auth_service
 import (
 	"context"
 
+	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
+	"github.com/buildbuddy-io/buildbuddy/server/util/authutil"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 
 	authpb "github.com/buildbuddy-io/buildbuddy/proto/auth"
 )
 
 type AuthService struct {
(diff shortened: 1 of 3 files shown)


iain-macdonald:
Adding Tyler because I need two reviewers for auth code!

enterprise/server/auth_service/auth_service.go
```diff
@@ -3,18 +3,30 @@ package auth_service
...
 }
 
 func Register(env *real_environment.RealEnv) {
-	env.SetAuthService(AuthService{})
+	env.SetAuthService(AuthService{authenticator: env.GetAuthenticator()})
 }
 
 func (a AuthService) Authenticate(ctx context.Context, req *authpb.AuthenticateRequest) (*authpb.AuthenticateResponse, error) {
-	return &authpb.AuthenticateResponse{}, nil
+	u, err := a.authenticator.AuthenticateGRPCRequest(ctx)
```
vadimberezniker, line 22: AuthenticatedGRPCRequest automatically puts the JWT into the context. If you use the JWT from the context you wouldn't need to mess with having expiration on UserInfo.

iain-macdonald: Done

vadimberezniker (approved):

tylerwilliams (approved):

PR #6744, opened 2024-06-07
Split fileToUpload struct into separate file and dir structs

bduffany:
While reading this code, I found the fileToUpload struct to be confusing because it can represent either a file or a directory. 

This PR adds a separate dirToUpload struct and simplifies the code by removing the conditional logic checking whether a fileToUpload is a file or directory.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"flag"
-	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -219,92 +216,85 @@ func newDirToUpload(instanceName string, digestFunction repb.DigestFunction_Valu
...
 		// Note: uploader.Upload closes the file after it is uploaded.
-		if err := uploader.Upload(uploadableFile.resourceName.GetDigest(), rsc); err != nil {
+		if err := uploader.Upload(uploadableFile.digest, f); err != nil {
 			return err
 		}
 	}
+	return nil
+}
 
+func uploadDirs(ctx context.Context, uploader *cachetools.BatchCASUploader, dirsToUpload []*dirToUpload) error {
```
sluongng, line 291: Nit: I think moving this inside UploadTree would make the whole thing easier to follow vs using a separate func.

sluongng (approved):
LGTM.

I also started doing this initially. Then after a bit more refactoring, I realized that it’s easier to follow the logic after I in-lined `DirectoryNode()` and got rid of the 2 structs `fileToUpload` and `dirToUpload` completely. Hope that future PRs will head toward the same direction.

enterprise/server/remote_execution/dirtools/dirtools.go
bduffany: done

bduffany:
> Then after a bit more refactoring, I realized that it’s easier to follow the logic after I in-lined DirectoryNode() and got rid of the 2 structs fileToUpload and dirToUpload completely. Hope that future PRs will head toward the same direction.

Yeah, I would like to eventually reach this state, but would like to gradually transform the code in ways where each change is more digestable and then once the code is in a state we're happy with, make the big 2 changes where we stop uploading directories unnecessarily and also make sure we track all uploads and/or call FindMissing before doing uploads.

iain-macdonald (approved):
LGTM, Approval

PR #6748, opened 2024-06-07
Support using github as a module backend

siggisim:
This allows you to use a github repo directly as a module provided it has a `MODULE.bazel` file (we could even lift that restriction, but keeping it simple for now).

Depending on the BuildBuddy repo at a commit sha (recommended, since the others can get stale):
```
bazel_dep(name = "buildbuddy", version = "github.buildbuddy-io+3b0abcc3d478b7a29ee7550eab36271076fff40c")
```

Depending on a tag:
```
bazel_dep(name = "buildbuddy", version = "github.buildbuddy-io+v1.0.0")
```

Depending on a branch:
```
bazel_dep(name = "buildbuddy", version = "github.buildbuddy-io+release_123")
```

Depending on head:
```
bazel_dep(name = "buildbuddy", version = "github.buildbuddy-io")
```

### DIFF
--- a/enterprise/server/registry/registry_server.go
+++ b/enterprise/server/registry/registry_server.go
@@ -3,9 +3,11 @@ package registry
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"strings"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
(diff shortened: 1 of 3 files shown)


enterprise/server/registry/registry_server.go
```diff
@@ -3,9 +3,11 @@ package registry
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
```
tylerwilliams, line 6: ioutil is dead, long live io

https://pkg.go.dev/io#ReadAll

enterprise/server/registry/registry_server.go
```diff
@@ -38,6 +40,13 @@ func (t *RegistryServer) Start() {
 	proxy := httputil.NewSingleHostReverseProxy(url)
 
 	http.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
+		urlParts := strings.Split(req.URL.Path, "/")
+
+		if len(urlParts) > 3 && urlParts[1] == "modules" && strings.HasPrefix(urlParts[3], "github.") {
+			handleGitHub(w, req)
```
tylerwilliams, line 46: up to you if it's worth it but you could have your top level method (or even this callsite) handle the reading and response writing and then your functional code that actually does a thing can return standard types and is more easily testable. totally a matter of preference tho

tylerwilliams (approved):
LGTM

enterprise/server/registry/registry_server.go
siggisim: Done

enterprise/server/registry/registry_server.go
siggisim: Done

PR #6750, opened 2024-06-07
Support using npm as a module backend

siggisim:
This allows you to use a npm package directly as a module.

Depending on an unscoped npm package at a version:
```
bazel_dep(name = "docusaurus", version = "npm+1.14.7")
```

Depending on a scoped npm package at a version:
```
bazel_dep(name = "docusaurus_core", version = "npm.docusaurus+3.4.0")
```

### DIFF
--- a/enterprise/server/registry/registry_server.go
+++ b/enterprise/server/registry/registry_server.go
@@ -47,7 +47,7 @@ func (t *RegistryServer) Start() {
 			bytes, status, err := handleGitHub(path)
 			w.WriteHeader(status)
 			if err != nil {
-				log.Printf("error serving github module %s: %s", path, err)
+				log.Errorf("error serving github module %s: %s", path, err)
 			}
 			w.Write(bytes)
 			return
@@ -57,7 +57,7 @@ func (t *RegistryServer) Start() {
 			bytes, status, err := handleNPM(path)
 			w.WriteHeader(status)
 			if err != nil {
-				log.Printf("error serving npm module %s: %s", path, err)
+				log.Errorf("error serving npm module %s: %s", path, err)
 			}
 			w.Write(bytes)
 			return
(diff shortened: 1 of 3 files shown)


enterprise/server/registry/registry_server.go
```diff
@@ -53,6 +53,16 @@ func (t *RegistryServer) Start() {
 			return
 		}
 
+		if len(pathPaths) > 3 && pathPaths[1] == "modules" && strings.HasPrefix(pathPaths[3], "npm") {
+			bytes, status, err := handleNPM(path)
+			w.WriteHeader(status)
+			if err != nil {
+				log.Printf("error serving npm module %s: %s", path, err)
```
tylerwilliams, line 60: log.Errorf if you want to get fancy

enterprise/server/registry/registry_server.go
```diff
@@ -53,6 +53,16 @@ func (t *RegistryServer) Start() {
 			return
 		}
 
+		if len(pathPaths) > 3 && pathPaths[1] == "modules" && strings.HasPrefix(pathPaths[3], "npm") {
+			bytes, status, err := handleNPM(path)
```
tylerwilliams, line 57: s/bytes/buf/g

tylerwilliams (approved):

enterprise/server/registry/registry_server.go
tylerwilliams: b/c bytes is a commonly used package already

enterprise/server/registry/registry_server.go
siggisim: Done

enterprise/server/registry/registry_server.go
siggisim: Done

PR #6755, opened 2024-06-10
[RB] Support configuring whether to run targets on the local vs remote runner

maggie-lou:
Currently for remote `run`s, the generated binary will not be executed remotely, and will be fetched and run locally (See here for [implementation](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/cmd/ci_runner/main.go#L157)). Add a configurable flag, because for CI people may want it to be executed remotely

**Related issues**: N/A


cli/remotebazel/remotebazel.go
```diff
@@ -704,7 +705,7 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 	fetchOutputs := false
 	runOutput := false
 	bazelArgs := arg.GetBazelArgs(opts.Args)
-	if len(bazelArgs) > 0 && (bazelArgs[0] == "build" || bazelArgs[0] == "run") {
+	if *fetchRemoteOutputs && len(bazelArgs) > 0 && (bazelArgs[0] == "build" || bazelArgs[0] == "run") {
```
bduffany, line 708: IIUC this changes the default for `bazel run` from running locally to running remotely. For running webservers, running remotely isn't a great default because there's no way to access the VM network.

Should we force running remotely for now if the command is `run`?

maggie-lou: I'll close this for now until we have a better idea of customer needs around this

maggie-lou: @bduffany Re-opening this. Currently I want this because it will be helpful for unit tests. But I could also see customers wanting this. For example if they're using remote bazel to build and publish a docker image, they might expect the target to run on the OS / arch / container image / with the compute resources that they've specified in their remote request

I think for a `build` it makes sense to default to downloading the outputs. For a `run`, I actually think it makes sense to default to running remotely, and of course there will still be the option to flip the flag. What do you think?

siggisim (approved):

PR #6760, opened 2024-06-10
Fix action cache fetching from /file/download endpoint

bduffany:
**Related issues**: N/A


bduffany:
(Working on adding a regression test)

maggie-lou (approved):

PR #6764, opened 2024-06-10
Fix prettier for zsh

maggie-lou:
I use zsh as a shell, and it does not have the mapfile command 

**Related issues**: N/A


bduffany (approved):

bduffany (commented):
this script should be getting executed with `/bin/bash`  because of the `#!` at the top of the file? maybe it's an issue with older versions of bash? (either way, LGTM)

maggie-lou:
Oh you're right - it was just an issue with an older version of bash

PR #6767, opened 2024-06-10
[raft] Add a deadline when stop store in test

luluz66:
**Related issues**: N/A
So if cannot gracefully stop in 10 seconds, we will hard stop grpc server.

### DIFF
--- a/enterprise/server/raft/testutil/testutil.go
+++ b/enterprise/server/raft/testutil/testutil.go
@@ -139,7 +139,10 @@ func (sf *StoreFactory) NewStore(t *testing.T) *TestingStore {
 	ts.Store = store
 
 	t.Cleanup(func() {
-		store.Stop(context.TODO())
+		ctx := context.Background()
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
LGTM, Approval

enterprise/server/raft/testutil/testutil.go
```diff
@@ -139,7 +139,10 @@ func (sf *StoreFactory) NewStore(t *testing.T) *TestingStore {
 	ts.Store = store
 
 	t.Cleanup(func() {
-		store.Stop(context.TODO())
+		ctx := context.Background()
```
tylerwilliams, line 142: Can we use env.GetServerContext here instead?

For local testing, the shutdown timeout is like 3 seconds

luluz66: we don't have a env here. changing the timeout from 10s to 3 seconds

PR #6774, opened 2024-06-11
[RB] Set the build user and add a remote bazel role

maggie-lou:
This is what the UI looks like

Screenshot 2024-06-11 at 3 18 52 PM

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3430

### DIFF
--- a/app/root/root.css
+++ b/app/root/root.css
@@ -984,6 +984,11 @@ code .comment {
   background: #cfd8dc;
 }
 
+.role-badge.HOSTED_BAZEL {
+  background: #a1a6a8;
+  color: white;
+}
+
 .empty-state .button {
   width: 100%;
   padding: 16px 32px;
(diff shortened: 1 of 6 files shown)


app/root/root.css
```diff
@@ -984,6 +984,11 @@ code .comment {
   background: #cfd8dc;
 }
 
+.role-badge.HOSTED_BAZEL {
+  background: #a1a6a8;
```
siggisim, line 988: Maybe we can use a more exciting color here like green 500 from https://materialui.co/colors/

siggisim (approved):

app/root/root.css
maggie-lou: What do you think of this? The green seemed a little too close to a "success" color
Screenshot 2024-06-11 at 3 41 01 PM

PR #6772, opened 2024-06-11
Upgrade rules_go, gazelle and rules_proto

sluongng:
Upgrade rules_go to v0.48.0.

This requires us to use the latest release of rules_proto (v6.0.0).
However, our rules_proto declaration was being shadowed by another rule's transitive dependencies.
Move rules_proto http_archive declarations up to avoid shadow-ing.

The latest version of rules_proto depends on `bazel_features` repo, but did not declare this as a transitive deps, so we also need to add that to our WORKSPACE setup.
The BzlMod setup has the correct transitive deps declaration so there is no need for it.

PR #6778, opened 2024-06-11
Upgrade gazelle to 0.37.0

sluongng:
Upgrade gazelle to the latest release.
This release comes with an upgrade on bazelbuild/buildtools, which made `buildifier` and `gazelle` format the BUILD files with a stricter sorting order.

To make the PR a bit easier to review, I have split the change out into separate commits with notes for each commit message.


sluongng:
Note that `go install github.com/bazelbuild/buildtools/buildifier@latest` is required for local setup so that a stricter `buildifier` format would be applied via `./buildfix.sh`.

We might also need to bump `bb` release version in that script after this was merged + build + release so that the latest `bb fix` is applied.

sluongng:
This was split off from https://github.com/buildbuddy-io/buildbuddy/pull/6772 to make the review a bit easier 😓

fmeum (approved):

PR #6781, opened 2024-06-11
Enable mysql query interpolation.

vadimberezniker:
Turns out it was the MySQL driver that was falling back to prepared statements, not GORM.
If the MySQL driver received a query with arguments, it forced the caller to use a prepare/execute combination which requires two round trips. MySQL has an option called 'interpolateParams' that does the interpolation client side in the driver instead of relying on the server. 

**Related issues**: N/A


tylerwilliams (approved):
cool :)

tylerwilliams (approved):

PR #6784, opened 2024-06-11
Add basic image support to OCI runtime isolation

bduffany:
* Pull and extract image layers with `go-containerregistry`
  * Pull concurrently with up to 8 goroutines
  * For now, use `tar` to extract, using `--no-same-user` to avoid permissions issues
* Set up rootfs as an overlayfs mount

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
@@ -2,13 +2,19 @@ package ociruntime_test
 
 import (
 	"context"
+	"io/fs"
 	"log"
+	"os/exec"
+	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 
 	"github.com/bazelbuild/rules_go/go/runfiles"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/container"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/containers/ociruntime"
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/platform"
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/util/oci"
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/testutil/testenv"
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
```diff
@@ -17,7 +23,31 @@ import (
...
+
+// Returns an image ref pointing to a remotely hosted busybox image.
+// Skips the test if we don't have mount permissions.
+// TODO: support rootless overlayfs mounts and get rid of this
+func realBusyboxImage(t *testing.T) string {
+	if !hasMountPermissions(t) {
+		t.Skipf("using a real container image with overlayfs requires mount permissions")
+	}
+	return "mirror.gcr.io/library/busybox"
+}
```
bduffany, line 46: This setup is pretty clunky but I wanted to make sure we could at least run some of the tests locally to make iteration a little faster until we figure out full rootless support. I don't think rootless support needs to be included with the initial release of this isolation type, though.

iain-macdonald (approved):
LGTM, Approval

enterprise/server/util/oci/oci.go
```diff
@@ -177,3 +185,94 @@ func Resolve(ctx context.Context, imageName string, platform *rgpb.Platform, cre
...
+		})
+	}
+	if err := eg.Wait(); err != nil {
+		return nil, err
+	}
+	return layers, nil
+}
+
+func tmpSuffix() string {
+	return fmt.Sprintf(".%d.tmp", rand.Int63n(1e16))
```
fmeum, line 277: Might want to use the new `math/rand/v2` here.

fmeum (approved):

fmeum:
> rpc error: code = FailedPrecondition desc = could not find a usable container runtime in PATH

How about downloading a hermetic `crun` binary with Bazel and adding it to the `go_test` as `data`? That would also make it easier to debug issues with the test locally.

enterprise/server/util/oci/oci.go
bduffany: Done

bduffany:
> How about downloading a hermetic crun binary with Bazel and adding it to the go_test as data? That would also make it easier to debug issues with the test locally.

Done in https://github.com/buildbuddy-io/buildbuddy/pull/6800 - they release a static executable, so I'm fetching it with http_file in that PR.

PR #6788, opened 2024-06-12
Add a remote authenticator

iain-macdonald:
**Related issues**: N/A


enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/buildbuddy-io/buildbuddy/server/util/lru"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+
+	authpb "github.com/buildbuddy-io/buildbuddy/proto/auth"
+)
+
+const (
+	jwtCacheSize = 10_000
```
tylerwilliams, line 23: can you add a comment here so that if someone needed to change this they would have enough information to do so?

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+var (
+	authHeaders = []string{authutil.APIKeyHeader}
+
+	remoteAuthTarget = flag.String("auth.remote_auth_target", "grpcs://remote.buildbuddy.dev", "The gRPC target of the remote authentication API.")
```
tylerwilliams, line 29: maybe we should default this to "" so that we don't end up with a situation where we're trying to auth between environments?

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+func NewRemoteAuthenticator() (*RemoteAuthenticator, error) {
+	conn, err := grpc_client.DialSimpleWithoutPooling(*remoteAuthTarget)
```
tylerwilliams, line 33: pq no DialSimple?

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (a *RemoteAuthenticator) AuthenticatedHTTPContext(w http.ResponseWriter, r *http.Request) context.Context {
+	return r.Context()
+}
+
+func (a *RemoteAuthenticator) SSOEnabled() bool {
+	return false
+}
+
+func (a *RemoteAuthenticator) AuthenticatedGRPCContext(ctx context.Context) context.Context {
```
tylerwilliams, line 95: I keep staring at this thinking there's a way to simplify it but b/c "hashAuthHeaders" can return an error it's tough.

tylerwilliams (commented):
Few nits but I think this looks pretty good. We'll see what @Vadim thinks :) e

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,182 @@
...
+	key, err := hashAuthHeaders(ctx)
```
vadimberezniker, line 96: As discussed, I don't think we need the hashing today. I don't think it's adding any value today.

enterprise/server/remoteauth/remoteauth.go
vadimberezniker, line 1: as discussed, we need to make sure jwt auth also works for executors to be able to talk to the cache proxy

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		jwt, err := a.authenticate(ctx)
```
vadimberezniker, line 98: Why do we call authenticate here if hashAuthHeaders fails?

vadimberezniker (changes requested):

enterprise/server/remoteauth/remoteauth.go
iain-macdonald: Good call, done.

enterprise/server/remoteauth/remoteauth.go
iain-macdonald: Oh yeah, done.

enterprise/server/remoteauth/remoteauth.go
iain-macdonald: Just copied this from somewhere. Switched to DialSimple.

enterprise/server/remoteauth/remoteauth.go
iain-macdonald: Reworked this a bit to address Vadim's comments, so it's a bit simpler now but LMK what you think.

enterprise/server/remoteauth/remoteauth.go
iain-macdonald: SG, done.

enterprise/server/remoteauth/remoteauth.go
iain-macdonald: Added a flow for JWT pass-through and a comment to address verifying JWTs locally that I'll address as a follow-up.

enterprise/server/remoteauth/remoteauth.go
iain-macdonald: I just figured that was a safe backup, but switched it to just using ctx in that case.

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,189 @@
...
+	if found {
+		return context.WithValue(ctx, authutil.ContextTokenStringKey, jwt)
+	}
+	jwt, err := a.authenticate(ctx)
+	if err != nil {
+		log.Debugf("Error remotely authenticating: %s", err)
+		return ctx
+	}
+	a.cache.Add(key, jwt)
+	return context.WithValue(ctx, authutil.ContextTokenStringKey, jwt)
```
bduffany, line 121: there's `authutil.AuthContextWithError()` which might be the "correct" way to return an error from this func - e.g. oidc.go will call `AuthContextFromClaims` which use that func when there's an error

iain-macdonald: oh nice, switched!

vadimberezniker (approved):

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,208 @@
...
+func (a *RemoteAuthenticator) AuthenticatedHTTPContext(w http.ResponseWriter, r *http.Request) context.Context {
+	return r.Context()
+}
+
+func (a *RemoteAuthenticator) SSOEnabled() bool {
+	return false
+}
+
+func (a *RemoteAuthenticator) AuthenticatedGRPCContext(ctx context.Context) context.Context {
+	// Pass valid JWTs through as is.
```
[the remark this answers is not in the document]
iain-macdonald: Done

enterprise/server/remoteauth/remoteauth.go
```diff
@@ -0,0 +1,208 @@
...
+	return *resp.Jwt, nil
+}
+
+func getAPIKey(ctx context.Context) string {
+	return getLastMetadataValue(ctx, authutil.APIKeyHeader)
+}
+
+// Returns a valid JWT from the incoming RPC metadata, or an error an invalid
+// JWT is present, or an empty string and no error if no JWT is provided.
+func getValidJWT(ctx context.Context, claimsCache *claims.ClaimsCache) (string, error) {
```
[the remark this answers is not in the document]
iain-macdonald: Sure, done

enterprise/server/oidc/oidc.go
```diff
@@ -303,11 +302,11 @@ func newOpenIDAuthenticator(ctx context.Context, env environment.Env, oauthProvi
 	}
 
 	claimsFunc := claims.ParseClaims
-	if *claimsCacheTTL > 0 {
-		claimsCache, err := claims.NewClaimsCache(ctx, *claimsCacheTTL)
-		if err != nil {
-			return nil, err
-		}
+	claimsCache, err := claims.NewClaimsCache(ctx)
```
tylerwilliams, line 305: seems fine but in general if err is nil it's assumed that the value is safe to use, and in this case it's nil, so that's a little error prone. This is the only callsite I see, but can you make sure to add a comment on func NewClaimsCache that this may return nil?

tylerwilliams (approved):

server/util/claims/claims.go
```diff
@@ -336,7 +337,11 @@ type ClaimsCache struct {
 	lru interfaces.LRU[*Claims]
 }
 
-func NewClaimsCache(ctx context.Context, ttl time.Duration) (*ClaimsCache, error) {
+func NewClaimsCache(ctx context.Context) (*ClaimsCache, error) {
```
tylerwilliams, line 340: nit: can you remove ctx from here (and on up the chain) too?

tylerwilliams (approved):
lg w/ the 2 nits

enterprise/server/oidc/oidc.go
iain-macdonald: Done

server/util/claims/claims.go
iain-macdonald: Done

PR #6790, opened 2024-06-12
[raft] nodeliveness: use a seperate mutex for stop

luluz66:
**Related issues**: buildbuddy-io/buildbuddy-internal#3496

The cancelFn in Release() is not run because it's waiting for the mutex held by
ensureValidLease. The purpose of the cancelFn was to make ensureValidLease exit
early so that it can release the lock.


tylerwilliams (approved):
looks ok but if we can I'd prefer we just remove Release and all this stuff entirely

luluz66:
Abandon in favor of https://github.com/buildbuddy-io/buildbuddy/pull/6791

PR #6791, opened 2024-06-12
[raft] nodeliveness: remove stop var and Release()

luluz66:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy/pull/6790
The cancelFn in Release() is not run because it's waiting for the mutex held by
ensureValidLease. The purpose of the cancelFn was to make ensureValidLease exit
early so that it can release the lock.

Since Release() method is only used in store.Stop(); and we are not really using
the functionality of clearing the lease, change Release() to Stop() and only
call the cancelFn.

Remove stop var and use sync.Once.


tylerwilliams (approved):
LGTM, Approval

PR #6793, opened 2024-06-12
build_event_handler: chunk event log by default

sluongng:
This is a saner default for folks with large Bazel build logs.
In some cases, it will prevent OOM in self-hosted OSS BuildBuddy.

Also fixed a minor typo in the next line.

### DIFF
--- a/server/build_event_protocol/build_event_handler/build_event_handler.go
+++ b/server/build_event_protocol/build_event_handler/build_event_handler.go
@@ -105,8 +105,8 @@ const (
 
 var (
 	chunkFileSizeBytes      = flag.Int("storage.chunk_file_size_bytes", 3_000_000 /* 3 MB */, "How many bytes to buffer in memory before flushing a chunk of build protocol data to disk.")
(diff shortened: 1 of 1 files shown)


bduffany:
> In some cases, it will prevent OOM in self-hosted OSS BuildBuddy.

Could you elaborate on how this prevents OOMs?

carolosf:
We are trying out build buddy.

@sluongng  has been helping us debug.

On a 16gb ec2 instance running a build buddy docker container.

I've seen the ec2 instance become unresponsive.

After some more digging it seemed to be that the available memory on the instance rapidly reduced during builds.

After some more debugging with the go profiling tools. 

We noticed that terminal-to-html was using 10gb of memory and more until the system memory was exhausted.

Enabling chunked event logs has improved things significantly.

The largest object during profiling is only on average about 20mb after the change.

Our build is in the region of 5-7 million lines of code and 150,000 tests.

sluongng:
https://buildbuddy.slack.com/archives/CUHBFVATU/p1718198013052839 should have more context with screenshots of heap dump provided by @carolosf 

This is probably the most important image 
image

bduffany (approved):

PR #6794, opened 2024-06-12
[RB] Improve feedback on CTRL-C

maggie-lou:
**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3387


cli/remotebazel/remotebazel.go
```diff
@@ -489,6 +489,12 @@ func streamLogs(ctx context.Context, bbClient bbspb.BuildBuddyServiceClient, inv
 	var chunks []*elpb.GetEventLogChunkResponse
 	wasLive := false
 	for {
+		select {
+		case <-stopChan:
+			return nil
+		default:
+		}
+
```
bduffany, line 497: instead of having a separate channel wdyt about using context cancellation? `signal.NotifyContext()` might be useful but if it winds up complicating things then nbd.

bduffany (approved):

cli/remotebazel/remotebazel.go
```diff
@@ -489,6 +489,12 @@ func streamLogs(ctx context.Context, bbClient bbspb.BuildBuddyServiceClient, inv
 	var chunks []*elpb.GetEventLogChunkResponse
 	wasLive := false
 	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
```
bduffany, line 496: instead of this check, should we just check for `ctx.Err() != nil` after calling GetEventLogChunk?

cli/remotebazel/remotebazel.go
```diff
@@ -683,14 +695,18 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 	healthChecker := healthcheck.NewHealthChecker("remote-bazel-client")
 	env := real_environment.NewRealEnv(healthChecker)
 
+	ctx = metadata.AppendToOutgoingContext(ctx, "x-buildbuddy-api-key", opts.APIKey)
+
+	// Handle interrupts to cancel the remote run.
+	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
+	defer cancel()
```
bduffany, line 702: should we move this after the `Run()` RPC to ensure that we cancel the remote invocation if it has been scheduled? (I think the previous code was doing that)

cli/remotebazel/remotebazel.go
```diff
@@ -779,17 +795,20 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
...
-	sigChan := make(chan os.Signal, 1)
 	go func() {
-		<-sigChan
-		_, err = bbClient.CancelExecutions(ctx, &inpb.CancelExecutionsRequest{
+		<-ctx.Done()
+		// Use a non-cancelled context to ensure the remote executions are
+		// canceled
+		nonCanceledCtx := context.Background()
+		nonCanceledCtx = metadata.AppendToOutgoingContext(nonCanceledCtx, "x-buildbuddy-api-key", opts.APIKey)
+		_, err = bbClient.CancelExecutions(nonCanceledCtx, &inpb.CancelExecutionsRequest{
```
bduffany, line 804: as of go 1.21 there's context.WithoutCancel for this https://pkg.go.dev/context#WithoutCancel

cli/remotebazel/remotebazel.go
maggie-lou: Ah yeah - good call. Fixed

cli/remotebazel/remotebazel.go
maggie-lou: I had put it before the `Run` RPC because if it takes a while for the runner to get scheduled and you try to CTRL-C, it will not be registered and the script will still continue. Updated to use `context.WithoutCancel`

cli/remotebazel/remotebazel.go
maggie-lou: Oh nice - updated

cli/remotebazel/remotebazel.go
maggie-lou: Actually the more I think about that case (remote runner takes a long time to get scheduled), it feels incorrect that we would wait until after the runner has been scheduled to cancel the executions it has created (this could take a long time if there is queueing). It seems like we should just cancel that remote run request immediately. Though I acknowledge there could be a race condition where the remote runner has created the invocation and we do not cancel it

maggie-lou: Actually the more I think about that case (remote runner takes a long time to get scheduled), it feels incorrect that we would wait until after the runner has been scheduled to cancel the executions it has created (this could take a long time if there is queueing). It seems like we should just cancel that remote run request immediately. Though I acknowledge there could be a race condition where the remote runner has created the invocation and we do not cancel it

I changed it back to previous behavior - where Run takes a cancelable context, but let me know if you still disagree

bduffany: If we ran with `async` and had the run request return the `execution_id` instead of just `invocation_id`, then it seems like we could probably figure out a way to cancel the execution based on the execution_id rather than the invocation ID?

bduffany: > I acknowledge there could be a race condition where the remote runner has created the invocation and we do not cancel it

To address this part more directly - I think we should try to avoid this race condition, since it's important for cancellation to work reliably. e.g. if you're running a `bazel run` that does a deployment, you want to be sure that it's cancelled if you press Ctrl+C

maggie-lou: Rather than have the remote bazel handler handle the case described above, I added some code in the hostedrunner to cancel the invocation if the context is canceled. Given the point you made above, this seemed like the correct behavior given that some customers are directly using the Run API

enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -422,6 +422,11 @@ func waitUntilInvocationExists(ctx context.Context, env environment.Env, executi
 	for {
 		select {
 		case <-ctx.Done():
+			log.Infof("Attempting to cancel invocation %s because the context for the remote runner has ended with %s", invocationID, ctx.Err())
+			// If the context is canceled, ensure the remote run is canceled
+			if err := env.GetRemoteExecutionService().Cancel(context.WithoutCancel(ctx), invocationID); err != nil {
+				return status.InternalErrorf("context canceled with %s but failed to cancel invocation %s", ctx.Err(), invocationID)
```
bduffany, line 428:
```suggestion
				return status.InternalErrorf("context canceled with %s but failed to cancel invocation %s: %s", ctx.Err(), invocationID, err)
```

bduffany (approved):

cli/remotebazel/remotebazel.go
```diff
@@ -779,28 +783,35 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
...
-		_, err = bbClient.CancelExecutions(ctx, &inpb.CancelExecutionsRequest{
+		<-ctx.Done()
+
+		if !isInvocationRunning {
+			return
+		}
+
+		// Use a non-cancelled context to ensure the remote executions are
+		// canceled
+		_, err = bbClient.CancelExecutions(context.WithoutCancel(ctx), &inpb.CancelExecutionsRequest{
```
bduffany, line 796: is this code needed now that it's done in hostedrunner.go?

maggie-lou: Yeah - the hostedrunner:Run call returns after the invocation has been created. We need this block if you want to cancel as the invocation is running

bduffany (approved):

PR #6795, opened 2024-06-12
[ci_runner] Modify Linux workflows to download the binary from the cache, rather than using the binary provisioned on the executors

maggie-lou:
This will make it easier to ship ci_runner changes to customers with self-hosted executors

**Related issues**: N/A

### DIFF
--- a/enterprise/server/workflow/service/service.go
+++ b/enterprise/server/workflow/service/service.go
@@ -52,6 +52,7 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 	"gopkg.in/yaml.v2"
 
+	ci_runner_bundle "github.com/buildbuddy-io/buildbuddy/enterprise/server/cmd/ci_runner/bundle"
 	ctxpb "github.com/buildbuddy-io/buildbuddy/proto/context"
 	inspb "github.com/buildbuddy-io/buildbuddy/proto/invocation_status"
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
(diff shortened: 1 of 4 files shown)


enterprise/server/workflow/service/service.go
```diff
@@ -1162,14 +1160,37 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
...
 
+	// NOTE: For non-Linux workflows, the executor is responsible for ensuring the
+	// buildbuddy_ci_runner binary exists at the workspace root when it sees
+	// the `workflow-id` platform property.
+	inputRootDigest, err := digest.ComputeForMessage(&repb.Directory{}, repb.DigestFunction_SHA256)
+	if err != nil {
+		return nil, err
+	}
+	if os == "" || os == platform.LinuxOperatingSystemName {
+		// On Linux, because we can build the binary with the apps with the right architecture,
```
bduffany, line 1171: nit: check arch as well, since we are starting to get more `linux/arm64` users

```suggestion
	if (os == "" || os == platform.LinuxOperatingSystemName) && (arch == "" || arch == "amd64") {
		// On Linux, because the apps are built for linux/amd64,
```

bduffany (approved):

enterprise/server/workflow/service/service.go
tylerwilliams: +1

tylerwilliams (commented):
See comments

PR #6800, opened 2024-06-12
Provision test OCI runtime binary with bazel

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/BUILD
+++ b/enterprise/server/remote_execution/containers/ociruntime/BUILD
@@ -32,11 +32,15 @@ go_library(
 go_test(
     name = "ociruntime_test",
     srcs = ["ociruntime_test.go"],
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/containers/ociruntime/BUILD
```diff
@@ -47,5 +51,17 @@ go_test(
         "//server/util/testing/flags",
         "@com_github_stretchr_testify//assert",
         "@com_github_stretchr_testify//require",
+        "@io_bazel_rules_go//go/runfiles:go_default_library",
     ],
 )
+
+genrule(
+    name = "crun_copy",
```
fmeum, line 59: I think that this could just be an `alias`

bduffany: nice - done

PR #6801, opened 2024-06-12
bazel: Add a musl cross toolchain targetting Linux amd64/arm64

fmeum:
The new toolchain produces binaries statically linked against musl.

Tested on `//cli` and `//enterprise/server/cmd/executor` from macOS.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -294,22 +294,3 @@ platform(
         "container-image": "docker://gcr.io/flame-public/iain-test:latest",
     },
 )
-
-# Target these platforms with --platforms for a static build linked against musl libc.
-platform(
-    name = "linux_x86_64_musl",
-    constraint_values = [
-        "@platforms//cpu:x86_64",
-        "@platforms//os:linux",
-        "//toolchains:musl_on",
-    ],
-)
-
-platform(
-    name = "linux_arm64_musl",
-    constraint_values = [
-        "@platforms//cpu:arm64",
-        "@platforms//os:linux",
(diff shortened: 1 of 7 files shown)


BUILD
```diff
@@ -294,3 +294,22 @@ platform(
...
+)
+
+platform(
+    name = "linux_arm64_musl",
+    constraint_values = [
+        "@platforms//cpu:arm64",
+        "@platforms//os:linux",
+        "//toolchains:musl_on",
+    ],
+)
```
sluongng, line 315: Nit: we do have `//platforms` for these

sluongng (approved):
Small nit, but LGTM

platforms/BUILD
```diff
@@ -59,3 +59,22 @@ platform(
     constraint_values = HOST_CONSTRAINTS,
     parents = [get_parent_from_constraints(HOST_CONSTRAINTS)],
 )
+
+# Target these platforms with --platforms for a static build linked against musl libc.
+platform(
```
fmeum, line 64: These platforms are really only used as target platforms, do I still need to place any `exec_properties` on them?

sluongng: Good question. I don't think there is any value in adding them today. But it could cause some confusion to why these don't come with `exec_properties`. 🤔 

If there is no way in Bazel to separate exec and target platform definitions, then I think adding some comment on top of these to explain that they should be used as target platform would suffice.

PR #6815, opened 2024-06-13
Use a simpler key for snapshot-specific AC entries

bduffany:
For snapshot-specific keys, _only_ include the snapshot ID in the key, leaving all other fields blank. This will make it possible to have a `--snapshot_id` arg to vmstart, which we can easily copy-paste from the UI, instead of needing to dig through executor logs.


maggie-lou (approved):
Good idea 👍 

I was previously thinking of adding a button to "copy snapshot key" in the UI, because we piped through the snapshot key to the FE when supporting invalidating a specific snapshot. But this is simpler

PR #6816, opened 2024-06-13
ci runner: try to reclaim disk space when usage is high

bduffany:
After running the action (and publishing the success/failure status for the outer invocation), attempt to save disk space by running `git gc` if disk usage exceeds 90%.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -41,6 +41,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/creack/pty"
 	"github.com/docker/go-units"
+	"github.com/elastic/gosigar"
 	"github.com/google/shlex"
 	"github.com/google/uuid"
 	"github.com/logrusorgru/aurora"
@@ -72,6 +73,10 @@ const (
 	// Name of the bazel output base dir. This is written under the workspace
 	// so that it can be cleaned up when the workspace is cleaned up.
 	outputBaseDirName = "output-base"
+
+	// Fraction of disk space that must be in use before we attempt to reclaim
+	// disk space.
+	highDiskUsageThreshold = 0.9
 
 	// Name of the dir where artifacts can be written.
 	// The CI runner provisions subdirectories under this directory, one per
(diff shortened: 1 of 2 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2272,3 +2285,50 @@ func runCredentialHelper() error {
...
+	}
+	usage, err := diskUsageFraction()
+	if err != nil {
+		return fmt.Errorf("get disk usage: %w", err)
+	}
+	if usage < highDiskUsageThreshold {
+		return nil
+	}
+
+	// We should be in the git repo root at this point - run git gc.
```
maggie-lou, line 2303: Previously we discussed running git gc at the end of every run. That should amortize the costs and make the behavior more predictable. WHat do you think of that approach instead?

bduffany: I'm not sure if the cost is actually amortized - if I run `git gc` in my local repo, it always seems to do a lot of work.

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2272,3 +2285,50 @@ func runCredentialHelper() error {
...
+	log.Infof("Running git gc...")
+	if _, err := git(ctx, os.Stderr, "gc"); err != nil {
```
sluongng, line 2305: Hmm, my concern is that with much garbage left over, git-gc could take a long time to complete. And this operation will block the PR check update, right?

I think we should default running these `git-maintenance` operations unconditionally instead:

```
git maintenance run loose-objects
git maintenance run incremental-repack
git maintenance run pack-refs
```

These commands are lighter weight and running them unconditionally each Invocation will divide out the total cleanup cost, avoiding 1 big blocking gc run (which, for larger repos, could cause OOM during packfile creation).

The caveat is that these are only available starting from Git 2.31.0

bduffany: > this operation will block the PR check update, right?

No, this cleanup is run after the workflow BES stream is closed, and the git statuses are published by the BES server.

> I think we should default running these git-maintenance operations unconditionally instead

These commands seem to behave better when running multiple times in a row, so running unconditionally SGTM (updated the PR)

sluongng (approved):
LGTM

maggie-lou (approved):
Small comments but overall LGTM

maggie-lou (approved):

PR #6833, opened 2024-06-14
Tweak CLI release workflow name

bduffany:
The current name looks like a workflow job ID rather than a workflow name, which had me confused for a second

**Related issues**: N/A


sluongng (approved):
LGTM

PR #6839, opened 2024-06-14
update flake-related target queries to what we actually need.

jdhollen:
- add a time series query for target flake stats
- remove pagination from target stats query (just not worth it for now since the query already sorts)
- add repo param so that two repos can have the same targets without us getting crossed up.
- add a few extra pieces of info to flake samples.

**Related issues**: N/A

### DIFF
--- a/proto/target.proto
+++ b/proto/target.proto
@@ -224,14 +224,11 @@ message GetTargetStatsRequest {
   // flakiest tests in the last 30 days.
   repeated string labels = 2;
 
-  // If many/all labels are requested, there may be multiple pages of results.
-  string page_token = 3;
+  // If specified, stats will be restricted to invocations in this repo.
+  string repo = 3;
 }
 
-message TargetStats {
-  // The target label that these stats are for.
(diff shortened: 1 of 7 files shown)


proto/target.proto
```diff
@@ -249,17 +246,51 @@ message TargetStats {
...
+  // If specified, stats will be restricted to invocations in this repo.
+  string repo = 3;
+
+  // If many/all labels are requested, there may be multiple pages of results.
+  string page_token = 4;
+}
+
+message DailyTargetStats {
+  // YYYY-MM-DD
+  string date = 1;
```
iain-macdonald, line 282: maybe add a TODO to use a Real Date Type™

iain-macdonald (approved):
LGTM, Approval

proto/target.proto
jdhollen: filed https://github.com/buildbuddy-io/buildbuddy-internal/issues/3533 since we have this in a few places.

PR #6847, opened 2024-06-17
Add custom resource client/server configuration

bduffany:
Allows configuring and requesting custom resources, but does not actually implement any scheduling logic yet.

Client config looks like this:

```
exec_properties = {
  "resources:foo": "3.14",
}
```

Executor config looks like this:

```
executor:
  custom_resources:
    - name: foo
      value: 8
```

**Related issues**: N/A


iain-macdonald (approved):
LGTM, Approval

PR #6851, opened 2024-06-17
Implement scheduling logic for custom resources

bduffany:
* Don't assign tasks to executors unless the executors satisfy the custom resource requirements
* Incorporate custom resource requirements into `canFit` logic in priority task scheduler

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3491


enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -395,26 +416,49 @@ func (q *PriorityTaskScheduler) stats() string {
...
-		if res.GetTaskSize().GetEstimatedMemoryBytes() == 0 {
-			log.CtxWarningf(q.rootContext, "Scheduling another unknown size task. THIS SHOULD NOT HAPPEN! res: %+v", res)
-		} else {
-			log.CtxInfof(q.rootContext, "Scheduling another task of size: %+v", res.GetTaskSize())
+	for _, r := range size.GetCustomResources() {
+		used, ok := q.customResourcesUsed[r.GetName()]
+		if !ok {
+			// The scheduler should never send us tasks that require resources
+			// we haven't set up in the config.
+			alert.UnexpectedEvent("missing_custom_resource", "Task requested custom resource %q which is not configured for this executor", r.GetName())
```
iain-macdonald, line 443: How is this going to be configured on the client side? Mostly want to make sure that if clients are experimenting / have typos in their repos then we're not getting alerted.

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -395,26 +416,49 @@ func (q *PriorityTaskScheduler) stats() string {
...
+		if customResource(r.GetValue()) > available {
+			return false
 		}
 	}
-	return willFit
+
+	// The scheduler should prevent CPU/memory requests that are <= 0.
+	// Alert if we get a task like this.
+	if size.GetEstimatedMemoryBytes() <= 0 {
+		alert.UnexpectedEvent("invalid_task_memory", "Requested memory %d is invalid", size.GetEstimatedMemoryBytes())
```
iain-macdonald, line 455: ditto here, isn't this just a client-specified string? (and below)

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -508,3 +553,16 @@ func (q *PriorityTaskScheduler) GetQueuedTaskReservations() []*scpb.EnqueueTaskR
...
 	return q.q.GetAll()
 }
+
+// customResourceCount represents custom resource values in such a way that
+// floating point rounding errors are not accumulated as values are added and
+// subtracted over time.
+type customResourceCount int64
+
+func customResource(value float32) customResourceCount {
+	return customResourceCount(value * 1e6)
```
iain-macdonald, line 563: It's been a while since I my computer architecture class, but IIRC multiplying floating point numbers doesn't affect their susceptibility to loss-of-precision because it's basically just changing the exponent, not adding more bits for the mantissa (does that sound correct?).

Can we get away with requiring that these be specified as integers for the time being?

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
bduffany: If the client asks for CPU or memory < 0, then it shouldn't make it this far due to the logic [here](https://github.com/buildbuddy-io/buildbuddy/blob/43174a2ff85e3047ca5860f040d5e004b2e3d850/enterprise/server/tasksize/tasksize.go#L399)

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
bduffany: the ship might have sailed on requiring these to be integers because we wanted these to work the same as Bazel's [`local_extra_resources`](https://bazel.build/reference/command-line-reference#flag--local_extra_resources) which uses floats for the values.

re. this multiplication, some loss of precision is OK - we just need the conversion from float to int to be deterministic, so that we don't accumulate floating point errors when adding and subtracting the same resource values over time.

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
bduffany: The task shouldn't make it this far if it asks for custom resources that the executor doesn't have, due to the logic [here](https://github.com/buildbuddy-io/buildbuddy/pull/6851/files#diff-fe896f3f44295b27c004e50a48cca5068273a3df48b98059bc62decb3c6df0caR504)

enterprise/server/scheduling/priority_task_scheduler/priority_task_scheduler.go
```diff
@@ -4,6 +4,7 @@ import (
 	"container/list"
 	"context"
 	"flag"
+	"fmt"
```
iain-macdonald, line 7: unused?

bduffany: woops, removed

iain-macdonald (approved):

PR #6853, opened 2024-06-17
Use `git gc --auto` instead of `git maintenance`

bduffany:
The CI runner image doesn't have a recent enough version of `git`: https://buildbuddy.buildbuddy.dev/invocation/3e19735b-975c-4436-9129-cd009d836dc2?role=CI_RUNNER&days=7&branch=master

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -2345,10 +2345,9 @@ func reclaimDiskSpace(ctx context.Context, log *buildEventReporter) error {
 }
 
 func runGitMaintenance(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


sluongng (approved):
hmm our 20.04 image seems to have some old packages

LGTM

maggie-lou:
Sorry for the delayed comment - from your last PR, you mentioned that `git gc` seems to take the same amount of time, even when run multiple times. Do we want to only run it when disk usage is particularly high (like you'd had it before)?

bduffany:
@maggie-lou The `--auto` flag seems to improve that behavior - I was running it without that flag when I was testing before

PR #6857, opened 2024-06-17
firecracker: mount only cgroup v2 in the guest instead of hybrid v1+v2

bduffany:
`crun` doesn't appear to be fully compatible with a "hybrid" cgroup setup (v1+v2) - when running the OCI runtime tests on Firecracker, I'm getting the error `cgroups in hybrid mode not supported` ([invocation link](https://buildbuddy.buildbuddy.io/invocation/a3f929c0-f756-4cf0-89a7-ddb9e37b6407?target=%2F%2Fenterprise%2Fserver%2Fremote_execution%2Fcontainers%2Fociruntime%3Aociruntime_test&targetStatus=6)) which comes from the code [here](https://github.com/containers/crun/blob/f44da38333321335611d45638401e99f5f9548f2/src/libcrun/cgroup.c#L314-L326)

This PR adds a flag to only enable cgroup2 within the guest. Since modern distros have ditched the hybrid setup in favor of v2-only, I would like to enable this everywhere after some testing in dev with docker-in-firecracker, so I went with an executor-level flag rather than a platform property.

**Related issues**: N/A


fmeum (approved):

tylerwilliams (approved):
long live v2

PR #6868, opened 2024-06-18
OCI runtime: implement Run()

bduffany:
Implements `Run`:

https://github.com/buildbuddy-io/buildbuddy/blob/de6e943b0adfe3b6c3158990b160883e0a6abd7e/enterprise/server/remote_execution/container/container.go#L200-L205

**Related issues**: N/A


fmeum:
I also saw this failure in a sandboxed run locally unless I enable pseudo terminal support in the Bazel sandbox. Maybe there is a way to not fail and just skip this mount?

bduffany:
@fmeum I just removed the `devpts` mount until we can figure out how to do it properly. In podman, it looks like the mount gets set up differently depending on certain conditions: https://github.com/containers/podman/blob/b8d95a5893572b37c8257407e964ad06ba87ade6/pkg/specgen/generate/oci_linux.go#L141-L173

fmeum (approved):

PR #6869, opened 2024-06-18
ci_runner: cleanup previously setup repo

sluongng:
Clean up well known files that could cause subsequent git operations to
fail.

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -1633,6 +1633,9 @@ func (ws *workspace) setup(ctx context.Context) error {
 		if err := os.Chdir(repoDirName); err != nil {
 			return status.WrapError(err, "cd")
 		}
+		if err := ws.cleanup(ctx); err != nil {
+			return err
+		}
 		err := ws.sync(ctx)
 		if err == nil {
 			return nil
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1690,6 +1693,27 @@ func (ws *workspace) applyPatch(ctx context.Context, bsClient bspb.ByteStreamCli
 	return nil
 }
 
+var gitTroubles = []string{
+	"index.lock",
+	"shallow.lock",
+	"HEAD.lock",
+	"hooks/post-checkout",
+	"config.lock",
```
bduffany, line 1701: how certain are we that it's safe to delete these files? could there be some background process running that's using these? And if we delete these lockfiles could those background processes wind up clobbering the git repo state?

bduffany (changes requested):

enterprise/server/cmd/ci_runner/main.go
siggisim: I'd prefer to wait for us having an example of this being an issue before checking this in. Maggie investigated the hang and didn't see any of these files.

sluongng:
closing this for now as it's not a problem currently.

PR #6873, opened 2024-06-18
Add instance name back to snapshot-specific key

bduffany:
We probably should remove `instance_name` from the snapshot key proto, since the key is basically the "digest" part of the resource name, and the instance name is normally separate from the digest. But that will be a pretty big refactor, so just add the instance name back to the snapshot-specific key to hold us over until then.

**Related issues**: N/A


maggie-lou (approved):
Just a reminder for my own memory: 
When reading the snapshot, we were using the instance name from the key (which was empty before this PR) https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/snaploader/snaploader.go#L399

When saving the snapshot, we were using the instance name from the master snapshot key: 
https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/snaploader/snaploader.go#L687

PR #6874, opened 2024-06-18
[RB] Support multiple bazel commands in remote runner request

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -761,7 +761,7 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 			CommitSha: repoConfig.CommitSHA,
 			Branch:    repoConfig.Ref,
 		},
-		BazelCommand:   strings.Join(bazelArgs, " "),
+		BazelCommands:  []string{strings.Join(bazelArgs, " ")},
 		Os:             reqOS,
 		Arch:           reqArch,
 		ContainerImage: *containerImage,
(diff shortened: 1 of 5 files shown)


siggisim (commented):
I think the two requests we're going to get if we add this are:
- I want the option to run my bazel commands to run in parallel instead of serial
- I want the option to run this command unconditionally (i.e. regardless of whether the previous step passed / failed)

I think we probably want to come up with answers on how (or if) we'd support those before we settle on this runner.proto change.

Also want to make sure we don't complicate the simple api too much (I just want to run a single bazel command) to accommodate these more advanced cases.

maggie-lou:
@siggisim updated the proto (not the code yet until we reach a decision) to take a struct instead, so it's more flexible. i.e. We could add a field like `continue_if_failed`. The first case - running multiple bazel commands on the same repo - isn't supported anyway, right?

bduffany:
> I want the option to run my bazel commands to run in parallel instead of serial

I think this could be accomplished with parallel requests to the API on the client side - we'd have to schedule multiple invocations anyway

> We could add a field like continue_if_failed

I guess it depends how people are using these - I guess typically people might have some setup, one or more build/test commands, then some "teardown". If there is more than one build/test command then the `continue_if_failed` isn't expressive enough I think, because we want to jump to the teardown step on failure rather than just continue.

If people want to do stuff like this, I wonder if we should reconsider the "steps" API that would allow arbitrary shell code.

siggisim:
> @siggisim updated the proto (not the code yet until we reach a decision) to take a struct instead, so it's more flexible. i.e. We could add a field like `continue_if_failed`. 

I don't hate this, but I defer to @tylerwilliams for API design stuff because he's better at it haha.

The one thing I want to make sure about is that we don't make this too complicated for the simple use-case. Like I'd like to be able to have a simple curl command that looks like (I know it's already more complicated than this - but this is my dream haha and every new nested struct we add makes this more complicated):
```
curl -v  -d '{"command": "test //...", repo: "github.com/buildbuddy-io/buildbuddy"}' -H 'x-buildbuddy-api-key: XXX'  -H 'Content-Type: application/json' https://app.buildbuddy.io/api/v1/Run
```

> The first case - running multiple bazel commands on the same repo - isn't supported anyway, right?

Was thinking for the parallel case we could spawn multiple runners.

maggie-lou:
Closing to prioritize allowing arbitrary shell commands

PR #6879, opened 2024-06-20
write out test target cached-ness to db

jdhollen:
**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/target_tracker/BUILD
+++ b/server/build_event_protocol/target_tracker/BUILD
@@ -27,8 +27,17 @@ go_library(
 
 go_test(
     name = "target_tracker_test",
-    size = "small",
     srcs = ["target_tracker_test.go"],
+    exec_properties = {
+        "test.workload-isolation-type": "firecracker",
+        "test.init-dockerd": "true",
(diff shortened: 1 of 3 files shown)


jdhollen:
I'm not sure what the "preferred" way to submit this is--with the defaults, i think i can split it up and just put the schema change in first?  happy to split that up if that's what we like, since it'll make cherrypicking the schema change a little easier.

tylerwilliams (approved):
This LGTM, I'll let lulu comment on any schema upgrade or other concerns

luluz66:
> I'm not sure what the "preferred" way to submit this is--with the defaults, i think i can split it up and just put the schema change in first? happy to split that up if that's what we like, since it'll make cherrypicking the schema change a little easier.

Yeah. change lgtm. prefer schema change to be a seperate pr just in case.

jdhollen:
> Yeah. change lgtm. prefer schema change to be a seperate pr just in case.

Sure sure, https://github.com/buildbuddy-io/buildbuddy/pull/6890, will plop this one in after.

luluz66 (approved):

PR #6880, opened 2024-06-20
Fix base branch ref in prettier.sh

bduffany:
The ref `master` won't exist after re-running a PR workflow from a clean workspace.

**Related issues**: N/A


maggie-lou (approved):
Thanks for fixing

PR #6882, opened 2024-06-20
Fix race condition in remote bazel integration test

maggie-lou:
**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3525

Fixes the following race conditions:
1. The health check initialized in remote bazel was trying to write to stdout at the same time as bazelisk was trying to redirect it. Removed the healthcheck, which is not needed for cli
2. t.Cleanup functions are run in LIFO order. We spun up a local app server (has a cleanup function to stop the server) and then used its grpc address to set flags `app.events_api_url` and `app.cache_api_url` (have cleanup functions to unset the flags). Because the flag cleanup ran first, the server was still using the flags as they were being unset. I refactored the rbetest cleanup code

### DIFF
--- a/enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
+++ b/enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
@@ -187,7 +187,7 @@ func TestWithPrivateRepo(t *testing.T) {
 		// to setup than a firecracker runner
 		"--runner_exec_properties=workload-isolation-type=none",
 		"--runner_exec_properties=container-image=",
-		"build",
+		"run",
 		":hello_world",
 		"--noenable_bzlmod",
 		fmt.Sprintf("--remote_header=x-buildbuddy-api-key=%s", env.APIKey1)})
@@ -206,10 +206,16 @@ func TestWithPrivateRepo(t *testing.T) {
 		Query:          &inpb.InvocationQuery{GroupId: env.GroupID1},
 	})
 	require.NoError(t, err)
-	// We expect 1 invocation for the ci_runner, and 1 invocation for the :hello_world
-	// build. Either should contain the log line we're looking for
+
 	require.Equal(t, 2, len(searchRsp.GetInvocation()))
-	invocationID := searchRsp.Invocation[0].InvocationId
+	// Find outer invocation because it will contain run output
+	var inv *inpb.Invocation
+	for _, i := range searchRsp.GetInvocation() {
+		if i.GetRole() == "HOSTED_BAZEL" {
+			inv = i
+		}
+	}
+	invocationID := inv.InvocationId
 
 	logResp, err := bbClient.GetEventLogChunk(ctx, &elpb.GetEventLogChunkRequest{
 		InvocationId: invocationID,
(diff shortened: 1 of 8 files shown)


enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
```diff
@@ -248,6 +249,13 @@ func runLocalServerAndExecutor(t *testing.T, githubToken string) (*rbetest.Env,
 	executors := env.AddExecutors(t, 1)
 	require.Equal(t, 1, len(executors))
 	flags.Set(t, "executor.enable_bare_runner", true)
+
+	t.Cleanup(func() {
+		for _, e := range executors {
+			env.RemoveExecutor(e)
+		}
```
bduffany, line 256: hm how does removing the executors fix the race condition? wonder if rbetest should do this automatically, since we already call ShutdownBuildBuddyServers automatically

bduffany (approved):

enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
bduffany: actually it looks like `NewRBETestEnv` does register some executor shutdown logic as a cleanup func already, but not exactly RemoveExecutor - any idea what the difference is that makes this code work?

maggie-lou: Hmm that's a good point I'd missed that. I'm wondering if it's because `executor.env.GetHealthChecker().WaitForGracefulShutdown()` was in the background instead of the foreground. Let me test that out

maggie-lou: Added more details to the commit description

maggie-lou:
Running the flaky tests 100x: https://app.buildbuddy.io/invocation/3f8f8710-8267-4e9e-a467-15236f3332e3?target=%2F%2Fenterprise%2Fserver%2Ftest%2Fintegration%2Fremote_bazel%3Aremote_bazel_test&targetStatus=5

enterprise/server/test/integration/remote_execution/rbetest/rbetest.go
```diff
@@ -699,6 +682,18 @@ func (r *Env) AddBuildBuddyServerWithOptions(opts *BuildBuddyServerOptions) *Bui
...
 	r.updateAppProxy()
+
+	grpcAddress := fmt.Sprintf("grpc://localhost:%d", server.GRPCPort())
+	u, err := url.Parse(grpcAddress)
+	require.NoError(t, err)
+
+	// Note: flags.Set registers a t.Cleanup function to clean up flags. Make sure
+	// that happens after the servers shut down and are no longer using the flags
+	flags.Set(t, "app.events_api_url", *u)
+	flags.Set(t, "app.cache_api_url", *u)
```
bduffany, line 693: instead of pointing these at an individual app, you can point them at the app load balancer address (`Env.AppProxy.GRPCTarget()`). I think we could do that in `NewRBETestEnv()` right after we create the load balancer, and then hopefully that would let us revert the other changes to how we're doing cleanup?

maggie-lou: Nice - that simplifies things a lot

bduffany (approved):

PR #6890, opened 2024-06-21
schema changes to store whether or not a test target was cached

jdhollen:
default value is false (i.e., not cached), which is effectively a no-op with prod.  there isn't a reasonable backfill path.

**Related issues**: N/A


tylerwilliams (approved):
reviewing against my will

PR #6891, opened 2024-06-21
OCI runtime: add standard image pull optimizations

bduffany:
Adds the usual optimizations to avoid excessive requests to the remote image registry:
* Implement `IsImageCached` so that `container.PullImageIfNecessary` can skip the call to `Pull` most of the time
* Dedupe pulls for the same image ref, since the executor would otherwise be making an excessive amount of requests to the registry when assigned several actions that need the same image within a short duration

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -793,7 +793,10 @@ type ImageStore struct {
 }
 
 func NewImageStore(layersDir string) *ImageStore {
-	return &ImageStore{layersDir: layersDir}
+	return &ImageStore{
+		layersDir:    layersDir,
+		cachedLayers: map[string][]ctr.Layer{},
+	}
 }
 
 // Pull downloads and extracts image layers to a directory, skipping layers
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -773,9 +783,69 @@ func layerPath(layersDir string, hash ctr.Hash) string {
...
+			return nil, res.Err
+		}
+		return res.Val.([]ctr.Layer), nil
+	}
+}
+
+// CachedLayers returns references to the cached image layers if the image
+// has been pulled. The second return value indicates whether the image has
+// been pulled - if false, the returned slice of layers will be nil.
+func (s *ImageStore) CachedLayers(imageName string) (layers []ctr.Layer, ok bool) {
```
iain-macdonald, line 842: This doesn't support checking for updated remote versions of tagged images, do you mind adding a TODO that explains that?

iain-macdonald, line 842: looks like the first return value is unused

iain-macdonald (approved):

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: for tagged images, we should be calling `PullImage` whenever the local image cache token expires (in `container.PullImageIfNecessary`), and the cached image layers should get updated here, I think

bduffany: I wound up using it in https://github.com/buildbuddy-io/buildbuddy/pull/6896 which is based on this one - would prefer to keep it in this PR if you don't mind, to avoid having to restack the PRs

fmeum (approved):

PR #6894, opened 2024-06-21
Make firecracker NAT a little smarter.

vadimberezniker:
If the default route has a source IP, use that for NAT instead of letting the kernel pick the IP.

Our bare metal hosts have two IPs and without this change MASQUERADE would always pick the private IP which cannot talk to the internet.

**Related issues**: N/A


maggie-lou (approved):

vadimberezniker:
This is solved in a better way in https://github.com/buildbuddy-io/buildbuddy-internal/pull/3553

PR #6896, opened 2024-06-21
Fix some bugs with OCI image handling

bduffany:
* Don't mount empty layers - if an image has multiple empty layers, then the overlay mount will fail, due to the same directory appearing twice in `lowerdir` (which is apparently invalid)
* Fix `layerDigests` logic - we were previously relying on `PullImage` to always be called, which sets `c.layerDigests`. But `PullImage` can be skipped if the image is already cached. Instead, compute the layer digests on the fly, using `imageStore.CachedLayers`

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -169,7 +169,6 @@ type ociContainer struct {
 	workDir string
 
 	imageRef         string
-	layerDigests     []ctr.Hash
 	overlayfsMounted bool
 }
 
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -382,8 +371,27 @@ func (c *ociContainer) createRootfs(ctx context.Context) error {
 
 	// Create an overlayfs with the pulled image layers.
 	var lowerDirs []string
-	for _, d := range c.layerDigests {
-		lowerDirs = append(lowerDirs, layerPath(c.layersRoot, d))
+	layers, ok := c.imageStore.CachedLayers(c.imageRef)
+	if !ok {
+		return fmt.Errorf("bad state: attempted to create rootfs before pulling image")
```
iain-macdonald, line 376: Change to InternalError?

iain-macdonald (approved):
LGTM, Approval

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: I think it makes sense to have this be a retryable error for now - in the places where we're calling this func, we're always wrapping with `UnavailableError` currently.

PR #6900, opened 2024-06-21
Add workspace proto

siggisim:
Changes from last proposal:
- changes is now a list instead of a map
- consolidated Node / Change into a single `Node` message
- simpler `Repo` message that doesn't include fields that we don't need

### DIFF
--- a/proto/workspace.proto
+++ b/proto/workspace.proto
@@ -34,6 +34,9 @@ message SaveWorkspaceRequest {
 message SaveWorkspaceResponse {
   // The response context.
   context.ResponseContext response_context = 1;
+
+  // The name of the workspace that was saved.
+  string workspace = 2;
 }
 
 message GetWorkspaceDirectoryRequest {
@@ -57,7 +60,8 @@ message GetWorkspaceDirectoryResponse {
   // The directory that was requested.
   Node directory = 3;
 
-  // The child nodes (files and directories) contained in this directory.
+  // The first level of child nodes (files and directories) contained in this
+  // directory.
   repeated Node child_nodes = 2;
 }
 
--- a/proto/workspace.proto
+++ b/proto/workspace.proto
@@ -16,9 +16,10 @@ message GetWorkspaceResponse {
   // The response context.
   context.ResponseContext response_context = 1;
 
-  // The requested workspace. 
+  // The requested workspace.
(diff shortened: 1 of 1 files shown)


proto/workspace.proto
```diff
@@ -0,0 +1,140 @@
...
+
+message SaveWorkspaceRequest {
+  // The request context.
+  context.RequestContext request_context = 1;
+
+  // The workspace to save.
+  Workspace workspace = 2;
+}
+
+message SaveWorkspaceResponse {
```
tylerwilliams, line 34: 100% not necessary, but you could return name here if you wanted to more easily match req/rsp

proto/workspace.proto
```diff
@@ -0,0 +1,140 @@
...
+
+message GetWorkspaceDirectoryResponse {
+  // The response context.
+  context.ResponseContext response_context = 1;
+
+  // The directory that was requested.
+  Node directory = 3;
+
+  // The child nodes (files and directories) contained in this directory.
+  repeated Node child_nodes = 2;
```
tylerwilliams, line 61: I'm cool with starting with this, but we might want to watch out b/c GetTree which is similar, has to be streaming to accommodate the flattened list of files under a parent directory.

proto/workspace.proto
```diff
@@ -0,0 +1,140 @@
...
+}
+
+message GetWorkspaceDirectoryResponse {
+  // The response context.
+  context.ResponseContext response_context = 1;
+
+  // The directory that was requested.
+  Node directory = 3;
+
+  // The child nodes (files and directories) contained in this directory.
```
tylerwilliams, line 60: nit: specify if this is all children or just one level deep?

tylerwilliams (approved):
LGTM (couple minor nits)

proto/workspace.proto
siggisim: Good idea, didn't think about matching these up. Done.

proto/workspace.proto
siggisim: Specified - just the first level.

proto/workspace.proto
siggisim: Specified this in the comment, but since this is just the first level of children - I'm hoping we won't have to do any streaming.

PR #6904, opened 2024-06-24
Disable runner-recycling for kythe actions

tylerwilliams:


enterprise/server/workflow/service/service.go
```diff
@@ -1254,6 +1255,19 @@ func (ws *workflowService) createActionForWorkflow(ctx context.Context, wf *tabl
...
+
+	if enableRunnerRecycling {
+		// Reuse the container/VM for the CI runner across executions if
+		// possible, and also keep the git repo around so it doesn't need to be
+		// re-cloned each time.
+		cmd.Platform.Properties = append(cmd.Platform.Properties, []*repb.Platform_Property{
+			{Name: "recycle-runner", Value: "true"},
+			{Name: "runner-recycling-max-wait", Value: (*workflowsRunnerMaxWait).String()},
+			{Name: "preserve-workspace", Value: "true"},
+			{Name: "use-self-hosted-executors", Value: useSelfHostedExecutors},
```
maggie-lou, line 1267: Do we want to maintain the self-hosted-executor prop regardless?

maggie-lou (commented):
Are you nervous about the build time of the kythe workflow if it's always on a clean runner?

enterprise/server/workflow/service/service.go
tylerwilliams: whoops, yeah, fixing.

bduffany, line 1267: this one doesn't look recycling related? otherwise LGTM

bduffany (approved):

enterprise/server/workflow/service/service.go
bduffany: woops Maggie already mentioned this

tylerwilliams:
> Are you nervous about the build time of the kythe workflow if it's always on a clean runner?

It's slow, but kythe annotations don't need to be realtime so that's not an issue. I think the huge amount of data possibly generated (it's 32GB on our repo alone) means putting all this in cache right now is not super fast anyway.

enterprise/server/workflow/service/service.go
tylerwilliams: Done.

PR #6907, opened 2024-06-24
[ci_runner] Support arbitrary commands in workflows

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -1,4 +1,10 @@
 actions:
+  - name: Maggie
+    container_image: ubuntu-20.04
+    steps:
+      - run: bazel help
+      - run: bazelisk help
+      - run: echo $PATH
   - name: Test
     container_image: ubuntu-20.04
     triggers:
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -1,10 +1,4 @@
 actions:
-  - name: Maggie
-    container_image: ubuntu-20.04
-    steps:
-      - run: bazel help
-      - run: bazelisk help
-      - run: echo $PATH
   - name: Test
     container_image: ubuntu-20.04
     triggers:
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -676,6 +676,17 @@ func run() error {
 			return status.WrapError(err, "failed to extract bazelisk")
 		}
 		*bazelCommand = bazeliskPath
+
+		if _, err := runBashCommand(ctx, fmt.Sprintf("sudo ln -s %s /usr/local/bin/bazel", *bazelCommand), nil, "", io.Discard); err != nil {
+			if !strings.Contains(err.Error(), "File exists") {
+				return status.InternalErrorf("could not create a symlink for bazel: %s", err.Error())
+			}
+		}
+		if _, err := runBashCommand(ctx, fmt.Sprintf("sudo ln -s %s /usr/local/bin/bazelisk", *bazelCommand), nil, "", io.Discard); err != nil {
+			if !strings.Contains(err.Error(), "File exists") {
+				return status.InternalErrorf("could not create a symlink for bazel: %s", err.Error())
+			}
+		}
 	}
 
 	if *shutdownAndExit {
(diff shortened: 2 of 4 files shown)


buildbuddy.yaml
```diff
@@ -1,4 +1,21 @@
...
+      - args: which bazelisk
+        continue_on_failure: true
+      - args: |
+          for i in {1..1}; do
+            j=99;
+            echo "j is $j";
+            echo "i is $i";
+            echo "$USER";
+          done
+        continue_on_failure: true
```
bduffany, line 18: instead of `continue_on_failure` can we have the user write this logic in the shell script? like `some_command_that_might_fail || true`

In general, I would prefer to leave control flow out of the YAML, because it's less flexible than bash, while increasing the complexity of the schema

Also just for familiarity, it would be nice to rename `commands` -> `steps` and `args` -> `run`

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -676,6 +676,17 @@ func run() error {
...
+		if _, err := runBashCommand(ctx, fmt.Sprintf("sudo ln -s %s /usr/local/bin/bazel", *bazelCommand), nil, "", io.Discard); err != nil {
+			if !strings.Contains(err.Error(), "File exists") {
+				return status.InternalErrorf("could not create a symlink for bazel: %s", err.Error())
+			}
+		}
+		if _, err := runBashCommand(ctx, fmt.Sprintf("sudo ln -s %s /usr/local/bin/bazelisk", *bazelCommand), nil, "", io.Discard); err != nil {
+			if !strings.Contains(err.Error(), "File exists") {
+				return status.InternalErrorf("could not create a symlink for bazel: %s", err.Error())
+			}
+		}
```
bduffany, line 689: instead of symlinking to /usr/local/bin I think it'd be better to add this dir to PATH (and symlink bazel to bazelisk)

```
// Symlink bazel to bazelisk
if err := os.Symlink(bazeliskBinaryName, filepath.Join(rootDir, bazelBinaryName)); err != nil {
  // ...
}
// Update PATH to include the root dir
os.Setenv("PATH", rootDir + ":" + os.Getenv("PATH"))
```

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -993,6 +1004,70 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 		}
 	}
 
+	// TODO(Maggie): Emit BES events for each bazel command
+	for i, step := range action.Steps {
+		// TODO(Maggie): Provision one sub-dir per bazel command
```
bduffany, line 1009: I think the current approach here seems OK? (one subdir per bash command)

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2131,6 +2206,21 @@ func gitRemoteName(repoURL string) string {
...
 
+func runBashCommand(ctx context.Context, cmd string, env map[string]string, dir string, outputSink io.Writer) (string, *commandErr) {
+	if err := printCommandLine(outputSink, cmd); err != nil {
+		return "", &commandErr{err, ""}
+	}
+
+	var buf bytes.Buffer
+	w := io.MultiWriter(outputSink, &buf)
+
+	if err := runCommand(ctx, "bash", []string{"-c", cmd}, env, dir, w); err != nil {
```
bduffany, line 2217: add `-eo pipefail` since it's a better default and also more familiar:

```suggestion
	if err := runCommand(ctx, "bash", []string{"-eo", "pipefail", "-c", cmd}, env, dir, w); err != nil {
```

enterprise/server/test/integration/ci_runner/ci_runner_test.go
```diff
@@ -324,6 +324,54 @@ func singleInvocation(t *testing.T, app *app.App, res *result) *inpb.Invocation
...
+actions:
+  - name: "Show bazel version"
+    triggers:
+      push: { branches: [ master ] }
+      pull_request: { branches: [ master ] }
+    steps:
+      - run: |
+          for i in {1..2}; do
+            echo "Loop $i: ";
+            ` + bazelPath + ` version;
```
bduffany, line 346: call `bazel` here to assert that it's available in PATH?

```suggestion
            bazel version;
```

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1960,36 +2035,36 @@ func (ws *workspace) fetch(ctx context.Context, remoteURL string, refs []string,
 	return nil
 }
 
-type gitError struct {
+type commandErr struct {
```
bduffany, line 2038: nit: use unabbreviated name since it's a type

```suggestion
type commandError struct {
```

enterprise/server/cmd/ci_runner/main.go
maggie-lou: I wasn't sure if the artifacts had to be in separate dirs for how we show them in the UI, so added a TODO so I didn't forget to look into it. Anyway just removed it

bduffany: Ah right, we render them as `bazel_commands[0]`. Maybe we could change these to `steps[0]` etc. once we are migrated

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -676,6 +676,21 @@ func run() error {
 			return status.WrapError(err, "failed to extract bazelisk")
 		}
 		*bazelCommand = bazeliskPath
+
+	}
+
+	// Symlink bazel to `bazelCommand` so that it can be easily invoked
+	if err := os.Symlink(*bazelCommand, filepath.Join(rootDir, bazelBinaryName)); err != nil {
+		if !strings.Contains(strings.ToLower(err.Error()), "file exists") {
```
bduffany, line 684: nit: it'd be slightly more robust to remove any existing symlink first - then you don't have to check for "file exists" and you also have a stronger guarantee that the resulting symlink has the correct target in the case where it already exists

```
bazelPath := filepath.Join(rootDir, bazelBinaryName)
if err := os.RemoveAll(bazelPath); err != nil {
  return status.WrapError(err, "remove existing bazel binary")
}
if err := os.Symlink(*bazelCommand, bazelPath); err != nil {
  return status.WrapError(err, "symlink bazel to bazelisk")
}
```

buildbuddy.yaml
```diff
@@ -1,4 +1,10 @@
 actions:
+  - name: Maggie
+    container_image: ubuntu-20.04
+    steps:
+      - run: bazel help
+      - run: bazelisk help
+      - run: echo $PATH
```
bduffany, line 7: rm?

```suggestion
```

bduffany (approved):

buildbuddy.yaml
maggie-lou: Oops yes thanks

PR #6908, opened 2024-06-24
[pebble cache] close the iter when e.samples is blocked

luluz66:
**Related issues**: N/A
Right now when e.samples channel is full, the iter can stay open full for a long
time.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -2547,7 +2547,9 @@ func (e *partitionEvictor) generateSamplesForEviction(quitChan chan struct{}) er
 	// We update the iter variable later on, so we need to wrap the Close call
 	// in a func to operate on the correct iterator instance.
 	defer func() {
-		iter.Close()
+		if iter != nil {
+			iter.Close()
+		}
 	}()
 
 	totalCount := 0
@@ -2593,7 +2595,9 @@ func (e *partitionEvictor) generateSamplesForEviction(quitChan chan struct{}) er
 			if err != nil {
 				return err
 			}
-			iter.Close()
+			if iter != nil {
+				iter.Close()
+			}
 			iter = newIter
 		}
 		totalCount += 1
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2650,9 +2654,22 @@ func (e *partitionEvictor) generateSamplesForEviction(quitChan chan struct{}) er
 			case e.samples <- sample:
 			case <-quitChan:
 				return nil
+			default:
+				// e.samples is full. Let's close the iter first, so that zombie
+				// tables can be cleaned up by pebble.
+				iter.Close()
+				shouldCreateNewIter = true
+				iter = nil
```
bduffany, line 2662: could `shouldCreateNewIter` be replaced by an `iter == nil` check?

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2650,9 +2654,22 @@ func (e *partitionEvictor) generateSamplesForEviction(quitChan chan struct{}) er
...
+				select {
+				case e.samples <- sample:
+				case <-quitChan:
+					return nil
+				}
```
bduffany, line 2667: If the evictor is not consuming from `e.samples` faster than we're sampling, it seems like we could potentially close the iterator on every loop iteration. Does this seem like it could be an issue? Wonder if the time-based logic here might make sense after all

luluz66: We sample 10 (deletes_per_eviction) or 500 (sample_pool_size) at the same time. The reason e.samples block is when the partition size hasn't reached the threshold; Once the threshold is reached, it should consume faster.

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: There are other conditions where we need to refresh the iter in the for loop.

tylerwilliams (approved):
LGTM, Approval

PR #6910, opened 2024-06-24
fix up loading state on flakes page a bunch.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/enterprise/app/tap/flakes.tsx
+++ b/enterprise/app/tap/flakes.tsx
@@ -294,9 +294,6 @@ export default class FlakesComponent extends React.Component<Props, State> {
       );
     }
 
-    console.log("hey");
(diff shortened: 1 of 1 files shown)


enterprise/app/tap/flakes.tsx
```diff
@@ -284,12 +294,13 @@ export default class FlakesComponent extends React.Component<Props, State> {
       );
     }
 
+    console.log("hey");
```
siggisim, line 297: rm

siggisim (approved):

enterprise/app/tap/flakes.tsx
jdhollen: crap crap crap

PR #6911, opened 2024-06-25
fetch_server: support sha1/sha384/sha512 digest function

hunshcn:


hunshcn:
@sluongng could you take a look?

server/remote_cache/digest/digest.go
```diff
@@ -51,6 +51,14 @@ func init() {
...
 		sizeBytes  int
 	}{
+		{
+			digestType: repb.DigestFunction_SHA512,
+			sizeBytes:  sha512.Size,
+		},
+		{
+			digestType: repb.DigestFunction_SHA384,
+			sizeBytes:  sha512.Size384,
+		},
```
sluongng, line 61: I don't think this is sufficient. We probably gonna need an update in `HashForDigestType` to link the digest function to the actual implementation of Go's `hash.Hash` interface.

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -116,12 +114,12 @@ func (p *FetchServer) FetchBlob(ctx context.Context, req *rapb.FetchBlobRequest)
...
 		var prefix string
 		if qualifier.GetName() == ChecksumQualifier {
-			if strings.HasPrefix(qualifier.GetValue(), sha256Prefix) {
-				checksumFunc = repb.DigestFunction_SHA256
-				prefix = sha256Prefix
-			} else if strings.HasPrefix(qualifier.GetValue(), blake3Prefix) {
-				checksumFunc = repb.DigestFunction_BLAKE3
-				prefix = blake3Prefix
+			for _, digestFunc := range []string{"sha1", "sha256", "sha384", "sha512", "blake3"} {
+				if strings.HasPrefix(qualifier.GetValue(), digestFunc+"-") {
```
sluongng, line 118: Move `digestFunc+"-"` to a variable outside the `if` statement so you would only compute it once.

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -316,6 +314,11 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 		if expectedChecksum != "" && checksumDigestRN.GetDigest().GetHash() != expectedChecksum {
 			return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, checksumDigestRN.GetDigest().Hash, expectedChecksum)
 		}
+		// for next request
+		_, err = cachetools.UploadFile(ctx, bsClient, remoteInstanceName, checksumFunc, tmpFilePath)
+		if err != nil {
+			return nil, status.UnavailableErrorf("failed to add object to cache: %s", err)
+		}
```
sluongng, line 321: Nit: the comment in this section should be updated.

Also, we do need a new test in `fetch_server_test.go` to properly validate this new logic.

sluongng (changes requested):
Much appreciate this PR. I think it's a step toward the right direction, but it's not yet ready.

The PR itself, after adding tests and docs, could be a bit big.
I strongly suggest you split this into 2 smaller PRs: 

- One PR adds writing a cache entry using `checksumFunc`.

- One PR for adding new hash functions support to buildbuddy. I would also stick with SHA512 only as I don't think SHA384 is widely used.

If you don't have the capacity/bandwidth to do this, please do let me know. I could add this to my TODO backlog.

hunshcn:
There are two commit. One is https://github.com/buildbuddy-io/buildbuddy/pull/6924

sluongng (commented):
A few nits left but this looks good to me as well.

server/remote_asset/fetch_server/fetch_server_test.go
```diff
@@ -184,10 +192,9 @@ func TestFetchBlobWithCache(t *testing.T) {
 			ctx, err := prefix.AttachUserPrefixToContext(ctx, te)
 			require.NoError(t, err)
 
-			err = te.GetCache().Set(ctx, digest.NewResourceName(&repb.Digest{
-				Hash:      tc.contentHash,
-				SizeBytes: contentSize,
-			}, "", resource.CacheType_CAS, tc.checksumFunc).ToProto(), []byte("content"))
+			contentDigest, err := digest.Compute(bytes.NewReader([]byte(content)), tc.checksumFunc)
+			require.NoError(t, err)
+			err = te.GetCache().Set(ctx, digest.NewResourceName(contentDigest, "", resource.CacheType_CAS, tc.checksumFunc).ToProto(), []byte(content))
```
bduffany, line 197: @sluongng this might be a pre-existing issue with the current test, but shouldn't we be storing to the cache with `tc.storageFunc` here rather than `tc.checksumFunc`? I thought the checksumFunc is only used when validating the content in the fetch server, and not used for caching purposes at all

server/remote_asset/fetch_server/fetch_server_test.go
```diff
@@ -215,67 +222,77 @@ func TestFetchBlobWithCache(t *testing.T) {
 			}, "", resource.CacheType_CAS, tc.storageFunc).ToProto())
 			require.NoError(t, err)
 			require.True(t, exist)
+
+			exist, err = te.GetCache().Contains(ctx, digest.NewResourceName(&repb.Digest{
+				Hash:      contentDigest.GetHash(),
+				SizeBytes: resp.GetBlobDigest().GetSizeBytes(),
+			}, "", resource.CacheType_CAS, tc.checksumFunc).ToProto())
+			require.NoError(t, err)
+			require.True(t, exist)
```
bduffany, line 231: I don't think we have really tested storing digests in the cache other than blake3 or sha256, so this might open us up to some new behavior that we should test more IMO. I would be fine supporting SHA512 etc. for validation purposes but when storing in the cache I'm not sure we want to start supporting new digest funcs just yet.

server/remote_asset/fetch_server/fetch_server_test.go
sluongng: We are already storing the cache with `storageFunc`.

However, for cases where the `storageFunc != checksumFunc` (i.e. most of rules_js use SHA512 while Bazel runs on SHA256), subsequent FetchBlob requests get no cache hits.

This PR makes it so that we double write: one cache entry with `storageFunc` and one cache entry with `checksumFunc`. So subsequent requests would pull the entry from our cache using `checksumFunc` instead of fetching the blob from the internet.

server/remote_asset/fetch_server/fetch_server_test.go
hunshcn: done.

But what should I do to test more IMO. ( maybe in next pr )

sluongng: The current PR added validations for other checksum types without writing the blob to cache as well as improved our existing test setup. The double-writes to the cache have been removed.

I think this is good enough for merging. We will discuss adding support for caching SHA512/SHA384 blobs internally and decide whether to add it in a separate PR.

sluongng (approved):
Hmm our CI is not well setup for PRs created by external contributors.

I am trying to trigger the remaining required workflows using another PR. https://github.com/buildbuddy-io/buildbuddy/pull/6948

Hopefully, the pipeline results are shared so we can merge this one.

sluongng:
Thanks so much for your contribution @hunshcn 🤝 
The PR has been merged.

PR #6917, opened 2024-06-25
Support hedging merged actions

iain-macdonald:
This PR adds a flag, `remote_execution.action_merging_hedge_count` that causes the execution server to execute actions additionally in the background when new execution requests are received and merged against pending executions, up to a limited number of total in-fight executions. This is done by tracking the number of in-flight executions per action in Redis alongside the other action-merging data, and making the execution server spawn additional executions in the background when actions are hedge-able. A few notes:
- When hedging is enabled, there will be 1 execution spawned per execution request up to `remote_execution.action_merging_hedge_count+1`.
- When hedge executions are run, the ExecutionServer will still return the merged-against execution ID to the client. If that execution is stuck, subsequent executions will be stuck too. The behavior change is that if the first execution is stuck, but a hedged execution succeeds, executions submitted after that success will be able to use the cached result instead of blocking on the stuck execution.
- The default value of `remote_execution.action_merging_hedge_count` is 0, which is the same, no-hedging behavior as we have today. 

There's also a new test in `remote_execution_test` that demonstrates that if an execution is submitted which blocks for a long time, hedged executions will unblock later executions.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/action_merger/action_merger.go
+++ b/enterprise/server/remote_execution/action_merger/action_merger.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
@@ -23,6 +22,11 @@ const (
 	// set to twice the length of `remote_execution.lease_duration` to give a
 	// short grace period in the event of missed leases.
 	DefaultClaimedExecutionTTL = 20 * time.Second
+
+	// Redis Hash keys for storing the canonical execution ID and the pending
+	// execution count.
+	executionIDKey    = "execution-id"
+	executionCountKey = "execution-count"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
This looks p good to me but @vadimberezniker is the expert on this code so I'll defer to him.

enterprise/server/remote_execution/action_merger/action_merger.go
```diff
@@ -26,18 +27,26 @@ const (
...
-		return "", err
+		return "", "", err
 	}
 	downloadString, err := adResource.DownloadString()
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
-	return fmt.Sprintf("pendingExecution/%s%s", userPrefix, downloadString), nil
+	return fmt.Sprintf("pendingExecution/%s%s", userPrefix, downloadString),
```
vadimberezniker, line 47: Instead of separating this information into two separate keys, why don't we create a new hash key that contains both pieces of information?
I was fine with having the second key until I saw the `strings.Replace` call.

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -629,6 +643,18 @@ func (s *ExecutionServer) execute(req *repb.ExecuteRequest, stream streamLike) e
...
+	// in the background.
+	if hedge {
+		action_merger.RecordHedgedExecution(ctx, s.rdb, adInstanceDigest)
+		log.CtxInfof(ctx, "Scheduling new hedged execution for %q for invocation %q", downloadString, invocationID)
+		hedgedExecutionID, err := s.Dispatch(ctx, req)
+		if err != nil {
+			log.CtxWarningf(ctx, "Error dispatching execution for %q: %s", downloadString, err)
+			return err
+		}
+		go s.waitExecution(ctx, &repb.WaitExecutionRequest{Name: hedgedExecutionID}, dummyStream{}, waitOpts{isExecuteRequest: true})
```
vadimberezniker, line 656: Do we need to have this goroutine reading updates? What would happen if we omitted this line?

vadimberezniker, line 656: Can you add a log here that records the hedged execution ID?

vadimberezniker (changes requested):

enterprise/server/remote_execution/action_merger/action_merger.go
iain-macdonald: Good idea, done

enterprise/server/remote_execution/execution_server/execution_server.go
iain-macdonald: Yup, done

iain-macdonald: Interesting, I didn't think it'd work without this but it seems to. Removed.

enterprise/server/remote_execution/action_merger/action_merger.go
```diff
@@ -22,12 +22,21 @@ const (
...
 )
 
 var (
 	enableActionMerging = flag.Bool("remote_execution.enable_action_merging", true, "If enabled, identical actions being executed concurrently are merged into a single execution.")
+	hedgedActionCount   = flag.Int("remote_execution.action_merging_hedge_count", 0, "When action merging is enabled, this flag controls how many additional, 'hedged' attempts an action is run in the background. Note that even hedged actions are run at most once per execution request.")
 )
 
+// Returns the redis key pointing to the hash storing action merging state. The
+// value stored here is a hash containing the canonical (first-submitted)
+// execution ID and the count of executions run for this action (for hedging).
```
vadimberezniker, line 39: You will want to change the name of the key to avoid rollout compatibility issues.

vadimberezniker (approved):

enterprise/server/remote_execution/action_merger/action_merger.go
iain-macdonald: Good call, added a version to both keys for future ease of use

vadimberezniker (approved):

PR #6922, opened 2024-06-25
[raft] rebalance based on replica count.

luluz66:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/raft/driver/driver.go
+++ b/enterprise/server/raft/driver/driver.go
@@ -29,7 +29,7 @@ var (
 
 const (
 	// If a node's disk is fuller than this (by percentage), it is not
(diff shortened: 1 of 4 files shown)


enterprise/server/raft/driver/driver.go
```diff
@@ -28,8 +29,12 @@ var (
 
 const (
 	// If a node's disk is fuller than this (by percentage), it is not
-	// eligible to receive ranges moved from other nodes.
+	// eligible to be used a rebalance or allocation target and we will
```
tylerwilliams, line 32: s/a rebalance/for a rebalance/g

tylerwilliams (approved):

PR #6928, opened 2024-06-26
Install API-Key propagating interceptor prior to auth interceptor

iain-macdonald:
https://github.com/buildbuddy-io/buildbuddy/pull/6788 adds a new authenticator that remotely authenticates gRPC requests. But that relies on the auth headers being present in the outgoing RPC context at the time that the authenticator is invoked, which is not the case with the way this code is currently written. This PR re-shuffles the extra-interceptors installation so that these interceptors are run before the auth interceptor. It also adds a JWT-propagating interceptor so that executors can remotely authenticate using the remote authenticator.

Ultimately, we should probably switch this to something fancier like a builder, but for now this should suffice.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/cache_proxy/cache_proxy.go
+++ b/enterprise/server/cmd/cache_proxy/cache_proxy.go
@@ -129,11 +129,16 @@ func main() {
 }
 
 func startGRPCServers(env *real_environment.RealEnv) error {
-	// Add the API-Key propagating interceptors.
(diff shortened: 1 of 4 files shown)


tylerwilliams (approved):
I think this looks ok but see what @vadimberezniker thinks :)

vadimberezniker (approved):

PR #6937, opened 2024-06-27
.bazelrc: write exec log on Windows

sluongng:
This was previously disabled on Windows because we were writing it to
`/tmp` and on Windows, there is not a fixed tmp path.

We have since switched the log path to our workspace and thus,
compatible with Windows.

Also changed the log to use the correct ZSTD extension, which is `.zst`.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -256,8 +256,7 @@ common --experimental_profile_include_primary_output
 common --noslim_profile
 # Include compact execution log
 # TODO(sluongng): make Bazel writes this to output_base automatically
-common:linux --experimental_execution_log_compact_file=bazel_compact_exec_log.binpb.zstd
-common:macos --experimental_execution_log_compact_file=bazel_compact_exec_log.binpb.zstd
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -256,8 +256,9 @@ common --experimental_profile_include_primary_output
 common --noslim_profile
 # Include compact execution log
 # TODO(sluongng): make Bazel writes this to output_base automatically
-common:linux --experimental_execution_log_compact_file=bazel_compact_exec_log.binpb.zstd
-common:macos --experimental_execution_log_compact_file=bazel_compact_exec_log.binpb.zstd
+common:linux --experimental_execution_log_compact_file=bazel_compact_exec_log.binpb.zst
+common:macos --experimental_execution_log_compact_file=bazel_compact_exec_log.binpb.zst
+common:windows --experimental_execution_log_compact_file=bazel_compact_exec_log.binpb.zst
```
fmeum, line 261: Does this still need to be platform-dependent?

sluongng: Good call. We can make it a generic flag now.

bduffany (approved):

PR #6944, opened 2024-06-27
[ci_runner] Fix github running statuses not being reported

maggie-lou:
I noticed that our github statuses were only being posted when workflows were "Queued" and "Finished", and were not updated when they were "Running". 

This was a regression from https://github.com/buildbuddy-io/buildbuddy/commit/861d44ea3b35fe6e29b5d583d565738edbe3096b where we started publishing the WorkspaceStatus event before the BuildMetadata event. We need the BuildMetadata event to be published first, because the [status reporter checks the role of the invocation](https://github.com/buildbuddy-io/buildbuddy/blob/master/server/build_event_protocol/build_status_reporter/build_status_reporter.go#L73).

**Related issues**: N/A


bduffany (approved):
nice find!

PR #6946, opened 2024-06-27
Enable local caching in byte_stream_server_proxy

iain-macdonald:
This PR contains the changes in https://github.com/buildbuddy-io/buildbuddy/pull/7048 -- review that one first.

This PR makes `byte_stream_server_proxy` use the `byte_stream_server` installed in the environment by `cache_proxy` in https://github.com/buildbuddy-io/buildbuddy/pull/7048 for caching data locally. It attempts to serve reads out of the local bytestreamserver, when the data is available, and writes data that's read through the proxy to the local bytestreamserver. There is a bit of a hack that converts downloaded resource names into uploaded resource names, but otherwise it's mostly just manipulating gRPC streams.

Note that the test is largely copied from `server/remote_cache/byte_stream_server/byte_stream_server_test.go` because that test has pretty good coverage of compression cases that I wanted to keep. I factored out some common code into a testutil.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
+++ b/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
@@ -65,7 +65,9 @@ func (s *ByteStreamServerProxy) Read(req *bspb.ReadRequest, stream bspb.ByteStre
 			}
 		}
 
-		stream.Send(rsp)
+		if err := stream.Send(rsp); err != nil {
+			return err
+		}
 		responseSent = true
 	}
 	return nil
@@ -114,7 +116,6 @@ func (s *ByteStreamServerProxy) readRemote(req *bspb.ReadRequest, stream bspb.By
 		localStream, err := s.local.Write(stream.Context())
 		if err != nil {
 			log.Warningf("error writing to local bytestream server: %s", err)
-			localWriteStream = nil
 		}
 		localWriteStream = &localWriter{
 			ctx:          stream.Context(),
@@ -127,13 +128,13 @@ func (s *ByteStreamServerProxy) readRemote(req *bspb.ReadRequest, stream bspb.By
 
 	for {
 		rsp, err := remoteReadStream.Recv()
-		if err == io.EOF {
+		if err != nil {
 			if localWriteStream != nil {
 				localWriteStream.close()
 			}
-			break
-		}
-		if err != nil {
+			if err == io.EOF {
+				break
+			}
 			return err
 		}
 
--- a/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
+++ b/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
@@ -139,9 +139,17 @@ func (s *ByteStreamServerProxy) readRemote(req *bspb.ReadRequest, stream bspb.By
 		}
 
 		if localWriteStream != nil {
-			localWriteStream.send(rsp.Data)
+			if err := localWriteStream.send(rsp.Data); err != nil {
+				log.Debugf("Error writing locally: %s", err)
+				localWriteStream = nil
+			}
 		}
 		if err = stream.Send(rsp); err != nil {
+			if localWriteStream != nil {
+				if err := localWriteStream.close(); err != nil {
+					log.Debugf("Error closing local write stream: %s", err)
+				}
+			}
 			return err
 		}
 	}
--- a/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
+++ b/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
@@ -6,56 +6,139 @@ import (
 	"io"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
-	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
 	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 
 	bspb "google.golang.org/genproto/googleapis/bytestream"
 )
 
 type ByteStreamServerProxy struct {
-	env         environment.Env
-	localCache  interfaces.Cache
-	remoteCache bspb.ByteStreamClient
+	env    environment.Env
+	local  bspb.ByteStreamClient
+	remote bspb.ByteStreamClient
 }
 
 func Register(env *real_environment.RealEnv) error {
-	byteStreamServer, err := NewByteStreamServerProxy(env)
+	proxy, err := New(env)
 	if err != nil {
 		return status.InternalErrorf("Error initializing ByteStreamServerProxy: %s", err)
 	}
-	env.SetByteStreamServer(byteStreamServer)
+	env.SetByteStreamServer(proxy)
 	return nil
 }
 
-func NewByteStreamServerProxy(env environment.Env) (*ByteStreamServerProxy, error) {
-	localCache := env.GetCache()
-	if localCache == nil {
-		return nil, status.FailedPreconditionError("A cache is required to enable the ByteStreamServerProxy")
+func New(env environment.Env) (*ByteStreamServerProxy, error) {
+	local := env.GetLocalByteStreamClient()
+	if local == nil {
+		return nil, fmt.Errorf("A local ByteStreamClient is required to enable ByteStreamServerProxy")
 	}
-	remoteCache := env.GetByteStreamClient()
-	if remoteCache == nil {
-		return nil, fmt.Errorf("A ByteStreamClient is required to enable ByteStreamServerProxy")
+	remote := env.GetByteStreamClient()
+	if remote == nil {
+		return nil, fmt.Errorf("A remote ByteStreamClient is required to enable ByteStreamServerProxy")
 	}
 	return &ByteStreamServerProxy{
-		env:         env,
-		localCache:  localCache,
-		remoteCache: remoteCache,
+		env:    env,
+		local:  local,
+		remote: remote,
 	}, nil
 }
 
 func (s *ByteStreamServerProxy) Read(req *bspb.ReadRequest, stream bspb.ByteStream_ReadServer) error {
-	remoteStream, err := s.remoteCache.Read(stream.Context(), req)
+	localReadStream, err := s.local.Read(stream.Context(), req)
 	if err != nil {
-		return err
+		return s.readRemote(req, stream)
 	}
+	responseSent := false
 	for {
-		rsp, err := remoteStream.Recv()
+		rsp, err := localReadStream.Recv()
 		if err == io.EOF {
 			break
 		}
 		if err != nil {
+			// If some responses were streamed to the client, just return the
+			// error. Otherwise, fall-back to remote.
+			if responseSent {
+				return err
+			} else {
+				return s.readRemote(req, stream)
+			}
+		}
+
+		stream.Send(rsp)
+		responseSent = true
+	}
+	return nil
+}
+
+// The Write() RPC requires the client keep track of some state. This struct
+// and its methods take care of that.
+type localWriter struct {
+	ctx          context.Context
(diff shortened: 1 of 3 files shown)


tylerwilliams (commented):
The stream code is so hard to follow (not your fault, grpc streams are tough). My understanding of this code is that it's starting async jobs during the write to populate the local cache. I wonder if a simpler approach where we just copy files from remote->local once a write succeeds would work?

Alternatively, we could do the copy synchronously, chunk by chunk, and if any error is encountered locally just set a flag and skip subsequent local writes?

wdyt?

iain-macdonald:
> The stream code is so hard to follow (not your fault, grpc streams are tough). My understanding of this code is that it's starting async jobs during the write to populate the local cache. I wonder if a simpler approach where we just copy files from remote->local once a write succeeds would work?
> 
> Alternatively, we could do the copy synchronously, chunk by chunk, and if any error is encountered locally just set a flag and skip subsequent local writes?
> 
> wdyt?



> The stream code is so hard to follow (not your fault, grpc streams are tough). My understanding of this code is that it's starting async jobs during the write to populate the local cache. I wonder if a simpler approach where we just copy files from remote->local once a write succeeds would work?
> 
> Alternatively, we could do the copy synchronously, chunk by chunk, and if any error is encountered locally just set a flag and skip subsequent local writes?
> 
> wdyt?

Agree that it's hard to follow, your understanding is correct. Of your two suggestions, I prefer the second (write local synchronously) because I don't want to increase load on the remote cache, so I'll make the changes here so that it behaves that way instead.

iain-macdonald:
I've been trying to refactor this per earlier comments, but haven't found a good way to do so while keeping byte_stream_server_proxy a wrapper of byte_stream_server. IMO that's worth the added complexity because there's a bunch of stuff in byte_stream_server relating to error handling and compressed/uncompressed input/output that I'd rather not duplicate, and it also makes future changes to byte_stream_server pass on to the proxy server.

That said, I did refactor this to make the dual-streaming stuff a bit easier to follow (at least in my opinion). Curious what you think of the new version and my points about the composition? Happy to chat on Slack / huddle tomorrow for higher bandwidth too.

iain-macdonald:
Alright, made the changes we discussed offline. This PR has ballooned a little bit, but most of it is moving code around / adding tests, so hopefully it's still a manageable review, but LMK if not.

iain-macdonald:
Gonna break this up into a handful of smaller PRs, stay tuned.

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -65,33 +148,67 @@ func (s *ByteStreamServerProxy) Read(req *bspb.ReadRequest, stream bspb.ByteStre
...
+		// Send to the local ByteStreamServer (if it hasn't errored)
+		localDone := req.GetFinishWrite()
+		if local != nil {
+			if err := local.Send(req); err != nil {
+				if err == io.EOF {
+					localDone = true
+				} else {
+					log.Infof("error writing to local bytestream server: %s", err)
+				}
+				local = nil
```
tylerwilliams, line 175: what's the reason for nilling out local here?

tylerwilliams (changes requested):
this looks a lot better than before, thanks for making those changes!

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,57 +6,140 @@ import (
...
+			// If some responses were streamed to the client, just return the
+			// error. Otherwise, fall-back to remote.
+			if responseSent {
+				return err
+			} else {
+				return s.readRemote(req, stream)
+			}
+		}
+
+		stream.Send(rsp)
```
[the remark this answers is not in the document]
iain-macdonald: whoops, yeah, fixed

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,57 +6,140 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	var localWriteStream *localWriter = nil
+	if req.ReadOffset == 0 {
+		localStream, err := s.local.Write(stream.Context())
+		if err != nil {
+			log.Warningf("error writing to local bytestream server: %s", err)
+			localWriteStream = nil
```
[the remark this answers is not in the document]
iain-macdonald: nah, fixed

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,57 +6,140 @@ import (
...
 		if err == io.EOF {
+			if localWriteStream != nil {
+				localWriteStream.close()
+			}
 			break
 		}
 		if err != nil {
 			return err
 		}
+
```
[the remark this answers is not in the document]
iain-macdonald: whoops, flipped this around so that stream is closed for any error

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
iain-macdonald: to prevent sending addition frames to the local server if it returned an error

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -65,33 +148,67 @@ func (s *ByteStreamServerProxy) Read(req *bspb.ReadRequest, stream bspb.ByteStre
...
 	for {
 		req, err := stream.Recv()
 		if err != nil {
 			return err
 		}
-		writeDone := req.GetFinishWrite()
-		if err := remoteStream.Send(req); err != nil {
+
+		// Send to the local ByteStreamServer (if it hasn't errored)
+		localDone := req.GetFinishWrite()
```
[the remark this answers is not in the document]
iain-macdonald: yeah, i think it's effectively intended to be an OR of req.GetFinishWrite() and local.Send() == io.EOF. Copied from here: https://github.com/buildbuddy-io/buildbuddy/blob/3eff2d5d76c811c0bdc2a659cdd76acc87db0fed/server/cache_proxy/cache_proxy.go#L262-L269

tylerwilliams (approved):
lgtm -- can you get vadim's approval on the streaming loop specifically before submitting?

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,92 +6,218 @@ import (
...
+	return err
+}
+
+func (s *ByteStreamServerProxy) readRemote(req *bspb.ReadRequest, stream bspb.ByteStream_ReadServer) error {
+	remoteReadStream, err := s.remote.Read(stream.Context(), req)
+	if err != nil {
+		return err
+	}
+
+	var localWriteStream *localWriter = nil
```
vadimberezniker, line 114: nit: don't need ` = nil`. pointers are always nil initialized.

vadimberezniker, line 114: instead of initializing this to nil, would the code be cleaner if you instead instantiated a discardingLocalWriter{} that discards all data?

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,92 +6,218 @@ import (
...
+	if req.ReadOffset == 0 {
+		localStream, err := s.local.Write(stream.Context())
+		if err != nil {
+			log.Warningf("error writing to local bytestream server: %s", err)
```
vadimberezniker, line 118: did you mean to initialize localWriteStream with a nil localStream? I think that would lead to a panic.

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,92 +6,218 @@ import (
...
+			initialized:  false,
+			offset:       int64(0),
+		}
+	}
+
+	for {
+		rsp, err := remoteReadStream.Recv()
+		if err != nil {
+			if localWriteStream != nil {
+				localWriteStream.close()
```
vadimberezniker, line 133: should this function be called commit instead of close()? even though this is an error case, close() will attempt to finish the write. the write would fail, but it doesn't seem like we should be trying to commit it in the first place on error.

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,92 +6,218 @@ import (
...
+		remote: remote,
 	}, nil
 }
 
 func (s *ByteStreamServerProxy) Read(req *bspb.ReadRequest, stream bspb.ByteStream_ReadServer) error {
-	remoteStream, err := s.remoteCache.Read(stream.Context(), req)
+	localReadStream, err := s.local.Read(stream.Context(), req)
 	if err != nil {
-		return err
+		return s.readRemote(req, stream)
```
vadimberezniker, line 50: I think you should be generous about logging on error-cases instead of silently ignoring them. Will make problems easier to debug in the future. Also please use Ctx* log functions where possible.

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -6,92 +6,218 @@ import (
...
+			}
+			if err == io.EOF {
+				break
+			}
 			return err
 		}
+
+		if localWriteStream != nil {
+			if err := localWriteStream.send(rsp.Data); err != nil {
+				log.Debugf("Error writing locally: %s", err)
```
vadimberezniker, line 143: I think should at least be an Info log so that we see it in prod as well? Same below.

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy_test.go
```diff
@@ -0,0 +1,335 @@
...
+		// uploads)
+		tc.bazelVersion = "5.0.0"
+		t.Run(tc.name+", bazel "+tc.bazelVersion, run)
+
+		tc.bazelVersion = "5.1.0"
+		t.Run(tc.name+", bazel "+tc.bazelVersion, run)
+	}
+}
+
+func newUUID(t *testing.T) string {
```
vadimberezniker, line 331: nit: might as well use uuid.New() instead

vadimberezniker (changes requested):

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
iain-macdonald: sg, done

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
iain-macdonald: whoops, thanks!

iain-macdonald: sure, done

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
iain-macdonald: oh nice catch, fixed

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
iain-macdonald: renamed and restructured to not call on errors except EOF

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
iain-macdonald: sg, done

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy_test.go
iain-macdonald: sure, fixed

vadimberezniker (approved):

PR #6949, opened 2024-06-28
fetch_server_test: demonstrate cache misses

sluongng:
As a follow up from #6911, provide a test to demonstrate cache misses on
subsequent FetchBlob requests where the digestFunc and the checksum
qualifier differ.

When this is fixed in the future, we should be able to flip the flag
accordingly.


bduffany (approved):

sluongng:
Hmm the dev workflow succeeded but the check status was not sent to Github.

Merging.

PR #6951, opened 2024-06-28
[RB] Add support to run multiple commands to Remote Run API

maggie-lou:
**Related issues**: N/A


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -413,21 +384,11 @@ func (r *buildEventReporter) Start(startTime time.Time) error {
 		startedEvent.Children = append(startedEvent.Children, &bespb.BuildEventId{Id: &bespb.BuildEventId_WorkflowConfigured{WorkflowConfigured: &bespb.BuildEventId_WorkflowConfiguredId{}}})
 		r.Printf("Streaming workflow logs to: %s", invocationURL(*invocationID))
 	} else {
-		startedEvent.Children = append(startedEvent.Children, &bespb.BuildEventId{Id: &bespb.BuildEventId_Pattern{Pattern: &bespb.BuildEventId_PatternExpandedId{Pattern: patterns}}})
 		r.Printf("Streaming remote runner logs to: %s", invocationURL(*invocationID))
 	}
 	if err := r.bep.Publish(startedEvent); err != nil {
 		return err
 	}
-	if !r.isWorkflow {
```
maggie-lou, line 422: If we're supporting multiple commands for remote bazel, it doesn't make sense for the outer invocation to emit this event. Unfortunately the UI will contain less information about the build (like the target that was run), even if only 1 bazel command is run

proto/runner.proto
```diff
@@ -17,10 +17,9 @@ message RunRequest {
   // A minimum of commit_sha or branch must be specified.
   git.RepoState repo_state = 3;
 
-  // The bazel command to run. Required.
-  // Example: "build :server"
-  // Example: "test //..."
-  string bazel_command = 4;
+  // The commands to run, in order. Required.
+  // Example: ["bazel build :server", "echo \"Hello World\""]
+  repeated string steps = 16;
```
bduffany, line 22: Having a repeated list of strings feels pretty similar to having a single string, since each script can have multiple commands (joined by `;` / `&&` etc.)

To make this a little easier in the common case, we could make it a single string called `run` (i.e. one step, and called "run" to match the workflow config YAML)

Alternatively, if we want to go for maximum extensibility, and allow adding more features like `name`, `if`, `env`, `uses`, etc., we could make it a repeated struct like `"steps": [ { "name": "...", "run": "..."} ]` (i.e. mirror the workflow config YAML)

wdyt?

enterprise/server/cmd/ci_runner/main.go
bduffany: wdyt about having a special case if there's only 1 command? (since that will probably be the most common case?) if it's hard because of the lossy bazel command => bash script conversion, maybe we can send up some extra command/pattern metadata to avoid losing this information?

proto/runner.proto
maggie-lou: I'm torn but I'm leaning towards the more extensible approach. The biggest reason is that I think we would want this as our CI product matures down the line, and we wouldn't have to require a migration

It also gives us more flexibility in making a better product experience. One feature I expect would be popular is having a `timeout` on certain steps, for example

That being said, we're trying not to prioritize a comprehensive CI suite right now, and that decision would open us to more customers asking us to add special features as opposed to if we were just like "do it all in bash". Plus each command is run in a separate bash process, which might lead to surprising behavior vs if we just ran it all in one block

CC @siggisim for additional thoughts

enterprise/server/cmd/ci_runner/main.go
maggie-lou: I thought about that, but the special case is still hard given the openness of bash scripts - 1 bash command could contain 3 bazel and 4 bash commands separated by && and ; 

But once we start processing the build logs for bazel invocations, we could start sending this again, and only using it to populate an invocation if there's only 1 bazel invocatoin

bduffany: I mean for stuff like `bb remote build //...` we only allow passing 1 command at the command line right? wonder if we'll lose the `build //...` information in this case, or if we could send up extra metadata to preserve that info

maggie-lou: I'd prefer not to add special logic around that, when we will already have a way to detect single bazel builds. But while we're in the intermediate phase of not processing the build logs yet, I can just re-add this code back in if you'd prefer

proto/runner.proto
maggie-lou: Siggi and I talked offline. We agreed that the struct approach is okay because while it's more complex, this is intended as an internal API. We can make an external API with a simpler interface, and extend the remote bazel cli so that fewer customers need to use the internal API

An idea I had is that we can also make a cli command that outputs a curl command with a sample Remote Run API call. So if you need to tweak one thing, you don't have to manually create a complicated request

enterprise/server/cmd/ci_runner/main.go
bduffany: What's the single way of detecting single builds that you're referring to?

bduffany (approved):

enterprise/server/cmd/ci_runner/main.go
maggie-lou: We'll have logic that's parsing the runner logs and looking for invocation links. We can use that to determine if there's just one bazel build

bduffany: We might have some difficulty with that approach because we want to publish the workflow invocation metadata as early as possible, and the metadata includes the command - but with the log scanning approach, we don't know whether there's more than one invocation until the second invocation has started

maggie-lou: I see. What if we added a name field to `Step`? I know we've wanted that anyway, so people could specify blocks like `Setup` and `Teardown`. Remote bazel could set the name to the bazel command

bduffany: That seems like a reasonable option if it winds up looking nice in the UI

PR #6954, opened 2024-06-28
New script to manually push multi-platform docker images

tempoz:
Since `rules_docker` is no longer supported, we are looking at moving to `rules_oci`. However, `rules_oci` doesn't support building images with a dockerfile since it's not hermetic. Instead, they recommend building your base image manually and uploading it to a registry, and then pulling it down by hash during your bazel build. This is a script to assist in the process of building such images. It's likely that instead of invoking it directly, we can make one-line bash scripts that do the pushes we want for us. I think another option for further work would be adding a warning about pushing an image to a prod repository for which there does not exist a corresponding dev repository.

**Related issues**: N/A


tylerwilliams:
I ran this successfully -- reported a few minor bugs to @tempoz but can confirm it does the thing.

bduffany (approved):

PR #6956, opened 2024-06-28
Consolidate invocation metadata logic

bduffany:
In https://github.com/buildbuddy-io/buildbuddy-internal/issues/3579 we were reporting incorrect statuses to GitHub after a seemingly harmless change to change the order of build events.

The root issue was that the build status reporter was only reporting a status once the WorkspaceStatus event was loaded, instead of waiting for _all_ declared metadata events.

The "quick fix" was to simply change the order of events so that we publish WorkspaceStatus last. This PR implements a more comprehensive fix, which is to improve how we check whether metadata is fully loaded. Specifically, this PR refactors the build status reporter to use the same logic that we use in the build event handler to decide whether the invocation metadata is ready to be flushed to the DB. This refactor involved moving the "unprocessedStartingEvents" logic into the accumulator so that the build status reporter can access it.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3579

### DIFF
--- a/server/build_event_protocol/accumulator/accumulator.go
+++ b/server/build_event_protocol/accumulator/accumulator.go
@@ -42,6 +42,12 @@ type Accumulator interface {
 	// not present.
 	Invocation() *inpb.Invocation
 
+	// MetadataIsLoaded returns whether all expected invocation-level metadata
+	// events have been received. It's generally not safe to call the other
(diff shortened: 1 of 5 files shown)


maggie-lou (approved):
Thanks for fixing this

PR #6962, opened 2024-06-28
Code search more button tweaks

siggisim:
- Removes use of OutlineButton
- Adds icon
- Renders "matches" or "match" depending on how many additional matches there are


tylerwilliams (approved):

tylerwilliams:
thanks :)

PR #6965, opened 2024-06-30
Fix 'bazel' status reported for each workflow

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -860,6 +860,28 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 	// because that error is instead surfaced in the caller func when calling
 	// `buildEventPublisher.Wait()`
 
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -860,6 +860,28 @@ func (ar *actionRunner) Run(ctx context.Context, ws *workspace) error {
 	// because that error is instead surfaced in the caller func when calling
 	// `buildEventPublisher.Wait()`
 
+	wfc := &bespb.WorkflowConfigured{
```
maggie-lou, line 863: nit: Could you add a comment that this has to be emitted first

maggie-lou (approved):

enterprise/server/cmd/ci_runner/main.go
bduffany: I have an in-progress PR that updates the BES server to not care about the order of these events, so will leave as-is

PR #6966, opened 2024-06-30
fetch_server: follow http_proxy env

hunshcn:


sluongng (commented):
This LGTM. Could you please add a small test to verify that requests are going through the proxy set in environment variables?

hunshcn:
> This LGTM. Could you please add a small test to verify that requests are going through the proxy set in environment variables?

Does this really need to be tested? They are all part of the standard library.

sluongng:
> Does this really need to be tested? They are all part of the standard library.

Agree that we don't need to validate the functionality of the standard library.

However, having a test checking that `HTTPS_PROXY` is respected here would help us avoid any accidental breakages when doing future refactoring. So think of the test as a way to protect this feature from future changes.

Again, it only needs to be a very small, basic test case. No need to check for all variables or all scenarios.

hunshcn:
I'm sorry that I didn't understand what I should add. Can you help me to improve it? @sluongng

sluongng:
Ah that's ok! No worry. I will take your commit as-is and add the test in a separate commit.

sluongng:
We merged this change via https://github.com/buildbuddy-io/buildbuddy/pull/6983. Thanks so much for your contribution 🙏

PR #6967, opened 2024-07-01
Stop() gRPC Servers started in RegisterLocalGRPCServer on t.Cleanup()

iain-macdonald:
**Related issues**: N/A


server/testutil/testenv/testenv.go
```diff
@@ -86,14 +86,15 @@ func init() {
 //
 // Register services to the server, then call LocalGRPCConn to get a connection
 // to the returned server.
-func RegisterLocalGRPCServer(te *real_environment.RealEnv) (*grpc.Server, func()) {
+func RegisterLocalGRPCServer(te *real_environment.RealEnv, t *testing.T) (*grpc.Server, func()) {
```
tylerwilliams, line 89: I think we have dueling conventions here: usually `testing.TB` is the first argument,  but then our own convention is to make `env` the first argument.

Probably the go convention should take precedence.

Context is a whole other thing: https://github.com/golang/go/issues/30246
env first: https://app.buildbuddy.dev/search/?q=.*RealEnv.*testing.T
t first: https://app.buildbuddy.dev/search/?q=.*testing.T.*RealEnv

tylerwilliams (changes requested):

server/testutil/testenv/testenv.go
iain-macdonald: sg, fixed all the occurrences this pr touches so t is before env

tylerwilliams (approved):

PR #6968, opened 2024-07-01
Fully shutdown healthchecker on t.Cleanup()

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/server/testutil/testenv/testenv.go
+++ b/server/testutil/testenv/testenv.go
@@ -160,6 +160,7 @@ func GetTestEnv(t testing.TB) *real_environment.RealEnv {
 	}
 
 	healthChecker := healthcheck.NewHealthChecker("test")
+	t.Cleanup(healthChecker.Shutdown)
 	te := real_environment.NewRealEnv(healthChecker)
 	c, err := memory_cache.NewMemoryCache(1000 * 1000 * 1000 /* 1GB */)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


server/testutil/testenv/testenv.go
```diff
@@ -137,6 +137,7 @@ func GetTestEnv(t testing.TB) *real_environment.RealEnv {
 	}
 
 	healthChecker := healthcheck.NewHealthChecker("test")
+	t.Cleanup(healthChecker.Shutdown)
```
bduffany, line 140: I think as written, the shutdown functions will run asynchronously, while new tests are allowed to start executing, which doesn't seem ideal 🤔 

I'd say we might want to call `WaitForGracefulShutdown` here but it looks like that has the side effect of invoking the health checker loop, which might not be desired in tests? and it might increase test duration a bunch depending on how long the graceful shutdown takes

tylerwilliams: Does it make sense to have a testonly healthchecker that does something different here?

bduffany: hmm actually it looks like NewHealthChecker() is already starting the health checker loop. So the only concern would be increasing test duration. Maybe worth a try though and seeing whether tests get significantly slower or not.

I think that another thing we shouldn't do in tests is register the `SIGINT` / `SIGTERM` handlers - those are preventing Ctrl+C from working when running the test locally.

PR #6970, opened 2024-07-01
recreate iterator every X minutes in sampler

luluz66:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -36,6 +36,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/proto"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/statusz"
+	"github.com/buildbuddy-io/buildbuddy/server/util/timeutil"
 	"github.com/buildbuddy-io/buildbuddy/server/util/tracing"
 	"github.com/docker/go-units"
 	"github.com/elastic/gosigar"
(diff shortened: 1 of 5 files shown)


enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2577,17 +2580,24 @@ func (e *partitionEvictor) generateSamplesForEviction(quitChan chan struct{}) er
 		shouldSleep := e.sizeBytes <= int64(SamplerSleepThreshold*float64(e.part.MaxSizeBytes))
 		e.mu.Unlock()
 		if shouldSleep {
+			samplerDelay := time.NewTimer(SamplerSleepDuration)
 			select {
 			case <-quitChan:
 				return nil
 			case <-samplerDelay.C:
 			}
+			timeutil.StopAndDrainTimer(samplerDelay)
```
vadimberezniker, line 2589: We should add some documentation to this function. Draining the timer is only necessary if you plan to reuse it or if there's a chance the timer would frequently not run to completion. Neither is true here. 
Here a simple time.After() would suffice. Same below.

enterprise/server/backends/pebble_cache/pebble_cache.go
```diff
@@ -2601,10 +2611,6 @@ func (e *partitionEvictor) generateSamplesForEviction(quitChan chan struct{}) er
 			iter = newIter
 		}
 		totalCount += 1
-		if totalCount > e.samplesPerBatch {
```
vadimberezniker, line 2604: I was thinking that the timer-based recreation would be a defensive check and that we would still keep the count-based check as well. Did you have any opinion on whether we should keep both or only keep the time-based one?

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: I changed to use "After" here. But I needed the timer below to avoid memory leak when case e.samples <- sample is run.

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: I kept both.

server/util/timeutil/timeutil.go
```diff
@@ -72,3 +73,13 @@ func StopAndDrainTimer(t *time.Timer) {
 		}
 	}
 }
+
+// TODO: combine StopAndDrainClockworkTimer and StopAndDrainTimer.
```
vadimberezniker, line 77: Does changing the argument type in StopAndDrainTimer to accept clockwork.Timer interface not work?

enterprise/server/backends/pebble_cache/pebble_cache.go
vadimberezniker: The count-based check is still missing.

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache.go
luluz66: Thanks!..I didn't catch that when I try to recover from a bad git restore command, and I was wondering why the test became flaky...

PR #6980, opened 2024-07-02
Remove runner.state from runner.String()

iain-macdonald:
https://github.com/buildbuddy-io/buildbuddy/pull/6968 caused failures in `//enterprise/server/test/integration/workflow:workflow_test` due to the fact that the `String()` function of runner does an unprotected read of `runner.state` which may race with one of the function that writes that field.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -232,9 +232,11 @@ func (r *taskRunner) String() string {
 	if err != nil {
 		ph = "<ERR!>"
 	}
+	// Note: we don't log r.state here as this can make log statements calling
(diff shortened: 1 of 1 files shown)


bduffany (commented):
I feel like it's a little too easy to forget the lock when writing `log.CtxWarningf("%s", r)` - wdyt about removing the `String()` interface to prevent this class of bug and adding a more explicit method like `DebugString()` ?

```
p.mu.RLock()
rs := r.DebugString()
p.mu.RUnlock()
```

iain-macdonald:
Updated per offline discussion to omit runner.State from runner.String() instead of the more complicated locking approach. PTAL!

bduffany (approved):

PR #6985, opened 2024-07-03
sluongng/gazelle new update repos

sluongng:
- **go deps.bzl: make sure to add -prune when update-repos**
- **patch gazelle to use go download for update-repos**

### DIFF
--- a/buildpatches/gazelle.patch
+++ b/buildpatches/gazelle.patch
@@ -39,3 +39,79 @@ index f5ae5e5..74678b4 100644
      deps = ["@com_github_bazelbuild_buildtools//build"],
  )
  
+diff --git a/language/go/modules.go b/language/go/modules.go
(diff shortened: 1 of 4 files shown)


buildpatches/gazelle.patch
```diff
@@ -39,3 +39,79 @@ index f5ae5e5..74678b4 100644
      deps = ["@com_github_bazelbuild_buildtools//build"],
  )
  
+diff --git a/language/go/modules.go b/language/go/modules.go
```
sluongng, line 42: @fmeum this patch could be interesting to you.

Instead of using `go list` output as-is, I filtered it using `go mod download` output, which seems to accurately track the dependencies after [graph pruning](https://go.dev/ref/mod#graph-pruning). Essentially matching the content of `go.mod` file.

This massively prune down our deps.bzl file.

PR #6991, opened 2024-07-08
podman: make podman-static archive deterministic

sluongng:
The action is poisoned by the creation date of the files on disk.
Fix this by adding `--mtime` to the `tar` packaging command.
Unfortunately this flag is only available on linux version of `tar`, so
restrict the target to only be executed on Linux.

Also ensure that the subsequent image layers are only built lazily to
avoid any extra downloads on our workflows runner.


bduffany (approved):
Thanks!

PR #6924, opened 2024-06-26
Add sha384/sha512 digest function

hunshcn:
As the pre-content supporting sha384/sha512 remote_asset.

Related https://github.com/buildbuddy-io/buildbuddy/pull/6911

PR #6997, opened 2024-07-09
fetch_server: support storing blob using checksum digest func

sluongng:
Originally suggested by @hunshcn in #6911.

In each FetchBlobRequest, there are 2 digest functions:

a. One digest function to verify the content of the downloaded blob.
This is usually set by the Bazel repository rules calling
`repository_ctx.download` and `repository_ctx.download_and_extract` API.
Usually, this is set by http_archive rules to SHA256, but in recent
community rules, such as rules_js and rules_proto, SHA512 and SHA384 are
also used.

b. One digest function to store the blob in cache for the client to
subsequently downloaded. This is usually set to either SHA1, SHA256, or
Blake3, which are the digest functions Bazel support in `startup
--digest_function`. The default is SHA256.

If the Bazel client is also configured with
`--experimental_remote_downloader` pointing to BuildBuddy, client would
get no cache hits from BuildBuddy side if (a) and (b) differs. This is
a result of FetchServer storing the downloaded blob using (b) while
looking up the cache using (a) before downloading.

In #6924, we added support for using these new digest functions in our
cache storage layer.

In #6911, we added support for verifying downloaded blobs using these
new digest functions to Fetch Server.

This PR added the ability to store downloaded blob in our cache using
(a), in additional to existing entry stored with (b). It allows
subsequent Fetch Requests to get cache hits from BuildBuddy at the cost
of storing the blob twice.

### DIFF
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -306,8 +306,15 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 	}()
 
 	// If the requested digestFunc is supplied and differ from the checksum sri,
-	// verify the downloaded file with the checksum sri before storing it to
-	// our cache.
+	// verify the downloaded file with the checksum sri before storing it to our cache.
+	//
+	// This will store the downloaded blob in our cache twice:
+	//  - One entry using the checksum digest function for future cache hits.
+	//  - One entry using the storage digest function for client to download.
+	//
+	// TODO(sluongng): We can track download information in a KV store with value
+	// pointing to the CAS entry. That way, we would only need to store the download
+	// blob once.
 	if checksumFunc != storageFunc {
 		checksumDigestRN, err := cachetools.ComputeFileDigest(tmpFilePath, remoteInstanceName, checksumFunc)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/buildbuddy-io/buildbuddy
 
-go 1.21.7
+go 1.22.4
```
sluongng, line 3: This is needed to use the new `for range <int>` syntax.

We are already using Go 1.22, this is only raising the repo's minimum supported SDK version so that the Go language server and other static analysis won't complain.

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -316,6 +323,12 @@ func mirrorToCache(ctx context.Context, bsClient bspb.ByteStreamClient, remoteIn
 		if expectedChecksum != "" && checksumDigestRN.GetDigest().GetHash() != expectedChecksum {
 			return nil, status.InvalidArgumentErrorf("response body checksum for %q was %q but wanted %q", uri, checksumDigestRN.GetDigest().Hash, expectedChecksum)
 		}
+		if _, err := cachetools.UploadFile(ctx, bsClient, remoteInstanceName, checksumFunc, tmpFilePath); err != nil {
+			// Best effort storing downloaded blob to our cache.
+			// This is Ok to fail because subsequent requests will simply get no cache hits
+			// and download blob again from upstream URL.
+			log.Warningf("failed to cache object with checksumFunc: %s", err)
```
bduffany, line 330:
```suggestion
			log.CtxWarningf(ctx, "Failed to cache object with checksumFunc: %s", err)
```

bduffany (approved):

PR #7003, opened 2024-07-10
Ignore error during network cleanup if the route doesn't exist

maggie-lou:
We are getting  vm_network_cleanup_failed alerts because of this

**Related issues**: Example https://buildbuddy-corp.slack.com/archives/C0154UFBCJ2/p1720610696894509

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -435,7 +435,9 @@ func SetupVethPair(ctx context.Context, netNamespace, vmIP string, vmIdx int) (_
 	exists, err := routeExists(ctx, cloneIP, cloneEndpointAddr)
 	if err != nil {
 		return nil, err
-	} else if !exists {
+	} else if exists {
+		return nil, status.UnavailableErrorf("ip route %s via %s already exists", cloneIP, cloneEndpointAddr)
+	} else {
 		err = runCommand(ctx, "ip", "route", "add", cloneIP, "via", cloneEndpointAddr)
 		if err != nil {
 			return nil, err
@@ -447,8 +449,6 @@ func SetupVethPair(ctx context.Context, netNamespace, vmIP string, vmIdx int) (_
 			}
 			return nil
 		})
-	} else {
-		log.Debugf("ip route %s via %s already exists", cloneIP, cloneEndpointAddr)
 	}
 
 	if IsSecondaryNetworkEnabled() {
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -441,7 +441,11 @@ func SetupVethPair(ctx context.Context, netNamespace, vmIP string, vmIdx int) (_
 			return nil, err
 		}
 		cleanupStack = append(cleanupStack, func(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):
LGTM

Any idea why we'd try to clean up a route that already was removed? Is this a symptom of some race condition?

maggie-lou:
> Any idea why we'd try to clean up a route that already was removed? Is this a symptom of some race condition?

I'm not positive to be honest.
The function that calls the cleanup (`firecracker::remove`) is in a sync.Once, so should never get called again.
We aren't clearing the cleanup function after it's been called, but for firecracker we create a new Container object even when we're loading a snapshot, so that shouldn't matter.
Brandon added some network locking, so another VM shouldn't be trying to create / delete the same route. @bduffany do you have any ideas how this could happen?

server/util/networking/networking.go
```diff
@@ -441,7 +441,11 @@ func SetupVethPair(ctx context.Context, netNamespace, vmIP string, vmIdx int) (_
...
 		cleanupStack = append(cleanupStack, func(ctx context.Context) error {
-			return runCommand(ctx, "ip", "route", "delete", cloneIP)
+			err := runCommand(ctx, "ip", "route", "delete", cloneIP)
+			if err != nil && !strings.Contains(err.Error(), "No such process") {
+				return err
+			}
+			return nil
 		})
 	} else {
 		log.Debugf("ip route %s via %s already exists", cloneIP, cloneEndpointAddr)
```
bduffany, line 451: The "routeExists" logic seems a little sketchy since we don't expect these routes to be used by anything other than VMs. Maybe we should promote this log line to Warning level or return an Unavailable error here?

maggie-lou: I checked and we haven't seen this log for over 2 weeks in dev - upgraded to an error

maggie-lou:
Okay had a chance to investigate this a little more

In dev, every time we got this `ip route delete` error, there were a couple errors like `ip route add default via 192.168.0.X" Error: Nexthop has invalid gateway.` a couple milliseconds apart. So it seems like the networking has a temporary glitch, and a lot of the networking commands fail for a couple milliseconds.

The errors setting up the network should be okay, because the executors should retry the task

The aspect I'm not sure about is whether the route does actually still exist when we fail to delete it, and it's incorrectly reported as not existing because of the network glitch. If we ignore the error, then we'll unlock that ip and another VM can claim it. This PR makes it so that step will fail if the route already exists at that point. This alert is firing relatively frequently, so I'm disposed to trying it out and seeing how it does @tylerwilliams @bduffany

bduffany:
@maggie-lou is it possible that `Nexthop has invalid gateway` is hinting at a legitimate error?

vadimberezniker:
`Nexthop has invalid gateway` is supposed to indicate that the gateway IP belongs to a subnet that is not attached to any interface. This would suggest something has already gone wrong prior to route setup.

bduffany:
I think I got to the bottom of this and fixed the issues in https://github.com/buildbuddy-io/buildbuddy/pull/7028

PR #7006, opened 2024-07-10
Add PSI metrics to UsageStats proto

bduffany:
The immediate use case for these metrics is to expose them in the UI to make it easier for users to see whether a task is slow due to thrashing (for each resource type: CPU, memory, and disk).

It may also be useful to incorporate this information into task sizing somehow. For example, we can consider a task size measurement to be unreliable if there was a lot of thrashing while the measurement was taken, and avoid storing it in Redis. Another idea is that we could apply a small percent penalty to counteract the negative feedback loop that can happen when a task runs during high pressure conditions.

**Related issues**: N/A

### DIFF
--- a/proto/remote_execution.proto
+++ b/proto/remote_execution.proto
@@ -2230,6 +2230,55 @@ message UsageStats {
 
   // Peak file system usage, for each mounted file system.
   repeated FileSystemUsage peak_file_system_usage = 4;
+
+  // CPU PSI metrics.
+  PSI cpu_pressure = 5;
+
+  // Memory PSI metrics.
+  PSI memory_pressure = 6;
+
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):
This LGTM assuming you're okay with the space usage per float (4 bytes?)

bduffany:
> This LGTM assuming you're okay with the space usage per float (4 bytes?)

If these values turn out to be sparse (i.e. usually 0) then they won't take up much space - I'd like to try populating these and if in practice these turn out to be taking up a lot of space, we can stop populating the `avg` numbers and just use the `total` value. I think the avg numbers would be interesting to see, so that we know whether stalling was occurring throughout the whole action or just for the tail end of the action

PR #7010, opened 2024-07-11
Add PSI stats implementation

bduffany:
* Read PSI metrics from cgroup2 fs
  * Refactor path templates a bit - instead of having a separate path template for each cgroup file we want to read, just keep track of the cgroup dir. All of the cgroup accounting files should be direct children of this dir.
  * Also remove deprecated path template flags - we aren't setting these anywhere and they are no longer needed.
* Update `UsageStats` util to allow tracking relative increases in pressure stall time (in the runner recycling case), similar to what we do for CPU time
* Add tests for `UsageStats` util, which also caught some bugs in the current implementation :grimacing: 

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/cgroup/BUILD
+++ b/enterprise/server/remote_execution/cgroup/BUILD
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "cgroup",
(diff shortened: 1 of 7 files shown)


bduffany:
turns out the path template changes don't work for cgroup v1 because the cgroup name is actually part of the directory name, e.g. `/sys/fs/cgroup/cpu,cpuacct/libpod_parent/libpod-{{.ContainerID}}` - need to rework this a bit

bduffany:
Tested in dev (which has cgroup v2 only) and in a GCP VM with a hybrid cgroup setup (v1+v2)

iain-macdonald (approved):
Thanks for adding that test :-)

PR #7011, opened 2024-07-11
Fix vulnerabilities in the executor image

tempoz:
~~We now use podman instead of docker for our executors, so we should be safe to remove docker, which was filled with gobinaries with fixable CVEs. Further, we upgrade the installed packages on the image to pick up security fixes. This results in a clean bill of health from `trivy`. Worked when deployed in dev.~~

Fully re-vamped. New stuff:
- Fixed all of the new deps that the `soci-snapshotter` upgrade dragged in to build with bazel
- Special callout to `containerd` for altering their generated `pb.go` files after generating them ([`go-fix-acronym`](https://github.com/containerd/protobuild/tree/main/cmd/go-fix-acronym) 😭)
- ~~Pinned `docker` and `containerd` debian packages to unaffected versions~~ Handled in a stacked PR
- ~~Upgraded `podman-static` and `soci-snapshotter` to new releases we've applied our patches to.~~ Handled in a stacked PR
- ~~Pasted over `containerd` executables with the version we select in `deps.bzl`~~ Handled in a stacked PR
- ~~Pasted over `runc` executables with the version we select in `deps.bzl`~~ Handled in a stacked PR
- ~~Moved `docker-credential-gcr` into a `container-layer`~~ Handled in a stacked PR
- ~~We finally ran into a naming collision with `health.proto`, switched to using the official version (identical code)~~ Handled in a separate PR
- ~~Pasted over `rootlesskit` executables with the version we select in `deps.bzl`~~ Handled in a stacked PR

To view the remaining vulnerabilities in the executor, install [`trivy`](https://github.com/aquasecurity/trivy) and run:
```sh
bazel run //enterprise/server/cmd/executor:executor_image --curses=no | \
grep 'Tagging [0-9a-f]* as bazel/enterprise/server/cmd/executor:executor_image' | \
awk '{print $2}' | \
read -l DOCKER_TAG && \
docker image tag $DOCKER_TAG executor:bazel-build && \
trivy image --ignore-unfixed executor:bazel-build
```

Note that we cannot ever fix the vulnerabilities in `docker-credential-gcr`, as it is unmaintained, and the vulnerabilities in `podman-static` are mostly fixed (except for the ones stemming from runc, which they pinned an old version of, discussion [here](https://github.com/containers/podman/issues/19795)) in `podman` v5.2.0, but the `podman-static` repo understandably does not yet have that version available, as it was released today.

**Related issues**: N/A


dockerfiles/executor_image/Dockerfile
```diff
@@ -6,48 +6,21 @@ FROM debian@sha256:fac2c0fd33e88dfd3bc88a872cfb78dcb167e74af6162d31724df69e482f8
...
     apt-get install -y \
-    coreutils curl apt-transport-https ca-certificates gnupg-agent software-properties-common fuse && \
+		# Install skopeo and umoci which we use to unpack OCI images when we're not using docker.
+		# Install iproute2 ("ip" command) to configure networking on host.
+    fuse skopeo umoci iproute2 amazon-ecr-credential-helper && \
+    apt-get upgrade -y && \
     rm -rf /var/lib/apt/lists/* && apt-get clean
 
-RUN ARCH=$(dpkg --print-architecture) && \
+RUN apt-get update && \
```
bduffany, line 15: nit: I think we could remove this whole RUN step by provisioning `tini` the same way we provision firecracker/jailer etc. (e.g. adding a `container_layer` and downloading the `tini` binary via an `http_file` provisioned by bazel)

dockerfiles/executor_image/Dockerfile
```diff
@@ -6,48 +6,21 @@ FROM debian@sha256:fac2c0fd33e88dfd3bc88a872cfb78dcb167e74af6162d31724df69e482f8
...
+		ARCH=$(dpkg --print-architecture) && \
     SHASUM=$(case $ARCH in \
         amd64) echo '93dcc18adc78c65a028a84799ecf8ad40c936fdfc5f2a57b1acda5a8117fa82c' ;; \
         arm64) echo '07952557df20bfd2a95f9bef198b445e006171969499a1d361bd9e6f8e5e0e81' ;; \
     esac) && \
     curl -o /tini -fsSL "https://github.com/krallin/tini/releases/download/v0.19.0/tini-${ARCH}" && \
     echo "$SHASUM /tini" | sha256sum -c && \
-    chmod +x /tini
-
-RUN install -m 0755 -d /etc/apt/keyrings && \
```
bduffany, line 21: since we're removing `docker` I think this means that on-prem users will have to install docker themselves if they want to manually run certain diagnostics like looking at container logs. I think this is fine, but we might want to document this somewhere, maybe in the on-prem executor setup docs (ideally have a script that people can paste into their terminal which installs docker).

bduffany (approved):

dockerfiles/executor_image/Dockerfile
siggisim: I think we still have quite a few on-prem customers using docker, so I don't feel super comfortable removing this. Can we just upgrade the versions here instead to pick up CVE fixes?

siggisim (changes requested):

dockerfiles/executor_image/Dockerfile
tempoz: We can't - or at least we couldn't when I tried a week and a half ago. Because it's not a package associated with the distribution, vulnerability scanning tools go by the most recent versions of the libraries it uses to determine whether or not a fix is available. We'd probably have to compile docker from scratch with updated dependencies if we wanted a clean scan, unless they released an updated version that addressed all the CVEs very recently :/

tempoz: That said, I'll look into it and make sure.

dockerfiles/executor_image/Dockerfile
tempoz: Cool, I'll see if I can get that working :D

dockerfiles/executor_image/Dockerfile
tempoz: To break it down:

- The latest release of `containerd` still has `CVE-2023-47108` from `otelgrpc` (fixed in version 0.46):
https://github.com/containerd/containerd/blob/v1.7.19/go.mod#L65
Their main branch has a new enough version of `otelgrpc`:
https://github.com/containerd/containerd/blob/ac0f34f0ec03c818797f085ee2fe39329689d77c/go.mod#L64
However, the change updating that dependency is three months old and the release is from a week ago, so it doesn't seem that they're eager to release it. We'd have to host one of their nightly builds somewhere we could pull it down from, since they only last 90 days:
https://github.com/containerd/containerd/actions?query=workflow%3ANightly

- The release of `rootlesskit` in the docker repository (v1.1.0) has `CVE-2023-45288` from `golang.org/x/net` (fixed in version 0.23.0):
https://github.com/rootless-containers/rootlesskit/blob/v1.1.0/go.mod#L25
This is fixed in the newest release, v2.1.0, so we can get that release and replace the existing binaries:
https://github.com/rootless-containers/rootlesskit/blob/v2.1.0/go.mod#L29

- The release `runc` in the docker repository (v1.1.12) has `CVE-2024-24786` from `google.golang.org/protobuf` (fixed in version 1.33.0):
https://github.com/opencontainers/runc/blob/v1.1.2/go.mod#L25
The only existing release in which this is fixed is their pre-release rc2, v1.2.0-rc.2, so we could once again download this release and overwrite the existing binaries:
https://github.com/opencontainers/runc/blob/v1.2.0-rc.2/go.mod#L25

- Finally, `docker-credential-gcr` has `CVE-2024-24557` and `GHSA-jq35-85cj-fj4p` from `docker` (fixed in versions 24.0.9, 25.0.2 and in versions 24.0.7, 23.0.8, 20.10.27, respectively:
https://github.com/GoogleCloudPlatform/docker-credential-gcr/blob/v2.1.22/go.mod#L17
As well as `CVE-2023-45288`, once again from `golang.org/x/net` (fixed in version 0.23.0):
https://github.com/GoogleCloudPlatform/docker-credential-gcr/blob/v2.1.22/go.mod#L21
These have not been fixed to this day, and the last commit to this repo was over six months ago, which probably has something to do with google sunsetting container registry. In this case, to fix it we would need to fork the repo and update the dependencies ourselves, and then host the built binaries somewhere the dockerfile can fetch them from.

I can do all of this, but I want to be sure this is what we want to do before I do it, as this is straying pretty far from the beaten path.

tempoz: I talked with Siggi, and the plan is to upgrade everything above that has an official release to pick up security fixes, and mark things with no release which includes fixes as "fix not yet available" in vanta.

tempoz:
These have all been handled in separate PRs

tempoz:
Almost all of these changes were handles separately, all that's left is to upgrade `soci-snapshotter` as a go dep. Handling that here:
https://github.com/buildbuddy-io/buildbuddy/pull/7465

PR #7015, opened 2024-07-12
Make browser tests run as part of the normal Test workflow action

bduffany:
There isn't much reason to have a separate `Browser tests` workflow anymore since it now runs with the same Bazel flags as the normal `Test` workflow, and also it is no longer an optional check.

**Related issues**: N/A

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -10,7 +10,7 @@ actions:
         branches:
           - "*"
     bazel_commands:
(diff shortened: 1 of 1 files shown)


jdhollen:
Wanna add them to `Test with bzlmod` too?

bduffany:
> Wanna add them to Test with bzlmod too?

good call, done

maggie-lou (approved):

jdhollen (approved):

iain-macdonald (approved):
🄻🄶🅃🄼

PR #7021, opened 2024-07-15
Report podman overhead in execution stats

bduffany:
Make sure that the reported CPU time includes the CPU time spent in the `podman run` command itself, and not just the child processes in the spawned container. On my local machine this adds about 60ms of additional reported CPU time for each command. This should significantly improve the task sizing accuracy for commands where podman itself accounts for most of the runtime.

**Related issues**: N/A


iain-macdonald (commented):
nice :-)

iain-macdonald (approved):
oof, forgot to hit approve

PR #7023, opened 2024-07-15
Log snapshot ID when resuming

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -639,7 +639,7 @@ func NewContainer(ctx context.Context, env environment.Env, task *repb.Execution
 				log.CtxInfof(ctx, "Failed to get VM snapshot for keyset %s: %s", snaploader.KeysetDebugString(ctx, c.env, c.SnapshotKeySet(), c.supportsRemoteSnapshots), err)
 			} else {
 				label = metrics.HitStatusLabel
-				log.CtxInfof(ctx, "Found snapshot for key %s", snaploader.KeyDebugString(ctx, c.env, snap.GetKey(), c.supportsRemoteSnapshots))
+				log.CtxInfof(ctx, "Found snapshot with ID %q for key %s", snap.GetVMMetadata().GetSnapshotId(), snaploader.KeyDebugString(ctx, c.env, snap.GetKey(), c.supportsRemoteSnapshots))
 			}
 			metrics.RecycleRunnerRequests.With(prometheus.Labels{
 				metrics.RecycleRunnerRequestStatusLabel: label,


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -639,7 +639,7 @@ func NewContainer(ctx context.Context, env environment.Env, task *repb.Execution
 				log.CtxInfof(ctx, "Failed to get VM snapshot for keyset %s: %s", snaploader.KeysetDebugString(ctx, c.env, c.SnapshotKeySet(), c.supportsRemoteSnapshots), err)
 			} else {
 				label = metrics.HitStatusLabel
-				log.CtxInfof(ctx, "Found snapshot for key %s", snaploader.KeyDebugString(ctx, c.env, snap.GetKey(), c.supportsRemoteSnapshots))
+				log.CtxInfof(ctx, "Found snapshot with ID %q for key %s", snap.GetVMMetadata().GetSnapshotId(), snaploader.KeyDebugString(ctx, c.env, snap.GetKey(), c.supportsRemoteSnapshots))
```
maggie-lou, line 642: Could you add this to snaploader.KeyDebugString instead so it applies everywhere?

bduffany: Done and renamed to `SnapshotDebugString` since the ID isn't part of the key

maggie-lou: It is a part of the key? Not sure why the field isn't ordered so maybe you missed it? 

https://github.com/buildbuddy-io/buildbuddy/blob/e8f1218846a4e093080c36aa1cbb8acbe58699ce/proto/firecracker.proto#L56

bduffany: It's only part of the key when calling `enterprise/tools/vmstart` - we usually leave that field blank and match based on branch info + platform etc.

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -673,7 +682,7 @@ func (l *FileCacheLoader) cacheActionResult(ctx context.Context, key *fcpb.Snaps
 		if err := cachetools.UploadActionResult(ctx, l.env.GetActionCacheClient(), acDigest, ar); err != nil {
 			return err
 		}
-		log.CtxInfof(ctx, "Cached master remote snapshot manifest %s", KeyDebugString(ctx, l.env, key, true /*remote*/))
+		log.CtxInfof(ctx, "Cached master remote snapshot %s", snapshotDebugString(ctx, l.env, key, true /*remote*/, opts.VMMetadata.GetSnapshotId()))
```
maggie-lou, line 685: Should snapshotID be empty here given that it's the master snapshot?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -710,7 +719,7 @@ func (l *FileCacheLoader) cacheActionResult(ctx context.Context, key *fcpb.Snaps
 	if _, err := l.env.GetFileCache().Write(ctx, manifestNode, b); err != nil {
 		return err
 	}
-	log.CtxInfof(ctx, "Cached master local snapshot manifest %s", KeyDebugString(ctx, l.env, key, false /*remote*/))
+	log.CtxInfof(ctx, "Cached master local snapshot manifest %s", snapshotDebugString(ctx, l.env, key, false /*remote*/, opts.VMMetadata.GetSnapshotId()))
```
maggie-lou, line 722: Should snapshotID be empty here given that it's the master snapshot?

maggie-lou (approved):

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: Yeah, that makes sense since we're logging the snapshot-specific message below  - also changed "Cached" to "Updated" to help clarify that if this debug string is copied from the logs, it's not going to point to the specific snapshot that we cached here.

enterprise/server/remote_execution/snaploader/snaploader.go

PR #7029, opened 2024-07-16
Dedupe mysql requests to lookup the same API key

tylerwilliams:


enterprise/server/backends/authdb/authdb.go
```diff
@@ -350,45 +356,51 @@ func (d *AuthDB) GetAPIKeyGroupFromAPIKey(ctx context.Context, apiKey string) (i
...
-	qb := d.newAPIKeyGroupQuery(sd, true /*=allowUserOwnedKeys*/)
-	keyClauses := query_builder.OrClauses{}
-	if !*encryptOldKeys {
-		keyClauses.AddOr("ak.value = ?", apiKey)
-	}
-	if d.apiKeyEncryptionKey != nil {
-		encryptedAPIKey, err := d.encryptAPIKey(apiKey)
-		if err != nil {
-			return nil, err
+	akgI, err, _ := d.apiKeyFetchGroup.Do(apiKey, func() (interface{}, error) {
```
vadimberezniker, line 359: I think the key should be `cacheKey`. There can be two concurrent requests for the same API key on different subdomains which may return different results.

vadimberezniker (changes requested):

enterprise/server/backends/authdb/authdb.go
tylerwilliams: Done.

PR #7040, opened 2024-07-17
Update instructions for podman-static releases

tempoz:
- Switch to ssh-based repo, since (I think) most of us use keys to communicate with github
- `git apply` is for applying patch files; switch to `git cherry-pick` to cherry-pick a commit
- the first positional argument of `git push` is the remote repository, then all subsequent arguments are the refs to push. Switch to `git push origin <tag>`

**Related issues**: N/A

### DIFF
--- a/dockerfiles/executor_image/README.md
+++ b/dockerfiles/executor_image/README.md
@@ -28,7 +28,7 @@ To build a new version of podman from the fork:
 
 ```shell
 # Clone the fork and ensure you have the master branch checked out
-git clone https://github.com/buildbuddy-io/podman-static
+git clone git@github.com:buildbuddy-io/podman-static.git
 cd podman-static
 git checkout master
 # Add upstream as a remote and fetch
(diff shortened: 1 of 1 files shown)


dockerfiles/executor_image/README.md
```diff
@@ -28,7 +28,7 @@ To build a new version of podman from the fork:
 
 ```shell
 # Clone the fork and ensure you have the master branch checked out
-git clone https://github.com/buildbuddy-io/podman-static
+git clone git@github.com:buildbuddy-io/podman-static.git
```
bduffany, line 31: this change is fine, but fwiw I use the `gh` CLI which installs a credential helper that makes `https://` work properly

bduffany (approved):
thanks for fixing this!

dockerfiles/executor_image/README.md
bduffany: ah ok, either way I don't think this matters too much - this PR LGTM :)

PR #7042, opened 2024-07-17
[FC] Remove lapic=notscdeadline bootarg

maggie-lou:
The corresponding bug has been fix in firecracker v1.8 🎉 

As a note, I'm planning to release each of these firecracker config changes independently and with some time between so it will be easier to debug if there are problems

**Related issues**: N/A


bduffany:
This LGTM but could you also make check that the firecracker-sleep-issue script doesn't reproduce on a GCP VM? It wasn't clear to me in that bug whether their fix was AMD-specific or if would also fix the issue we were seeing on our intel VMs

tylerwilliams (approved):

PR #7043, opened 2024-07-17
[FC] Update guest kernel config

maggie-lou:
Update the guest kernel config to match the options recommended in the firecracker repo: https://github.com/firecracker-microvm/firecracker/blob/main/resources/guest_configs/microvm-kernel-ci-x86_64-5.10.config

We were previously using their 4.14 recommended config.

I changed a couple things from the firecracker config:
* Enabling FUSE to match this PR https://github.com/buildbuddy-io/buildbuddy/pull/945/files
* Enabling CONFIG_NETFILTER, CONFIG_IPV6, and related options which were needed for docker initialization to properly work (we were getting `nat` related networking errors)

I built the kernel by running `VERSION=v5.15 enterprise/vmsupport/kernel/rebuild.sh`.

**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3620


bduffany (approved):
LGTM - does this change let us remove `acpi=off`?

maggie-lou:
@bduffany Nice - this change combined with removing "noapic" (which I merged in a previous PR) lets us remove "acpi=off"

maggie-lou:
This actually causes docker to fail to initialize within the VM with `docker init timed out after 30s`, even with `acpi=off`. There is not a panic in the logs anymore though. Marking as WIP while I investigate more

maggie-lou:
> This actually causes docker to fail to initialize within the VM with `docker init timed out after 30s`, even with `acpi=off`. There is not a panic in the logs anymore though. Marking as WIP while I investigate more

I was able to fix this by enabling some network-related config options. These were set in our previous config, but disabled in the new one

tylerwilliams:
Nice :)

PR #7044, opened 2024-07-17
Add a flag that causes the executor to run 'podman system reset -f' on startup

iain-macdonald:
This is intended to help BYOR customers who 1. don't update/restart their executors that often and 2. occasionally run into issues upgrading their executors due to stale podman state. Siggi suggested we do something a bit more sophisticated like try to detect when the podman version changes, but I don't think that's sufficient as changes to some of the configuration options can cause inter-version issues too (e.g. changing how we apply `--transient-store`), so it would need to be a managed version within our code. For now, this seems simpler, but open to suggestions.

**Related issues**: N/A


iain-macdonald:
Closing, Brandon tells me the customer who might benefit from this persists their `/var/lib/containers` across executor restarts, so this won't be that useful.

PR #7048, opened 2024-07-18
In Cache Proxy, start a local ByteStreamServer on the internal gRPC port

iain-macdonald:
…and make it available in the environment. This local ByteStreamServer's going to be used by the ByteStreamProxyServer in https://github.com/buildbuddy-io/buildbuddy/pull/6946

The bufconn stuff is for future testing, when there'll need to be a "public" and an internal bufconn for connecting to the two different bytestreamservers (proxied and vanilla).

**Related issues**: N/A


server/real_environment/real_environment.go
```diff
@@ -225,6 +227,13 @@ func (r *RealEnv) SetLocalBufconnListener(l *bufconn.Listener) {
 	r.localBufconnListener = l
 }
 
+func (r *RealEnv) GetInternalLocalBufconnListener() *bufconn.Listener {
+	return r.internalLocalBufconnListener
+}
+func (r *RealEnv) SetInternalLocalBufconnListener(l *bufconn.Listener) {
+	r.internalLocalBufconnListener = l
+}
```
tylerwilliams, line 235: I don't think we should use this for new prod stuff -- if it's for testing could we put it somewhere else (testenv?) or call it TestOnlyBlah()?

tylerwilliams (changes requested):
LGTM minus the bufcon stuff

server/real_environment/real_environment.go
iain-macdonald: Yeah, this is only used for testing in https://github.com/buildbuddy-io/buildbuddy/pull/6946 I renamed these functions, and the ones for the other bufconn to `...ForTesting()`. It looks like we should be able to move these out of real_environment, but that change would touch a bunch of other files so I'll pursue it separately.

tylerwilliams (approved):

PR #7055, opened 2024-07-19
Add a couple of new action-merging metrics

iain-macdonald:
This PR adds two new metrics that track action-merging performance:
- `buildbuddy_remote_execution_merged_actions_per_execution`: a distribution tracking the number of merged actions per execution.
- `buildbuddy_remote_execution_merged_action_submit_time_offset_usec`: a distribution tracking the "submit time offset" for merged actions. The "submit time offset" is the wall time at which an action that is merged was submitted to the execution server minus the wall time at which the original execution it was merged against was submitted to the execution server. In a [previous PR](https://github.com/buildbuddy-io/buildbuddy/pull/6630) I called this the "time saved" due to action merging, but that's not really correct.

I'd like to add these as prometheus metrics because Siggi suggested that the current action-merging hedging implementation is a bit too eager in running hedges, and that I explore only running hedged executions when the original execution has been running for a little while, or only after a few other actions have been submitted (i.e. hedge every 5th submitted action). I looked into figuring out what those values should be by querying the Executions table in Clickhouse, but this: 1. requires running a kind of tricky query, 2. against a modifiable production database and 3. requires a bit of effort to keep up-to-date on, so I'd prefer to keep track of these basic metrics instead.

There are a couple of caveats with this PR. First, because we're writing to and reading from Redis for merged actions as well as executed actions, this will increase the load on Redis by about the rate of merged actions. This is about 10-50 QPS at peak (The query is `rate(sum(buildbuddy_remote_execution_merged_actions{region="${region}"}))`), and because we'll write and read, it'll be close to 100QPS extra at peak. This seems fine because we do 100k-150k QPS to Redis at peak, almost 50k of which are hash writes.

Second, the `buildbuddy_remote_execution_merged_actions_per_execution` metric is recorded once per submitted action, or many times per merged-against execution. This is a little bit awkward, and I thought about recording it on execution finalization so only the final value is recorded, but that would miss recording values if the merged-against execution doesn't finish correctly, for example because it times out, so I think the over-recording is worth it in this case.
 
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/action_merger/action_merger.go
+++ b/enterprise/server/remote_execution/action_merger/action_merger.go
@@ -167,11 +167,11 @@ func recordMetrics(ctx context.Context, rdb redis.UniversalClient, key string, g
 	if err := hash.Err(); err != nil {
 		log.Debugf("Error reading action-merging state from Redis: %s", err)
 	}
-	recordCountMetrics(ctx, hash, groupIdForMetrics)
+	recordCountMetric(ctx, hash, groupIdForMetrics)
 	recordSubmitTimeOffsetMetric(ctx, hash, groupIdForMetrics)
 }
 
-func recordCountMetrics(ctx context.Context, hash *redis.StringStringMapCmd, groupIdForMetrics string) {
+func recordCountMetric(ctx context.Context, hash *redis.StringStringMapCmd, groupIdForMetrics string) {
 	rawCount, ok := hash.Val()[actionCountKey]
 	if !ok {
 		return
--- a/enterprise/server/remote_execution/action_merger/action_merger.go
+++ b/enterprise/server/remote_execution/action_merger/action_merger.go
@@ -136,15 +136,7 @@ func RecordHedgedExecution(ctx context.Context, rdb redis.UniversalClient, adRes
 		return err
 	}
 
-	pipe := rdb.TxPipeline()
-	pipe.HIncrBy(ctx, key, executionCountKey, 1)
-	pipe.HIncrBy(ctx, key, actionCountKey, 1)
-	if _, err = pipe.Exec(ctx); err != nil {
-		return err
-	}
-
-	recordMetrics(ctx, rdb, key, groupIdForMetrics)
-	return nil
+	return rdb.HIncrBy(ctx, key, executionCountKey, 1).Err()
 }
 
 // Records a merged execution.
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/action_merger/action_merger.go
```diff
@@ -117,12 +130,79 @@ func RecordClaimedExecution(ctx context.Context, rdb redis.UniversalClient, exec
...
+	if !ok {
+		return
+	}
+	count, err := strconv.Atoi(rawCount)
+	if err != nil {
+		return
+	}
+
+	metrics.RemoteExecutionMergedActionsPerExecution.
+		With(prometheus.Labels{metrics.GroupID: groupIdForMetrics}).
```
vadimberezniker, line 185: It seems like this metric is going to be a little inaccurate. We are going to record multiple observations for the same digest every time an action is merged. And I think because RecordMergedExecution and RecordHedgedExecution can both be called it might be observed multiple times for the same execution request.
First request: no observation
Second request with hedge: Observe(1), Observe(2) <-- this seems wrong?
Third request without hedge: Observe(3)
In terms of repeating observations, I think this is the best we can do but we shouldn't count multiples times when a request is hedged.

vadimberezniker (changes requested):

enterprise/server/remote_execution/action_merger/action_merger.go
iain-macdonald: Oh shoot, good catch! Fixed.

enterprise/server/remote_execution/action_merger/action_merger.go
```diff
@@ -117,14 +130,69 @@ func RecordClaimedExecution(ctx context.Context, rdb redis.UniversalClient, exec
...
 	key, err := redisKeyForPendingExecutionID(ctx, adResource)
 	if err != nil {
 		return err
 	}
+
 	return rdb.HIncrBy(ctx, key, executionCountKey, 1).Err()
 }
 
+// Records a merged execution.
+func RecordMergedExecution(ctx context.Context, rdb redis.UniversalClient, adResource *digest.ResourceName, groupIdForMetrics string) error {
```
bduffany, line 143: nit: mention where the execution is being recorded?

enterprise/server/remote_execution/action_merger/action_merger.go
```diff
@@ -117,14 +130,69 @@ func RecordClaimedExecution(ctx context.Context, rdb redis.UniversalClient, exec
...
+	hash := rdb.HGetAll(ctx, key)
+	if err := hash.Err(); err != nil {
+		log.Debugf("Error reading action-merging state from Redis: %s", err)
+	}
+	recordCountMetric(ctx, hash, groupIdForMetrics)
+	recordSubmitTimeOffsetMetric(ctx, hash, groupIdForMetrics)
+	return nil
+}
+
+func recordCountMetric(ctx context.Context, hash *redis.StringStringMapCmd, groupIdForMetrics string) {
```
bduffany, line 162: nit: you could write `hash, err := rdb.HGetAll(ctx, key).Result()` so that `hash` winds up being a plain old `map[string]string` instead of this `*redis.StringStringMapCmd` type

bduffany (approved):

enterprise/server/remote_execution/action_merger/action_merger.go
iain-macdonald: done

enterprise/server/remote_execution/action_merger/action_merger.go
iain-macdonald: done

vadimberezniker (approved):

PR #7058, opened 2024-07-22
pebble cache: add a flag to enable bloom filter creation.

jdhollen:
It's kinda tough to get a read on how much of a difference this makes in a real-ish environment.

I ran a little test harness where I hacked up cacheload to:

- write a few million 500-1000 byte entries to the pebble cache
- call findmissing with random (i.e., near-certain miss) digests as fast as possible

CPU profiles showed table read CPU decreasing by ~33% on FindMissing misses (so, disk reads and table loads decreased) in exchange for essentially-zero (<.1%) CPU hit on checking bloom filters.

qps perf, on the other hand, was pretty variable. i won't pretend to have a meaningful read on it; in both cases my load test environment came up with something like 30-40k read qps (single process requesting + receiving).  Performance across runs wasn't consistent enough to pretend i'm not being influenced by my machine conditions.  It ~ felt ~ like the fastest runs were with bloom filters on, but some bloom filter runs were slower than the fastest no-filter runs.

There's some memory overhead from having the bloom filters for each sstable loaded up, but that's kinda tough to judge given that my load test is ~bogus.  The documentation says the bloom.FilterPolicy(10) should take "approximately" 10 bits per key, which is perfectly fine for us assuming it's accurate.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache.go
@@ -501,7 +501,7 @@ func defaultPebbleOptions(el *pebbleEventListener) *pebble.Options {
 			DiskSlow:        el.DiskSlow,
 		},
 	}
-	if *enableBloomFilter {
+	if *enableTableBloomFilter {
 		opts.Levels = []pebble.LevelOptions{
 			pebble.LevelOptions{
 				FilterPolicy: pebble.BloomFilterPolicy(10),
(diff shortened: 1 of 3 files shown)


vadimberezniker:
IIRC bloom filters are only used for specialized operations which we are not using, or at least that used to be the case

tylerwilliams (approved):
// FilterPolicy implements the FilterPolicy interface from the pebble package.
//
// The integer value is the approximate number of bits used per key. A good
// value is 10, which yields a filter with ~ 1% false positive rate.

tylerwilliams:
> IIRC bloom filters are only used for specialized operations which we are not using, or at least that used to be the case

We used to rely on iter.SeekGE() / iter.Next() in FindMissing but now we call Get, so I think this may have some benefits for the case when FindMissing is called with a digest that has not been written yet.

jdhollen:
> > IIRC bloom filters are only used for specialized operations which we are not using, or at least that used to be the case
> 
> We used to rely on iter.SeekGE() / iter.Next() in FindMissing but now we call Get, so I think this may have some benefits for the case when FindMissing is called with a digest that has not been written yet.

Right--this should help primarily with FindMissing but I'm hoping it will have some knock-on effects because we'll be loading fewer tables during FindMissing calls as well.

PR #7061, opened 2024-07-22
[ci_runner] Populate child invocation cards in UI for bazel commands in bash

maggie-lou:
On the outer workflow invocation page, we have child invocation cards linking to each inner invocation. We emit the ChildInvocationsConfigured and ChildInvocationCompleted events to trigger the UI to show these cards.

Now that we're accepting arbitrary bash commands, we must scan the command output logs to identify bazel commands, so we know to emit these events.

**Related issues**: N/A

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -772,14 +772,6 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 		Env:            envVars,
 		ExecProperties: platform.Properties,
 		RunRemotely:    *runRemotely,
-		Steps: []*rnpb.Step {
-			{
-				Run: "bazel build //server/util/lru/...",
-			},
-			{
-				Run: "bazel build //server/util/lru/... && bazel build //server/util/lru/... && bazel version",
-			},
-		},
 	}
 	req.GetRepoState().Patch = append(req.GetRepoState().Patch, repoConfig.Patches...)
 
@@ -989,8 +981,8 @@ func parseArgs(commandLineArgs []string) (bazelArgs []string, execArgs []string,
 	// app backend as the remote runner.
 	bazelArgs = arg.Remove(bazelArgs, "bes_backend")
 	bazelArgs = arg.Remove(bazelArgs, "remote_cache")
-	bazelArgs = append(bazelArgs, "--bes_backend=grpc://10.128.15.217:1985")
-	bazelArgs = append(bazelArgs, "--remote_cache=grpc://10.128.15.217:1985")
+	bazelArgs = append(bazelArgs, "--bes_backend="+*remoteRunner)
+	bazelArgs = append(bazelArgs, "--remote_cache="+*remoteRunner)
 
 	return bazelArgs, execArgs, nil
 }
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -570,7 +570,7 @@ func downloadFile(ctx context.Context, bsClient bspb.ByteStreamClient, resourceN
 }
 
 func lookupBazelInvocationOutputs(ctx context.Context, bbClient bbspb.BuildBuddyServiceClient, invocationID string) ([]*bespb.File, error) {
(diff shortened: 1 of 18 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2526,11 +2626,23 @@ func runBazelWrapper() error {
...
+	iid, err := newUUID()
+	if err != nil {
+		return err
+	}
+	args = appendBazelSubcommandArgs(args, fmt.Sprintf("--invocation_id=%s", iid))
+	args = append(startupArgs, args...)
+
+	io.WriteString(os.Stdout, fmt.Sprintf("BB command for invocation [%s]: [%s]", iid, strings.Join(os.Args[1:], " ")))
+	start := time.Now()
+	err = runCommand(context.Background(), bazelCmd, args, nil, "", os.Stderr)
```
maggie-lou, line 2639: @bduffany You've previously mentioned how we should use `exec` to avoid having child processes. I couldn't think of a better way to capture the exit code of the bazel commands though. Let me know if you have any better ideas

Another way would be to stop including this information in the BES events and have the UI fetch each child invocation to get the status, but that felt like a poor solution because we may have to fetch the invocations thousands of times for long builds

bduffany: hm I wonder if we could simplify by storing the parent workflow invocation ID in the invocation table, and then when we call LookupInvocation for displaying the invocation in the UI, we also fetch child invocations and return their metadata back to the UI?

I think we already set the necessary metadata on the child invocation (since we render the parent workflow ID in the breadcrumbs) but just aren't storing it in the DB

(tbh I'm not sure why I didn't consider this option before... are there any downsides you can think of?)

bduffany: oh, I think that's what you meant by

> Another way would be to stop including this information in the BES events and have the UI fetch each child invocation to get the status, but that felt like a poor solution because we may have to fetch the invocations thousands of times for long builds

I think it's fine as long as we're only fetching the child invocations' DB row, and not their full BES stream

maggie-lou: Sounds good - this will also fix a confusing UI experience where sometimes from the outer invocation, the duration and status of the builds is incorrect, because there are 2 sources of truth (outer invocation uses BES event, inner invocation uses invocation DB row)

maggie-lou: @bduffany Cleaned up this PR after refactoring the FE to fetch the child invocations directly from the db

bduffany (approved):

cli/remotebazel/remotebazel.go
```diff
@@ -813,25 +825,42 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
...
+		if err != nil {
+			return fmt.Errorf("could not retrieve invocation: %s", err)
+		}
+		if len(inRsp.GetInvocation()) == 0 {
+			return fmt.Errorf("invocation not found")
+		}
+		return nil
+	})
+	eg.Go(func() error {
+		exRsp, err = bbClient.GetExecution(ctx, &espb.GetExecutionRequest{ExecutionLookup: &espb.ExecutionLookup{
```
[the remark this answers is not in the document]
maggie-lou: `Invocation` only has `bazel_exit_code` which is a string. I figured we should try to return the exact exit code the bazel command returned, so this simulates running a normal bazel command as closely as possible

proto/build_event_stream.proto
```diff
@@ -1456,6 +1456,7 @@ message WorkflowConfigured {
 // Event describing a workflow command that completed.
 // Note: the event ID holds the invocation ID that identifies the command.
 // Next Tag: 6
+// TODO(Maggie): Deprecate - duplicate with ChildInvocationCompleted
```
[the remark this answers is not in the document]
maggie-lou: Good idea - I also cleaned up the FE reference to this event (it's unneeded because ChildInvocationCompleted currently contains the same data) and added a TODO to rename it to `RemoteRunnerStepCompleted`

PR #7066, opened 2024-07-23
Refactor `SetupVethPair` to prepare for reuse with OCI containers

bduffany:
* Remove `vmIdx` parameter of `SetupVethPair` - it is only used to keep track of the next host IP that we should try to reserve. This can instead be an internal implementation detail of `HostNetAllocator`, instead of relying on the caller to "hint" at what the next "likely free" index will be.
* Remove TAP device configuration from `CreateVethPair` and into a separate function (more context/discussion to follow in Slack)

**Related issues**: N/A

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -244,24 +244,18 @@ func (a *HostNetAllocator) Get() (*HostNet, error) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 	for attempt := 0; attempt < numAssignableNetworks; attempt++ {
-		netIdx := a.idx % numAssignableNetworks
+		netIdx := a.idx
 		a.idx = (a.idx + 1) % numAssignableNetworks
-		if a.tryLock(netIdx) {
-			return &HostNet{
-				netIdx: netIdx,
-				unlock: func() { a.unlock(netIdx) },
-			}, nil
+		if a.inUse[netIdx] {
+			continue
 		}
+		a.inUse[netIdx] = true
+		return &HostNet{
+			netIdx: netIdx,
+			unlock: func() { a.unlock(netIdx) },
+		}, nil
 	}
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
LGTM but getting a review from @vadimberezniker and @maggie-lou too probably wouldn't hurt :)

server/util/networking/networking.go
```diff
@@ -236,9 +240,12 @@ func (n *HostNet) Unlock() {
 }
 
 // Get assigns a host network IP for the given VM index.
-func (a *HostNetAllocator) Get(vmIdx int) (*HostNet, error) {
+func (a *HostNetAllocator) Get() (*HostNet, error) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
 	for attempt := 0; attempt < numAssignableNetworks; attempt++ {
-		netIdx := (vmIdx + attempt) % numAssignableNetworks
+		netIdx := a.idx % numAssignableNetworks
```
maggie-lou, line 247: nit: don't need to % numAssignableNetworks if you're doing it in the line below

server/util/networking/networking.go
```diff
@@ -236,9 +240,12 @@ func (n *HostNet) Unlock() {
...
+		a.idx = (a.idx + 1) % numAssignableNetworks
 		if a.tryLock(netIdx) {
```
maggie-lou, line 249: nit: Maybe rename this given we're not locking anymore?

maggie-lou (approved):

server/util/networking/networking.go
bduffany: good catch, removed

server/util/networking/networking.go
bduffany: I just inlined the logic to avoid confusion

vadimberezniker (approved):

PR #7068, opened 2024-07-23
Do more to pick the correct size buffer

tylerwilliams:
For TreeCache entries, make it possible to know the exact buffer size. And for CAS objects, use the digest size.

### DIFF
--- a/enterprise/server/backends/distributed/distributed.go
+++ b/enterprise/server/backends/distributed/distributed.go
@@ -1076,6 +1076,9 @@ func (c *Cache) distributedReader(ctx context.Context, rn *rspb.ResourceName, of
 	return nil, status.NotFoundErrorf("Exhausted all peers attempting to read %q.", rn.GetDigest().GetHash())
 }
 
+// Below, in Get(), this value is the max initial allocatable buffer size.
+// Set it somewhat conservatively so that we're not DOSed by someone crafting
+// remote_instance_names that match this just to use memory.
 const maxInitialByteBufferSize = (1024 * 1024 * 4)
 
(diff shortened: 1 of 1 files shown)


enterprise/server/backends/distributed/distributed.go
```diff
@@ -1080,7 +1082,23 @@ func (c *Cache) Get(ctx context.Context, rn *rspb.ResourceName) ([]byte, error)
...
+
+	var buf *bytes.Buffer
+	if rn.GetCacheType() == rspb.CacheType_CAS {
+		// If this is a CAS object, size the buffer to fit exactly.
+		buf = bytes.NewBuffer(make([]byte, 0, int(rn.GetDigest().GetSizeBytes())))
+	} else if strings.HasPrefix(rn.GetInstanceName(), content_addressable_storage_server.TreeCacheRemoteInstanceName) {
+		// If this is a TreeCache entry that we wrote; pull the size
+		// from the remote instance name.
+		parts := strings.Split(rn.GetInstanceName(), "/")
+		if s, err := strconv.Atoi(parts[len(parts)-1]); err == nil {
```
jdhollen, line 1094: this line looks like a well-crafted instance name could cause us to allocate an arbitrary-length array (if i'm understanding right).  could we clamp this value to something? (lmk if i'm misunderstanding)

siggisim (approved):

enterprise/server/backends/distributed/distributed.go
tylerwilliams: yeah that's a good point. I'll cap it to ~few MB

tylerwilliams: Done.

jdhollen (approved):

PR #7071, opened 2024-07-24
Support deprecation warnings in the console

tylerwilliams:
Big ups @fmeum for the help :)

Allow for publishing a deprecation warning in the bazel console when a request uses some feature that we're planning to drop. Do this by returning an error in the bytestream.Write call of the profile -- but still let that upload succeed so we don't impact profile uploads.

### DIFF
--- a/server/util/bazel_deprecation/bazel_deprecation.go
+++ b/server/util/bazel_deprecation/bazel_deprecation.go
@@ -6,7 +6,6 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
 	"github.com/buildbuddy-io/buildbuddy/server/util/bazel_request"
 	"github.com/buildbuddy-io/buildbuddy/server/util/claims"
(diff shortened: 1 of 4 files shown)


server/util/bazel_deprecation/bazel_deprecation.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if w.getWarningCount(ctx) >= 1 {
+		// Already warned.
+		return nil
+	}
+
+	if isAnonymousBuild(ctx) {
+		defer w.incrementWarningCount(ctx)
+		return status.FailedPreconditionError("Anonymous access is deprecated; please create an account at https://app.buildbuddy.io/")
```
fmeum, line 69: How about prefixing this so that it's clear that this message is coming from BuildBuddy, not Bazel itself? The structure of the error message doesn't make that clear.

tylerwilliams: Done.

siggisim (approved):

PR #7076, opened 2024-07-24
Locally cache unary requests in content_addressable_storage_server_proxy

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/content_addressable_storage_server_proxy/content_addressable_storage_server_proxy.go
+++ b/enterprise/server/content_addressable_storage_server_proxy/content_addressable_storage_server_proxy.go
@@ -88,7 +88,22 @@ func (s *CASServerProxy) BatchReadBlobs(ctx context.Context, req *repb.BatchRead
 	if len(mergedResp.Responses) == len(req.Digests) {
 		return &mergedResp, nil
 	}
+
+	// digest.Diff returns a set of differences between two sets of digests,
+	// but the protocol requires the server return multiple responses if the
+	// same digest is requested multiple times. Count the number of client
+	// requests per blob so we can duplicate responses that many times before
+	// returning to the client.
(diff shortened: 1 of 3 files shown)


iain-macdonald:
Ping, any chance you can take a look at this today?

enterprise/server/content_addressable_storage_server_proxy/content_addressable_storage_server_proxy.go
```diff
@@ -6,58 +6,131 @@ import (
...
+	}
+	for _, resp := range localResp.Responses {
+		if resp.Status.Code == int32(codes.OK) {
+			mergedResp.Responses = append(mergedResp.Responses, resp)
+			mergedDigests = append(mergedDigests, resp.Digest)
+		}
+	}
+	if len(mergedResp.Responses) == len(req.Digests) {
+		return &mergedResp, nil
+	}
```
tylerwilliams, line 90: I think it is legal (in the protocol) for a client to send the same digest twice. If that happens, will local/remote BatchReadBlobs return both or dedupe?

If it returns both, this code is correct. If it dedupes, then I think you have an issue where you'd fall through and digests might be an empty list?

tylerwilliams (approved):

enterprise/server/content_addressable_storage_server_proxy/content_addressable_storage_server_proxy.go
iain-macdonald: This was actually working due to a bug in `digest.Diff` that I fixed in https://github.com/buildbuddy-io/buildbuddy/commit/d8461cff320849f1f9dc8aef322a15a525f76f10. I added a test, rebased, and fixed the bug.

PR #7078, opened 2024-07-24
Don't write to mysql so much

tylerwilliams:
all credit to @jdhollen and @bduffany for identifying this -- this is old code that we still have hanging around but isn't really doing anything useful anymore (and is causing unnecessary load on the database).

Write once the first time through the loop to set stage = executing, and then let everything else work the way it was before.

### DIFF
--- a/enterprise/server/remote_execution/execution_server/execution_server.go
+++ b/enterprise/server/remote_execution/execution_server/execution_server.go
@@ -832,7 +832,7 @@ func (s *ExecutionServer) waitExecution(ctx context.Context, req *repb.WaitExecu
 	}
 }
 
-func loopAfterTimeout(ctx context.Context, timeout time.Duration, f func() bool ) {
+func loopAfterTimeout(ctx context.Context, timeout time.Duration, f func() bool) {
 	ticker := time.NewTicker(timeout)
 	defer ticker.Stop()
 	for {
--- a/enterprise/server/remote_execution/execution_server/execution_server.go
+++ b/enterprise/server/remote_execution/execution_server/execution_server.go
@@ -832,7 +832,7 @@ func (s *ExecutionServer) waitExecution(ctx context.Context, req *repb.WaitExecu
 	}
 }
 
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -919,7 +901,15 @@ func (s *ExecutionServer) PublishOperation(stream repb.Execution_PublishOperatio
...
-		taskID = op.GetName()
+
+		// If this is the first op received from Stream, update the
+		// execution status in MySQL.
+		if op.GetName() != taskID {
+			taskID = op.GetName()
+			if err := s.updateExecution(ctx, taskID, stage, lastOp); err != nil {
+				return err
+			}
+		}
```
bduffany, line 912: previously we'd skip this write if the execution took less than 5s. Does it seem worth trying to preserve that behavior?

bduffany (approved):

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -883,29 +883,11 @@ func (s *ExecutionServer) PublishOperation(stream repb.Execution_PublishOperatio
 	if err != nil {
 		return err
 	}
-	lastOp := &longrunning.Operation{}
-	lastWrite := time.Now()
+	var lastOp *longrunning.Operation
 	taskID := ""
 	// Once the executor has called PublishOperation, we're in EXECUTING stage.
 	stage := repb.ExecutionStage_EXECUTING
 	mu := sync.Mutex{}
```
bduffany, line 890: nit: I think the mutex isn't needed anymore since we don't have any concurrent calls to `updateExecution` in this func?

enterprise/server/remote_execution/execution_server/execution_server.go
bduffany: resolved offline

jdhollen (approved):

bduffany (approved):

PR #7079, opened 2024-07-24
[raft] wait till the replicas are loaded and caught up.

luluz66:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3625

Added a health check in raft to check whether replicas that started from disk on
ready to serve.

1. Before we call StartOnDiskReplica, we will mark them "started". Note that,
   replicas created by AddReplica or Split won't have this status.
2. The first time when the store received AddRange from the replica with a non-empty range, we 
will initiate a goroutine to start wait for the replica to catch up. 
3. For some zombie shards that were created by a failed attempt to split, we
   will wait up to maxWaitTimeForReplicaRange.

Also add the init status to statuz to make debugging easier.

### DIFF
--- a/enterprise/server/raft/store/store.go
+++ b/enterprise/server/raft/store/store.go
@@ -76,6 +76,10 @@ var (
 
 const (
 	deleteSessionsRateLimit = 1
+	// The number of go routines we use to wait for the replicas to be ready
+	readyNumGoRoutines           = 100
+	checkReplicaCaughtUpInterval = 1 * time.Second
+	maxWaitTimeForReplicaRange   = 30 * time.Second
 )
 
 type Store struct {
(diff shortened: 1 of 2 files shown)


enterprise/server/raft/store/store.go
```diff
@@ -1314,6 +1337,235 @@ func (s *Store) Usage() *rfpb.StoreUsage {
...
+		return "started"
+	case replicaInitStatusInProgress:
+		return "in-progress"
+	case replicaInitStatusCompleted:
+		return "completed"
+	}
+	return "unknown"
+}
+
+func (w *replicaStatusWaiter) MaybeWaitReplicaToCatchupInBackground(repl *replica.Replica) error {
```
tylerwilliams, line 1408: nit: MaybeWaitForReplicaToCatchupInBackground or just WaitReplicaToCatchupInBackground

tylerwilliams (approved):

enterprise/server/raft/store/store.go
luluz66: Done

PR #7080, opened 2024-07-24
Fix container stats baseline for recycled runners

bduffany:
The current call to `Reset()` sets the stats baseline to the last observed value, but if a runner has background processes running, then the last observed value won't account for any usage from background processes since the last call to `Exec()` finished.

There is some evidence that this is causing task sizes to be inflated - some tasks with runner recycling enabled have very high pressure stall times which are close to or exceed the task's execution duration, which doesn't make much sense. But this could be explained by the fact that the current `Reset()` implementation ignores any resource usage since the previous task finished its `Exec()` call (e.g. while it was uploading outputs).

**Related issues**: N/A


iain-macdonald (approved):
LGTM, Approval

PR #7087, opened 2024-07-25
Add a firecracker test with networking after resuming from snapshot

bduffany:
The current networking setup is designed to ensure that networking can still work after resuming from snapshot, following this guide: https://github.com/firecracker-microvm/firecracker/blob/main/docs%2Fsnapshotting%2Fnetwork-for-clones.md

This PR adds a dedicated test to make sure we're testing networking after resuming from snapshot.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
@@ -1161,6 +1161,56 @@ func TestFirecrackerRunWithNetwork(t *testing.T) {
 	assert.Contains(t, string(res.Stdout), "64 bytes from "+googleDNS)
 }
 
+func TestSnapshotAndResumeWithNetwork(t *testing.T) {
+	ctx := context.Background()
+	env := getTestEnv(ctx, t, envOpts{})
+	rootDir := testfs.MakeTempDir(t)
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -1161,6 +1161,55 @@ func TestFirecrackerRunWithNetwork(t *testing.T) {
...
+	assert.Contains(t, string(res.Stdout), "64 bytes from "+googleDNS)
+
+	err = c.Pause(ctx)
+	require.NoError(t, err)
+
+	err = c.Unpause(ctx)
+	require.NoError(t, err)
+
+	res = c.Exec(ctx, cmd, &interfaces.Stdio{})
+	require.NoError(t, res.Error)
```
maggie-lou, line 1207: Sanity check add 

assert.Equal(t, 0, res.ExitCode)
	assert.Contains(t, string(res.Stdout), "64 bytes from "+googleDNS)

here too?

bduffany: woops, thanks for catching this! (it's the whole point of the test haha)

PR #7092, opened 2024-07-26
Implement networking for OCI runtime

bduffany:
Not much interesting new code is being added here except a new `iptables` rule in `SetupVethPair` to prevent traffic between containers. Everything else is mostly plumbing / tests.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy/pull/7093

### DIFF
--- a/server/testutil/testnetworking/testnetworking.go
+++ b/server/testutil/testnetworking/testnetworking.go
@@ -1,0 +1,41 @@
+package testnetworking
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/buildbuddy-io/buildbuddy/server/testutil/testfs"
+	"github.com/stretchr/testify/require"
+)
+
+// Setup sets up the test to be able to call networking functions.
+// If skips the test if the required net tools aren't available.
+func Setup(t *testing.T) {
+	// Ensure ip tools are in PATH
+	os.Setenv("PATH", os.Getenv("PATH")+":/usr/sbin:/sbin")
+
+	// Make sure the 'ip' tool is available and that we have the necessary
+	// permissions to use it.
+	cmd := []string{"ip", "link"}
+	if os.Getuid() != 0 {
+		cmd = append([]string{"sudo", "--non-interactive"}, cmd...)
+	}
+	if b, err := exec.Command(cmd[0], cmd[1:]...).CombinedOutput(); err != nil {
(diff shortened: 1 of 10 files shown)


fmeum (approved):

server/util/networking/networking.go
```diff
@@ -443,23 +429,56 @@ func SetupVethPair(ctx context.Context, netNamespace string) (_ *VethPair, err e
...
+	iptablesRules := [][]string{
+		// Allow forwarding traffic from the host side of the veth pair to the
+		// device associated with the configured route prefix (usually the
+		// default route). This is necessary on hosts with default-deny policies
+		// in place.
+		{"FORWARD", "-i", veth1, "-o", device, "-j", "ACCEPT"},
+
+		// Drop any traffic from the namespace that is targeting another
+		// namespace.
+		{"FORWARD", "-s", network.CIDR(), "-d", containerNetworkingCIDR, "-j", "DROP"},
```
vadimberezniker, line 441: Containers shouldn't be addressing any private addresses so you might as well block all the ranges in `PrivateIPRanges`.

vadimberezniker (approved):

iain-macdonald (approved):
LGTM, Approval

server/util/networking/networking.go
bduffany: We also have `executor.blackhole_private_ranges` which should disallow traffic to private ranges, but it felt somewhat incidental that `containerNetworkingCIDR` happens to be one of the private IP ranges so I thought it was worth an explicit rule. 

If we want to consolidate the handling of these DROP rules and maybe get rid of that blackhole_private_ranges flag, I can do that in a separate PR - wdyt?

server/testutil/testnetworking/testnetworking.go
```diff
@@ -0,0 +1,41 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/buildbuddy-io/buildbuddy/server/testutil/testfs"
+	"github.com/stretchr/testify/require"
+)
+
+// Setup sets up the test to be able to call networking functions.
+// If skips the test if the required net tools aren't available.
```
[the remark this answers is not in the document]
bduffany: woops, fixed

server/util/networking/networking.go
bduffany: (Will discuss more offline and follow up in another PR)

PR #7093, opened 2024-07-26
Mount /dev/shm in OCI containers

bduffany:
Not having `/dev/shm/` mounted was causing webdriver tests to fail. When run with `--verbose` mode, Chromium reports failures relating to `/dev/shm`. Confirmed that tests pass on a local executor after enabling this mount.

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -588,13 +588,12 @@ func (c *ociContainer) createSpec(cmd *repb.Command) (*specs.Spec, error) {
 				Source:      filepath.Join(c.bundlePath(), "hosts"),
 				Options:     []string{"bind", "rprivate"},
 			},
-			// TODO: shm
-			// {
-			// 		Destination: "/dev/shm",
-			// 		Type:        "bind",
-			// 		Source:      "/run/containers/storage/overlay-containers/99133d16f4f9d0678f87972c01209e308ebafc074f333822805a633620f12507/userdata/shm",
-			// 		Options:     []string{"bind", "rprivate", "nosuid", "noexec", "nodev"},
-			// },
+			{
+				Destination: "/dev/shm",
+				Type:        "tmpfs",
+				Source:      "tmpfs",
+				Options:     []string{"rw", "nosuid", "nodev", "noexec", "relatime", "size=64000k", "inode64"},
+			},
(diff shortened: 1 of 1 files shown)


fmeum (approved):

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -588,13 +588,12 @@ func (c *ociContainer) createSpec(cmd *repb.Command) (*specs.Spec, error) {
...
-			// 		Destination: "/dev/shm",
-			// 		Type:        "bind",
-			// 		Source:      "/run/containers/storage/overlay-containers/99133d16f4f9d0678f87972c01209e308ebafc074f333822805a633620f12507/userdata/shm",
-			// 		Options:     []string{"bind", "rprivate", "nosuid", "noexec", "nodev"},
-			// },
+			{
+				Destination: "/dev/shm",
+				Type:        "tmpfs",
+				Source:      "tmpfs",
+				Options:     []string{"rw", "nosuid", "nodev", "noexec", "relatime", "size=64000k", "inode64"},
```
fmeum, line 595: Not sure which of these isn't supported in CI. Could we maybe print the output of something like `journalctl` to see kernel logs when the test fails?

bduffany: The failures are happening only in dev and seem to be fixed after this revert PR https://github.com/buildbuddy-io/buildbuddy/pull/7096

iain-macdonald: Does it make sense to make the size configurable via a flag, or do you think we're unlikely to have to change it for running on different machines?

iain-macdonald (commented):
See comments

bduffany:
> Does it make sense to make the size configurable via a flag, or do you think we're unlikely to have to change it for running on different machines?

This is the default size that podman uses and it seems to be working OK for people so far, so I think we can wait until someone asks to make it configurable

iain-macdonald:
> > Does it make sense to make the size configurable via a flag, or do you think we're unlikely to have to change it for running on different machines?
> 
> This is the default size that podman uses and it seems to be working OK for people so far, so I think we can wait until someone asks to make it configurable

sg!

iain-macdonald (approved):

bduffany:
Going to merge because the failures in dev aren't specifically related to this PR.

PR #7098, opened 2024-07-26
Print an --override_repository arg after building the kernel

bduffany:
Allows skipping the GCS upload / download and building the executor directly against the built `vmlinux` artifact. Useful for quickly testing kernel config options.

Also bump the default kernel version to v5.15 since that's what we're using.

**Related issues**: N/A


bduffany:
Going to put a little more thought into just storing these in git instead (when I have more time). Seems like it might be fine since our workflows are cloning snapshots consistently, and GH workflows clone the BB repo pretty quickly

PR #7099, opened 2024-07-26
[FC] Fix OCI tests by enabling SECCOMP

maggie-lou:
Fix the OCI tests. They run in firecracker containers, and after the kernel config update, started failing with errors like ` create container: {"msg":"prctl: Invalid argument","level":"error","time":"2024-07-26T01:55:10.449777Z"}`. Setting the `CONFIG_SECCOMP=y` kernel option makes them run successfully again.

**Related issues**: N/A


bduffany (approved):
nice! thanks for fixing this

PR #7102, opened 2024-07-29
Add an option to only fetch invocation metadata from GetInvocation

maggie-lou:
**Related issues**: BE PR to support FE PR https://github.com/buildbuddy-io/buildbuddy/pull/7138

* Also added support for the FE to handle multiple `childInvocationsConfigured` events, because we will emit one event per bazel command detected in bash blocks


proto/invocation.proto
```diff
@@ -167,6 +167,10 @@ enum InvocationPermission {
 message InvocationLookup {
   // The invocation_id: a UUID generated by the bazel run.
   string invocation_id = 1;
+
+  // If true, includes additional build metadata including the scorecard,
+  // BES events, structured command line, and console buffer.
+  bool include_metadata = 2;
```
bduffany, line 173: nit: I think we consider the metadata to be the stuff in the DB, so this naming seems to do the opposite of what we might expect.

Also I think instead of having this in the GetInvocation API and doing a bunch of fetches on the client, we should have the server fetch the workflow child invocation metadata (`Invocation` proto) and have a `child_invocations` field in the returned `Invocation`

maggie-lou: We fetch the parent invocation every ~3 seconds. My plan is:
1. If there is a new ChildConfiguredEvent that we haven't seen yet, the FE will fetch that invocation to pull the bazel command to populate the card.
2. If there is a new ChildCompletedEvent that we havent seen yet, the FE will fetch the invocation to pull the final status and duration to populate the card.
3. The FE memoizes the Child<Configured, Completed> events

That way we don't have to keep fetching the children, whether in the FE or BE. We need very limited data from the child invocations, and we don't need any additional data while each child invocation is in process

(I'll think on a better name of the field as well)

bduffany: > If there is a new ChildConfiguredEvent that we haven't seen yet

I think then you'd need to parse the bazel command output in order to publish these events right?

> This could save us thousands of DB reads for long workflows

Looking up invocation metadata in the DB is extremely cheap especially when running the same query multiple times (it should be cached in memory). I don't think we need to worry about this especially since we already poll the DB for looking up the parent invocation - so it's not like we're adding an order of magnitude more additional reads

maggie-lou: Ah I see - I'd misunderstood your original comment. Unfortunately we'd still need to detect when bazel is called to publish other BES events - like `TargetCompleted`, which we use to render workflow artifacts

Also where do we cache DB reads in memory?

maggie-lou: Discussed offline and decided to keep going with this strategy (https://github.com/buildbuddy-io/buildbuddy/pull/7102#discussion_r1695537032)

app/invocation/child_invocations.tsx
```diff
@@ -16,13 +16,14 @@ export default class ChildInvocations extends React.Component<ChildInvocationPro
...
-      ? workflowConfiguredEvent.invocation
-      : [];
+    const childInvocationConfiguredEvents = this.props.model.childInvocationsConfigured;
+    let invocations = [];
+    for (let i = 0; i < childInvocationConfiguredEvents.length; i++) {
+      const event = childInvocationConfiguredEvents[i];
+      for (let inv of event.invocation) {
+        invocations.push(inv);
+      }
+    }
```
bduffany, line 26: should these typescript changes go in a separate PR? at a quick glance they seem unrelated to the `MetadataOnly` support unless I'm missing something?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -112,6 +113,14 @@ func (s *BuildBuddyServer) GetInvocation(ctx context.Context, req *inpb.GetInvoc
 	if req.GetLookup().GetInvocationId() == "" {
 		return nil, status.InvalidArgumentErrorf("GetInvocationRequest must contain a valid invocation_id")
 	}
+	if req.Lookup.MetadataOnly {
```
bduffany, line 116:
```suggestion
	if req.GetLookup().GetMetadataOnly() {
```

bduffany (approved):

app/invocation/child_invocations.tsx
maggie-lou: They need to go out before the follow up PR #7138 so I just threw them in here.  But it is a little randomly shoved into here 😅

PR #7107, opened 2024-07-29
Only index lowercase ngrams (still use case in regex match tho).

tylerwilliams:
The sparse ngrams are longer than trigrams so I don't want to do case expansion on them (casE, caSe, cAse, Case, caSE...).

### DIFF
--- a/codesearch/token/token_test.go
+++ b/codesearch/token/token_test.go
@@ -4,9 +4,12 @@ import (
 	"bytes"
 	"io"
 	"sort"
+	"strings"
 	"testing"
+	"unicode"
 
 	"github.com/RoaringBitmap/roaring"
+	"github.com/buildbuddy-io/buildbuddy/codesearch/types"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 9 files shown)


codesearch/token/token_test.go
```diff
@@ -67,39 +72,43 @@ func oldTokenize(f io.Reader) []string {
...
-		if err != nil {
-			break
-		}
-		tokens = append(tokens, string(tok.Ngram()))
-	}
+	tokens := tokenizeBuf("this is a string", NewWhitespaceTokenizer())
 	assert.Equal(t, []string{"this", "is", "a", "string"}, tokens)
 }
 
+// func TestBuildAllNgramsMatchesTrigrams(t *testing.T) {
```
siggisim, line 104: Intentional?

siggisim (approved):

codesearch/token/token_test.go
tylerwilliams: yeah gonna fix in a followup -- requires changing the tri tokenizer to stop tokenizing across lines

PR #7108, opened 2024-07-29
Fix OCI runtime stats

bduffany:
* Fix peak memory getting reported as 0
* Port https://github.com/buildbuddy-io/buildbuddy/pull/7021 to OCI runtime so that we account for the runtime process stats


tylerwilliams (approved):
LGTM, Approval

PR #7111, opened 2024-07-29
Reduce OCI capabilities set

bduffany:
In our current podman implementation, we are granting fewer capabilities than what's currently in the hard-coded list.

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #7112, opened 2024-07-29
Clean up how devices are mounted in OCI containers

bduffany:
We don't need to manually mount /dev/null, /dev/zero etc. - this happens automatically.

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #7124, opened 2024-07-30
Add histograms for enqueued task sizes

bduffany:
Wound up just doing it per-probe since `adjustTaskSize` incorporates each node's total size, and the adjusted task size is the source of truth for the task size.

**Related issues**: N/A

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -850,7 +850,7 @@ var (
 		Subsystem: "remote_execution",
 		Name:      "enqueued_task_milli_cpu",
 		Help:      "Milli-CPU prediction of enqueued tasks.",
-		Buckets:   exponentialBucketRange(1, 64_000, 1.2),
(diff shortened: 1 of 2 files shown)


iain-macdonald (approved):

server/metrics/metrics.go
```diff
@@ -845,6 +845,22 @@ var (
 		StatusHumanReadableLabel,
 	})
 
+	RemoteExecutionEnqueuedTaskMilliCPU = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: bbNamespace,
+		Subsystem: "remote_execution",
+		Name:      "enqueued_task_milli_cpu",
+		Help:      "Milli-CPU prediction of enqueued tasks.",
+		Buckets:   exponentialBucketRange(1, 64_000, 1.2),
```
vadimberezniker, line 853: fwiw, our new hardware has 128 cores so we should at least go up to that

vadimberezniker (approved):

PR #7126, opened 2024-07-30
Roll-forward 'Fix container stats baseline for recycled runners (#7080)'

bduffany:
Context: https://buildbuddy-corp.slack.com/archives/C06N9AHA4JX/p1722370297422949

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -213,14 +213,12 @@ type UsageStats struct {
 	baselineCPUPressure, baselineMemoryPressure, baselineIOPressure *repb.PSI
 }
 
-// Reset resets resource usage counters in preparation for a new task, so that
-// the new task's resource usage can be accounted for. It should be called
-// at the beginning of Exec() in the container lifecycle.
-func (s *UsageStats) Reset() {
-	if s.last == nil {
-		// No observations yet; nothing to do.
-		return
-	}
+// SetBaseline records the current stats readings just before executing a new
+// task. It should be called at the beginning of Exec() in the container
+// lifecycle to ensure that the reported stats are only accounting for the
+// container's resource usage during the Exec() call itself.
+func (s *UsageStats) SetBaseline(lifetimeStats *repb.UsageStats) {
+	s.last = lifetimeStats.CloneVT()
 	s.last.MemoryBytes = 0
 	s.baselineCPUNanos = s.last.GetCpuNanos()
 	s.baselineCPUPressure = s.last.GetCpuPressure()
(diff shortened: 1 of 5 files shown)


iain-macdonald (commented):
Looking at this PR, I can't really figure out how the differences between it and 7080 will fix yesterday's bug. Do you have an idea of what went wrong, or if not do you think it'd make sense to pursue a lower risk strategy like just logging the baselines so we can inspect them (or something)?

bduffany:
@iain-macdonald

>  I can't really figure out how the differences between it and 7080 will fix yesterday's bug

The issue is that we should always be setting the baseline equal to the "lifetime stats" of the container, i.e. the raw cgroup stats. This way, when the task is complete, we subtract the last observed cgroup stats from the baseline, to get the usage incurred just for that task.

The issue with the original PR was that we were passing `TaskStats()` to `SetBaseline()`, instead of the raw cgroup stats. So we were setting the baseline to a value that is much lower than it should be, causing the reported CPU to be much higher.

This whole "baseline" approach is ugly / error-prone as we've seen - ideally, we'd just use a new cgroup for each task that we `exec()`. However, this has its own difficulties:
- `podman exec` doesn't have a `--cgroup` flag so it seems like doing this could get complicated. We might have to spawn a wrapper process which forks into its own cgroup inside the container.
- We'd need a way to report usage for persistent workers. It might be possible to move the persistent worker to a new cgroup between executions but that seems somewhat complicated.

> do you think it'd make sense to pursue a lower risk strategy like just logging the baselines so we can inspect them

Yeah, I could try to add some additional logging to test whether this PR would fix the issue that we're currently seeing with the inflated task sizes, causing runaway executors.

The issue that this PR is trying to solve is this: say we execute 2 tasks in the same recycled runner. The timeline looks like this:

1. Task A starts running - we call Create() to create a new runner
2. we call Pause() after task A completes
3. after pausing, we call [Stats()](https://github.com/buildbuddy-io/buildbuddy/blob/af1af37078eabc84816015a8bba58211f52ea57f/enterprise/server/remote_execution/runner/runner.go#L623)
4. Task B starts running. We call `Unpause()` to resume from a recycled runner. If there are BG processes in the runner, usage starts accumulating here again.
5. We call pull_image and fetch action inputs.
6. We record the execution_start_timestamp.
7. `Exec()` - we call `Reset()` here. Internally, this sets the baseline to the observation that we took in step (3).
8. After `Exec()` finished, we record the execution_completed_timestamp.

The problem is steps 4-7. The baseline for task B gets set to the value observed in step 3. So between steps 4 and 6, there is extra usage we are accumulating, which isn't getting attributed to the exec_duration. So the numerator in the `(cpu_usage / exec_duration)` calculation can be inflated, causing the milli-CPU calculation to be much higher than it should be, depending on how much work went untracked. There doesn't even have to be a lot of work that goes untracked - there just has to be a high ratio between the work that was done, and the exec_duration.

I can add some logging to try and get a sense of how much work is going untracked.

bduffany:
@iain-macdonald I sent https://github.com/buildbuddy-io/buildbuddy/pull/7135 as a precursor to this PR. The idea with that PR is to test whether most of the "High computed CPU usage" logs we're seeing (which result in runaway executor queues) are for execution IDs that have a high logged CPU usage between Unpause() and Exec().

iain-macdonald:
Sounds good! Let's see what info we get once that other PR is deployed.

PR #7131, opened 2024-07-31
Fix Firecracker CPU stats

bduffany:
**Related issues**: Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3664

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -106,7 +106,7 @@ const (
 	//
 	// NOTE: this is part of the snapshot cache key, so bumping this version
 	// will make existing cached snapshots unusable.
-	GuestAPIVersion = "11"
+	GuestAPIVersion = "12"
 
 	// How long to wait when dialing the vmexec server inside the VM.
 	vSocketDialTimeout = 60 * time.Second
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sync"
 	"syscall"
 	"testing"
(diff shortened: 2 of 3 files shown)


iain-macdonald (commented):
I have the same concern here as with the other PR. Seems like it's possible there's something up with our measurement/usage of these baselines. Can we try to figure out whether they're returning sane values and how they'll affect our CPU estimates before changing behaviour based on them?

bduffany:
@iain-macdonald

re. the risk of this change, this PR will not affect CPU estimates because we completely bypass the task sizer for Firecracker. We do this because VMs have hard memory limits and we don't to OOM if the same command later needs more memory.

re. correctness of this change - I ran `stress-ng --timeout=${N}s --cpu=1` for various values of `N` -- this command should burn roughly `N*1000` CPU-millis, plus a little bit extra due to process overhead. I set `recycle-runner=true` because this issue only happens when recycling runners.

Before this PR (note that the 10s run consumes 30K milliCPU, which is unexpected):

```
timeout:  1s cpu_millis: 1130.0
timeout:  2s cpu_millis: 2110.0
timeout:  3s cpu_millis: 4210.0
timeout:  4s cpu_millis: 6190.0
timeout:  5s cpu_millis: 9240.0
timeout:  6s cpu_millis: 12250.0
timeout:  7s cpu_millis: 16280.0
timeout:  8s cpu_millis: 20320.0
timeout:  9s cpu_millis: 25350.0
timeout:  10s cpu_millis: 30420.0
```

After this PR:

```
timeout:  1s cpu_millis: 1010.0
timeout:  2s cpu_millis: 2020.0
timeout:  3s cpu_millis: 3040.0
timeout:  4s cpu_millis: 4020.0
timeout:  5s cpu_millis: 5050.0
timeout:  6s cpu_millis: 6040.0
timeout:  7s cpu_millis: 7070.0
timeout:  8s cpu_millis: 8040.0
timeout:  9s cpu_millis: 9050.0
timeout:  10s cpu_millis: 10080.0
```

Here is the full test command I used (requires a small change to `bb execute` to report executionMetadata, which I'll send separately):

`INSTANCE_NAME=$(date +s) ; for timeout in {1..10} ; do bb --verbose=1 execute --remote_instance_name=$INSTANCE_NAME --remote_executor=grpc://localhost:1985 --exec_properties=workload-isolation-type=firecracker --exec_properties=container-image=docker://ecerulm/stress-ng --exec_properties=recycle-runner=true -- stress-ng --cpu=1 --timeout=${timeout}s 2>&1 | grep 'Execution metadata' | perl -pe 's@.*?: @@' | jq -r '.usageStats.cpuNanos' | xargs -I{} python3 -c 'print("timeout: ", "'$timeout's", "cpu_millis:", {} / 1e6)' ; done`

This test along with the unit test that I added gives me much higher confidence that things are working as expected especially compared to what we have today. WDYT?

maggie-lou (approved):
LGTM - thanks for sharing your testing strategy

iain-macdonald (approved):
Thanks for the detailed explanation. This seems good to me, especially given that we don't use these CPU estimates for scheduling decisions.

PR #7138, opened 2024-07-31
child fetch ui

maggie-lou:
- Add an option to exclude invocation metadata from GetInvocation
- Rename field
- Fetch child invocations from DB
- Consolidate two approaches
- Cleanup
- Revert FE changes to make rollout safe
- Fetch child invocations from the db instead of BES events

**Related issues**: N/A


bduffany (commented):
hm github is showing there's nothing to review?

maggie-lou:
> hm github is showing there's nothing to review?

Huh - that was weird. Something weird happened with git history but should be fixed now

app/invocation/child_invocations.tsx
```diff
@@ -1,61 +1,22 @@
...
-        return "queued";
-      }
-    };
-
-    for (const invocation of invocations) {
-      results.push({ invocation, status: getStatus(invocation), durationMillis: this.getDurationMillis(invocation) });
-    }
-
-    if (!results.length) return null;
+    if (!this.props.childInvocations.length) return null;
```
bduffany, line 11: nice simplification here :)

app/invocation/invocation.tsx
```diff
@@ -189,30 +196,88 @@ export default class InvocationComponent extends React.Component<Props, State> {
       this.fetchRunnerExecution();
     }
 
+    try {
+      const inv = await this.fetchInvocation(this.props.invocationId, false /*metadataOnly*/);
```
bduffany, line 200:
```suggestion
      const invocation = await this.fetchInvocation(this.props.invocationId, false /*metadataOnly*/);
```

app/invocation/invocation.tsx
```diff
@@ -189,30 +196,88 @@ export default class InvocationComponent extends React.Component<Props, State> {
...
+    request.lookup.metadataOnly = metadataOnly;
+    const invResp = await rpcService.service.getInvocation(request);
+    console.log(invResp);
+    if (!invResp.invocation || invResp.invocation.length === 0) {
+      throw new BuildBuddyError("NotFound", "Invocation not found.");
+    }
+    return invResp.invocation[0];
+  }
+
+  async fetchChildInvocations(model: InvocationModel): Promise<invocation.Invocation[]> {
```
bduffany, line 231: I know we had discussed this a bit offline but seeing this logic written out I feel like it turned out to be a bit too complex - I feel like it would simplify things a lot if we fetched these on the server when fetching the parent invocation, and returned the child invocations as a field of the Invocation proto.

If we're worried about repeatedly fetching the same DB rows, the server could return a pagination token back to the client (in the GetInvocationResponse), and e.g. skip fetching the first N child invocations that have been finalized. This is how we do log fetching as well - we return a token back to the client that it can then pass to the server to skip fetching any log chunks that have been finalized.

wdyt?

maggie-lou: Hmm I'd probably still prefer the FE approach. The BE approach sounds like the "more correct", scalable option if we needed this functionality in more places, but given that we only need it in one place, it seems preferable to scope the complexity to one small UI element. The primary concern for me is the additional load on the DB

I may be misunderstanding, but I don't think the token approach would work for this use case? Because we want to skip fetching the child invocations while they're *not* finalized (i.e. while they're still running). Then the BE won't have a good mechanism to learn about when they have finally completed so it can fetch the children again to get the final data

maggie-lou: Another option is to add an option to the BE API to fetch children invocations, but the FE manages the state on whether to set that option

Then we can simplify the FE logic that fetches the child invocations, but it would still retain the slightly complex logic on whether we should refetch the children. wdyt about that approach? It's kind of a hybrid between the two

bduffany: > given that we only need it in one place, it seems preferable to scope the complexity to one small UI element

I'm mostly just worried about adding more complexity to invocation.tsx - this component has slowly evolved into a pretty complicated state machine so that it can handle workflows (including "queued" workflows) and regular invocations, and I feel like the extra fetching winds up adding even more extra state that we have to worry about.

In general, doing multiple fetches from the same component in React winds up getting complicated really fast because when state changes in the component we have to think about whether any in-flight fetches need to be canceled.

> I don't think the token approach would work for this use case? Because we want to skip fetching the child invocations while they're not finalized (i.e. while they're still running). Then the BE won't have a good mechanism to learn about when they have finally completed

The ChildInvocationConfigured events in the BES are always returned in the same order, so we can think of them as an ordered list. The server could just return an index of the first child invocation that may still be in progress and need to be fetched. It can pass that back to the client and the client can blindly pass it back to the server on each GetInvocation request. This way all of the bookkeeping is done on the server and the only state the client needs to care about is the `page_token`.

Example:
- Client calls GetInvocation for the outer workflow invocation, initially setting `page_token = 0`
- Server loads workflow invocation metadata and all BES events like it does today
- Server sees that there are 5 child invocations, from BES events
  - Since token is 0, it fetches `children[0:]` from the DB (i.e. all of them)
  - Let's say 4 out of 5 are complete, and the 5th one is in progress. It returns `page_token = 4` back to the client
- Client calls GetInvocation again, to refresh, setting `page_token = 4` (just passthrough from the server)
- Server loads workflow invocation metadata and all BES events like it does today
- Server sees that there are still 5 child invocations, from BES events
  - Since `page_token` is 4, server fetches `children[4:]` from the DB
  - Let's say the 5th invocation is complete this time. The server sends back `page_token = 5` to the client.
- On the next fetch, the client sends `page_token = 5`, then on the next GetInvocation request, the server won't need to fetch any child invocations, assuming there are no new child invocations in the BES stream.

maggie-lou: @bduffany Refactored to implement the second approach I mentioned - let me know what you think. [This PR ](https://github.com/buildbuddy-io/buildbuddy/pull/7181) should come first. It should address your concern about doing multiple fetches from the FE

The token approach is not ideal because we * don't * want to keep fetching the children while they're running. We only update the child invocation card once at the start and once at the end. WIth the token approach as you described, if there's a 3 hour child invocation, we'll continue to fetch it once every 3 sec for 3 hours, even though we don't need any additional info from it

bduffany (approved):
I still think that fetching just the most recent child invocation (using a page token) would be simpler while also being fine in terms of DB load (we are already polling the DB for the outer invocation, and reads from the invocation table are negligible compared to e.g. the heavy Execution write load that we have today). If we want to stop polling the DB, then we have other ways we can do that (e.g. redis caching or pubsub notifications) which would work at a lower layer and would reduce load across the board (i.e. top-level invocation polling would get more efficient too) instead of just for the child invocation polling.

But I think it's in a much cleaner state in the latest iteration, and would be fine with merging as-is :+1:

PR #7143, opened 2024-08-01
More automated CLI login flow

bduffany:
Running `bb login` now opens a web auth flow that is fully automated in the "happy path" (already logged in to bb, only a member of one org, org has personal API keys enabled).

The CLI starts a local server which then redirects the user to BB. BB then redirects back to the local server, with a "token" param set to a newly created personal API key (in the happy path).

Edge cases:
- If not logged in to the UI, we show the login page first, then redirect to the CLI login page.
- If the user is a member of 0 orgs, we take them through the "Create org" flow, then redirect them back to the CLI login page.
- If the user is a member of more than one org, we show a group picker (screenshot below). Customers can pass `--org` to the `bb login` command, which pre-selects the org and avoids showing this picker.
- If the user is not a member of the group specified by `--org` then we direct them to the org join page for that org. 

Since there are various edge cases, this PR includes a few webdriver tests to make sure we don't break this flow accidentally.

The landing page (in the app) looks like this:

image

After selecting an org (or if the user is only a member of one org), the app redirects back to the CLI, attempts to store the token, and then if that succeeds, redirects back to the app with this page:

image

(If writing the token fails, it shows an error page instead)

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3665


siggisim:
This looks awesome - could we have the cli redirect to a success page rather than render it on it's own? That way we can do styling / etc more easily.

bduffany:
> could we have the cli redirect to a success page rather than render it on it's own? 

Done (and updated the screenshot)

siggisim (approved):
Love this!

PR #7148, opened 2024-08-02
Add comments to iptables rules created by the executor

bduffany:
Sometimes I want to delete all `iptables` rules created by the executor. This is a lot easier if each rule has a common string that I can grep for. This PR adds "buildbuddy-executor" as a comment to each rule, using the xt_comment module (if it's loaded).

**Related issues**: N/A

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -53,6 +53,10 @@ const (
 
 	// CIDR matching all container networks on the host.
 	containerNetworkingCIDR = "192.168.0.0/16"
+
+	// Comment added to each iptables rule to identify it as being created by
+	// BuildBuddy.
+	iptablesComment = "buildbuddy-executor"
 )
 
 // runCommand runs the provided command, prepending sudo if the calling user is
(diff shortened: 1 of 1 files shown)


vadimberezniker (commented):
Would it be simpler to put our custom rules into a separate chain? Then you could easily nuke all of the rules in the chain.

server/util/networking/networking.go
```diff
@@ -280,6 +284,22 @@ func (a *HostNetAllocator) unlock(netIdx int) {
 	a.inUse[netIdx] = false
 }
 
+// Returns whether the xt_comment kernel module is available.
+var xtCommentLoaded = sync.OnceValue(func() bool {
+	b, err := exec.Command("lsmod").CombinedOutput()
+	if err != nil {
+		log.Warningf("lsmod failed: %s", err)
```
tylerwilliams, line 291: is this going to be noisy on mac, or is this code not included?

tylerwilliams (commented):
See comments

server/util/networking/networking.go
bduffany: this code should only be included on linux - going to close this PR though because I think Vadim's suggestion of using a chain might be a better way of doing it since it doesn't depend on this xt_comment module being available

PR #7149, opened 2024-08-02
Fix a bug in digest.Diff() by making it use value equality instead of pointer equality.

iain-macdonald:
The current implementation of `digest.Diff` uses pointer equality for testing digest equality. This means that it will incorrectly declare two digests to be different if they are pointers to different versions of the same digest, rather than checking their values to determine they are the same.

This is only used in the migration cache and has likely resulted in over-migrating data during the double read phase. From reading the code, it appears the new, correct behavior is desired as it's reading digests from two different backends, so the current implementation will always declare them all to be different, as they'll be pointers into different allocated protos in memory.

I added a new test that demonstrates the bug.

**Related issues**: N/A

### DIFF
--- a/server/remote_cache/digest/digest.go
+++ b/server/remote_cache/digest/digest.go
@@ -583,24 +583,24 @@ func Diff(s1 []*repb.Digest, s2 []*repb.Digest) (missingFromS1 []*repb.Digest, m
 	missingFromS1 = make([]*repb.Digest, 0)
 	missingFromS2 = make([]*repb.Digest, 0)
 
-	s1Set := make(map[Key]*repb.Digest, len(s1))
+	s1Set := make(map[Key]struct{}, len(s1))
 	for _, d := range s1 {
-		s1Set[NewKey(d)] = d
+		s1Set[NewKey(d)] = struct{}{}
 	}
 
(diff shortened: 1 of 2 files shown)


server/remote_cache/digest/digest.go
```diff
@@ -583,22 +583,23 @@ func Diff(s1 []*repb.Digest, s2 []*repb.Digest) (missingFromS1 []*repb.Digest, m
 	missingFromS1 = make([]*repb.Digest, 0)
 	missingFromS2 = make([]*repb.Digest, 0)
 
-	s1Set := make(map[*repb.Digest]struct{}, len(s1))
+	s1Set := make(map[Key]*repb.Digest, len(s1))
```
vadimberezniker, line 586: It seems redundant to store both a Key and a *Digest which contain the same data. What do you think about calling key.ToDigest() when populating the slices at the end?

vadimberezniker (approved):

server/remote_cache/digest/digest.go
iain-macdonald: Sure, it's either redundant map values or redundant conversion back into digest protos. I guessed storing pointer values in the map would be slightly more performant than re-allocating new Keys for each set of differences, but probably either way is fine. Changed.

PR #7150, opened 2024-08-02
Don't waste memory

tylerwilliams:
Decode runes directly into the interned slice of runes.

### DIFF
--- a/codesearch/token/token.go
+++ b/codesearch/token/token.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"unicode"
 	"unicode/utf8"
+	"unsafe"
 
 	"github.com/buildbuddy-io/buildbuddy/codesearch/sparse"
 	"github.com/buildbuddy-io/buildbuddy/codesearch/types"
@@ -330,9 +331,10 @@ func NewSparseNgramTokenizer(mods ...Option) *SparseNgramTokenizer {
 
 		hasher:      fnv.New32(),
 		alreadySeen: sparse.NewSet(1<<32 - 1),
-		ngrams:      make([]string, 0),
-		s:           make([]rune, 0),
-		st:          make([]hashAndPosition, 0),
+
+		ngrams: make([]string, 0),
+		s:      make([]rune, bufio.MaxScanTokenSize),
+		st:     make([]hashAndPosition, 0),
 
 		stringTemp: make([]byte, utf8.UTFMax*opts.MaxNgramLength),
 	}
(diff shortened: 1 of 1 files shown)


codesearch/token/token.go
```diff
@@ -351,7 +353,8 @@ func (tt *SparseNgramTokenizer) Type() types.FieldType {
 }
 
 func (tt *SparseNgramTokenizer) Ngram() []byte {
-	return []byte(tt.ngrams[len(tt.ngrams)-1])
+	gram := tt.ngrams[len(tt.ngrams)-1]
+	return unsafe.Slice(unsafe.StringData(gram), len(gram))
```
jdhollen, line 357: Just confirming that I checked documentation and agree this is all operating on byte counts and not runes

codesearch/token/token.go
```diff
@@ -362,11 +365,20 @@ func (tt *SparseNgramTokenizer) refillLine() error {
...
-		loweredLine := strings.ToLower(tt.scanner.Text())
-		if len(loweredLine) == 0 {
+		buf := tt.scanner.Bytes()
+		lineLength := len(buf)
+
+		if lineLength == 0 {
 			continue
 		}
-		tt.s = []rune(loweredLine)
+		tt.s = tt.s[:lineLength]
```
jdhollen, line 374: This is weird, right? You're indexing into runes with a byte offset? I assume I'm misreading but figure it's faster to ask.

codesearch/token/token.go
```diff
@@ -330,9 +331,10 @@ func NewSparseNgramTokenizer(mods ...Option) *SparseNgramTokenizer {
 
 		hasher:      fnv.New32(),
 		alreadySeen: sparse.NewSet(1<<32 - 1),
-		ngrams:      make([]string, 0),
-		s:           make([]rune, 0),
-		st:          make([]hashAndPosition, 0),
+
+		ngrams: make([]string, 0),
```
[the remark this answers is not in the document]
tylerwilliams: it's because there's a newline between st and ngrams. I can remove so they all line up

codesearch/token/token.go
tylerwilliams: ayyyy

codesearch/token/token.go
tylerwilliams: it is weird! I actually had to google this. The gist of it is that `tt.s` is a rune buffer with sufficient capacity to hold any line we'd read. We scan a line of bytes from the scanner and ensure the length of `tt.s` is at least that long so that we have room to put all decoded runes in there (rune count being <= byte count, depending on encoding).

Once we've decoded all runes and know the exact length, we set `tt.s`'s length to the number of runes (capacity is unchanged, and it's the same memory).

jdhollen (approved):

PR #7153, opened 2024-08-02
Prevent ui from crashing when our proto is older

sluongng:


sluongng:
example crash: https://sluongng.buildbuddy.io/invocation/e1a61abc-390e-444d-a4cf-8aa9737ad876

app/invocation/invocation_error_card.tsx
```diff
@@ -197,6 +197,10 @@ function formatFailureDescription(failureDetail: failure_details.IFailureDetail)
   }
   // TODO: handle other FailureDetail fields
 
+  // Our proto could be out-of-date compare to Bazel and cause code to be undefined
+  // when the failureDetail code is new.
+  // Make sure that code is always a string.
+  code = code ?? "";
```
jdhollen, line 203: rather than blanking it, maybe we could just switch to the raw error code above? e.g.,

```
code = failure_details.TargetPatterns.Code[failureDetail.targetPatterns.code] ?? failureDetail.targetPatterns.code;
```

jdhollen: or, well, that plus the string cast

sluongng: Ah yeah, that would be much nicer.

Applied the fix.

jdhollen (approved):

PR #7154, opened 2024-08-02
cli/print: remove size limit when parsing compact exec log proto

sluongng:


cli/printlog/compact/compact.go
```diff
@@ -55,7 +55,7 @@ func PrintCompactExecLog(path string, raw bool, sort bool) error {
 	if raw {
 		entry := &spb.ExecLogEntry{}
 		for {
-			err := protodelim.UnmarshalFrom(reader, entry)
+			err := protodelim.UnmarshalOptions{MaxSize: -1}.UnmarshalFrom(reader, entry)
```
fmeum, line 58: As far as I can tell this is only relevant for [this check](https://github.com/protocolbuffers/protobuf-go/blob/f09e239720f1bcd654d565107b1d4feea184d552/encoding/protodelim/protodelim.go#L111-L122), it's not used for buffer presizing. Is there a downside to keeping this unbounded? I'm not certain that `ExecLogEntry` couldn't grow beyond 4 MB in pathological cases (e.g. when a rule doesn't use param files).

sluongng: I definitely have run into this error before, which is why I had this PR open 🤔

It has been a while ago though, so I no longer have a copy of that log. Since the `print` command is mostly used for debug/troubleshooting purposes, I think adding this option should be fine?

fmeum: Oh wait, you are removing the size *limit*, not removing the option that removes the limit. Nvm, ship it

fmeum (approved):

PR #7168, opened 2024-08-06
github/workflows: exclude flaky test

sluongng:
Exclude
//enterprise/server/test/integration/remote_execution:remote_execution_test
from running locally on ARM64.

This constantly fail on ARM64 and create false signals.

### DIFF
--- a/.github/workflows/test-executor-linux-arm64.yaml
+++ b/.github/workflows/test-executor-linux-arm64.yaml
@@ -9,7 +9,7 @@ on:
       - "*"
 
 jobs:
-  build:
+  "Local-Exec ARM64 Test":
     runs-on: ubuntu-22.04-16cpu-arm64
     if: "!contains(github.event.head_commit.message, 'ci skip')"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-executor-linux-arm64.yaml
```diff
@@ -9,7 +9,7 @@ on:
       - "*"
 
 jobs:
-  build:
+  "Local-Exec ARM64 Test":
```
bduffany, line 12: nit: this will show up in the GitHub UI like `Test linux/arm64 executor / Local-Exec ARM64 Test` - the arm64 part is redundant and the local-exec part is not that important IMO since the BES UI will show "remote execution off" in the header

```suggestion
  "test":
```

bduffany (approved):

.github/workflows/test-executor-linux-arm64.yaml
```diff
@@ -46,7 +46,8 @@ jobs:
           )
           bazel test "${BUILD_FLAGS[@]}" --test_tag_filters=-performance,-docker \
             //enterprise/server/remote_execution/... \
```
bduffany, line 48: I think `--` is needed to fix the build failure

```suggestion
            -- \
            //enterprise/server/remote_execution/... \
```

bduffany (approved):

PR #7170, opened 2024-08-06
fix up loading state for flaky test link

jdhollen:
**Related issues**: N/A


siggisim (approved):
lgtm, i also noticed there is a little extra padding in between the icon and the "Recently flaky" text.

i think if you remove the gap on .flakey-target-chip it looks a bit better better

PR #7174, opened 2024-08-06
fix icon margin on flaky target chip on targets tab.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/app/root/root.css
+++ b/app/root/root.css
@@ -1258,10 +1258,11 @@ code .comment {
 .flaky-target-chip {
   display: flex;
   margin-left: 8px;
+  gap: 8px;
 }
 
 .flaky-target-chip .icon {
-  margin-right: 8px;
+  margin-right: 0;
 }
 
 .test-subtitle {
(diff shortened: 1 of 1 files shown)


app/root/root.css
```diff
@@ -1258,7 +1258,10 @@ code .comment {
 .flaky-target-chip {
   display: flex;
   margin-left: 8px;
-  gap: 8px;
+}
+
+.flaky-target-chip .icon {
+  margin-right: 8px;
```
siggisim, line 1264: Thoughts on making this 0? Then it's just controlled by the gap, which is I think what's used in other contexts and looks nice.

siggisim (approved):

app/root/root.css
jdhollen: like this?

siggisim (approved):

app/root/root.css
siggisim: yup

PR #7179, opened 2024-08-07
Update executor image, but do not update soci-snapshotter go dep

tempoz:
- **updated Dockerfile**
- **move dockerfile run steps to container layers**
- **update versions for container layers**

**Related issues**: N/A


enterprise/server/cmd/executor/BUILD
```diff
@@ -80,6 +81,120 @@ go_binary(
...
+        "        \"credHelpers\": {",
+        "                \"asia.gcr.io\": \"gcr\",",
+        "                \"eu.gcr.io\": \"gcr\",",
+        "                \"gcr.io\": \"gcr\",",
+        "                \"marketplace.gcr.io\": \"gcr\",",
+        "                \"us.gcr.io\": \"gcr\"",
+        "        }",
+        "}",
+    ],
+)
```
bduffany, line 130: nit: this could be simplified by storing the JSON as a plain file under //dockerfiles/executor_image/data/docker/config.json

deps.bzl
```diff
@@ -5,9 +5,9 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file"
 # When changing this version, a new release of podman-static may be needed.
 # See dockerfiles/executor_image/README.md for instructions.
 # The checksums below will also need to be updated.
-PODMAN_VERSION = "v4.9.4"
-PODMAN_STATIC_SHA256_AMD64 = "ac14152d2ef5cb25abbd615893cd56355f951f075a7ae663f0b2c1bc5d3fb77e"
-PODMAN_STATIC_SHA256_ARM64 = "72ab4a9c6ae9d5d00200070cb43b0921117898925a284bfacbfb3873aa82e598"
+PODMAN_VERSION = "v5.1.1-1"
+PODMAN_STATIC_SHA256_AMD64 = "a172ed5383234e6b44a0de8d908bb0aa1baff8beeb56214898c3a7930a3c903d"
+PODMAN_STATIC_SHA256_ARM64 = "3f932655d812b26525063f1b13a0fbab918ea536aee3311e69805c075460ad5f"
```
bduffany, line 10: nit, split the podman upgrade into a separate PR? since it's a major version upgrade it's more risky and would be nice to have as a separate commit that we can bisect.

deps.bzl
```diff
@@ -6834,20 +6834,15 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
...
-        sha256 = "8b420b9e94433f7fe74f71a26a34ee2c0bed6535bdb4c2176a91611e4054d0d0",
-        urls = ["https://storage.googleapis.com/buildbuddy-tools/binaries/soci-snapshotter/soci-store-v0.0.15-linux-amd64-race"],
+    http_archive(
+        name = "com_github_buildbuddy_io_soci_snapshotter-static-linux-amd64",
+        build_file_content = "\n".join([
+            'package(default_visibility = ["//visibility:public"])',
+            'filegroup(name = "soci-store.bin", srcs = ["soci-store"])',
+            'filegroup(name = "soci-store-race.bin", srcs = ["soci-store-race"])',
+        ]),
+        urls = ["https://github.com/buildbuddy-io/soci-snapshotter/releases/download/release%2Fbuildbuddy-v0.7.0/soci-snapshotter-0.7.0-linux-amd64-static.tar.gz"],
```
bduffany, line 6844: wondering why the tag changed from `v0.0.15` => `v0.7.0` - do our tags just match the upstream but with our patches applied?

deps.bzl
```diff
@@ -6991,3 +6986,104 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
...
+    )
+    http_file(
+        name = "com_github_opencontainers_runc_runc-linux-arm64",
+        urls = ["https://github.com/opencontainers/runc/releases/download/v1.2.0-rc.2/runc.arm64"],
+        sha256 = "3741f07946c54a23ee443de84506c6e397d0d350f1f16f81ae4dd2c7d39d2651",
+        downloaded_file_path = "runc",
+        executable = True,
+    )
+
+    http_archive(
```
bduffany, line 7062: would be nice to split the containerd changes into a separate PR too, if it's not too difficult

dockerfiles/executor_image/Dockerfile
```diff
@@ -4,50 +4,36 @@
...
-    rm -rf /var/lib/apt/lists/* && apt-get clean
-
-# Install skopeo and umoci which we use to unpack OCI images when we're not using docker.
-# Install iproute2 ("ip" command) to configure networking on host.
-RUN apt-get update && \
-    apt-get install -y skopeo umoci iproute2 amazon-ecr-credential-helper && \
+    apt-get update && apt-get install -y \
+    docker-ce=${DOCKER_VERSION} \
+    docker-ce-cli=${DOCKER_VERSION} \
+    containerd.io=${CONTAINERD_DEB_VERSION} \
```
bduffany, line 33: It looks like we're overwriting certain binaries via image layers - I think that will increase the image size significantly? might make sense to either overwrite those files in this step (instead of adding an overwrite layer) or just delete the files in this step (and sprinkle some comments that these deletions should be kept in sync with the container_layer)

enterprise/server/cmd/executor/BUILD
```diff
@@ -80,6 +81,120 @@ go_binary(
...
+        "@platforms//cpu:x86_64": [
+            "@com_github_googlecloudplatform_docker-credential-gcr-linux-amd64//:docker-credential-gcr.bin",
+        ],
+        "@platforms//cpu:arm64": [
+            "@com_github_googlecloudplatform_docker-credential-gcr-linux-arm64//:docker-credential-gcr.bin",
+        ],
+        "//conditions:default": [],
+    }),
+    tags = ["manual"],
+)
```
bduffany, line 112: The `tini` and `docker-credential-gcr` changes feel pretty low-risk and self-contained, would be nice to move those into a separate PR

PR #7181, opened 2024-08-07
[ci_runner] Write parent_invocation_id to the db

maggie-lou:
**Related issues**: N/A


proto/invocation.proto
```diff
@@ -168,9 +177,8 @@ message InvocationLookup {
   // The invocation_id: a UUID generated by the bazel run.
   string invocation_id = 1;
 
-  // If true, only return build metadata stored in the db. Excludes data like
-  // the scorecard, BES events, structured command line, and console buffer.
-  bool metadata_only = 2;
+  // Whether to inline child invocations in the invocation response.
```
bduffany, line 180: nit: clarify whether only the  metadata is fetched vs metadata + BES events?

server/tables/tables.go
```diff
@@ -171,6 +171,8 @@ type Invocation struct {
 	RemoteExecutionEnabled bool
 
 	Tags string
+
+	ParentInvocationID string
```
bduffany, line 175: should we store this as a `UUID` `[]byte` instead, similar to `InvocationUUID` above? (IIRC @luluz66 has more context for these UUID fields)

server/util/clickhouse/schema/schema.go
```diff
@@ -115,6 +115,7 @@ type Invocation struct {
 	UploadLocalResultsEnabled         bool
 	RemoteExecutionEnabled            bool
 	Tags                              []string `gorm:"type:Array(String);"`
+	ParentInvocationID                string
```
bduffany, line 118: ditto here, should this be a `UUID []byte`?

server/backends/invocationdb/invocationdb.go
```diff
@@ -181,6 +181,12 @@ func (d *InvocationDB) LookupInvocation(ctx context.Context, invocationID string
 	return ti, nil
 }
 
+func (d *InvocationDB) LookupChildInvocations(ctx context.Context, parentInvocationID string) ([]*tables.Invocation, error) {
+	rq := d.h.NewQuery(ctx, "invocationdb_get_invocation").Raw(
```
bduffany, line 185:
```suggestion
	rq := d.h.NewQuery(ctx, "invocationdb_get_child_invocations").Raw(
```

server/tables/tables.go
bduffany, line 175: this field needs a non-unique index, otherwise the child invocations query will require a full table scan

luluz66: InvocationUUID was added initially for the targetStatus table in mysql. It was a foreign key for that table. We use Binary(16) to save space.

bduffany: OK, I think these are fine as-is then. Thanks for explaining!

maggie-lou: Ah good call - added

bduffany (approved):
sweet!

server/util/clickhouse/schema/schema.go
luluz66, line 118: You want to add this field to "ToInvocationFromPrimaryDB"

server/tables/tables.go
maggie-lou: @luluz66 mentioned that we might want to store parent invocation ID as a UUID in clickhouse, so that its format matches InvocationUUID if we every want to join on the field or compare them

@luluz66 @bduffany Could you take another look and make sure this looks right?

server/util/clickhouse/schema/schema.go
```diff
@@ -125,11 +127,16 @@ func (i *Invocation) ExcludedFields() []string {
...
 }
 
 func (i *Invocation) AdditionalFields() []string {
-	return []string{}
+	return []string{
+		// Invocation IDs are stored as UUIDs (without dashes), so store the parent invocation ID
+		// as a UUID, so it can be easily joined.
+		"ParentInvocationUUID",
+	}
 }
```
bduffany, line 140: why is ParentInvocationUUID is in this list but not InvocationUUID? (sorry I'm not super familiar with the clickhouse schema)

bduffany (approved):
LGTM, just a question

server/util/clickhouse/schema/schema.go
maggie-lou: We save InvocationUUID explicitly in the mysql db (https://github.com/buildbuddy-io/buildbuddy/blob/07185758c263aae0210fc63b4149af8172f698f6/server/tables/tables.go#L1037). @luluz66 there aren't any reasons why I'd have to store my new UUID in mysql right? We can just convert it to UUID before storing in click house?

luluz66 (approved):

PR #7183, opened 2024-08-07
Use container_layer instead of Dockerfile when possible in the executor image

tempoz:
- **move dockerfile run steps to container layers**
- **docker_credential_gcr config now lives in a file**

**Related issues**: N/A


dockerfiles/executor_image/Dockerfile
```diff
@@ -39,13 +32,17 @@ RUN DOCKER_VERSION="5:26.1.4-1~debian.12~bookworm" && \
...
     rm \
-    # keep these in sync with the binaries we are manually adding
+    # keep these in sync with the binaries we are manually
+    # adding in enterprise/server/cmd/executor/BUILD.
+    #
     # rootlesskit
-    /usr/bin/rootlesskit
-    /usr/bin/rootlesskit-docker-proxy
+    /usr/bin/rootlesskit \
+    /usr/bin/rootlesskit-docker-proxy \
```
bduffany, line 45: does the fix here need to be backported to the branch that this one is based on?

bduffany (approved):

dockerfiles/executor_image/Dockerfile
tempoz: yup, did it.

PR #7189, opened 2024-08-08
invocation_exec_log_card: fix invalid href local runner entries

sluongng:
While working on #7166, I discovered that clicking on `local` actions in
the execution log card would crash our UI.

This is because local spawn entry does not have `digest` and thus,
render the entries in the card with `href=/invocation/<uuid>$#action`,
and clicking on it would cause our browser to crash.

Don't render the invalid URL here.

### DIFF
--- a/app/invocation/invocation_exec_log_card.tsx
+++ b/app/invocation/invocation_exec_log_card.tsx
@@ -169,13 +169,13 @@ export default class InvocationExecLogCardComponent extends React.Component<Prop
   }
 
   getActionPageLink(entry: tools.protos.ExecLogEntry) {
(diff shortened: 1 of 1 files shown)


sluongng:
Reproduce steps:

Go to https://app.buildbuddy.io/invocation/267b1271-7215-40fc-acc9-6b9d1ea8f1f8#execution
filter by "Runner == local" and click on any action. This will result in a UI crash.

siggisim (approved):
I'd consider inverting this for simplicity, i.e.

```
if (!entry.spawn?.digest) {
   return undefined;
}

... original code
```

PR #7199, opened 2024-08-08
Migrate to new singleflight package

bduffany:
The new package respects context cancellation, which fixes 3 classes of issues:
- A context cancellation from the original caller can affect other callers waiting on that operation to complete, resulting in errors which might be non-retryable and result in a bad client experience.
- A cancelled request from a deduped caller must still wait for the original caller to complete its request, even if the client cancelled their request or their deadline has exceeded.
- A possible workaround for the above 2 issues was to use `DoChan` and use a background context for the deduped operation, but this results in a third type of issue where potentially expensive operations are kept running in the background even if no one is waiting for them to complete.

**Related issues**: N/A


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -820,14 +820,8 @@ func (ff *BatchFileFetcher) bytestreamReadFiles(ctx context.Context, instanceNam
 		return fp0, nil
 	})
 	if err != nil {
-		// Ensure that an unavailable error is always returned so that
-		// the download can be retried if we hit a transient error.
-		if !status.IsUnavailableError(err) {
-			err = status.UnavailableError(err.Error())
-		}
```
bduffany, line 827: This code was working around the context cancellation issue in the x/sync package - if the original caller cancels their context then it would cause other actions to fail with a potentially non-retryable error. With the new package, this logic can be removed.

vadimberezniker (approved):

bduffany:
(need 1 more LGTM for auth code)

enterprise/server/remote_execution/dirtools/dirtools.go
tylerwilliams: Kinda slow still but double checking -- you're sure it's OK to remove this?

tylerwilliams (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
bduffany: yup, this was just working around the limitations of the old package. The singleflight package has unit tests that confirm this but I also sanity checked it here: https://go.dev/play/p/ksdZXjzd1sI

PR #7201, opened 2024-08-09
invocation_action_card: handle exception parsing action digest

sluongng:
If user try to navigate to a url such as
`http://app.buildbuddy.io/invocation/<uuid>#action` today, their UI
would crash and unable to recover.

Fix this by returning an alert on the screen instead.
With this, users still see the invocation UI and can navigate away.

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -267,7 +267,7 @@ export default class InvocationActionCardComponent extends React.Component<Props
       // execution table lookup.
       const executeResponseDigest = parseActionDigest(executeResponseDigestParam);
       if (!executeResponseDigest) {
-        alert_service.error("Missing execute response digest in URL");
+        alert_service.error("Invalid execute response digest in URL");
         return;
       }
       this.executeResponseRPC = this.fetchExecuteResponseByDigest(executeResponseDigest);
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -108,6 +108,10 @@ export default class InvocationActionCardComponent extends React.Component<Props
       return;
     }
     const digest = parseActionDigest(digestParam);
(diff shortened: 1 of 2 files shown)


app/invocation/invocation_action_card.tsx
```diff
@@ -217,6 +221,10 @@ export default class InvocationActionCardComponent extends React.Component<Props
       return;
     }
     const digest = parseActionDigest(digestParam);
+    if (!digest) {
```
bduffany, line 224: nit: if you pass `digestParam ?? ""` to this func then you can remove the redundant check on line 219

bduffany (approved):
This change seems fine as a defensive check. Out of curiosity though, are we reaching these invalid URLs from some link in the app? (And if so, should we avoid rendering it as a link in the case where the action digest is missing?)

sluongng:
yeah we were hitting it before https://github.com/buildbuddy-io/buildbuddy/pull/7189

Now that is patched, this PR should help future-proof this class of issue.

app/invocation/invocation_action_card.tsx
sluongng: Done

app/invocation/invocation_action_card.tsx
```diff
@@ -262,9 +266,17 @@ export default class InvocationActionCardComponent extends React.Component<Props
       // If we have the executeResponseDigest in the URL, we can skip the
       // execution table lookup.
       const executeResponseDigest = parseActionDigest(executeResponseDigestParam);
+      if (!executeResponseDigest) {
+        alert_service.error("Missing execute response digest in URL");
```
bduffany, line 270: sorry I missed this earlier. It should say "Invalid", not "Missing", because we checked whether it's non-empty above.
```suggestion
        alert_service.error("Invalid execute response digest in URL");
```

bduffany (approved):

PR #7202, opened 2024-08-09
invocation_exec_log_card: fix sort direction radio group

sluongng:
This was using the same name as the radio buttons for the remote
executions card.
Changing sort direction when both 4 radio buttons are available cause
the checked status to not render correctly.

Rename the radio inputs of exec log.
Also tighten the onChange handler to not rely on the input name to
determine the state.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3587

### DIFF
--- a/app/invocation/invocation_exec_log_card.tsx
+++ b/app/invocation/invocation_exec_log_card.tsx
@@ -268,7 +268,7 @@ export default class InvocationExecLogCardComponent extends React.Component<Prop
                       checked={this.state.direction == "desc"}
                       onChange={this.handleSortDirectionChange.bind(this)}
                       value="desc"
-                      name="exec-log-direction"
+                      name="execLogDirection"
                       type="radio"
                     />
                     <label htmlFor="direction-desc">Desc</label>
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_exec_log_card.tsx
```diff
@@ -257,9 +255,9 @@ export default class InvocationExecLogCardComponent extends React.Component<Prop
                     <input
                       id="direction-asc"
                       checked={this.state.direction == "asc"}
-                      onChange={this.handleInputChange.bind(this)}
+                      onChange={this.handleSortDirectionChange.bind(this)}
                       value="asc"
-                      name="direction"
+                      name="exec-log-direction"
```
bduffany, line 260: in other "name" fields in the past I think we've used `camelCase`

```suggestion
                      name="execLogDirection"
```

app/invocation/invocation_exec_log_card.tsx
```diff
@@ -268,9 +266,9 @@ export default class InvocationExecLogCardComponent extends React.Component<Prop
                     <input
                       id="direction-desc"
                       checked={this.state.direction == "desc"}
-                      onChange={this.handleInputChange.bind(this)}
+                      onChange={this.handleSortDirectionChange.bind(this)}
                       value="desc"
-                      name="direction"
+                      name="exec-log-direction"
```
bduffany, line 271:
```suggestion
                      name="execLogDirection"
```

bduffany (approved):

PR #7206, opened 2024-08-09
Fix `googleapis` module extension usage

fmeum:
This fixes running `bazel mod tidy --enable_bzlmod` in the repo.

Since we do not reference any repos generated by the `switched_rules` extension (otherwise CI would have caught the mistake), we can also drop the `use_repo`.

**Related PRs**: #7205

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -19,7 +19,6 @@ switched_rules.use_languages(
     go = True,
     grpc = True,
 )
-use_repo(switched_rules, "com_google_googleapis_imports")
 
 # Needed for com_google_protobuf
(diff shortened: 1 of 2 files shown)


tylerwilliams (approved):

vadimberezniker (approved):

bduffany (approved):

bduffany:
@fmeum If you want to increase the disk usage to fix the test failure in this PR, I think that'd be fine:

```
actions:
  - name: Test with BzlMod
    # ...
    resource_requests:
      disk: 27GB
```

PR #7217, opened 2024-08-12
ci_runner: avoid url.*.insteadOf clobbering

sluongng:
Running
`git config --global url.https://github.com/.insteadOf <original>`
multiple times will replace the previous config value with the new value
of each call.

Instead of doing that, run the very first attempt with `--replace-all`
to wipe all existing values with our new value.  For subsequent runs,
use `--add` to append the new values to existing config.

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -2471,13 +2471,15 @@ func configureGlobalCredentialHelper(ctx context.Context) error {
 // necessarily have any SSH private keys available that we can use to clone
 // repos over SSH.
 func configureGlobalURLRewrites(ctx context.Context) error {
+	flag := "--replace-all"
 	for original, replacement := range map[string]string{
 		"ssh://git@github.com:": "https://github.com/",
 		"git@github.com:":       "https://github.com/",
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2471,13 +2471,15 @@ func configureGlobalCredentialHelper(ctx context.Context) error {
 // necessarily have any SSH private keys available that we can use to clone
 // repos over SSH.
 func configureGlobalURLRewrites(ctx context.Context) error {
+	flag := "--replace-all"
```
bduffany, line 2474:
```suggestion
		// Discard any existing rewrites initially, then append.
		flag := "--replace-all"
```

bduffany (approved):

enterprise/server/cmd/ci_runner/main.go
sluongng: hmm the indentation looks weird here on Github UI.
Let's see if it lands correctly

sluongng: It was a little off. Pulled + reformat + pushed

PR #7219, opened 2024-08-12
Update `soci-snapshotter` version in executor image

tempoz:
Fixed the panic; soci changed where they added default values from where they were used to when they were parsed, so corrected to the new config loading function:

https://github.com/buildbuddy-io/soci-snapshotter/commit/15935fa304085232215608325765d80c3d4e0a5a

**Related issues**: N/A

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -6834,20 +6834,15 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
         urls = ["https://github.com/containerd/stargz-snapshotter/releases/download/v0.11.4/stargz-snapshotter-v0.11.4-linux-amd64.tar.gz"],
     )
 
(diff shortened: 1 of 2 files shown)


bduffany:
Iain is in this week so I added him as the reviewer

iain-macdonald (commented):
It looks like you renamed the executable file. Do you have to update `--executor.podman.soci_store_binary` as well so the store can be referenced and started by the executor?

https://github.com/buildbuddy-io/buildbuddy/blob/6d72df4de5bdea65247847025d1cc4e00db99a75/enterprise/server/remote_execution/soci_store/soci_store_linux.go#L41

iain-macdonald (approved):

PR #7221, opened 2024-08-12
Add a simpler way to prevent recycling

bduffany:
Allow preventing runner recycling by creating a file `.BUILDBUDDY_DO_NOT_RECYCLE` in the workspace root.

The current approach of sending SIGKILL to the runner process is not as explicit (we happen to skip recycling on SIGKILL, but it's not very obvious), is a little confusing because SIGKILL usually indicates an OOM error, is not well-tested, and is not as flexible because it doesn't work in cases where the action should succeed but just skip recycling.

**Related issues**: N/A


tylerwilliams:
I feel like the file seems like a cleaner approach here?

bduffany:
> I feel like the file seems like a cleaner approach here?

OK, file approach it is then. I was on the fence

tylerwilliams (approved):

enterprise/server/remote_execution/runner/runner_test.go
```diff
@@ -46,28 +47,42 @@ const (
 var (
 	// set by x_defs in BUILD file
 	testworkerRunfilePath string
+)
 
-	defaultCfg = &RunnerPoolOptions{
+func defaultCfg() *RunnerPoolOptions {
```
bduffany, line 52: was seeing some weird test errors that were caused by the `PoolOptions` subfields persisting across test runs. Fixed by making this a func so that each test generates a clean struct each time.

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -291,7 +296,17 @@ func (r *taskRunner) DownloadInputs(ctx context.Context, ioStats *repb.IOStats)
 }
 
 // Run runs the task that is currently bound to the command runner.
-func (r *taskRunner) Run(ctx context.Context) *interfaces.CommandResult {
+func (r *taskRunner) Run(ctx context.Context) (res *interfaces.CommandResult) {
+	defer func() {
+		// Allow tasks to create a special file to skip recycling.
+		exists, err := disk.FileExists(ctx, filepath.Join(r.Workspace.Path(), doNotRecycleMarkerFile))
+		if err != nil {
```
maggie-lou, line 303: I guess we should never/rarely see this case, but should we conservatively set DoNotRecycle=true if there's an error reading the file too?

enterprise/server/remote_execution/runner/runner_test.go
```diff
@@ -830,3 +849,24 @@ func TestRunnerPool_RecycleAfterCreateFailed_CallsRemove(t *testing.T) {
...
+			if createFile {
+				task.ExecutionTask.Command.Arguments = []string{
+					"touch", ".BUILDBUDDY_DO_NOT_RECYCLE",
+				}
+			}
+			r, err := pool.Get(ctx, task)
+			require.NoError(t, err)
+			res := r.Run(ctx)
+			assert.Equal(t, createFile, res.DoNotRecycle)
+			pool.TryRecycle(ctx, r, false)
```
maggie-lou, line 869: What's the reasoning for having this in the test?

maggie-lou (commented):
Sorry I know this is already merged - just some small comments

enterprise/server/remote_execution/runner/runner.go
bduffany: I think long term that is a better approach, but I'd like to sanity check whether these warnings get logged in practice before changing behavior here

enterprise/server/remote_execution/runner/runner_test.go
bduffany: for better or worse, the executor would call `TryRecycle()` with a value of `false` in this case, so I figured it makes sense to do that here as well, just to drive the runner's lifecycle to completion.

There are no assertions here (since there is no return value) but this will at least make sure we don't panic.

PR #7222, opened 2024-08-12
Don't recycle workflow VM if workspace lock is still held

bduffany:
If the workspace lock is still held at the end of the workflow, create the file ".BUILDBUDDY_DO_NOT_RECYCLE" in the action working dir to prevent recycling.

This should fix an issue where workflows can get stuck with the error message `Another command (pid=XXX) is running. Waiting for it to complete on the server (server_pid=XXX)...`

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -633,6 +633,11 @@ func run() error {
 	}
 	os.Setenv("BUILDBUDDY_CI_RUNNER_ABSPATH", absPath)
 
+	// Store the original task workspace dir since we change directories later.
+	taskWorkspaceDir, err := os.Getwd()
+	if err != nil {
+		return err
+	}
 	// Change the current working directory to respect WORKDIR_OVERRIDE, if set.
 	if wd := os.Getenv("WORKDIR_OVERRIDE"); wd != "" {
 		if err := os.MkdirAll(wd, 0755); err != nil {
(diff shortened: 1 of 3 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2506,10 +2534,24 @@ func runCredentialHelper() error {
...
 }
 
+// getBazelCommand returns a bazel command (executable and args) with our custom
+// bazelrc options applied.
+func getBazelCommand(rootPath, bazelCmd string, args ...string) ([]string, error) {
+	startupArgs, err := customBazelrcOptions(rootPath, "")
+	if err != nil {
+		return nil, err
+	}
+	args = append([]string{bazelCmd}, append(startupArgs, args...)...)
```
maggie-lou, line 2544: It's gross that syscall.Exec requires you to pass bazelCmd as both the executable in the first argument and in the list of arguments. Can you clarify this behavior with a comment, or ideally by default not include bazelCmd in the arguments? In most cases people will probably use exec.CommandContext (which we use in `runCommand`) where we don't need this behavior, and people will likely forget to strip it out of the arguments

maggie-lou: Also to simplify, we probably don't need to apply the custom .bazelrc for the workspace lock check

maggie-lou:
This approach sounds good to me as a short term solution to unblock Paxos. As a more optimized solution given that failure to recycle can be somewhat destructive performance-wise, should we try to kill any remaining bazel commands so we can still recycle the snapshot?

bduffany:
> As a more optimized solution given that failure to recycle can be somewhat destructive performance-wise, should we try to kill any remaining bazel commands so we can still recycle the snapshot?

This seems like a good idea but also seems a little tricky to implement properly, since I'm not sure if killing the bazel command will also kill any action subprocesses that it spawns. We probably want to clean those up as well.

enterprise/server/cmd/ci_runner/main.go
bduffany: Changed to not include the bazelCmd in the arguments.

I think we do need the custom .bazelrc since it includes the startup options and I think we need the startup options to match when we run this `--noblock_for_lock` command

maggie-lou (approved):

PR #7223, opened 2024-08-12
Fix .bazelrc logic when bazel_workspace_dir is set

bduffany:
`buildbuddy.yaml` supports a `bazel_workspace_dir` option for testing repos that contain more than one bazel workspace (e.g. rules_xcodeproj contains multiple workspaces to run various integration tests).

We weren't taking this into account when computing the bazel args used to apply our custom bazelrc, meaning we'd treat the workspace bazelrc with lower precedence than we should have.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -717,9 +717,13 @@ func run() error {
 		if err != nil {
 			log.Warningf("Failed to read BuildBuddy config; will run `bazel shutdown` from repo root: %s", err)
 		}
-		wsPath := ""
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		wsPath := cwd
 		if cfg != nil {
-			wsPath = bazelWorkspacePath(cfg)
+			wsPath = filepath.Join(cwd, bazelWorkspacePath(cfg))
 		}
 		args, err := bazelArgsWithCustomBazelrc(rootDir, wsPath, "shutdown")
 		if err != nil {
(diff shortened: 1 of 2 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -717,9 +717,13 @@ func run() error {
 		if err != nil {
 			log.Warningf("Failed to read BuildBuddy config; will run `bazel shutdown` from repo root: %s", err)
 		}
-		wsPath := ""
+		cwd, err := os.Getwd()
```
maggie-lou, line 720: We do a lot of changing directories, so this logic feels a bit brittle. Could we reuse this logic here `bazelWorkspaceAbsPath := filepath.Join(ws.rootDir, repoDirName, action.BazelWorkspaceDir)` and consolidate it in `bazelArgsWithCustomBazelrc`?

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2521,8 +2521,13 @@ func runBazelWrapper() error {
 		return err
 	}
 
+	// Get the current bazel workspace path where we expect to find the
+	// workspace rc file. If this fails just treat it like we aren't in a
+	// workspace.
+	workspacePath, _ := currentBazelWorkspaceAbsPath()
```
maggie-lou, line 2527: Should we at least log the error to the executor?

enterprise/server/cmd/ci_runner/main.go
bduffany: Done

enterprise/server/cmd/ci_runner/main.go
bduffany: Changed it to just return an error - if there's an error here other than NotFound then it's probably some sort of filesystem error, and it shouldn't matter much whether we return it here or if we let bazel run and return the error

maggie-lou (approved):

PR #7225, opened 2024-08-13
Add some parallelism to file cache linking.

vadimberezniker:
In my tests, it increases performance by more than 5x.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -22,6 +22,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/third_party/singleflight"
 	"golang.org/x/sync/errgroup"
+	"google.golang.org/protobuf/types/known/durationpb"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	rspb "github.com/buildbuddy-io/buildbuddy/proto/resource"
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -680,16 +718,27 @@ func (ff *BatchFileFetcher) FetchFiles(filesToFetch FileMap, opts *DownloadTreeO
...
 
-	currentBatchRequestSize := int64(0)
-	eg, ctx := errgroup.WithContext(ff.ctx)
+	linkEG, _ := errgroup.WithContext(ff.ctx)
+	if runtime.GOOS == "darwin" {
+		// Macs don't appear to like too much parallelism, with overall
+		// performance dropping with high parallelism.
+		linkEG.SetLimit(5)
+	} else {
+		linkEG.SetLimit(100)
```
tylerwilliams, line 728: does it make sense to max out at runtime.GOMAXPROCS(0)?

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -702,67 +751,76 @@ func (ff *BatchFileFetcher) FetchFiles(filesToFetch FileMap, opts *DownloadTreeO
...
+			fetchQueue <- digestToFetch{d: d, fps: filePointers}
+			return nil
+		})
+	}
+
+	// Read work off the fetchQueue channel and generate batch read requests
+	// to download data.
+	eg, ctx := errgroup.WithContext(ff.ctx)
+	enqueueWorkDone := make(chan struct{})
+	go func() {
```
tylerwilliams, line 771: did you consider using the errgroup for this goroutine as well? That might let you get rid of enqueueWorkDone and just wait on the eg

tylerwilliams (approved):

tylerwilliams:
Also: great find :) This is really cool

enterprise/server/remote_execution/dirtools/dirtools.go
vadimberezniker: done

PR #7227, opened 2024-08-13
Improve DoNotRecycle logging

bduffany:
- In `ci_runner`, make sure logs are sent on the BES stream so that they appear in the workflow invocation logs
- In the executor, log a message when we don't reycle because of the DO_NOT_RECYCLE file

**Related issues**: N/A


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -762,6 +762,8 @@ func run() error {
 // to be called after the current invocation has completed, to avoid blocking
 // the invocation status from being reported.
 func (ws *workspace) prepareRunnerForNextInvocation(ctx context.Context, taskWorkspaceDir string) {
+	log := ws.log
+
```
maggie-lou, line 766: nit: I'd prefer if you just explicitly use ws.log below to stay consistent with the rest of the file

maggie-lou (approved):

PR #7229, opened 2024-08-13
docs: document static env vars in workflows config

sluongng:


maggie-lou (approved):

docs/workflows-config.md
```diff
@@ -404,6 +404,7 @@ A named group of Bazel commands that run when triggered.
 - **`bazel_workspace_dir`** (`string`): A subdirectory within the repo
   containing the bazel workspace for this action. By default, this is
   assumed to be the repo root directory.
+- **`env`** (`map[string]string`): List of static environment variables and their values.
```
bduffany, line 407: nit: use more language-agnostic type description

```suggestion
- **`env`** (`map` with string values): Map of static environment variables and their values.
```

bduffany, line 407: nit: can you move this up between "user" and "git_fetch_filters" ? I think it fits better there. The rough layout would be

- name / triggers (basic workflow config)
- basic action properties (platform + env)
- git fetch config (less common)
- bazel config + commands

sluongng: Makes sense. Done.

bduffany (approved):

PR #7230, opened 2024-08-13
Fix rules/go/index.bzl to account for rules_go changes

dzbarsky:
Stop using some attributes that are going away.

### DIFF
--- a/rules/go/index.bzl
+++ b/rules/go/index.bzl
@@ -6,12 +6,13 @@ def _go_sdk_tool_impl(ctx):
     # bazel to be able to create an executable symlink to that tool and to be
     # able to recreate the symlink if the tool changes.
     go = go_context(ctx)
-    tool_path = paths.join(go.sdk_root.dirname, ctx.attr.goroot_relative_path)
+    sdk = go.sdk
+    tool_path = paths.join(sdk.root_file.dirname, ctx.attr.goroot_relative_path)
     tool = None
 
     sdk_tools = []
-    sdk_tools.append(go.go)  # `go` binary
-    sdk_tools.extend(go.sdk_tools)  # all other tools, e.g. `gofmt`
+    sdk_tools.append(sdk.go)  # `go` binary
+    sdk_tools.extend(sdk.tools.to_list())  # all other tools, e.g. `gofmt`
 
     for f in sdk_tools:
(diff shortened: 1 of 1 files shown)


rules/go/index.bzl
```diff
@@ -6,12 +6,13 @@ def _go_sdk_tool_impl(ctx):
...
-    tool_path = paths.join(go.sdk_root.dirname, ctx.attr.goroot_relative_path)
+    sdk = go.sdk
+    tool_path = paths.join(sdk.root_file.dirname, ctx.attr.goroot_relative_path)
     tool = None
 
     sdk_tools = []
-    sdk_tools.append(go.go)  # `go` binary
-    sdk_tools.extend(go.sdk_tools)  # all other tools, e.g. `gofmt`
+    sdk_tools.append(sdk.go)  # `go` binary
+    sdk_tools.extend(sdk.tools.to_list())  # all other tools, e.g. `gofmt`
```
sluongng, line 15:
```suggestion
    sdk_tools.extend(sdk.tools)  # all other tools, e.g. `gofmt`
```


```
ERROR: /root/workspace/repo-root/BUILD:194:12: in go_sdk_tool rule //:go:
Traceback (most recent call last):
        File "/root/workspace/repo-root/rules/go/index.bzl", line 15, column 31, in _go_sdk_tool_impl
                sdk_tools.extend(sdk.tools.to_list())  # all other tools, e.g. `gofmt`
Error: 'list' value has no field or method 'to_list'
```

sluongng:
I recreated the PR with a fix here https://github.com/buildbuddy-io/buildbuddy/pull/7237

rules/go/index.bzl
dzbarsky: @sluongng ah this is to account for https://github.com/bazelbuild/rules_go/commit/3994841ad80c48c06db8ce7d65c8a5cb9840e2d1 , so you'll want to fold it into a rules_go upgrade. I'll just close out this PR for now since it's not mergeable in current form

sluongng: cc: @fmeum 

This means that the next rules_go release will be a breaking change. So we should provide some docs and guide folks on how to upgrade in a release note. Not sure if we need to bump the compatibility level in bzlmod either.

PR #7231, opened 2024-08-13
[ci_runner] Support using git credentials pre-configured on self-hosted executors

maggie-lou:
This should allow us to support github enterprise for customers who have configured it with a global credential helper on their self-hosted executors

**Related issues**: N/A


bduffany (approved):
LGTM, as a followup could we also disable our credential helper stuff when this option is set? (maybe add a TODO)

maggie-lou:
> LGTM, as a followup could we also disable our credential helper stuff when this option is set? (maybe add a TODO)

I think this PR already does that? Or are you referring to something else?

https://github.com/buildbuddy-io/buildbuddy/pull/7231/files#diff-a1778f788b2996ecb914bb66cf733fe40e53a23e1d348e392d9e19db1f1b8d77L2045

maggie-lou:
@bduffany Do you think there's ever a case where we'd want to enable osxkeychain? Because it requires human intervention, I can't ever imagine it being useful for CI

bduffany:
I think it's more about using whatever credential helper the user wants to configure - e.g. OSX keychain is the default, but the user might have something in their ~/.gitconfig that they'd want to use. e.g. on my machine I use `gh` as the credential helper and I could see stuff like that being somewhat common

maggie-lou:
@bduffany Here's the follow up PR https://github.com/buildbuddy-io/buildbuddy/pull/7267

I think in practice we'll actually be okay with toggling the option because we [include the repo URL as a platform property](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/hostedrunner/hostedrunner.go#L159) and when we toggle the option, the URL should switch between using ssh / https

PR #7233, opened 2024-08-13
Send re-index requests on commit hook

tylerwilliams:
When we receive a commit hook, if codesearch is enabled, send an async request to reindex the repo.


proto/index.proto
```diff
@@ -1,23 +1,21 @@
...
 }
 
-message GitRepoResponse {}
-
 message IndexRequest {
   context.RequestContext request_context = 1;
 
-  GitRepoRequest git_repo = 2;
+  git.GitRepo git_repo = 2;
+  git.RepoState repo_state = 3;
```
tylerwilliams, line 18: Removed the weirdly-named bespoke gitRepoRequest and just used the protos we already have for this.

codesearch/server/server.go
```diff
@@ -70,12 +71,17 @@ type codesearchServer struct {
 	scratchDirectory string
 }
 
-func makeDoc(name, repo, commitSha string, buf []byte) (types.Document, error) {
+func makeDoc(name, repoURLString, commitSha string, buf []byte) (types.Document, error) {
```
tylerwilliams, line 75: the changes in this file are just manual fix-ups to accommodate the new proto format.

enterprise/server/githubapp/githubapp.go
```diff
@@ -414,10 +416,45 @@ func (a *GitHubApp) maybeTriggerBuildBuddyWorkflow(ctx context.Context, eventTyp
 	if err != nil {
 		return err
 	}
+	if err := a.MaybeReindexRepo(ctx, row.GitRepository, wd, repoURL, tok.GetToken()); err != nil {
+		log.Debugf("Not indexing repo: %s", err)
+	}
```
tylerwilliams, line 421: This is the main significant change: if we receive a webhook, we check if the user has codesearch enabled and possibly send an async RPC to the codesearch endpoint to re-index a repo.

enterprise/server/githubapp/githubapp.go
```diff
@@ -414,10 +416,45 @@ func (a *GitHubApp) maybeTriggerBuildBuddyWorkflow(ctx context.Context, eventTyp
...
 	return a.env.GetWorkflowService().HandleRepositoryEvent(
 		ctx, row.GitRepository, wd, tok.GetToken())
 }
 
+func (a *GitHubApp) MaybeReindexRepo(ctx context.Context, repo *tables.GitRepository, wd *interfaces.WebhookData, repoURL *gitutil.RepoURL, accessToken string) error {
+	codesearchService := a.env.GetCodesearchService()
+	if codesearchService == nil {
+		return nil
+	}
+	if wd.PushedBranch != "main" && wd.PushedBranch != "master" {
```
bduffany, line 431: nit: we get the default branch name in the webhook payload - also probably good to check that this is a push event

```suggestion
	if wd.EventName != webhook_data.EventName.Push || wd.PushedBranch != wd.TargetRepoDefaultBranch {
```

bduffany (approved):

enterprise/server/githubapp/githubapp.go
```diff
@@ -414,10 +416,45 @@ func (a *GitHubApp) maybeTriggerBuildBuddyWorkflow(ctx context.Context, eventTyp
...
+	if err != nil {
+		return err
+	}
+	if !g.CodeSearchEnabled {
+		return nil
+	}
+	_, err = codesearchService.Index(ctx, &csinpb.IndexRequest{
+		GitRepo: &gitpb.GitRepo{
+			RepoUrl:     repoURL.String(),
+			AccessToken: accessToken,
```
maggie-lou, line 444: You aren't using this access token anywhere - you'll have to set it on the repo url to access private repos

(And so I don't forget later - if you do add the token to the repo url, make sure not to log it)

maggie-lou (commented):
Do you have a rough sense for how long indexing takes / how resource intensive it is? If multiple PRs land on master in quick succession, do we want some way to kill previous index attempts? Is this safe if multiple goroutines are trying to index the same repo at the same time? (for ex. from a brief glance it looks like we're writing the results to pebble - can we guarantee that the later commits will always be the last ones to write their index files to the db? is there a chance this could become too resource intensive if large orgs are merging very frequently to master?)

enterprise/server/githubapp/githubapp.go
tylerwilliams: Yup, I am planning to do this in a followup because this CL was already getting too big!

tylerwilliams:
> Do you have a rough sense for how long indexing takes / how resource intensive it is? If multiple PRs land on master in quick succession, do we want some way to kill previous index attempts? Is this safe if multiple goroutines are trying to index the same repo at the same time? (for ex. from a brief glance it looks like we're writing the results to pebble - can we guarantee that the later commits will always be the last ones to write their index files to the db? is there a chance this could become too resource intensive if large orgs are merging very frequently to master?)

These are all good questions! I don't have all the answers yet -- indexing our own repo takes about 12 seconds, the linux kernel takes about a minute.

Not in this CL, but I am planning to use that token sent in the RPC and possibly use go-git or something rather than checking stuff out on disk. Also I will start honoring the replacement strategy -- so one index will blow away previous indexes of the same repo. It probably does make sense to have some form of sequencing -- I'll look into doing this as a queue on the indexing server maybe?

Longer term it's reasonable we might start doing delta indexing, the index works fine for this, but it felt like a hard starting point so I thought I'd start by just always indexing everything.

PR #7238, opened 2024-08-14
Use a configurable fraction of stall time in PSI-adjusted duration

bduffany:
Adds a knob to control the fraction of pressure stall time that we apply as a correction to the execution duration, and adjusts the effective value of the knob from 100% to 75%. This is needed because (for reasons that are still not 100% clear) the PSI stall time is occasionally reported to be very close to the execution duration, resulting in a high computed value for the average milliCPU usage.

Based on analysis of logs, using only 75% of the reported pressure stall time instead of 100% would have fixed all problematic tasks from the last 24h (>16 CPU), putting them under the 30 CPU mark. For non-problematic tasks, this should not affecting task sizes by too much, and preserve enough backpressure to prevent negative feedback loops, since we are still preserving most (75%) of the PSI correction.

Analyzed logs using [this script](https://github.com/buildbuddy-io/buildbuddy/blob/tasksize-parse-logs/parse.py), processing 1600 log entries over the last 24h, where the computed milliCPU was > 16K.

```
millicpu using 0.75*stall_duration:
- min: 2984.2107274066457
- max: 28998.04379154327
- average: 5991.706074873706
- p50: 4686.979561300274
- p90: 8858.734140841832
- p95: 17038.070814486146
- p99: 19499.865554629403

millicpu using 1.00*stall_duration:
- min: 16001.436653717952
- max: 7725139.126009313
- average: 96027.33574049846
- p50: 28856.248638381527
- p90: 140770.50112569623
- p95: 274212.55830794346
- p99: 1238244.4639119518

N=1644
```

### DIFF
--- a/enterprise/server/tasksize/tasksize.go
+++ b/enterprise/server/tasksize/tasksize.go
@@ -26,8 +26,9 @@ import (
 )
 
 var (
-	useMeasuredSizes = flag.Bool("remote_execution.use_measured_task_sizes", false, "Whether to use measured usage stats to determine task sizes.")
-	modelEnabled     = flag.Bool("remote_execution.task_size_model.enabled", false, "Whether to enable model-based task size prediction.")
+	useMeasuredSizes    = flag.Bool("remote_execution.use_measured_task_sizes", false, "Whether to use measured usage stats to determine task sizes.")
+	modelEnabled        = flag.Bool("remote_execution.task_size_model.enabled", false, "Whether to enable model-based task size prediction.")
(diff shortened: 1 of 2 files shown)


tylerwilliams:
I'm OK with this stopgap if you promise to remove this code and fix it properly once we get a better handle on the PSI metrics and why they are seemingly incorrect. SG?

tylerwilliams (approved):

bduffany:
@tylerwilliams SG - I'm treating this with high priority and will do some stress testing after next week

PR #7253, opened 2024-08-16
Add option to replay_invocation to fetch and print log chunks from blobstore

bduffany:
If log chunking is enabled then we strip the Progress output before storing it in blobstore, so the logs are not visible in the replayed invocation.

This PR adds a flag to make these logs accessible - for now we just dump the chunks to stdout/stderr, but a future improvement would be to create synthetic Progress events for these so that we can show them in the UI too.

The logs are also visible in the original invocation but this lets us see the logs in case the UI crashes when trying to view logs, which happened today.

**Related issues**: N/A


tools/replay_invocation/replay_invocation.go
```diff
@@ -181,5 +194,27 @@ func main() {
 			break
 		}
 	}
+
+	// Print invocation log chunks.
+	// TODO: replay these as synthetic progress events.
+	if *printLogs {
+		logsBlobstorePrefix := eventlog.GetEventLogPathFromInvocationIdAndAttempt(*invocationID, uint64(*attemptNumber))
+		log.Infof("Fetching log chunks from %s_*", logsBlobstorePrefix)
+		for i := uint16(0); ; i++ {
```
tempoz, line 203: Is there a reason to read and print the chunks separately? If not we can use `chunkstore.Read`

tools/replay_invocation/replay_invocation.go
```diff
@@ -181,5 +194,27 @@ func main() {
...
+			chunkID := chunkstore.ChunkIndexAsStringId(i)
+			blobName := fmt.Sprintf("%s_%s", logsBlobstorePrefix, chunkID)
```
tempoz, line 205:
```suggestion
			blobName := chunkstore.ChunkName(logsBlobstorePrefix, chunkID)
```

tempoz (approved):

tools/replay_invocation/replay_invocation.go
bduffany: Done. It turned out to be a little trickier than I had hoped, because I had to decide on a good buffer size for the reads.

tools/replay_invocation/replay_invocation.go
bduffany: Using Reader now instead

tempoz (approved):
LGTM

PR #7257, opened 2024-08-16
Update `rules_go` from 0.48.0 -> 0.49.0

tempoz:
Fixes a build problem I get with vet with a newer libc

**Related issues**: N/A


sluongng (changes requested):
Please also bump the rules_go version inside MODULE.bazel file and go.mod file.

tempoz:
Son updated this already

PR #7260, opened 2024-08-16
Consolidate FE logic to fetch execute response

maggie-lou:
* Clean up the logic in invocation_action_card to fallback to fetching the execute response by action digest
* Move the logic to fetch execute response to a util file so the workflow rerun button can use it too (this will let it fallback to fetching the execute response by action digest too)

**Related issues**: N/A

### DIFF
--- a/app/invocation/workflow_rerun_button.tsx
+++ b/app/invocation/workflow_rerun_button.tsx
@@ -23,6 +23,8 @@ import { ChevronDown, RefreshCw } from "lucide-react";
 import Long from "long";
 import { User } from "../auth/user";
 import { firecracker } from "../../proto/firecracker_ts_proto";
+import { fetchExecuteResponseWithFallback } from "./execution_response_utils";
+import ExecuteResponse = build.bazel.remote.execution.v2.ExecuteResponse;
 
 export interface WorkflowRerunButtonProps {
   model: InvocationModel;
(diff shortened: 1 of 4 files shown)


app/invocation/workflow_rerun_button.tsx
```diff
@@ -186,6 +157,32 @@ export default class WorkflowRerunButton extends React.Component<WorkflowRerunBu
     );
   }
 
+  private async getExecuteResponse(): Promise<ExecuteResponse> {
```
bduffany, line 160: I'm not sure we want the AC fallback for the workflow re-run button, since when the user clicks re-run, we probably should guarantee that we fetch the execution for the specific re-run attempt? IIUC fetching from AC could wind up fetching a past execution with the same params?

maggie-lou: I think the AC digest will change every time because we set the invocation_id as a parameter to the ci_runner command

Anyway this isn't that important - I just thought the behavior was confusing that we could load the execute response via the action card, but not the workflow rerun button. But I'm fine closing this if you think the current logic is cleaner

bduffany: As a bit more context I think I mostly kept the AC fallback so that we could roll out the executeResponseDigest stuff a little more safely - i.e. if there's a bug with that logic then we still have something to show that wasn't worse than what we were already doing. I eventually think it'd be good to remove the AC fallback because it's showing the latest execution of the action rather than the one relevant to the particular invocation you're looking at - I feel like the latest ActionResult is not really that useful and probably a bit confusing since it's showing something that could have been executed by a different invocation - wdyt?

PR #7088, opened 2024-07-25
[RB] Support for macs

maggie-lou:
**Related issues**: Based off https://github.com/buildbuddy-io/buildbuddy/pull/6714

PR #7266, opened 2024-08-20
[ci_runner] Fix fetching execute response in the UI

maggie-lou:
This will fix the workflow rerun button not working

**Related issues**: There was a regression from this [PR](https://github.com/buildbuddy-io/buildbuddy/pull/7088) where I switched the workflow service to use BLAKE3 instead of SHA256


app/invocation/invocation_model.tsx
```diff
@@ -405,6 +405,9 @@ export default class InvocationModel {
   getDigestFunction() {
     const digestFnName = this.optionsMap.get("digest_function");
     if (!digestFnName) {
+      if (this.isWorkflowInvocation() || this.isHostedBazelInvocation()) {
+        return build.bazel.remote.execution.v2.DigestFunction.Value.BLAKE3;
+      }
```
bduffany, line 410: it's sort of silly but for other things like this (e.g. remote_instance_name) we have just been passing nop flags to the CI runner like `--remote_instance_name=...` so that we don't have to add too much special casing for the CI runner. So in this case you could pass `--digest_function=blake3` to the CI runner action, and the `this.optionsMap.get("digest_function")` above should return the right thing

bduffany (approved):

PR #7268, opened 2024-08-20
Dont index non-utf8 content

tylerwilliams:


jdhollen (approved):

codesearch/cmd/cli/cli.go
```diff
@@ -142,12 +143,9 @@ func makeDoc(name string) (types.Document, error) {
 
 	// Skip long files.
```
tylerwilliams, line 144: I reordered these clauses to be more conservative. We attempt to throw out a file first by size, then by reading the first few bytes, then finally by checking validutf8 on all the bytes.

PR #7278, opened 2024-08-23
redact: replace with placeholder

sluongng:
We have received some report about our redact logic wrongly remove regex
flag values that contain an `@` character in side. Although I am not
sure if we could make our regex any narrower without reducing the
effectiveness of it, I think it's a confusing experience for our user to
pass some regex via CLI just to get a different regex via our UI.

Instead of silently replace the matched value with an empty string,
effectively yanking out the redaction, replace it using a static
placeholder instead.

### DIFF
--- a/server/util/redact/redact.go
+++ b/server/util/redact/redact.go
@@ -98,7 +98,7 @@ var (
 )
 
 func stripURLSecrets(input string) string {
-	return urlSecretRegex.ReplaceAllString(input, "")
+	return urlSecretRegex.ReplaceAllString(input, "<REDACTED>@")
 }
 
 // Strips URL secrets from the provided flag value, if there is a value.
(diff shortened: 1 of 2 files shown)


bduffany (approved):
This LGTM - as a sanity check, could you make sure that if people are setting `REPO_URL` via `workspace_status.sh` like `echo REPO_URL=$(git remote get-url origin)`, and there's a token included in the `origin` remote URL, that the `<REDACTED>` part doesn't show up in the invocation header?

sluongng:
I set it up locally and got this event `bes.json`

```json
{
  "id": {
    "workspaceStatus": {}
  },
  "workspaceStatus": {
    "item": [
      ...
      {
        "key": "REPO_URL",
        "value": "git@github.com:sluongng/bb-rbe-test.git"
      }
    ]
  }
}
```

However the BuildBuddy raw event will be

```json
{
    "id": {
        "workspaceStatus": {}
    },
    "workspaceStatus": {
        "item": [
            ...
            {
                "key": "REPO_URL",
                "value": "ssh://github.com/sluongng/bb-rbe-test.git"
            }
        ]
    }
}
```

The header is rendered as

```html
<a class="link-wrapper detail clickable" href="/history/repo/sluongng/bb-rbe-test">
  <SVG ...>...</svg>
  sluongng/bb-rbe-test
</a>
```

PR #7279, opened 2024-08-23
Set local CAS client in cache_proxy initialization so it doesn't crash.

iain-macdonald:
Forgot to add this in https://github.com/buildbuddy-io/buildbuddy/pull/7076

**Related issues**: N/A


tylerwilliams (approved):
LGTM, Approval

PR #7288, opened 2024-08-26
Add mitigation for kernel bug in PSI calculation

bduffany:
After investigation, there is strong evidence of a bug somewhere in the kernel [code](https://github.com/torvalds/linux/blob/5be63fc19fcaa4c236b307420483578a56986a37/kernel/sched/psi.c) responsible for reporting PSI durations. I reported a bug here (with more details + small repro program): https://bugzilla.kernel.org/show_bug.cgi?id=219194

If we look at [logs](http://go/bad-cpu-stall) from the last 7d of nonsensical task sizes, almost all of the CPU stall durations for these tasks seem to be about 4.2s, i.e. the suspicious UINT32_MAX nanoseconds number mentioned in the bug report.

At some point, it would be nice to diagnose / fix this bug upstream, but I think it makes sense to add a mitigation for this bug in our code, since any upstream fix would likely to take a long time to develop, get reviewed, be released, and then become available in all environments where the executor is deployed.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/runner/runner.go
+++ b/enterprise/server/remote_execution/runner/runner.go
@@ -309,17 +309,20 @@ func (r *taskRunner) Run(ctx context.Context) (res *interfaces.CommandResult) {
 		// reported.
 		// Also, skip recycling in this case, because the nonsensical result
 		// will persist across tasks.
-		durationUsec := time.Since(start).Microseconds()
+		runDuration := time.Since(start)
 		stats := res.UsageStats
-		if stats.GetCpuPressure().GetFull().GetTotal() > durationUsec {
+		if cpuStallDuration := time.Duration(stats.GetCpuPressure().GetFull().GetTotal()) * time.Microsecond; cpuStallDuration > runDuration {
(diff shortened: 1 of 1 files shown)


iain-macdonald (commented):
Can you add a little bit of logging for when this happens, and maybe also when we receive those suspicious values from the kernel?

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -297,7 +297,33 @@ func (r *taskRunner) DownloadInputs(ctx context.Context, ioStats *repb.IOStats)
...
+			res.DoNotRecycle = true
+		}
+		if stats.GetMemoryPressure().GetFull().GetTotal() > durationUsec {
+			stats.MemoryPressure = nil
+			res.DoNotRecycle = true
+		}
+		if stats.GetIoPressure().GetFull().GetTotal() > durationUsec {
+			stats.IoPressure = nil
+			res.DoNotRecycle = true
+		}
```
tylerwilliams, line 325: logging the values in each case would be nice so we could see if they are all UINT32_MAX

tylerwilliams (approved):

tylerwilliams:
Good find and thanks for filing a bug upstream!

bduffany:
@iain-macdonald Done

PR #7293, opened 2024-08-26
[ci_runner] Move custom kythe code out of binary

maggie-lou:
Use the new bash steps to simplify this codepath.

### DIFF
--- a/enterprise/server/workflow/config/config.go
+++ b/enterprise/server/workflow/config/config.go
@@ -181,7 +181,7 @@ func checkoutKythe() string {
 
 func buildWithKythe(cacheURL string) string {
 	buf := `export KYTHE_DIR="$BUILDBUDDY_CI_RUNNER_ROOT_DIR/kythe-v0.0.67"\n`
-	buf += "build --override_repository kythe_release=$KYTHE_DIR"
+	buf += "bazel --bazelrc=$KYTHE_DIR/extractors.bazelrc build build --override_repository kythe_release=$KYTHE_DIR"
 	if cacheURL != "" {
 		buf += " --remote_cache=" + cacheURL + " --experimental_remote_cache_compression"
 	}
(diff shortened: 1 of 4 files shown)


enterprise/server/workflow/config/config.go
```diff
@@ -167,16 +168,33 @@ func NewConfig(r io.Reader) (*BuildBuddyConfig, error) {
 	return cfg, nil
 }
 
+const kytheDownloadURL = "https://storage.googleapis.com/buildbuddy-tools/archives/kythe-v0.0.67.tar.gz"
+
+func checkoutKythe() string {
```
maggie-lou, line 173: nit: I find both bash blocks more readable when it's a single block with `` because the spacing is more obvious

tylerwilliams: PTAL

enterprise/server/workflow/config/config.go
```diff
@@ -167,16 +168,35 @@ func NewConfig(r io.Reader) (*BuildBuddyConfig, error) {
...
+	buf := fmt.Sprintf(`
+export KYTHE_DIR="$BUILDBUDDY_CI_RUNNER_ROOT_DIR/kythe-v0.0.67"
+if [ ! -d "$KYTHE_DIR" ]; then
+  mkdir -p "$KYTHE_DIR"
+  curl -sL "%s" | tar -xz -C "$KYTHE_DIR" --strip-components 1
+fi`, kytheDownloadURL)
+	return buf
+}
+
+func buildWithKythe(cacheURL string) string {
```
bduffany, line 183: nit: to avoid this plumbing you could also use `--config=buildbuddy_remote_cache` which gets configured in our generated bazelrc

bduffany (approved):
nice

enterprise/server/workflow/config/config.go
tylerwilliams: Done.

PR #7299, opened 2024-08-26
Clean up parent invocation ID on children if the parent is retried

maggie-lou:
This refactor https://github.com/buildbuddy-io/buildbuddy/pull/7138 has the UI fetch child invocations for workflows from the DB. I noticed that [some workflows](https://app.buildbuddy.dev/invocation/536a0418-b197-447c-b73b-14a1fc544e60?role=CI_RUNNER) have duplicate child invocations. I think this is happening because the original ci_runner execution was retried due to an executor restart, but the UI is still fetching children from earlier attempts

**Related issues**: N/A


bduffany (commented):
(Discussed offline - this is superseded by https://github.com/buildbuddy-io/buildbuddy/pull/7318)

PR #7304, opened 2024-08-26
Support termination grace period on RBE

bduffany:
This PR adds an exec property `termination-grace-period` that matches bazel's semantics of `--local_termination_grace_seconds` - start a graceful shutdown when the timeout is reached, then after the grace period has elapsed, do a force-shutdown. The grace period is 0 by default, but eventually it might be nice to have it default to 15s to match bazel's local default.

The proposed implementation adds a `Signal()` function to the `CommandContainer` interface, which works similarly to `podman kill`. For now, implementations are written only for `oci`, `podman`, and `bare` isolation types - other isolation types just log a warning when graceful termination would normally be started. We can implement `firecracker` too, but it's slightly more complicated because of the streaming vmexec RPC.

The implementation is done in such a way that it should not interfere with executor shutdown or invocation cancellation. Executor shutdown is still able to force-kill tasks even if they are in their termination grace period, and they will be retried like we do today.

**Related issues**: N/A


enterprise/server/remote_execution/executor/executor.go
```diff
@@ -249,14 +272,38 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
...
 
+	// If graceful termination is requested, start a goroutine to request
+	// termination after the termination timeout.
+	gracefullyTerminated := make(chan struct{})
+	if execTimeouts.TerminateAfter < execTimeouts.ForceKillAfter {
+		go func() {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(time.Until(terminateAt)):
```
vadimberezniker, line 297: pre 1.23 this will leak the timer :)

vadimberezniker (approved):

iain-macdonald (approved):

enterprise/server/remote_execution/executor/executor.go
bduffany: heh I totally knew that :P

tylerwilliams (approved):

PR #7318, opened 2024-08-27
Fetch child invocations by run ID to account for retries

maggie-lou:
This PR re-enables two PRs that support showing child invocations in the UI for workflows with bash commands

- Revert "Revert "[ci_runner] Fetch child invocations from the db instead of BES events" (#7305)"
- Revert "Revert "[ci_runner] Populate child invocation cards in UI for bazel commands in bash" (#7298)"

**Related issues**: N/A

### DIFF
--- a/app/invocation/BUILD
+++ b/app/invocation/BUILD
@@ -97,7 +97,8 @@ ts_library(
     deps = [
         "//app/components/link",
         "//app/format",
-        "//proto:build_event_stream_ts_proto",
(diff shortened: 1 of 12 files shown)


maggie-lou:
I realized that these commits should actually go out separately, because we should write the new data to the DB for some time before switching the UI to use the new data source. Here's the separated run ID PR: https://github.com/buildbuddy-io/buildbuddy/pull/7333

bduffany (approved):

PR #7319, opened 2024-08-27
Serve GetTree requests out of the local cache via BatchReadBlobs

iain-macdonald:
This is a pretty simple implementation of GetTree. I looked into trying to reuse the implementation in content_addressable_storage_server, but I think that adding support for locally writing remotely read directories will add enough complexity that it's not warranted, at least for now.

I also refactored the three instances of `gRPCMaxSize` out into `rpcutil`, using the lower of the two values defined in the code, rather than adding a fourth redundant definition of this constant.

**Related issues**: N/A


tylerwilliams (approved):

iain-macdonald:
@vadimberezniker I'd like a third pair of eyes on this so will wait for your review.

vadimberezniker (approved):

PR #7324, opened 2024-08-28
Add debug logging to runner

maggie-lou:
Add some temporary debug logging to help debug a customer issue where there is a 3 min delay between when the executor claims to be executing a task and the command starts executing: https://buildbuddy-corp.slack.com/archives/C019DS52XAR/p1724833048078539?thread_ts=1710341219.838939&cid=C019DS52XAR

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -437,11 +437,15 @@ type VM interface {
 // PullImageIfNecessary pulls the image configured for the container if it
 // is not cached locally.
 func PullImageIfNecessary(ctx context.Context, env environment.Env, ctr CommandContainer, creds oci.Credentials, imageRef string) error {
-	ctx, span := tracing.StartSpan(ctx)
-	defer span.End()
-	if *debugUseLocalImagesOnly {
+	if *debugUseLocalImagesOnly || imageRef == "" {
 		return nil
 	}
+
+	ctx, span := tracing.StartSpan(ctx)
+	defer span.End()
+	log.CtxDebugf(ctx, "Preparing to pull image")
+	defer log.CtxDebugf(ctx, "Finished pulling image")
+
 	cacheAuth := env.GetImageCacheAuthenticator()
 	if cacheAuth == nil || env.GetAuthenticator() == nil {
 		// If we don't have an authenticator available, fall back to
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/containers/bare/bare.go
```diff
@@ -90,6 +92,7 @@ func (c *bareCommandContainer) exec(ctx context.Context, cmd *repb.Command, work
 	}
 
 	if *enableLogFiles {
+		log.CtxDebugf(ctx, "Enabling log files")
```
bduffany, line 95:
```suggestion
		log.CtxDebugf(ctx, "Creating log files")
```

enterprise/server/remote_execution/runner/runner.go
```diff
@@ -212,7 +212,8 @@ func (r *taskRunner) String() string {
 		truncate(r.key.InstanceName, 8, "..."), truncate(ph, 8, ""))
 }
 
-func (r *taskRunner) pullCredentials() (oci.Credentials, error) {
+func (r *taskRunner) pullCredentials(ctx context.Context) (oci.Credentials, error) {
+	log.CtxDebugf(ctx, "Pulling credentials")
```
bduffany, line 216: fyi this doesn't pull credentials, it just returns a struct

enterprise/server/remote_execution/container/container.go
```diff
@@ -437,11 +437,15 @@ type VM interface {
 // PullImageIfNecessary pulls the image configured for the container if it
 // is not cached locally.
 func PullImageIfNecessary(ctx context.Context, env environment.Env, ctr CommandContainer, creds oci.Credentials, imageRef string) error {
-	ctx, span := tracing.StartSpan(ctx)
-	defer span.End()
-	if *debugUseLocalImagesOnly {
+	if *debugUseLocalImagesOnly || imageRef == "" {
```
tylerwilliams, line 440: 👀

tylerwilliams (approved):

bduffany (approved):

enterprise/server/remote_execution/container/container.go
maggie-lou: This is the only thing that seems like it could even take time... I'm suspicious

enterprise/server/remote_execution/runner/runner.go
iain-macdonald: We should probably rename it to not be a verb-noun when it doesn't do the verb. Something like `getCredentials()` or `getPullCredentials()`, but not urgent!

enterprise/server/remote_execution/container/container.go
maggie-lou: It is for the bare runner, which we often use for mac executors

iain-macdonald: gah sorry, i thought this was in podman.go, my bad

PR #7326, opened 2024-08-28
Disable flaky cache requests card assertions

bduffany:
**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3719

### DIFF
--- a/enterprise/server/test/webdriver/invocation/invocation_test.go
+++ b/enterprise/server/test/webdriver/invocation/invocation_test.go
@@ -70,12 +70,15 @@ func TestAuthenticatedInvocation_CacheEnabled(t *testing.T) {
 	require.Contains(t, details, "Cache on")
 	require.Contains(t, details, "Remote execution off")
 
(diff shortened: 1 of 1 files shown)


maggie-lou:
Hot take - I'd kind of prefer leaving the flaky test so there's some incentive to fix it

tylerwilliams:
> Hot take - I'd kind of prefer leaving the flaky test so there's some incentive to fix it

I mostly agree (and think we should fix this!), but when it blocks the releases from succeeding and someone has to notice and go mash the retry button a bunch (waiting each time to see if it's successful), it feels like a net negative.

tylerwilliams (approved):

maggie-lou:
I kind of feel like it's the on-call person's job to fix the test in that case? I'm on call next week so I'll put my money where my mouth is and fix it when it inevitably burns me :P

tylerwilliams:
lol I feel like this is a be careful what you wish for situation -- but I'm happy to close this for now and we can chat offline about prioritizing the flakes that need fixing!

PR #7331, opened 2024-08-28
Cache Proxy: Update remote-cache-atime of locally-accessed cache entries

iain-macdonald:
Whoops, this PR turned out a bit bigger than I expected. Most of it's tests, though, so hopefully it's still palatable but let me know if not.

This PR adds an `AtimeUpdater` interface and implementation that exposes functions for enqueueing atime updates to be written to the remote cache. This is intended to address the issue that artifacts served out of the Cache Proxy's local cache won't have their remote atimes updated and thus may expire prematurely. The `AtimeUpdater` is best effort and will drop requested updates if the number of pending updates exceeds flag-configured thresholds. Internally, it maintains a set of "updates" (effectively FindMissingBlobsRequests) per group, with one update per instance-name/digest-type per group. It flushes these to the remote cache in a round robin fashion on an interval that's controlled by a flag, sending one update per group per tick to try and achieve some degree of fairness between groups. Lastly, this PR hooks up the BS proxy and CAS proxy to the atime updater, enqueuing updates for locally accessed cache entries.

Interesting things that may warrant discussion:
- I chose not to send updates through a channel, instead passing them synchronously, copying them into the set of pending updates, and then returning. This may be slow or block in some cases, but I tried to prevent that.
- I bet the updater is passing arguments by value instead of by pointer a bit too much and that should probably be cleaned up for performance.
- I'm not sure it's necessary to enqueue updates for BS.Read() and CAS.BatchGetBlobs(), but I couldn't convince myself it's _not_ necessary...

**Related issues**: N/A

### DIFF
--- a/enterprise/server/atime_updater/atime_updater.go
+++ b/enterprise/server/atime_updater/atime_updater.go
@@ -1,0 +1,236 @@
+package atime_updater
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
+	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
+	"github.com/buildbuddy-io/buildbuddy/server/remote_cache/digest"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
+
+	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
+)
+
+var (
+	// The maximum gRPC message size is 4MB. Each digest proto is
+	// 64 bytes (hash) + 8 bytes (size) + 2 bytes (tags) = 74 bytes. The
+	// remaining fields in FindMissingBlobsRequest should be small (1kb?), so
+	// we can conceivably fit 4MB / 74bytes = 14.1k of these. Cap it at 10k for
+	// now and we can increase it (or increase the update frequency) if we drop
+	// a lot of updates. Note this also affects memory usage of the proxy!
+	atimeUpdaterMaxDigestsPerUpdate = flag.Int("cache_proxy.remote_atime_max_digests_per_update", 10*1000, "The maximum number of blob digests to send in a single request for updating blob access times in the remote cache.")
+	atimeUpdaterMaxUpdatesPerGroup  = flag.Int("cache_proxy.remote_atime_max_updates_per_group", 25, "The maximum number of FindMissingBlobRequests to accumulate per group for updating blob access times in the remote cache.")
+	atimeUpdaterBatchUpdateInterval = flag.Duration("cache_proxy.remote_atime_update_interval", 30*time.Second, "The time interval to wait between sending access time updates to the remote cache.")
+	atimeUpdaterRpcTimeout          = flag.Duration("cache_proxy.remote_atime_rpc_timeout", 1*time.Second, "RPC timeout to use when updating blob access times in the remote cache.")
+)
+
+// A pending batch of atime updates (basically a FindMissingBlobsRequest).
+type atimeUpdate struct {
+	instanceName   string
+	digests        map[digest.Key]struct{} // stored as a set for de-duping.
+	digestFunction repb.DigestFunction_Value
+}
+
+func (u *atimeUpdate) toProto() *repb.FindMissingBlobsRequest {
+	req := repb.FindMissingBlobsRequest{
+		InstanceName:   u.instanceName,
+		BlobDigests:    []*repb.Digest{},
+		DigestFunction: u.digestFunction,
+	}
+	for d := range u.digests {
+		req.BlobDigests = append(req.BlobDigests, d.ToDigest())
+	}
+	return &req
+}
+
+// The set of pending access time updates to send per group. There should be at
+// most one of these per group, but each contains many atimeUpdate{}s, as
+// there is one per (groupID, instance-name, digest-function) tuple. However,
+// the update will only keep one such atimeUpdate{} and if it gets too big will
+// discard additional digests until it's flushed.
+type atimeUpdates struct {
+	mu      sync.Mutex // protects jwt, updates, and atimeUpdate.digests.
+	jwt     string
+	updates []*atimeUpdate
+}
+
+type atimeUpdater struct {
+	authenticator interfaces.Authenticator
+
+	mu      sync.Mutex               // protects updates
+	updates map[string]*atimeUpdates // pending updates keyed by groupID.
+
+	maxDigestsPerUpdate int
+	maxUpdatesPerGroup  int
+	updateInterval      time.Duration
+	rpcTimeout          time.Duration
+
+	remote repb.ContentAddressableStorageClient
+}
+
+type Opts struct {
+	TickerForTesting <-chan time.Time
+}
+
+func Register(env *real_environment.RealEnv, opts Opts) error {
+	authenticator := env.GetAuthenticator()
+	if authenticator == nil {
+		return fmt.Errorf("An Authenticator is required to enable the AtimeUpdater.")
+	}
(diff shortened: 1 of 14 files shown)


enterprise/server/atime_updater/atime_updater.go
```diff
@@ -0,0 +1,236 @@
...
+type atimeUpdate struct {
+	instanceName   string
+	digests        map[digest.Key]struct{} // stored as a set for de-duping.
+	digestFunction repb.DigestFunction_Value
+}
+
+func (u *atimeUpdate) toProto() *repb.FindMissingBlobsRequest {
+	req := repb.FindMissingBlobsRequest{
+		InstanceName:   u.instanceName,
+		BlobDigests:    []*repb.Digest{},
```
tylerwilliams, line 41: you can allocate this to the correct length / capacity (to reduce allocations) since you know how long u.digests is:

either: ```
...
BlobDigests: make([]*repb.Digest, len(u.digests)),
...

for i, d := range listOfDigests {
  req.BlobDigests[i] = d.ToDigest()
}

or if you just want to do capacity:

...
BlobDigests: make([]*repb.Digest, 0, len(u.digests)),
...

for d := range u.digests {
  req.BlobDigests = append(req.BlobDigests, d.ToDigest())
} 
```

File: enterprise/server/atime_updater/atime_updater.go [hunk 2]
```diff
@@ -0,0 +1,236 @@
...
+	// now and we can increase it (or increase the update frequency) if we drop
+	// a lot of updates. Note this also affects memory usage of the proxy!
+	atimeUpdaterMaxDigestsPerUpdate = flag.Int("cache_proxy.remote_atime_max_digests_per_update", 10*1000, "The maximum number of blob digests to send in a single request for updating blob access times in the remote cache.")
+	atimeUpdaterMaxUpdatesPerGroup  = flag.Int("cache_proxy.remote_atime_max_updates_per_group", 25, "The maximum number of FindMissingBlobRequests to accumulate per group for updating blob access times in the remote cache.")
+	atimeUpdaterBatchUpdateInterval = flag.Duration("cache_proxy.remote_atime_update_interval", 30*time.Second, "The time interval to wait between sending access time updates to the remote cache.")
+	atimeUpdaterRpcTimeout          = flag.Duration("cache_proxy.remote_atime_rpc_timeout", 1*time.Second, "RPC timeout to use when updating blob access times in the remote cache.")
+)
+
+// A pending batch of atime updates (basically a FindMissingBlobsRequest).
+type atimeUpdate struct {
```
tylerwilliams, line 32: did you try to use FindMissingBlobsRequest directly for this and it was painful?

enterprise/server/atime_updater/atime_updater.go
```diff
@@ -0,0 +1,236 @@
...
+	u.Enqueue(ctx, req.InstanceName, req.BlobDigests, req.DigestFunction)
+}
+
+func (u *atimeUpdater) start(opts Opts) {
+	tickerChan := opts.TickerForTesting
+	if tickerChan == nil {
+		ticker := time.NewTicker(u.updateInterval)
+		tickerChan = ticker.C
+		defer ticker.Stop()
+	}
```
tylerwilliams, line 195: would it work to add a `TickerForTestingOnly() chan time.Time` method to atimeUpdate and grab the one you create here? Then you could omit Opts

tylerwilliams (approved):
mostly LGTM -- see what @vadimberezniker thinks?

(Prolly should also run these tests with --config=race and make sure it's happy if you haven't already)

enterprise/server/atime_updater/atime_updater.go
iain-macdonald: sg, done

enterprise/server/atime_updater/atime_updater.go
iain-macdonald: yeah because to support de-duping digests i need to store them in a set. added a comment explaining.

iain-macdonald:
> mostly LGTM -- see what @vadimberezniker thinks?
> 
> (Prolly should also run these tests with --config=race and make sure it's happy if you haven't already)

Thanks! Yeah, I ran it with the race detector and `--runs_per_test=100` and it came out clean.

enterprise/server/atime_updater/atime_updater.go
```diff
@@ -0,0 +1,239 @@
...
+var (
+	// The maximum gRPC message size is 4MB. Each digest proto is
+	// 64 bytes (hash) + 8 bytes (size) + 2 bytes (tags) = 74 bytes. The
+	// remaining fields in FindMissingBlobsRequest should be small (1kb?), so
+	// we can conceivably fit 4MB / 74bytes = 14.1k of these. Cap it at 10k for
+	// now and we can increase it (or increase the update frequency) if we drop
+	// a lot of updates. Note this also affects memory usage of the proxy!
+	atimeUpdaterMaxDigestsPerUpdate = flag.Int("cache_proxy.remote_atime_max_digests_per_update", 10*1000, "The maximum number of blob digests to send in a single request for updating blob access times in the remote cache.")
+	atimeUpdaterMaxUpdatesPerGroup  = flag.Int("cache_proxy.remote_atime_max_updates_per_group", 25, "The maximum number of FindMissingBlobRequests to accumulate per group for updating blob access times in the remote cache.")
+	atimeUpdaterBatchUpdateInterval = flag.Duration("cache_proxy.remote_atime_update_interval", 30*time.Second, "The time interval to wait between sending access time updates to the remote cache.")
```
vadimberezniker, line 27: We will need to monitor performance and potentially reduce this interval. I think it'll be healthier for app load to get smaller but more frequent batches.

enterprise/server/atime_updater/atime_updater.go
```diff
@@ -0,0 +1,239 @@
...
+	updater := atimeUpdater{
+		authenticator:       authenticator,
+		updates:             make(map[string]*atimeUpdates),
+		maxDigestsPerUpdate: *atimeUpdaterMaxDigestsPerUpdate,
+		maxUpdatesPerGroup:  *atimeUpdaterMaxUpdatesPerGroup,
+		updateInterval:      *atimeUpdaterBatchUpdateInterval,
+		rpcTimeout:          *atimeUpdaterRpcTimeout,
+		remote:              remote,
+	}
+	go updater.start(opts)
```
vadimberezniker, line 100: Doesn't need to be in this PR, but I would like to see a best-effort attempt to drain the pending atime updates when shutdown is trigerrede.

enterprise/server/atime_updater/atime_updater.go
```diff
@@ -0,0 +1,239 @@
...
+	maxDigestsPerUpdate int
+	maxUpdatesPerGroup  int
+	updateInterval      time.Duration
+	rpcTimeout          time.Duration
+
+	remote repb.ContentAddressableStorageClient
+}
+
+type Opts struct {
+	TickerForTesting <-chan time.Time
```
vadimberezniker, line 79: nit: this is also achievable using the clockwork fake clock. I would like to see us standardize on a single approach for testing behaviors that depend on time.

vadimberezniker (approved):

enterprise/server/atime_updater/atime_updater.go
iain-macdonald: Yup, will keep an eye on it

enterprise/server/atime_updater/atime_updater.go
iain-macdonald: Good call, added a TODO

enterprise/server/atime_updater/atime_updater.go
iain-macdonald: Thanks for adding that! Switched all the tests here to use it and dropped the `Opts`.

PR #7332, opened 2024-08-29
Add output inlining support to the action cache

fmeum:
Bazel will start requesting output file inlining with https://github.com/bazelbuild/bazel/pull/23428.

**Related issues**: N/A


server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -173,6 +174,7 @@ func (s *ActionCacheServer) GetActionResult(ctx context.Context, req *repb.GetAc
 	if err := ValidateActionResult(ctx, s.cache, req.GetInstanceName(), req.GetDigestFunction(), rsp); err != nil {
 		return nil, status.NotFoundErrorf("ActionResult (%s) not found: %s", d, err)
 	}
+	s.maybeInlineOutputFiles(ctx, req, rsp, 4*1024*1024)
```
fmeum, line 177: Is the maximum message size configurable or do we have a constant for this somewhere?

I made it a parameter since this number will need to be modified when we use this for `ExecuteRequest` (since there are more fields than just `ActionResult` in `ExecuteResponse`).

server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -242,3 +244,59 @@ func (s *ActionCacheServer) UpdateActionResult(ctx context.Context, req *repb.Up
 	}
 	return req.ActionResult, nil
 }
+
+// Inlines the contents of output files requested to be inlined as long as the
+// total size of the ActionResult is below maxResultSize.
+func (s *ActionCacheServer) maybeInlineOutputFiles(ctx context.Context, req *repb.GetActionResultRequest, ar *repb.ActionResult, maxResultSize int) {
```
fmeum, line 250: I could use some hints on how to write tests for `action_cache_server`.

server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -242,3 +244,58 @@ func (s *ActionCacheServer) UpdateActionResult(ctx context.Context, req *repb.Up
...
+	if len(actuallyInlinedFiles) == 0 {
+		return
+	}
+
+	g, gCtx := errgroup.WithContext(ctx)
+	for _, f := range actuallyInlinedFiles {
+		fc := f
+		g.Go(func() error {
+			rn := digest.NewResourceName(fc.GetDigest(), req.GetInstanceName(), rspb.CacheType_CAS, req.GetDigestFunction()).ToProto()
+			blob, err := s.cache.Get(gCtx, rn)
```
fmeum, line 291: Does this update the usual download metrics?

server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -242,3 +244,58 @@ func (s *ActionCacheServer) UpdateActionResult(ctx context.Context, req *repb.Up
...
+			if err != nil {
+				// Inlining is best-effort ("The server MAY omit inlining...").
+				return nil
+			}
```
bduffany, line 295: nit: should we add a prometheus metric here so that we have a way to track client usage / hit rate for these?

server/remote_cache/action_cache_server/action_cache_server.go
bduffany: Good catch - hits/misses are tracked in the API layer rather than the storage layer (cache) so I think you'd need to construct a HitTracker here and track any hits/misses.

One interesting thing is that each hit tracked will appear as a separate row in the cache requests card, even though there is only a single client "request," though this seems OK for now. In the future, it'd be neat to somehow attach the AC request info to these rows, so that we can show them in the UI as "inlined" or something.

server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -242,3 +244,58 @@ func (s *ActionCacheServer) UpdateActionResult(ctx context.Context, req *repb.Up
...
+		}
+		budget -= totalSize
+		actuallyInlinedFiles = append(actuallyInlinedFiles, f)
+	}
+
+	if len(actuallyInlinedFiles) == 0 {
+		return
+	}
+
+	g, gCtx := errgroup.WithContext(ctx)
```
bduffany, line 286: nit: maybe `g.SetLimit(16)` or something, to guard against requests that are requesting lots of small output files

server/remote_cache/action_cache_server/action_cache_server.go
bduffany: Looks like `action_cache_server_test` doesn't exist yet - it would be awesome if you could add that. You could use `content_addressable_storage_server_test` as an example.

server/remote_cache/action_cache_server/action_cache_server.go
bduffany: We configure a max recv message size in grpc_server.go, but not a max send msg size. I'm not sure if there is a default limit imposed by the grpc lib that we should reference, or if there are other effective limits we'd need to consider.

server/remote_cache/action_cache_server/action_cache_server.go
tylerwilliams, line 291: since we're going to hold all these digests in memory anyway, you're already reasonably sure they are smaller than your ~4MB limit, you could use cache.GetMulti here instead

server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -242,3 +244,58 @@ func (s *ActionCacheServer) UpdateActionResult(ctx context.Context, req *repb.Up
...
+			continue
+		}
+		// An additional "contents" field requires 1 byte for the tag field
+		// (5:LEN), the bytes for the varint encoding of the length of the
+		// contents and the contents themselves.
+		totalSize := 1 + int64(binary.Size(uint64(contentsSize))) + contentsSize
+		if budget < totalSize {
+			continue
+		}
+		budget -= totalSize
```
sluongng, line 278: Nit: we can add an if `budget <= 0` at the beginning of the loop and break.

server/remote_cache/action_cache_server/action_cache_server.go
```diff
@@ -242,3 +244,58 @@ func (s *ActionCacheServer) UpdateActionResult(ctx context.Context, req *repb.Up
...
+
+	g, gCtx := errgroup.WithContext(ctx)
+	for _, f := range actuallyInlinedFiles {
+		fc := f
+		g.Go(func() error {
+			rn := digest.NewResourceName(fc.GetDigest(), req.GetInstanceName(), rspb.CacheType_CAS, req.GetDigestFunction()).ToProto()
+			blob, err := s.cache.Get(gCtx, rn)
+			if err != nil {
+				// Inlining is best-effort ("The server MAY omit inlining...").
+				return nil
```
sluongng, line 294: Small nit on the error handling here (which should still apply when we switch to GetMulti): in the ValidateActionResult step right before this, we made sure that all the blobs referenced by the ActionResult are available before returning the action result. Else, we just return a not found error so Bazel could re-run the action.

The same principle should be applied here: if the referenced blob cannot be found in our cache, we should just return not found rather than try to not inline it. Because in subsequent requests from Bazel to download the blob, it might error out with cryptic message on the client side.

Bonus if we can combine some of these inline logic with the checkFilesExist in ValidateActionResult somehow (i.e. no need to validate blobs which we gonna inline). But it's probably not needed in the first iteration.

server/remote_cache/action_cache_server/action_cache_server.go
fmeum: I can no longer do that as I am sending the metrics.

server/testutil/testmetrics/testmetrics.go
```diff
@@ -67,6 +68,32 @@ func HistogramVecValues(t testing.TB, metric *prometheus.HistogramVec) []Histogr
 	return out
 }
 
+// AssertHistogramSamples asserts that the given histogram metric recorded the
+// expected samples.
+func AssertHistogramSamples(t testing.TB, metric *prometheus.HistogramVec, samples ...float64) {
```
fmeum, line 73: @bduffany I found this helper easier to use than the existing one for my use case as I don't explicitly specify the buckets. What do you think?

PR #7333, opened 2024-08-29
Add run ID to invocations to link parent invocation attempts with children

maggie-lou:
This PR adds the concept of a "run ID" - a unique identifier for each invocation retry attempt - so that we can fetch children invocations for only the current retry attempt

We should not fetch children invocations using the parent's invocation ID, because if the parent has been retried, we will fetch children from all retry attempts

**Related issues**: N/A

### DIFF
--- a/server/tables/tables.go
+++ b/server/tables/tables.go
@@ -157,6 +157,7 @@ type Invocation struct {
 	InvocationUUID                   []byte `gorm:"size:16;default:NULL;uniqueIndex:invocation_invocation_uuid;unique"`
 	Success                          bool
 	Attempt                          uint64 `gorm:"not null;default:0"`
+	RunID                            string
 	BazelExitCode                    string
 
 	// The user-specified setting of how to download outputs from remote cache.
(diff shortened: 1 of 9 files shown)


server/tables/tables.go
```diff
@@ -173,6 +174,7 @@ type Invocation struct {
 	Tags string
 
 	ParentInvocationID string `gorm:"index:parent_invocation_id_index"`
```
bduffany, line 176: should we remove this index on parent_invocation_id ?

bduffany (approved):

server/tables/tables.go
maggie-lou: Yeah I'll clean this up separately. Do you think it's still worth keeping that field? We could derive it from joining on ParentRunID instead

bduffany: hm I guess we use this for rendering the parent workflow invocation ID in the breadcrumb, so we might want to keep it for now.

maggie-lou: No I don't think we use this field anywhere because I just added it. In the breadcrumb, I think we use the fact that the parent invocation ID is in the build metadata

We don't have a specific use case for it now, so I'll just delete the field and the index

PR #7334, opened 2024-08-29
[ci_runner] Support build remotely and running locally via bash command

maggie-lou:
We have a remote bazel feature that builds the target remotely and runs the executable locally. This happens automatically in the CLI, as opposed to building the target remotely, and then the user manually running the executable that was fetched locally (here's the [PR](https://github.com/buildbuddy-io/buildbuddy/commit/fa9409e55400bd34c19b7e7282d4468999c056fc) where the behavior was introduced). This PR adds support in the new Steps syntax - this is the last missing feature before feature parity with BazelCommands

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -73,6 +73,8 @@ const (
 	// Name of the bazel output base dir. This is written under the workspace
 	// so that it can be cleaned up when the workspace is cleaned up.
 	outputBaseDirName = "output-base"
+	// Name of the dir where we write bazel run scripts.
+	runScriptDirName = "bazel-run-scripts"
 
 	// Fraction of disk space that must be in use before we attempt to reclaim
 	// disk space.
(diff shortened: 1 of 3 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -1391,6 +1432,16 @@ func (ar *actionRunner) workspaceStatusEvent() *bespb.BuildEvent {
...
 
+// Returns the bazel command - i.e. `build` or `run`
+func getBazelCommand(bazelArgs []string) string {
+	for _, arg := range bazelArgs {
+		if !strings.HasPrefix(arg, "--") {
+			return arg
+		}
+	}
+	return ""
+}
```
bduffany, line 1443: I think ideally we'd centralize any parsing logic in the CLI since it has more information about the bazel flag schema (`cli/parser` package). Could the CLI maybe detect whether we're doing a remote run and then pass --script_path as an additional arg?

bduffany (approved):

enterprise/server/cmd/ci_runner/main.go
maggie-lou: Good idea - done

PR #7336, opened 2024-08-29
Small improvements to replay_action

bduffany:
- Make log output less noisy and more useful - switch some logs to `debug` level and add `run=i/N` context to all execution logs
- Log stdout+stderr if `--jobs=1`, not just when `--n=1`.
- Bring back `--skip_copy_files` - even though we're using batched FindMissing requests to reduce the amount of work, it can still take a long time to make all the FindMissing requests.

**Related issues**: N/A

### DIFF
--- a/enterprise/tools/replay_action/replay_action.go
+++ b/enterprise/tools/replay_action/replay_action.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


enterprise/tools/replay_action/replay_action.go
```diff
@@ -70,7 +72,7 @@ func copyFile(srcCtx, targetCtx context.Context, fmb *FindMissingBatcher, to, fr
 		return err
 	}
 	if exists {
-		log.Infof("Copy %s: already exists", digest.String(outd.GetDigest()))
+		log.Debugf("Copy %s: already exists", digest.String(outd.GetDigest()))
```
iain-macdonald, line 75: CtxDebugf?

enterprise/tools/replay_action/replay_action.go
```diff
@@ -203,7 +210,8 @@ func main() {
 		log.Fatalf("Error fetching action: %s", err.Error())
 	}
 	// If remote_executor and target_executor are not the same, copy the files.
-	if inCopyMode() {
+	if inCopyMode() && !*skipCopyFiles {
+		log.Infof("Copying files to target.")
```
iain-macdonald, line 214: CtxInfof?

iain-macdonald (approved):

PR #7342, opened 2024-08-30
ci_runner: fix insteadOf replacement

sluongng:
When the remote url is `ssh://` then the subsequent colon will be used
for server port. We should be using forward-slash `/` for repo namespace
separator.


sluongng:
Context: https://buildbuddy-corp.slack.com/archives/C07GMM2VBLY/p1724919026663969?thread_ts=1723402039.050079&cid=C07GMM2VBLY

enterprise/server/cmd/ci_runner/main.go
```diff
@@ -2495,7 +2495,7 @@ func configureGlobalURLRewrites(ctx context.Context) error {
 	// Discard any existing rewrites initially, then append.
 	flag := "--replace-all"
 	for original, replacement := range map[string]string{
-		"ssh://git@github.com:": "https://github.com/",
+		"ssh://git@github.com/": "https://github.com/",
```
bduffany, line 2498: Leaving the server port blank seems to work for me:

```
git clone ssh://git@github.com:/buildbuddy-io/buildbuddy /tmp/buildbuddy.git
```

Should we add a new entry instead of overwriting the existing one?

bduffany, line 2498:
```suggestion
		"ssh://git@github.com:/": "https://github.com/",
		"ssh://git@github.com/": "https://github.com/",
```

sluongng: Makes sense. I think it's relatively rare to specify a SSH port with `github.com` because it's always 22. But there's no harm in adding it.

bduffany (approved):

PR #7344, opened 2024-08-30
math/rand: replace rand.Read with crypto/rand

sluongng:
For more info: https://pkg.go.dev/math/rand#Read

```
Deprecated: For almost all use cases, crypto/rand.Read is more appropriate.
If a deterministic source is required, use math/rand/v2.ChaCha8.Read.
```

Replace it with crypto/rand.

### DIFF
--- a/enterprise/server/backends/pebble_cache/pebble_cache_test.go
+++ b/enterprise/server/backends/pebble_cache/pebble_cache_test.go
@@ -3,6 +3,7 @@ package pebble_cache_test
 import (
 	"bytes"
 	"context"
+	crand "crypto/rand"
 	"fmt"
 	"hash/crc32"
 	"io"
@@ -2348,7 +2349,7 @@ func TestMigrateVersions(t *testing.T) {
 
 func generateKMSKey(t *testing.T, kmsDir string, id string) string {
 	key := make([]byte, 32)
-	_, err := rand.Read(key)
+	_, err := crand.Read(key)
 	require.NoError(t, err)
 	err = os.WriteFile(filepath.Join(kmsDir, id), key, 0644)
 	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


sluongng:
based on https://github.com/buildbuddy-io/buildbuddy/pull/7343

enterprise/server/certgenerator/certgenerator.go
```diff
@@ -3,7 +3,7 @@ package main
 import (
 	"context"
 	"crypto/ecdsa"
-	"crypto/rand"
+	crand "crypto/rand"
```
tylerwilliams, line 6: I don't think this should be an aliased import -- we should not be importing math/rand in most of these files.

(If it did need to be aliased, the aliased imports are down below every other group of imports)

enterprise/server/backends/pebble_cache/pebble_cache_test.go
```diff
@@ -3,6 +3,7 @@ package pebble_cache_test
 import (
 	"bytes"
 	"context"
+	crand "crypto/rand"
```
vadimberezniker, line 6: please move this down below with the other aliased imports

vadimberezniker (approved):

enterprise/server/backends/pebble_cache/pebble_cache_test.go
tylerwilliams: is there a reason we need to have both rands in this file?

sluongng: Some tests use `math/rand.IntN()` which is cleaner with `math/rand`.

I plan to switch most of these to `math/rand/v2` in the future, but for now, I created a smaller helper func to use `crypto/rand` in this file instead.

PR #7346, opened 2024-08-30
Allow admins to provision user API keys for other group users

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -696,7 +697,7 @@ func (d *AuthDB) authorizeNewAPIKeyCapabilities(ctx context.Context, userID, gro
 	return d.authorizeGroupAdminRole(ctx, groupID)
 }
 
-func (d *AuthDB) CreateUserAPIKey(ctx context.Context, groupID, label string, caps []akpb.ApiKey_Capability) (*tables.APIKey, error) {
+func (d *AuthDB) CreateUserAPIKey(ctx context.Context, groupID, forUserID, label string, caps []akpb.ApiKey_Capability) (*tables.APIKey, error) {
 	if !*userOwnedKeysEnabled {
 		return nil, status.UnimplementedError("not implemented")
 	}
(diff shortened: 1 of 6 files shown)


server/interfaces/interfaces.go
```diff
@@ -479,7 +479,11 @@ type AuthDB interface {
 	GetUserAPIKeys(ctx context.Context, groupID string) ([]*tables.APIKey, error)
 
 	// CreateUserAPIKey creates a user-owned API key within the group.
-	CreateUserAPIKey(ctx context.Context, groupID, label string, capabilities []akpb.ApiKey_Capability) (*tables.APIKey, error)
+	// If "forUserID" is empty then the key is created for the authenticated
+	// user. Otherwise, the key will be owned by the given user ID, if they
```
tylerwilliams, line 483: given how few places this is called -- would it make sense to simplify the semantics and require forUserID always be set?

bduffany: I thought it was better to model this as an "optional parameter" - the UI doesn't set this param yet, so it would either add extra churn to make it "required" at the API layer, or require some backwards-compatibility logic in the API layer to default this param to the request_context.user_id if it's not set in the request.

The backwards compatibility logic seemed fine, but I also was thinking about how we wanted to move away from passing around groupID / userID as explicit params, instead move towards using request_context as the source of truth everywhere instead (it's not great that this method even takes a groupID). So with that mental model, I figured it was slightly better to not be setting the user ID "option" in most cases, and instead only set it in the less common case (when creating a key for another user).

tylerwilliams: I think what I would do is what you said: `default this param to the request_context.user_id if it's not set in the request` and fix up any call sites to explicitly set the param. Then in a week or two you can remove the line of code that defaults the value to request_context.user_id and everything is in a nice state.

The benefit is that CreateUserAPIKey doesn't have two codepaths (error prone! scary for an important api!) it only has one, and you can avoid code like this:

```
	if forUserID == "" {
		forUserID = u.GetUserID()
	}

	if forUserID == "" {
		return nil, status.InvalidArgumentError("could not determine user ID from request")
	}

	if forUserID != u.GetUserID() {
```

Sorry I know this is probably a little annoying, I'm pushing because I think we should strive to keep any auth / api key related code as straightforward as possible.

bduffany: Fair enough - I mostly just feel uneasy about proliferating auth-related params as raw strings rather than more trusted types like `interfaces.UserInfo`... but in this case we sort of have to deal with the raw string either way, so I'm fine with either approach :shrug:

tylerwilliams (approved):

enterprise/server/backends/authdb/authdb.go
```diff
@@ -713,7 +717,30 @@ func (d *AuthDB) CreateUserAPIKey(ctx context.Context, groupID, label string, ca
...
+		}
+		if !slices.Contains(caps, akpb.ApiKey_ORG_ADMIN_CAPABILITY) {
+			return nil, status.PermissionDeniedError("org admin permission is required to create an API key for the requested user")
+		}
+
+		// When creating a key for the non-authenticated user, validate that
+		// they are a member of the requested group.
+		ok, err := d.isGroupMember(ctx, groupID, userID)
+		if err != nil {
+			log.Errorf("Failed to query group memberships: %s", err)
```
vadimberezniker, line 735: warning maybe?

vadimberezniker (approved):

enterprise/server/backends/authdb/authdb.go
bduffany: done

PR #7347, opened 2024-08-30
Add some basic hit/miss metrics to cache proxy servers that do stuff.

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
+++ b/enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
@@ -6,9 +6,11 @@ import (
 	"io"
 
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
+	"github.com/buildbuddy-io/buildbuddy/server/metrics"
 	"github.com/buildbuddy-io/buildbuddy/server/real_environment"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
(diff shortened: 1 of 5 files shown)


enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -61,23 +65,29 @@ func (s *ByteStreamServerProxy) Read(req *bspb.ReadRequest, stream bspb.ByteStre
...
 			} else {
+				metrics.ByteStreamProxyReads.With(
+					prometheus.Labels{metrics.CacheHitMissStatus: "miss"}).Inc()
 				return s.readRemote(req, stream)
 			}
 		}
 
 		if err := stream.Send(rsp); err != nil {
+			metrics.ByteStreamProxyReads.With(
+				prometheus.Labels{metrics.CacheHitMissStatus: "hit"}).Inc()
```
tylerwilliams, line 84: i'm dumb -- why do want to record a hit if err != nil?

tylerwilliams (approved):

enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
iain-macdonald: this error would come from `stream.Send()` which should be a client-side error, so the entry was found in the local cache, but we just weren't able to send it back to the client for some reason (disconnect or something).

PR #7352, opened 2024-09-02
Don't use container runtimes bundled with podman-static release

bduffany:
This PR should fix the memory corruption bug ("corrupted size vs. prev_size") mentioned in https://github.com/buildbuddy-io/buildbuddy-internal/issues/3631

It turns out that `podman-static` includes its own versions of `crun` and `runc` under `/usr/local/bin`, which has higher precedence in PATH than the ones we provision under /usr/bin. So when upgrading to the latest `podman-static` release, we inadvertently upgraded `crun` to version 1.16.1, which has a [double-free bug](https://github.com/containers/crun/issues/1537). This was causing the "corrupted size vs. prev_size" errors we were seeing in some teed actions.

The double-free bug won't be fixed until 1.16.2, so this PR effectively just downgrades us to 1.15 until 1.16.2 is available.

### DIFF
--- a/enterprise/server/remote_execution/containers/podman/BUILD
+++ b/enterprise/server/remote_execution/containers/podman/BUILD
@@ -82,6 +82,11 @@ genrule(
         # Remove README.md (not needed)
         rm "$$TMPDIR/README.md"
 
+        # Remove container runtimes that are included in podman-static
+        # release, since we provision the container runtime separately.
+        rm "$$TMPDIR/usr/local/bin/crun"
+        rm "$$TMPDIR/usr/local/bin/runc"
+
         # Replace /etc/containers/registries.conf with our custom config.
         rm "$$TMPDIR/etc/containers/registries.conf"
         cp "$$REGISTRIES_CONF" "$$TMPDIR/etc/containers/registries.conf"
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/containers/podman/BUILD
```diff
@@ -82,6 +82,11 @@ genrule(
         # Remove README.md (not needed)
         rm "$$TMPDIR/README.md"
 
+        # Remove container runtimes that are included in podman-static
+        # release, since we provision the container runtime separately.
+        rm "$$TMPDIR/usr/local/bin/crun"
+        rm "$$TMPDIR/usr/local/bin/runc"
```
iain-macdonald, line 88: do you think it'd make sense to run a command that verifies there's no crun/runc after these are removed and fails if not, just in case an upstream change somehow reintroduces them?

iain-macdonald (approved):

enterprise/server/remote_execution/containers/podman/BUILD
bduffany: These `rm` commands will fail if the `crun` / `runc` binaries didn't actually exist at these paths. I think that should be good enough - wdyt?

vadimberezniker (approved):

enterprise/server/remote_execution/containers/podman/BUILD
iain-macdonald: sgtm

PR #7353, opened 2024-09-03
tsconfig: account for static builds

sluongng:
Ensure that tsconfig include the platform paths of --config=static and
--config=static-arm64.


bduffany (approved):
:face_with_spiral_eyes:

PR #7359, opened 2024-09-03
Implement container network pooling

bduffany:
* Adds a `ContainerNetworkPool` object that allows pooling networks for OCI containers only (*)
* When adding networks to the pool, we unassign the IP so it does not immediately get reused. This also deletes the default route.
* When taking networks from the pool, we assign a new IP as well as a new default route.
* When the executor shuts down, all pooled networks are cleaned up, and new networks may not be added to the pool.

Tested performance by running 1000 actions via 32 concurrently running `bb execute` commands, where each action just runs `ping -c1 8.8.8.8`

Results:
- Pooling disabled: 53s
- Pooling enabled: 6.9s

(*) It should be possible to also do this for VM networks, but it will take a bit of refactoring to do in a clean way, since Firecracker's setup and cleanup logic is very manual compared to the `ContainerNetwork` object that is used for OCI containers. I've started this refactor in https://github.com/buildbuddy-io/buildbuddy/pull/7360/files, with the eventual goal being that we have a common `VethNetwork` interface that can be managed using a single pool implementation.

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -19,6 +20,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/uuid"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/unix"
 )
 
@@ -53,6 +55,17 @@ const (
 
 	// CIDR matching all container networks on the host.
 	containerNetworkingCIDR = "192.168.0.0/16"
+)
+
+var (
+	// Default pool size limit to use when network pooling is enabled and the
+	// default size limit is requested.
+	//
+	// This value is big enough to allow an executor to burst from 0%
+	// utilization to 100% utilization while allowing all tasks to use a pooled
+	// network. (The number 4 is based on the current min CPU task size estimate
+	// of 250m)
+	defaultContainerNetworkPoolSizeLimit = runtime.NumCPU() * 4
 )
 
 // runCommand runs the provided command, prepending sudo if the calling user is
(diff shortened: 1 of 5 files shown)


server/util/networking/networking.go
```diff
@@ -421,12 +571,12 @@ func SetupVethPair(ctx context.Context, netNamespace string) (_ *VethPair, err e
...
 		// in place.
-		{"FORWARD", "-i", veth1, "-o", device, "-j", "ACCEPT"},
-		{"FORWARD", "-i", device, "-o", veth1, "-j", "ACCEPT"},
+		{"FORWARD", "-i", vp.hostDevice, "-o", device, "-j", "ACCEPT"},
+		{"FORWARD", "-i", device, "-o", vp.hostDevice, "-j", "ACCEPT"},
 
 		// Drop any traffic from the namespace that is targeting another
 		// namespace.
-		{"FORWARD", "-s", network.CIDR(), "-d", containerNetworkingCIDR, "-j", "DROP"},
+		{"FORWARD", "-i", vp.hostDevice, "-d", containerNetworkingCIDR, "-j", "DROP"},
```
bduffany, line 579: I updated this rule to use the source interface instead of the source IP so that we don't have to reconfigure this rule when we update the IP on the interface due to pooling. It still passes the network isolation tests in `networking_test.go`

tylerwilliams (approved):
This LGTM but you should probably have Vadim give it a once over too :)

server/util/networking/networking.go
```diff
@@ -213,6 +226,136 @@ func RemoveNetNamespace(ctx context.Context, netNamespace string) error {
...
+
+	head, tail := p.resources[0], p.resources[1:]
+	p.resources = tail
+	return head
+}
+
+// Add adds a veth pair to the pool.
+// It returns whether the veth pair was successfully added.
+// The caller should clean up the veth pair if this returns false.
+func (p *ContainerNetworkPool) Add(ctx context.Context, n *ContainerNetwork) (ok bool) {
```
iain-macdonald, line 317: nit: don't need a named return var

iain-macdonald (approved):

vadimberezniker (approved):

server/util/networking/networking.go
bduffany: When a func returns a non-obvious value such as `bool` I think using a named value helps for documentation purposes.

PR #7363, opened 2024-09-03
[ci_runner] Log end time of the script execution

maggie-lou:
Currently workflows don't log the git cleanup command to the UI, so it's difficult to determine how long it took to run and the actual end time of the execution 

How it will look after this change (Currently it only has the top "Running git maintenance..." line)

Screenshot 2024-09-04 at 12 29 32 PM

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -741,7 +741,7 @@ func run() error {
 	ws.prepareRunnerForNextInvocation(ctx, taskWorkspaceDir)
 
 	// Print an empty line to display the end time of the workflow
-	printCommandLine(ws.log, "")
+	ws.log.Printf("%sRemote run completed at %s%s", ansiGray, formatNowUTC(), ansiReset)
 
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -740,6 +740,9 @@ func run() error {
 
 	ws.prepareRunnerForNextInvocation(ctx, taskWorkspaceDir)
 
+	// Print an empty line to display the end time of the workflow
+	printCommandLine(ws.log, "")
```
bduffany, line 744: The blank shell prompt looks a little awkward to me - could we instead do a normal log here instead of calling `printCommandLine`, and write something like `Workflow completed in <duration>`?

maggie-lou: I included the timestamp instead of duration because the invocation link already includes a duration. I find the timestamp helpful because I can never remember if the duration on the invocation page includes queueing, setting up the runner etc

bduffany (approved):

PR #7364, opened 2024-09-03
2layer treecache

tylerwilliams:
Store tree cache blobs separately from pointers.

The blobs themselves are flattened lists of a directory and all it's children: these are stored in the cache. Separately, a pointer based only on the directory name and a seed is stored in the action cache: this points to the blob.

Also, update the tree cache seed.

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -235,6 +235,9 @@ const (
 	// the root dir, and so on.
 	TreeCacheLookupLevel = "level"
 
+	// TreeCache operation "read" or "write"
+	TreeCacheOperation = "op"
+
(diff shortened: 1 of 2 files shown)


siggisim (approved):
Makes me a little nervous to make every treecache read / write two calls instead of one, but I'm down to try this out.

tylerwilliams:
> Makes me a little nervous to make every treecache read / write two calls instead of one, but I'm down to try this out.

Same. Once inlining is out, we can use that and make it one again for small stuff.

jdhollen (approved):

jdhollen (approved):

PR #7365, opened 2024-09-03
[ci_runner] Rename the log that writes only to the executor logs to reduce ambiguity

maggie-lou:
If you want the ci_runner to write to the output that is visible to customers, you must use workspace.log. We frequently confused this with the logger that only writes to the backend executor logs

**Related issues**: N/A


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -701,9 +701,9 @@ func run() error {
 	}
 
 	if *shutdownAndExit {
```
bduffany, line 703: just as a note, in this shutdownAndExit case there is no BES stream to flush the logs to, so these logs will just sit buffered in memory - but there is only a small amount being buffered so I think it's fine to keep it like this for simplicity.

bduffany (approved):
thanks for doing this :pray:

PR #7372, opened 2024-09-04
Allow groups to set action priorities within their own queues

bduffany:
Allow users to pass `--remote_execution_priority` to set relative action priorities within their group. Per the remote API spec, a higher priority value means that the action should be executed later. For now, we limit values to be between -1000 and 1000.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/execution_server/execution_server.go
+++ b/enterprise/server/remote_execution/execution_server/execution_server.go
@@ -631,6 +631,7 @@ func (s *ExecutionServer) dispatch(ctx context.Context, req *repb.ExecuteRequest
 		PredictedTaskSize: predictedSize,
 		ExecutorGroupId:   pool.GroupID,
 		TaskGroupId:       taskGroupID,
+		Priority:          req.GetExecutionPolicy().GetPriority(),
 	}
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/platform/platform.go
```diff
@@ -71,6 +71,7 @@ const (
 
 	RecycleRunnerPropertyName            = "recycle-runner"
 	AffinityRoutingPropertyName          = "affinity-routing"
+	PriorityPropertyName                 = "priority"
```
tylerwilliams, line 74: can we use https://bazel.build/reference/command-line-reference#flag--remote_execution_priority instead of a platform prop for this?

bduffany: ah nice, didn't know about that.

Kind of ugly that it uses a backwards notion of priority (higher priority = executes later??) but it's probably better to stick with something that's already in the spec.

I also went ahead and limited the priorities so that they have to be between -1000 and 1000, just in case we want more flexibility later in how we interpret that field.

tylerwilliams (approved):

iain-macdonald (approved):

PR #7375, opened 2024-09-05
Don't start executor if expected Xcode versions are not present

iain-macdonald:
Today when we provision and start up mac executors we do a bunch of stuff to install the OS and Xcode and simulators, and then start up the executor on the machine. However, if any of the prerequisite steps, like installing the OS or Xcodes, fail then we just skip 'em and then the running executor won't be able to serve some RBE requests because it doesn't have all of the correct dependencies. This PR is the first of two that'll make starting the executor fail if any of the desired Xcodes aren't present on the host machine, which is better than quietly starting up and failing to serve RBE requests. It adds a new flag, `--executor.required_xcode_versions` that specifies a comma-delimited list of Xcode versions that must be present on the host system for the executor to come up healthy. There's another one coming that'll set that flag for our mac deployments.

**Related issues**: N/A

### DIFF
--- a/server/xcode/xcode_darwin.go
+++ b/server/xcode/xcode_darwin.go
@@ -10,6 +10,7 @@ package xcode
 import "C"
 
 import (
+	"flag"
 	"fmt"
 	"io/fs"
 	"os"
@@ -30,6 +31,8 @@ const developerDirectoryPath = "Contents/Developer"
 const filePrefix = "file://"
 const defaultXcodeVersion = "default-xcode-version"
 
+var requiredXcodeVersions = flag.String("executor.required_xcode_versions", "", "Comma-delimited list of Xcode versions required on the host system. If any of the provided Xcode versions cannot be found on the host, the executor will fail to start.")
+
 type xcodeLocator struct {
 	versions map[string]*xcodeVersion
 }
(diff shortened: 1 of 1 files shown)


server/xcode/xcode_darwin.go
```diff
@@ -49,9 +52,25 @@ type xcodePlist struct {
...
+// --executor.required_xcode_versions were found, killing the executor if not.
+func (x *xcodeLocator) verify() {
+	if *requiredXcodeVersions == "" {
+		return
+	}
+
+	requiredXcodes := strings.Split(*requiredXcodeVersions, ",")
+	for _, requiredXcode := range requiredXcodes {
+		if _, ok := x.versions[requiredXcode]; !ok {
+			log.Fatalf("Failed to locate required Xcode version %s", requiredXcode)
```
bduffany, line 69: In our mac setup, I think the executor will be automatically restarted if it gets killed. Might be worth doing a quick sanity check to make sure we don't start burning 100% CPU in this case (because we're repeatedly starting the executor and killing it in a tight loop)? If that does turn out to be the case then we might need to add a little delay here

iain-macdonald: That's my understanding as well. I think a better way to address that issue would be to `launchctl unload` the executor if it fails to come up healthy here: https://github.com/buildbuddy-io/buildbuddy-internal/blob/3e140a2cb644bd629d6ac84ae22f5b843aa17b80/enterprise/deployment/macstadium/roles/executor/handlers/main.yaml#L21

WDYT?

bduffany: I think I would prefer to have it crash-loop because that's more consistent with how it works on k8s (CrashLoopBackoff). I think we don't need the "backoff" part though and could probably just hard-code a 5s delay or something. Would be ideal if we also had a counter metric that we incremented on each crash loop.

This way we also don't have to do as much manual intervention - we can just install the necessary xcode version and it'll come back up on its own, rather than having to re-run the whole playbook.

Maybe @tylerwilliams has some thoughts here as well

iain-macdonald: That seems fine to me, I went ahead and added a `sleep()` before the `log.Fatal()` here.

In the playbooks we install Xcode and don't proceed to the next steps until we've verified the installation was successful, so the partial-install case you describe shouldn't happen.

bduffany: ugh my mistake, I think the delay should actually be handled at a higher level rather than doing it in the executor - maybe there's some retry delay parameter we can set in the launch agent XML?

iain-macdonald: Discussed offline, but adding here for more context. Per https://apple.stackexchange.com/questions/436219/launch-agent-launches-my-task-every-10-seconds-even-though-the-startinterval-fil this is already how launchctl behaves, so I don't think we need to do anything special here.

bduffany (approved):

PR #7377, opened 2024-09-05
Add darwin_arm64 workflow

bduffany:
**Related issues**: N/A


tylerwilliams (approved):

buildbuddy.yaml
```diff
@@ -60,6 +60,29 @@ actions:
...
+      pull_request:
+        branches:
+          - "*"
+    # TODO: Fix the tests below on Mac, and re-enable.
+    bazel_commands:
+      - >-
+        test --config=mac-workflows --test_tag_filters=-performance,-webdriver,-docker,-bare
+        --
+        //...
+        -//server/backends/disk_cache:all
```
tylerwilliams, line 79: Could we use a tag to exclude these so we don't need to update this workflow file or does that not work?

bduffany: I think that should work - will send a separate PR for that

iain-macdonald (approved):

PR #7378, opened 2024-09-05
split node_modules dirs into their own child treecaches

jdhollen:
**Related issues**: N/A


server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
```diff
@@ -456,12 +458,52 @@ func makeTreeCacheActionResult(blob *rspb.ResourceName) ([]byte, error) {
...
+			eg.Go(func() error {
+				childBuf, err := proto.Marshal(childCache)
+				if err != nil {
+					return err
+				}
+				childBlob, err := makeTreeCacheDigest(treeCachePointer.GetDigestFunction(), childBuf)
+				if err != nil {
+					return err
+				}
+				if err := s.cache.Set(egCtx, childBlob.ToProto(), childBuf); err != nil {
```
tylerwilliams, line 491: nit: wanna add a `if not s.cache.Contains()` checks before any CAS sets we do?

tylerwilliams (approved):
LGTM

(i think you're already doing this but making a test that triggers the split-out-dir case would be nice)

server/remote_cache/content_addressable_storage_server/content_addressable_storage_server.go
jdhollen: Yup, done

jdhollen:
> LGTM
> 
> (i think you're already doing this but making a test that triggers the split-out-dir case would be nice)

Added a test--the splitting itself is invisible at the call layer, but I've set the flags in the test case such that splitting happens.

PR #7380, opened 2024-09-05
close reader in fetch server

luluz66:
**Related issues**: N/A
There is a pebble_db_handle_leak alert this morning indicating Reader() is not
properly closed.


sluongng (approved):
LGTM. Good find!

PR #7390, opened 2024-09-06
actually wait for treecache writes to finish.

jdhollen:
by my read, we are basically always killing the trecache write content via a direct `defer cancelFunc()` call (we never actually make it to the block of code where we wait on the cache write error group)

this change should make it so that we return the rpc but hang onto the context for up to a second to finish cache writes.

**Related issues**: N/A

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -218,6 +218,9 @@ const (
 	// The TreeCache status: hit/miss/invalid_entry.
 	TreeCacheLookupStatus = "status"
 
+	// The TreeCache set status: success/deadline_exceeded/other_error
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
I think this looks right but sent it to vadim to double check

jdhollen:
This is makes the cas server tests time out, which probably isn't a good sign.  having a look

jdhollen:
Eh, adding the extra test yesterday made this flaky.  I'm sharding it for now, i'll make file creation a little faster and/or create fewer files to speed things up separately.

vadimberezniker (approved):

PR #7395, opened 2024-09-06
Only start active raft ranges

tylerwilliams:


luluz66:
The approach looks good to me in general. One question I have is that if it works with bootstrapping.

enterprise/server/raft/store/store.go
```diff
@@ -265,20 +265,41 @@ func NewWithArgs(env environment.Env, rootDir string, nodeHost *dragonboat.NodeH
...
+			previouslyStartedReplicas = append(previouslyStartedReplicas, replicaDescriptor)
+		}
+	}
+
+	ctx := context.Background()
+
+	// Scan the metarange and start any clusters we own that have not been
+	// removed. If previouslyStartedReplicas is an empty list, then
+	// LookupActiveReplicas will return nil, nil, and the following loop
+	// will be a no-op.
```
tylerwilliams, line 293: @luluz66 added a comment to explain how this works on bootstrap / bringup.

luluz66 (approved):

PR #7398, opened 2024-09-06
Fix refetching invocation with live-streaming invocation logs

maggie-lou:
Now that live-streaming invocation logs is enabled, we are constantly triggering `componentDidUpdate` which calls `scheduleRefetch`. That keeps calling `clearTimeout`, so `timeoutRef` the function to fetch the invocation never gets called

I noticed this because workflows should refetch their child invocations to display on the UI and I noticed they weren't updating

**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation.tsx
+++ b/app/invocation/invocation.tsx
@@ -90,8 +90,7 @@ export default class InvocationComponent extends React.Component<Props, State> {
     childInvocations: [],
   };
 
-  private timeoutRef: number = 0;
-  private refetchInProgress: boolean = false;
+  private timeoutRef: number | undefined;
   private logsModel?: InvocationLogsModel;
   private logsSubscription?: Subscription;
   private modelChangedSubscription?: Subscription;
@@ -185,7 +184,9 @@ export default class InvocationComponent extends React.Component<Props, State> {
   }
 
   componentWillUnmount() {
-    clearTimeout(this.timeoutRef);
+    if (this.timeoutRef) {
+      clearTimeout(this.timeoutRef);
+    }
     this.logsModel?.stopFetching();
     this.logsSubscription?.unsubscribe();
     this.runnerExecutionRPC?.cancel();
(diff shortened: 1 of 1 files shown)


app/invocation/invocation.tsx
```diff
@@ -262,16 +263,19 @@ export default class InvocationComponent extends React.Component<Props, State> {
...
+      this.refetchInProgress = true;
+      // Refetch invocation data in 3 seconds to update status.
+      this.timeoutRef = window.setTimeout(async () => {
+        // Before fetching the invocation, wait for the runner execution to be
+        // fetched, so we don't keep canceling the execution fetch if it takes
+        // longer than the invocation poll interval.
+        if (this.runnerExecutionRPC) await this.runnerExecutionRPC;
+
+        this.fetchInvocation();
+        this.refetchInProgress = false;
```
jdhollen, line 276: sort of on the flip side of this issue--it seems like we've always allowed multiple fetchInvocation() calls to spawn at the same time (if one takes a long time).  maybe we should be waiting until the invocation fetch is done before we set `refetchInProgress = false`?  i might be missing some quirky edge case where we're genuinely relying on this to force-update the UI, but the change in this PR would already break that (if it exists).

app/invocation/invocation.tsx
```diff
@@ -262,16 +263,19 @@ export default class InvocationComponent extends React.Component<Props, State> {
...
-    this.timeoutRef = window.setTimeout(async () => {
-      // Before fetching the invocation, wait for the runner execution to be
-      // fetched, so we don't keep canceling the execution fetch if it takes
-      // longer than the invocation poll interval.
-      if (this.runnerExecutionRPC) await this.runnerExecutionRPC;
-
-      this.fetchInvocation();
-    }, 3000);
+    if (!this.refetchInProgress) {
+      this.refetchInProgress = true;
```
bduffany, line 267: nit: to avoid introducing more component state (`this.refetchInProgress`) we could check whether there's an existing timeout and avoid starting a new one if one already exists

```
scheduleRefetch() {
  this.timeoutRef ??= setTimeout(async () => {
    /* ... */
    this.timeoutRef = null;
  }, 3000);
}
```

app/invocation/invocation.tsx
maggie-lou: I think I addressed this by adding `await` - could you take a look and see if that's what you meant?

jdhollen (approved):

app/invocation/invocation.tsx
bduffany: @maggie-lou The `await` here doesn't actually await anything because `fetchInvocation()` doesn't return a promise

maggie-lou: I'm going to be honest - I have a loose understanding of how this works, but my understanding is that the return type is inferred and `fetchInvocation` implicitly returns a promise

bduffany: `fetchInvocation` would implicitly return a promise if it were marked `async` but because it's not marked async, there's no implicit transform applied to the return value. In this case it's always returning `undefined` which is the default return value when nothing gets returned (i.e. no `return` statement)

maggie-lou: Would adding `async` fix this?

maggie-lou: https://github.com/buildbuddy-io/buildbuddy/pull/7430

I verified that the previous code didn't work, and that this new version does work, with some logging and a sleep

PR #7399, opened 2024-09-09
website: add support to run from within the directory

sluongng:
This change allows us to do

```
cd website && bazel run //website:start
```

instead of having to run the target from the workspace root dir.

Also tell git to ignore the left over exec log.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,4 +56,4 @@ buildbuddy-key.pem
 enterprise/config/user-configs
 
 # Bazel compact execution log
-/bazel_compact_exec_log.binpb.zst
+bazel_compact_exec_log.binpb.zst
--- a/rules/yarn/index.bzl
+++ b/rules/yarn/index.bzl
@@ -2,7 +2,7 @@ DEFAULT_CMD_TPL = """
 # NOTE: BazelBinResolverPlugin in docusaurus.config.js depends on ROOTDIR being set
 # to the original execution working directory.
 export ROOTDIR=$$(pwd) &&
-export PACKAGEDIR=$$(dirname $(location {package})) &&
+export PACKAGEDIR=\\$$BUILD_WORKSPACE_DIRECTORY/$$(dirname $(location {package})) &&
 export PATH=$$ROOTDIR/$$(dirname $(location {yarn})):$$ROOTDIR/$$(dirname $(location {node})):$$PATH &&
 cd $$PACKAGEDIR &&
 yarn install &&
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -56,4 +56,4 @@ buildbuddy-key.pem
 enterprise/config/user-configs
 
 # Bazel compact execution log
-/bazel_compact_exec_log.binpb.zst
+bazel_compact_exec_log.binpb.zst
```
bduffany, line 59: does it work to instead prefix the compact exec log file path with `%workspace%/`? (otherwise I think these files will keep building up as we run `bazel` within different dirs)

rules/yarn/index.bzl
```diff
@@ -2,7 +2,7 @@ DEFAULT_CMD_TPL = """
 # NOTE: BazelBinResolverPlugin in docusaurus.config.js depends on ROOTDIR being set
 # to the original execution working directory.
 export ROOTDIR=$$(pwd) &&
-export PACKAGEDIR=$$(dirname $(location {package})) &&
+export PACKAGEDIR=\\$$BUILD_WORKSPACE_DIRECTORY/$$(dirname $(location {package})) &&
```
bduffany, line 5: hm why is the `\\` needed? (here and below)

.gitignore
sluongng:
```
ERROR: /Users/sluongng/work/buildbuddy/buildbuddy/website/%workspace%/bazel_compact_exec_log.binpb.zst (No such file or directory)
ERROR: Error initializing execution log: /Users/sluongng/work/buildbuddy/buildbuddy/website/%workspace%/bazel_compact_exec_log.binpb.zst (No such file or directory)
```

Unfortunately, the flag is a bit dumb at the moment. 
Hopefully, upstream Bazel would make it write to output base automatically instead of us having to specify the exact location.

cc: @fmeum

rules/yarn/index.bzl
sluongng: without the back slashes, Bazel would fail with

```
/bin/bash: line 1: BUILD_WORKSPACE_DIRECTORY: unbound variable
```

when it tries to replace the variable with relevant Make variables.

Escaping excludes the variable from the substitution.

bduffany: any idea why `$$PATH` doesn't need the `\\` but `$$BUILD_WORKSPACE_DIRECTORY` does?

bduffany (approved):

rules/yarn/index.bzl
sluongng:
```
> cat   bazel-bin/website/start.sh

export PATH=/private/var/tmp/_bazel_sluongng/06e573a93bc2d6a9cad4ad41f00b4310/execroot/buildbuddy/external/yarn/bin/yarnpkg/bin:/private/var/tmp/_bazel_sluongng/06e573a93bc2d6a9cad4ad41f00b4310/execroot/buildbuddy/external/nodejs_darwin_arm64/bin/nodejs/bin:/bin:/usr/bin:/usr/local/bin &&
cd $BUILD_WORKSPACE_DIRECTORY/website &&
yarn install &&
yarn start
```

This is the output in `bazel-bin/website/start.sh` on my laptop.
The `$PATH` env var was replaced by Bazel's Makevars while the `$BUILD_WORKSPACE_DIRECTORY` is kept as a bash variable.

So it worked because PATH is a legal makevars 😅

PR #7402, opened 2024-09-09
[RB] Show full child command in UI

maggie-lou:
Currently, the UI will only show the command and the pattern - i.e. `build //...`. Show the full explicit command line to include all flags specified by the user. This will bring the UI closer to how it looked before the bash steps refactor

Note that this won't apply automatically for workflows - some more refactoring is required to consolidate the deprecated `BazelCommand` syntax with the new `Steps` syntax

Before this PR:
Screenshot 2024-10-17 at 11 42 49 AM

After:
Screenshot 2024-10-17 at 11 42 33 AM

**Related issues**: N/A

### DIFF
--- a/app/invocation/child_invocation_card.tsx
+++ b/app/invocation/child_invocation_card.tsx
@@ -4,6 +4,7 @@ import { invocation } from "../../proto/invocation_ts_proto";
 import { CheckCircle, PlayCircle, XCircle, CircleSlash } from "lucide-react";
 import Link from "../components/link/link";
 import { invocation_status } from "../../proto/invocation_status_ts_proto";
+import InvocationModel from "./invocation_model";
 
 type CommandStatus = "failed" | "succeeded" | "in-progress" | "not-run";
 
@@ -52,14 +53,38 @@ export default class ChildInvocationCard extends React.Component<ChildInvocation
     }
   }
 
+  private simplifyCommandLine(commandLine: string): string {
+    // Remove the bazel command string
+    if (commandLine.startsWith("bazel ")) {
+      commandLine = commandLine.slice("bazel ".length);
+    }
+
+    // Strip all --remote_header flags
+    commandLine = commandLine.replace(/\s*--remote_header=[^\s]+/g, "");
+    // Strip the options we manually added
+    commandLine = commandLine.replace(" --config=buildbuddy_bes_backend", "");
+    commandLine = commandLine.replace(" --config=buildbuddy_bes_results_url", "");
+    commandLine = commandLine.replace(" --config=buildbuddy_remote_cache", "");
+    commandLine = commandLine.replace(/\s*--invocation_id=[^\s]+/g, "");
+
+    return commandLine;
+  }
+
   render() {
-    const status = this.getStatus();
     const inv = this.props.invocation;
-    const command = `${inv.command} ${inv.pattern.join(" ")}`;
+    const invModel = new InvocationModel(inv);
+
+    const status = this.getStatus();
+    let command = invModel.explicitCommandLine();
+    if (command == "") {
+      command = `${inv.command} ${inv.pattern.join(" ")}`;
+    }
+    command = this.simplifyCommandLine(command);
+
     return (
       <Link
         className={`child-invocation-card status-${status} ${this.isClickable(status) ? "clickable" : ""}`}
-        href={this.isClickable(status) ? `/invocation/${this.props.invocation.invocationId}` : undefined}>
+        href={this.isClickable(status) ? `/invocation/${inv.invocationId}` : undefined}>
         <div className="icon-container">{this.renderStatusIcon(status)}</div>
         <div className="command">{command}</div>
         <div className="duration">{this.getDurationLabel(status)}</div>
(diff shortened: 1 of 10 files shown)


app/invocation/child_invocation_card.tsx
```diff
@@ -52,14 +53,38 @@ export default class ChildInvocationCard extends React.Component<ChildInvocation
...
+
+    // Strip all --remote_header flags
+    commandLine = commandLine.replace(/\s*--remote_header=[^\s]+/g, "");
+    // Strip the options we manually added
+    commandLine = commandLine.replace(" --config=buildbuddy_bes_backend", "");
+    commandLine = commandLine.replace(" --config=buildbuddy_bes_results_url", "");
+    commandLine = commandLine.replace(" --config=buildbuddy_remote_cache", "");
+    commandLine = commandLine.replace(/\s*--invocation_id=[^\s]+/g, "");
+
+    return commandLine;
```
bduffany, line 70: We have `EXPLICIT_COMMAND_LINE` that is intended as a way to override the bazel command displayed in the UI in cases where there is some extra arg fiddling going on (e.g. the CLI) - can we set that build metadata and use it directly instead of "undoing" the arg fiddling here?

maggie-lou: I realized that the redactor strips EXPLICIT_COMMAND_LINE from the build metadata, so that is actually an ineffective way to override the bazel command displayed in the UI (Here's a [PR](https://github.com/buildbuddy-io/buildbuddy/pull/7734) to remove the code from the CLI. Alternatively, we could add it back and risk there being places we are incorrectly redacting it, which was the [original impetus to remove it](https://github.com/buildbuddy-io/buildbuddy/pull/4525))

bduffany: > the [redactor strips EXPLICIT_COMMAND_LINE from the build metadata](https://github.com/buildbuddy-io/buildbuddy/blob/master/server/util/redact/redact.go#L196)

I think the linked code is redacting the CommandLine event, not the BuildMetadata event?

bduffany: after a closer look, that linked code is stripping any nested occurrences of `--build_metadata=EXPLICIT_COMMAND_LINE=` within the explicit command line (which shouldn't normally happen), but not the EXPLICIT_COMMAND_LINE build metadata is still preserved in the `BuildMetadata` event

maggie-lou:
> It's a bit unfortunate that we have to load all of the build events for each child invocation to be able to show this, but I think there's probably not a better alternative atm 🤔

Yeah... The only other option I can think of is parsing the bash code which sounds like a nightmare and very error prone

We could probably be a little more clever in optimizing this. i.e. If we've already fetched a child invocation before, we could make sure to not fetch it again (we already do this to some extent, but could be even stricter). Or we could set the explicit command line on the invocation object in the BE, so we don't need to return all the build events for the children to the FE client. Is there a primary concern you're nervous about?

server/backends/invocationdb/invocationdb.go
```diff
@@ -181,10 +182,22 @@ func (d *InvocationDB) LookupInvocation(ctx context.Context, invocationID string
...
 }
 
-func (d *InvocationDB) LookupChildInvocations(ctx context.Context, parentRunID string) ([]*tables.Invocation, error) {
+func (d *InvocationDB) LookupChildInvocations(ctx context.Context, parentRunID string) ([]string, error) {
 	rq := d.h.NewQuery(ctx, "invocationdb_get_child_invocations").Raw(
-		`SELECT * FROM "Invocations" WHERE parent_run_id = ? ORDER BY created_at_usec`, parentRunID)
-	return db.ScanAll(rq, &tables.Invocation{})
+		`SELECT invocation_id FROM "Invocations" WHERE parent_run_id = ? ORDER BY created_at_usec`, parentRunID)
+	iids := make([]string, 0)
+	err := rq.IterateRaw(func(ctx context.Context, row *sql.Rows) error {
```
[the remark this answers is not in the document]
maggie-lou: I used `sql.Rows` here because we're only reading the invocationID and it seemed simpler to read a single field with it

bduffany:
@maggie-lou I think it's fine since we're only returning the top-level events to the client - loading all the events from blobstore can just be a little slow for larger invocations. We could maybe store the explicit command line in a better way, but for now it's probably fine

server/backends/invocationdb/invocationdb.go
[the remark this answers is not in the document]
bduffany: The resulting code will actually be simpler using `ScanEach`:

```
db.ScanEach(rq, func(ctx context.Context, inv *tables.Invocation) error {
  iids = append(iids, inv.InvocationID)
  return nil
})
```

not a big deal either way, just figured I'd point out that this is a tool that exists.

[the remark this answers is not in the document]
maggie-lou: Oh I see - thanks! That is simpler. I was under the assumption that the struct passed in to ScanEach should only have the fields fetched from the db

PR #7405, opened 2024-09-09
Add basic file-based logging for OCI containers

bduffany:
Writes `Run()` logs to `$BUNDLE_PATH/logs/{stdout,stderr}` and `Exec()` logs to `$BUNDLE_PATH/logs/exec-$N.{stdout,stderr}` where `$BUNDLE_PATH` is where we write other data for the container (OCI runtime spec, rootfs mount, etc.) and `$N` is a counter that increments with each `Exec()` call, starting from 1

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3752


bduffany:
Discussed offline with @tylerwilliams - will close this in favor of a different approach that doesn't involve buffering logs in memory, and works across all container types

PR #7408, opened 2024-09-09
Break out compact exec log spawns and files into their own tabs

siggisim:
This breaks compact execution log spawns into their own tab called `Spawns` (so you don't have to scroll on the `Executions` tab to find remote actions).

Also adds a new tab called `Files` when the compact execution log is enabled. This allows you to see and search for all input / output files in the execution log.

These tabs are both pretty basic for now, but we can make them fancier in subsequent PRs.


bduffany (approved):
I feel like spawns/executions and files/artifacts could get easily confused, and also that we're starting to have too many tabs on the invocation page, but can't really think of any better alternatives atm - will probably be easier to merge then tweak as we test this out ourselves / get more feedback from users

PR #7409, opened 2024-09-09
Fix flaky webdriver test

maggie-lou:
The webdriver test does a sample build and pulls the api key from `docs/setup`. The api key takes a moment to populate so it was fetching an empty api key, and the test was failing because it was trying to do an unauthenticated build.

**Related issues**: N/A

### DIFF
--- a/server/testutil/webtester/webtester.go
+++ b/server/testutil/webtester/webtester.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"regexp"
 	"runtime"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


server/testutil/webtester/webtester.go
```diff
@@ -440,7 +441,26 @@ func GetBazelBuildFlags(wt *WebTester, appBaseURL string, opts ...SetupPageOptio
 	for _, fn := range opts {
 		fn(wt)
 	}
-	rawBazelrc := wt.Find(`[data-header=".bazelrc"] .contents`).Text()
+
+	refetch := true
+	attempts := 1
+	var rawBazelrc string
+	for refetch && attempts <= 5 {
+		rawBazelrc = wt.Find(`[data-header=".bazelrc"] .contents`).Text()
```
jdhollen, line 449: Since this is actually an async call on the page, you can use the builtin `wt.driver.WaitWithTimeout` to wait for this element to exist

server/testutil/webtester/webtester.go
```diff
@@ -440,7 +441,26 @@ func GetBazelBuildFlags(wt *WebTester, appBaseURL string, opts ...SetupPageOptio
...
+
+		// API key takes a moment to populate on the page, so keep reloading the text
+		// until it is fetched.
+		re := regexp.MustCompile(`build --remote_header=x-buildbuddy-api-key=(\S+)`)
```
bduffany, line 453: I think this test flakiness is actually surfacing a legitimate bug which is that we're not displaying a loading spinner here. I call it a "bug" rather than a UX issue because if the user is on a slow/flaky network connection, this content will never be displayed to the user, and they'll have no indication that it should be displayed. If we showed a loading spinner, it would indicate that we're blocked on a network call somehow. In the test, we could just wait for the loading spinner to go away, rather than doing a regex match.

maggie-lou: Ack - I'm going to prioritize some bazelcon related features for now, so created a ticket to track this https://github.com/buildbuddy-io/buildbuddy-internal/issues/3760

This flaky test somewhat frequently impedes the release. @jdhollen @bduffany how would you feel about merging this as is for now given that it's just a test?

bduffany: @maggie-lou I can send a fix for this today if you want to re-assign that issue to me

maggie-lou: 🙏

bduffany:
Fixed in https://github.com/buildbuddy-io/buildbuddy/pull/7435

PR #7412, opened 2024-09-10
fetch_server: add support for bazel qualifiers

sluongng:
In addition to the checksum.sri qualifier from remote api spec,
Bazel includes it's own set of qualifiers to each FetchBlob requests.

https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java;l=76-85;drc=618c0abbfe518f4e29de523a2e63ca9179050e94

This change adds initial support for the http_header qualifier as well
as acknoledge the existence of bazel.canonical_id header (without
actually using it). Relevant tests were also added to demonstrate the
current behavior.

In a future patch, we may start rejecting unknown qualifiers as
recommended by the Remote Asset API spec and thus, flip the assertions
in the test. See https://github.com/bazelbuild/remote-apis/issues/301
for more info.

Another change to consider in the future is support for custom
url-specific header credentials. This is being proposed in Bazel via
https://github.com/bazelbuild/bazel/pull/23578.

### DIFF
--- a/server/remote_asset/fetch_server/fetch_server.go
+++ b/server/remote_asset/fetch_server/fetch_server.go
@@ -31,8 +31,10 @@ import (
 )
 
 const (
-	ChecksumQualifier = "checksum.sri"
-	maxHTTPTimeout    = 60 * time.Minute
+	ChecksumQualifier              = "checksum.sri"
+	BazelCanonicalIDQualifier      = "bazel.canonical_id"
+	BazelHttpHeaderPrefixQualifier = "http_header:"
+	maxHTTPTimeout                 = 60 * time.Minute
 )
 
 type FetchServer struct {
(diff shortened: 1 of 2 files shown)


server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -31,8 +31,10 @@ import (
 )
 
 const (
-	ChecksumQualifier = "checksum.sri"
-	maxHTTPTimeout    = 60 * time.Minute
+	ChecksumQualifier              = "checksum.sri"
+	QualifierBazelCanonicalID      = "bazel.canonical_id"
+	QualifierBazelHttpHeaderPrefix = "http_header:"
```
bduffany, line 36: nit, for consistency w/ the existing code, put the `Qualifier` towards the end of these names? (or rename `ChecksumQualifier` => `QualifierChecksum` but imo that doesn't read as nicely)

```suggestion
	ChecksumQualifier              = "checksum.sri"
	BazelCanonicalIDQualifier      = "bazel.canonical_id"
	BazelHttpHeaderQualifierPrefix = "http_header:"
```

server/remote_asset/fetch_server/fetch_server.go
```diff
@@ -100,6 +102,22 @@ func timeoutHTTPClient(ctx context.Context, protoTimeout *durationpb.Duration) *
 	}
 }
 
+func handleCheckSumQualifier(qualifier *rapb.Qualifier) (repb.DigestFunction_Value, string, error) {
```
bduffany, line 105:
```suggestion
// parseChecksumQualifier returns a digest function and digest hash
// given a "checksum.sri" qualifier.
func parseChecksumQualifier(qualifier *rapb.Qualifier) (repb.DigestFunction_Value, string, error) {
```

bduffany (approved):
nice :+1:

PR #7414, opened 2024-09-10
Schema changes for serving Executions page using ClickHouse

bduffany:
This PR adds the missing pieces of data needed to render the Executions page via ClickHouse. Updates both the ClickHouse table schema as well as its proto representation (`StoredExecution`).

**Related issues**: N/A

### DIFF
--- a/server/util/clickhouse/schema/schema.go
+++ b/server/util/clickhouse/schema/schema.go
@@ -192,6 +192,10 @@ type Execution struct {
 	CachedResult bool
 	DoNotCache   bool
 
+	// Long string fields
+	CommandSnippet string
+	StatusMessage  string
+
 	// Fields from Invocations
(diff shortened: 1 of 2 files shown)


server/util/clickhouse/schema/schema.go
```diff
@@ -192,6 +192,10 @@ type Execution struct {
 	CachedResult bool
 	DoNotCache   bool
 
+	// Long string fields
+	CommandSnippet string
```
luluz66, line 196: need to remove these from ExcludedFields()

bduffany: done

luluz66 (approved):

bduffany:
Updated with `output_path` per offline discussion.

PR #7416, opened 2024-09-10
[ci_runner] Redact all logs

maggie-lou:
**Related issues**: https://buildbuddy-corp.slack.com/archives/C019DS52XAR/p1725982537416079?thread_ts=1710341219.838939&cid=C019DS52XAR

### DIFF
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -946,7 +946,7 @@ func newInvocationLog() *invocationLog {
 func (invLog *invocationLog) Write(b []byte) (int, error) {
 	output := string(b)
 
-	redacted := redact.RedactTxt(output)
+	redacted := redact.RedactText(output)
 
 	invLog.writeListener(redacted)
 	_, err := invLog.writer.Write([]byte(redacted))
--- a/enterprise/server/cmd/ci_runner/main.go
+++ b/enterprise/server/cmd/ci_runner/main.go
@@ -946,13 +946,10 @@ func newInvocationLog() *invocationLog {
 func (invLog *invocationLog) Write(b []byte) (int, error) {
 	output := string(b)
 
-	redacted, err := redact.RedactCommand(output)
-	if err != nil {
-		return 0, err
-	}
+	redacted := redact.RedactTxt(output)
 
 	invLog.writeListener(redacted)
-	_, err = invLog.writer.Write([]byte(redacted))
+	_, err := invLog.writer.Write([]byte(redacted))
 
 	// Return the size of the original buffer even if a redacted size was written,
 	// or clients will return a short write error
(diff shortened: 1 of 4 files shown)


enterprise/server/cmd/ci_runner/main.go
```diff
@@ -936,19 +934,29 @@ func (r *buildEventReporter) Printf(format string, vals ...interface{}) {
...
 	return invLog
 }
 
 func (invLog *invocationLog) Write(b []byte) (int, error) {
-	invLog.writeListener(b)
-	n, err := invLog.writer.Write(b)
-	return n, err
+	output := string(b)
+
+	redacted, err := redact.RedactCommand(output)
```
bduffany, line 949: `RedactCommand` parses all input using `shlex`, but invocations contain a lot more output than just commands. Are there specific regex-replace rules we could apply here instead?

maggie-lou: I considered that too, but thought I'd just stick with the existing logic for simplicity. Are you concerned about the performance implications?

bduffany: `shlex.Split()` doesn't handle arbitrary text - it's a shell tokenizer. For example, `shlex.Split("Hello 'w o r l d'")` returns `[]string{"Hello", "w o r l d"}`. `shlex.Split(" ' ")` fails because it's not valid shell code (there's an unterminated closing quote).

maggie-lou: Refactored

server/util/redact/redact.go
```diff
@@ -206,6 +207,41 @@ func redactCmdLine(tokens []string) {
 	stripExplicitCommandLineFromCmdLine(tokens)
 }
 
+func RedactTxt(txt string) string {
+	redacted := stripURLSecrets(txt)
+	redacted = redactRemoteHeaders(redacted)
+	redacted = redactAPIKeys(redacted)
+	return redacted
```
bduffany, line 214: uber nit: it feels slightly safer to process the `txt` variable directly, to avoid accidents like `redacted = redactRemoveHeaders(txt)` which redacts the original string instead of the intermediate redacted result

```suggestion
	txt = stripURLSecrets(txt)
	txt = redactRemoteHeaders(txt)
	txt = redactAPIKeys(txt)
	return redacted
```

server/util/redact/redact.go
```diff
@@ -206,6 +207,41 @@ func redactCmdLine(tokens []string) {
 	stripExplicitCommandLineFromCmdLine(tokens)
 }
 
+func RedactTxt(txt string) string {
```
bduffany, line 210: nit: avoid abbreviation since it's an exported func

```suggestion
func RedactText(txt string) string {
```

server/util/redact/redact.go
```diff
@@ -679,27 +718,10 @@ func (r *StreamingRedactor) RedactAPIKeysWithSlowRegexp(ctx context.Context, eve
...
-
-	txt = []byte(apiKeyAtPattern.ReplaceAllString(string(txt), "$1<REDACTED>@"))
-
-	// Replace the literal API key set up via the BuildBuddy config, which does not
-	// need to conform to the way we generate API keys.
-	if configuredKey := *apiKey; configuredKey != "" {
-		txt = []byte(strings.ReplaceAll(string(txt), configuredKey, "<REDACTED>"))
-	}
-
+	redacted := redactAPIKeys(string(txt))
```
bduffany, line 721: same comment here, feels safer to use `txt = ` here instead of introducing a `redacted` variable

bduffany (approved):

PR #7424, opened 2024-09-11
Cache Proxy: Add a couple of metrics tracking atime_updater performance

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/atime_updater/atime_updater.go
+++ b/enterprise/server/atime_updater/atime_updater.go
@@ -132,12 +132,7 @@ func (u *atimeUpdater) Enqueue(ctx context.Context, instanceName string, digests
 	// requested updates. Too bad. Keep track of the counts for metrics below.
 	keys := map[digest.Key]int{}
 	for _, digestProto := range digests {
-		key := digest.NewKey(digestProto)
-		if count, ok := keys[key]; ok {
-			keys[key] = count + 1
-		} else {
-			keys[key] = 1
-		}
+		keys[digest.NewKey(digestProto)]++
 	}
 
 	// TODO(iain): this locking may be a bit overzealous. Hopefully not though.
--- a/enterprise/server/atime_updater/atime_updater.go
+++ b/enterprise/server/atime_updater/atime_updater.go
@@ -8,11 +8,14 @@ import (
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/interfaces"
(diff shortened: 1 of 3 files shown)


enterprise/server/atime_updater/atime_updater.go
```diff
@@ -126,10 +129,15 @@ func (u *atimeUpdater) Enqueue(ctx context.Context, instanceName string, digests
...
 
 	// Uniqueify the incoming digests. Note this wrecks the order of the
-	// requested updates. Too bad.
-	keys := map[digest.Key]struct{}{}
+	// requested updates. Too bad. Keep track of the counts for metrics below.
+	keys := map[digest.Key]int{}
 	for _, digestProto := range digests {
-		keys[digest.NewKey(digestProto)] = struct{}{}
+		key := digest.NewKey(digestProto)
+		if count, ok := keys[key]; ok {
```
vadimberezniker, line 136: You can replace this code with `keys[key]++`

vadimberezniker: If a key doesn't exist, you get the zero value back which is actually `0` in this case since the value is an int.

enterprise/server/atime_updater/atime_updater.go
```diff
@@ -162,13 +175,47 @@ func (u *atimeUpdater) Enqueue(ctx context.Context, instanceName string, digests
...
+			dropped += count
+			continue
+		}
+		if _, ok := pendingUpdate.digests[key]; ok {
+			duplicate += count
+			continue
 		}
 		pendingUpdate.digests[key] = struct{}{}
+		enqueued++
+		duplicate += count - 1
```
vadimberezniker, line 197: why is this considered a duplicate?

vadimberezniker (approved):

enterprise/server/atime_updater/atime_updater.go
iain-macdonald: Done

enterprise/server/atime_updater/atime_updater.go
iain-macdonald: The first one is enqueued and any others are considered duplicates. Agree that it's slightly odd that the "too_many_batches" above is recorded once per instance, but I figured that measuring duplicates as those that actually make it into the queue was reasonable. We can always change it if we need a slightly different definition in the future.

PR #7425, opened 2024-09-11
Consolidate / clean up VM networking logic

bduffany:
- Move namespace, veth pair, and tap device setup/cleanup out of firecracker.go and into a new `VMNetwork` struct, similar to the current `ContainerNetwork` struct.
- This sets us up for VM network pooling: `VMNetwork` and `ContainerNetwork` can both implement a common interface like `Poolable { BeforeAddToPool(), BeforeTakeFromPool() }`. `NetworkPool` can then hold a generic list of `Poolable` rather than `ContainerNetwork`.

**Related issues**: N/A


server/util/networking/networking.go
```diff
@@ -642,31 +590,92 @@ func (s cleanupStack) Cleanup(ctx context.Context) error {
...
+		// If we failed to fully set up the network, make sure to clean up any
+		// resources that were partially set up.
+		if err != nil {
+			_ = cleanupStack.Cleanup(ctx)
+		}
+	}()
+
+	// Create a net namespace.
+	netns, err := createUniqueNetNamespace(ctx)
+	if err != nil {
```
maggie-lou, line 618: Previously we only warned if we got a "file exists" error here - was it intentional to remove that?

server/util/networking/networking.go
```diff
@@ -642,31 +590,92 @@ func (s cleanupStack) Cleanup(ctx context.Context) error {
...
+		return nil, status.WrapError(err, "create net namespace")
 	}
-	if err := runCommand(ctx, namespace(vethPair.netns, "iptables", "--wait", "-t", "nat", "-A", "PREROUTING", "-i", vethPair.namespacedDevice, "-d", vethPair.network.NamespacedIP(), "-j", "DNAT", "--to", vmIP)...); err != nil {
-		return err
+	cleanupStack = append(cleanupStack, func(ctx context.Context) error {
+		return netns.Delete(ctx)
+	})
+
+	// Create a veth pair with one end in the namespace.
+	vethPair, err := setupVethPair(ctx, netns)
```
maggie-lou, line 626: I'm not very familiar with the setup - just basing off the old code. Does it matter that we're now doing this before creating the tap device?

bduffany: Nope, the order of creating the interfaces doesn't matter - the tap / veth devices are independent except for the NAT rule that translates the tap IP to/from the namespaced veth IP, and that NAT rule setup will succeed as long as the namespaced veth device exists when we run it

server/util/networking/networking.go
bduffany: I removed that code because that error doesn't get logged in practice, which makes sense because net namespace names are generated using a uuid - so it shouldn't be possible for a namespace to already exist when we create it

vadimberezniker (approved):

PR #7426, opened 2024-09-11
Reapply "website: add support to run from within the directory (#7399)" (#7421)

sluongng:
Instead of using the variable directly, use a proxy `WS_DIR` variable so
that if BUILD_WORKSPACE_DIRECTORY is unset, we just gona pass and empty
string to `cd`.

Also replaced all the `&&` with `set -e` to simplify the script.

### DIFF
--- a/rules/yarn/index.bzl
+++ b/rules/yarn/index.bzl
@@ -1,23 +1,31 @@
 DEFAULT_CMD_TPL = """
+set -e
+
 # NOTE: BazelBinResolverPlugin in docusaurus.config.js depends on ROOTDIR being set
 # to the original execution working directory.
-export ROOTDIR=$$(pwd) &&
-export PACKAGEDIR=$$(dirname $(location {package})) &&
-export PATH=$$ROOTDIR/$$(dirname $(location {yarn})):$$ROOTDIR/$$(dirname $(location {node})):$$PATH &&
-cd $$PACKAGEDIR &&
-yarn install &&
-yarn {command} &&
-cd build &&
-tar -cvf ../build.tar * &&
-cd $$ROOTDIR &&
+export ROOTDIR=$$(pwd)
+export PACKAGEDIR=$$(dirname $(location {package}))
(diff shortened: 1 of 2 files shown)


rules/yarn/index.bzl
```diff
@@ -1,23 +1,31 @@
...
-export PATH=$$(pwd)/$$(dirname $(location {yarn})):$$(pwd)/$$(dirname $(location {node})):$$PATH &&
-cd $$(dirname $(location {package})) &&
-yarn install &&
+set -e
+export PATH=$$(pwd)/$$(dirname $(location {yarn})):$$(pwd)/$$(dirname $(location {node})):$$PATH
+if [ -n "\\$$BUILD_WORKSPACE_DIRECTORY" ]; then
+    WS_DIR="\\$$BUILD_WORKSPACE_DIRECTORY/"
+else
+    WS_DIR=""
+fi
```
bduffany, line 26: any idea why BUILD_WORKSPACE_DIRECTORY is sometimes defined and other times isn't? I'm a little hesitant to roll this forward without understanding that piece of it. (If it's taking too much time to figure out, I think it's also fine to save this change for a later date - running bazel from within a subdir seems like a relatively rare use case)

sluongng: So if you have an `sh_binary(name="foo")` and you do `bazel run //:foo` then Bazel would set that environment variable before executing `foo.sh`.

However, if there is another target depending on foo and uses it at build time, let's say `genrule(name="bar", srcs=["//:foo"])`. Then `bazel  build //:bar` would execute foo.sh inside a sandbox with hermetic environment variables (thus, no `BUILD_WORKSPACE_DIRECTORY` is set).

It seems like currently, we use `DEFAULT_CMD_TPL` script in build sandbox, while `EXECUTABLE_CMD_TPL` is for development with `bazel run`. So I think it's relatively safe to set it exclusively for `EXECUTABLE_CMD_TPL`

PR #7429, opened 2024-09-11
Make firecracker CoW tests run on CI

bduffany:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
@@ -706,6 +706,7 @@ func TestFirecracker_RemoteSnapshotSharing(t *testing.T) {
 		},
 		ExecutorConfig: cfg,
 	}
+	instanceName := "test-instance-name"
 	task := &repb.ExecutionTask{
 		Command: &repb.Command{
 			// Note: platform must match in order to share snapshots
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker_test.go
```diff
@@ -798,10 +802,14 @@ func TestFirecracker_RemoteSnapshotSharing(t *testing.T) {
...
 
 	// Should still be able to start from the original snapshot if we use
-	// a snapshot key containing the original VM's snapshot ID
+	// a snapshot key containing the original VM's snapshot ID.
+	// Note that when using a snapshot ID as the key, we only include the
+	// snapshot_id and instance_name fields.
 	workDirForkOriginalSnapshot := testfs.MakeDirAll(t, rootDir, "work-fork-og-snapshot")
-	originalSnapshotKey := baseVM.SnapshotKeySet().GetBranchKey()
-	originalSnapshotKey.SnapshotId = baseSnapshotId
+	originalSnapshotKey := &fcpb.SnapshotKey{
```
bduffany, line 809: There were 2 issues here:
- We weren't doing anything with `originalSnapshotKey` - from the comment we should be setting it in ContainerOpts so that we resume from this specific snapshot
- I forgot to update this logic here when we simplified snapshot-specific keys so that they only include the instance_name + snapshot_id.

bduffany:
yikes, the tests are running extremely slowly. going to start on a separate PR to do the file-locking thing with the IP addresses, so that we can shard this test.

maggie-lou (approved):

PR #7438, opened 2024-09-12
change default lookaside_cache_ttl to 1 minute

luluz66:
**Related issues**: N/A
The default of `cache.distributed_cache.lookaside_cache_ttl` is 15 minutes, and
it's supposed to be << `atime_update_threshold`. but the default of 
`cache.pebble.atime_update_threshold` is 10 minutes.


tylerwilliams (approved):

luluz66:
Changing the default of `cache.distributed_cache.lookaside_cache_ttl` instead since lookaside_cache by default is turned off.

PR #7439, opened 2024-09-12
Allow multiple firecracker test shards to run concurrently on a single machine

bduffany:
Follow-up from https://github.com/buildbuddy-io/buildbuddy/pull/7429

* Add optional `flock`-based IP range locking to prevent concurrent test shards from trying to use the same IP range concurrently.
  * This is useful for firecracker tests, as well as running multiple firecracker executors locally. 
* Don't try to clean up network namespaces in tests, because they might be in use.
   * I don't expect that too many of these will accumulate, but if it does happen, we can find a way to clean up "orphaned" namespaces, maybe by associating them with lockfiles.
* Don't try to clean up the shared executor root directory in tests either, to prevent converted ext4 image files from disappearing mid-test.
  * Added a TODO to provision these images as data deps of the test instead of converting during the test.
* Set `shard_count` on `firecracker_test`. (This won't actually buy us a ton of concurrency until we disable exclusive task scheduling on the bare pool.)

**Related issues**: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3758

### DIFF
--- a/server/util/networking/networking.go
+++ b/server/util/networking/networking.go
@@ -8,10 +8,12 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"slices"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
@@ -30,6 +32,7 @@ var (
 	blackholePrivateRanges        = flag.Bool("executor.blackhole_private_ranges", false, "If true, no traffic will be allowed to RFC1918 ranges.")
 	preserveExistingNetNamespaces = flag.Bool("executor.preserve_existing_netns", false, "Preserve existing bb-executor net namespaces. By default all \"bb-executor\" net namespaces are removed on executor startup, but if multiple executors are running on the same machine this behavior should be disabled to prevent them interfering with each other.")
 	natSourcePortRange            = flag.String("executor.nat_source_port_range", "", "If set, restrict the source ports for NATed traffic to this range. ")
+	networkLockDir                = flag.String("executor.network_lock_directory", "", "If set, use this directory to store lockfiles for allocated IP ranges. This is required if running multiple executors within the same networking environment.")
 
 	// Private IP ranges, as defined in RFC1918.
 	PrivateIPRanges = []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "169.254.0.0/16"}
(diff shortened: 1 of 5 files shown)


server/util/networking/networking.go
```diff
@@ -410,10 +439,42 @@ func (a *HostNetAllocator) Get() (*HostNet, error) {
...
+// It does not block - if the lock is already held then it returns a nil file
+// handle.
+// If a non-nil file handle is returned, the file must be closed to release the
+// lock.
+func (a *HostNetAllocator) tryFlock(netIdx int) (*os.File, error) {
+	if err := os.MkdirAll(*networkLockDir, 0755); err != nil {
+		return nil, fmt.Errorf("make lock dir %q: %s", *networkLockDir, err)
+	}
+	path := filepath.Join(*networkLockDir, fmt.Sprintf("ip_range.%d.lock", netIdx))
+	f, err := os.Create(path)
```
maggie-lou, line 452: Sanity check - if path already exists (because another process already holds the lock), will os.Create return the same file descriptor?

bduffany: it'll return a different file descriptor (file descriptors are local to a process) but the returned descriptor will point to the same underlying file (inode), so `flock` should work here. (I also checked this with some debug logging)

maggie-lou (approved):
LGTM - but maybe check with Iain because I remember he had some thoughts/alternate ideas for other ways to implement this

bduffany:
> LGTM - but maybe check with Iain because I remember he had some thoughts/alternate ideas for other ways to implement this

Sure - for reference, I dug up the thread: https://buildbuddy-corp.slack.com/archives/C04R9V7H28N/p1697734759336109

A couple of thoughts after reading through it:
* We talked about tying namespaces to IP addresses by using a naming convention like `/var/run/netns/bb-executor-192-168-XXX-XXX`, but in the new world of network pooling, IP addresses and net namespaces are no longer 1-1 (because we unassign/reassign IPs when adding to/taking from the pool, but keep the net namespace around). So, I think the problem is now mostly scoped to just locking IP ranges, and net namespaces are not really as relevant.
* Towards the end of the thread we talked about possibly using a library for flocking or rolling our own. I think the current implementation of just making a `flock()` system call is simple enough that pulling in a dependency is not warranted, nor is it worth spending the time rolling our own. For cross-process coordination that requires more complexity than just acquiring a lock (e.g. storing data), I think sqlite is probably a better solution - e.g. podman uses sqlite to maintain container state without needing a daemon process.

iain-macdonald (approved):

PR #7440, opened 2024-09-12
ociruntime: decompress using go-containerregistry

sluongng:
Avoid handling gzip decompression via `tar` command.
Replace it with `layer.Uncompressed()` from go-containerregistry which
supports both gzip and zstd format.

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -1160,8 +1160,7 @@ func (s *ImageStore) pull(ctx context.Context, imageName string, creds oci.Crede
 			// the credentials in the key here too.
 			key := hash.Strings(destDir, creds.Username, creds.Password)
 			_, _, err = s.layerPullGroup.Do(ctx, key, func(ctx context.Context) (any, error) {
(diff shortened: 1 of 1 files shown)


bduffany (approved):

sluongng:
@bduffany I finally got time to add test to this PR.

Using the nix-built image, we gonna get an error like this before this PR
https://buildbuddy.buildbuddy.io/invocation/4a31d98a-b9f0-4f03-8f92-3505ef7b6df5?target=%2F%2Fenterprise%2Fserver%2Fremote_execution%2Fcontainers%2Fociruntime%3Aociruntime_test&targetStatus=6

After this PR landed, the issue was gone.

However, Im not sure if we should use the partner-provided image in our test. This is also a network-based test as we are fetching images from ghcr.io 🤔

bduffany:
> However, Im not sure if we should use the partner-provided image in our test. This is also a network-based test as we are fetching images from ghcr.io 🤔

Yeah, let's omit this test for now. We eventually want to have a "test registry" that runs locally in the test container for these types of tests, with the images provisioned via Bazel

sluongng:
Let me see if I can set it up after migrating stuffs to rules_oci.

PR #7448, opened 2024-09-12
[RB] Add bazel query button for build targets

maggie-lou:
Help customers answer "Why did this target build?"

**Related issues**: N/A


app/invocation/cache_requests_card.tsx
```diff
@@ -611,6 +636,24 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
...
+            <DialogHeader>
+              <DialogTitle>GitHub Sync Required</DialogTitle>
+            </DialogHeader>
+            <DialogBody>
+              {/*TODO(Maggie): Link to a remote bazel blog post*/}
+              <p>To use Remote Bazel backed UI features, you must link a GitHub repository.</p>
+            </DialogBody>
+            <DialogFooter>
+              <OutlinedButton onClick={this.onCloseWorkflowUpsell.bind(this)}>Cancel</OutlinedButton>
+              <Button onClick={this.onClickWorkflowUpsell.bind(this)}>Link a Repository</Button>
```
bduffany, line 653: nit: use `LinkButton` here instead of manually calling `window.open()` - it behaves the same as `<a>` so you can set `target` as well.

app/invocation/cache_requests_card.tsx
```diff
@@ -611,6 +636,24 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
         className={
           this.getGroupBy() === cache.GetCacheScoreCardRequest.GroupBy.GROUP_BY_TARGET ? "group-by-target" : ""
         }>
+        <Modal
+          className="workflow-upsell-modal"
+          isOpen={this.state.isWorkflowUpsellOpen}
+          onRequestClose={this.onCloseWorkflowUpsell.bind(this)}>
+          <Dialog>
+            <DialogHeader>
+              <DialogTitle>GitHub Sync Required</DialogTitle>
```
bduffany, line 645:
```suggestion
              <DialogTitle>GitHub link required</DialogTitle>
```

app/invocation/invocation.tsx
```diff
@@ -135,6 +135,13 @@ export default class InvocationComponent extends React.Component<Props, State> {
         this.modelChangedSubscription = this.state.model?.onChange.subscribe(() => this.forceUpdate());
       }
     }
+
+    if (this.props.search.get("openChild") == "true" && this.state.childInvocations.length > 0) {
+      const childInvocationId = this.state.childInvocations[0].invocationId;
+      const url = `/invocation/${childInvocationId}?queued=true`;
+      window.location.href = url;
+    }
```
bduffany, line 143: I think this might wind up feeling janky? it might take a little while before the child invocation loads, then suddenly a new page opens without the user's explicit intent.

app/invocation/cache_requests_card.tsx
```diff
@@ -657,6 +700,24 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
...
+                          </div>
+                        )}>
+                        <HelpCircle
+                          onClick={this.executeRemoteBazelQuery.bind(this, group.results[0]?.targetId!)}
+                          className="download-button icon"
+                          role="button"
+                        />
+                      </Tooltip>
+                    </>
+                  )}
```
bduffany, line 720: should we add this to the "Target" page instead (or in addition)?

app/util/remote_runner.tsx
```diff
@@ -0,0 +1,38 @@
...
+    .run(request)
+    .then((response: runner.RunResponse) => {
+      let url = `/invocation/${response.invocationId}?queued=true`;
+      if (autoOpenChild) {
+        url += "&openChild=true";
+      }
+      window.open(url, "_blank");
+    })
+    .catch((error) => {
+      alert(error);
```
bduffany, line 36: nit: use `error_service.handleError(error)` instead

app/util/remote_runner.tsx
```diff
@@ -0,0 +1,38 @@
...
+  request.gitRepo.repoUrl = invocationModel.getRepo();
+  let step = new runner.Step();
+  step.run = command;
+  request.steps = [step];
+  let state = new git.RepoState();
+  state.commitSha = invocationModel.getCommit();
+  state.branch = invocationModel.getBranchName();
+  request.repoState = state;
+  request.async = true;
+  request.runRemotely = true;
```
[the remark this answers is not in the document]
maggie-lou: Oh nice - TIL

app/invocation/invocation.tsx
maggie-lou: Yeah I wasn't sure which is better. You can't see the query graph unless you click on the child invocation, which is the valuable part. I feel like it could be unclear the intended outcome if it isn't automatic. @siggisim any thoughts on this? (Once the child invocation containing the query graph loads, should we automatically open it?)

jdhollen: i know it's a separate invocation, but don't we kinda want this to show up on this invocation's page + inline the results? we could still link out to the parent obviously, but staying put and showing loading state + the results here seems reasonable to me

siggisim: Maybe for now we can just have them click on the child invocation - and in the future we can come up with some clever way of rendering the graph on the parent workflow page.

maggie-lou: @jdhollen The one downside with your approach is that this could be relatively slow if it's on a clean runner. I feel like linking to the remote bazel run gives more context on what's happening

Are people opposed to automatically opening the child invocation page? We could do it in the same tab if that's less intrusive. As someone who hadn't used our UI to run a bazel query, I wouldn't have known that clicking the child invocation would show a different view of a graph (because usually it contains the same logs as the parent)

app/invocation/cache_requests_card.tsx
maggie-lou: Done. I like having it on the cache table as well, because that seems more commonly used by customers

app/invocation/invocation.tsx
bduffany: Opening in the same tab seems OK to me for now (one of the `router.replace*` methods) - I still think the experience will be a bit jarring so it would be nice if we could figure out a generally better way for displaying child invocation results inside the parent workflow. (This would be useful for workflows too)

maggie-lou: @bduffany Doesn't `window.location.href` open it in the same tab?

bduffany: It does, but it will reload the whole app (it will request the JS from the server, re-execute the whole JS, re-auth, etc.) rather than just re-rendering the app that is already loaded). We generally do all in-app navigations using the `router.navigate()` functions except in certain special cases like switching subdomains.

maggie-lou: Got it - thanks for explaining. Changed

jdhollen: @maggie-lou i'm fine with auto-linking out until we've got something fancy to plop in (like siggi said), but i do think we'll eventually want this to become sort of "okay, that's in progress" status that updates inline + a link to more details while the query is running + then a graph + link to the invocation when it's done.

but yeah, just linking out for now is definitely fine

siggisim (approved):

PR #7449, opened 2024-09-12
Add a simple searcher test

tylerwilliams:
Gonna change the truncate method -- wanted to add a test first.

### DIFF
--- a/codesearch/types/types.go
+++ b/codesearch/types/types.go
@@ -93,9 +93,7 @@ type Highlighter interface {
 
 type Query interface {
 	SQuery() string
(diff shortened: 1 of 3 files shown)


codesearch/types/types.go
```diff
@@ -93,9 +93,7 @@ type Highlighter interface {
 
 type Query interface {
 	SQuery() string
-	ParsedQuery() string
 	Scorer() Scorer
-	Highlighter() Highlighter
```
tylerwilliams, line 98: these aren't needed in the interface.

siggisim (approved):

PR #7452, opened 2024-09-13
go.mod: upgrade github.com/GoogleCloudPlatform/cloudsql-proxy to latest

sluongng:
We received some feedbacks from customer about our transitive
dependencies being vulnerable:
- github.com/jackc/pgproto3 v1.1.0 -> CVE-2024-27304, severity=9.8
- github.com/jackc/pgproto3/v2 v2.3.2 -> CVE-2024-27304, severity=9.8

After investigation, we confirm that these Go modules are NOT used in
our build graph and was only included as transitive dependencies.
Removing them from deps.bzl does not affect/change our build result.

However, to please the security scanner, let's upgrade the package that
pulled in these transitive deps: cloudsql-proxy.

This upgrade also include a bump in go-fuse/v2, which includes our
submitted change. We can now retire our custom patch for good.


fmeum (approved):
You also need to drop the patch in the Bzlmod setup.

What do you think, are we ready for turning on `--enable_bzlmod` unconditionally and exclusively use `go_deps` to manage Go dependencies? That would have avoided this issue in the first place.

sluongng:
> What do you think, are we ready for turning on --enable_bzlmod unconditionally and exclusively use go_deps to manage Go dependencies? That would have avoided this issue in the first place.

We gonna need to move our internal repo to bzlmod so that it does not rely on the deps.bzl file.

iain-macdonald (approved):
LGTM but check with Brandon because it looks like he tried to upgrade go-fuse [earlier this year](https://github.com/buildbuddy-io/buildbuddy/pull/5763) and then [reverted](https://github.com/buildbuddy-io/buildbuddy/pull/5773) it.

sluongng:
Pinned go-fuse/v2 to a landed commit of Zoey https://github.com/hanwen/go-fuse/commit/db8da281991c8a71c8393824878b436c3feec137

PR #7453, opened 2024-09-13
    go.mod: upgrade containerd/containerd to v1.7.22

sluongng:
We received some feedbacks from customer about our transitive
dependency being vulnerable:
- github.com/containerd/containerd v1.5.8 -> CVE-2021-43816, severity=9.1

After investigation, we confirm that this Go modules are NOT used in
our build graph and was only included as transitive dependencies.
The actual version we used was v1.7.14

However, to please the security scanner, let's upgrade the packages that
pulled in this transitive deps, so we can get rid of the vulnerable
deps completely. These included:

```bash
go get github.com/ClickHouse/clickhouse-go/v2@latest
go get gorm.io/driver/clickhouse@latest

# Also needed to resolve some version conflict
go get github.com/docker/docker@latest
go get go.opentelemetry.io/otel/sdk@latest
go get gorm.io/driver/{mysql,postgres,sqlite}@latest
```

Because docker was upgraded, we need to switch away from using
deprecated APIs by updating `docker.go` accordingly.

Base on #7452.


iain-macdonald (approved):

sluongng:
There was a merge conflict with master's latest changes and it was not trivial to resolve, so I recreated the changes, one by one on top of `master`, and split them into separate commits instead.

tempoz (approved):
LGTM

classic containerd upgrade, adding a thousand deps 😣

PR #7454, opened 2024-09-13
Add `bb explain` command

fmeum:
The new `bb explain` command accepts two compact execution logs and computes a structural diff of the two, without first converting them into the expanded execution log format.

The command internally uses a protobuf schema to represent structural diffs. For now, this format is not exposed and only meant for usage in tests and to allow more flexible rendering, e.g. in the UI.

Sample output (adding a new public method to Bazel's `Label` class):
```
Turbine //src/main/java/com/google/devtools/build/lib/cmdline:cmdline (bazel-out/darwin_arm64-fastbuild/bin/src/main/java/com/google/devtools/build/lib/cmdline/libcmdline-hjar.jar)
  inputs changed:
    src/main/java/com/google/devtools/build/lib/cmdline/Label.java
  transitively invalidated:
    542 Javac
    296 Turbine

Javac //src/main/java/com/google/devtools/build/lib/cmdline:cmdline (bazel-out/darwin_arm64-fastbuild/bin/src/main/java/com/google/devtools/build/lib/cmdline/libcmdline.jar)
  inputs changed:
    src/main/java/com/google/devtools/build/lib/cmdline/Label.java
  transitively invalidated:
    1 Genrule
    1 JavaDeployJar
```

**Related issues**: N/A


sluongng (commented):
I gave it a quick read. The core logic in compactgraph.go makes sense. I feel like you would benefit from a special `SortedSet` struct that includes a map and a slice underneath to keep the order of the elements underneath.

Some exported func could use a simple test to demonstrate how they behave. That way we could validate the functionality overtime.

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,478 @@
...
+	"google.golang.org/protobuf/encoding/protodelim"
+
+	spawnproto "github.com/buildbuddy-io/buildbuddy/proto/spawn"
+)
+
+// CompactGraph represents the compact execution log as a directed acyclic graph of spawns.
+// The keys are the output paths of all spawns in the graph and each spawn contains references to other execution log
+// entries, such as input files, directories, and input sets. The edges between spawns are tracked implicitly by
+// matching the output paths of the spawns to the input paths of the referenced entries.
+type CompactGraph map[string]*Spawn
```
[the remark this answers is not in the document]
fmeum: The graph representation treats the digests as opaque byte blobs, so it doesn't need to store the hash function name at the top level. Since we return `Digest`s as part of the diffs, I am now passing the name down to the individual `File` nodes during creation.

Since an input can be consumed by a large number of spawns, I think that we would want to avoid building such an index. When it happens, we should definitely make this a struct, but I also don't want to promise API stability for this package yet :-D

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,478 @@
...
+
+// ReadCompactLog reads a compact execution log from the given reader and returns the graph of spawns, the hash function
+// used to compute the file digests, and an error if any.
+func ReadCompactLog(in io.Reader) (CompactGraph, string, error) {
```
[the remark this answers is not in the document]
fmeum: I looked into `printlog/compact`, but I think that it's only superficially similar (as in, it also parses the compact exec log). I don't think there is anything we could deduplicate (`compactgraph` keeps the compact entries with metadata attached, `printlog/compact` converts them directly).

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,478 @@
...
+			DiffType:      spawn_diff.SpawnDiff_NEW_ONLY,
+		})
+	}
+
+	commonOutputs := setIntersection(oldPrimaryOutputs, newPrimaryOutputs)
+	type diffResult struct {
+		spawnDiff   *spawn_diff.SpawnDiff
+		localChange bool
+		affectedBy  []string
+	}
```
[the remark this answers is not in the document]
fmeum: This type is only used in this function and only as an intermediate to arrange data in a `sync.Map`. Wouldn't lifting it out of the function unnecessarily pollute the top level?

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,478 @@
...
+		for _, inputSet := range n.TransitiveSets {
+			visitor(inputSet)
+		}
+	case *Spawn:
+		visitor(n.Tools)
+		visitor(n.Inputs)
+	}
+}
+
+func (cg *CompactGraph) primaryOutputs() []string {
```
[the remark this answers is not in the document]
fmeum: We could, but as we only access this data once, storing it would also be a waste (and complicate `CompactGraph` since we would now be storing the result of a pure function of its other contents). Since we are iterating over the graph many times anyway, I wouldn't expect this to have meaningful CPU time impact.

cli/explain/explain.go
```diff
@@ -0,0 +1,171 @@
...
+	for _, spawnDiff := range spawnDiffs {
+		writeSpawnDiff(os.Stdout, spawnDiff)
+	}
+	return 0, nil
+}
+
+func diff(oldPath, newPath string) ([]*spawn_diff.SpawnDiff, error) {
+	readsEG := errgroup.Group{}
+	var oldGraph compactgraph.CompactGraph
+	var oldHashFunction string
```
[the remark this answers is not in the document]
fmeum: See my comment above, it's not needed in the graph structure itself, only in its `File` nodes (which do now store it).

cli/explain/explain.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/buildbuddy-io/buildbuddy/cli/log"
+	"github.com/buildbuddy-io/buildbuddy/proto/spawn_diff"
+	"golang.org/x/sync/errgroup"
+
+	gocmp "github.com/google/go-cmp/cmp"
+)
+
+const (
+	usage = `
+usage: bb explain <old compact execution log> <new compact execution log>
```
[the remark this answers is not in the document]
fmeum: Good point, changed this to use flags!

fmeum:
I am working on tests now. I am leaning towards integration tests that diff two very small execution logs emitted by real Bazel invocations.

fmeum:
I added support for more directories and symlinks as well as integration tests for basic scenarios.

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,506 @@
...
+
+// reduceToRoots returns the sorted subset of outputs that are not inputs to any other spawn in the graph.
+func (cg *CompactGraph) reduceToRoots(outputs []string) []string {
+	rootsSet := make(map[string]struct{})
+	for _, output := range outputs {
+		rootsSet[output] = struct{}{}
+	}
+
+	// Visit all nodes in the graph and remove those with incoming edges from the set of roots.
+	var toVisit []interface{}
```
sluongng, line 177: nit: prefer using `any` over `interface{}`

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,506 @@
...
+		node, toVisit = toVisit[0], toVisit[1:]
+		visited[node] = struct{}{}
+		switch n := node.(type) {
+		case *File:
+		case *Directory:
+			delete(rootsSet, n.Path())
+		}
+		cg.visitSuccessors(node, func(input interface{}) {
+			markForVisit(input)
+		})
```
sluongng, line 198:
```suggestion
		cg.visitSuccessors(node, markForVisit)
```

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	roots := maps.Keys(rootsSet)
+	slices.Sort(roots)
+	return roots
+}
+
+// sortedPrimaryOutputs returns the primary output paths of the spawns in topological order.
+func (cg *CompactGraph) sortedPrimaryOutputs() []string {
+	var toVisit []interface{}
```
sluongng, line 208: nit: this could be `[]*Spawn` no?

tools/tools.go
```diff
@@ -0,0 +1,9 @@
+//go:build tools
+
+// This file is used to mark Go modules as used that are only (deps of) tools run during the build.
+package tools
+
+import (
+	// Used by cli/explain/compactgraph/testdata/generate to generate test data.
+	_ "github.com/otiai10/copy"
+)
```
sluongng, line 9: I think you already included this dep in `generate.go` so this file should not be needed right?

Otherwise, move this import to `tools/deps/deps.go` as we are keeping our tool deps there.

cli/cli_command/cli_command.go
```diff
@@ -97,4 +98,8 @@ var Commands = []Command{
 		Help:    "Prints bb cli version info.",
 		Handler: versioncmd.HandleVersion,
 	},
+	{
+		Name:    "explain",
+		Handler: explain.HandleExplain,
```
sluongng, line 103: Nit: missing Help

cli/explain/compactgraph/testdata/generate.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		log.Fatalf("Failed to create temp output base: %s", err)
+	}
+	defer os.RemoveAll(outputBase)
+	// Bazel's output base can contain files with no write permissions.
+	defer filepath.WalkDir(outputBase, func(path string, d fs.DirEntry, err error) error {
+		return os.Chmod(path, 0755)
+	})
+	cmd := exec.Command(
+		"bazel",
```
sluongng, line 190: Nit: assuming this is Bazelisk, it would be nice to get a `.bazelversion` setup here so we can reproduce/update these a bit easier in the future.

sluongng (approved):
A few nits left, but this looks like a good first iteration.

cli/explain/compactgraph/compactgraph.go
```diff
@@ -0,0 +1,506 @@
...
+		spawnDiff   *spawn_diff.SpawnDiff
+		localChange bool
+		affectedBy  []string
+	}
+	diffResults := sync.Map{}
+	diffWG := sync.WaitGroup{}
+	for _, output := range commonOutputs {
+		diffWG.Add(1)
+		go func() {
+			aSpawn := old[output]
```
bduffany, line 116: nit: where possible I like to put `defer wg.Done()` at the top, to minimize the distance between the `Add()` / `Done()` calls and prevent accidentally adding a `return` later that would prevent the `Done()` from getting called.

```suggestion
		diffWG.Add(1)
		go func() {
			defer diffWG.Done()	
			aSpawn := old[output]
```

cli/explain/compactgraph/testdata/generate.go
```diff
@@ -0,0 +1,224 @@
...
+	cmd.Stderr = os.Stderr
+	if err = cmd.Run(); err != nil {
+		// Allow failures due to no tests as we always run with `bazel test`.
+		if exitErr, ok := err.(*exec.ExitError); !ok || exitErr.ExitCode() != 4 {
+			log.Fatalf("Failed to run command: %s", err)
+		}
+	}
+}
+
+func extractTxtar(dir string, tar string) {
```
bduffany, line 212: TIL about txtar - looks useful :)

bduffany (approved):

cli/explain/compactgraph/compactgraph.go
fmeum: Ah, no, because the nodes in the graph are not just spawns, but also any kind of `Input`.

cli/explain/compactgraph/testdata/generate.go
fmeum: Go 1.23 even has `os.CopyFS` in the stdlib, but unfortunately we can't use it here (it errors out when a file already exists).

PR #7455, opened 2024-09-13
Make OCI image cache directory versioned and configurable

bduffany:
- Add `--executor.oci.image_cache_root` flag which sets the directory where image layers are cached.
- Make the image cache versioned - store layers under a `v1/` subdirectory so that if we make backwards-incompatible changes to the storage format, we can migrate more smoothly.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -44,11 +44,12 @@ import (
 )
 
 var (
-	Runtime     = flag.String("executor.oci.runtime", "", "OCI runtime")
-	runtimeRoot = flag.String("executor.oci.runtime_root", "", "Root directory for storage of container state (see <runtime> --help for default)")
-	pidsLimit   = flag.Int64("executor.oci.pids_limit", 2048, "PID limit for OCI runtime. Set to -1 for unlimited PIDs.")
-	dns         = flag.String("executor.oci.dns", "8.8.8.8", "Specifies a custom DNS server for use inside OCI containers. If set to the empty string, mount /etc/resolv.conf from the host.")
-	netPoolSize = flag.Int("executor.oci.network_pool_size", 0, "Limit on the number of networks to be reused between containers. Setting to 0 disables pooling. Setting to -1 uses the recommended default.")
+	Runtime        = flag.String("executor.oci.runtime", "", "OCI runtime")
+	runtimeRoot    = flag.String("executor.oci.runtime_root", "", "Root directory for storage of container state (see <runtime> --help for default)")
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -107,16 +113,16 @@ type provider struct {
 	// Each subdirectory corresponds to a created container instance.
 	containersRoot string
 
-	// Root directory where all image layer contents will be located.
+	// Root directory where cached images are stored.
 	// This directory is structured like the following:
 	//
-	// - {layersRoot}/
+	// - {imageCacheRoot}/v1/
```
iain-macdonald, line 119: update comment to ref imageCacheVersion?

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -152,11 +158,14 @@ func NewProvider(env environment.Env, buildRoot string) (*provider, error) {
 	if err := os.MkdirAll(containersRoot, 0755); err != nil {
 		return nil, err
 	}
-	layersRoot := filepath.Join(buildRoot, "executor", "oci", "layers")
-	if err := os.MkdirAll(layersRoot, 0755); err != nil {
+	imgRoot := *imageCacheRoot
+	if imgRoot == "" {
```
iain-macdonald, line 162: this seems like misconfiguration, do you think it'd make more sense to log.Fatal() here?

iain-macdonald, line 162: Nit: WDYT of setting the default value in the flag instead of here?

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: done

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: considered that, but it feels a bit odd for a constructor function to be modifying a global variable.

bduffany:
> Assuming we'll add a new persistent volume to store this stuff on, I think we probably want to do something to clean up old versions of the image cache.

I can run a manual cleanup script for this.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: actually nvm, I did this intentionally. This comment is specifically describing the V1 layout, so I think having it be a reference is potentially misleading.

iain-macdonald (approved):

PR #7458, opened 2024-09-13
Attempt to stop replicas we can't delete

tylerwilliams:
Fixes: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3765

### DIFF
--- a/enterprise/server/raft/store/store.go
+++ b/enterprise/server/raft/store/store.go
@@ -954,15 +954,13 @@ func (s *Store) syncRequestDeleteReplica(ctx context.Context, rangeID, replicaID
 		}
 		return err
 	})
-	if err != nil {
-		return status.InternalErrorf("failed to request delete replica for c%dn%d: %s", rangeID, replicaID, err)
-	}
-	return nil
+	return err
 }
 
 func (s *Store) syncRequestStopAndDeleteReplica(ctx context.Context, rangeID, replicaID uint64) error {
-	if err := s.syncRequestDeleteReplica(ctx, rangeID, replicaID); err != nil {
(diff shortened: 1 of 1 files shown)


enterprise/server/raft/store/store.go
```diff
@@ -1211,9 +1225,13 @@ func (s *Store) cleanupZombieNodes(ctx context.Context) {
 						return
 					}
 					s.log.Debugf("Removing zombie node: %+v...", potentialZombie)
-					if err := s.syncRequestDeleteReplica(ctx, potentialZombie.ShardID, potentialZombie.ReplicaID); err != nil {
-						s.log.Errorf("Error request delete zombie replica c%dn%d: %s", potentialZombie.ShardID, potentialZombie.ReplicaID, err)
-						return
+					err := s.syncRequestDeleteReplica(ctx, potentialZombie.ShardID, potentialZombie.ReplicaID)
+					if err != nil {
```
luluz66, line 1229: shall we only do that in the case of request rejected?

tylerwilliams: Done.

enterprise/server/raft/store/store.go
```diff
@@ -954,8 +954,29 @@ func (s *Store) syncRequestDeleteReplica(ctx context.Context, rangeID, replicaID
 		}
 		return err
 	})
-	if err != nil {
-		return status.InternalErrorf("failed to request delete replica for c%dn%d: %s", rangeID, replicaID, err)
+	return err
+}
+
+func (s *Store) syncRequestStopAndDeleteReplica(ctx context.Context, rangeID, replicaID uint64) error {
```
luluz66, line 960: nit: add a comment on what's the difference between this method and the syncRequestDelete method.

luluz66 (approved):

PR #7461, opened 2024-09-16
fetch_server: support uri-specific header qualifier

sluongng:
This allows Bazel to propagate credentials from netrc or cred helpers
to fetch external dependencies from BuildBuddy.

**Related issues**: N/A


tempoz (approved):
LGTM

PR #7462, opened 2024-09-16
Update path-to-regexp for dependabot#170

tempoz:
Upgrades our 1.x.x path-to-regexp dependency to a non-vulnerable version. We still depend on 2.2.1 transitively through serve-handler, which has not been updated in two years. We may have to patch it to resolve this dependabot alert.

**Related issues**: https://github.com/buildbuddy-io/buildbuddy/security/dependabot/170


tempoz:
This was handled in another PR

PR #7465, opened 2024-09-16
Upgrade `soci-snapshotter` to the version we use with our executors

tempoz:
We updated `soci-snapshotter` to pick up security fixes for vanta in the executor; this just mirrors that upgrade.

**Related issues**: N/A


deps.bzl
```diff
@@ -2394,8 +2442,12 @@ def install_go_mod_dependencies(workspace_name = "buildbuddy"):
...
         name = "com_github_hanwen_go_fuse_v2",
         importpath = "github.com/hanwen/go-fuse/v2",
-        sum = "h1:0Aex0XJ12fMbuuwyowIbUKBs976RxRuBxlYVK3eoHTA=",
-        version = "v2.5.2-0.20240816120607-db8da281991c",
+        patch_args = ["-p1"],
+        # TODO: Remove when/if hanwen/go-fuse#525 is merged. Prevents a kernel
+        # panic during init if /dev/null does not exist.
+        patches = ["@{}//buildpatches:com_github_hanwen_go_fuse_v2.patch".format(workspace_name)],
+        sum = "h1:OQBE8zVemSocRxA4OaFJbjJ5hlpCmIWbGr7r0M4uoQQ=",
+        version = "v2.5.1",
```
sluongng, line 2450: We already upgraded this to the exact commit with the fix.

deps.bzl
```diff
@@ -3759,20 +3805,20 @@ def install_go_mod_dependencies(workspace_name = "buildbuddy"):
...
+        sum = "h1:8SG7/vwALn54lVB/0yZ/MMwhFrPYtpEHQb2IpWsCzug=",
+        version = "v1.1.0",
     )
     go_repository(
         name = "com_github_opencontainers_runc",
         importpath = "github.com/opencontainers/runc",
-        sum = "h1:y2EZDS8sNng4Ksf0GUYNhKbTShZJPJg1FiXJNH/uoCk=",
-        version = "v1.1.7",
+        sum = "h1:L44KXEpKmfWDcS02aeGm8QNTFXTo2D+8MYGDIJ/GDEs=",
+        version = "v1.1.5",
```
sluongng, line 3815: Note: this seems to be a downgrade.

sluongng (changes requested):
Kinda weird that the MacOS ci runs are failing to build go-fuse. I don't think we should be using FUSE on MacOS so I suspect we can mark some of our targets to be linux-exclusive 🤔 

A few nits, but the PR overall looks good to me.

Great job on patching the external deps for bzlmod 👏

sluongng:
cc: @fmeum example of bzlmod visibility issue.

fmeum:
> cc: @fmeum example of bzlmod visibility issue.

Looks like the new override capabilities in 7.4.0 would fix this, right? When 7.4.0 is out, I'll try to fully migrate us over.

sluongng:
Yeah I think we should test it as soon as rc1 is available to confirm it early.

deps.bzl
tempoz: Good call; fixed.

buildpatches/com_github_hanwen_go_fuse_v2.patch
```diff
@@ -0,0 +1,56 @@
...
++		reservedFDs = append(reservedFDs, fd)
++		fd, err = syscall.Dup(fd)
++		if err != nil {
++			panic(fmt.Sprintf("Dup(%q): %v", fd, err))
++		}
+ 	}
+ }
+ 
+-- 
+2.39.2
```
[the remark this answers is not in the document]
tempoz: Done!

deps.bzl
tempoz: You're right; reverted.

enterprise/server/cmd/executor/BUILD
```diff
@@ -198,6 +198,21 @@ container_layer(
     tars = ["//enterprise/server/remote_execution/containers/podman:podman-static.tar.gz"],
 )
 
+container_layer(
+    name = "podman_static_runc_layer",
+    directory = "/usr/local/bin",
```
sluongng, line 203: Just FYI I got this from the latest image

```
> docker run --rm -it --entrypoint="" gcr.io/flame-public/buildbuddy-executor-enterprise:latest which runc
/usr/bin/runc
```

So not sure if writing runc to `/usr/local/bin` will take precedence.

PR #7467, opened 2024-09-16
server-side changes for supporting generic filters in invocation search 

jdhollen:
this is partially done, but there's not really a need to flag-guard it.

leaving some todos to add remaining filter types as they're a little different from the primary ones in this pr.

**Related issues**: N/A

### DIFF
--- a/proto/stat_filter.proto
+++ b/proto/stat_filter.proto
@@ -146,15 +146,7 @@ enum FilterType {
     supported_objects:
       [INVOCATION_OBJECTS, EXECUTION_OBJECTS];
   }];
-  TAG_FILTER_TYPE = 6 [(filter_type_options) = {
-    category:
-      STRING_FILTER_CATEGORY;
-    database_column_name:
-      "tag";
-    supported_objects:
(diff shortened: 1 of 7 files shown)


proto/stat_filter.proto
```diff
@@ -56,4 +58,219 @@ message Dimension {
...
+
+message FilterValue {
+  repeated string string_value = 1;
+  repeated int64 int_value = 2;
+}
+
+enum FilterType {
+  UNKNOWN_FILTER_TYPE = 0 [(filter_type_options) = {
+    supported_objects:
+      [NO_OBJECTS];
```
bduffany, line 107: nit: could NO_OBJECTS be represented as an empty list instead? (and maybe rename NO_OBJECTS => UNKNOWN_OBJECTS to match the usual UNKNOWN_* convention?)

bduffany (approved):

proto/stat_filter.proto
jdhollen: Yeah, this is better, guess I got tunnel vision from how I handled filter categories.

PR #7469, opened 2024-09-16
bazel button explain

maggie-lou:
- [RB] Add bazel query button for build targets
- PR feedback
- Add button to targets page
- Use router.replace
- DNR - Bazel button bb explain

**Related issues**: N/A


app/invocation/cache_requests_card.tsx
```diff
@@ -626,6 +693,43 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
         className={
           this.getGroupBy() === cache.GetCacheScoreCardRequest.GroupBy.GROUP_BY_TARGET ? "group-by-target" : ""
         }>
+        {capabilities.config.bazelButtonsEnabled && (
```
siggisim, line 696: I'm cool to submit this as-is behind a flag - but before we flip the flag in prod, let's find a location for the button that doesn't take up so much vertical space by itself.

siggisim (approved):

app/invocation/cache_requests_card.tsx
maggie-lou: Sounds good. Would you prefer if the button is on the same line as the title of the cache requests card? Or open to other suggestions

siggisim: Yeah maybe something like that - it feels more like a secondary action rather than a primary action

app/invocation/invocation.css
```diff
@@ -530,6 +530,41 @@
...
+#cache-miss-popup .title {
+  font-weight: 600;
+  font-size: 15px;
+  background: white;
+  padding: 4px;
+}
+
+#cache-miss-popup button {
+  margin-top: 5px;
+}
```
bduffany, line 548: nit, use a class rather than ID here? not a big deal but we don't usually use IDs for stuff

app/invocation/invocation.css
```diff
@@ -530,6 +530,41 @@
...
+
+.debug-cache-miss-container button {
+  font-size: 14px;
+}
+
+.debug-cache-miss-button {
+  padding-left: 5px;
+}
+
+.checkbox-row {
```
bduffany, line 562: nit: make this selector more specific like `.debug-cache-miss-blah .checkbox-row`

bduffany (approved):
looks good - excited to try this out :+1:

app/invocation/invocation.css
bduffany: For now, the way that we increase specificity is by either making the selector more deeply nested, or adding additional classnames.

e.g.

```
/* more deeply nested */
.cache-miss-popup .button.blah { ... }

/* add more classnames */
.cache-miss-popup .title.title { ... } /* duplicate .title selector for higher specificity */
```

In the long run, I think we need a better system, but this what we've been using so far :shrug:

maggie-lou: I am having trouble getting the CSS to apply to the popup itself (not the children elements). Just going to stick with the ID for now. Personally find it a bit more readable than the duplicated class name

bduffany: You should be able to write `.popup.cache-miss-popup` to get higher specificity than any of the `.popup` rules in `popup.css`. If that still doesn't work let me know and I can try to restructure our CSS setup to make it more friendly to overriding.

The reason I am trying to avoid IDs here is because it's introducing a "new way" to override specificity that we aren't doing anywhere else in the app, so it's adding a new thing that someone approaching the codebase would have to learn. The fewer conventions we are using throughout the app, the better.

maggie-lou: Ahhh thank you - TIL how to use no space to target multiple class names on the same element. That fixed things for me. Updated

bduffany: np, thanks for taking the time to fix that :)

PR #7484, opened 2024-09-18
track uploaded snapshot bytes.

jdhollen:
**Related issues**: N/A

### DIFF
--- a/cli/upload/upload.go
+++ b/cli/upload/upload.go
@@ -110,7 +110,7 @@ func uploadFile(args []string) error {
 		return err
 	}
 
-	_, err = cachetools.UploadFromReader(ctx, bsClient, ind, f)
+	_, _, err = cachetools.UploadFromReader(ctx, bsClient, ind, f)
 	if err != nil {
 		return err
 	}
--- a/enterprise/server/remote_execution/snaputil/BUILD
+++ b/enterprise/server/remote_execution/snaputil/BUILD
@@ -10,6 +10,7 @@ go_library(
         "//proto:remote_execution_go_proto",
         "//proto:resource_go_proto",
         "//server/interfaces",
+        "//server/metrics",
         "//server/remote_cache/cachetools",
         "//server/remote_cache/digest",
         "//server/util/log",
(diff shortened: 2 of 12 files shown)


jdhollen:
this is my dumb idea for how to do this, please tell me if you think i should do it a different way.

tylerwilliams (approved):

maggie-lou:
How does this differ from this metric? https://github.com/buildbuddy-io/buildbuddy/blob/3ba0238272a89f4b928f8629bdc232b1de0810f6/server/metrics/metrics.go#L1281

Dashboard: https://metrics.buildbuddy.io/d/KI6NxDWSk/workflow-snapshotting?orgId=1&refresh=1m&editPanel=14

jdhollen:
> How does this differ from this metric?
> 
> https://github.com/buildbuddy-io/buildbuddy/blob/3ba0238272a89f4b928f8629bdc232b1de0810f6/server/metrics/metrics.go#L1281
> 
> Dashboard: https://metrics.buildbuddy.io/d/KI6NxDWSk/workflow-snapshotting?orgId=1&refresh=1m&editPanel=14

We also call `snaputil.Cache` for everything in `range enumerateFiles(opts)` -- is this actually empty for some reason? i was worried that there might be a meaningful discrepancy here for whatever reason (if i'm not misunderstanding)

maggie-lou:
> We also call snaputil.Cache for everything in range enumerateFiles(opts) -- is this actually empty for some reason? i was worried that there might be a meaningful discrepancy here for whatever reason (if i'm not misunderstanding)

Hmmm I guess it could be interesting to see if those are unexpectedly large. Many of the files in `range enumerateFiles(opts)` are static though - it's contains things like the kernel and initrd images

At the very least, should we consolidate the two metrics?

jdhollen:
yup exactly--this should be static but it'd be good to see if there's unexpectedly a bunch of data flooding into the cache from somewhere.

and sure, i didn't think of that, we could just make some kind of "datatype" attr or something on the new metric and ditch the cow one?  does that sound good?

maggie-lou:
Sure sounds good. The old metric has a `fileName` label that we could use. The only slight preference for using the old metric is that we'll have more data if we want to query metric history, but prometheus doesn't keep data for that long anyway. If you switch to the new metric, reminder to update the grafana dashboard that reads from it

PR #7503, opened 2024-09-19
Use the correct context for local CAS proxy writes.

iain-macdonald:
context.Background() doesn't have creds so these writes were always failing in dev. Whoops!

**Related issues**: N/A


tylerwilliams (approved):
LGTM

PR #7510, opened 2024-09-20
Export kythe outputs in workflow

tylerwilliams:


enterprise/server/workflow/config/config.go
```diff
@@ -186,6 +186,24 @@ export KYTHE_DIR="$BUILDBUDDY_CI_RUNNER_ROOT_DIR/%s"
...
+find -L bazel-out/ -name *.go.kzip -print0 | xargs -r0 $KYTHE_DIR/tools/kzip merge --output output.go.kzip
+find -L bazel-out/ -name *.protobuf.kzip -print0 | xargs -r0 $KYTHE_DIR/tools/kzip merge --output output.protobuf.kzip
+
+if [ -f output.go.kzip ]; then
+  "$KYTHE_DIR/indexers/go_indexer" -continue output.go.kzip >> entries
+fi
+if [ -f output.protobuf.kzip ]; then
+  "$KYTHE_DIR/indexers/proto_indexer -index_file output.protobuf.kzip >> entries
+fi
+mv entries $BUILDBUDDY_ARTIFACTS_DIRECTORY/kythe_entries_for_buildbuddy
```
tylerwilliams, line 202: @maggie-lou is this the correct way to expose output files from this workflow?

maggie-lou: Yeah that looks right

maggie-lou (approved):

PR #7515, opened 2024-09-20
Cache Proxy: add a few graphs to the Grafana dashboard

iain-macdonald:
This PR adds two sections to the Grafana dashboard:
* "gRPC (cache proxy)" which contains all of the same graphs as "gRPC (buildbuddy-app)" but pointed at the cache proxy instead and
* "Cache proxy" which contains graphs tracking the hit/miss/partial performance of the ByteStream Proxy Server and CAS Proxy Server. CAS Proxy requests are tracked both by request (on the left axis) and by digest (on the right axis).

**Related issues**: N/A


tools/metrics/grafana/dashboards/buildbuddy.json
```diff
@@ -16139,7 +16744,7 @@
     ]
   },
   "timezone": "",
-  "title": "BuildBuddy Metrics",
-  "uid": "1rsE5yoGz",
+  "title": "BuildBuddy Metricssss",
+  "uid": "d246b6c6-7247-4b7e-849e-07f405d4ccb6",
```
bduffany, line 16748: lol intentional?

bduffany (commented):
Instead of adding a new grpc panel could you configure the existing one to repeat for each `job` like we do for the golang one?

tools/metrics/grafana/dashboards/buildbuddy.json
tylerwilliams: did a 🐍 write this

iain-macdonald:
> Instead of adding a new grpc panel could you configure the existing one to repeat for each `job` like we do for the golang one?

sounds good, done!

tools/metrics/grafana/dashboards/buildbuddy.json
```diff
@@ -16140,6 +16313,6 @@
   },
   "timezone": "",
   "title": "BuildBuddy Metrics",
-  "uid": "1rsE5yoGz",
+  "uid": "d246b6c6-7247-4b7e-849e-07f405d4ccb6",
```
bduffany, line 16316: any idea why it's trying to update the uid? this uid appears in the URL so this might break some things

bduffany (approved):

tools/metrics/grafana/dashboards/buildbuddy.json
iain-macdonald: Hmm, not sure, maybe because I saved it under a different name due to that other bug I mentioned to you. In any case, reverted.

PR #7524, opened 2024-09-23
Allow a few more operations with user key management

bduffany:
- Allow assigning API key capabilities when creating a key for a user other than the authenticated user (ORG_ADMIN keys only)
- Allow listing keys for users other than the authenticated user (ORG_ADMIN keys only)

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -681,8 +681,12 @@ func (d *AuthDB) authorizeNewAPIKeyCapabilities(ctx context.Context, userID, gro
 	if userID != "" {
 		// Capabilities assigned to the user-level key should not exceed the
 		// capabilities of the currently authenticated user.
+		//
+		// As a special case, we bypass this check for ORG_ADMIN keys, to allow
+		// user provisioning agents to assign cache capabilities, without having
+		// to grant those capabilities to the agent.
 		requestedCapabilities := capabilities.ToInt(caps)
-		if requestedCapabilities&capabilities.ToInt(userCapabilities) != requestedCapabilities {
+		if requestedCapabilities&capabilities.ToInt(userCapabilities) != requestedCapabilities && !slices.Contains(userCapabilities, akpb.ApiKey_ORG_ADMIN_CAPABILITY) {
 			return status.PermissionDeniedError("user does not have permission to assign these API key capabilities")
 		}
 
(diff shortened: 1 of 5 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -760,7 +760,18 @@ func (s *BuildBuddyServer) GetUserApiKeys(ctx context.Context, req *akpb.GetApiK
 		return nil, status.UnimplementedError("Not Implemented")
 	}
 	groupID := req.GetRequestContext().GetGroupId()
-	tableKeys, err := authDB.GetUserAPIKeys(ctx, groupID)
+
+	u, err := s.env.GetAuthenticator().AuthenticatedUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: always set req.user_id
```
tylerwilliams, line 768: might want to file a bug if you're not going to send a followup right after this?

tylerwilliams (approved):

server/buildbuddy_server/buildbuddy_server.go
bduffany: This has to be done over multiple rollouts so I'll file a bug just so I don't forget.

enterprise/server/backends/authdb/authdb.go
```diff
@@ -971,6 +978,26 @@ func (d *AuthDB) GetUserAPIKeys(ctx context.Context, groupID string) ([]*tables.
...
+	// required, and the user must be a member of the org.
+	if userID != u.GetUserID() {
+		caps, err := capabilities.ForAuthenticatedUserGroup(ctx, d.env, groupID)
+		if err != nil {
+			return nil, err
+		}
+		if !slices.Contains(caps, akpb.ApiKey_ORG_ADMIN_CAPABILITY) {
+			return nil, status.PermissionDeniedError("missing required capability")
+		}
+		ok, err := d.isGroupMember(ctx, groupID, userID)
```
vadimberezniker, line 991: Is this redundant with `ForAuthenticatedUserGroup` which also checks group membership?

vadimberezniker (approved):

enterprise/server/backends/authdb/authdb.go
bduffany: `ForAuthenticatedUserGroup` uses the authenticated user ID rather than the requested `user_id` which is different in this case

PR #7526, opened 2024-09-23
[raft] don't start eviction when we are still initializing replicas

luluz66:
**Related issues**: N/A


enterprise/server/raft/store/store.go
```diff
@@ -477,6 +477,7 @@ func (s *Store) AddEventListener() <-chan events.Event {
 // Start starts a new grpc server which exposes an API that can be used to manage
 // ranges on this node.
 func (s *Store) Start() error {
+	s.usages.Start()
```
tylerwilliams, line 480: is there a corresponding call to Stop?

tylerwilliams (approved):

enterprise/server/raft/store/store.go
luluz66: yup

PR #7527, opened 2024-09-23
Add optional CPU limit to OCI runtime.

vadimberezniker:
We can use this to set an upper bound on CPUs so that the CPU count looks uniform regardless of the underlying hardware.

https://github.com/buildbuddy-io/buildbuddy-internal/issues/3853

**Related issues**: N/A

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -48,7 +48,7 @@ var (
 	runtimeRoot    = flag.String("executor.oci.runtime_root", "", "Root directory for storage of container state (see <runtime> --help for default)")
 	imageCacheRoot = flag.String("executor.oci.image_cache_root", "", "Root directory for cached OCI images. Defaults to './executor/oci/images' relative to the configured executor.root_directory")
 	pidsLimit      = flag.Int64("executor.oci.pids_limit", 2048, "PID limit for OCI runtime. Set to -1 for unlimited PIDs.")
-	cpuLimit       = flag.Int("executor.oci.cpu_limit", 0, "Hard limit for CPU resources, expressed as CPU count. Default is no limit.")
+	cpuLimit       = flag.Int("executor.oci.cpu_limit", 0, "Hard limit for CPU resources, expressed as CPU count. Default (0) is no limit.")
 	dns            = flag.String("executor.oci.dns", "8.8.8.8", "Specifies a custom DNS server for use inside OCI containers. If set to the empty string, mount /etc/resolv.conf from the host.")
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -639,6 +640,15 @@ func (c *ociContainer) createSpec(ctx context.Context, cmd *repb.Command) (*spec
 	if err != nil {
 		return nil, fmt.Errorf("get container user: %w", err)
 	}
+
+	cpuSpecs := &specs.LinuxCPU{}
+	if *cpuLimit != 0 {
+		cpuSpecs = &specs.LinuxCPU{
+			Quota:  pointer(int64(*cpuLimit * 100_000)),
+			Period: pointer(uint64(100_000)),
```
tylerwilliams, line 648: a little comment about what units we're using here would be helpful for the reader

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -48,6 +48,7 @@ var (
 	runtimeRoot    = flag.String("executor.oci.runtime_root", "", "Root directory for storage of container state (see <runtime> --help for default)")
 	imageCacheRoot = flag.String("executor.oci.image_cache_root", "", "Root directory for cached OCI images. Defaults to './executor/oci/images' relative to the configured executor.root_directory")
 	pidsLimit      = flag.Int64("executor.oci.pids_limit", 2048, "PID limit for OCI runtime. Set to -1 for unlimited PIDs.")
+	cpuLimit       = flag.Int("executor.oci.cpu_limit", 0, "Hard limit for CPU resources, expressed as CPU count. Default is no limit.")
```
tylerwilliams, line 51: nit: Default (0) means no limit

tylerwilliams (approved):

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany, line 648: nit: might be a little clearer as

```suggestion
	if *cpuLimit != 0 {
		period := 100*time.Millisecond	
		cpuSpecs = &specs.LinuxCPU{
			Quota:  pointer(int64(*cpuLimit) * period.Microseconds())),
			Period: pointer(uint64(period.Microseconds())),
```

bduffany (approved):

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
vadimberezniker: fixed

vadimberezniker: done, thanks

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
vadimberezniker: done

PR #7530, opened 2024-09-23
Fix checkstyle take 2

maggie-lou:
Apply flags in CI only, because they refer to a .bazelrc config that is initialized by the ci_runner

**Related issues**: N/A

### DIFF
--- a/tools/checkstyle/checkstyle.sh
+++ b/tools/checkstyle/checkstyle.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 
-CI_INPUT="${1:-}"
-
 GO_PATH="$(readlink ./go)"
 GOFMT_PATH="$(readlink ./gofmt)"
 BB_PATH="$(readlink ./cli/cmd/bb/bb_/bb)"
@@ -50,7 +48,7 @@ run BuildFiles \
 
 run GoDeps \
   env GO_PATH="$GO_PATH" \
-  tools/fix_go_deps.sh --diff "$CI_INPUT"
+  tools/fix_go_deps.sh --diff
 
 run GoFormat \
   "$GOFMT_PATH" -d .
--- a/tools/checkstyle/checkstyle.sh
+++ b/tools/checkstyle/checkstyle.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
 
+CI_INPUT="${1:-}"
+
 GO_PATH="$(readlink ./go)"
 GOFMT_PATH="$(readlink ./gofmt)"
 BB_PATH="$(readlink ./cli/cmd/bb/bb_/bb)"
@@ -48,7 +50,7 @@ run BuildFiles \
 
 run GoDeps \
   env GO_PATH="$GO_PATH" \
-  tools/fix_go_deps.sh --diff
+  tools/fix_go_deps.sh --diff "$CI_INPUT"
 
 run GoFormat \
   "$GOFMT_PATH" -d .
--- a/tools/fix_go_deps.sh
+++ b/tools/fix_go_deps.sh
@@ -5,24 +5,24 @@ set -euo pipefail
 : "${GAZELLE_PATH:=}"
 : "${BAZEL_COMMAND:=bazelisk}"
 
-DIFF_MODE=0
-if [[ "${1:-}" == "-d" ]] || [[ "${1:-}" == "--diff" ]]; then
-  DIFF_MODE=1
-fi
-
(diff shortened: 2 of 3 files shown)


tools/fix_go_deps.sh
```diff
@@ -5,21 +5,27 @@ set -euo pipefail
...
+fi
+
+
+GAZELLE_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true //:gazelle "$CI_FLAGS" --)
 if [[ "$GAZELLE_PATH" ]]; then
   GAZELLE_COMMAND=("$GAZELLE_PATH")
 fi
 
-GO_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true //:go --config=buildbuddy_bes_results_url --config=buildbuddy_bes_backend --)
+GO_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true //:go "$CI_FLAGS" --)
```
bduffany, line 24: `"$CI_FLAGS"` will be expanded to a single variable - if you define CI_FLAGS as an array, like

```
CI_FLAGS=()
if ... then
  CI_FLAGS=(--config=buildbuddy_bes_backend --config=buildbuddy_results_url)
fi
```

then you could use array expansion here:

```suggestion
GO_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true "${CI_FLAGS[@]}" //:go --)
```

tools/checkstyle/checkstyle.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
 
+CI_INPUT="${1:-}"
```
bduffany, line 4: nit: maybe check `[[ "$CI" == true ]]` instead? we set this env var for workflows. Seems like it could simplify flag parsing etc

maggie-lou: Done - good idea

tools/fix_go_deps.sh
maggie-lou: Ah yes - realized this, thanks

tools/fix_go_deps.sh
```diff
@@ -5,12 +5,17 @@ set -euo pipefail
 : "${GAZELLE_PATH:=}"
 : "${BAZEL_COMMAND:=bazelisk}"
 
-GAZELLE_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true //:gazelle --config=buildbuddy_bes_backend --config=buildbuddy_bes_results_url --)
+CI_FLAGS=""
+if [[ "$CI" == true ]]; then
+  CI_FLAGS=(--config=buildbuddy_bes_backend --config=buildbuddy_bes_results_url)
+fi
+
+GAZELLE_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true //:gazelle "${CI_FLAGS[@]}" --)
```
bduffany, line 13:
```suggestion
GAZELLE_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true "${CI_FLAGS[@]}" //:gazelle --)
```

tools/fix_go_deps.sh
```diff
@@ -5,12 +5,17 @@ set -euo pipefail
...
 if [[ "$GAZELLE_PATH" ]]; then
   GAZELLE_COMMAND=("$GAZELLE_PATH")
 fi
 
-GO_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true //:go --config=buildbuddy_bes_results_url --config=buildbuddy_bes_backend --)
+GO_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true //:go "${CI_FLAGS[@]}" --)
```
bduffany, line 18: nit: put bazel flags before the target for consistency (or if you prefer putting them after the target, could you update the existing flags?)

```suggestion
GO_COMMAND=("$BAZEL_COMMAND" run --build_metadata=DISABLE_COMMIT_STATUS_REPORTING=true "${CI_FLAGS[@]}" //:go --)
```

bduffany (approved):

PR #7537, opened 2024-09-24
Set API key in pr.yaml Bazel commands

iain-macdonald:
**Related issues**: N/A

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -16,8 +16,14 @@ jobs:
 
       - name: Build
         run: |
-          bazelisk build --config=untrusted-ci //...
+          bazelisk build \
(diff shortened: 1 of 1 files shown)


vadimberezniker:
I'll defer to @bduffany  I believe it was intentionally left off for the untrusted CI configuration.

iain-macdonald:
> I'll defer to @bduffany I believe it was intentionally left off for the untrusted CI configuration.

I was wondering if that was the case, but couldn't quite understand what the risk is?

iain-macdonald:
Confirmed that pr.yaml only runs if a member of the BuildBuddy org approves the PR, so I don't think this creates any additional security risk, though LMK if I'm missing something.

bduffany (approved):

PR #7539, opened 2024-09-24
explain: Prepare for breaking changes to the compact execution log

fmeum:
The compact execution log has undergone breaking changes in https://github.com/bazelbuild/bazel/commit/c2f539c899e3801e7166da096bb19bb53a772810 in order to better support runfiles trees. This PR prepares for adding support for Bazel versions including that commit by:
* updating the proto
* adapting the `explain` command to the new proto while preserving the ability to parse existing exec logs
* making it easier (and more hermetic) to test against multiple Bazel versions

**Related issues**: N/A


cli/explain/compactgraph/compactgraph.go
```diff
@@ -262,14 +262,8 @@ func (cg *CompactGraph) visitSuccessors(node any, visitor func(input any)) {
...
-		for _, file := range n.Files {
-			visitor(file)
-		}
-		for _, dir := range n.Directories {
-			visitor(dir)
-		}
-		for _, inputSet := range n.TransitiveSets {
-			visitor(inputSet)
+		for _, input := range n.Inputs {
+			visitor(input)
```
sluongng, line 266: Do you need an if/else here to support the old exec logs before 7.4?

If we don't intend to support the older version, we gonna need a warning message asking users to upgrade.

proto/spawn.proto
```diff
@@ -252,33 +262,106 @@ message ExecLogEntry {
...
+    // belonging to this set.
+    repeated uint32 input_ids = 5;
+    // Entry IDs of other input sets contained in this set.
+    repeated uint32 transitive_set_ids = 4;
+
+    // Merged into input_ids for backwards compatibility with Bazel < 7.4.0.
+    repeated uint32 file_ids = 1 [deprecated = true];
+    repeated uint32 directory_ids = 2 [deprecated = true];
+    repeated uint32 unresolved_symlink_ids = 3 [deprecated = true];
+  }
```
sluongng, line 282: We might need to also modify the cli/print package to use the new proto fields? Perhaps in a separate PR though.

cli/explain/compactgraph/compactgraph.go
fmeum: `n` is a custom type here, not one of the protos, and `n.Inputs` contains the new and old proto fields merged together, so we really are supporting both versions.

proto/spawn.proto
fmeum: Yeah, but that's going to be a pretty large change by itself as we also need to flatten the new `RunfilesTree` message. I will build it into `explain` first as the expanded execution log will probably be deprecated in Bazel 8, but we could still look into making `printlog` work for both.

fmeum:
I fixed the test failures by making `printlog` syntactically compatible with the new proto. It still wouldn't work on logs generated by a new version of Bazel.

cli/explain/compactgraph/compactgraph.go
sluongng: I see, so we pre-flatten the InputSet to reduce the nested loops inside the switch later. Nice.

sluongng (approved):
lgtm

PR #7542, opened 2024-09-24
[raft] move handle events before replica initiation

luluz66:
**Related issues**: N/A
When replica.setRange, it will send usage event to the store events channel. But
we started handling events after the replicas are initiated. This might cause
the events channel to be blocked.


tylerwilliams (approved):

tylerwilliams:
do you want to add some error logging so we know if this kind of thing happens again?

luluz66:
Actually. this is not an issue, becuase in replica there is a default case when we notify usage.

PR #7546, opened 2024-09-25
Proxy Kythe apis to codesearch server

tylerwilliams:
1) Build ts protos of kythe stuff we're going to use in the UI.
2) Add a util/grpc_forward library called from grpc_server.go that forwards traffic, only if forwarding is configured, to some backend server.

I considered just manually implementing these methods in buildbuddy_service and calling the kythe backend service (writing out the proxy basically), but there are between 10 and 24 methods we'd want to proxy to the kythe backend, so this seemed cleaner and simpler. It also allows us to proxy other stuff in the future (like the codesearch APIs, for example).

LMK what you think, I'm open.


server/util/grpc_forward/grpc_forward.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/buildbuddy-io/buildbuddy/server/util/flag"
+	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
+	"github.com/buildbuddy-io/buildbuddy/server/util/status"
+	"github.com/mwitkow/grpc-proxy/proxy"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+var (
+	kytheBackendToProxy = flag.String("app.kythe_backend_to_proxy", "", "If set, proxy kythe traffic to this backend")
```
bduffany, line 17: nit: the other proxy-ish flags are `-cache_proxy.remote_cache` and `-build_event_proxy.hosts` - maybe `-kythe_proxy.target` would be a good name for this?

bduffany (approved):

server/util/grpc_forward/grpc_forward.go
```diff
@@ -0,0 +1,74 @@
...
+func init() {
+	mu.Lock()
+	backendConnectionPools = make(map[string]*grpc_client.ClientConnPool)
+	mu.Unlock()
+}
+
+type directorFunc func(ctx context.Context, fullMethodName string) (context.Context, *grpc.ClientConn, error)
+
+func getProxyDirector() directorFunc {
+	if *kytheBackendToProxy != "" {
```
vadimberezniker, line 32: if *kytheBackendToProxy == "" {
  return nil
}
the rest

server/util/grpc_forward/grpc_forward.go
```diff
@@ -0,0 +1,74 @@
...
+		return func(ctx context.Context, fullMethodName string) (context.Context, *grpc.ClientConn, error) {
+			target := ""
+			if strings.HasPrefix(fullMethodName, "/kythe.service") {
```
vadimberezniker, line 35: If this is meant to be a general forwarding util, there shouldn't be any kythe-specific references either here or in the flag name.

vadimberezniker (changes requested):

server/util/grpc_forward/grpc_forward.go
tylerwilliams: Ack (removed flag)

server/util/grpc_forward/grpc_forward.go
tylerwilliams: Ack. (removed flag)

server/util/grpc_forward/grpc_forward.go
tylerwilliams: Done.

server/util/grpc_forward/grpc_forward.go
```diff
@@ -0,0 +1,102 @@
...
+	once                   sync.Once
+	mu                     sync.RWMutex
+	backendConnectionPools map[string]*grpc_client.ClientConnPool
+)
+
+func lookupProxyTarget(fullMethodName string) (string, error) {
+	target := ""
+	for _, pair := range *proxyTargets {
+		if strings.HasPrefix(fullMethodName, pair.Prefix) {
+			target = pair.Target
```
vadimberezniker, line 35: Why not directly return here?

bduffany (approved):

vadimberezniker (approved):

server/util/grpc_forward/grpc_forward.go
tylerwilliams: Done.

PR #7550, opened 2024-09-25
reapi: favor platform set in Action over Command

sluongng:
In REAPI v2.2, the usage of platform in Command was deprecated over in
Action. Although most of the older REAPI clients, such as Bazel, are
setting it in both, newer clients such as Buck2 would only set it in
Action message.

This change replace most of existing `command.GetPlatform()` calls with
`action.GetPlatform()` while using the command value as fallback.

Closes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3863

### DIFF
--- a/enterprise/server/remote_execution/platform/platform.go
+++ b/enterprise/server/remote_execution/platform/platform.go
@@ -259,7 +259,11 @@ func (p *ExecutorProperties) SupportsIsolation(c ContainerType) bool {
 // executor-specific overrides applied via the ApplyOverrides function.
 func ParseProperties(task *repb.ExecutionTask) (*Properties, error) {
 	m := map[string]string{}
-	for _, prop := range task.GetCommand().GetPlatform().GetProperties() {
+	p := task.GetAction().GetPlatform()
+	if p == nil || len(p.GetProperties()) == 0 {
+		p = task.GetCommand().GetPlatform()
+	}
+	for _, prop := range p.GetProperties() {
 		m[strings.ToLower(prop.GetName())] = strings.TrimSpace(prop.GetValue())
 	}
 	for _, prop := range task.GetPlatformOverrides().GetProperties() {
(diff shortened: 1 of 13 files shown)


bduffany (commented):
as a preliminary step, wdyt about adding some logging in execution_server to see whether we get any actions in practice where the action platform is empty but the command platform isn't? If in practice, the action platform is always set, then I think we could skip adding this fallback logic everywhere to simplify things a bit.

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -259,7 +259,11 @@ func (p *ExecutorProperties) SupportsIsolation(c ContainerType) bool {
 // executor-specific overrides applied via the ApplyOverrides function.
 func ParseProperties(task *repb.ExecutionTask) (*Properties, error) {
 	m := map[string]string{}
-	for _, prop := range task.GetCommand().GetPlatform().GetProperties() {
+	p := task.GetAction().GetPlatform()
+	if p == nil || len(p.GetProperties()) == 0 {
+		p = task.GetCommand().GetPlatform()
+	}
```
bduffany, line 265: nit: can you consolidate this fallback logic into a helper func in `platform.go` like `platform.GetProto(action, command)`

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -696,6 +700,18 @@ func FindValue(platform *repb.Platform, name string) string {
...
+// property name (ignoring case) and returns the value of that property if it
+// exists, otherwise "".
+func FindValueInTask(task *repb.ExecutionTask, name string) string {
+	p := task.GetAction().GetPlatform()
+	if p == nil || len(p.GetProperties()) == 0 {
+		p = task.GetCommand().GetPlatform()
+	}
+	value, _ := findValue(p, name)
+	return value
+}
```
bduffany, line 713: nit: I think if we had a helper func like `platform.GetProto(action, command)` that returns the effective platform, then we could instead write `p := platform.GetProto(task.GetAction(), task.GetCommand())` and `platform.FindValue(p, "some-prop")` instead of using this helper func. It's slightly more verbose but helps minimize the API surface of this package - and we only need this logic in a couple of places.

sluongng:
we implemented this in another PR. Closing

PR #7551, opened 2024-09-25
Aspect_rules_swc: upgrade to v2.0.0

sluongng:
Upgrade swc to the latest version so that we are not stuck with the
older aspect transitive dependencies such as their aspect_bazel_lib.

This should unlock us from experimenting with other aspect rules such as
rules_oci.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -465,7 +465,7 @@ load("@aspect_rules_swc//swc:repositories.bzl", "swc_register_toolchains")
 
 swc_register_toolchains(
     name = "swc",
-    swc_version =  "v1.3.78", # LATEST_SWC_VERSION
+    swc_version = "v1.3.78",  # LATEST_SWC_VERSION
 )
 
 # Web testing
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -465,7 +465,7 @@ load("@aspect_rules_swc//swc:repositories.bzl", "swc_register_toolchains")
 
 swc_register_toolchains(
     name = "swc",
-    swc_version = "v1.3.78",  # LATEST_SWC_VERSION
+    swc_version = "v1.3.78",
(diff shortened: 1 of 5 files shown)


WORKSPACE
```diff
@@ -450,23 +450,22 @@ load("@build_bazel_rules_nodejs//toolchains/esbuild:esbuild_repositories.bzl", "
...
 
 rules_swc_dependencies()
 
-load("@aspect_rules_swc//swc:repositories.bzl", "swc_register_toolchains")
+load("@aspect_rules_swc//swc:repositories.bzl", "LATEST_SWC_VERSION", "swc_register_toolchains")
 
 swc_register_toolchains(
     name = "swc",
-    swc_version = "v1.2.141",
+    swc_version = LATEST_SWC_VERSION,
```
bduffany, line 468: Can we pin a version here? (Even if the LATEST_SWC_VERSION is hard-coded, it'd be nice to copy it here so we know what version we're using at a glance)

WORKSPACE
```diff
@@ -466,7 +465,7 @@ load("@aspect_rules_swc//swc:repositories.bzl", "swc_register_toolchains")
 
 swc_register_toolchains(
     name = "swc",
-    swc_version = "v1.2.141",
+    swc_version = "v1.3.78",  # LATEST_SWC_VERSION
```
bduffany, line 468: nit: this comment could easily go out of date

```suggestion
    swc_version = "v1.3.78",
```

bduffany (approved):

WORKSPACE
sluongng: I was thinking that it would help urge whoever upgrading our rules_swc next to also upgrade the swc_version 🌝 

But that's probably not needed.

PR #7553, opened 2024-09-25
Make --config=local builds work on Arm Macs when also using local executors

vanja-p:
Without these flags, remote execution fails because bazel tries to build for the linux_arm64 platform, and the locally running RBE doesn't support that.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -41,6 +41,10 @@ common:remote --config=remote-shared
 common:remote --config=cache
 common:remote --remote_executor=grpcs://buildbuddy.buildbuddy.io
 
+
+common:linux --platforms=//platforms:linux_x86_64
+common:linux --extra_execution_platforms=//platforms:linux_x86_64
+
 #common:remote-linux --config=remote
 #common:remote-linux --platforms=//platforms:linux_x86_64
 #common:remote-linux --extra_execution_platforms=//platforms:linux_x86_64
--- a/.bazelrc
+++ b/.bazelrc
@@ -250,6 +250,10 @@ common:macos --repo_env=BAZEL_NO_APPLE_CPP_TOOLCHAIN=1
 common:macos --macos_minimum_os=12.0
 common:macos --host_macos_minimum_os=12.0
 
+# Make --config=local builds work on Arm Macs when also using local executors.
+common:macos --platforms=//platforms:macos_arm64
+common:macos --extra_execution_platforms=//platforms:macos_arm64
+
 # Run Webdriver tests with --config=webdriver-debug to debug webdriver tests locally.
 # See server/testutil/webtester/webtester.go for more details.
 common:webdriver-debug --test_arg=-webdriver_headless=false
--- a/.bazelrc
+++ b/.bazelrc
@@ -35,13 +35,15 @@ common:remote-shared --remote_timeout=600
 common:remote-shared --remote_download_minimal
 common:remote-shared --jobs=100
 common:remote-shared --verbose_failures
-common:remote-shared --platforms=//platforms:linux_x86_64
-common:remote-shared --extra_execution_platforms=//platforms:linux_x86_64
 
 # Build with --config=remote to use BuildBuddy RBE.
 common:remote --config=remote-shared
 common:remote --config=cache
 common:remote --remote_executor=grpcs://buildbuddy.buildbuddy.io
+
+#common:remote-linux --config=remote
+#common:remote-linux --platforms=//platforms:linux_x86_64
+#common:remote-linux --extra_execution_platforms=//platforms:linux_x86_64
 
 # Build with --config=remote-dev to use BuildBuddy RBE.
 common:remote-dev --config=remote-shared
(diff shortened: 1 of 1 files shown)


vadimberezniker (approved):
Did you double check that building with --config=remote switches back to Linux platform? I expect that to be the case but it would be good to double check before merging.

vanja-p:
Yep, I checked that -config=remote switches back to Linux platform, but it seems like the `Test (darwin_amd64)` workflow fails with this change. I don't see an obvious way to fix that.

.bazelrc
```diff
@@ -250,6 +250,10 @@ common:macos --repo_env=BAZEL_NO_APPLE_CPP_TOOLCHAIN=1
 common:macos --macos_minimum_os=12.0
 common:macos --host_macos_minimum_os=12.0
 
+# Make --config=local builds work on Arm Macs when also using local executors.
+common:macos --platforms=//platforms:macos_arm64
+common:macos --extra_execution_platforms=//platforms:macos_arm64
```
bduffany, line 255: I think the `arm64` part is causing the macos workflows to fail because we have both amd64 and arm64 macs.

@sluongng do you remember why we are setting `--platforms=//platforms:linux_x86_64` in our `remote-shared` config? seems like this should be in some linux-specific config, no?

sluongng: I think we set it there because it's the most popular option among our team. We rarely build targeting MacOS remotely because our MacOS setup is usually (a) slower than our local MacBook and (b) at the time we only had amd64 mac executor while most/all of us were on M1 or later (arm64).

I don't mind having target linux_x86_64 moved to `common:remote-linux`, instead of `common:remote-shared`, and having a `common:remote-macos` for this.

Perhaps something like this?

```
--- a/.bazelrc
+++ b/.bazelrc
@@ -35,21 +35,23 @@ common:remote-shared --remote_timeout=600

 # Build with --config=remote to use BuildBuddy RBE.
 common:remote --config=remote-shared
 common:remote --config=cache
 common:remote --remote_executor=grpcs://buildbuddy.buildbuddy.io

+common:remote-linux --config=remote
+common:remote-linux --platforms=//platforms:linux_x86_64
+common:remote-linux --extra_execution_platforms=//platforms:linux_x86_64
```

tylerwilliams: +1 to moving this -- i think we should keep the configs straightforward and if people are doing funny things (target linux from mac) they can set the flags to do that?

bduffany: @sluongng your suggestion sounds good except maybe put the linux config in a shared config like `remote-linux-shared` so that we can easily reference it in other environments.

vanja-p: Removing the platforms from remote-shared doesn't seem to be enough to make building on a local RBE on a mac work, if we don't also set the platforms flags to `//platforms:macos_arm64` when building on arm macs. The builds still fail with the same error: `rpc error: code = Unavailable desc = No registered executors in pool "" with os "linux" with arch "amd64".` I'm not sure how the `Test (darwin_amd64)` and `Test (darwin_arm64)` workflows pass without setting a platform. Maybe because they don't use remote execution?

So that means we either need to specify the flags directly on each build (which seems ok to be) or create a config and specify that.

[--enable_platform_specific_config](https://bazel.build/reference/command-line-reference#flag--enable_platform_specific_config) can only set a config to windows, mac, or linux, depending on the host OS (and not architecture). Since we have enable_platform_specific_config set for every config in our bazelrc, and we have both arm64 and amd64 executors, we can't use common:mac for this. We could create a config `mac-arm64`, but I'm not sure if this worth it.

Also, it seems that multiple workflows rely on the current platforms flags in remote-shared, and fail when they are removed. Example: https://buildbuddy.buildbuddy.io/invocation/b631d886-a42f-4ef7-b7bf-8715c710ce1d.

Here are my ideas:
1. Don't change any configs. Just document that to run against a local RBE on mac, you must specify the platforms flags.
2. Create a `local-mac` config, which inherits from `local` and also sets the platforms configs. Document that this needs to be used when running against a local RBE on mac.
3. Move the existing platforms flags from `remote-shared` to `linux-workflows` or just `linux`. This won't be a problem until we have non amd64 executors on linux.

3 can be done independently of 1 or 2. I'm leaning towards 1, since the work there seems most proportional to the results.

bduffany: I think by default, when using `--config=remote-shared`, we want to use the "host" config, which I think should solve all of these issues - is there some special "host" platform value we could reference in `--platforms`? I vaguely remember something like that, but not sure if I'm misremembering. (If not, maybe we could define one ourselves by defining a platform `alias` that selects the correct platform based on `select()`, or using a repository_rule or something like that)

vanja-p: My understanding of https://bazel.build/extending/platforms follows. Assume I could be very wrong, but this is what I think.

There are 3 platform settings: host, target, and execution, with corresponding flags being `--host_platform`, `--platforms`, and `--extra_execution_platforms`. Host defaults to where bazel is running. Target is where the resulting built binary needs to run and it defaults to host. Execution is where all the actions run, and I'm not sure what the default is.

When we use the remote configs, we care about the execution platform. The remote config means we're running on against a deployed instance of RBE, and so we want to use the linux_x86_64 platform. This is probably true for most builds we do, regardless of whether they come from mac, linux, or windows.

When running RBE locally on an arm mac, and then sending builds to it, those builds need to use `--extra_execution_platforms=//platforms:macos_arm64`. I confirmed that the other flags don't need to be changed. Note that `--config=local` doesn't inherit from `remote-shared`, and so it uses the default host and target flags, since `--platforms` isn't set.

This made me think that we don't need `--platforms=//platforms:linux_x86_64` in `remote-shared`, so I removed it, but some actions fail: https://buildbuddy.buildbuddy.io/invocation/6f7b0a65-9c4e-43a3-a1eb-025ee6d64aea

After looking into this more, I think we can dynamically set the execution platform in the `local` config, using one of the rules we already have, with some changes, so that's what I'm doing now.

bduffany (approved):

PR #7568, opened 2024-09-26
sluongng/init rules oci

sluongng:
- **gazelle: upgrade to v0.39.0**
- **Remove for_bzlmod.patch files**
- **Init rules_oci**
- **add push example**


deployment/BUILD
```diff
@@ -39,3 +39,12 @@ write_flag_to_file(
     flag = ":image_tag",
     visibility = ["//visibility:public"],
 )
+
+# TODO: rules OCI only support tag file names with the ".txt" extension.
+genrule(
```
fmeum, line 44: Let's use skylib `copy_file` with `allow_symlink` instead for better RBE integration.

fmeum (approved):
Looks good to me (and pretty clean)!

deployment/BUILD
sluongng: Hmm I do like `copy_file` but I don't think it's better for RBE integration.

See https://github.com/bazelbuild/bazel-skylib/blob/e8e9d218ff563f93e3a7f7547ea532e66b210620/rules/private/copy_common.bzl#L43
and https://github.com/bazelbuild/bazel-skylib/issues/533 for more information.

fmeum: These exec properties should only be used when the rule runs a non-internal action, but with `allow_symlink` it just runs the internal symlink action which doesn't even take exec properties. That's why I think it's going to perform better.

sluongng: ah right. We have `allow_symlink=True` here already. Good call!

enterprise/server/cmd/server/BUILD
```diff
@@ -130,3 +132,27 @@ container_image(
     tags = ["manual"],
     visibility = ["//visibility:public"],
 )
+
+# TODO(sluongng): Verify the layering of the image.
```
sluongng, line 136: Worth noting: I am aware that the current pkg_tar setup is wrong.

- It does not pull in the much-needed runfiles of the go binaries
- There are also some layout changes on the final oci image vs the old image that we should address.

Since this PR is already fairly sized and does not break anything, I will be merging this as-is. A subsequent PR will be sent to correct the rules_pkg usage and oci image layout.

PR #7572, opened 2024-09-26
[RB] Fix a couple bazel button bugs

maggie-lou:
* Add a missing space in a command
* Only add bazel flags to bazel commands
* Support platform properties
* For the bb explain button, use commit shas rather than branches when comparing 2 invocations
**Related issues**: N/A

### DIFF
--- a/app/util/remote_runner.tsx
+++ b/app/util/remote_runner.tsx
@@ -71,8 +71,9 @@ export function triggerRemoteRun(
 
 // commandWithRemoteRunnerFlags adds useful flags to bazel commands run on a remote runner
 export function commandWithRemoteRunnerFlags(command: string): string {
-  // The buildbuddy_ configs point to the corresponding env that created the workflow
-  // action (Ex. If it was created in dev, points to the dev app)
+  // The buildbuddy_ configs point to the corresponding env that created the remote runner
+  // action (Ex. If it was created in dev, points to the dev app).
+  // These configs are defined in a .bazelrc the ci_runner creates in ci_runner/main.go
   const addlFlags =
     "--remote_cache_compression --config=buildbuddy_bes_backend --config=buildbuddy_bes_results_url --config=buildbuddy_remote_cache";
   command = appendBazelSubCommandArgs(command, addlFlags);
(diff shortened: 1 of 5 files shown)


app/util/remote_runner.tsx
```diff
@@ -53,6 +69,16 @@ export function triggerRemoteRun(invocationModel: InvocationModel, command: stri
     });
 }
 
+// commandWithRemoteRunnerFlags adds useful flags to bazel commands run on a remote runner
+export function commandWithRemoteRunnerFlags(command: string): string {
+  // The buildbuddy_ configs point to the corresponding env that created the workflow
+  // action (Ex. If it was created in dev, points to the dev app)
+  const addlFlags =
```
jdhollen, line 76: can we document the connection between this and `ci_runner/main.go` both here and there? this is kind of a brittle connection right now.

jdhollen (approved):

app/util/remote_runner.tsx
jdhollen: (fine to do separate from this pr)

app/util/remote_runner.tsx
```diff
@@ -53,6 +69,16 @@ export function triggerRemoteRun(invocationModel: InvocationModel, command: stri
...
+    "--remote_cache_compression --config=buildbuddy_bes_backend --config=buildbuddy_bes_results_url --config=buildbuddy_remote_cache";
```
bduffany, line 77: hm I thought we applied the BES flags by default? maybe only the remote cache ones are needed? otherwise lgtm

bduffany (approved):

app/util/remote_runner.tsx
maggie-lou: The problem is we have this in our .bazelrc

common --bes_results_url=https://app.buildbuddy.io/invocation/
common --bes_backend=grpcs://remote.buildbuddy.io

app/util/remote_runner.tsx
maggie-lou: Good idea - done

PR #7573, opened 2024-09-26
cli: surface BES forwarding RPC errors when in synchronous mode

bduffany:
When in `synchronous` mode (e.g. when `CI=true`) we should not throw away the `PublishBuildToolEventStream` error, and instead surface it to the client.

**Related issues**: N/A

### DIFF
--- a/server/build_event_protocol/build_event_server/build_event_server.go
+++ b/server/build_event_protocol/build_event_server/build_event_server.go
@@ -87,11 +87,18 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 
 	eg, ctx := errgroup.WithContext(ctx)
 	forwardingStreams := make([]pepb.PublishBuildEvent_PublishBuildToolEventStreamClient, 0)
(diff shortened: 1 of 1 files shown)


server/build_event_protocol/build_event_server/build_event_server.go
```diff
@@ -87,11 +87,18 @@ func (s *BuildEventProtocolServer) PublishBuildToolEventStream(stream pepb.Publi
 
 	eg, ctx := errgroup.WithContext(ctx)
 	forwardingStreams := make([]pepb.PublishBuildEvent_PublishBuildToolEventStreamClient, 0)
+	var closeStreamsOnce sync.Once
+	defer closeStreamsOnce.Do(func() { closeForwardingStreams(forwardingStreams) })
```
bduffany, line 91: moved this `defer` up here in case we ever support multiple forwarding streams in synchronous mode - an error in one stream would mean that we `return err`, so we need to clean up any other streams that got created

siggisim (approved):

PR #7582, opened 2024-09-26
enable init_on_alloc and init_on_free in firecracker vms

jdhollen:
theoretically this should help keep snapshot size down because we'll wipe freed memory back to zeroes (to help with squishing down pages / compression)

@tylerwilliams flagged this, i'm just mailing it

**Related issues**: N/A


tylerwilliams (approved):
you might want to put behind a flag and flip on only for dev?

tylerwilliams:
> you might want to put behind a flag and flip on only for dev?

or promise that it's ok to keep, that's fine too :)

jdhollen:
> you might want to put behind a flag and flip on only for dev?

sure sure

PR #7596, opened 2024-09-27
Return execution metadata even when execution fails

vanja-p:
This allows us to show information like the executor host ID when the executor fails due to some transient issue.

**Related issues**: http://go/b/937

Originally, I was going to use the Operation.metadata field for this, but I think it's clunky, and with a few small changes, it seems safe to return an Operation.result even when the execution failed.

The executions tab used to look like 
image

Now it looks like 
image

Note that the executor host ID is filled in.


bduffany (approved):
Looks great!

tylerwilliams (approved):
LGTM

PR #7602, opened 2024-09-30
proto: add sync to help bootstrap dev env easier

sluongng:
With this change, `bazel run proto:sync` will sync all the generated
code from bazel-out to our workspace.

Manage the pbsync go binary directly with Bazel in case the dev machine
has no Go sdk preinstalled.


tylerwilliams:
Could you add some context about why we want this? I don't think we should ever be checking generated protos into our repo, and I thought we already had several tools (written by Zoey and Vadim) to do this, but maybe I don't understand!

sluongng:
I think I was discussing with some users about how to set up our repository for local development to test their patches to BuildBuddy. An external user is not aware of `pbsync` helper to set up their BuildBuddy repo to be more `gopls` friendly, so having a `bazel run proto:sync` would be nice.

Gona close this for now as nothing has come up from that conversation 5-6 months ago.
We can always revive this if there are more demands.

PR #7603, opened 2024-09-30
explain: Various bug fixes and improvements

fmeum:
This PR gets explain to run on a large (~250MB) execution log provided by a customer.

* Added support for failed spawns and exit code changes
* Added support for invalid outputs
* By default, new/old only executions are no longer reported in detail
* With a new `--verbose` flag, top-level new/old only executions as well as certain "expected" executions (e.g. test XML generation, non-hermetic test logs) are listed
* Improved handling of test spawns, which have a large number of special cases related to optional output files and extra spawns
* Fixed a bug in the algorithm finding roots in the graph (Go doesn't fall through switch case labels by default)
* CPU profiling can be enabled via `--cpuprofile`

**Related issues**: N/A


fmeum:
Any idea why my PR has the `Test` workflow run out of disk space? Has it been a close call for a while?

bduffany:
@fmeum we don't clean up Bazel's cache automatically between runs, so I think over time it fills up and we have to clear out the disks manually. (Tracked in https://github.com/buildbuddy-io/buildbuddy-internal/issues/2069)

For now you can go to the "Workflows" tab then "Invalidate all VM snapshots" in the 3-dot menu for the repo.

bduffany (approved):

PR #7615, opened 2024-10-01
Add a ubuntu_2204 rbe image

tylerwilliams:
This is pretty much identical to the 20.04 image but I had to make the following changes:
 - install `python-is-python3` and remove the TODO
 - install `gpg-agent` and `mkdir -p /root/.gnupg` so it works
 - add a `-y` flag to add-apt-repository so it doesn't just hang when building the image

Diff:
```
tylerw@lunchbox:~/buildbuddy$ diff dockerfiles/rbe-ubuntu20-04/Dockerfile dockerfiles/rbe-ubuntu22-04/Dockerfile
1c1
< FROM marketplace.gcr.io/google/ubuntu2004@sha256:9e3ead2510253310bd5b44d254ab810c8f37b19a35716379220fa72b38e8c818
---
> FROM marketplace.gcr.io/google/ubuntu2204@sha256:b70be2d80535d3fe99e2bc484133da588728add8fc15956d124c478b42e61f2c
11a12
>       python-is-python3 \
24a26,33
> RUN apt-get update && \
>     apt-get install -y --no-install-recommends \
>     gpg-agent \
>     && \
>     apt-get clean && rm -rf /var/lib/apt/lists/* \
>     && \
>     mkdir -p mkdir -p /root/.gnupg
>
37c46
<     add-apt-repository ppa:git-core/ppa && \
---
>     add-apt-repository ppa:git-core/ppa -y && \
92,95d100
<
< # Alias `python` to `python2` to be compatible with the Ubuntu 16.04 RBE image.
< # Note, after Ubuntu 20.04, we should probably install `python-is-python3` instead.
< RUN update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1
tylerw@lunchbox:~/buildbuddy$
```

I tested it very lightly by batching `enterprise/server/remote_execution/containers/firecracker/firecracker_test.go` to use this latest image and all tests passed. I then reverted this change so it's not part of this CL.


dockerfiles/rbe-ubuntu22-04/Dockerfile
```diff
@@ -0,0 +1,121 @@
+FROM marketplace.gcr.io/google/ubuntu2204@sha256:b70be2d80535d3fe99e2bc484133da588728add8fc15956d124c478b42e61f2c
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Python 2 and 3
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+      python2 \
+      python2-dev \
```
bduffany, line 9: wdyt about removing python2 and then only installing it later if it turns out to be needed?

```suggestion
# Python 3
RUN apt-get update && \
    apt-get install -y --no-install-recommends \
```

dockerfiles/rbe-ubuntu22-04/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+ENV CC=/usr/local/bin/clang
+ENV GCOV=/dev/null
+ENV LD_LIBRARY_PATH=/usr/local/lib
+
+ENV LC_ALL=C.UTF-8
+ENV LANGUAGE=C.UTF-8
+ENV LANG=C.UTF-8
+
+ENV PATH="$PATH:/usr/local/go/bin"
```
bduffany, line 119: We could probably remove all of these env vars, since actions should be setting any env vars they need via the toolchain

dockerfiles/rbe-ubuntu22-04/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+# JRE / JDK (headless).
+# Don't install recommended packages since that would include
+# some alsa/a11y packages which aren't needed for server installs.
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+      default-jre-headless \
+      default-jdk-headless \
+      && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
```
bduffany, line 24: I think we could remove Java from newer images - Bazel's remote JDK works well and is easy to use

```suggestion
```

dockerfiles/rbe-ubuntu22-04/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    gpg-agent \
+    && \
+    apt-get clean && rm -rf /var/lib/apt/lists/* \
+    && \
+    mkdir -p mkdir -p /root/.gnupg
```
bduffany, line 32: out of curiosity - why is gpg-agent needed?

also, typo:

```suggestion
    mkdir -p /root/.gnupg
```

dockerfiles/rbe-ubuntu22-04/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+# GCC, make
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+      build-essential \
+      && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
```
bduffany, line 39: as a heads up, if we want to make this image available as one of the default images, we'd also need to update buildbuddy-toolchain with the GCC version that gets installed here

enterprise/dockerfiles/rbe-ubuntu22-04-workflows/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM gcr.io/flame-public/rbe-ubuntu22-04@sha256:f068d42c574bff8e5c08620eb74d1cbda1c34c1e7f784252c5f5e675a0192409
+
+# Install bazelisk
+RUN curl -Lo /usr/local/bin/bazelisk https://github.com/bazelbuild/bazelisk/releases/download/v1.15.0/bazelisk-linux-amd64 && \
+    chmod +x /usr/local/bin/bazelisk
+
+# Pre-download/extract bazel so that Bazel can skip that work on first run,
+# at least for CI runs on the BB repo itself.
+RUN USE_BAZEL_VERSION=6.0.0 bazelisk version
```
bduffany, line 9: we could probably remove these - we provision our own version of bazelisk using the `ci_runner` and place it in PATH, and also Bazel 6.0.0 is out of date at this point (not to mention, this step is installing bazel only for the `root` user and workflows run by non-root as default now)

enterprise/dockerfiles/rbe-ubuntu22-04-workflows/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# Provision a non-root user named "buildbuddy" and set up passwordless sudo.
+# Non-root users are needed for some bazel toolchains, such as hermetic python.
+# Also add them to the docker group so they can use docker.
+RUN apt-get update && apt-get install -y sudo && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN useradd --create-home buildbuddy --groups sudo,docker && \
+    echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
```
bduffany, line 17: if we did these steps in the regular image instead, and removed the bazel installation steps, then we could avoid the need for a `-workflows` image which would be pretty nice

dockerfiles/rbe-ubuntu22-04/Dockerfile
tempoz: We need `gpg-agent` for adding keys to `apt`.

tempoz: Though I'm now thinking `gnupg` is the package doing that, and we install it during only the steps we need it.

enterprise/dockerfiles/rbe-ubuntu22-04-workflows/Dockerfile
tylerwilliams: Done.

dockerfiles/rbe-ubuntu22-04/Dockerfile
tylerwilliams: Ack.

dockerfiles/rbe-ubuntu22-04/Dockerfile
tylerwilliams: Done.

tylerwilliams (commented):
I think I've addressed all comments if you want to take another look at this.

dockerfiles/rbe-ubuntu22-04/Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+# Python 3
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+      python3 \
+      python3-dev \
+      python-is-python3 \
+      && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# gpc-agent (for add-apt-repository below)
```
tempoz, line 14:
```suggestion
# gpg-agent (for add-apt-repository below)
```

tempoz (approved):
The image itself works just fine, but it's worth noting that it won't work out-of-the-box with the current `buildbuddy-toolchain`, as it needs the toolchain to set `gcc-version` to 11.

Other than that, this looks good to me! (Though I want to remove the msan clang stuff for cross-platform to arm reasons, but I can handle that separately).

bduffany (approved):
LGTM

PR #7623, opened 2024-10-01
Draft PR for rbe-ubuntu-2204 so we can test building on arm

tempoz:
- Use new buildbuddy-toolchain that supports arm by default
- Remove hard-coded x86_64 bits
- Add rootDirs to tsconfig for new platforms

**Related issues**: N/A


tempoz:
This is obsoleted by later efforts.

PR #7624, opened 2024-10-01
server: Use a strict CSP for all pages

fmeum:
The CSP allows at most the current origin (and subdomains) as well as a few inline scripts pinned by hashes.

### DIFF
--- a/server/http/csp/csp.go
+++ b/server/http/csp/csp.go
@@ -1,0 +1,5 @@
+package csp
+
+// Nonce is the context key type for the per-request CSP nonce.
+// The associated value is always of type template.HTMLAttr.
+type Nonce struct{}
--- a/server/http/csp/csp.go
+++ b/server/http/csp/csp.go
@@ -1,10 +1,11 @@
 package csp
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 
-	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
+	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 )
 
 // Nonce is the context key type for the per-request CSP nonce.
@@ -16,8 +17,8 @@ const ReportingEndpoint = "/csp-report"
 var ReportingHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 	report, err := io.ReadAll(r.Body)
 	if err != nil {
-		alert.UnexpectedEvent("csp_violation_report_failure", err.Error())
+		log.CtxWarning(r.Context(), fmt.Sprintf("CSP report failure: %s", err))
 		return
 	}
-	alert.UnexpectedEvent("csp_violation", string(report))
+	log.CtxWarning(r.Context(), fmt.Sprintf("CSP violation: %s", string(report)))
 })
--- a/server/http/csp/csp.go
+++ b/server/http/csp/csp.go
@@ -1,7 +1,6 @@
 package csp
 
 import (
-	"fmt"
 	"io"
 	"net/http"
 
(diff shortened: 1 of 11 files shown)


server/http/interceptors/interceptors.go
```diff
@@ -37,15 +38,41 @@ import (
...
 
+const ContentSecurityPolicyHeader = "Content-Security-Policy"
+
+func GetContentSecurityPolicy(scriptHashes ...[sha256.Size]byte) string {
+	scriptSrcs := make([]string, 0, len(scriptHashes))
+	for _, scriptHash := range scriptHashes {
+		scriptSrcs = append(scriptSrcs, fmt.Sprintf("'sha256-%s'", base64.StdEncoding.EncodeToString(scriptHash[:])))
+	}
+	return strings.Join([]string{
+		"default-src 'self'",
```
fmeum, line 51: I clicked around in the app and it worked, but please double-check this list and try to think of other kinds of access it may need.

server/http/interceptors/interceptors.go
```diff
@@ -37,15 +38,41 @@ import (
...
+		"style-src 'self' https://fonts.googleapis.com/css",
+		// The fixed hash is for the inline script that injects the Google Tag Manager. Additional hashes can be
+		// added by the caller.
+		"script-src 'self' https://www.googletagmanager.com 'sha256-3dO3w7Y+Pvr8nNuM9JioBpuBGrxgMBNZhPSqYDD4qn4=' " + strings.Join(scriptSrcs, " "),
```
bduffany, line 55: nit: wdyt about using a [nonce-based](https://web.dev/articles/strict-csp#structure) approach instead of hash-based? I am not an expert in CSP but it seems like nonce-based might be a little easier to maintain since we don't have to update these hashes if the contents change

fmeum: A hash-based approach generally has slightly better security guarantees (think `</script>` injection into server-generated data), but they don't apply to our current usage. It also makes the page itself static, which can help with caching.

A nonce-based approach doesn't require updating the hashes, which are currently only two which are probably very stable, but you are right that this could turn out to be more maintainable in the future. 

It's a close call for me. I did update the PR to use the nonce-based approach since it makes it easier for users to customize the index page: they just need to add `{{.Nonce}}` to their `script` tags rather than having to adding hashes to the Go constant.

bduffany (approved):

server/static/static.go
```diff
@@ -213,12 +216,14 @@ func serveIndexTemplate(ctx context.Context, env environment.Env, tpl *template.
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	nonce := interceptors.SetContentSecurityPolicy(w.Header())
```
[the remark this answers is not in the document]
fmeum: Yes, this was wastefully generating two nonces for the index page, with only the latter CSP header being sent. I am now storing the nonce in the context, which is indeed much nicer. Thanks for the suggestion!

fmeum:
@bduffany I added reporting to a new endpoint and also switched to report only mode when I observed that we may be using a [dep that unnecessarily uses `eval`](https://github.com/protobufjs/protobuf.js/issues/1754). We can enable this in dev and react to reports (or decide to weaken the policy).

server/http/csp/csp.go
```diff
@@ -0,0 +1,23 @@
...
+	"net/http"
+
+	"github.com/buildbuddy-io/buildbuddy/server/util/alert"
+)
+
+// Nonce is the context key type for the per-request CSP nonce.
+// The associated value is always of type template.HTMLAttr.
+type Nonce struct{}
+
+const ReportingEndpoint = "/csp-report"
```
bduffany, line 14: it makes me a little nervous that anyone could `curl` this endpoint and trigger an alert for us - maybe just `log.CtxInfo()` for now?

fmeum: Updated to `log.CtxWarning`

bduffany: Warning-level logs are considered errors in our canarying process, and can trigger the apps to be rolled back if too many warnings are logged - we might want to use `Info` until we're more confident that there shouldn't be any CSP violations

bduffany (approved):
LGTM (if we change Warning => Info)

server/http/csp/csp.go
fmeum: TIL, thanks!

PR #7628, opened 2024-10-02
Use unique index instead of table lock for usage inserts

bduffany:
The table locking approach doesn't play nicely with usage queries running concurrently; switch to a unique index for now.

It's harder to add new label columns when using a unique index, but we can think of alternate approaches as part of migrating this table to ClickHouse. For example, we could hash the usage labels and only include that hash in the unique index, rather than each individual label column.


bduffany:
(This approach won't work because we're hitting the 3072 byte limit on unique index size)

PR #7630, opened 2024-10-02
ociruntime: handle images with high layer count

sluongng:
When the action required an image with more than 20 layers, our mount
will fail with

```
create OCI bundle: create rootfs: mount overlayfs: no such file or directory
```

After some digging, it seems like 20 is the current limit of the number
of lowerdir allowed in each mount call.

Add special logic to break down images with more than 20 layers into
groups of 20. For each group, create an overlayfs mount called
"merged<group-id>" in the same bundle dir. The final overlayfs will then
be composed of these "merged" groups as lowerdirs.

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -68,10 +68,8 @@ const (
 	// directories can be cleaned up.
 	imageCacheVersion = "v1" // TODO: add automatic cleanup if this is bumped.
 
-	// The kernel's page size determines the maximum length of the "mount" syscall's options string.
-	// Actual value could be different on different systems, but 4096 is a common value.
-	// Use "getconf PAGE_SIZE" to check.
-	kernelPageSize = 4096
+	// Maximum length of overlayfs mount options string.
+	maxMntOptsLength = 4095
 )
 
 //go:embed seccomp.json
@@ -601,7 +599,7 @@ func (c *ociContainer) createRootfs(ctx context.Context) error {
 			len(mergedWorkdir)+len(mergedUpperdir),
 			len(workdir)+len(upperdir),
 		)
-		if len(newLowerDirs) == 1 || mntOptsLen <= kernelPageSize {
+		if len(newLowerDirs) == 1 || mntOptsLen <= maxMntOptsLength {
 			lowerDirs = newLowerDirs
 			continue
 		}
@@ -639,8 +637,8 @@ func (c *ociContainer) createRootfs(ctx context.Context) error {
 	// TODO: do this mount inside a namespace so that it gets removed even if
 	// the executor crashes (also needed for rootless support)
 	options := fmt.Sprintf(optionsTpl, strings.Join(lowerDirs, ":"), upperdir, workdir)
-	if len(options) > kernelPageSize {
-		return fmt.Errorf("mount options too long: %d / %d. Consider using container image with fewer layers.", len(options), kernelPageSize)
+	if len(options) > maxMntOptsLength {
+		return fmt.Errorf("mount options too long: %d / %d. Consider using container image with fewer layers.", len(options), maxMntOptsLength)
 	}
 	log.CtxDebugf(ctx, "Mounting overlayfs to %q, options=%q, length=%d", c.rootfsPath(), options, len(options))
 	if err := unix.Mount("none", c.rootfsPath(), "overlay", 0, options); err != nil {
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -619,6 +619,9 @@ func (c *ociContainer) createRootfs(ctx context.Context) error {
 		slices.Reverse(lowerDirs)
 		mntOpts := fmt.Sprintf(optionsTpl, strings.Join(lowerDirs, ":"), mergedUpperdir, mergedWorkdir)
 		log.CtxDebugf(ctx, "Mounting merged overlayfs to %q, options=%q, len=%d", merged, mntOpts, len(mntOpts))
+		if len(mntOpts) > maxMntOptsLength {
+			return fmt.Errorf("mount options too long: %d / %d. Consider using container image with fewer layers.", len(mntOpts), maxMntOptsLength)
+		}
 		if err := unix.Mount("none", merged, "overlay", 0, mntOpts); err != nil {
 			return fmt.Errorf("mount overlayfs: %w", err)
 		}
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -67,6 +67,11 @@ const (
 	// backwards-compatible changes to image cache storage, and older version
 	// directories can be cleaned up.
 	imageCacheVersion = "v1" // TODO: add automatic cleanup if this is bumped.
+
+	// The kernel's page size determines the maximum length of the "mount" syscall's options string.
+	// Actual value could be different on different systems, but 4096 is a common value.
+	// Use "getconf PAGE_SIZE" to check.
+	kernelPageSize = 4096
 )
 
 //go:embed seccomp.json
@@ -559,8 +564,23 @@ func (c *ociContainer) createRootfs(ctx context.Context) error {
 	if !ok {
 		return fmt.Errorf("bad state: attempted to create rootfs before pulling image")
 	}
+
+	// Create workdir and upperdir.
+	workdir := filepath.Join(c.bundlePath(), "tmp", "rootfs.work")
+	if err := os.MkdirAll(workdir, 0755); err != nil {
+		return fmt.Errorf("create overlay workdir: %w", err)
+	}
+	upperdir := filepath.Join(c.bundlePath(), "tmp", "rootfs.upper")
+	if err := os.MkdirAll(upperdir, 0755); err != nil {
+		return fmt.Errorf("create overlay upperdir: %w", err)
+	}
+
+	// - userxattr is needed for compatibility with older kernels
+	// - volatile disables fsync, as a performance optimization
+	optionsTpl := "lowerdir=%s,upperdir=%s,workdir=%s,userxattr,volatile"
+	tplLen := len(optionsTpl) - 3*len("%s")
 	var lowerDirs []string
-	for i, layer := range image.Layers {
+	for _, layer := range image.Layers {
 		path := layerPath(c.imageCacheRoot, layer.DiffID)
 		// Skip empty dirs - these can cause conflicts since they will always
 		// have the same digest, and also just add more overhead.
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
```diff
@@ -922,6 +922,41 @@ func TestOverlayfsEdgeCases(t *testing.T) {
 	assert.Equal(t, 0, res.ExitCode)
 }
 
+func TestOverlayfsHighLayerCount(t *testing.T) {
+	setupNetworking(t)
+
+	image := "ghcr.io/avdv/nix-build@sha256:5f731adacf7290352fed6c1960dfb56ec3fdb31a376d0f2170961fbc96944d50"
```
bduffany, line 928: The layering logic is starting to get complicated - it's kind of hard to understand the layer ordering now that the "groups" have to be in reverse-manifest-order and so do the individual layers within those merged groups.

So I feel like it'd be better to have a more explicit test on how we're handling the layers, rather than an integration test. For example, create a fake slice of 30 `image.Layer` structs, pass that to a func like `GetOverlayMounts()` which returns some data needed to create the overlays, then assert that it returns 2 dirs, one with 10 layers and another with 20 layers (and make sure the order of the layers is correct). The "20" const could also be a parameter and it could be set to 3 to make the test easier to follow.

Doing a more specialized test like this would also let us avoid depending on this heavyweight image to verify correctness

sluongng: Yeah with the latest version of this PR, I do think that the logic is a bit convoluted (multiple slices.Reverse calls) and could use some tests.

I will look for a way to export part of `createRootfs()` logic for testing.

sluongng: I gave this a try and when I tried to do the refactoring, making the test inspecting the directory and mount structure felt... wrong? It felt like we were testing the implementation details, which would lock us out of changing it up in the future.

Instead, I think we should just test the higher-level effect which matters to our users.
So I added a test that creates X layers on top of busybox container image.
Each layer includes an `a.txt` file with a different value and we simply assert that the last layer wins by running `cat /a.txt`.

Let me know what you think.

sluongng:
Hmm it seems like the test is failing at 1 + 21 layers. (my test is working! 🎉 )

I will put the PR in draft so I can investigate the implementation a bit more.

sluongng:
Turn out it's just an extra `slices.Reverse` that I need to remove 👀 

The tests all passed now. I also ran the old test against the Nix image manually and it passed as well.

sluongng:
What is curious though is with the current test setup, I can create an oci image with 2000 layers and it would not fail. So I wonder if the problem was not only layer count alone but perhaps a combination of layer count and layer content 🤔

I did not dig into it too much, but I left a TODO there to investigate further in the future.

bduffany:
> What is curious though is with the current test setup, I can create an oci image with 2000 layers and it would not fail. So I wonder if the problem was not only layer count alone but perhaps a combination of layer count and layer content

hm where did you see that the layer limit is 20? I can't find much documentation on this

sluongng:
It was not documented. I was simply testing against the Nix image previously and noticed it would start failing from the 21st layer. I inspected the image and found nothing special with the 20th or 21st layers so I concluded that there was a limit in the lower level, assuming overlayfs mount. 🤔

bduffany:
maybe we should try a bit more experimentation to root-cause the issue before going forward with this PR? (It doesn't seem super urgent to get these huge nix images working right? if it's not fixing an urgent issue then I'd sort of prefer to make sure we fully understand the root cause before adding a workaround)

sluongng:
yeah, let me dig into this a bit more.

I was surprised by the new test passed.

sluongng:
@bduffany I split up the test and the fix into 2 separate commits.
Without the fix, my test seems to be failing consistently starting from 21 layers now 🤔 

Example: https://buildbuddy.buildbuddy.io/invocation/b3ff7e6d-676b-456f-900f-68823ff6957d

sluongng:
From the firecracker log
```
[    9.938631] overlayfs: failed to resolve '/workspace/_tmp/c904fefdd4370b46205df2f829f288f0/buildbuddy-test-3965904397/executor/oci/run/2d3c4dd71e3c2570c6edbc0238e3dbcfdf9c26b': -2
[   10.513903] overlayfs: failed to resolve '/workspace/_tmp/c904fefdd4370b46205df2f829f288f0/buildbuddy-test-3496524003/executor/oci/images/v1/sha256/392d0f9da58e2e07bb713eb270d8c2002bde94d67059a88': -2
```

It seems like there is a character limit when calling `mount` syscall.
And with our current approach, the limit is around 20-21 dirs.
From then, the syscall args will get cut off, resulting in an incomplete path.
For example, this

```
...
upperdir=/workspace/_tmp/c904fefdd4370b46205df2f829f288f0/buildbuddy-test-3965904397/executor/oci/run/2d3c4dd71e3c2570c6edbc0238e3dbcfdf9c26b70d96a933b99572f77ea39b2f/tmp/rootfs.upper,
workdir=...
```

was turned into
```
...
upperdir=/workspace/_tmp/c904fefdd4370b46205df2f829f288f0/buildbuddy-test-3965904397/executor/oci/run/2d3c4dd71e3c2570c6edbc0238e3dbcfdf9c26b
```

bduffany:
Ah, nice find. From this issue it seems like the limit might be `PAGESIZE` https://github.com/moby/moby/issues/26380

bduffany:
In my testing it looks like the limit is 4095, which might be explained by the string needing a 0 byte at the end as a null terminator.

https://gist.github.com/bduffany/21a67c69407cb7d6813b25b4cca0f286

So I think we should directly track the total length of the mount options, and if any given lowerdir would push the mount option length over 4095, then we can create an intermediate lowerdir.

sluongng:
Yeah I did a bit of digging on podman side as well. Similar optimization

https://github.com/containers/storage/blob/250a170db0630f521bfb56548c5e7c5c76527205/drivers/overlay/overlay.go#L93-L102

I think the longer-term solution here is to use a shorter parent path for all these image layers. I think the image layers could be stored in a CAS-like manner as well to help deduplicating some.

bduffany:
> I think the longer-term solution here is to use a shorter parent path for all these image layers.

Just to clarify, are you planning to implement this now? I think this would be good to do eventually, but think we should stick with the current approach of combining lowerdirs, since we'll want to do that either way - since even if we use shorter parent paths, we can still run into the 4095-char limit.

> I think the image layers could be stored in a CAS-like manner as well to help deduplicating some.

FYI we already store layers using the `DiffID` which uniquely identifies the layer contents.

sluongng:
> Just to clarify, are you planning to implement this now?

No. I will go with your suggestion.
Going with the 4095 char limits sounds good to me.

> FYI we already store layers using the DiffID which uniquely identifies the layer contents.

Ah neat! 👍

sluongng:
a few notes since the last version:

- I switched to using `unix.Mount` instead of `syscall` because the `x/sys/unix` API following the recommendation of the syscall doc https://pkg.go.dev/syscall

- I wanted to add a test for 2000 layers to test the error when too many layers are used but it seems like our firecracker test would consistently fail for some other reason(?). Because other oci runtimes only support up-to 128 layers, I decided to test for 128 instead and call it.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -67,6 +67,11 @@ const (
 	// backwards-compatible changes to image cache storage, and older version
 	// directories can be cleaned up.
 	imageCacheVersion = "v1" // TODO: add automatic cleanup if this is bumped.
+
+	// The kernel's page size determines the maximum length of the "mount" syscall's options string.
+	// Actual value could be different on different systems, but 4096 is a common value.
+	// Use "getconf PAGE_SIZE" to check.
+	kernelPageSize = 4096
```
bduffany, line 74: in my testing, the mount options string has to be <= 4095 chars long (4096 results in an error)

Also, I'm not 100% sure if the limit is actually related to page size or if it's just arbitrarily hard-coded somewhere to be 4K. We could check the kernel source to verify, or for now just call this `mountOptionsLengthLimit = 4095` and avoid mentioning page size for now.

enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
```diff
@@ -34,6 +41,14 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/buildbuddy-io/buildbuddy/server/util/testing/flags"
 	"github.com/buildbuddy-io/buildbuddy/server/util/uuid"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/registry"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
bduffany, line 46: nit: instead of aliasing to `v1` I think an alias of `containerregistry` or something shorter like `registry` would be more clear.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: changed this to `maxMntOptsLength` to avoid mentioning the kernel's page size.

enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
sluongng: renamed

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -567,28 +590,58 @@ func (c *ociContainer) createRootfs(ctx context.Context) error {
...
+		if err := os.MkdirAll(mergedUpperdir, 0755); err != nil {
+			return fmt.Errorf("create overlay upperdir: %w", err)
+		}
+		merged := filepath.Join(c.bundlePath(), "tmp", fmt.Sprintf("merged%d", len(c.mergedMounts)))
+		if err := os.MkdirAll(merged, 0755); err != nil {
+			return fmt.Errorf("create overlay merged: %w", err)
+		}
+		slices.Reverse(lowerDirs)
+		mntOpts := fmt.Sprintf(optionsTpl, strings.Join(lowerDirs, ":"), mergedUpperdir, mergedWorkdir)
+		log.CtxDebugf(ctx, "Mounting merged overlayfs to %q, options=%q, len=%d", merged, mntOpts, len(mntOpts))
```
bduffany, line 621: nit: mind adding a sanity check here that the mount options length is <= 4095, and returning an error? In case we have some logic issue above (in the future), or in case we have a single mount dir that is really long, it would be nice to return a clear error message instead of falling back to the "no such file or directory" error

enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
```diff
@@ -922,6 +940,128 @@ func TestOverlayfsEdgeCases(t *testing.T) {
...
+					content:   b.Bytes(),
+				})
+				require.NoError(t, err)
+				layers = append(layers, layer)
+			}
+			testImg, err := mutate.AppendLayers(busyboxImg, layers...)
+			require.NoError(t, err)
+
+			// Start registry and push our new image there
+			ts := httptest.NewServer(registry.New(registry.Logger(log.New(io.Discard, "", 0))))
```
bduffany, line 1030: need a `defer ts.Close()` here?

bduffany (approved):

enterprise/server/remote_execution/containers/ociruntime/ociruntime_test.go
sluongng: Good call. Initially I was playing around with our own `testhttp` package which use `t.Cleanup` automatically. Later on I switched to the stdlib `httptest` and forgot about `Close()` 🤦

PR #7635, opened 2024-10-02
Make use-self-hosted-executors=false work as expected

bduffany:
Context: https://buildbuddy-corp.slack.com/archives/C05GZM1EGM8/p1727895825533649

Our expectation (as well as one of our users' expectations) was that setting `use-self-hosted-executors=false` would override the org-level "Default to self-hosted executors" setting, but it turns out that this override behavior only works in one direction (i.e. when setting `use-self-hosted-executors=true`).

This PR makes it work as expected by updating the logic to consider whether `use-self-hosted-executors` was set explicitly or not.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -986,7 +986,10 @@ func NewSchedulerServerWithOptions(env environment.Env, options *Options) (*Sche
 	return s, nil
 }
 
-func (s *SchedulerServer) GetPoolInfo(ctx context.Context, os, requestedPool, workflowID string, useSelfHosted bool) (*interfaces.PoolInfo, error) {
+func (s *SchedulerServer) GetPoolInfo(ctx context.Context, os, requestedPool, workflowID string, useSelfHosted *bool) (*interfaces.PoolInfo, error) {
+	requestedSelfHosted := useSelfHosted != nil && *useSelfHosted
+	requestedShared := useSelfHosted != nil && !*useSelfHosted
+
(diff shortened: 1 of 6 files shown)


siggisim (approved):

iain-macdonald (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -986,7 +986,10 @@ func NewSchedulerServerWithOptions(env environment.Env, options *Options) (*Sche
 	return s, nil
 }
 
-func (s *SchedulerServer) GetPoolInfo(ctx context.Context, os, requestedPool, workflowID string, useSelfHosted bool) (*interfaces.PoolInfo, error) {
+func (s *SchedulerServer) GetPoolInfo(ctx context.Context, os, requestedPool, workflowID string, useSelfHosted *bool) (*interfaces.PoolInfo, error) {
```
tylerwilliams, line 989: is a *bool better than making an enum type and passing that here?

I think you could define an enum like this:
```
const (
  selfHostedPoolRequested = iota
  sharedPoolRequested
)
```

then parse your platform property into that enum and pass it.

bduffany: I prefer `*bool` over a custom enum type since `use-self-hosted-executors` is an optional bool (unset/nil, false, or true).

I could see `*bool` being considered a little unusual though so I threw an enum into `interfaces.go`, lmk what you think

tylerwilliams (approved):
LGTM (thanks!)

PR #7638, opened 2024-10-02
[RB] Add public API

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/proto/api/v1/remote_runner.proto
+++ b/proto/api/v1/remote_runner.proto
@@ -34,6 +34,16 @@ message RunRequest {
   // If true, start the runner but do not wait for the action to be scheduled
   // before returning a response.
   bool async = 8;
+
+  // Whether to use github credentials configured on the executor.
+  //
+  // By default, we require https for git operations and generate short-lived
+  // access tokens using our github app installation.
+  // If GitHub SSH access is already configured on the runners, set this
+  // to true to skip doing that and use the configured auth.
+  //
+  // This is only supported for bare runners.
+  bool use_system_git_credentials = 9;
 }
 
 message Step {
--- a/proto/api/v1/remote_runner.proto
+++ b/proto/api/v1/remote_runner.proto
@@ -17,6 +17,7 @@ message RunRequest {
   string commit_sha = 3;
 
   // Bash commands to run, in order.
+  // If a step fails, subsequent steps are not run.
   // Ex. [{"run": "bazel build :server"}, {"run": "echo \"Hello World\""}]
   repeated Step steps = 4;
 
@@ -24,7 +25,7 @@ message RunRequest {
   map<string, string> env = 5;
 
   // Platform properties to apply to the runner.
-  // Ex. ["OSFamily":"linux"}, {"Arch":"amd64"}]
+  // Ex. {"OSFamily":"linux", "Arch":"amd64"}
   map<string, string> platform_properties = 6;
 
   // Max time before run should be canceled.
--- a/proto/api/v1/remote_runner.proto
+++ b/proto/api/v1/remote_runner.proto
@@ -1,0 +1,46 @@
+syntax = "proto3";
+
+package api.v1;
+
+import "google/rpc/status.proto";
+
+message RemoteRunRequest {
+  // URL of the repo the workflow is running for
+  // Ex. "https://github.com/some-user/acme"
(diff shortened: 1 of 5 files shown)


proto/api/v1/remote_runner.proto
```diff
@@ -0,0 +1,47 @@
...
+  // Bash commands to run, in order.
+  // Ex. [{"run": "bazel build :server"}, {"run": "echo \"Hello World\""}]
+  repeated Step steps = 4;
+
+  // Environment variables to set in the runner env.
+  map<string, string> env = 5;
+
+  // Platform properties to apply to the runner.
+  // Ex. [{"name":"OSFamily","value":"linux"}, {"name":"Arch","value":"amd64"}]
+  repeated build.bazel.remote.execution.v2.Platform.Property
```
siggisim, line 29: I wonder if we should just do map<string, string> to make this easier to consume.

proto/api/v1/remote_runner.proto
```diff
@@ -0,0 +1,47 @@
...
+      platform_properties = 6;
+
+  // Max time before run should be canceled.
+  string timeout = 7;
```
siggisim, line 33: What's the format of this? Should we do int seconds instead?

siggisim (commented):
Love it! Couple of nits

proto/api/v1/remote_runner.proto
maggie-lou: It's a duration variable in the BE. Added a comment but let me know if you still think an int is better

proto/api/v1/remote_runner.proto
maggie-lou: Ack - changed

proto/api/v1/remote_runner.proto
bduffany: In `buildbuddy.yaml` we allow passing timeouts like `30m` etc. - I think it'd be nice to be able to specify it this way, if we document it. Timeout seconds also sounds good though

(edit: looks I replied the same time as Maggie)

siggisim (approved):

bduffany (approved):

proto/api/v1/remote_runner.proto
```diff
@@ -0,0 +1,56 @@
...
+
+  // Bash commands to run, in order.
+  // Ex. [{"run": "bazel build :server"}, {"run": "echo \"Hello World\""}]
+  repeated Step steps = 4;
+
+  // Environment variables to set in the runner env.
+  map<string, string> env = 5;
+
+  // Platform properties to apply to the runner.
+  // Ex. ["OSFamily":"linux"}, {"Arch":"amd64"}]
```
bduffany, line 27:
```suggestion
  // Ex. {"OSFamily":"linux", "Arch":"amd64"}
```

proto/api/v1/remote_runner.proto
```diff
@@ -0,0 +1,56 @@
...
+  string repo = 1;
+
+  // Git refs to configure the remote git workspace (at least one of `branch` or
+  // `commit_sha` must be set). If only `branch` is set, will run from the tip
+  // of the branch. If only `commit_sha` is set, reporting will not contain the
+  // branch name.
+  string branch = 2;
+  string commit_sha = 3;
+
+  // Bash commands to run, in order.
```
bduffany, line 19:
```suggestion
  // Bash commands to run, in order.
  // If a step fails, subsequent steps are not run.
```

proto/api/v1/remote_runner.proto
```diff
@@ -0,0 +1,56 @@
...
+  // access tokens using our github app installation.
+  // If GitHub SSH access is already configured on the runners, set this
+  // to true to skip doing that and use the configured auth.
+  //
+  // This is only supported for bare runners.
+  bool use_system_git_credentials = 9;
+}
+
+message Step {
+  string run = 1;
```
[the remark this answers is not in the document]
maggie-lou: Two slight preferences for run: 1. It can be a bash block and doesn't have to be a single command 2. 'run' is what github actions uses 

I don't feel strongly though. What do you think?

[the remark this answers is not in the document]
maggie-lou: Going to merge as is for now - but let me know if you feel strongly about changing it

PR #7639, opened 2024-10-02
client changes to use user-selected date range for flakes.

jdhollen:
- Splits out the filter date picker to a separate component.
- When the user clicks a "recently flaky" button, we still default them to the 7-day view.

**Related issues**: N/A


enterprise/app/tap/flakes.tsx
```diff
@@ -215,10 +243,7 @@ export default class FlakesComponent extends React.Component<Props, State> {
       <div className="container">
         <h3 className="flakes-list-header">Sample flakes for {targetLabel}</h3>
         {!this.state.pendingFlakeSamplesRequest && !(this.state.flakeSamples?.samples.length ?? 0) && (
-          <div>
-            No samples found--if it looks like there have been flakes in the last seven days, their logs may have
-            expired from the remote cache.
-          </div>
+          <div>No samples found--their logs may have expired from the remote cache.</div>
```
bduffany, line 246:
```suggestion
          <div>No samples found. Their logs may have expired from the remote cache.</div>
```

bduffany (approved):

PR #7646, opened 2024-10-03
integration/podman: pin busybox image to stable

sluongng:
The unstable version has some weird behavior that affect how signal is
handled and failed our TestSignal test consistently.

Let's downgrade our image from latest to stable and pin it with a
SHA256.


bduffany (approved):
LGTM - mind adding this TODO line to `TestSignal`?

```
// TODO(http://go/b/3924): Figure out why this test fails on newer versions of busybox.
```

PR #7648, opened 2024-10-03
Enable reader/writer roles by default

bduffany:


siggisim (approved):

bduffany:
+Tyler since this is an auth change

tylerwilliams (approved):

PR #7650, opened 2024-10-03
server: Allow Monaco's font loading in the CSP

fmeum:


bduffany (approved):

fmeum:
@bduffany When I go to `app.buildbuddy.dev` and open the Developer Tools, I still see a CSP violation caused by the `eval` in `@protobuf.js/inquire` that I thought I had patched out. You can still find it in https://storage.googleapis.com/buildbuddy-static-dev/release/db8a1a21797d901f1379696140c534ad2d999d88/app_bundle/app.js though if you search for `eval(`. Is the bundle updated less regularly than the Go server (the CSP has already been updated with the changes in my previous PR)?

bduffany:
@fmeum The Go server + app release are kept in lock-step (according to the git commit SHA). We are doing a release today and also various people might be pushing dev for testing, so that could potentially be the reason. Do you still see that `eval()` string in the JS bundle if you run the app locally? If not, it's probably just dev being stale

fmeum:
I don't see it locally, I'll just wait for the server to update.

PR #7651, opened 2024-10-03
[RB] Implement public API

maggie-lou:
**Related issues**: N/A


enterprise/server/api/api_server.go
```diff
@@ -576,5 +580,38 @@ func (s *APIServer) ExecuteWorkflow(ctx context.Context, req *apipb.ExecuteWorkf
...
+	for _, s := range req.GetSteps() {
+		steps = append(steps, &rnpb.Step{Run: s.Run})
+	}
+	execProps := make([]*repb.Platform_Property, 0, len(req.GetPlatformProperties()))
+	for k, v := range req.GetPlatformProperties() {
+		execProps = append(execProps, &repb.Platform_Property{
+			Name:  k,
+			Value: v,
+		})
+	}
```
bduffany, line 598: req.GetPlatformProperties() will have the props in an arbitrary order - idk if the internal `Run()` API already sorts platform properties by key internally, but if not, we should do that, since it will affect caching / snapshot key etc. (and maybe also make a note of it here)

bduffany (approved):

enterprise/server/api/api_server.go
maggie-lou: The internal Run API does sort them already (and resolve duplicates)

PR #7652, opened 2024-10-03
server: Allow connections to region servers in the CSP

fmeum:
Also add more domains for Google Analytics.

### DIFF
--- a/server/http/interceptors/interceptors.go
+++ b/server/http/interceptors/interceptors.go
@@ -45,9 +45,7 @@ var (
 
 const contentSecurityPolicyReportingEndpointName = "csp-endpoint"
 
-var contentSecurityPolicyTemplate = createContentSecurityTemplate()
-
-func createContentSecurityTemplate() string {
+func getContentSecurityPolicyHeaderValue(nonce string) string {
 	var regionConnectSrcs []string
 	for _, r := range region.Protos() {
 		regionConnectSrcs = append(regionConnectSrcs, r.Subdomains)
@@ -70,8 +68,7 @@ func createContentSecurityTemplate() string {
 		"report-to " + contentSecurityPolicyReportingEndpointName,
 		"report-uri " + csp.ReportingEndpoint,
 		// libsodium.js requires 'wasm-unsafe-eval' to avoid a fallback to asm.js.
-		// KEEP THIS LAST as setContentSecurityPolicy appends the nonce.
-		"script-src 'self' 'strict-dynamic' 'wasm-unsafe-eval'",
+		fmt.Sprintf("script-src 'self' 'strict-dynamic' 'wasm-unsafe-eval' 'nonce-%s'", nonce),
 	}, ";")
 }
 
@@ -83,7 +80,7 @@ func setContentSecurityPolicy(h http.Header) template.HTMLAttr {
 	}
 	nonce := base64.StdEncoding.EncodeToString(nonceBytes)
 	// TODO: Enable this by dropping the "-Report-Only" suffix.
-	h.Set("Content-Security-Policy-Report-Only", contentSecurityPolicyTemplate+fmt.Sprintf(" 'nonce-%s'", nonce))
+	h.Set("Content-Security-Policy-Report-Only", getContentSecurityPolicyHeaderValue(nonce))
 	return template.HTMLAttr(fmt.Sprintf(`nonce="%s"`, nonce))
 }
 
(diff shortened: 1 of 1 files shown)


server/http/interceptors/interceptors.go
```diff
@@ -65,21 +67,34 @@ var contentSecurityPolicyTemplate = strings.Join([]string{
...
+		} else {
+			eventsAPIOrigin = "http://" + host
+		}
+	} else {
+		eventsAPIOrigin = eventsAPIURL.Scheme + "//:" + eventsAPIURL.Host
+	}
+
 	// TODO: Enable this by dropping the "-Report-Only" suffix.
-	h.Set("Content-Security-Policy-Report-Only", fmt.Sprintf(contentSecurityPolicyTemplate, nonce))
+	h.Set("Content-Security-Policy-Report-Only", fmt.Sprintf(contentSecurityPolicyTemplate, nonce, eventsAPIOrigin))
```
bduffany, line 91: nit: I missed this earlier, but could we build the CSP header using a func that returns a string (like `getContentSecurityPolicy(nonce, eventsAPIOrigin string) string`)  then do string concatenation to fill in the parameterized values? The `%s` approach is a bit difficult to follow because the string is so long, so the reader has to hunt to find the `%s` occurrences and match them up with the parameters here.

server/endpoint_urls/events_api_url/events_api_url.go
```diff
@@ -15,3 +15,12 @@ func WithPath(path string) *url.URL {
 func String() string {
 	return eventsAPIURL.String()
 }
+
+func URL() *url.URL {
+	if eventsAPIURL == nil {
+		return nil
+	}
+	// Make a copy so the caller can't accidentally modify the original.
+	u := *eventsAPIURL
```
bduffany, line 24: IIUC `eventsAPIURL` is a native gRPC endpoint (e.g. remote.buildbuddy.io) and the app doesn't speak gRPC, so I think the app shouldn't need to communicate with this URL?

fmeum: When I visit https://app.buildbuddy.dev and go to `Executions`, I get CSP violations for `https://events.buildbuddy.dev/rpc/...`. Is that unexpected?

bduffany: (I think you meant `Executors`? I was able to repro it there but not the Executions tab)

I think it's because of this "Test region" stuff - @siggisim can we remove this?

https://github.com/buildbuddy-io/buildbuddy-internal/blob/446e52484f833a13c42ea063e1e2c8b13be61015/enterprise/config/buildbuddy.dev.yaml#L284-L286

siggisim: Yeah we can remove that

bduffany: Alternatively, I think the correct thing to do is to look at that "regions" config and allowlist each of the configured subdomains.

server/http/interceptors/interceptors.go
```diff
@@ -45,26 +45,35 @@ var (
...
+		"frame-ancestors 'none'",
+		"base-uri 'none'",
+		"block-all-mixed-content",
+		// libsodium.js requires data: for wasm.
+		"connect-src 'self' data: https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com " + strings.Join(regionConnectSrcs, " "),
+		"report-to " + contentSecurityPolicyReportingEndpointName,
+		"report-uri " + csp.ReportingEndpoint,
+		// libsodium.js requires 'wasm-unsafe-eval' to avoid a fallback to asm.js.
+		// KEEP THIS LAST as setContentSecurityPolicy appends the nonce.
+		"script-src 'self' 'strict-dynamic' 'wasm-unsafe-eval'",
```
bduffany, line 74: nit: I think it's fine to pass the nonce in as a param to this function and call it for each request - we don't get a high volume of requests for the root UI endpoint (`/`) so the cost of rebuilding this string for each request should be negligible, and is worth the added simplicity IMO.

bduffany (approved):

server/http/interceptors/interceptors.go
bduffany: Yeah, it should be negligible - we can always profile later and optimize if needed, but I wouldn't expect this code path to show up in a CPU profile

PR #7660, opened 2024-10-04
Remove flakiness in remote_execution_test.TestActionMerging_Hedging

vanja-p:
In the previous version, it was possible for the 2 slow executions to be accepted and in the executing state, but not actually check for the existence of fname before the unlock command ran.

Now, there is synchronization between the scripts, so unlock will only run after the first 2 executions checked for the existence of fname.

### DIFF
--- a/enterprise/server/test/integration/remote_execution/remote_execution_test.go
+++ b/enterprise/server/test/integration/remote_execution/remote_execution_test.go
@@ -1812,16 +1812,19 @@ func TestActionMerging_Hedging(t *testing.T) {
 	rbe.AddBuildBuddyServer()
 	rbe.AddExecutor(t)
 
-	// This script runs quickly and outputs "FAST" if `fname` exists.
-	// Otherwise it runs slowly and outputs "SLOW".
+	// This script runs quickly and outputs "FAST" if `fname` exists. Otherwise
+	// it writes a line to `counter`, runs slowly, and then outputs "SLOW".
 	fname := fmt.Sprintf("/tmp/%s", uuid.New().String())
+	counter := fmt.Sprintf("/tmp/%s", uuid.New().String())
 	flakyScript := fmt.Sprintf(`FILE="%s"
+COUNTER="%s"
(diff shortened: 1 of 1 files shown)


enterprise/server/test/integration/remote_execution/remote_execution_test.go
```diff
@@ -1812,16 +1812,19 @@ func TestActionMerging_Hedging(t *testing.T) {
...
-	// Otherwise it runs slowly and outputs "SLOW".
+	// This script runs quickly and outputs "FAST" if `fname` exists. Otherwise
+	// it writes a line to `counter`, runs slowly, and then outputs "SLOW".
 	fname := fmt.Sprintf("/tmp/%s", uuid.New().String())
+	counter := fmt.Sprintf("/tmp/%s", uuid.New().String())
 	flakyScript := fmt.Sprintf(`FILE="%s"
+COUNTER="%s"
 if [ -f $FILE ]; then
-    echo "FAST"
+	echo "FAST"
```
iain-macdonald, line 1822: nit: Mind using four spaces instead of a tab for consistency with the rest of this file?

iain-macdonald (approved):
Thanks for fixing this! I swear this test wasn't flaky when I added it 😅

enterprise/server/test/integration/remote_execution/remote_execution_test.go
```diff
@@ -1860,7 +1873,7 @@ fi`, fname)
 	require.Equal(t, op1, op3, "expected action to be merged")
 
 	// Let the hedged execution finish.
-	time.Sleep(10 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
```
bduffany, line 1876: No need to do this right now (since this PR seems like an improvement over what we've got today) but instead of using shell scripts and coordinating this logic using sleeps (which are likely to fail under load), we could use `RunInterceptor` to write the execution logic in Go and do this coordination e.g. using channels.

PR #7680, opened 2024-10-07
[CI] Add better snapshot fallback behavior for merge queues

maggie-lou:
**Related issues**: Fixes https://buildbuddy.slack.com/archives/CUY16GNK1/p1728328827179879?thread_ts=1726242879.319709&cid=CUY16GNK1

### DIFF
--- a/enterprise/server/remote_execution/snaploader/snaploader.go
+++ b/enterprise/server/remote_execution/snaploader/snaploader.go
@@ -77,6 +77,32 @@ func (l *FileCacheLoader) SnapshotKeySet(ctx context.Context, task *repb.Executi
 		fallbackKey.Ref = ref
 		keys.FallbackKeys = append(keys.FallbackKeys, fallbackKey)
 	}
+
+	// We only write a snapshot for the pushed git branch.
+	// We do not update the snapshot for any fallback key(s) that we may have
+	// read from, i.e. ones corresponding to the PR's base branch or the repo's
+	// default branch.
+	writeKey := branchKey
+
+	// For merge queue branches, we should save the snapshot to the default branch (like `main`),
+	// to support our fallback branch behavior.
+	//
+	// If someone is using merge queues, they're unlikely to run CI on the default
+	// branch itself, so there will never be a fallback branch hit. Instead, we should
+	// fallback to the latest merge queue snapshot.
+	if strings.HasPrefix(branchRef, "gh-readonly-queue") {
+		writeKey = branchKey.CloneVT()
+		// We're expecting branches in the format `gh-readonly-queue/<default branch name>/<pr branch name>
+		splitBranch := strings.Split(branchRef, "/")
+		if len(splitBranch) == 3 {
+			defaultBranch := splitBranch[1]
+			writeKey.Ref = defaultBranch
+		} else {
+			log.Errorf("Unexpected merge queue branch name %s: expected 3 '/' separated parts", branchRef)
+		}
+	}
+	keys.WriteKey = writeKey
+
 	return keys, nil
 }
 
@@ -573,8 +599,6 @@ func (l *FileCacheLoader) UnpackSnapshot(ctx context.Context, snapshot *Snapshot
 }
 
 func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotKey, opts *CacheSnapshotOptions) error {
-	key = generateMergeQueueSupportedKey(key)
-
 	vmConfig, err := anypb.New(opts.VMConfiguration)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -572,6 +573,8 @@ func (l *FileCacheLoader) UnpackSnapshot(ctx context.Context, snapshot *Snapshot
 }
 
 func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotKey, opts *CacheSnapshotOptions) error {
+	key = generateMergeQueueSupportedKey(key)
```
tylerwilliams, line 576: Why is it safe to only mutate this key inside of this one method?

enterprise/server/remote_execution/snaploader/snaploader.go
```diff
@@ -659,6 +662,33 @@ func (l *FileCacheLoader) CacheSnapshot(ctx context.Context, key *fcpb.SnapshotK
...
 }
 
+// For merge queue branches, we should save the snapshot to the default branch (like `main`),
+// to support our fallback branch behavior.
+//
+// By default, if there is not a snapshot for the current branch,
+// we will fallback to a snapshot for the default branch.
+// If someone is using merge queues, they're unlikely to run CI on the default
+// branch itself, so there will never be a fallback branch hit. Instead, we should
+// fallback to the latest merge queue snapshot.
```
bduffany, line 672: > If someone is using merge queues, they're unlikely to run CI on the default branch itself

This is a bit surprising - is this what we're seeing in practice?

enterprise/server/remote_execution/snaploader/snaploader.go
maggie-lou: Let me know if this description helps explain the logic

Incoming workflows triggered by webhook will automatically have fallback_branch=main

So a PR workflow might have { branch=myb, fallback=main }
A merge queue workflow might have { branch = mergeq, fallback=main }

1. With this PR, the merge queue workflow will save a snapshot to ref=main
2. An incoming PR workflow will not find a snapshot for ref=myb, so will fallback to ref=main (it will pick the snapshot from step #1) 
3. An incoming merge queue workflow will not find a snapshot for ref=mergeq, so will fall back to ref=main (it will pick the snapshot from step #1). It will then overwrite the snapshot at ref=main. This is good, as it represents the latest state of the merge queue, which should soon represent main (as long as CI plasses)

enterprise/server/remote_execution/snaploader/snaploader.go
maggie-lou: Talked offline but TLDR:

I figured that if a customer is running CI on a merge branch, which will functionally become their main branch immediately after CI passes, there is no need to then run CI again on the main branch as it will be the same

enterprise/server/remote_execution/snaploader/snaploader.go
bduffany: wdyt about asking the customer(s) to run CI on main instead? it addresses the conditional part of "as long as CI passes" because main should always be green, and it shouldn't add much to their costs because the build should be fully cached?

bduffany: (asking because the branching model for snapshotting is already pretty confusing, so I think the more we can do to reduce the amount of edge cases to consider, the better)

maggie-lou: Given that merge queues are a fairly common use case, I feel like we should support it out of the box and without requiring customers to pay more for it. I could see a user cleaning up the workflow on main (because it functionally doesn't do anything), and that unexpectedly breaking snapshot sharing, which is also unexpected. What do you think?

bduffany: I think overall, saving the snapshot to the "main" key seems fine. To Tyler's point, I wonder if this can be modeled in a more direct way - we have SnapshotKeySet which holds all the keys associated with the snapshot - could we maybe store some sort of "target" key there (i.e. the key that will be overwritten) and then pass it directly to this func?

maggie-lou: Updated to be more explicit

bduffany (approved):
Do we need to update the task router as well to be aware of this new logic? (Fine to do as a follow-up)

maggie-lou:
> Do we need to update the task router as well to be aware of this new logic? (Fine to do as a follow-up)

@bduffany I don't think any updates are required. We already have logic to route based on default branch (though it isn't enabled in prod), and the default branch hasn't changed with this PR. Are you thinking of anything specific?

bduffany:
@maggie-lou Thanks for checking that - I wasn't sure if any updates were needed. (Do you want to turn on that flag in prod?)

maggie-lou:
> @maggie-lou Thanks for checking that - I wasn't sure if any updates were needed. (Do you want to turn on that flag in prod?)

Yeah I got super side-tracked after adding that flag. I'm going to focus on performance soon and will revisit testing that flag in prod

PR #7685, opened 2024-10-08
proto: update Bazel protos

sluongng:
Update all relevant Bazel protos.

Skip on the deprecations removal in spawn.proto as it would break our
existing code, which supports the older compact exec log format.

### DIFF
--- a/proto/failure_details.proto
+++ b/proto/failure_details.proto
@@ -150,6 +150,7 @@ message FailureDetail {
     ModCommand mod_command = 183;
     BuildReport build_report = 184;
     Skyfocus skyfocus = 185;
+    RemoteAnalysisCaching remote_analysis_caching = 186;
   }
 
   reserved 102;         // For internal use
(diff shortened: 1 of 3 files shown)


proto/failure_details.proto
```diff
@@ -579,7 +596,7 @@ message Command {
     STARLARK_OPTIONS_PARSE_FAILURE = 10 [(metadata) = { exit_code: 2 }];
     ARGUMENTS_NOT_RECOGNIZED = 11 [(metadata) = { exit_code: 2 }];
     NOT_IN_WORKSPACE = 12 [(metadata) = { exit_code: 2 }];
-    SPACES_IN_WORKSPACE_PATH = 13 [(metadata) = { exit_code: 36 }];
```
fmeum, line 582: We should keep this as older Bazel versions can still throw it.

fmeum (approved):

proto/failure_details.proto
sluongng: Great call!

PR #7688, opened 2024-10-08
Remove some AnyPB unwrapping in ExecutionServer.PublishOperation

vanja-p:
When execution is completed, we used to call operation.ExtractExecuteResponse 3 times. Now it's called once.

Maybe this has some performance benefit, but I mostly made the change because I found it confusing that we are unpacking 3 times.


tylerwilliams (approved):
LGTM but maybe let @vadimberezniker have a look too before submitting.

vadimberezniker (approved):

PR #7699, opened 2024-10-09
Proxy kythe backend for client (UI)

tylerwilliams:
Plumbing!

Originally was not going to do this but it seems simpler than trying to modify the rpc_service.ts code to call other services.

There are other kythe services, but Jim's prototype only used these three so that's all I'm exposing for now:

```
type KytheService interface {
	Nodes(ctx context.Context, req *kgrpb.NodesRequest) (*kgrpb.NodesReply, error)
	Decorations(ctx context.Context, req *kxrpb.DecorationsRequest) (*kxrpb.DecorationsReply, error)
	CrossReferences(ctx context.Context, req *kxrpb.CrossReferencesRequest) (*kxrpb.CrossReferencesReply, error)
}
```

Adding more is just copy and paste tho.


siggisim (approved):

tylerwilliams:
Ugh, the tests caught me -- these protos don't have request / response context. So gotta figure out something there

siggisim (approved):

PR #7700, opened 2024-10-09
server: Add nonce to `AutoSizer`

fmeum:
`AutoSizer` creates style elements that need a nonce to pass our CSP.

Also add CSP v2 fallback sources for browsers that don't support CSP v3 (`strict-dynamic`).


server/http/csp/csp.go
```diff
@@ -8,7 +8,8 @@ import (
 )
 
 // Nonce is the context key type for the per-request CSP nonce.
-// The associated value is always of type template.HTMLAttr.
+// The associated value is always of type template.HTML and should be inserted
+// between the double quotes of a nonce = "..." attribute.
```
bduffany, line 12: nit: I'm not an expert in Go templates but it feels slightly odd to put HTML inside of quotes, since it's an HTML attribute value. Should this be `string`?

bduffany (approved):

PR #7702, opened 2024-10-09
Fix panic in debug logging

bduffany:
Fix a panic when res.Digest is nil

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -641,9 +641,13 @@ func (ff *BatchFileFetcher) batchDownloadFiles(ctx context.Context, req *repb.Ba
 	ff.statsMu.Unlock()
 
 	fileCache := ff.env.GetFileCache()
-	for _, res := range responses {
+	for i, res := range responses {
 		if res.Err != nil {
-			log.CtxInfof(ctx, "Failed to download %s: %s", digest.String(res.Digest), res.Err)
+			digestStr := "<?>"
+			if i >= 0 && i < len(req.Digests) && req.Digests[i] != nil {
+				digestStr = digest.String(req.Digests[i])
+			}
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -641,9 +641,13 @@ func (ff *BatchFileFetcher) batchDownloadFiles(ctx context.Context, req *repb.Ba
 	ff.statsMu.Unlock()
 
 	fileCache := ff.env.GetFileCache()
-	for _, res := range responses {
+	for i, res := range responses {
 		if res.Err != nil {
-			log.CtxInfof(ctx, "Failed to download %s: %s", digest.String(res.Digest), res.Err)
+			digestStr := "<?>"
+			if i >= 0 && i < len(req.Digests) && req.Digests[i] != nil {
+				digestStr = digest.String(req.Digests[i])
```
tylerwilliams, line 648: digest.String(res.Digest) is just for logging -- maybe you could make that method return "<nil digest>"?

bduffany: Done - went with `<nil>` since I think that'll make it more obvious in the logs that it's nil

server/remote_cache/digest/digest.go
```diff
@@ -547,8 +547,12 @@ func MissingDigestError(d *repb.Digest) error {
 	}
 }
 
-// String returns the digest formatted as "HASH/SIZE".
+// String returns the digest formatted as "HASH/SIZE" or the string "<nil>"
+// if the digest is nil.
```
tylerwilliams, line 551: nit: maybe add a comment about not using this and pointing to digest.Key for using digests as keys?

tylerwilliams (approved):

server/remote_cache/digest/digest.go
bduffany: Good call, done

PR #7704, opened 2024-10-09
docker: Install en_US.UTF-8 locale in rbe-ubuntu20-04

fmeum:
Java compilation actions unconditionally use this locale.

**Related issues**: N/A


bduffany:
> Are the docker image deployed automatically after this is merged or does that involve any manual action?

We have to build and push these images manually. I sent https://github.com/buildbuddy-io/buildbuddy/pull/7708 to consolidate the push targets - we'll want to push all of the ubuntu 20.04 targets in the new consolidated package, and then we'll need to update the constants in `platform.go`. I believe the `buildbuddy-toolchain` references the `ubuntu-20.04` image using the `:latest` tag so that part doesn't need updating, but we should hold off on pushing the `:latest` tag until we've done a bit of testing.

iain-macdonald (approved):

PR #7710, opened 2024-10-09
[raft] use seperate sessions for eviction and startShard

luluz66:
**Related issues**: N/A

### DIFF
--- a/enterprise/server/raft/store/store.go
+++ b/enterprise/server/raft/store/store.go
@@ -97,12 +97,16 @@ type Store struct {
 	grpcServer    *grpc.Server
 	apiClient     *client.APIClient
 	liveness      *nodeliveness.Liveness
-	session       *client.Session
-	// use a seperate session for operations like split, add, remove replica.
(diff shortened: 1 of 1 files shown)


enterprise/server/raft/store/store.go
```diff
@@ -202,22 +207,26 @@ func NewWithArgs(env environment.Env, rootDir string, nodeHost *dragonboat.NodeH
 	clock := env.GetClock()
 	session := client.NewSessionWithClock(clock)
 	txnSession := client.NewSessionWithClock(clock)
+	evictionSession := client.NewSessionWithClock(clock)
+	shardStarterSession := client.NewSessionWithClock(clock)
```
tylerwilliams, line 211: nit: add a comment about why there are separate sessions here?

tylerwilliams (approved):

enterprise/server/raft/store/store.go
luluz66: Done.

PR #7715, opened 2024-10-09
Allow admin API keys for a 'parent' group to affect child groups.

vadimberezniker:
Groups are considered related if they have the same SAML Metadata URL.

Only keys for groups marked as 'parent' are allowed access to related groups.

**Related issues**: N/A

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -236,6 +236,10 @@ func (g *apiKeyGroup) GetCapabilities() int32 {
 	return g.Capabilities
 }
 
+func (g *apiKeyGroup) HasCapability(cap akpb.ApiKey_Capability) bool {
+	return g.Capabilities&int32(cap) != 0
+}
+
 func (g *apiKeyGroup) GetUseGroupOwnedExecutors() bool {
 	return g.UseGroupOwnedExecutors
 }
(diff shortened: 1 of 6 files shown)


bduffany (approved):

enterprise/server/backends/authdb/authdb.go
```diff
@@ -339,6 +351,32 @@ func (d *AuthDB) fillDecryptedAPIKey(ak *tables.APIKey) error {
 	return nil
 }
 
+func (d *AuthDB) fillChildGroupIDs(ctx context.Context, akg *apiKeyGroup) error {
+	// If the group is not designated as a parent then don't bother doing a
+	// query.
+	if !akg.IsParent {
+		return nil
+	}
+	if akg.GetCapabilities()&int32(akpb.ApiKey_ORG_ADMIN_CAPABILITY) == 0 {
```
tylerwilliams, line 360: do we not have a "HasCapability" util method yet? We should add one to prevent errors in bit checking

tylerwilliams (approved):

enterprise/server/backends/authdb/authdb.go
vadimberezniker: This is a bit of a special case because we typically handle capabilities checks at a higher level where we do have a HasCapability function whereas we are still dealing with the raw capabilities here. I added one here as well but it might not get a lot of mileage.

PR #7717, opened 2024-10-10
server: Enforce CSP with `--app.strict_csp_enabled`

fmeum:
There haven't been any CSP reports in dev since the latest fix. The flag is only enabled in dev, but any violations reported after this point should be reported in a release-blocking fashion.

Also enables the flag in webdriver tests.


bduffany (approved):
LGTM (modulo Siggi's comment)

bduffany:
(As a side note, before turning this on in prod, we should turn on enforced CSP for local development, so that we can't accidentally introduce new content that works locally but then CSP failures go unnoticed in dev/prod)

fmeum:
Well, that's another way to discover all product features. I'll take a look and include a fix for this violation.

fmeum:
I talked to Tyler and confirmed that warnings in dev can't cause releases to be rolled back, but I will monitor this for a while before we enable it in prod.

PR #7720, opened 2024-10-10
Spanner integration

JonathanPerry651:
I'm hoping to add spanner integration to buildbuddy, which will give us vastly improved operational characteristics.

This is a checkpoint which does not pass tests, as I am beginning to run up against a bunch of code that appears not to use gorm to interface with the database.  Specifically the first one I came across was here - https://github.com/buildbuddy-io/buildbuddy/blob/master/server/tables/tables.go#L929, but there are definitely several others.

So I want to stop and ask advice - would you guys want me to start porting your code to fully use gorm here?  Or take some other approach?

Thanks!


tylerwilliams:
Hey @JonathanPerry651, thanks for sending a CL.

I think the best way for us to support Spanner in BuildBuddy is via the [PostgreSQL interface](https://cloud.google.com/spanner/docs/postgresql-interface). This might even work out of the box, but I haven't tested. Have you tried?

Adding native spanner support is difficult due to how spanner schemas work, and the fact that we occasionally run queries outside of the gorm object model. Importantly, we don't test with spanner or use it ourselves, so it's likely that support will fall behind.

For those reasons, I'm happy to accept CLs that add support for Spanner via the PostgreSQL interface, but I don't think we're ready to accept native spanner support at this time.

JonathanPerry651:
You are obviously correct, I'll work on that approach.  Thank you.

bduffany:
Closing due to inactivity

PR #7724, opened 2024-10-10
Allow an Admin API key for a 'parent' group to create new groups.

vadimberezniker:
The new groups will inherit the SAML IDP Metadata URL of the original group.

**Related issues**: N/A

### DIFF
--- a/server/buildbuddy_server/buildbuddy_server.go
+++ b/server/buildbuddy_server/buildbuddy_server.go
@@ -493,6 +493,10 @@ func (s *BuildBuddyServer) CreateGroup(ctx context.Context, req *grpb.CreateGrou
 	}
 
 	groupOwnedDomain := ""
+	// Groups created by a user via the UI can have "auto-join by domain"
+	// enabled for low-friction onboarding. Groups created using an API key
+	// are intended to be managed manually so we do not currently allow
+	// joining by domain.
 	if req.GetAutoPopulateFromOwnedDomain() && u.GetUserID() != "" {
 		user, err := userDB.GetUser(ctx)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -511,6 +515,17 @@ func (s *BuildBuddyServer) CreateGroup(ctx context.Context, req *grpb.CreateGrou
 		DeveloperOrgCreationEnabled: req.GetDeveloperOrgCreationEnabled(),
 		UseGroupOwnedExecutors:      req.GetUseGroupOwnedExecutors(),
 	}
+
+	if u.HasCapability(akpb.ApiKey_ORG_ADMIN_CAPABILITY) && u.GetUserID() == "" {
```
tylerwilliams, line 519: could you add a comment about what this is doing & why?

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -495,13 +495,17 @@ func (s *BuildBuddyServer) CreateGroup(ctx context.Context, req *grpb.CreateGrou
 	}
 
 	groupOwnedDomain := ""
-	if req.GetAutoPopulateFromOwnedDomain() {
+	if req.GetAutoPopulateFromOwnedDomain() && u.GetUserID() != "" {
```
tylerwilliams, line 498: nit: a comment here (about why we check userID != "") would be nice too

tylerwilliams (approved):
lgtm w/ 2 minor nits

server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -509,6 +517,20 @@ func (s *BuildBuddyServer) CreateGroup(ctx context.Context, req *grpb.CreateGrou
 		DeveloperOrgCreationEnabled: req.GetDeveloperOrgCreationEnabled(),
 		UseGroupOwnedExecutors:      req.GetUseGroupOwnedExecutors(),
 	}
+
+	// For groups created using an API Key allow the SAML IDP Metadata URL
+	// to be inherited if the API Key group is marked as a 'parent' group.
+	// This allows the new group to be managed using a parent group API key.
+	if u.HasCapability(akpb.ApiKey_ORG_ADMIN_CAPABILITY) && u.GetUserID() == "" {
```
bduffany, line 524: nit: the `UserID == ""` check doesn't seem necessary here? (it might make sense to eventually support ORG_ADMIN for personal API keys)

bduffany (approved):

server/buildbuddy_server/buildbuddy_server.go
vadimberezniker: This API is also by the web frontend where u.GetUserID() != "" and I don't want this behavior to occur if someone creates a group via the UI. Once the customer is done testing, the plan is to promote the relevant APIs to the public API where we wouldn't need to do this check anymore.

PR #7730, opened 2024-10-11
explain: Add support for the new exec log format

fmeum:
In Bazel 7.4.0 and Bazel 8, the compact execlog format has changed and now represents runfiles directories via a dedicated `RunfilesTree` method, which is both more space efficient and also allows attribution of changes to runfiles.

This change adds support for the new format and, to get the new tests to pass, also fixes a number of bugs:
* `TransitivelyInvalidated` no longer counts diamond reverse deps and spawns with multiple outputs twice.
* Source directories and tree artifacts are now diffed consistently, with the latter still being treated as an explanation for args changes.
* Track both a logical and a resolved path for inputs content diffs to make attribution of runfiles symlinks possible.
* Topological sorting did not visit all edges.
* Tool contents are no longer diffed separately as they are a subset of inputs.
* Bazel settings that make execlogs incomparable now result in an error.


fmeum:
Apologies for the large diff, but implementing support for this feature without fixing the mentioned bugs was just too confusing for me.

fmeum:
Bzlmod tests fail with
```
ERROR: While parsing option --startup_time=18446744073709550929: '18446744073709550929' is not a long
```
even on retries. Maybe something is off with the clock on that executor?

bduffany:
@fmeum we've seen that error pop up from time to time but haven't yet figured out why it happens. It usually goes away with a re-run.

One thing that's suspicious is that `18446744073709550929` starts with a lot of the same digits as uint64_max; since `startup_time` is a duration, it kind of seems like bazel is subtracting two timestamps that it's expecting to be monotonically increasing, but aren't. I vaguely suspect this could be related to VM snapshotting

cli/explain/compactgraph/compactgraph_test.go
```diff
@@ -16,141 +16,255 @@ import (
...
-	assert.Equal(t, "TestRunner", sd3.Mnemonic)
-	assert.Empty(t, sd3.GetModified().GetTransitivelyInvalidated())
-	assert.True(t, sd3.GetModified().GetExpected())
-	require.Len(t, sd3.GetModified().GetDiffs(), 1)
-	sd3d1 := sd3.GetModified().Diffs[0]
-	require.IsType(t, &spawn_diff.Diff_OutputContents{}, sd3d1.Diff)
-	require.Len(t, sd3d1.GetOutputContents().GetFileDiffs(), 1)
-	sd3fd1 := sd3d1.GetOutputContents().GetFileDiffs()[0]
-	assert.Regexp(t, "^bazel-out/[^/]+/testlogs/src/test/java/com/example/lib/lib_test/test.xml$", sd3fd1.GetOldFile().GetPath())
+	t.Run("7.3.1", func(t *testing.T) {
```
sluongng, line 19: Nit: If the tests are different between versions, then it would be nicer to split them up into separate Test funcs. That way rules_go sharding logic could be applied neatly in the future.

If the tests are identical between versions, we can use a test table with nested t.Run() calls inside to mark with test is which.
```go
for _, bazelVersion := []string{"7.3.1", "8.0.0"} {
  t.Run(bazelVersion, func (t *testing.T) {
    for _, tc := testCases {
      t.Run(tc.Name, func (t *testing.T) { ... })
    }
  })
}
```

this way we can filter down the subtest with `--test_filter=`

cli/explain/compactgraph/input.go
```diff
@@ -269,17 +277,239 @@ func protoToInputSet(s *spawn.ExecLogEntry_InputSet, previousInputs map[uint32]I
...
+	transitiveSets []*SymlinkEntrySet
+
+	shallowPathHash    Hash
+	shallowContentHash Hash
+}
+
+var emptySymlinkEntrySet = protoToSymlinkEntrySet(&spawn.ExecLogEntry_SymlinkEntrySet{}, nil)
+
+func (s *SymlinkEntrySet) Path() string {
+	panic(fmt.Sprintf("SymlinkEntrySet %s doesn't have a path", s.String()))
```
sluongng, line 308: Let's use log.Fatalf instead.

cli/explain/compactgraph/input.go
```diff
@@ -269,17 +277,239 @@ func protoToInputSet(s *spawn.ExecLogEntry_InputSet, previousInputs map[uint32]I
...
+}
+func (s *SymlinkEntrySet) ShallowPathHash() Hash    { return s.shallowPathHash }
+func (s *SymlinkEntrySet) ShallowContentHash() Hash { return s.shallowContentHash }
+func (s *SymlinkEntrySet) Proto() any {
+	panic(fmt.Sprintf("SymlinkEntrySet %s doesn't support Proto()", s.String()))
```
sluongng, line 313: ditto: use log.Fatalf instead.

cli/explain/compactgraph/input.go
```diff
@@ -269,17 +277,239 @@ func protoToInputSet(s *spawn.ExecLogEntry_InputSet, previousInputs map[uint32]I
...
+	// 4. Root symlinks.
+	for runfilesPath, artifact := range iterateAsRunfiles(r.RootSymlinks, noFilter) {
+		m[runfilesPath] = artifact
+	}
+	// 5. The repo mapping manifest at its fixed location.
+	if r.RepoMappingManifest != nil {
+		m["_repo_mapping"] = r.RepoMappingManifest
+	}
+	// 6. The existence of the <workspace runfiles directory>/.runfile file, similar to empty files, is a pure function
+	// of the other paths.
```
sluongng, line 419: Should there be some code for (6) here?

proto/spawn_diff.proto
```diff
@@ -12,7 +12,7 @@ message SpawnDiff {
   oneof diff {
     OldOnly old_only = 5;
     NewOnly new_only = 6;
-    Modified modified = 7;
+    Common common = 7;
```
sluongng, line 15: Nit: Unclear why this rename is needed.

proto/spawn_diff.proto
```diff
@@ -33,7 +33,6 @@ message Modified {
 message Diff {
   oneof diff {
     StringSetDiff tool_paths = 1;
-    FileSetDiff tool_contents = 2;
```
sluongng, line 36: Do we need to add `reserved 2` to the message?

proto/spawn_diff.proto
```diff
@@ -66,17 +65,19 @@ message FileSetDiff {
...
   }
   oneof new {
-    tools.protos.ExecLogEntry.File new_file = 2;
-    tools.protos.ExecLogEntry.Directory new_directory = 4;
-    tools.protos.ExecLogEntry.UnresolvedSymlink new_symlink = 6;
-    string new_invalid_output = 8;
+    tools.protos.ExecLogEntry.File new_file = 3;
+    tools.protos.ExecLogEntry.Directory new_directory = 5;
+    tools.protos.ExecLogEntry.UnresolvedSymlink new_symlink = 7;
+    string new_invalid_output = 9;
```
sluongng, line 80: Is this a safe change? Not sure if we should add it as tag 9 instead of tag 1

sluongng (approved):
I have a bunch of nits.
The logic inside compactgraph.go is growing too big for me to consume all in one reading. And the diff is big as well...
I feel like we gonna need to break the graph structure down to smaller self-contained packages (I.e `/compactgraph/runfiles`) so that we can test each separately.

I also dislike the sprinkle use of `panic` and ignoring errors returned by `binary.Write`. I think they are harmless right now, but could make it harder to troubleshoot issues in the future.

The rest looks sane to me.

cli/explain/compactgraph/BUILD.bazel
```diff
@@ -19,12 +19,16 @@ go_library(
...
-    srcs = ["compactgraph_test.go"],
+    srcs = [
+        "compactgraph_test.go",
+        "input_test.go",
+    ],
     data = [
         "//cli/explain/compactgraph/testdata:all_logs",
     ],
+    embed = [":compactgraph"],
+    race = "on",
```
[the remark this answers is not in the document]
fmeum: I didn't know we run with `race` in CI. I don't quite like that this introduces a difference between local and CI test execution, but I agree that enabling `race` just on this test doesn't make sense.

bduffany (approved):

proto/spawn_diff.proto
fmeum: I found that `Modified` is not quite the right term as users don't directly modify a spawn, so I wanted to opt for something that fits better with `OldOnly` and `NewOnly`. 

It doesn't have to be part of this PR though, I changed the name back to make the diff easier to read.

proto/spawn_diff.proto
fmeum: We are the only consumers of this proto so far, the `explain` command doesn't provide a user-facing way to produce it. We can think about adding a proto output flag at some point, but so far I would like to keep it experimental.

proto/spawn_diff.proto
fmeum: See above, this proto is internal only so far.

cli/explain/compactgraph/compactgraph_test.go
fmeum: I split out the tests that are different depending on the version. The once that aren't already use `t.Run`.

fmeum:
> ignoring errors returned by `binary.Write`

We are writing to a `hash.Hash`, which can never fail.

PR #7731, opened 2024-10-11
Add CPU timeline to UsageStats proto

bduffany:
Understanding the distribution of CPU usage (as opposed to just the average) may help us improve our task sizing model, and is also useful information to present to users in order to understand how effectively their actions are utilizing available resources.

### DIFF
--- a/proto/remote_execution.proto
+++ b/proto/remote_execution.proto
@@ -2199,6 +2199,10 @@ message UsageStats {
   // Total number of CPU-nanoseconds consumed by the task.
   int64 cpu_nanos = 2;
 
+  // CPU usage timeline. Each sample represents the approximate CPU usage
+  // accumulated over each period.
+  Timeseries cpu_timeline = 8;
+
   // Most recently recorded total memory usage of the task. This field is only
   // used for real-time metrics and shouldn't be used as a "summary" metric for
   // the task (peak_memory_bytes is a more useful summary metric).
@@ -2239,6 +2243,23 @@ message UsageStats {
 
   // IO PSI metrics.
   PSI io_pressure = 7;
+
+  message Timeseries {
+    // Time when recording started.
+    google.protobuf.Timestamp start = 1;
+
+    // Time when the last sample was taken.
+    google.protobuf.Timestamp end = 2;
+
+    // Duration between each sample.
+    google.protobuf.Duration period = 3;
+    
+    // TODO: compress the samples to save space
+
+    // Observed samples. The interpretation of these samples may depend on the
+    // metric being reported.
+    repeated int64 samples = 4;
+  }
 }
 
 // Pressure Stall Information, commonly known as PSI.


proto/remote_execution.proto
```diff
@@ -2199,6 +2199,27 @@ message UsageStats {
...
+  // executed a task that ran for 200ms, and during the first 100ms we observed
+  // 3140 total cpu-millis used, and during the second 200ms we observed 111
+  // total cpu-millis used. Then a valid histogram representing this execution
+  // could be:
+  //
+  // { buckets: [100, 3100], counts: [0.1, 0.1], bucket_width: 100 }
+  //
+  // Note: this histogram does not convey any information about *when* the CPU
+  // usage occurred; it only reports information on how frequently we observed
+  // each range of CPU usages.
```
bduffany, line 2220: Another option that preserves timing information would be to report CPU usage at every 100ms interval and then if a task runs for a very long time we can "collapse" intervals so that we only report every second or every minute, to keep the proto reasonably small. I figured this approach would be slightly more complex but it may be more useful / interesting to report it this way? curious if anyone has thoughts here

tylerwilliams: delta compression would probably work well here if your CPU stats are quantized and the period is fixed

tylerwilliams: You can also use run length encoding for the intervals if you want.

Anyway before commiting to a format, you probably want to run a long-ish task, gather the data and encode it and see if your format makes sense.

vanja-p: For users, some timeline information is definitely useful (compared to a histogram), but I suspect that when they really need this data (like a large test), they'll want to know which process or function used CPU when. That means running some sort of profiler. So I think less precise data here is more than enough to give them an overview. 100ms windows are probably too small. 1s seems good. Maybe 10s is all you need.

I like the idea of delta compression, though it does make the data a bit harder to use. If the numbers can be reasonably small, each sample could be 2-3 bytes. Though, you could get also small numbers if samples were rounded to the nearest deci-period. For example, let's say the period is 1s. Instead of a sample value of 3517ms indicating that the action used 3.517 cores during that period, use a sample of 35 deciseconds to indicate that the action used ~3.5 cores during that period.

I'm just guessing, but RLE is probably not worth it. Analyzing the data in something like a SQL query becomes way harder then. Also, I think it requires a repeated message instead of a primitive field, which means you don't benefit from packed fields.

proto/remote_execution.proto
```diff
@@ -2239,6 +2242,15 @@ message UsageStats {
...
   // IO PSI metrics.
   PSI io_pressure = 7;
+
+  message Timeseries {
+    // Duration between each sample.
+    google.protobuf.Duration period = 1;
+    
+    // Observed samples.
+    // TODO: maybe do delta-compression and/or RLE to reduce data size.
+    repeated int32 samples = 2;
```
vanja-p, line 2252: this could be a uint32, unless there's some metric that could be negative.

proto/remote_execution.proto
bduffany: For some more background: the main motivation for this PR is to start collecting more data in order to improve our automatic task sizing. This is sort of a pre-requisite for being able to apply hard-limits on task resource usage, since we don't want to apply automatic hard limits if we aren't confident about what the limits should be. There is more discussion in this issue: https://github.com/buildbuddy-io/buildbuddy-internal/issues/3965

We currently estimate task sizes by using average CPU (total cpu-milliseconds used divided by execution duration), which turns out to be a poor estimate for certain types of workloads - e.g. some VM workloads spend a lot of work up-front decompressing docker images using multiple cores, and then once the image is decompressed, the test is single-threaded and/or IO bound, resulting in a lot less CPU being used throughout the test. So using the average CPU to auto-size these workloads doesn't work very well.

One idea we discussed was recording the CPU usage over time and then somehow improving our task sizing model using that information. Our current idea is to use p90 CPU usage (this is just a starting point which will likely have some drawbacks; maybe some more analysis could be done to understand how to more effectively use this info).

Encoding aside, I'm curious if you have any thoughts about this overall approach?

But in terms of the implementation here I do agree that we could probably just collect samples in 1s intervals and report every single sample as a plain old int32. That's just 14 KB per timeseries per hour of execution time.

vanja-p: I meant to talk with you about this at Bazelcon, but then I forgot. I'll set up a VC for tomorrow. I have a bunch of questions before I can have any well informed thoughts on this.

proto/remote_execution.proto
bduffany: I think we are following some best practice about using signed ints because it makes it less likely to run into bugs due to overflow when doing math on the values.

proto/remote_execution.proto
bduffany: We chatted a bit offline and brainstormed various ideas about how to improve the CPU task sizing situation, but still think that measuring this CPU timeline is useful.

proto/remote_execution.proto
```diff
@@ -2239,6 +2243,24 @@ message UsageStats {
...
+    google.protobuf.Timestamp end = 2;
+
+    // Duration between each sample.
+    google.protobuf.Duration period = 3;
+    
+    // Observed samples. The interpretation of these samples may depend on the
+    // metric being reported.
+    //
+    // TODO: maybe do quantization + delta-compression and/or RLE to reduce data
+    // size.
```
tylerwilliams, line 2261: I think if we do this we should just add a bytes field and link to a util function where we use roaring or something.

tylerwilliams (approved):

vanja-p (approved):

proto/remote_execution.proto
bduffany: OK - will keep this in mind

PR #7734, opened 2024-10-11
Clean up unused explicit command line overwriting

maggie-lou:
**Related issues**: N/A

### DIFF
--- a/app/invocation/invocation_model.tsx
+++ b/app/invocation/invocation_model.tsx
@@ -853,19 +853,6 @@ export default class InvocationModel {
   }
 
   explicitCommandLine() {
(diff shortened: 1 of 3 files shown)


bduffany (changes requested):
hm I don't think this is unused - if I run `bb test //... --bes_backend=grpc://localhost:1985 --bes_results_url=http://localhost:8080/invocation/`

before this PR I see

image


after this PR I see

image

(notice the sidecar stuff showing up in the explicit command line)

maggie-lou:
@bduffany You're right - I misunderstood the code and had a typo when I tried to use EXPLICIT_COMMAND_LINE which caused it to not work, and incorrectly assumed that meant it was getting redacted. Will modify my other PR to use it

PR #7736, opened 2024-10-14
Use go proto target for cel.dev/expr

jschaf:
Looks like a copy-paste bug since the target is the same for the `proto` rule and the `proto go` rule for cel.dev/expr.

Discovered while trying to get the Spanner Go client library to compile.

This code, introduced in https://github.com/buildbuddy-io/buildbuddy/pull/7465, was fantastically helpful in figuring that out. 

**Related issues**: N/A


jschaf:
@tempoz is likely the best reviewer.

tylerwilliams (approved):

PR #7744, opened 2024-10-17
Fix oci.RuntimePlatform to set variant for arm64

bduffany:


iain-macdonald (approved):

tempoz:
I think the problem may actually be that `match.Platform`  is not as sophisticated as `index.matchesPlatform`, which we can access through a `descriptor.Descriptor`. The arm64 images `docker buildx` creates don't specify a variant, so this PR would still result in them being rejected.

PR #7745, opened 2024-10-17
Prevent delete_build_root_on_startup from wiping image cache

bduffany:
Running the executor with `--executor.delete_build_root_on_startup` deletes the contents of `--executor.build_root_directory`. This is where temporary files are stored (action workspaces, firecracker chroot dirs, etc.).

Unfortunately, this is also where cached firecracker / OCI images are stored (under an `executor` subdir).

This PR adds a temporary fix for this issue which is to only delete the `executor` directory if a separate flag `executor.delete_image_cache_on_startup` is set.

Eventually, we should move persistent files and temporary files to separate directories. Ideally, cached artifacts would be somehow stored in filecache - this is a little tricky though because OCI image layers in particular are represented as extracted directories, not single files.

### DIFF
--- a/enterprise/server/cmd/executor/executor.go
+++ b/enterprise/server/cmd/executor/executor.go
@@ -304,9 +304,12 @@ func cleanBuildRoot() {
 		return
 	}
 
+	log.Infof("Cleaning build root directory")
+
+	n := 0
 	for _, entry := range entries {
-		// TODO: move cached firecracker/ociruntime images to a different
-		// directory.
+		// TODO(http://go/b/3994): move cached firecracker/ociruntime images to
+		// a different directory.
 		if entry.Name() == "executor" && !*deleteImageCacheOnStartup {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


enterprise/server/cmd/executor/executor.go
```diff
@@ -297,3 +292,29 @@ func main() {
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return
+		}
+		log.Warningf("Failed to read build root dir: %s", err)
+		return
+	}
+
+	for _, entry := range entries {
+		// TODO: move cached firecracker/ociruntime images to a different
```
iain-macdonald, line 308: mind filing a bug and referencing it here?

enterprise/server/cmd/executor/executor.go
```diff
@@ -297,3 +292,29 @@ func main() {
...
+		// directory.
+		if entry.Name() == "executor" && !*deleteImageCacheOnStartup {
```
iain-macdonald, line 310: This name is flag-defined, right? Can you make sure it's correct even if the value of the flag is overridden?

enterprise/server/cmd/executor/executor.go
bduffany: Done

enterprise/server/cmd/executor/executor.go
bduffany: It's not flag-defined - it's a hard-coded string that we sprinkled in a few different places. I'll leave the code as-is, since we want to move image cache artifacts to filecache eventually anyway

iain-macdonald (approved):

PR #7746, opened 2024-10-17
Use `go-containerregistry`'s image resolution instead of rolling our own

tempoz:
`go-containerregistry` will handle image resolution by platform so long as the `WithPlatform` option is provided.

**Related issues**: N/A


bduffany (approved):
LGTM, but CC @iain-macdonald just in case the previous code was written this way intentionally

Mind also cherry-picking this test commit? https://github.com/buildbuddy-io/buildbuddy/commit/02ddf361840bce09661332f8a23345e32774daca

deps.bzl
```diff
@@ -2205,8 +2205,8 @@ def install_go_mod_dependencies(workspace_name = "buildbuddy"):
     go_repository(
         name = "com_github_google_go_containerregistry",
         importpath = "github.com/google/go-containerregistry",
-        sum = "h1:MMkSh+tjSdnmJZO7ljvEqV1DjfekB6VUEAZgy3a+TQE=",
-        version = "v0.15.2",
+        sum = "h1:B1wPJ1SN/S7pB+ZAimcciVD+r+yV/l/DSArMxlbwseo=",
```
bduffany, line 2208: nit: are the version bumps needed here?

enterprise/server/util/oci/oci.go
```diff
@@ -144,39 +151,21 @@ func Resolve(ctx context.Context, imageName string, platform *rgpb.Platform, cre
...
-		return imgs[0], nil
-	case types.OCIManifestSchema1, types.DockerManifestSchema2:
-		img, err := remoteDesc.Image()
-		if err != nil {
+	img, err := remoteDesc.Image()
+	if err != nil {
+		switch remoteDesc.MediaType {
+		// This is an "image index", a meta-manifest that contains a list of
+		// {platform props, manifest hash} properties to allow client to decide
+		// which manifest they want to use based on platform.
```
iain-macdonald, line 159: Can you update this comment to explain that these should be resolved correctly by `Image()` so that's why this is an error?

iain-macdonald (approved):
Not exactly sure why I wrote it the way that it was. If this approach works in your testing, it LGTM

deps.bzl
tempoz: No, just something I tried hoping it would resolve the problem. I'll revert it.

enterprise/server/util/oci/oci.go
```diff
@@ -144,39 +151,21 @@ func Resolve(ctx context.Context, imageName string, platform *rgpb.Platform, cre
...
-			return nil, status.NotFoundErrorf("could not find suitable image in image index")
-		}
-		if len(imgs) > 1 {
-			return nil, status.NotFoundErrorf("found multiple matching images in image index")
-		}
-		return imgs[0], nil
-	case types.OCIManifestSchema1, types.DockerManifestSchema2:
-		img, err := remoteDesc.Image()
-		if err != nil {
+	img, err := remoteDesc.Image()
```
tempoz, line 154:
```suggestion
    // Image() should resolve both images and image indices to an appropriate image
	img, err := remoteDesc.Image()
```

tempoz:
> Mind also cherry-picking this test commit? `02ddf36`

Done!

bduffany (approved):

PR #7751, opened 2024-10-18
Don't hard-code x86 platform in `remote-shared` config

tempoz:
**Related issues**: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -36,8 +36,6 @@ common:remote-shared --remote_timeout=600
 common:remote-shared --remote_download_minimal
 common:remote-shared --jobs=100
 common:remote-shared --verbose_failures
-common:remote-shared --platforms=//platforms:linux_x86_64
-common:remote-shared --extra_execution_platforms=//platforms:linux_x86_64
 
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -36,8 +36,6 @@ common:remote-shared --remote_timeout=600
 common:remote-shared --remote_download_minimal
 common:remote-shared --jobs=100
 common:remote-shared --verbose_failures
-common:remote-shared --platforms=//platforms:linux_x86_64
-common:remote-shared --extra_execution_platforms=//platforms:linux_x86_64
```
sluongng, line 40: This will probably break using `--config=remote` on MacOS today.

Instead, I think we can change this to

```
common:target-x86 --platforms=//platforms:linux_x86_64
common:target-x86 --extra_execution_platforms=//platforms:linux_x86_64

...

common:remote --config=target-x86
common:remote-arm64 --config=target-arm64 # when arm64 is implemented
```

wdyt?

tempoz:
This is being handled in a different PR.

PR #7753, opened 2024-10-18
Revert "gosdk: upgrade to 1.23.2"

iain-macdonald:
Reverts buildbuddy-io/buildbuddy#7718 due to a newly introduced panic: https://go-review.googlesource.com/c/go/+/611496


sluongng:
It might be sufficient to just revert the internal repo bc our binaries are built from there

sluongng (approved):

PR #7757, opened 2024-10-21
filecache: restrict initial scan to group-specific dirs

bduffany:
The filecache has a `_tmp` dir which we should not scan, and we are also considering moving OCI + firecracker images to this cache directory, to prevent `executor.delete_build_root_on_startup` from wiping cached images.


iain-macdonald (approved):

vadimberezniker (approved):
What's the context? Why are there other files under the filecache dir?

bduffany:
> What's the context?

Updated the PR description

PR #7761, opened 2024-10-21
flakes: fall back to test.log if test.xml is not found

jdhollen:


iain-macdonald (approved):

iain-macdonald (approved):
need to cherrypick?

iain-macdonald:
> need to cherrypick?

oops, nvm

PR #7766, opened 2024-10-21
Improve shell token handling in remotebazel

bduffany:
* `strings.Join(bazelArgs, " ")` is incorrect if any of the bazel args contains a space or a shell special character - fix by adding a `shlex.Quote()` function (added a new `shlex` package since there are a few other places where I think we could use this)
* One user is depending on the current incorrect behavior to be able to pass arbitrary commands. To support this use case, add a new `--script` arg for passing arbitrary shell scripts

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -35,6 +35,7 @@ import (
 	"github.com/buildbuddy-io/buildbuddy/server/util/bazel"
 	"github.com/buildbuddy-io/buildbuddy/server/util/grpc_client"
 	"github.com/buildbuddy-io/buildbuddy/server/util/rexec"
+	"github.com/buildbuddy-io/buildbuddy/server/util/shlex"
 	"github.com/buildbuddy-io/buildbuddy/server/util/status"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
@@ -77,6 +78,7 @@ var (
 	useSystemGitCredentials = RemoteFlagset.Bool("use_system_git_credentials", false, "Whether to use github auth pre-configured on the remote runner. If false, require https and an access token for git access.")
 	runFromBranch           = RemoteFlagset.String("run_from_branch", "", "A GitHub branch to base the remote run off. If unset, the remote workspace will mirror your local workspace.")
 	runFromCommit           = RemoteFlagset.String("run_from_commit", "", "A GitHub commit SHA to base the remote run off. If unset, the remote workspace will mirror your local workspace.")
+	script                  = RemoteFlagset.String("script", "", "Shell code to run instead of a Bazel command.")
 
 	defaultBranchRefs = []string{"refs/heads/main", "refs/heads/master"}
 )
@@ -793,13 +795,13 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 	req.GetRepoState().Patch = append(req.GetRepoState().Patch, repoConfig.Patches...)
 
 	// TODO(Maggie): Clean up after we've migrated fully to use `Steps`
-	stepsMode := os.Getenv("STEPS_MODE") == "1"
+	stepsMode := os.Getenv("STEPS_MODE") == "1" || *script != ""
 	if stepsMode {
-		req.Steps = []*rnpb.Step{
-			{
-				Run: fmt.Sprintf("bazel %s", strings.Join(bazelArgs, " ")),
-			},
+		run := *script
+		if run == "" {
+			run = fmt.Sprintf("bazel %s", shlex.Quote(bazelArgs...))
 		}
+		req.Steps = []*rnpb.Step{{Run: run}}
 	} else {
 		req.BazelCommand = strings.Join(bazelArgs, " ")
 	}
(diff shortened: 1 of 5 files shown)


goniz:
I am the "one user" that you referenced 🫢😅
In my use-case I need to run `bazel run ..` and then `bazel test ..` in the same microVM. 
If the CLI would allow me to specify multiple steps this would be great as I don't really need a shell script.

Example:
```bash
bazel remote run \
  --step "bazel run //workflow_setup" \
  --step "bazel test //target"
```

cli/remotebazel/remotebazel.go
```diff
@@ -77,6 +78,7 @@ var (
 	useSystemGitCredentials = RemoteFlagset.Bool("use_system_git_credentials", false, "Whether to use github auth pre-configured on the remote runner. If false, require https and an access token for git access.")
 	runFromBranch           = RemoteFlagset.String("run_from_branch", "", "A GitHub branch to base the remote run off. If unset, the remote workspace will mirror your local workspace.")
 	runFromCommit           = RemoteFlagset.String("run_from_commit", "", "A GitHub commit SHA to base the remote run off. If unset, the remote workspace will mirror your local workspace.")
+	script                  = RemoteFlagset.String("script", "", "Shell code to run instead of a Bazel command.")
```
maggie-lou, line 81: Would this work with a multi-line script? That seems to be a popular use case

Ex.
`
bazel build //...
bazel test //... 
`

If so, could you add an example to the docs?

maggie-lou: Also @siggisim because I know you have some thoughts on the interface for supporting bash commands in the remote bazel CLI

bduffany:
@goniz you're totally fine haha

After this PR, you'd specify that command like

```
bb remote run --script="bazel run //workflow_setup && bazel test //target"
```

or multi-line:

```
bb remote run --script='
  bazel run //workflow_setup
  bazel test //target
'
```

we could also support multiple `--step` arguments if that's easier?

cli/remotebazel/remotebazel.go
```diff
@@ -1037,14 +1039,20 @@ func parseArgs(commandLineArgs []string) (bazelArgs []string, execArgs []string,
...
-	bazelArgs = append(bazelArgs, "--config=buildbuddy_remote_cache")
+	if len(bazelArgs) > 0 {
+		if *script != "" {
+			return nil, nil, fmt.Errorf("cannot pass both a bazel command and --script")
+		}
+
+		// Ensure all bazel remote runs use the remote cache.
+		// The goal is to keep remote workloads close to our servers, so use the same
+		// app backend as the remote runner.
+		bazelArgs = arg.Remove(bazelArgs, "bes_backend")
```
maggie-lou, line 1050: Maybe we should add these in the [bazel wrapper in ci_runner:main ](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/cmd/ci_runner/main.go#L2683) instead of here, to guarantee that these flags are set on every bazel command. With the current implementation, these will no longer be applied if a user uses --script

server/util/shlex/shlex.go
```diff
@@ -0,0 +1,64 @@
...
+//
+// Example:
+//
+//	shlex.Split("  foo --bar='/Quoted/Path/With Spaces'  ")
+//	// Returns: []string{"foo", "--bar=/Quoted/Path/With Spaces"}
+func Split(command string) ([]string, error) {
+	return gshlex.Split(command)
+}
+
+// Quote escapes the given shell tokens so that each argument would be treated
```
maggie-lou, line 28: Talked offline about this, but could you add more comments to the code and to the PR description giving context for why this is needed

maggie-lou:
Broadly I'm concerned that adding --script will break some assumptions we're making that the bazel command will be set in `opts.Args`

For example, we have some [logic here](https://github.com/buildbuddy-io/buildbuddy/blob/master/cli/remotebazel/remotebazel.go#L940) around building a target remotely, and then fetching it and running it locally

I think we should make the code more clear that either you can pass a bazel command or a bash script, but not both. (This is asserted in one function, but the rest of the code makes it seem like we can expect bazelArgs / opts.Args to contain the command to run) And we should have some more documentation on what features you can expect when using either option

goniz:
Just to clarify our actual use-case, we actually wanted a way to run a workflow action with concurrency limit like with GitHub Actions's concurrency groups.
Since it's not supported (yet?) we went with a GitHub action that triggers the build using Bazel Remote with a concurrency group set on the action.

If we had the option to do `bb remote run --workflow "action-name"` and it would tail the log until completion etc. it would be more ergonomic / simple for us then `--script`.

@maggie-lou @bduffany

maggie-lou:
Hey @goniz - would it work to poll the status of the workflow in your GitHub action? 

To trigger the workflow, you can use the [`ExecuteWorkflow` API](https://www.buildbuddy.io/docs/enterprise-api/#executeworkflow), which will return an invocation ID. 

If you want to tail the logs in the GitHub action, you can use our [`GetLog` API ](https://www.buildbuddy.io/docs/enterprise-api/#getinvocation) until `GetLogResponse.NextPageToken` is empty. This is how remote bazel streams the logs.

If you just need to check the status of the invocation, you can use the [GetInvocation API](https://www.buildbuddy.io/docs/enterprise-api/#getinvocation). The `bazelExitCode` and `success` fields will not be set until the invocation is complete.

PR #7768, opened 2024-10-21
Handle duplicate aspect-related build events in the event index

maggie-lou:
I noticed we were seeing duplicate build events for a given label, and realized that one of them had an `Aspect` ID (For example, you can see in this [invocation](https://buildbuddy.buildbuddy.io/invocation/1bd24aa9-0d79-47bc-8366-4347cae84ece#artifacts) we render the build artifact twice). I believe the aspect, named "ValidateTarget" is run because we set `--experimental_use_validation_aspect=true`


bduffany (approved):
nice find!

PR #7769, opened 2024-10-21
Use AggregatedList instead of List to find candidate instance templates in edit_instance_template.

iain-macdonald:
AggregatedList returns regional and global instance templates. List does not.


tools/edit_instance_template/edit_instance_template.go
```diff
@@ -51,26 +51,40 @@ func parseKubeLabels(tpl *computepb.InstanceTemplate) map[string]string {
...
+				continue
+			}
+			log.Debugf("Found editable instance template %s", tpl.GetName())
+			templates = append(templates, tpl)
 		}
-		templates = append(templates, tpl)
 	}
+
+	if numTemplates == 0 {
+		log.Fatalf("No instance templates matching '%s' returned by "+
```
vadimberezniker, line 84: nit: it's usually bad form to log.Fatalf from outside of main. return an error here, the caller is already calling log.Fatalf on errors.

vadimberezniker (approved):

tools/edit_instance_template/edit_instance_template.go
iain-macdonald: done

PR #7770, opened 2024-10-21
[RB] Fix fetching remote build outputs

maggie-lou:
Remote bazel can be used to build a target on a remote runner, and then the CLI will fetch the build outputs to be used locally.

A couple things were broken:
* The CLI uses certain build events to identify and fetch build outputs, but [this PR](https://github.com/buildbuddy-io/buildbuddy/pull/4490) moved target-level events from the `Event` field to `TargetGroups`. This PR updates the CLI to use the new field
* Verifies that the build uses `--remote_upload_local_results` so that remote build outputs can be fetched from the cache

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -42,6 +42,7 @@ import (
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc/metadata"
 
+	cmnpb "github.com/buildbuddy-io/buildbuddy/proto/api/v1/common"
 	bespb "github.com/buildbuddy-io/buildbuddy/proto/build_event_stream"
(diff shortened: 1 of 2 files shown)


cli/remotebazel/remotebazel.go
```diff
@@ -1120,3 +1123,13 @@ func contains(m map[string]string, elem string) bool {
 	_, ok := m[elem]
 	return ok
 }
+
+// Returns the bazel command - i.e. `build` or `run`
+func getBazelCommand(bazelArgs []string) string {
```
bduffany, line 1128: nit: use `parser.GetBazelCommandAndIndex` here

bduffany (approved):

PR #7771, opened 2024-10-21
refactor: get rid of unneeded prefixing logic for queries

jdhollen:
this cleans up some freaky string concats in query code that i've never really liked much--these callsites were all being used correctly, but a messed up call could theoretically lead to injection.

doesn't seem to me that we actually need `i.` on the search service queries, and that was the only place this was being used.


jdhollen:
+ @tylerwilliams also--the only sensitive thing in here to me is the perms check--i guess i can imagine a world where we don't like that we don't explicitly name the invocations table (`role` instead of `i.role`), but that feels less error-prone than these prefixes do, to be honest.

bduffany (approved):

tylerwilliams (approved):

PR #7776, opened 2024-10-22
Record CPU usage timelines

bduffany:
Populates the CPU timeseries field added in https://github.com/buildbuddy-io/buildbuddy/pull/7731

### DIFF
--- a/enterprise/server/remote_execution/container/BUILD
+++ b/enterprise/server/remote_execution/container/BUILD
@@ -23,8 +23,11 @@ go_library(
         "//server/util/status",
         "//server/util/tracing",
         "//server/util/unixcred",
+        "@com_github_jonboulle_clockwork//:clockwork",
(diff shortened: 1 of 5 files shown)


bduffany:
Discussed offline with Tyler - will try out an approach based on compression of the raw samples using something like zstd, since it's more flexible

vanja-p (commented):
I was in the middle of reviewing this so I'll just leave the comments I had, even though they don't matter if you're trying a different approach.

I'm not sure how the alternative approach will look, but my bias is to keep the data model simple and not have unnecessary precision. This data will have much variance, so keeping a high precision is more likely to be misleading than useful.

I would also make it easy for users to see this data.

vanja-p:
Also, any custom format will be harder to analyze in SQL (or any other language where you don't have parsing code). That said, I'm not sure how easy it would be to analyze the proto you already added in SQL.

PR #7780, opened 2024-10-22
[FC] Ensure VM cleanup, even if creating a snapshot fails

maggie-lou:
We're suspicious that the failure to unmount VBDs has contributed to disk buildup on our executors. The logs show that most of the failures are a result of "context canceled"

This PR lets us retry failed VBD unmounts with an extended context, and also ensures we run additional VM cleanup (like networking cleanup), even if creating the snapshot fails.

Slack: https://buildbuddy-corp.slack.com/archives/C06N9AHA4JX/p1729535844869529?thread_ts=1729527018.654009&cid=C06N9AHA4JX

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -2372,7 +2372,7 @@ func (c *FirecrackerContainer) remove(ctx context.Context) error {
 		c.vfsServer = nil
 	}
 
-	if err := c.unmountAllVBDs(ctx); err != nil {
+	if err := c.unmountAllVBDs(ctx, true /*logErrors*/); err != nil {
 		// Don't log the err - unmountAllVBDs logs it internally.
 		lastErr = err
 	}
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -2489,6 +2495,16 @@ func (c *FirecrackerContainer) pause(ctx context.Context) error {
...
 
+	// Even if we fail to create a snapshot, we still want to cleanup the VM.
+	deferRemove := true
+	defer func() {
+		if deferRemove {
+			if err := c.Remove(ctx); err != nil {
+				log.Warningf("Failed to remove container during early cleanup: %s", err)
+			}
+		}
+	}()
```
bduffany, line 2506: In `runner.go` > `TryRecycle`, we call `Remove()` if `Pause()` fails. So I think this might not be needed?

bduffany: OK I think this makes sense - LGTM

bduffany (approved):

PR #7782, opened 2024-10-22
Fix replay_action not waiting for execute response

bduffany:


enterprise/tools/replay_action/replay_action.go
```diff
@@ -394,14 +394,22 @@ func (r *Replayer) execute(ctx, srcCtx, targetCtx context.Context, action *repb.
...
 		DigestFunction:  sourceExecutionRN.GetDigestFunction(),
 	}
+	executeErr := make(chan error, *n)
 	for i := 1; i <= *n; i++ {
 		i := i
 		r.executeGroup.Go(func() error {
-			execute(targetCtx, r.execClient, r.destBSClient, i, sourceExecutionRN, execReq)
-			return nil
+			err := execute(targetCtx, r.execClient, r.destBSClient, i, sourceExecutionRN, execReq)
+			executeErr <- err
```
iain-macdonald, line 402: do you want to do this if err is nil? sorry, not sure what happens if you send nil to a chan

bduffany: yeah, this is just collecting all of the return values. I'd call this 'results' or something but I feel like 'result' is too overloaded of a term in this context :thinking:

iain-macdonald: sg, may want to log these somewhere but up to you

iain-macdonald (approved):

vanja-p (approved):

enterprise/tools/replay_action/replay_action.go
bduffany: Done (we were logging only the last error before, but now we're logging every error)

PR #7785, opened 2024-10-22
oci runtime: respect ENTRYPOINT

bduffany:
Context: https://buildbuddy-corp.slack.com/archives/C05GZM1EGM8/p1729631149867599?thread_ts=1727895825.533649&cid=C05GZM1EGM8

* Respect ENTRYPOINT in image config (one user's actions are working in podman but not OCI because they are relying on ENTRYPOINT)
* Consolidate/cleanup some test registry code
* Add a test to ensure that we respect ENTRYPOINT


server/testutil/testregistry/testregistry.go
```diff
@@ -97,3 +99,20 @@ func (r *Registry) PushRandomImage(t *testing.T) string {
...
+// The rlocationpath should be set via x_defs in the BUILD file, and the
+// rlocationpath target should be an OCI image target (e.g. oci.pull)
+func ImageFromRlocationpath(t *testing.T, rlocationpath string) v1.Image {
+	indexPath, err := runfiles.Rlocation(rlocationpath)
+	require.NoError(t, err)
+	idx, err := layout.ImageIndexFromPath(indexPath)
+	require.NoError(t, err)
+	m, err := idx.IndexManifest()
+	require.NoError(t, err)
+	require.Equal(t, 1, len(m.Manifests))
```
fmeum, line 113: Nit:
```suggestion
	require.Len(t, 1, m.Manifests)
```

fmeum (approved):

sluongng (approved):
1. Nice, I did not know we have a `testregistry` package 😎 

2. Could be pedantic, but it would be nice to get 2 things tested with EntryPoint support.
  + Recycle runner. Does the Entrypoint script get runs multiple times or only once?

  + Working directory. Not sure if the image's entrypoint script could be expecting a different working directory?


We can also add these in a separate PR when we need to verify them.

server/testutil/testregistry/testregistry.go
bduffany: Done

bduffany:
> Recycle runner. Does the Entrypoint script get runs multiple times or only once?

ENTRYPOINT is executed during `Create()` but not `Exec()`. So if it's just setting env vars or something, then the executed actions wouldn't see those env vars. I don't know if it's worth adding a test for this, but we can maybe think about this later if people run into issues

> Working directory. Not sure if the image's entrypoint script could be expecting a different working directory?

ENTRYPOINT doesn't affect the working dir - the entrypoint process is still run from the action workdir. After thinking about it a bit, I don't think this needs to be tested explicitly - entrypoint does respect working dir, but it also respects env, cgroups, user, etc. and it doesn't seem worth testing all of these other properties of the process too

sluongng:
Hmm... what if the image has a `WORKDIR` directive before the `ENTRYPOINT`?

For example

```
FROM busybox
CMD mkdir -p /foo/bar
WORKDIR /foo/bar
ENTRYPOINT ["touch", "abc"]
```

where would the `abc` file get created? 🤔

bduffany:
@sluongng we currently ignore `WORKDIR` whether or not `ENTRYPOINT` is used, so `abc` would be created in the action workspace. This is also the case for podman.

Also, I am not sure if this is part of your question or not, but setting `WORKDIR` before `ENTRYPOINT` is the same as setting it after.

sluongng:
Ohhh. I confused with something else then 🤔 

Please proceed 🙏

PR #7786, opened 2024-10-23
Jh/2024 10 21 filter status field

jdhollen:


server/util/filter/filter.go
```diff
@@ -5,7 +5,9 @@ import (
 	"slices"
 	"strings"
 
+	ispb "github.com/buildbuddy-io/buildbuddy/proto/invocation_status"
```
iain-macdonald, line 8: nit: move to aliased imports section

iain-macdonald (approved):

PR #7787, opened 2024-10-23
server: Allow inline styles when code editor is enabled

fmeum:
Monaco is highly incompatible with a CSP that forbids inline styles.

### DIFF
--- a/server/features/features.go
+++ b/server/features/features.go
@@ -1,0 +1,8 @@
+package features
+
+import "github.com/buildbuddy-io/buildbuddy/server/util/flag"
+
+var (
+	CodeEditorEnabled   = flag.Bool("app.code_editor_enabled", false, "If set, code editor functionality will be enabled.")
+	CodeEditorV2Enabled = flag.Bool("app.code_editor_v2_enabled", false, "If set, show v2 of code editor that stores state on server instead of local storage.")
+)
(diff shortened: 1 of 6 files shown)


server/http/interceptors/interceptors.go
```diff
@@ -66,20 +75,23 @@ func getContentSecurityPolicyHeaderValue(nonce string) string {
...
+func setContentSecurityPolicy(h http.Header, u *url.URL) string {
 	nonceBytes := make([]byte, 16)
 	_, err := rand.Read(nonceBytes)
 	if err != nil {
 		panic(fmt.Sprintf("Failed to generate nonce: %s", err))
 	}
 	nonce := base64.StdEncoding.EncodeToString(nonceBytes)
+	// Allow inline styles for /code/ to support the Monaco editor.
+	// https://github.com/microsoft/monaco-editor/issues/271
+	allowInlineStyles := u.Path == "/code" || strings.HasPrefix(u.Path, "/code/")
```
bduffany, line 92: This server-side path check won't work unfortunately, because we have client-side navigation. Maybe we could gate this behind whether the code feature is enabled or not? I think it's behind a flag already

server/features/features.go
```diff
@@ -0,0 +1,8 @@
+package features
```
bduffany, line 1: I think this is a useful addition because there have been a few times where having a feature flag defined in `static.go` hasn't been convenient because of certain dependency chains.

It might be confusing having some flags defined here and some defined in `static.go`, though.

Maybe add a TODO here to move the other feature flags from `static.go` into this file?

bduffany (approved):

server/features/features.go
fmeum: I added the TODO to the existing flags in `static.go`. Should I submit a follow-up change that moves them?

PR #7788, opened 2024-10-23
After executing, don't upload outputs that already exist remotely

vanja-p:
In local and dev experiments, this prevents the upload of about 30% of bytes, and makes the output_upload stage about 40% faster. The speed increase isn't really meaningful, since the output_upload stage only takes 13ms on average.

Some things I'm not sure about:

1. Instead of creating a new metric, should I add a dimension to an existing metric, like `buildbuddy_remote_execution_file_upload_size_bytes`? I'm leaning towards creating this new metric because we don't need a histogram for skipped bytes.
2. Instead of adding a new metric, should I just report the reduced amount to `buildbuddy_remote_execution_file_upload_size_bytes`? If I did that, it would be harder to know how many bytes were skipped, and I think it's still useful to know how many bytes would be uploaded without the optimization.
3. Should I update the attributes in the trace, or add another attribute here: https://github.com/buildbuddy-io/buildbuddy/blob/36907cc2adc9c49c4d80e2ee07140770935b33de/enterprise/server/remote_execution/workspace/workspace.go#L330-L331

Closes buildbuddy-io/buildbuddy-internal#3849

### DIFF
--- a/enterprise/server/remote_execution/dirtools/dirtools.go
+++ b/enterprise/server/remote_execution/dirtools/dirtools.go
@@ -273,10 +273,15 @@ func (f *fileToUpload) FileNode() *repb.FileNode {
 	}
 }
 
-func uploadMissingFiles(ctx context.Context, uploader *cachetools.BatchCASUploader, env environment.Env, filesToUpload []*fileToUpload, instanceName string, digestFunction repb.DigestFunction_Value) error {
+func uploadMissingFiles(ctx context.Context, uploader *cachetools.BatchCASUploader, env environment.Env, filesToUpload []*fileToUpload, instanceName string, digestFunction repb.DigestFunction_Value) (skippedFiles, skippedBytes int64, _ error) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	batches := make(chan []*fileToUpload, 1)
+
+	type batchResult struct {
+		files                      []*fileToUpload
+		skippedFiles, skippedBytes int64
+	}
+	batches := make(chan batchResult, 1)
 	var wg sync.WaitGroup
 	cas := env.GetContentAddressableStorageClient()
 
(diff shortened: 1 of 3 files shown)


bduffany (commented):
re. the questions in the PR description, I think anywhere that we have a metric like `uploaded` or `transferred`, we should only report the count of bytes sent over the network, otherwise it's confusing. If we want to know the total output size I think we could derive that from uploaded + skipped? (`file_upload_size_bytes` is a histogram but histograms implicitly have a `*_sum` metric which is a counter). Or if you want a histogram for total output size, that could be a separate histogram metric, but I think it's maybe not that useful?

vanja-p:
That makes sense. I kept my skipped metric, and changed the existing uploaded metrics to ignore skipped files.

I'll fix this up a bit in the next change so that stdout, stderr, and auxiliary logs count towards the uploaded metrics.

enterprise/server/remote_execution/dirtools/dirtools.go
```diff
@@ -459,9 +529,13 @@ func UploadTree(ctx context.Context, env environment.Env, dirHelper *DirHelper,
...
 	// Upload output files to the remote cache and also add them to the local
 	// cache since they are likely to be used as inputs to subsequent actions.
-	if err := uploadFiles(ctx, uploader, env.GetFileCache(), filesToUpload); err != nil {
+	skippedFiles, skippedBytes, err := uploadMissingFiles(ctx, uploader, env, filesToUpload, instanceName, digestFunction)
+	if err != nil {
 		return nil, err
 	}
+	metrics.SkippedOutputBytes.Add(float64(skippedBytes))
+	txInfo.FileCount -= skippedFiles
+	txInfo.BytesTransferred -= skippedBytes
```
bduffany, line 538: nit: this subtraction after the fact feels slightly error prone, wdyt about passing txInfo as a param to uploadMissingFiles and having that func be responsible for incrementing FileCount/BytesTransferred? (would also let us avoid having 3 return values from that func)

bduffany (approved):

enterprise/server/remote_execution/dirtools/dirtools.go
vanja-p: I'll do this in next change, since I have a few other related cleanups.

PR #7790, opened 2024-10-23
oci runtime: preserve file ownership

bduffany:
Fixes the issue mentioned here: https://buildbuddy-corp.slack.com/archives/C05GZM1EGM8/p1729695120908659?thread_ts=1727895825.533649&cid=C05GZM1EGM8


sluongng (approved):
Lgtm

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1247,7 +1247,7 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
 	defer os.RemoveAll(tempUnpackDir)
 
 	// TODO: avoid tar command.
-	cmd := exec.CommandContext(ctx, "tar", "--no-same-owner", "--extract", "--directory", tempUnpackDir)
+	cmd := exec.CommandContext(ctx, "tar", "--extract", "--directory", tempUnpackDir)
```
bduffany, line 1250: This is the only non-test change in this PR (removing `--no-same-owner`). This flag was causing permissions errors when running as non-root, but a while back I decided not to support non-root for `ociruntime` for now, since it adds a lot of complexity.

PR #7793, opened 2024-10-24
Ensure that `//cli` and `//server` do not depend on `//enterprise/...`

fmeum:
Our license file states that everything outside `//enterprise/...` that isn't third-party is MIT licensed. However, this doesn't rule out that building any target outside `//enterprise/...` depends on code that isn't MIT.

This change moves `dirtools` under `//server` and adds tests that catch regressions.


sluongng (commented):
might need to update the package visibility in server/cache/dirtools/BUILD

tylerwilliams (approved):
Are there visibility rules we can enable too?

siggisim (approved):

fmeum:
> Are there visibility rules we can enable too?

I deleted the exception.

PR #7794, opened 2024-10-24
oci runtime: support setting CPU hard-limits based on task size

bduffany:
Add a flag to set a CPU hard limit relative to the task size.

The eventual goal would be to set this new flag to `0` so that we are setting exact hard limits based on the task size, and the app would be responsible for building any "headroom" needed into the task size, since the app has more task sizing information and knows how "confident" it is about the task size.

Note that we currently also have `executor.oci.cpu_limit` which sets an absolute hard-limit. This absolute limit is now applied after the relative limit is applied.

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -44,12 +44,13 @@ import (
 )
 
 var (
-	Runtime     = flag.String("executor.oci.runtime", "", "OCI runtime")
-	runtimeRoot = flag.String("executor.oci.runtime_root", "", "Root directory for storage of container state (see <runtime> --help for default)")
-	pidsLimit   = flag.Int64("executor.oci.pids_limit", 2048, "PID limit for OCI runtime. Set to -1 for unlimited PIDs.")
-	cpuLimit    = flag.Int("executor.oci.cpu_limit", 0, "Hard limit for CPU resources, expressed as CPU count. Default (0) is no limit.")
-	dns         = flag.String("executor.oci.dns", "8.8.8.8", "Specifies a custom DNS server for use inside OCI containers. If set to the empty string, mount /etc/resolv.conf from the host.")
-	netPoolSize = flag.Int("executor.oci.network_pool_size", 0, "Limit on the number of networks to be reused between containers. Setting to 0 disables pooling. Setting to -1 uses the recommended default.")
+	Runtime          = flag.String("executor.oci.runtime", "", "OCI runtime")
+	runtimeRoot      = flag.String("executor.oci.runtime_root", "", "Root directory for storage of container state (see <runtime> --help for default)")
+	pidsLimit        = flag.Int64("executor.oci.pids_limit", 2048, "PID limit for OCI runtime. Set to -1 for unlimited PIDs.")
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -44,12 +44,13 @@ import (
...
-	dns         = flag.String("executor.oci.dns", "8.8.8.8", "Specifies a custom DNS server for use inside OCI containers. If set to the empty string, mount /etc/resolv.conf from the host.")
-	netPoolSize = flag.Int("executor.oci.network_pool_size", 0, "Limit on the number of networks to be reused between containers. Setting to 0 disables pooling. Setting to -1 uses the recommended default.")
+	Runtime          = flag.String("executor.oci.runtime", "", "OCI runtime")
+	runtimeRoot      = flag.String("executor.oci.runtime_root", "", "Root directory for storage of container state (see <runtime> --help for default)")
+	pidsLimit        = flag.Int64("executor.oci.pids_limit", 2048, "PID limit for OCI runtime. Set to -1 for unlimited PIDs.")
+	cpuLimit         = flag.Int("executor.oci.cpu_limit", 0, "Hard limit for CPU resources, expressed as CPU count. Default (0) is no limit.")
+	cpuRelativeLimit = flag.Int("executor.oci.cpu_relative_limit", -1, "CPU hard limit added to the task size, expressed as CPU count. If set to -1, only use cpu_limit as a hard limit.")
```
tylerwilliams, line 51: Should this be millicpu or is 1 cpu (1000 milliCPU the min overhead?)

Could we call this executor.oci.task_buffer_millicpu or executor.oci.task_overhead_millicpu to make it clear that it's an amount of CPU to be added to the task size hard limit?

tylerwilliams (commented):
See comments

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -692,13 +697,18 @@ func (c *ociContainer) createSpec(ctx context.Context, cmd *repb.Command) (*spec
 		return nil, fmt.Errorf("get container user: %w", err)
 	}
 
-	cpuSpecs := &specs.LinuxCPU{}
+	var milliCPUHardLimit int64
+	if *cpuRelativeLimit >= 0 {
+		milliCPUHardLimit = c.milliCPU + int64(*cpuRelativeLimit)*1000
+	}
 	if *cpuLimit != 0 {
```
vanja-p, line 704: This wasn't handled before, but what happens if this flag is negative? Maybe that should just be ignored.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: Changed to milli-CPU and went with `executor.oci.task_millicpu_headroom` - `buffer` is a bit too overloaded imo and `overhead` sounds like some fixed cost that we're trying to offset

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: Changed to `> 0`

vanja-p (approved):

bduffany:
Closing since the plan is now to set hard limits via CgroupSettings in the app

PR #7796, opened 2024-10-24
support string array filters in clickhouse (for tags)

jdhollen:


server/interfaces/interfaces.go
```diff
@@ -343,6 +343,8 @@ type DB interface {
 	GORM(ctx context.Context, name string) *gorm.DB
 
 	NowFunc() time.Time
+
+	DialectName() string
```
iain-macdonald, line 347: mind adding a todo to return an enum?

iain-macdonald (approved):

server/interfaces/interfaces.go
jdhollen: sure

PR #7799, opened 2024-10-25
Include stdout, stderr, and auxiliary logs in file upload metrics

vanja-p:
Now the `file_upload_count` and `file_upload_size_bytes` metrics will include these files and bytes.

Also addresses the last comment from buildbuddy-io/buildbuddy/pull/7788

### DIFF
--- a/enterprise/server/remote_execution/workspace/workspace.go
+++ b/enterprise/server/remote_execution/workspace/workspace.go
@@ -324,7 +324,7 @@ func (ws *Workspace) UploadOutputs(ctx context.Context, cmd *repb.Command, execu
 	if err := eg.Wait(); err != nil {
 		return nil, err
 	}
-	txInfo.FileCount += 2 //for stdout and stderr
+	txInfo.FileCount += 2 // for stdout and stderr
 	txInfo.BytesTransferred += int64(len(cmdResult.Stdout) + len(cmdResult.Stderr))
 	txInfo.FileCount += int64(len(cmdResult.AuxiliaryLogs))
 	for _, b := range cmdResult.AuxiliaryLogs {
--- a/enterprise/server/remote_execution/workspace/workspace.go
+++ b/enterprise/server/remote_execution/workspace/workspace.go
@@ -324,6 +324,12 @@ func (ws *Workspace) UploadOutputs(ctx context.Context, cmd *repb.Command, execu
 	if err := eg.Wait(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


vanja-p:
One thing I'm not sure how to resolve. `cachetools.BatchCASUploader` can avoid duplicate uploads in the batch: https://github.com/buildbuddy-io/buildbuddy/blob/974a8bb054f8cb75224f9ac39a731faa21fe3d35/server/remote_cache/cachetools/cachetools.go#L615-L620

So maybe these should also be counted as skipped bytes/files and removed from the `file_upload_...` metrics, but this operates on all operations of the batch, regardless of the blob type (file, dir, tree).

Currently:
- each file increments file_upload_count and file_upload_size based on the size of the file
- each directory also increments file_upload_count and file_upload_size based on the size of the Directory proto
- tree uploads are ignored for these metrics.

So it's not clear how to reconcile this. Should uploaded trees count as files? That seems like it could duplicate the file count. What about their bytes?

Why do directories count as uploaded files? This doesn't seem consistent with the docs: https://github.com/buildbuddy-io/buildbuddy/blob/6bb9e682a16c2a231e6409bdbec89593983bfc1f/proto/remote_execution.proto#L2189-L2193

My thinking is to leave this as is, despite the fact that it's inconsistent. Alternatively, include trees in the `file_upload_...` metrics and account for uploads skipped by `cachetools.BatchCASUploader`.

tylerwilliams (approved):
LGTM but can you make sure @bduffany is cool with it before submitting?

bduffany:
@vanja-p I think there isn't much value distinguishing between files and Directory/Tree protos - the thing we really care about is just measuring everything that we upload. So ideally we'd count all uploaded blobs in that file_upload_count / file_upload_size_bytes.

If it's clearer, we could rename those to `cas_upload_count` / `cas_upload_size_bytes` (should be a compatible change since I don't believe these are consumed as JSON anywhere)

vanja-p:
Cool, that actually makes it a bit cleaner to implement since I can use the BatchCASUploader to keep track of uploads. I do have to separately track digests that were already present with FindMissing, and that's a bit ugly.

Maybe the BatchCASUploader should handle calling FindMissing and then it could keep track of all the stats, but I didn't think it was worth making it even more complicated. Also, because its method accept an io.Reader, that means we would have to open files that are already present and we don't need to read them.

vanja-p:
I fixed a bunch of tests because the result of dirtools.UploadTree changed.

bduffany (approved):

PR #7801, opened 2024-10-25
server: Fix logspam caused by regression in otelhttp

fmeum:
Since #7769, which updated `otelhttp`, the following line is logged many times per second:

```
2024/10/25 12:52:58 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/request.(*RespWriterWrapper).writeHeader (resp_writer_wrapper.go:78)
```

This is caused by a regression in the module, which has been fixed in https://github.com/open-telemetry/opentelemetry-go-contrib/commit/54d69164e7e6e337b3da441573cfea85279c2a3b.


bduffany:
CC @tylerwilliams in case it's worth cherry-picking this (there are about ~10K of these errors logged per minute)

PR #7806, opened 2024-10-25
[Cache Proxy] halt atime_updater goroutine on server shutdown

iain-macdonald:


enterprise/server/atime_updater/atime_updater.go
```diff
@@ -318,3 +331,8 @@ func (u *atimeUpdater) update(ctx context.Context, groupID string, jwt string, r
 		log.CtxWarningf(ctx, "Error sending FindMissingBlobs request to update remote atimes for group %s: %s", groupID, err)
 	}
 }
+
+func (u *atimeUpdater) shutdown(ctx context.Context) error {
+	u.quit <- struct{}{}
```
vadimberezniker, line 336: we usually `close` the quit channel rather than sending a value. it's the same in this case, but if you ever have more than one goroutine only one would get the close signal!

vadimberezniker (approved):

PR #7809, opened 2024-10-25
Record CPU usage timelines, take 2

bduffany:
* Use streamvbyte to compress raw samples collected from the executor.
* With the raw sample approach, the executors have to do minimal processing (as clients) and the apps have maximum flexibility to process/analyze data as desired.
* Results from CPU data for `remote_execution_test` which takes 279s:

```
5667 samples
streamvbyte: 17.8K in 14.286µs
streamvbyte_delta: 7.5K in 13.725µs
zstd: 22.7K in 43.22µs
```


server/util/timeseries/timeseries.go
```diff
@@ -0,0 +1,88 @@
+package timeseries
```
tylerwilliams, line 1: can you send this separately first?

bduffany: sent https://github.com/buildbuddy-io/buildbuddy/pull/7810/files

vanja-p:
If you stored this in simple proto fields like
```
repeated int64 timestamps
repeated int64 cpu_samples
```
and you used delta compression, I think each sample would take 1-2 bytes for the timestamp and 1-2 bytes for the sample. For 5667 samples, that works out to between 11.33K and 22.66K. It should be simple to implement without holding all samples in memory.

tylerwilliams (approved):

bduffany:
@vanja-p `repeated int64` with delta compression seems to work well when testing with the bb repo.

Here are the compression ratios when using various methods:

```
Stats over 582 actions
uncompressed               avg: 1.000   p90: 1.000   p99: 1.000
streamvbyte_delta          avg: 0.327   p90: 0.335   p99: 0.342
pb_repeated_int64          avg: 0.498   p90: 0.756   p99: 0.757
pb_repeated_int64_delta    avg: 0.136   p90: 0.145   p99: 0.156
```

I'll update this PR to switch to int64 + delta compression instead.

vanja-p (approved):

enterprise/server/remote_execution/container/container.go
```diff
@@ -263,9 +293,32 @@ func (s *UsageStats) TaskStats() *repb.UsageStats {
...
+		return
+	}
+	// Update timestamps
+	ts := now.UnixMilli()
+	tsDelta := ts - s.timelineState.lastTimestamp
+	s.timeline.Timestamps = append(s.timeline.Timestamps, tsDelta)
+	s.timelineState.lastTimestamp = ts
+
+	// Update CPU samples
+	cpu := (s.last.GetCpuNanos() - s.baselineCPUNanos) / 1e6
```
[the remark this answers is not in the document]
bduffany: Done

PR #7823, opened 2024-10-29
[RB] Support remote header overrides

maggie-lou:


proto/api/v1/remote_runner.proto
```diff
@@ -28,6 +28,17 @@ message RunRequest {
...
+  // Platform properties that should be set via remote header override.
+  // These apply to the remote runner (and not the commands run by the runner).
+  //
+  // These are similar to `platform_properties`, except they will not be
+  // included in the snapshot key (which contains regular platform properties)
+  // and won't be displayed on the executions page.
+  //
+  // They are useful for passing secrets that you don't want displayed or
+  // short-lived credentials that you don't want invalidating the snapshot key.
+  map<string, string> remote_header_overrides = 10;
```
bduffany, line 40: I think I would prefer calling this `remote_headers` and blindly pass them through as gRPC headers in the Execute request. So people would have to set `"x-buildbuddy-platform.NAME": "value"` rather than just `"name": "value"`. This would allow people to set whatever headers they want, not just platform overrides, and is also more consistent with our docs on how to pass secret headers today (`--remote_header=x-buildbuddy-platform.SECRET_NAME=SECRET_VALUE`)

If you were aiming for something more user friendly, then another option is to call this `secret_platform_properties` and avoid mentioning anything about remote headers, since `remote_header_overrides` seems like a bit of an implementation detail, rather than directly communicating the intended purpose. Personally though, I would go with `remote_headers` since it's more flexible and is also more consistent with the normal RBE APIs.

maggie-lou: Good call - updated

cli/remotebazel/remotebazel.go
```diff
@@ -78,6 +78,7 @@ var (
 	remoteRunner            = RemoteFlagset.String("remote_runner", defaultRemoteExecutionURL, "The Buildbuddy grpc target the remote runner should run on.")
 	timeout                 = RemoteFlagset.Duration("timeout", 0, "If set, requests that have exceeded this timeout will be canceled automatically. (Ex. --timeout=15m; --timeout=2h)")
 	execPropsFlag           = bbflag.New(RemoteFlagset, "runner_exec_properties", []string{}, "Exec properties that will apply to the *ci runner execution*. Key-value pairs should be separated by '=' (Ex. --runner_exec_properties=NAME=VALUE). Can be specified more than once. NOTE: If you want to apply an exec property to the bazel command that's run on the runner, just pass at the end of the command (Ex. bb remote build //... --remote_default_exec_properties=OSFamily=linux).")
+	remoteHeaders           = bbflag.New(RemoteFlagset, "remote_header", []string{}, "Remote headers to be applied to the execution request for the remote run. Can be used to set platform properties containing secrets (Ex. --remote_header=x-buildbuddy-platform.SECRET_NAME=SECRET_VALUE). Can be specified more than once.")
```
bduffany, line 81: for the flag name specifically, we might want to call it `--runner_remote_header`. The "runner" prefix helps clarify which flags are applied to the runner, and which flags are applied to the inner bazel invocation

For the APIs, it seems fine to omit the "runner" prefix since it seems like that's the convention we're currently using.

bduffany (approved):

PR #7826, opened 2024-10-29
Update to new `buildbuddy-toolchain`

tempoz:
- **update some toolchain and platform stuff**
- **add necessary rootDirs for new platform names**

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -40,8 +40,7 @@ common:remote-shared --remote_timeout=600
 common:remote-shared --remote_download_minimal
 common:remote-shared --jobs=100
 common:remote-shared --verbose_failures
-common:remote-shared --platforms=//platforms:linux_x86_64
-common:remote-shared --extra_execution_platforms=//platforms:linux_x86_64
+common:remote-shared --extra_execution_platforms=@buildbuddy_toolchain//:platform_linux
 
 # Build with --config=remote to use BuildBuddy RBE.
 common:remote --config=remote-shared
--- a/.bazelrc
+++ b/.bazelrc
@@ -45,14 +45,15 @@ common:remote-shared --verbose_failures
 common:remote --config=remote-shared
 common:remote --config=cache
 common:remote --remote_executor=grpcs://buildbuddy.buildbuddy.io
-common:remote --platforms=@buildbuddy_toolchain//:platform_linux
-common:remote --extra_execution_platforms=@buildbuddy_toolchain//:platform_linux
+common:remote --platforms=@buildbuddy_toolchain//:platform_linux_x86_64
+common:remote --extra_execution_platforms=@buildbuddy_toolchain//:platform_linux_x86_64
 
-# Specify arch to do cross-platform builds on remote
-common:remote-linux-amd64 --config=remote
-common:remote-linux-amd64 --extra_execution_platforms=@buildbuddy_toolchain//:platform_linux_x86_64
-
-common:remote-linux-arm64 --config=remote
+# Specify arch to do cross-platform builds on remote until the go toolchain can
+# accomodate multiple execution platforms
+common:remote-linux-arm64 --config=remote-shared
+common:remote-linux-arm64 --config=cache
+common:remote-linux-arm64 --remote_executor=grpcs://buildbuddy.buildbuddy.io
+common:remote-linux-arm64 --platforms=@buildbuddy_toolchain//:platform_linux_arm64
 common:remote-linux-arm64 --extra_execution_platforms=@buildbuddy_toolchain//:platform_linux_arm64
 
 # Build with --config=remote-dev to use BuildBuddy RBE.
--- a/.bazelrc
+++ b/.bazelrc
@@ -40,17 +40,32 @@ common:remote-shared --remote_timeout=600
 common:remote-shared --remote_download_minimal
 common:remote-shared --jobs=100
 common:remote-shared --verbose_failures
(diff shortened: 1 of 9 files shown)


.bazelrc
```diff
@@ -40,8 +40,7 @@ common:remote-shared --remote_timeout=600
 common:remote-shared --remote_download_minimal
 common:remote-shared --jobs=100
 common:remote-shared --verbose_failures
-common:remote-shared --platforms=//platforms:linux_x86_64
-common:remote-shared --extra_execution_platforms=//platforms:linux_x86_64
+common:remote-shared --extra_execution_platforms=@buildbuddy_toolchain//:platform_linux
```
sluongng, line 43: Without a target platform, we will not be able to build remotely from a MacOS host.
If you are planning to add a `remote-linux-arm64` config in `.bazelrc`, I think we can move these flags from `remote-shared` to `remote` so that `remote-linux-arm64` can inherit from `remote-shared`

MODULE.bazel
```diff
@@ -546,7 +546,3 @@ use_repo(
     "busybox",
     "busybox_linux_amd64",
 )
-
-register_toolchains(
-    "//toolchains:ubuntu_cc_toolchain",
-)
```
sluongng, line 552: Let's keep `register_toolchains` in 2 places: `WORKSPACE` for our existing workflow and `MODULE.bazel` for future bzlmod workflow.

The plan is to remove `WORKSPACE.bzlmod` when we manage to migrate away from the remaining dependencies.

WORKSPACE.bzlmod
```diff
@@ -195,3 +195,7 @@ browser_repositories(chromium = True)
 load("@io_bazel_rules_webtesting//web:go_repositories.bzl", web_test_go_repositories = "go_repositories")
 
 web_test_go_repositories()
+
+register_toolchains(
+    "@buildbuddy_toolchain//:ubuntu_cc_toolchain",
+)
```
sluongng, line 201: ditto: move this to `MODULE.bazel`

tsconfig.json
```diff
@@ -38,7 +38,13 @@
...
       "bazel-out/macos_arm64-opt/bin",
       "bazel-out/macos_x86_64-fastbuild/bin",
-      "bazel-out/macos_x86_64-opt/bin"
+      "bazel-out/macos_x86_64-opt/bin",
+      "bazel-out/platform_linux_arm64-fastbuild/bin",
+      "bazel-out/platform_linux_arm64-opt/bin",
+      "bazel-out/platform_linux-fastbuild/bin",
+      "bazel-out/platform_linux-opt/bin",
+      "bazel-out/platform_linux_x86_64-fastbuild/bin",
+      "bazel-out/platform_linux_x86_64-opt/bin"
```
sluongng, line 47: Why do we need both `platform_linux-fastbuild` and `platform_linux_x86_64-fastbuild` here?

sluongng (changes requested):

tsconfig.json
tempoz: this thing needs directories for every platform string or the build fails when you build with that platform. So long as both `platform_linux` and `platform_linux_x86_64` exist as platforms, we need both.

WORKSPACE.bzlmod
tempoz: Can't; we need the `@buildbuddy_toolchain` repo created by the `buildbuddy` call, and we can't `load` the `buildbuddy` call in `MODULE.bazel`. We'd need to make it an extension, I think? Which I have not looked into how to do.

MODULE.bazel
tempoz: Can't do it; see comment below.

WORKSPACE.bzlmod
sluongng: Oh you are right.

Let's use your approach for now. I will add support for bzlmod in our toolchain and change this part later.

sluongng (approved):
lgtm

.bazelrc
```diff
@@ -40,19 +40,34 @@ common:remote-shared --remote_timeout=600
 common:remote-shared --remote_download_minimal
 common:remote-shared --jobs=100
 common:remote-shared --verbose_failures
-common:remote-shared --platforms=//platforms:linux_x86_64
-common:remote-shared --extra_execution_platforms=//platforms:linux_x86_64
```
sluongng, line 44: I think removing the targeted platform out of 'remote-shared' config broke this PR's CI.

We have a bunch of stuff using 'remote-shared'

```
grep '=remote-shared' .bazelrc
common:remote --config=remote-shared
common:remote-dev --config=remote-shared
common:probers --config=remote-shared
common:probers-dev --config=remote-shared
common:linux-workflows --config=remote-shared
```

and those would require the same target+exec platform flags.

So you might need to move these platform flags to a separate config, then add those configs into probers, probers-dev and linux-workflows to unblock CI.

.bazelrc
tempoz: Done!

.bazelrc
tempoz: Done!

sluongng:
I added 3 changes to the PR:

- Move the platform flags to a separate bazel config so that we don't have to repeat multiple lines multiple times. In the future, we can just replace this with `--platforms=` after we have registered all the exec platforms statically.

- Add the new platform flags to `website/docusaurus.config.js`. This unblocks CI

- Remove the deps upgrade under `website`. This can be sent as a separate PR.

PR #7828, opened 2024-10-29
[Cache Proxy] don't attempt to locally read/write content for organizations with encryption enabled

iain-macdonald:


enterprise/server/byte_stream_server_proxy/byte_stream_server_proxy.go
```diff
@@ -47,16 +53,23 @@ func New(env environment.Env) (*ByteStreamServerProxy, error) {
...
 	}, nil
 }
 
 func (s *ByteStreamServerProxy) Read(req *bspb.ReadRequest, stream bspb.ByteStream_ReadServer) error {
 	ctx, spn := tracing.StartSpan(stream.Context())
 	defer spn.End()
 
+	if authutil.EncryptionEnabled(ctx, s.authenticator) {
+		metrics.ByteStreamProxyReads.With(
+			prometheus.Labels{metrics.CacheHitMissStatus: "miss"}).Inc()
```
vadimberezniker, line 69: do you want a separate label here? it might throw off the cache hit stats if encrypted data is always counted as a miss.

vadimberezniker (approved):

PR #7832, opened 2024-10-30
[raft] fix flaky txn test

luluz66:


enterprise/server/raft/store/store_test.go
```diff
@@ -1121,6 +1096,8 @@ func TestReplaceDeadReplica(t *testing.T) {
 			break
 		}
 	}
+
+	log.Infof("** check replica caught up.")
```
tylerwilliams, line 1100: intended?

tylerwilliams (approved):

enterprise/server/raft/store/store_test.go
luluz66: removed

PR #7833, opened 2024-10-31
authdb: allow admin cap to update+delete user api key

sluongng:
Modify authdb.go so that credentials with admin capability
can Update and Delete user's api keys.

### DIFF
--- a/enterprise/server/backends/authdb/authdb.go
+++ b/enterprise/server/backends/authdb/authdb.go
@@ -967,9 +967,19 @@ func (d *AuthDB) authorizeAPIKeyWrite(ctx context.Context, h interfaces.DB, apiK
 	if err != nil {
 		return nil, err
 	}
-	acl := perms.ToACLProto(&uidpb.UserId{Id: key.UserID}, key.GroupID, key.Perms)
-	if err := perms.AuthorizeWrite(&user, acl); err != nil {
-		return nil, err
+	// ORG_ADMIN capability grants special permissions to retrieve any key
+	// within the authenticated group.
+	// If the authenticated user doesn't have this capability within the group
+	// associated with the key, then perform the usual ACL check on the key.
+	caps, err := capabilities.ForAuthenticatedUserGroup(ctx, d.env, key.GroupID)
+	if err != nil {
+		return nil, status.WrapError(err, "get capabilities")
+	}
+	if !slices.Contains(caps, akpb.ApiKey_ORG_ADMIN_CAPABILITY) {
+		acl := perms.ToACLProto(&uidpb.UserId{Id: key.UserID}, key.GroupID, key.Perms)
+		if err := perms.AuthorizeWrite(&user, acl); err != nil {
+			return nil, err
+		}
 	}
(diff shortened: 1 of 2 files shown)


enterprise/server/backends/authdb/authdb.go
```diff
@@ -967,16 +967,19 @@ func (d *AuthDB) authorizeAPIKeyWrite(ctx context.Context, h interfaces.DB, apiK
 	if err != nil {
 		return nil, err
 	}
+	err = authutil.AuthorizeOrgAdmin(user, key.GroupID)
+	if err == nil {
+		// admins have write permissions on all keys in the group.
+		return key, nil
+	}
```
bduffany, line 974: `AuthorizeOrgAdmin` is not quite what we want here - we only want to allow API keys with ORG_ADMIN capability. See the authdb changes in https://github.com/buildbuddy-io/buildbuddy/pull/7666

enterprise/server/backends/userdb/userdb_test.go
```diff
@@ -966,17 +964,23 @@ func TestUserOwnedKeys_GetUpdateDeletePermissions(t *testing.T) {
...
 			if test.Accessor != "" {
 				accessorCtx = authUserCtx(accessorCtx, env, t, test.Accessor)
 			}
-			keys, err := adb.GetUserAPIKeys(accessorCtx, test.Accessor, ownerGroup.GroupID)
-			// Only the owner should be able to view or update the API key,
-			// regardless of role.
-			isAuthorized := test.Owner == test.Accessor
+			keys, err := adb.GetUserAPIKeys(accessorCtx, test.Owner, ownerGroup.GroupID)
+			// Only the owner and group's admin should be able to view or update the API key.
+			isAuthorized := test.Owner == test.Accessor || test.Accessor == groupAdminID
```
bduffany, line 983: I mentioned in the other comment but this isn't the behavior we want - we want to check ORG_ADMIN capability on the authenticated API key, not group admin role on the authenticated user.

The test code exercising ORG_ADMIN accessing personal keys is here: https://github.com/buildbuddy-io/buildbuddy/blob/3fd4dd9e23fb21995f6db370803d208a269ee8f7/enterprise/server/backends/userdb/userdb_test.go#L1410-L1431

It's already testing Create/List/Read functionality; it just needs Update/Delete tests to be added.

bduffany (changes requested):

enterprise/server/backends/authdb/authdb.go
sluongng: I was conflating admin role and admin capability. 🤦 

Will fix.

enterprise/server/backends/userdb/userdb_test.go
sluongng: I changed the implementation but this test would still require an update to check for the admin user. I did add the other test to validate admin capability from an API key though.

bduffany (approved):

sluongng:
Since this is auth-related code, gona merge after another +1 review

PR #7835, opened 2024-10-31
[RB] Migrate to use `Steps`

maggie-lou:
In preparation for this PR deprecating usage of `BazelCommands`, migrate call-sites to use the new `Steps` syntax

Also adds a way to name the outer invocation of your remote run

This PR should be merged before https://github.com/buildbuddy-io/buildbuddy/pull/7747


cli/remotebazel/remotebazel.go
```diff
@@ -1134,3 +1128,15 @@ func contains(m map[string]string, elem string) bool {
...
+func getSimplifiedCommand(bazelArgs []string) string {
+	filteredArgs := make([]string, 0)
+	for _, arg := range bazelArgs {
+		if !strings.HasPrefix(arg, "--") {
+			filteredArgs = append(filteredArgs, arg)
+		}
+	}
+
+	return strings.Join(filteredArgs, " ")
+}
```
bduffany, line 1142: I think this doesn't work for things like `bazel run -c opt //foo`? Can you use `parser.GetBazelCommandAndIndex` here to get the bazel command?

For the target pattern(s), it's probably not super straightforward to parse those, but you could also add a func like `parser.GetPatterns()` that does something basic for now (like any flag that doesn't start with `-` and  has `//` or `:` can be treated as a target) and add a TODO there to parse it properly.

(More generally, I think ideally we'd consolidate any parsing code into `parser.go` so that all of the hacks/known issues with flag parsing are consolidated in a central place rather than scattered in various places and hard to track down.)

maggie-lou: Done

cli/parser/parser.go
```diff
@@ -1069,6 +1069,23 @@ func GetBazelCommandAndIndex(args []string) (string, int) {
 	return "", -1
 }
 
+// GetBazelPatterns makes a best-attempt effort to return the pattern in a bazel command.
```
bduffany, line 1072: `GetFirstTargetPattern` maybe? (`Patterns` sort of indicates that it would return a `[]string` with each pattern - there can be multiple patterns in a command)

bduffany (approved):

PR #7836, opened 2024-10-31
Don't run podman network setup container if podman isn't enabled

bduffany:
If podman is not enabled then don't run the warmup container (which ensures the correct rule ordering) but do continue to run the other networking setup.


tylerwilliams (approved):
LGTM (thanks!)

vadimberezniker (approved):

PR #7842, opened 2024-11-01
[raft] fix shutdown order

luluz66:
1. stop liveness before leasekeeper. Previously, after we shutdown leasekeeper, liveness keepLeaseAlive which will send cas requests to the store. The leasekeeper will then continue queuing instructions causing leaseAgent.runloop unable to finish in time.
2. When we stop leaseagent, we will also stop the keepLeaseAlive. 
3. In lease.ensureValidLease, check if the context deadline is reached.

### DIFF
--- a/enterprise/server/raft/leasekeeper/leasekeeper.go
+++ b/enterprise/server/raft/leasekeeper/leasekeeper.go
@@ -166,9 +166,7 @@ func (la *leaseAgent) stop() {
 	la.cancel()
 	la.eg.Wait()
 
-	if la.l != nil {
-		la.l.Stop()
-	}
+	la.l.Stop()
 }
 
 func (la *leaseAgent) runloop() {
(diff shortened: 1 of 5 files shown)


enterprise/server/raft/leasekeeper/leasekeeper.go
```diff
@@ -165,6 +165,10 @@ func (la *leaseAgent) doSingleInstruction(ctx context.Context, instruction *leas
 func (la *leaseAgent) stop() {
 	la.cancel()
 	la.eg.Wait()
+
+	if la.l != nil {
```
tylerwilliams, line 169: do you need this check?

tylerwilliams (approved):

enterprise/server/raft/leasekeeper/leasekeeper.go
luluz66: Done.

PR #7843, opened 2024-11-01
Set bazel exit code for disconnected invocations in GetInvocation

maggie-lou:
https://buildbuddy-corp.slack.com/archives/C04S4M5U3BK/p1730321133924089?thread_ts=1730318886.632589&cid=C04S4M5U3BK

### DIFF
--- a/enterprise/server/api/api_server.go
+++ b/enterprise/server/api/api_server.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/backends/prom"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/hostedrunner"
+	"github.com/buildbuddy-io/buildbuddy/proto/invocation_status"
 	"github.com/buildbuddy-io/buildbuddy/proto/workflow"
 	"github.com/buildbuddy-io/buildbuddy/server/build_event_protocol/build_event_handler"
 	"github.com/buildbuddy-io/buildbuddy/server/environment"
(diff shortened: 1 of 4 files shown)


enterprise/server/api/api_server.go
```diff
@@ -103,6 +104,10 @@ func (s *APIServer) GetInvocation(ctx context.Context, req *apipb.GetInvocationR
 
 	invocations := []*apipb.Invocation{}
 	err = db.ScanEach(rq, func(ctx context.Context, ti *tables.Invocation) error {
+		bazelExitCode := ti.BazelExitCode
+		if ti.InvocationStatus == int64(invocation_status.InvocationStatus_DISCONNECTED_INVOCATION_STATUS) {
+			bazelExitCode = "disconnected"
+		}
```
tylerwilliams, line 110: wdyt about adding invocation status to the returned object rather than adding logic here in the api layer (which is mostly just copying to / from our internal stuff)?

maggie-lou: I considered that too, but I think it would be most helpful as a string and not an int, which would require some translation logic here anyway. I figured it was simpler to reuse an existing field which also has the benefit of having richer, more detailed information

bduffany: I agree with Tyler's comment - storing "disconnected" in `bazelExitCode` doesn't "type-check" because the invocation BES status (PARTIAL/DISCONNECTED/COMPLETED) is representing something different than the bazel exit code.

re `string` vs `int` I think probably a proto enum is more appropriate? (we might have to copy the private InvocationStatus enum to the API proto definition). enums will appear as strings in the JSON-marshaled object, e.g. `Status: inspb.InvocationStatus_DISCONNECTED_INVOCATION_STATUS` in Go will appear as `"status": "DISCONNECTED_INVOCATION_STATUS"` in JSON I think.

maggie-lou: Okay that's fair. Changed

tylerwilliams (approved):

PR #7845, opened 2024-11-04
server: More CSP fixes and cleaner reports

fmeum:
* Reports are now filtered (rudimentarily) to remove noise such as reports triggered by browser extensions.
* Reports are logged as raw JSON so that they are easier to discover via the GCP Log Explorer.
* Allow `data:` sources for images as that's very low risk and Monaco appears to require it, although it's unclear why.
* Allow workers when Monaco is enabled.

Work towards https://github.com/buildbuddy-io/buildbuddy-internal/issues/3911

### DIFF
--- a/server/http/csp/csp.go
+++ b/server/http/csp/csp.go
@@ -1,6 +1,7 @@
 package csp
 
 import (
+	"encoding/json"
 	"io"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


server/http/csp/csp.go
```diff
@@ -15,10 +16,49 @@ type Nonce struct{}
...
+	var filtered []string
+	for _, anyReport := range all {
+		var actualReport map[string]any
+		report, ok := anyReport.(map[string]any)
+		if !ok {
+			// All individual reports must be JSON objects.
+			continue
+		}
+		if actualReport, _ = report["csp-report"].(map[string]any); actualReport != nil {
+			if !ok || actualReport["source-file"] == "moz-extension" {
```
bduffany, line 48: `!ok` here will always evaluate to true I think?

```suggestion
			if actualReport["source-file"] == "moz-extension" {
```

server/http/csp/csp.go
```diff
@@ -15,10 +16,49 @@ type Nonce struct{}
...
+				continue
+			}
+		} else if actualReport, _ = report["body"].(map[string]any); actualReport != nil {
+			if !ok || actualReport["sourceFile"] == "chrome-extension" {
```
bduffany, line 52:
```suggestion
			if actualReport["sourceFile"] == "chrome-extension" {
```

bduffany (approved):

bduffany:
auto-merge strikes again :sweat_smile:

fmeum:
Woops, will send a follow-up.

PR #7854, opened 2024-11-06
Remove NUMA setting in jailer config

bduffany:
This prevents using all available CPUs on machines that have multiple nodes, since it gets translated to `cpuset.cpus` and `cpuset.mems`: https://github.com/firecracker-microvm/firecracker-go-sdk/blob/f74f43bb036d832e6599a5716b669b4f04242c80/jailer.go#L141-L144


bduffany:
Going to start with just randomizing the node for now, per https://github.com/buildbuddy-io/buildbuddy-internal/issues/4087#issuecomment-2460493694

PR #7856, opened 2024-11-06
Fix nil digest error logged when calling BatchReadBlobs

bduffany:
When `cachetools.BatchReadBlobs` returns an error for an individual blob, we forget to set `Digest` in the `BlobResponse` which results in a `Digest <nil> not found` error being logged.


tylerwilliams (approved):
Nice

PR #7857, opened 2024-11-06
Switch back to firecracker-go-sdk upstream repo

bduffany:
We no longer use the `UffdPrivileged` memory backend (regular `Uffd` works now) so we can stop depending on our forked repo.


tylerwilliams (approved):
🔥

PR #7861, opened 2024-11-07
Report cgroup `io.stat` info in UsageStats

bduffany:


bduffany:
err... this is incomplete - need to update `UsageStats` in `container.go` as well so that we correctly handle the runner recycling case.

tylerwilliams (approved):

PR #7864, opened 2024-11-08
Report PSI prom metrics as a fraction of exec duration

bduffany:
Currently, it's hard to interpret the stall time histogram because it's reporting absolute duration. High stall durations might be fine if the task ran for a relatively long time. So report stall durations as a fraction of execution duration instead.

### DIFF
--- a/enterprise/server/remote_execution/executor/executor.go
+++ b/enterprise/server/remote_execution/executor/executor.go
@@ -475,9 +475,10 @@ func (m *ActionMetrics) Report(ctx context.Context) {
 	}
 	// If the isolation type supports it, report PSI metrics.
 	if md != nil && (m.Isolation == string(platform.PodmanContainerType) || m.Isolation == string(platform.OCIContainerType)) {
-		observePSI("cpu", md.GetUsageStats().GetCpuPressure())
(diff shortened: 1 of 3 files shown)


jdhollen:
I guess there isn't a great way to report both of these without just.. reporting both?  I feel like magnitude is sometimes useful on its own as well, but understand why you'd what to have the percentage.

bduffany:
@jdhollen how do you interpret magnitude without knowing the associated duration for each task? would it help to instead have executor-wide metrics for total pressure stalling? (I would like to add these as well, separately)

jdhollen:
> @jdhollen how do you interpret magnitude without knowing the associated duration for each task? would it help to instead have executor-wide metrics for total pressure stalling across all processes (i.e. `/proc/pressure/*` contents)? (I would like to add these as well, separately)

I think both are useful in their moments--having an overall total in addition to the percentage seems good enough to me.  The totals without duration would let you see the difference between "we've massively overscheduled" and "we just kinda tip over the edge every once in a while", but i guess you could roughly accomplish the same using running task counts or something??

jdhollen (approved):

PR #7867, opened 2024-11-08
Clean up firecracker cgroups manually

bduffany:
The cgroups created by jailer aren't deleted automatically. From the [jailer docs](https://github.com/firecracker-microvm/firecracker/blob/main/docs/jailer.md#observations):

> It’s up to the user to handle cleanup after running the jailer. One way to do this involves registering handlers with the cgroup notify_on_release mechanism, while being wary about potential race conditions (the instance crashing before the subscription process is complete, for example).

Rather than using `notify_on_release`, this PR just deletes the cgroup after the firecracker process exits, since the cgroup should no longer have any pids present at that point.

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/4101

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -2465,6 +2465,16 @@ func (c *FirecrackerContainer) stopMachine(ctx context.Context) error {
 	if err := c.machine.StopVMM(); err != nil {
 		return status.WrapError(err, "kill firecracker")
 	}
+
+	defer func() {
+		// Once the VM exits, delete the cgroup.
+		// Jailer docs say that this cleanup must be handled by us:
+		// https://github.com/firecracker-microvm/firecracker/blob/main/docs/jailer.md#observations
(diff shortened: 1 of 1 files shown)


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -2465,6 +2465,16 @@ func (c *FirecrackerContainer) stopMachine(ctx context.Context) error {
 	if err := c.machine.StopVMM(); err != nil {
 		return status.WrapError(err, "kill firecracker")
 	}
+
```
maggie-lou, line 2468: Should this go after the `c.machine.Wait` block, as opposed to in a defer? I'm seeing a couple `wait for firecracker to exit: context deadline exceeded` errors in the logs. In that case, I think this removal code will also fail because there will still be a live process in the cgroup

bduffany: yeah, I guess whether this is deferred or not, we'll still probably have cases where the cgroup might not be cleaned up. Just moved after the `machine.Wait()` for now

maggie-lou (approved):

PR #7870, opened 2024-11-08
[COW] Do not lock the lru when unmapping

maggie-lou:
The background lru eviction channel is not keeping up, because we are locking the LRU when unmapping chunks. This means that the unmaps are happening serially. Unmapping is slow, and we do not need the LRU to be locked while different chunks are being unmapped

This situation will now be possible: 
```
evict m1
> m1.mu.Lock()
> lru.Lock()
> lru.Unlock()
> blocks before we unmap and unlock m1

// Another goroutine
evict m2
> m2.mu.Lock()
> lru.Lock() <- This is a safe operation, because we are only taking the lock to see if the lru contains the m2 key. There is no reason we need to wait for m1 to finish unmapping beforehand

... 
```

Fixes https://buildbuddy-corp.slack.com/archives/C0154UFBCJ2/p1731014993153979

### DIFF
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -51,6 +51,9 @@ const (
 
 	// Number of goroutines to run concurrently to convert a file to a COWStore.
 	fileConversionConcurrency = 8
+
+	// Number of goroutines to run concurrently to handle LRU evictions.
+	lruEvictionConcurrency = 2
 )
 
 var maxEagerFetchesPerSec = flag.Int("executor.snaploader_max_eager_fetches_per_sec", 1000, "Max number of chunks snaploader can eagerly fetch in the background per second.")
@@ -1325,7 +1328,9 @@ func NewMmapLRU() (*MmapLRU, error) {
 		return nil, err
 	}
 	ml.lru = l
-	ml.evictorGroup.Go(ml.processEvictions)
+	for range lruEvictionConcurrency {
+		ml.evictorGroup.Go(ml.processEvictions)
+	}
 	return ml, nil
 }
 
--- a/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
+++ b/enterprise/server/remote_execution/copy_on_write/copy_on_write.go
@@ -1379,10 +1379,12 @@ func (ml *MmapLRU) processEviction(m *Mmap) {
 	// the LRU lock, in order to avoid deadlocks.
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	ml.mu.Lock()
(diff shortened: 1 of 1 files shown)


maggie-lou:
The locking logic was originally added in [this PR](https://github.com/buildbuddy-io/buildbuddy/pull/5532). As a sanity check, here is a walk through of why this PR will not introduce a regression:

```
m1.Unmap()
> m1.Lock()
> actually unmap the data
> lru.Remove(m1)
        > block before the lru acquires the lock 

// Another goroutine
m2.Read()
> m2.Lock()
> m2.initMap()
       > lru.Add(m2)
               > lru.Lock()
               > actually add m2 to LRU
               > lru.evict(m1) 
                       > remove m1 from the actual LRU
                       > (in background) processEviction(m1)
                                  > m1.Lock() - cannot proceed until the original block resumes (first few lines)
              > lru.Unlock().  (first block cannot continue until this runs, because it's waiting for the lock)


// The first goroutine can continue with lru.Remove(m1), as the lru has been unlocked

```

bduffany (commented):
I read through `TestCOW_MmapLRUDoesNotDeadlock` again and I think it should give us pretty high confidence that this is safe - mind taking another glance at that test and make sure it's not doing anything silly? Would be helpful to have an extra set of eyes in case I missed something

Also, I think now that we're not holding the LRU lock - it should be safe to add another goroutine to the `evictorGroup`, instead of having just 1 goroutine - I think that change is also needed in order to make sure we keep up with eviction and fix the UnexpectedEvent alerts?

maggie-lou:
> `TestCOW_MmapLRUDoesNotDeadlock` 

The test looks good to me 👍  Good point on the eviction concurrency

enterprise/server/remote_execution/copy_on_write/copy_on_write.go
```diff
@@ -1368,9 +1371,15 @@ func (ml *MmapLRU) Close() error {
...
 	for m := range ml.evictions {
-		ml.processEviction(m)
+		eg.Go(func() error {
+			ml.processEviction(m)
+			return nil
+		})
 	}
+	eg.Wait()
 	return nil
 }
```
bduffany, line 1384: Starting a goroutine for every eviction seems expensive - I think instead, we want to do this on line 128:

```
for range lruConcurrency {
	ml.evictorGroup.Go(ml.processEvictions)
}
```

then each `processEvictions` worker will pop values off the same channel and stop when it's closed

maggie-lou: I wasn't sure if it was safe for 2 goroutines to concurrently range over a channel, but looks like that's okay

bduffany (approved):
LGTM

tylerwilliams (approved):

PR #7874, opened 2024-11-08
[FC] Add more logging to tell if VBD was unmounted in the background

maggie-lou:
Even after this [change](https://github.com/buildbuddy-io/buildbuddy/pull/7780), we are still seeing error logs "Failed to unmount VBD: context canceled". However the [unmount happens in a goroutine, and continues even if the context is canceled.](https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/remote_execution/vbd/vbd.go#L132)

Add some more logging so we can see if we are eventually unmounting these VBDs, or if there is a goroutine leak and something is permanently blocking them from being unmounted. In prod, where most of these errors occur, we only have info-level logging. In the more limited cases I saw in dev, there were debug logs that the VBD was eventually getting unmounted. So I don't think this is an actual problem, but the logging should make it more clear

### DIFF
--- a/enterprise/server/remote_execution/vbd/vbd.go
+++ b/enterprise/server/remote_execution/vbd/vbd.go
@@ -135,7 +135,8 @@ func (f *FS) Unmount(ctx context.Context) error {
 	case err := <-resultCh:
 		return err
 	case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):

bduffany (approved):
nice

PR #7877, opened 2024-11-12
workflows: add Action Names and Env to Run Workflow button

sluongng:
Let user decide which action to run from our web UI.
The field accepts multiple action names separated by comma (`,`).

Custom input could be provided via environment variables.
The environment variables should be specified in comma-separated key-value pairs.
For example: `VAR1=value1,VAR2=value2`.

### DIFF
--- a/enterprise/app/workflows/workflows.tsx
+++ b/enterprise/app/workflows/workflows.tsx
@@ -418,14 +418,15 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
       .executeWorkflow(
         new workflow.ExecuteWorkflowRequest({
           pushedRepoUrl: this.props.repoUrl,
+          // Parse "var=val" pairs from the input field to Record<string, string>
           env: this.state.runWorkflowActionNames
             .split(",")
             .filter((n) => n.includes("="))
-            .reduce((acc: Record<string, string>, n) => {
-              const [k, v] = n.split("=");
-              if (k && v) {
-                acc[k] = v;
-              }
+            .map((n) => n.trim().split("="))
+            .filter(([k, v]) => k && v)
+            .map(([k, v]) => [k.trim(), v.trim()])
+            .reduce((acc: Record<string, string>, [k, v]) => {
+              acc[k] = v;
               return acc;
             }, {}),
           actionNames: this.state.runWorkflowEnvs
--- a/enterprise/app/workflows/workflows.tsx
+++ b/enterprise/app/workflows/workflows.tsx
@@ -419,16 +419,14 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
         new workflow.ExecuteWorkflowRequest({
           pushedRepoUrl: this.props.repoUrl,
           // Parse "var1=val1,var2=val2" string from the input field to Record<string, string>
-          env: this.state.runWorkflowEnv
-            .split(",")
-            .filter((n) => n.includes("="))
-            .map((n) => n.trim().split("="))
-            .filter((parts) => parts.length)
-            .map(([name, ...valueParts]) => [name, valueParts.join("=")])
-            .reduce((acc: Record<string, string>, [k, v]) => {
-              acc[k] = v;
-              return acc;
-            }, {}),
+          env: Object.fromEntries<string>(
+            this.state.runWorkflowEnv
+              .split(",")
+              .filter((n) => n.includes("="))
+              .map((n) => n.trim().split("="))
+              .filter((parts) => parts.length)
+              .map(([name, ...valueParts]) => [name, valueParts.join("=")])
+          ),
           actionNames: this.state.runWorkflowActionNames
             .split(",")
             .map((n) => n.trim())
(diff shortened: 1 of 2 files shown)


maggie-lou (approved):
This looks good to me as a basic approach. Supporting `Env: "key1=value1,key2=value2"` sounds okay to me as well for now

sluongng:
Added support for `env` as well. PTAL.

enterprise/app/workflows/workflows.tsx
```diff
@@ -370,6 +371,8 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
   state: RepoItemState = {
     isMenuOpen: false,
     showRunWorkflowInput: false,
+    runWorkflowActionNames: "",
+    runWorkflowEnvs: "",
```
maggie-lou, line 375: nit: `runWorkflowEnv`

enterprise/app/workflows/workflows.tsx
```diff
@@ -415,6 +418,21 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
       .executeWorkflow(
         new workflow.ExecuteWorkflowRequest({
           pushedRepoUrl: this.props.repoUrl,
+          // Parse "var=val" pairs from the input field to Record<string, string>
+          env: this.state.runWorkflowActionNames
```
maggie-lou, line 422: Looks like you accidentally swapped the variables - should be runWorkflowEnvs here and actionNames below

enterprise/app/workflows/workflows.tsx
```diff
@@ -415,6 +418,21 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
...
+            .split(",")
+            .filter((n) => n.includes("="))
```
maggie-lou, line 424: Rather than just filtering out invalid input, maybe we should return an error message and some additional validation that the input looks correct

enterprise/app/workflows/workflows.tsx
sluongng: fixed

enterprise/app/workflows/workflows.tsx
sluongng: oh great call. Fixed.

enterprise/app/workflows/workflows.tsx
sluongng: I was hoping to get away with a simple implementation first and add UI improvements to this later.

I expect the current `Env` field to be sort-lived, and that we should add a table input (2 columns) for users to add the environment variables in later. That would save us from this whole validation madness. Wdyt?

sluongng:
```diff
--- a/enterprise/app/workflows/workflows.tsx
+++ b/enterprise/app/workflows/workflows.tsx
@@ -414,21 +414,28 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
     this.setState({ runWorkflowActionStatuses: null, isWorkflowRunning: true });
     this.onCloseMenu();

+    // Parse "var1=val1,var2=val2" string from the input field to Record<string, string>
+    const env: Record<string, string> = {};
+    const envPairs = this.state.runWorkflowEnv.split(",")
+    if (!envPairs.reduce((acc, pair) => {
+      if (!acc) return false;
+      const halves = pair.split("=");
+      if (halves.length !== 2) return false;
+      // We just care about variable not being empty, value can be empty
+      if (halves[0].trim().length === 0) return false;
+
+      env[halves[0].trim()] = halves[1].trim();
+      return true;
+    }, true)) {
+      errorService.handleError("Invalid environment variable format. Please use 'VAR1=value1,VAR2=value2'.");
+      return;
+    }
+
     rpcService.service
       .executeWorkflow(
         new workflow.ExecuteWorkflowRequest({
           pushedRepoUrl: this.props.repoUrl,
-          // Parse "var1=val1,var2=val2" string from the input field to Record<string, string>
-          env: this.state.runWorkflowEnv
-            .split(",")
-            .filter((n) => n.includes("="))
-            .map((n) => n.trim().split("="))
-            .filter(([k, v]) => k && v)
-            .map(([k, v]) => [k.trim(), v.trim()])
-            .reduce((acc: Record<string, string>, [k, v]) => {
-              acc[k] = v;
-              return acc;
-            }, {}),
+          env: env,
           actionNames: this.state.runWorkflowActionNames
             .split(",")
             .map((n) => n.trim())
```
fwiw, here is what I cooked up locally which is really ugly 😆

enterprise/app/workflows/workflows.tsx
```diff
@@ -415,6 +418,21 @@ class RepoItem extends React.Component<RepoItemProps, RepoItemState> {
       .executeWorkflow(
         new workflow.ExecuteWorkflowRequest({
           pushedRepoUrl: this.props.repoUrl,
+          // Parse "var1=val1,var2=val2" string from the input field to Record<string, string>
+          env: this.state.runWorkflowEnv
+            .split(",")
+            .filter((n) => n.includes("="))
+            .map((n) => n.trim().split("="))
+            .filter(([k, v]) => k && v)
+            .map(([k, v]) => [k.trim(), v.trim()])
```
bduffany, line 427: nit: handle values with `=`

```suggestion
            .filter((parts) => parts.length)
            .map(([name, ...valueParts]) => [name, valueParts.join("=")])
```

bduffany: also `Object.fromEntries()` can help you form the env map from your `[k, v]` pairs

sluongng: Nice tip. Using fromEntries saved me from that .reduce 🙌

PR #7879, opened 2024-11-12
[RB] Support running a bash script from the CLI

maggie-lou:


maggie-lou:
Replaces https://github.com/buildbuddy-io/buildbuddy/pull/7766

bduffany (approved):

docs/remote-bazel-introduction.md
```diff
@@ -205,6 +207,29 @@ Remote Bazel flags, which go after `remote`.
 bb --verbose=1 remote build //...
 ```
 
+#### Running bash scripts
+
+To run arbitrary bash code on the remote runner, use the `--script` flag.
+
+```bash
+bb remote --script ls
```
bduffany, line 215: nit: I wrap the `ls` in quotes to reinforce that it accepts a single string argument, not a list of shell tokens

```suggestion
bb remote --script="ls"
```

bduffany: I think this will probably be a common source of confusion so another option is to make `--script` accept a path to a shell script

maggie-lou: Added an example for how to read the contents from the path to a shell script

PR #7897, opened 2024-11-14
Clarify docs for execution properties

vanja-p:
I think this was the intended meaning of the docs.


docs/rbe-platforms.md
```diff
@@ -150,7 +150,7 @@ These properties can be used in different ways:
 - Set `exec_properties` in the execution platform definition.
 - Set `exec_properties` in each BUILD target.
 - Set `--remote_default_exec_properties=KEY=VALUE` in `.bazelrc` or at the Bazel command line.
-  Note that these properties are not apply if you are already using a platform.
+  Note that these properties are not applied if you are already setting a platform on a target.
```
sluongng, line 153: Hmm I think `or a target` might make more sense here?

"Exec properties" could be set in 3(?) places, this flag, platform definition, and target definition.

sluongng (approved):

PR #7898, opened 2024-11-14
Record IO stats only for build_root device

bduffany:
Also show the IO stats in the UI - this was useful for E2E testing locally; figured it's a small enough change to include with this PR.

Here's the results from an sh_test that ran `head -c 10000000000 > /tmp/file.txt ; sync`

image

### DIFF
--- a/enterprise/server/remote_execution/block_io/block_io_unix.go
+++ b/enterprise/server/remote_execution/block_io/block_io_unix.go
@@ -37,8 +37,8 @@ func LookupDevice(path string) (*Device, error) {
 	if !ok {
 		return nil, err
 	}
-	major := unix.Major(stat.Rdev)
-	minor := unix.Minor(stat.Rdev)
+	major := unix.Major(uint64(stat.Rdev))
+	minor := unix.Minor(uint64(stat.Rdev))
 	blkDev := &Device{
 		Maj: int64(major),
 		Min: int64(minor),
--- a/enterprise/server/remote_execution/block_io/block_io_unix.go
+++ b/enterprise/server/remote_execution/block_io/block_io_unix.go
@@ -1,4 +1,4 @@
-//go:build unix
+//go:build (linux || darwin) && !android && !ios
 
 package block_io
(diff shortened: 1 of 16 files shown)


enterprise/server/remote_execution/cgroup/BUILD
```diff
@@ -5,11 +5,15 @@ go_library(
     srcs = ["cgroup.go"],
     importpath = "github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/cgroup",
     visibility = ["//visibility:public"],
-    deps = [
-        "//proto:remote_execution_go_proto",
-        "//server/util/log",
-        "//server/util/status",
-    ],
+    deps = select({
```
vanja-p, line 8: same thing about using compatible_with here

enterprise/server/remote_execution/block_io/block_io_unix.go
```diff
@@ -0,0 +1,107 @@
...
+		Maj: int64(major),
+		Min: int64(minor),
+	}
+
+	return blkDev, nil
+}
+
+// ParseMajMin parses major/minor block device numbers formatted like
+// "MAJ:MIN"
+func ParseMajMin(str string) (*Device, error) {
```
vanja-p, line 52: This should probably return two ints instead of Device. There's no guarantee that the string represents a real device.

enterprise/server/remote_execution/cgroup/BUILD
bduffany: gazelle added this diff after I added the `//go:build` directive. gopls complains without that directive, because it's trying to build on all platforms. I added `target_compatible_with` in addition, so that this target doesn't get built on windows

enterprise/server/remote_execution/block_io/block_io_unix.go
bduffany: good point - done

vanja-p (approved):

PR #7899, opened 2024-11-14
Add a proto for cgroup settings

bduffany:
Adds a proto representing all of the cgroup2 options that we can set per-task: https://docs.kernel.org/admin-guide/cgroup-v2.html

This will allow the app's task sizer to control resource limits in addition to task sizes. Executor implementation to follow.

### DIFF
--- a/proto/scheduler.proto
+++ b/proto/scheduler.proto
@@ -116,6 +116,140 @@ message TaskSize {
   repeated CustomResource custom_resources = 4;
 }
 
+// CgroupSettings defines Linux cgroup2 options for an execution.
+//
+// Where applicable, the value -1 means "unlimited", which translates to the
+// "max" string in cgroup2 terms.
+//
+// Note: for numeric fields, a corresponding field prefixed with "use_" is
+// defined, which indicates whether the executor should apply the value. If it's
+// false, the setting should not be applied, meaning the default value is used
+// instead. This convention is used instead of defining 0 to mean "do not set,"
+// because in some cases it's valid to set a value to 0. proto3 optional is not
+// used because of the awkward Go API.
+message CgroupSettings {
+  // Proportion of CPU given to this task relative to other tasks in the parent
+  // cgroup. This provides for a best-effort CPU guarantee.
+  //
+  // Values 1 to 10000 are supported.
+  //
+  // Maps to "cpu.weight" in cgroup2.
+  int64 cpu_weight = 1;
+  bool use_cpu_weight = 2;
+
+  // Maximum CPU usage allowed per quota period.
+  //
+  // Maps to the "cpu.max" quota field in cgroup2.
+  int64 cpu_quota_limit_usec = 3;
+  // How often the CPU quota is refreshed. Longer periods may allow for higher
+  // burst CPU usage but may result in more stalling if the quota is exhausted
+  // very early in the period.
+  //
+  // Maps to the "cpu.max" period field in cgroup2.
+  int64 cpu_quota_period_usec = 4;
+  bool use_cpu_quota = 5;
+
+  // CPU time that can be "borrowed" from other quota periods to allow for burst
+  // CPU usage, in usec.
+  //
+  // Maps to "cpu.max.burst" in cgroup2.
+  int64 cpu_max_burst_usec = 6;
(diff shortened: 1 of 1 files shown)


proto/scheduler.proto
```diff
@@ -116,6 +116,119 @@ message TaskSize {
...
+
+  // Maximum CPU usage allowed per quota period.
+  //
+  // Maps to the "cpu.max" quota field in cgroup2.
+  int64 cfs_quota_limit_usec = 3;
+  // How often the CPU quota is refreshed. Longer periods may allow for higher
+  // burst CPU usage but may result in more stalling if the quota is exhausted
+  // very early in the period.
+  //
+  // Maps to the "cpu.max" period field in cgroup2.
```
vanja-p, line 148: I don't think this is `cpu.max`. Maybe `cpu.period`?

vanja-p (approved):

proto/scheduler.proto
bduffany: In cgroup2, `cpu.max` gets set to a value containing 2 numbers, like `$MAX $PERIOD`

vanja-p: Ah, got it.

proto/scheduler.proto
```diff
@@ -116,6 +116,140 @@ message TaskSize {
...
+// used because of the awkward Go API.
+message CgroupSettings {
+  // Proportion of CPU given to this task relative to other tasks in the parent
+  // cgroup. This provides for a best-effort CPU guarantee.
+  //
+  // Values 1 to 10000 are supported.
+  //
+  // Maps to "cpu.weight" in cgroup2.
+  int64 cpu_weight = 1;
+  bool use_cpu_weight = 2;
```
tylerwilliams, line 138: lol proto3 sucks i guess?

it's a little hacky but you could just ignore stuff < 0 and remove all the bools in here. I think having them is gonna lead to fairly annoying code to generate this proto and read from it

bduffany: We need to distinguish between all of these cases:
(1) The field being unset. In this case, the executor should not set the cgroup value and leave it at the default. I think this behavior makes sense because of (a) old executor clients and (b) potential future additions to cgroup2
(2) Set to "max" value - we might want to explicitly set a value to be "unlimited." Unclear whether we'll want to do this in practice, but it feels like we're losing flexibility if we don't allow this to be represented.
(3) Explicitly set to 0
(4) Explicitly set to a value > 0

- Ignoring stuff < 0 doesn't allow for case (2) and also doesn't let us distinguish between (1) and (3).
- Using proto3 optional is fine but it means that the code will read like `if settings.CpuWeight != nil { doStuffWith(*settings.CpuWeight) }` I've gotten comments on PRs in the past that pointers to primitive types feel weird/unconventional because they require awkward code like this
 
An alternative would be to use a wrapper proto that's basically `oneof { max, int value, float value }` but this seems a bit much, idk.

vanja-p: Don't optional fields generate GetCpuWeights and HasCpuWeights methods?

bduffany: it only generates `Get*` not `Has*`:

the generated field is

```
	CpuWeight                *int64                        `protobuf:"varint,1,opt,name=cpu_weight,json=cpuWeight,proto3,oneof" json:"cpu_weight,omitempty"`
```

and only this method is generated:

```
func (x *CgroupSettings) GetCpuWeight() int64 {
	if x != nil && x.CpuWeight != nil {
		return *x.CpuWeight
	}
	return 0
}
```

bduffany: (I'm fine with proto3 optional + `nil` checks if you both are cool with it? since it lets us get rid of these `use_*` fields)

vanja-p: proto3 optional + nil checks seems good to me. It doesn't seem any harder to use than bool+field.

bduffany: I switched to a string representation per an offline discussion w/ Tyler. It allows representing unset, 0, non-zero, and "max" - the max value being represented as -1 with the optional approach has the downside of potentially storing it in the DB and messing up metrics.

PR #7900, opened 2024-11-14
Fix peer dependencies

tempoz:
This addresses all peer dependency warnings we get from yarn except for the `dagre-d3-react` ones, which would require a new release of that library to fix.

In doing so, it:

- Upgrades our typescript version

- Switches to using `useResizeDetector` from the deprecated `withResizeDetector`

- Switches to using `Range` instead of the removed `RangeWithKey` and `RangeKeyDict` instead of the removed `OnChangeProps`

- Removes code for handling the `days` field of a `Range`, as that field no longer exists

- Removes the `undefined` second positional parameter from `handleDefaultConfigClicked`, as that function no longer accepts a second positional parameter

- Removes fallback values when using unary `+` to convert a `Long` to a `number`, as that now always returns a `number`

- Changes our target in `tsconfig.json` from `es2016` to `es2018` to support the use of named capturing groups in `trace_events.ts`

- Removes the `resolutions` block from our `package.json`, as it is no longer necessary


enterprise/app/trends/heatmap.tsx
```diff
@@ -524,17 +518,24 @@ class HeatmapComponentInternal extends React.Component<HeatmapProps, State> {
   }
 
   render() {
-    const width = this.getWidth() - CHART_MARGINS.left - CHART_MARGINS.right;
-    const height = this.getHeight() - CHART_MARGINS.top - CHART_MARGINS.bottom;
+    const { width, ref } = useResizeDetector({
+      handleHeight: false,
+      refreshMode: "throttle",
+      refreshRate: 500,
+    });
```
bduffany, line 525: does this work? I thought React hooks (functions prefixed with `use`) only worked with functional components, and weren't supported for class components

tempoz: Then you're probably right; I couldn't test this locally and so was going to test it in dev when I got a free moment. How do you recommend we resolve this? The library just removed the `with` version, and I wasn't aware of the differing use in class and functional components until just now.

bduffany: Personally I'd ignore the deprecation warning

bduffany: (and revert to the old version - unless the issue is that this is setting off some security scanner?)

enterprise/app/filter/date_picker_button.tsx
```diff
@@ -54,17 +54,8 @@ export default class DatePickerButton extends React.Component<Props, State> {
...
-    const selection = (range as { selection: CustomDateRange }).selection;
-    if (selection.days) {
-      router.setQuery({
-        ...Object.fromEntries(this.props.search.entries()),
-        [START_DATE_PARAM_NAME]: "",
-        [END_DATE_PARAM_NAME]: "",
-        [LAST_N_DAYS_PARAM_NAME]: String(selection.days),
-      });
-      return;
-    }
```
jdhollen, line 67: i'm not sure we really want to remove this--iiuc, this changes "last 7 days" to a specific 7-day date range, so any bookmark set off of this setting will now be referring to a specific date range.

enterprise/app/tap/flakes.tsx
```diff
@@ -265,11 +265,11 @@ export default class FlakesComponent extends React.Component<Props, State> {
     if (!stats) {
       return "0%";
     }
-    const totalFlakes = (+stats.flakyRuns ?? 0) + (+stats.likelyFlakyRuns ?? 0);
+    const totalFlakes = +stats.flakyRuns + +stats.likelyFlakyRuns;
```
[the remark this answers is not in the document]
tempoz: I think it made it so that Longs can't fail to be converted to numbers?

enterprise/app/filter/date_picker_button.tsx
tempoz: good call; I got really confused by the type change here and misunderstood that we would still be getting a CustomDateRange out of `range.selection`. Fixed!

enterprise/app/tap/flakes.tsx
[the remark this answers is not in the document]
tempoz: also I think this is likely a result of upgrading typescript, not proto

enterprise/app/trends/heatmap.tsx
tempoz: I switched the `HeatmapComponent` to a functional component instead of a class component, and it appears to work based on my limited testing I did on my local machine with no real data.

jdhollen (approved):

PR #7901, opened 2024-11-14
Deprecate ambiguous SchedulingMetadata.task_size field

bduffany:
The `SchedulingMetadata.task_size` field is both lossy (since it is supposed to contain the "default" task size, but gets overwritten) and redundant (the executor allegedly uses this for scheduling but in practice it actually uses EnqueueTaskReservationRequest.task_size).

Deprecate this field and add clearer, not-lossy alternatives. (We'll keep setting it for some time, probably, because we do have some places in the code where it's unfortunately used - e.g. ExecutedActionMetadata.task_size gets set to this value, but probably should be getting set to the task_size from EnqueueTaskReservationRequest, which represents its "final" size for local scheduling

### DIFF
--- a/proto/scheduler.proto
+++ b/proto/scheduler.proto
@@ -118,27 +118,31 @@ message TaskSize {
 
 // Next ID: 9
 message SchedulingMetadata {
-  // Task size used for scheduling purposes, when the scheduler is deciding
-  // which executors (if any) may execute a task, and also when an executor is
-  // deciding which task to dequeue. Executors may see a different value of this
-  // field than what the scheduler sees, depending on measured_task_size or
-  // predicted_task_size. See documentation of those fields for more info.
-  TaskSize task_size = 1;
+  // DEPRECATED. This field has two different meanings:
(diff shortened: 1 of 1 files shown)


proto/scheduler.proto
```diff
@@ -206,7 +218,13 @@ message ReEnqueueTaskResponse {
 
 message EnqueueTaskReservationRequest {
   string task_id = 1;
-  TaskSize task_size = 2;
+
+  // The size of the task that determines how many resources the executor should
+  // allocate for this task.
+  //
+  // DEPRECATED: use scheduling_metadata.scheduled_size instead.
+  TaskSize task_size = 2 [deprecated = true];
```
vanja-p, line 226: The comment on `SchedulingMetadata.task_size` says to use this field, and `scheduled_size` doesn't exist.

bduffany: woops, I thought I reverted this change. will revert

vanja-p (approved):

vadimberezniker (approved):

PR #7904, opened 2024-11-15
Add IOPS/BPS to task size proto

bduffany:


proto/scheduler.proto
```diff
@@ -107,6 +107,18 @@ message TaskSize {
...
+  int64 estimated_disk_wbps = 5;
+
+  // The task's estimated disk bytes read per second.
+  int64 estimated_disk_rbps = 6;
+
+  // The task's estimated disk read IO operations per second.
+  int64 estimated_disk_riops = 7;
+
+  // The task's estimated disk write IO operations per second.
+  int64 estimated_disk_wiops = 8;
```
tylerwilliams, line 120: nit: all fields in here start with estimated_ which seems sort of redundant

tylerwilliams (approved):
LGTM with nits

proto/scheduler.proto
bduffany: yeah, it's a pre-existing issue, but I got rid of `estimated_` for now and we can rename the other fields later (should be wire-compatible)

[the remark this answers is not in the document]
bduffany: good call, done

PR #7905, opened 2024-11-15
Add disk usage stats to ClickHouse

bduffany:
This is needed so that if we add a default task size for disk usage, we can analyze historical executions to see what the default should be. Depends on https://github.com/buildbuddy-io/buildbuddy/pull/7898


tylerwilliams (approved):
LGTM

PR #7907, opened 2024-11-15
Schedule work more fairly on heterogeneous executors

tylerwilliams:
Distribution for the following nodes:
```
nodes = append(nodes, &testNode{1, "1", "h-1", 30_000})
nodes = append(nodes, &testNode{2, "2", "h-2", 30_000})
nodes = append(nodes, &testNode{3, "3", "h-3", 30_000})
nodes = append(nodes, &testNode{4, "4", "h-4", 44_000})
nodes = append(nodes, &testNode{5, "5", "h-5", 44_000})
nodes = append(nodes, &testNode{6, "6", "h-6", 44_000})
nodes = append(nodes, &testNode{7, "7", "h-7", 512_000})
nodes = append(nodes, &testNode{8, "8", "h-8", 512_000})
nodes = append(nodes, &testNode{9, "9", "h-9", 512_000})
```
Looks like:
```
executor: "1", numTasks: 12
executor: "2", numTasks: 15
executor: "3", numTasks: 19
executor: "4", numTasks: 27
executor: "5", numTasks: 18
executor: "6", numTasks: 24
executor: "7", numTasks: 285
executor: "8", numTasks: 283
executor: "9", numTasks: 317
```

### DIFF
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -1759,10 +1758,8 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 			// continue with for loop
 		}
 
-		// Pop the next highest ranked node off the list that has not
 		rankedNode := rankedNodes[0]
 		rankedNodes = rankedNodes[1:]
-
 		attempts++
 		if opts.maxAttempts > 0 && attempts > opts.maxAttempts {
 			return status.ResourceExhaustedErrorf("could not enqueue task reservation to executor")
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -642,8 +642,7 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode, 
 // To ensure fairness and diversity (mostly for tests with small number of
 // nodes):
 //   - the returned number of nodes is (maxSize / minSize) times greater than
-//
-// the original number of nodes
+//     the original number of nodes
 //   - all original nodes will be returned at least once
 func weightedResample(nodes []*executionNode) []*executionNode {
 	if len(nodes) == 0 {
@@ -658,6 +657,7 @@ func weightedResample(nodes []*executionNode) []*executionNode {
 		return iCPU < jCPU
 	})
 
+	unsampledOriginalNodes := make(map[*executionNode]struct{}, len(nodes))
 	cumulativeSum := make([]float64, len(nodes))
 	for i := 0; i < len(nodes); i++ {
 		cpu := float64(nodes[i].GetAssignableMilliCpu())
@@ -671,13 +671,22 @@ func weightedResample(nodes []*executionNode) []*executionNode {
 		if cpu < smallest {
 			smallest = cpu
 		}
+		unsampledOriginalNodes[nodes[i]] = struct{}{}
 	}
 
 	samples := make([]*executionNode, len(nodes)*int(largest/smallest))
 	for i := 0; i < len(samples); i++ {
 		val := rand.Float64() * cumulativeSum[len(cumulativeSum)-1]
 		n := sort.Search(len(cumulativeSum), func(i int) bool { return cumulativeSum[i] > val })
-		samples[i] = nodes[n]
+		sampledNode := nodes[n]
+		samples[i] = sampledNode
+		delete(unsampledOriginalNodes, sampledNode)
+	}
+	// Ensure that any of the original nodeset that was not sampled gets
+	// added. This ensures that all nodes are represented, so that node
+	// specific task routing ("route to this one node") still works.
+	for unsampledNode := range unsampledOriginalNodes {
+		samples = append(samples, unsampledNode)
 	}
 	for i := range samples {
 		j := rand.Intn(i + 1)
--- a/enterprise/server/scheduling/scheduler_server/scheduler_server.go
+++ b/enterprise/server/scheduling/scheduler_server/scheduler_server.go
@@ -1789,7 +1789,6 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 
 	attempts := 0
 	var rankedNodes []interfaces.RankedExecutionNode
-	usedNodes := make([]interfaces.RankedExecutionNode, 0, probeCount)
 
 	nonPreferredDelay := getNonPreferredSchedulingDelay(platform.GetProto(task.GetAction(), cmd))
 	delayable := enqueueRequest.GetDelay() == nil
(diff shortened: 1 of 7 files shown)


enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -1743,6 +1797,7 @@ func (s *SchedulerServer) enqueueTaskReservations(ctx context.Context, enqueueRe
 			if len(candidateNodes) == 0 {
 				return errTaskSizeTooLarge(pool, os, arch, enqueueRequest.GetTaskSize())
 			}
+			candidateNodes = weightedResample(candidateNodes)
```
bduffany, line 1800: `task_router.RankNodes()` does a random shuffle on the non-preferred nodes, so it will undo the work done here - probably the task_router needs to be in charge of this?

tylerwilliams: I don't think the order matters in this case, it's just the population of nodes that I'm after. LMK if you still think this is wrong.

tylerwilliams:
Added a fix that ensures all original nodes are always present, so that task-specific routing is not broken.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -634,6 +635,66 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode,
...
+	for i := 0; i < len(samples); i++ {
+		val := rand.Float64() * cumulativeSum[len(cumulativeSum)-1]
+		n := sort.Search(len(cumulativeSum), func(i int) bool { return cumulativeSum[i] > val })
+		sampledNode := nodes[n]
+		samples[i] = sampledNode
+		delete(unsampledOriginalNodes, sampledNode)
+	}
+	// Ensure that any of the original nodeset that was not sampled gets
+	// added. This ensures that all nodes are represented, so that node
+	// specific task routing ("route to this one node") still works.
```
bduffany, line 687: this part will mess with the relative frequencies and therefore the weighting, right? (this wouldn't be a concern if we did this weightedResample() in task_router, I think? i.e. task router could do the rand.Shuffle() only on the nodes that showed up in the randomly sampled list above, then append the randomly shuffled nodes that we missed in the sampling process)

bduffany: in other words I think it'd be better (more accurate and also clearer) to change `rand.Shuffle()` in task_router to a function like `weightedShuffle()` which is essentially this function plus deduping

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -634,6 +635,66 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode,
...
+	}
+	largest := float64(0)
+	smallest := float64(1e6)
+
+	sort.Slice(nodes, func(i, j int) bool {
+		iCPU := nodes[i].GetAssignableMilliCpu()
+		jCPU := nodes[j].GetAssignableMilliCpu()
+		return iCPU < jCPU
+	})
+
```
bduffany, line 659: is this sorting step needed? the loop just below is getting the min/max, and the sampling doesn't seem to depend on the original nodes being sorted (only the cumulative sums being correct)

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -634,6 +635,66 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode,
...
+// assignable CPU. This means that larger nodes have an opportunity to appear
+// more often than smaller nodes, increasing their chance of being hit.
+//
+// To ensure fairness and diversity (mostly for tests with small number of
+// nodes):
+//   - the returned number of nodes is (maxSize / minSize) times greater than
+//     the original number of nodes
+//   - all original nodes will be returned at least once
+func weightedResample(nodes []*executionNode) []*executionNode {
+	if len(nodes) == 0 {
```
iain-macdonald, line 648: Probably fine to make this a "<= 1"

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -634,6 +635,66 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode,
...
+		val := rand.Float64() * cumulativeSum[len(cumulativeSum)-1]
+		n := sort.Search(len(cumulativeSum), func(i int) bool { return cumulativeSum[i] > val })
+		sampledNode := nodes[n]
+		samples[i] = sampledNode
+		delete(unsampledOriginalNodes, sampledNode)
+	}
+	// Ensure that any of the original nodeset that was not sampled gets
+	// added. This ensures that all nodes are represented, so that node
+	// specific task routing ("route to this one node") still works.
+	for unsampledNode := range unsampledOriginalNodes {
```
iain-macdonald, line 688: why not just seed the output slice with every input node and then append from the weighted set after? then i think you can avoid the `unsampledOriginalNodes`

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -634,6 +635,66 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode,
...
+		samples = append(samples, unsampledNode)
+	}
+	for i := range samples {
```
iain-macdonald, line 691: rand.Shuffle()?

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -634,6 +635,66 @@ func (np *nodePool) fetchExecutionNodes(ctx context.Context) ([]*executionNode,
...
+// To ensure fairness and diversity (mostly for tests with small number of
+// nodes):
+//   - the returned number of nodes is (maxSize / minSize) times greater than
+//     the original number of nodes
+//   - all original nodes will be returned at least once
+func weightedResample(nodes []*executionNode) []*executionNode {
+	if len(nodes) == 0 {
+		return nodes
+	}
+	largest := float64(0)
```
iain-macdonald, line 651: probably fine as-is but you could just use nodes[i].GetAssignableMilliCpu() to avoid any weird edge cases here

iain-macdonald (approved):

bduffany (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
tylerwilliams: Done. I like this better too!

enterprise/server/scheduling/scheduler_server/scheduler_server.go
tylerwilliams: removed.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
tylerwilliams: Done.

enterprise/server/scheduling/scheduler_server/scheduler_server.go
tylerwilliams: The code currently only adds back original-set nodes if they were not sampled, which I think gives a more accurate representation than seeding with that list. (Basically, if they are sampled we don't need to re-add them, whereas with the method you suggest, we'd add them twice I think?)

enterprise/server/scheduling/task_router/task_router.go
```diff
@@ -124,6 +176,10 @@ func (tr *taskRouter) RankNodes(ctx context.Context, action *repb.Action, cmd *r
 		return nonePreferred(nodes)
 	}
 
+	// Resample nodes by CPU weight so that nodes with more resources
+	// are more likely to get more tasks.
+	nodes = weightedResample(nodes)
```
bduffany, line 181: I think in the `nonePreferred` cases above, we still want to do a `weightedResample`, no? why not do the weightedResample on line 108 and keep the existing `nonePreferred` code that directly returns the nodes?

PR #7908, opened 2024-11-15
[RB] Reset flags between tests

maggie-lou:


enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
```diff
@@ -186,6 +198,10 @@ func TestWithPublicRepo(t *testing.T) {
 	})
 	require.NoError(t, err)
 	require.Contains(t, string(logResp.GetBuffer()), "Usage: bazel <command> <options>")
+
+	t.Cleanup(func() {
+		resetFlags(t)
+	})
```
bduffany, line 204: I think this cleanup should go at the top of each test so that it runs even if the `require` calls fail above?

bduffany (approved):
LGTM but would be good to get Zoey's thoughts on the flag lib changes

enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
maggie-lou: Ack - fixed

tempoz:
I made `flags.Set` for this purpose: https://github.com/buildbuddy-io/buildbuddy/blob/master/server/util/testing/flags/flags.go#L32-L42

`authdb_test` has a lot of examples of different usages of it: https://github.com/buildbuddy-io/buildbuddy/blob/master/enterprise/server/backends/authdb/authdb_test.go

The primary difference is that it registers the cleanup on its own and that you pass the actual value you want as opposed to the string representation of the value.

This should help simplify the PR ^-^

PR #7914, opened 2024-11-18
Add function to apply cgroup settings

bduffany:
The plan is to use this function to configure the cgroups for firecracker+ociruntime.

### DIFF
--- a/enterprise/server/remote_execution/cgroup/cgroup.go
+++ b/enterprise/server/remote_execution/cgroup/cgroup.go
@@ -32,19 +32,13 @@ const (
 	cidPlaceholder = "{{.ContainerID}}"
 )
 
-// min/max weight values for cgroup2 weight-based control
-const (
-	minWeight = 1
-	maxWeight = 10_000
-)
-
 // Setup configures the cgroup at the given path with the given settings.
 // Any settings for cgroup controllers that aren't enabled are ignored.
 // IO limits are applied to the given block device, if specified.
 func Setup(ctx context.Context, path string, s *scpb.CgroupSettings, blockDevice *block_io.Device) error {
-	m, err := settingsMap(s, blockDevice)
-	if err != nil {
-		return err
+	m := settingsMap(s, blockDevice)
+	if len(m) == 0 {
+		return nil
 	}
 	enabledControllers, err := ParentEnabledControllers(path)
 	if err != nil {
@@ -58,7 +52,7 @@ func Setup(ctx context.Context, path string, s *scpb.CgroupSettings, blockDevice
 		}
 		settingFilePath := filepath.Join(path, name)
 		if err := os.WriteFile(settingFilePath, []byte(value), 0); err != nil {
-			return err
+			return fmt.Errorf("write %q to cgroup file %q: %w", value, name, err)
 		}
 	}
 	return nil
@@ -79,19 +73,21 @@ func ParentEnabledControllers(path string) (map[string]bool, error) {
 	return enabled, nil
 }
 
-func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) (map[string]string, error) {
+func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) map[string]string {
 	m := map[string]string{}
 	if s == nil {
-		return m, nil
+		return m
 	}
 	if s.CpuWeight != nil {
-		if s.GetCpuWeight() < minWeight {
-			return nil, fmt.Errorf("invalid cpu weight %d: must be in range [1, 10000]", s.GetCpuWeight())
-		}
 		m["cpu.weight"] = strconv.Itoa(int(s.GetCpuWeight()))
 	}
-	if s.CpuQuotaPeriodUsec != nil {
-		m["cpu.max"] = fmt.Sprintf("%d %d", s.GetCpuQuotaLimitUsec(), s.GetCpuQuotaPeriodUsec())
+	if s.CpuQuotaLimitUsec != nil {
+		if s.CpuQuotaPeriodUsec == nil {
+			// Keep current or default period (100ms) but update quota.
+			m["cpu.max"] = strconv.Itoa(int(s.GetCpuQuotaPeriodUsec()))
+		} else {
+			m["cpu.max"] = fmt.Sprintf("%d %d", s.GetCpuQuotaLimitUsec(), s.GetCpuQuotaPeriodUsec())
+		}
 	}
 	if s.CpuMaxBurstUsec != nil {
 		m["cpu.max.burst"] = strconv.Itoa(int(s.GetCpuMaxBurstUsec()))
@@ -128,9 +124,6 @@ func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) (map[stri
 			m["io.latency"] = fmt.Sprintf("%d:%d target=%d", blockDevice.Maj, blockDevice.Min, s.GetBlockIoLatencyTargetUsec())
 		}
 		if s.BlockIoWeight != nil {
-			if s.GetBlockIoWeight() < minWeight {
-				return nil, fmt.Errorf("invalid cpu weight %d: must be in range [1, 10000]", s.GetCpuWeight())
-			}
 			m["io.weight"] = fmt.Sprintf("%d:%d %d", blockDevice.Maj, blockDevice.Min, s.GetBlockIoWeight())
 		}
 		var limitFields []string
@@ -153,7 +146,7 @@ func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) (map[stri
 			m["io.max"] = fmt.Sprintf("%d:%d %s", blockDevice.Maj, blockDevice.Min, strings.Join(limitFields, " "))
 		}
 	}
-	return m, nil
+	return m
 }
 
 func fmtPercent(v float32) string {
--- a/enterprise/server/remote_execution/cgroup/cgroup.go
+++ b/enterprise/server/remote_execution/cgroup/cgroup.go
@@ -36,7 +36,10 @@ const (
 // Any settings for cgroup controllers that aren't enabled are ignored.
 // IO limits are applied to the given block device, if specified.
 func Setup(ctx context.Context, path string, s *scpb.CgroupSettings, blockDevice *block_io.Device) error {
-	m := settingsMap(s, blockDevice)
+	m, err := settingsMap(s, blockDevice)
+	if err != nil {
+		return err
+	}
 	if len(m) == 0 {
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,105 @@ const (
...
+// IO limits are applied to the given block device, if specified.
+func Setup(ctx context.Context, path string, s *scpb.CgroupSettings, blockDevice *block_io.Device) error {
+	m, err := settingsMap(s, blockDevice)
+	if err != nil {
+		return err
+	}
+	for name, value := range m {
+		settingFilePath := filepath.Join(path, name)
+		if err := os.WriteFile(settingFilePath, []byte(value), 0); err != nil {
+			if os.IsNotExist(err) {
```
vanja-p, line 51: we want to quietly fail if the cgroup doesn't exist?

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,105 @@ const (
...
+			}
+		}
+	}
+	return nil
+}
+
+func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) (map[string]string, error) {
+	m := map[string]string{}
+	if s.CpuWeight != nil {
+		m["cpu.weight"] = fmt.Sprintf("%d", s.GetCpuWeight())
```
vanja-p, line 64: i think you want to clampWeight here as well

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,105 @@ const (
...
+	}
+	if s.CpuQuotaPeriodUsec != nil {
+		m["cpu.max"] = fmt.Sprintf("%d %s", s.GetCpuQuotaLimitUsec(), strconv.Itoa(int(s.GetCpuQuotaPeriodUsec())))
```
vanja-p, line 67: why formatting as int for one field and string for the other?

Also, do you want to clamp these values to their ranges?

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,105 @@ const (
...
+				log.CtxWarningf(ctx, "Failed to set up cgroup file %q: %s", settingFilePath, err)
+			} else {
+				return fmt.Errorf("write %q: %w", settingFilePath, err)
+			}
+		}
+	}
+	return nil
+}
+
+func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) (map[string]string, error) {
```
vanja-p, line 61: this never returns an error

enterprise/server/remote_execution/cgroup/cgroup_test.go
```diff
@@ -4,13 +4,83 @@ import (
...
+		{
+			name:     "nil",
+			settings: nil,
+			expected: map[string]string{},
+		},
+		{
+			name:     "all fields unset",
+			settings: nil,
+			expected: map[string]string{},
+		},
```
vanja-p, line 32: these 2 tests look the same

enterprise/server/remote_execution/cgroup/cgroup_test.go
```diff
@@ -4,13 +4,83 @@ import (
...
+				"memory.high":      "777000000",
+				"memory.max":       "800000000",
+				"memory.low":       "100000000",
+				"memory.min":       "50000000",
+				"memory.swap.high": "800000000",
+				"memory.swap.max":  "1000000000",
+				"io.latency":       "279:8 target=100000",
+				"io.weight":        "279:8 300",
+				"io.max":           "279:8 riops=1000 wiops=500 rbps=4096000 wbps=1024000",
+			},
```
vanja-p, line 72: can you add a test to verify that clamping works?

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,105 @@ const (
...
+			}
+		}
+		if len(limitFields) > 0 {
+			m["io.max"] = fmt.Sprintf("%d:%d %s", blockDevice.Maj, blockDevice.Min, strings.Join(limitFields, " "))
+		}
+	}
+	return m, nil
+}
+
+func clampWeight(v int64) int64 {
```
vanja-p, line 126: I'm thinking that it might be better to fail requests with invalid values instead of clamping them. That will make it faster to find bugs in the execution and scheduler servers.

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: The intent was to avoid failing for on-prem setups that might not have certain cgroup controllers enabled, but in my testing we wind up seeing EPERM in this case instead of ENOENT. I changed this logic to explicitly check whether the controller is enabled and just log a warning for now if not.

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: it returns an error now that we're validating weights instead of clamping

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: validating the weight now instead

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: oops, fixed. I didn't see a min/max for CPU quota in docs?

enterprise/server/remote_execution/cgroup/cgroup.go
bduffany: agreed, done

enterprise/server/remote_execution/cgroup/cgroup_test.go
bduffany: woops, meant to use an empty proto here.

enterprise/server/remote_execution/cgroup/cgroup_test.go
bduffany: added a test to make sure we return an error instead, since we're not clamping anymore

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,134 @@ const (
...
+	return enabled, nil
+}
+
+func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) (map[string]string, error) {
+	m := map[string]string{}
+	if s == nil {
+		return m, nil
+	}
+	if s.CpuWeight != nil {
+		if s.GetCpuWeight() < minWeight {
```
vanja-p, line 88: also check the maxWeight

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,134 @@ const (
...
+			return nil, fmt.Errorf("invalid cpu weight %d: must be in range [1, 10000]", s.GetCpuWeight())
+		}
+		m["cpu.weight"] = strconv.Itoa(int(s.GetCpuWeight()))
+	}
+	if s.CpuQuotaPeriodUsec != nil {
+		m["cpu.max"] = fmt.Sprintf("%d %d", s.GetCpuQuotaLimitUsec(), s.GetCpuQuotaPeriodUsec())
```
vanja-p, line 94: for cgroups1, quota period must be between 1000 microseconds and 1 second (https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/resource_management_guide/sec-cpu#sec-cpu). I assume it's the same for cgroups2. I guess the limit just needs to be non-negative, or maybe bigger than 0.

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,134 @@ const (
...
+		}
+		if len(limitFields) > 0 {
+			m["io.max"] = fmt.Sprintf("%d:%d %s", blockDevice.Maj, blockDevice.Min, strings.Join(limitFields, " "))
+		}
+	}
+	return m, nil
+}
+
+func fmtPercent(v float32) string {
+	return fmt.Sprintf("%.2f", v)
```
vanja-p, line 160: I think all the percentages needs to be between 0 and 100

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,134 @@ const (
...
+	}
+	if s.SwapLimitBytes != nil {
+		m["memory.swap.max"] = strconv.Itoa(int(s.GetSwapLimitBytes()))
+	}
+	if blockDevice != nil {
+		if s.BlockIoLatencyTargetUsec != nil {
+			m["io.latency"] = fmt.Sprintf("%d:%d target=%d", blockDevice.Maj, blockDevice.Min, s.GetBlockIoLatencyTargetUsec())
+		}
+		if s.BlockIoWeight != nil {
+			if s.GetBlockIoWeight() < minWeight {
```
vanja-p, line 131: check the max weight as well

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,134 @@ const (
...
+		return fmt.Errorf("read enabled controllers: %w", err)
+	}
+	for name, value := range m {
+		controller, _, _ := strings.Cut(name, ".")
+		if !enabledControllers[controller] {
+			log.CtxWarningf(ctx, "Skipping cgroup %q setting for disabled cgroup controller %q", name, controller)
+			continue
+		}
+		settingFilePath := filepath.Join(path, name)
+		if err := os.WriteFile(settingFilePath, []byte(value), 0); err != nil {
```
vanja-p, line 60: What happens with os.WriteFile if we try to write invalid values? Does it return an error or ignore the write? If it returns an error, maybe we don't need our own error checking.

If we need to do our own error checking, maybe we should check that all values are non-negative.

enterprise/server/remote_execution/cgroup/cgroup.go
vanja-p, line 94: Do you want to support changing just the period or just the limit? Right now, they're tied together in a weird way:
- If you set just CpuQuotaPeriodUsec, then limit gets set to 0.
- If you set just CpuQuotaLimitUsec, it's ignored.

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,127 @@ const (
...
+	if s == nil {
+		return m
+	}
+	if s.CpuWeight != nil {
+		m["cpu.weight"] = strconv.Itoa(int(s.GetCpuWeight()))
+	}
+	if s.CpuQuotaLimitUsec != nil {
+		if s.CpuQuotaPeriodUsec == nil {
+			// Keep current or default period (100ms) but update quota.
+			m["cpu.max"] = strconv.Itoa(int(s.GetCpuQuotaPeriodUsec()))
```
vanja-p, line 87: i think you meant to write the limit here instead of the period

enterprise/server/remote_execution/cgroup/cgroup_test.go
```diff
@@ -4,13 +4,84 @@ import (
...
+		},
```
vanja-p, line 75: add a test that return EINVAL

enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,127 @@ const (
...
+
+func settingsMap(s *scpb.CgroupSettings, blockDevice *block_io.Device) map[string]string {
+	m := map[string]string{}
+	if s == nil {
+		return m
+	}
+	if s.CpuWeight != nil {
+		m["cpu.weight"] = strconv.Itoa(int(s.GetCpuWeight()))
+	}
+	if s.CpuQuotaLimitUsec != nil {
```
vanja-p, line 84: If just the period is set, this silently ignores it. Maybe we should return an error.

vanja-p (approved):

PR #7916, opened 2024-11-19
Apply cgroup settings from scheduler to firecracker and oci containers

bduffany:
Tested E2E by updating execution_server to hard-code `SchedulingMetadata.CgroupSettings`, and verified that they are applied both for firecracker VMs and OCI containers.


enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1517,12 +1546,11 @@ func (c *FirecrackerContainer) getJailerConfig(kernelImagePath string) (*fcclien
 		Stdout:         c.vmLogWriter(),
 		Stderr:         c.vmLogWriter(),
 		CgroupVersion:  cgroupVersion,
-		CgroupArgs:     cgroupArgs,
 		// Use the root cgroup as the cgroup parent rather than the default
```
vanja-p, line 1549: This comment is a bit stale now

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1517,12 +1546,11 @@ func (c *FirecrackerContainer) getJailerConfig(kernelImagePath string) (*fcclien
...
 		// "firecracker" subdir, so that VM cgroups are siblings of OCI
 		// container cgroups. This is needed because CPU weight is applied to
 		// sibling cgroups in a hierarchy.
-		ParentCgroup: fcclient.String(""),
+		ParentCgroup: fcclient.String(c.cgroupParent),
```
vanja-p, line 1553: I think you want to use c.cgroupPath() here. Otherwise that cgroup is never used.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -814,8 +856,7 @@ func (c *ociContainer) createSpec(ctx context.Context, cmd *repb.Command) (*spec
 			"org.opencontainers.image.stopSignal": unix.SIGTERM.String(),
 		},
 		Linux: &specs.Linux{
-			// TODO: set up cgroups
-			CgroupsPath: "",
+			CgroupsPath: filepath.Join(c.cgroupRootRelativePath()),
```
vanja-p, line 859: why the filepath.Join?

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: updated the comment - to get the effective cgroup path, the jailer prepends /sys/fs/cgroup and appends the VM ID.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: removed

vanja-p (approved):

PR #7921, opened 2024-11-19
Ensure action cgroups are placed under the k8s container cgroup

bduffany:
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/4140

### DIFF
--- a/enterprise/server/cmd/executor/executor_linux.go
+++ b/enterprise/server/cmd/executor/executor_linux.go
@@ -4,13 +4,42 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 
+	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/cgroup"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/vbd"
+	"github.com/buildbuddy-io/buildbuddy/server/util/flag"
 	"github.com/buildbuddy-io/buildbuddy/server/util/log"
 	"github.com/buildbuddy-io/buildbuddy/server/util/networking"
 )
+
+var (
+	inheritCgroup = flag.Bool("executor.inherit_cgroup", false, "Ensure that action cgroups inherit from the executor's starting cgroup. This flag will be removed in the future.", flag.Internal)
+)
+
+// getActionsCgroupParent returns the parent cgroup under which all action child
+// cgroups should be created. The returned cgroup path is relative to the cgroup
+// FS root path.
+func getActionsCgroupParent() (string, error) {
+	if !*inheritCgroup {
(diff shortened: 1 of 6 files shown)


enterprise/server/remote_execution/cgroup/cgroup.go
```diff
@@ -31,6 +32,153 @@ const (
...
+	if controllers := parts[1]; controllers != "" {
+		return "", ErrV1NotSupported
+	}
+	return strings.Join(parts[2:], ":"), nil
+}
+
+// Setup configures the cgroup at the given path with the given settings.
+// Any settings for cgroup controllers that aren't enabled are ignored.
+// IO limits are applied to the given block device, if specified.
+func Setup(ctx context.Context, path string, s *scpb.CgroupSettings, blockDevice *block_io.Device) error {
```
vanja-p, line 64: i think this diff is from your previous PR

bduffany: woops, fixed

enterprise/server/cmd/executor/executor_linux.go
```diff
@@ -4,14 +4,43 @@ package main
...
+		if errors.Is(err, cgroup.ErrV1NotSupported) {
+			log.Warningf("Note: executor is running under cgroup v1, which has limited support. Some functionality may not work as expected.")
+			return "", nil
+		}
+	}
+	// Return the executor cgroup directly, for now.
+	// TODO(https://github.com/buildbuddy-io/buildbuddy-internal/issues/4110):
+	// restructure cgroups so that all actions are run in a single cgroup and
+	// the executor runs as a sibling of that.
+	return strings.TrimPrefix(relpath, string(os.PathSeparator)), nil
```
vanja-p, line 41: why not remove the `/` in cgroup.GetCurrent? That would make it easier to use.

vanja-p (approved):

enterprise/server/cmd/executor/executor_linux.go
bduffany: no good reason - done

vadimberezniker (approved):
Fine with me as long as it doesn't interfere with us assigning CPUs to the executor cgroup, whether it's through k8s or through us messing with the cgroup directly.

bduffany:
> Fine with me as long as it doesn't interfere with us assigning CPUs to the executor cgroup, whether it's through k8s or through us messing with the cgroup directly.

Discussed offline - this PR will make it so we can no longer control the executor cgroup independently from the action cgroups, but I will send a follow-up PR to properly structure the cgroups so that actions run in a sibling cgroup of the executor.

tylerwilliams (approved):

PR #7924, opened 2024-11-20
Fix filecache scan deleting entries added by concurrent actions

bduffany:
During the initial filecache scan, we repeatedly call `AddFile()` on each scanned file.

The current implementation of `AddFile()` is:
1. Evict any existing entry. If an entry does exist, unlink it from the destination path, via the OnEvict callback.
2. Hardlink the file from the source path to its dest path.

For the filecache scan specifically, this logic is not ideal for two reasons:
* At best, it's unnecessary. Passing the same argument to `os.Link()` twice is always an error: either `EEXIST` (file already exists) or `ENOENT` (no such file or directory). So we're performing a bunch of syscalls which are guaranteed to return errors, which we then ignore. This is a bunch of unnecessary work.
* At worst, an action can race with the filecache background scan to update a cache key (e.g. during input download or output upload). If the scan calls `AddFile()` on a file it discovered, but an action was running concurrently and already called `AddFile()` for the same key, then step (1) above will invoke the OnEvict callback, which deletes the cache entry and unlinks the existing file. Because we're linking the path from itself, we then fail to add the file.

### DIFF
--- a/enterprise/server/remote_execution/filecache/filecache.go
+++ b/enterprise/server/remote_execution/filecache/filecache.go
@@ -166,9 +166,7 @@ func (c *fileCache) nodeFromPathAndSize(fullPath string, sizeBytes int64) (strin
 func (c *fileCache) scanDir() {
 	dirCount := 0
 	fileCount := 0
-	errCount := 0
 	scanStart := time.Now()
-	var addErr error
 	walkFn := func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -298,15 +288,27 @@ func (c *fileCache) addFileToGroup(groupID string, node *repb.FileNode, existing
...
+		c.l.Remove(k)
+		if err := disk.EnsureDirectoryExists(filepath.Dir(fp)); err != nil {
+			return err
+		}
+		if err := cloneOrLink(groupID, existingFilePath, fp); err != nil {
+			return err
+		}
+	} else if c.l.Contains(k) {
+		return nil
 	}
```
tylerwilliams, line 310: this logic is a little confusing to me -- would it be clearer if we just enumerated the special case and unindented the rest?
```
if fp == existingFilePath && c.l.Contains(k) {
  return nil // skip adding the same file twice which could cause an eviction
}
c.l.Remove(k)
if err := disk.EnsureDirectoryExists(filepath.Dir(fp)); err != nil {
	return err
}
if err := cloneOrLink(groupID, existingFilePath, fp); err != nil {
	return err
}
```

tylerwilliams (approved):

enterprise/server/remote_execution/filecache/filecache.go
bduffany: Done, but guarded the `Remove()`, `EnsureDirectoryExists()`, and `cloneOrLink()` calls with `if fp != existingFilePath { }`

PR #7928, opened 2024-11-20
Update dev QA script to remove podman dependency in dev

bduffany:
We're testing the bb repo on an old commit that still runs the ci_runner_test with `workload-isolation-type=podman` against dev.

### DIFF
--- a/tools/dev_qa.py
+++ b/tools/dev_qa.py
@@ -33,7 +33,7 @@ REPO_CONFIGS = [
     {
         "name": "buildbuddy",
         "repo_url": "https://github.com/buildbuddy-io/buildbuddy",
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):

tools/dev_qa.py
```diff
@@ -33,7 +33,7 @@
     {
         "name": "buildbuddy",
         "repo_url": "https://github.com/buildbuddy-io/buildbuddy",
-        "commit_sha": "70416ce9f040ba487579dfc773b8f426396bca48",
```
maggie-lou, line 36: Reminder to update here too: https://github.com/buildbuddy-io/buildbuddy-internal/blob/master/.github/workflows/dev-qa.yaml#L37

PR #7929, opened 2024-11-20
Add executor flag to create child cgroups for executor and actions

bduffany:
Add a flag to allow the executor to create child cgroups for the executor and action processes.

This should fix several issues:
- The current cpu weight for action processes is being applied relative to cgroup siblings in the cgroup root on the host node. This is not desired - CPU resources should be distributed _within_ the executor container cgroup, so that we respect quota/limits etc. and ensure that we don't mess with the relative weights between other pods on the node.
- Placing the executor within a sibling cgroup of action cgroups allows us to assign CPU weight to the executor process itself, which gets weighted against all action cgroups.
- k8s pod metrics should now work properly (assuming those metrics are driven by scraping the cgroup metrics from the container cgroup)
- On k8s we're currently creating action cgroups in the host node cgroup namespace, which means we're leaving garbage behind that accumulates until the node is restarted. This should be fixed as well, since k8s will tear down the executor container cgroup for us, which now includes all of the per-action cgroups.

Tested locally by running the executor in a cgroup that would simulate a k8s container cgroup.

The idle state looks like this:

```
/sys/fs/cgroup/executor-container-test/buildbuddy.executor/cgroup.procs:
- 1873893: executor (./bazel-bin/enterprise/server/cmd/executor/executo...)
/sys/fs/cgroup/executor-container-test/buildbuddy.executor.tasks/cgroup.procs:
/sys/fs/cgroup/executor-container-test/cgroup.procs:
```

When running a `sleep 5` action, it looks like this:

```
/sys/fs/cgroup/executor-container-test/buildbuddy.executor/cgroup.procs:
- 1873893: executor (./bazel-bin/enterprise/server/cmd/executor/executo...)
- 1891167: 3 (crun --log-format=json --cgroup-manager=cgroupfs r...)
/sys/fs/cgroup/executor-container-test/buildbuddy.executor.tasks/cgroup.procs:
/sys/fs/cgroup/executor-container-test/buildbuddy.executor.tasks/d2947fe257999362e84f96eb4e9e67349116e3fbd1c3e0b983b95c4e4946defe/cgroup.procs:
- 1891169: sleep (sleep 5 ...)
/sys/fs/cgroup/executor-container-test/cgroup.procs:
```

Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/4112
Partially addresses https://github.com/buildbuddy-io/buildbuddy-internal/issues/4110 since we are not configuring these cgroups yet.

### DIFF
--- a/enterprise/server/cmd/executor/executor_linux.go
+++ b/enterprise/server/cmd/executor/executor_linux.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
+	"strconv"
 
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/cgroup"
 	"github.com/buildbuddy-io/buildbuddy/enterprise/server/remote_execution/vbd"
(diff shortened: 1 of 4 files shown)


enterprise/server/cmd/executor/executor_linux.go
```diff
@@ -16,29 +18,63 @@ import (
...
+const (
+	// cgroup name for the executor process as well as any child processes that
+	// aren't explicitly run in separate cgroups (e.g. executable tools).
+	executorCgroupName = "buildbuddy.executor"
+
+	// cgroup name for the parent cgroup for action executions.
+	taskCgroupName = "buildbuddy.executor.tasks"
+)
+
+// setupCgroups moves the executor process to its own child cgroup, and sets up
```
vanja-p, line 33: document what this returns

vanja-p (approved):

enterprise/server/cmd/executor/executor_linux.go
bduffany: Done

PR #7930, opened 2024-11-20
Don't copy the workspace VBD contents during workspace sync

bduffany:
Since the workspace VBD now exposes the workspace block device contents as a plain old file (via FUSE), we can directly pass that file path to `ImageToDirectory()`, rather than first creating a full copy of the workspace contents. (This resolves the TODO here - this wasn't easily doable with NBD before, but is now possible with VBD)

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -1653,16 +1653,12 @@ func (c *FirecrackerContainer) copyOutputsToWorkspace(ctx context.Context) error
 
 	workspaceExt4Path := filepath.Join(c.getChroot(), workspaceFSName)
 
-	if *enableVBD {
-		// Reassemble the workspace chunks into a single file.
-		// TODO(bduffany): figure out how to avoid doing this work, e.g. by
-		// mounting the workspace disk as a local NBD on the host.
-		// For now, this approach is acceptable because firecracker workspaces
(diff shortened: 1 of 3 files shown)


vadimberezniker (approved):

tylerwilliams (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1653,16 +1653,12 @@ func (c *FirecrackerContainer) copyOutputsToWorkspace(ctx context.Context) error
...
 	workspaceExt4Path := filepath.Join(c.getChroot(), workspaceFSName)
 
-	if *enableVBD {
-		// Reassemble the workspace chunks into a single file.
-		// TODO(bduffany): figure out how to avoid doing this work, e.g. by
-		// mounting the workspace disk as a local NBD on the host.
-		// For now, this approach is acceptable because firecracker workspaces
-		// are typically small or empty (e.g. workflows run in $HOME rather than
-		// the workspace dir).
-		if err := c.workspaceStore.WriteFile(workspaceExt4Path); err != nil {
```
maggie-lou, line 1663: nit: We can delete this method now right?

maggie-lou (approved):
LGTM - one small comment

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: nice catch, done

PR #7931, opened 2024-11-20
Store filecache entries in subdirs

tylerwilliams:
Add a test and bring this CL back behind a flag (disabled).

Fix the bug we encountered  before, which was that existing filecache files were not being renamed during the startup scan.


enterprise/server/remote_execution/filecache/filecache_test.go
```diff
@@ -352,6 +358,77 @@ func TestScanWithConcurrentAdd(t *testing.T) {
...
+	ctx := context.Background()
+	fcDir := testfs.MakeTempDir(t)
+
+	baseDir := testfs.MakeTempDir(t)
+
+	{
+		fc, err := filecache.NewFileCache(fcDir, 4096*10, false)
+		if err != nil {
+			t.Fatal(err)
+		}
```
bduffany, line 371: nit - might be worth tossing a `WaitForDirectoryScanToComplete` here - idk if the old scanner goroutine will clash with the cache created in the next block below

tylerwilliams: Done.

bduffany (approved):

vadimberezniker (approved):

Issue #7896, opened 2024-11-14
[RB/CLI] `bb remote` does not work in a Git worktree


goniz:
**Describe the bug**
Trying to run `bb remote` in a Git worktree directory, it fails with:
`remote config: determine remote: the git repository must have a remote configured to use remote Bazel`

**To Reproduce**

1. Clone a git repo
2. Create a new worktree using `git worktree add /path/to/worktree-dir $ref`
3. `cd /path/to/worktree-dir`
4. `bb remote run $label`

**Additional context**
A worktree path does not have a `.git` directory , which probably is the cause for the failure(?)

maggie-lou:
Sounds good. I've been meaning to refactor out go-git anyway. There have been a series of unhandled edge cases

PR #7933, opened 2024-11-21
[RB] Refactor CLI to not use go-git when determining the remote

maggie-lou:
Our usage of go-git in the remote bazel CLI doesn't handle a series of edge cases and has resulted in some performance problems

To make these changes easier to review, I'll refactor out go-git in a series of PRs

This PR also adds some debug logging to the CLI for how long each git command takes, as customers often wonder where time is spent by the CLI

Related issues:
https://github.com/buildbuddy-io/buildbuddy/issues/7896
https://github.com/buildbuddy-io/buildbuddy-internal/issues/3717
https://github.com/buildbuddy-io/buildbuddy-internal/issues/1682
https://buildbuddy-corp.slack.com/archives/C04S4M5U3BK/p1731617106465189

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -116,6 +116,13 @@ type RunOpts struct {
 	WorkspaceFilePath string
 }
 
+type gitRemote struct {
+	name string
+	url  string
+	// What the remote is used for, like 'fetch' or 'pull'
+	use string
+}
+
 type RepoConfig struct {
 	Root          string
 	URL           string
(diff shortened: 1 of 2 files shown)


bduffany (approved):

cli/remotebazel/remotebazel.go
```diff
@@ -125,69 +132,106 @@ type RepoConfig struct {
...
+		url:  parts[1],
+	}
+	if len(parts) >= 3 {
+		use := parts[2]
+		use = strings.TrimSuffix(use, ")")
+		use = strings.TrimPrefix(use, "(")
+		r.use = use
+	}
+
+	return r, nil
```
sluongng, line 233: FWIW, over in `gh` cli, they use a regex to parse this string https://github.com/cli/cli/blob/b133f9cce0767071c004708c83b7d934187a63d0/git/client.go#L22

sluongng (approved):
LGTM. Thanks so much.

cli/remotebazel/remotebazel.go
maggie-lou: Oh nice that's cleaner - thanks

PR #7937, opened 2024-11-21
Simplify how auxiliary metadata is set in  ExecuteTaskAndStreamResults

vanja-p:
Set it in only one place instead of two.


vanja-p:
whoops, including extra files and changes. Fixing it.

vanja-p:
ok fixed. PTAL

bduffany (approved):
The current logic was somewhat intentional as it would allow us to sprinkle more auxiliary metadata and it would "just work". This LGTM but if we add more metadata, it will probably evolve back to how it is now :shrug:

vanja-p:
> This LGTM but if we add more metadata, it might just evolve back to how it is now 🤷

Yep, you were right about this. Also, one thing I didn't catch is that before this change, we only appended ExecutionAuxiliaryMetadata on the last send, and now we do it on each send, which isn't necessary.

PR #7941, opened 2024-11-22
Set cgroup settings in the app

bduffany:
Behind a flag, have the app apply cgroup settings for Linux executions, instead of the executor.


vanja-p (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -419,14 +420,23 @@ func (h *executorHandle) EnqueueTaskReservation(ctx context.Context, req *scpb.E
...
+
 	// Just before enqueueing, resize the task to match the measured or
 	// predicted task size, up to the executor's limits. This late-resizing
 	// approach ensures that we always determine schedulability using the
-	// "default" estimate, which is stable. This way, tasks can't suddenly
-	// become unschedulable due to a change in measurements or an adjustment to
-	// the prediction model.
+	// default + requested estimates, which are stable. This way, tasks can't
+	// suddenly become unschedulable due to a change in measurements or an
+	// adjustment to the prediction model.
```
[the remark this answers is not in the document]
tylerwilliams: +1 I don't understand this comment either! Was in here the other day scratching my head lol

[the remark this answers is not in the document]
bduffany: There's a `nodesThatFit` function which determines which set of nodes that a task can fit on. This determines which nodes we will try to enqueue the task on (i.e. which nodes we will call `EnqueueTaskReservation` on).

`nodesThatFit` doesn't take measured size or predicted size into account. This is because we don't want to wind up in a situation where a "greedy" task executes on a large node and uses up all of its resources, then suddenly will only schedule on large nodes.

So, the comment is trying to convey that because we already made a scheduling decision, `adjustTaskSize` needs to be careful not to increase the size so much that the task is no longer schedulable (because it will no longer fit on the node).

I updated the comment to hopefully make it more clear why we are doing this. That said, we could make this less confusing using one or more of these changes:

* Enforce global maximum task sizes using hard limits, to mitigate the "greedy task" scenario
  * One caveat here is that we actually cannot enforce limits on macOS.
* Have `nodesThatFit` only check user-requested size. In other words, unless the user specifically requested 8 CPU cores and 16GB memory, I think we should just run a task on any node in the pool, and just use whatever resources it has available.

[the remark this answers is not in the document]
vanja-p: I think your comment change helps a lot so this LGTM.

Since I spent a bunch if time thinking about your reply, here's what I found and how I'm thinking about it.
- At the points where `nodesThatFit` and `adjustTaskSize` are called, `EnqueueTaskReservationRequest.task_size and EnqueueTaskReservationRequest.scheduling_metadata.task_size` are based only on the default and requested size. This is done [here in the execution_server](https://github.com/buildbuddy-io/buildbuddy/blob/f82cab6bfdc46bc14eb9c1a1969f96a2f783f721/enterprise/server/remote_execution/execution_server/execution_server.go#L608).
- `adjustTasksize` is called after we pick the executor. This is obvious because it's attached to the `executorHandle` type.
- `nodesThatFit` treats the requested and default size as mandatory. That is, we won't even try to run the action if we can't fit those sizes on any executor, presumably because we want to honor the user's request, and the action might fail if it doesn't have enough resources. The default size really doesn't matter though, because it's relatively small.
- `adjustTaskSize` allows the measured or predicted size to serve as an optimization. That is, allow the action to use more or less resources, but only within the limits of the executor we already picked. Here we assumed that the action can pass within the limits of the executor, but maybe it will run faster with higher limits.
- This is complicated by the fact that [measured](https://github.com/buildbuddy-io/buildbuddy/blob/f82cab6bfdc46bc14eb9c1a1969f96a2f783f721/enterprise/server/tasksize/tasksize.go#L133) and [predicted](https://github.com/buildbuddy-io/buildbuddy/blob/f82cab6bfdc46bc14eb9c1a1969f96a2f783f721/enterprise/server/tasksize_model/tasksize_model.go#L206) sizes aren't set when the user specifies EstimatedComputeUnits, but they are when the user sets EstimatedCPU or EstimatedMemory. So `adjustTaskSize` doesn't do anything when EstimatedComputeUnits is set.

My thoughts on how this could be improved:
- Changing `nodesThatFit` to only consider requested size makes sense, but I doubt it would have any affect. I suspect that the default size for any task is smaller than the smallest executor.
- I think it makes sense to cap the measured and predicted size, but instead of capping them to the size of the picked executor, I would pick some constant cap that is smaller than the smallest executor. This would make it easier to reason about and give more predictable performance.
- Hard limits are definitely a good idea. Ideally, they smaller than the smallest executor, and then the hard limit can be the same as the cap on the requested, predicted, and measured sizes.
  - For macs, the hard limit could be the size of the smallest mac. Yes the action could run on a bigger mac and use more, but it shouldn't be guaranteed that.
  - Even if the action ends up bursting over N CPU, we can still set a hard limit for the CPU it can ask for in the future.

vanja-p (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
```diff
@@ -419,14 +420,26 @@ func (h *executorHandle) EnqueueTaskReservation(ctx context.Context, req *scpb.E
...
+
+	// At this point, we haven't yet used the measured size or predicted size at
+	// all when deciding how to schedule this task. This is intentional - we
+	// don't want a task to suddenly become unschedulable on some nodes in a
+	// heterogenous pool, just because it executed one time on one of the larger
+	// nodes, and it greedily used up lots of resources on that node.
+	//
+	// So, once we've decided to schedule a task on a node, only then do we
+	// apply the measured and predicted sizes - but we limit to the node's
+	// capacity, to ensure that it is still schedulable within the node.
```
tylerwilliams, line 435: I understand this comment better now, thanks for clarifying it a bit.

I still find the name `adjustTaskSize` very confusing -- because what I believe is happening is:
 - we compute several different task sizes using different methods
 - when selecting executors to run a task, we use the default estimate, not the measured or predicted size (right?)
 - when running a task on a node, we use the best size (measured or predicted or default or something)

I mentally think of the full code as doing:
executors := filterExecutors(allExecutors, task.GetDefaultSize())
...
mostAccurateTaskSize := selectMostAccurateTaskSize(task)
req.SchedulingMetadata.CgroupSettings = GetCgroupSettings(mostAccurateTaskSize)
...

So it's a bit confusing that "adjustTaskSize" is called adjust (what is being adjusted??) and that it mutates req, rather than just selecting the best size or something.

enterprise/server/tasksize/tasksize.go
```diff
@@ -29,7 +29,12 @@ var (
 	useMeasuredSizes    = flag.Bool("remote_execution.use_measured_task_sizes", false, "Whether to use measured usage stats to determine task sizes.")
 	modelEnabled        = flag.Bool("remote_execution.task_size_model.enabled", false, "Whether to enable model-based task size prediction.")
 	psiCorrectionFactor = flag.Float64("remote_execution.task_size_psi_correction", 1.0, "What percentage of full-stall time should be subtracted from the execution duration.")
-	milliCPULimit       = flag.Int64("remote_execution.task_size_millicpu_limit", 7500, "Limit placed on milliCPU calculated from task execution statistics.")
+	cpuQuotaLimit       = flag.Duration("remote_execution.cpu_quota_limit", 30*100*time.Millisecond /*30 cores*/, "Maximum CPU time allowed for each quota period.")
+	cpuQuotaPeriod      = flag.Duration("remote_execution.cpu_quota_period", 100*time.Millisecond, "How often the CPU quota is refreshed.")
+	pidLimit            = flag.Int64("remote_execution.pids_limit", 2048, "Maximum number of processes allowed per task at any time.")
```
tylerwilliams, line 34: is this going to give anyone a haircut?

tylerwilliams (approved):

enterprise/server/scheduling/scheduler_server/scheduler_server.go
bduffany: I agree that the "adjust" terminology is confusing - changed it to a function that returns the most accurate size like in your example. We still need to set `req.TaskSize` and `req.SchedulingMetadata.TaskSize` though, because those fields determine the actual size on the executor. (If that part is still confusing, that's part of the reason why I thought we should deprecate `req.SchedulingMetadata.TaskSize` - though I think it could still be clearer)

enterprise/server/tasksize/tasksize.go
bduffany: No, this is the current effective value

PR #7943, opened 2024-11-22
Limit total input tree setup parallelism across all actions.

vadimberezniker:
Pipe the operations through a global queue with a fixed number of workers.


tylerwilliams (approved):
LGTM

PR #7950, opened 2024-11-25
userdb: deduplicate groupids when create user

sluongng:
Under certain conditions, we could be adding users to the default group
twice and got this error as the result:

```
Error 1062 (23000): Duplicate entry 'US16246136860140428328-GR0000000000000000000' for key 'UserGroups.PRIMARY'
```

Ensure that all ids in groupids is unique by using a map instead of a
slice.


sluongng:
Context: https://buildbuddy-corp.slack.com/archives/C0495TM9UUE/p1726649737517399

bduffany (approved):

sluongng:
The baremetal tests are failing due to unrelated reasons (tested from `master` and got the same failures). Merging this.

PR #7954, opened 2024-11-25
Disable block_io on macos for now

bduffany:
`/proc/mounts` is not a thing on macOS.


tylerwilliams (approved):

bduffany:
errr, I think this can be simplified by just doing a single `stat` syscall - taking a closer look...

PR #7959, opened 2024-11-26
Extract only output paths from firecracker workspaces

bduffany:


vanja-p (approved):

vadimberezniker (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -2870,3 +2867,28 @@ func getRandomNUMANode() (int, error) {
...
+func workspacePathsToExtract(task *repb.ExecutionTask) []string {
+	if platform.IsTrue(platform.FindEffectiveValue(task, platform.PreserveWorkspacePropertyName)) {
+		return []string{"/"}
+	}
+
+	// Special files
+	// TODO: declare this list as a constant somewhere?
+	paths := []string{
+		".BUILDBUDDY_DO_NOT_RECYCLE",
+		".BUILDBUDDY_INVALIDATE_SNAPSHOT",
```
bduffany, line 2885: @maggie-lou FYI I preemptively added the new file here from your in-flight PR. After our PRs are both merged I'm thinking of maybe adding a list to `container.go` with these special file names

PR #7960, opened 2024-11-26
metrics: record task queue length with executor pool name

sluongng:
For some executor deployments, users might want to implement autoscaling differently for different executor pools.

Decorate our queue length metric with pool name and export it through prom API.

### DIFF
--- a/enterprise/server/backends/prom/prom.go
+++ b/enterprise/server/backends/prom/prom.go
@@ -40,13 +40,13 @@ var (
 	MetricConfigs = []*MetricConfig{
 		{
 			sourceMetricName: "buildbuddy_remote_execution_queue_length",
-			LabelNames:       []string{podNameLabel},
+			LabelNames:       []string{podNameLabel, metrics.ExecutorPool},
 			ExportedFamily: &dto.MetricFamily{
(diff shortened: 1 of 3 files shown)


sluongng:
Context: https://buildbuddy-corp.slack.com/archives/C0495TM9UUE/p1732549575265739

bduffany:
Another question: are these `exported_*` metrics scraped from self-hosted executors? I don't think we have the capability to scrape self-hosted executors, so I'd be somewhat surprised if this works - but @luluz66 knows how this works better than I do

goniz:
Hi,
The request originated for the purpose of implementing scaling to zero on our self hosted executors (GPU cluster) to reduce cost.
Anyhow, even if the `exported_` metrics were to include self-hosted executors it's still not possible right now due to the scheduler not queuing work for non-registered pools.
Meaning failing builds during scale to 0 instead of scaling up.
We decided to skip this for now and to scale to 1 during non-working hours.

Thank you @sluongng @bduffany @luluz66 !

sluongng:
We discussed this internally and we think it might be better for us to add a separate metric to support the "scale to zero" use case for separate executor pools. It will take a bit more effort to implement, but it should work much better. Closing this PR for now.

PR #7961, opened 2024-11-26
Run firecracker workspace conversion in action cgroup

bduffany:
Workspace ext4 conversion is IO-heavy. Do this work in the action cgroup so that we can limit IO and CPU usage of this conversion, and also report it in execution metrics.


bduffany:
I'm a little on the fence about this change because workspace conversion feels a little bit like it should be part of the "prepare inputs" phase and not really count toward the action execution, but on the other hand, it might make sense to move the "prepare inputs" phase inside the action cgroup too. i.e. this PR would be a step towards measuring/controling all of the resource usage for all phases of an action, not just the execution phase.

PR #7964, opened 2024-11-26
[RB] Fix test

maggie-lou:
I realized there is a small mistake in this test. During a `bazel run` on the remote runner, the execution output is always in the outer parent invocation, not the child invocation. So this test would always succeed, even if something is broken

### DIFF
--- a/enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
+++ b/enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
@@ -493,7 +493,7 @@ func TestBuildRemotelyRunLocally(t *testing.T) {
 
 	var parentInv *inpb.Invocation
 	for _, inv := range searchRsp.GetInvocation() {
-		if inv.Command != "run" {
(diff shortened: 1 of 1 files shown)


enterprise/server/test/integration/remote_bazel/remote_bazel_test.go
```diff
@@ -491,16 +491,16 @@ func TestBuildRemotelyRunLocally(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, 2, len(searchRsp.GetInvocation()))
 
-	var childInv *inpb.Invocation
+	var parentInv *inpb.Invocation
 	for _, inv := range searchRsp.GetInvocation() {
-		if inv.Command == "run" {
-			childInv = inv
+		if inv.Command != "run" {
```
bduffany, line 496: nit: maybe `inv.GetParentRunId() == ""` ? (to directly check that it's the root invocation)

bduffany (approved):

PR #7965, opened 2024-11-26
[COW] Don't build on windows

maggie-lou:
The window build is broken because certain syscalls are not supported on windows: https://github.com/buildbuddy-io/buildbuddy/actions/runs/12037002728/job/33560293374


sluongng:
Since this package is only used for firecracker runner, I think we can simply copy the current approach from the firecracker package and add
```
    target_compatible_with = [
        "@platforms//os:linux",
    ],
```
to the go_lib target.

bduffany (approved):

PR #7967, opened 2024-11-26
Set cached_result=true when there is a cache hit in the execution ser…

vanja-p:
…vice

Without this, these cache hits look like executions to bazel. I don't think it matters to functionality, but it does underreport the cache hit ratio.


bduffany (approved):
From what I could tell, it looks like this field is only ever plumbed through to bazel's remote exec log file, so this LGTM

PR #7974, opened 2024-11-27
Write to the action cache in the execution server

vanja-p:
This is done when the executor sends a completed action to PublishOperation, and only when the executor populates `ExecutionAuxiliaryMetadata.execution_task`. This is for 2 reasons:
1. Writing failed actions to the cache requires the invocation ID, and PublishOperation doesn't get it currently.
2. This allows us to avoid a period of time where both the executor and the app write to the cache.

I didn't include the executor changes in this PR. My plan was to let this roll out, and then send an executor PR that would start populating `ExecutionAuxiliaryMetadata.execution_task` and stop writing to the cache. This avoids double writes, but it does still leave a window where we have no writes, if the executor is updated before the app. Let me know what you think.

There are a few other behavior changes that will result from these changes:
1. Actions written to the cache will no longer contain the `auxiliary_metadata`, `io_stats`, and `usage_stats` fields. They all contain buildbuddy specific data, so I don't think bazel needs them, but I'm not sure if some other part of the app reads from the cache and expects these fields to be there.
2. The same fields are also removed from the Operation that we send to the execution's pubsub channel, which I think ends up getting sent to bazel.
3. The cached ExecuteResponses still have these fields, and importantly, auxiliary_metadata now also includes the ExecutionTask, which might make these entries much bigger.

I could undo any of these changes if you think they are too risky.


enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -1090,13 +1125,33 @@ func (s *ExecutionServer) cacheExecuteResponse(ctx context.Context, taskID strin
...
+func (s *ExecutionServer) cacheActionResult(ctx context.Context, actionResourceName *digest.ResourceName, response *repb.ExecuteResponse, md *espb.ExecutionAuxiliaryMetadata) error {
+	if response.GetCachedResult() || md.GetExecutionTask() == nil {
+		return nil
+	}
+	// The action executed and the executor sent us an ExecutionTask in the
+	// auxiliary metadata, which indicates we need to write to the cache.
+	action := md.GetExecutionTask().GetAction()
+	// If the action failed or do_not_cache is set, upload information about the error via a failed
+	// ActionResult under an invocation-specific digest, which will not ever be seen by bazel but
+	// may be viewed via the Buildbuddy UI.
```
bduffany, line 1137: I forgot to mention this earlier - we don't need to store failed ActionResults in the cache anymore, because the UI relies on the ExecuteResponse stored with `cacheExecuteResponse` to render the results now (both in the success and failure cases)

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -1090,13 +1125,33 @@ func (s *ExecutionServer) cacheExecuteResponse(ctx context.Context, taskID strin
...
+	cacheableResourceName := actionResourceName
+	if action.GetDoNotCache() || response.GetStatus().GetCode() != 0 || response.GetResult().GetExitCode() != 0 {
+		resultDigest, err := digest.AddInvocationIDToDigest(actionResourceName.GetDigest(), actionResourceName.GetDigestFunction(), md.GetExecutionTask().GetInvocationId())
+		if err != nil {
+			return status.UnavailableErrorf("Error uploading action result: %s", err.Error())
+		}
+		cacheableResourceName = digest.NewResourceName(resultDigest, actionResourceName.GetInstanceName(), rspb.CacheType_AC, actionResourceName.GetDigestFunction())
```
bduffany, line 1144: I think these failed ActionResults are no longer used, so we can short-circuit in this case. The UI fetches the response that was stored with cacheExecuteResponse now, which includes the complete response and works for both failed and succeeded actions.

proto/execution_stats.proto
```diff
@@ -135,6 +135,9 @@ message Execution {
 message ExecutionAuxiliaryMetadata {
   // Platform overrides set via remote header.
   build.bazel.remote.execution.v2.Platform platform_overrides = 1;
+
+  // The ExecutionTask that the executor received from the scheduler.
+  build.bazel.remote.execution.v2.ExecutionTask execution_task = 2;
```
bduffany, line 140: once we start populating this, we probably want to keep an eye on the prometheus metrics for AC write size and make sure it looks OK

vanja-p: Would that be the metric on the right of this screenshot?

image

enterprise/server/remote_execution/execution_server/execution_server.go
vanja-p: I sent https://github.com/buildbuddy-io/buildbuddy/pull/7985 to first remove these writes.

proto/execution_stats.proto
bduffany: The UI uses the auxiliary metadata (if you were looking for references using `AuxiliaryMetadata` you'll miss the UI references because they are `camelCase`)

bduffany: re. metrics, `Upload throughput` accounts for CAS blobs as well - you could either run a one-off query to restrict to the AC label or update that dashboard to split out those charts by AC vs CAS

vanja-p: Ah, that's a good point about the UI. I decided to play it safe and not remove any existing data. Instead, the ExecutionTask will come as a separate auxiliary metadata, which can easily be removed without messing with the other metadata. PTAL.

Now that I don't need the invocation ID, another option is to not use the presence of the ExecutionTask to trigger cache writes from the app. All the necessary data is already available in PublishOperation. One downside of this is we might double write to the cache, especially in the case of self-hosted executors that aren't getting updated.

bduffany: I think the double-write for older executor clients (and during the rollout) seems fine - if we notice it being a problem we could ask people to upgrade and/or gate the logic behind some sort of version check.

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -1007,31 +1008,61 @@ func (s *ExecutionServer) PublishOperation(stream repb.Execution_PublishOperatio
 			return err
 		}
 
+		if op.GetName() == "" {
```
bduffany, line 1011: this change feels a little risky, is it needed?

bduffany (approved):

proto/execution_stats.proto
vanja-p: Here is the version without avoiding double-writes: https://github.com/buildbuddy-io/buildbuddy/pull/8009. Still waiting for Tyler's thoughts on which one he prefers.

enterprise/server/remote_execution/execution_server/execution_server.go
vanja-p: I could revert this change, but as far as I can tell, there's no way for PublishOperation to do anything useful without the name being an execution ID.

proto/execution_stats.proto
vanja-p: Tyler's OK with the double writes, so ignore this PR in favor of https://github.com/buildbuddy-io/buildbuddy/pull/8009

PR #7975, opened 2024-11-27
Document ScheduleTaskRequest.serialized_task

vanja-p:


proto/scheduler.proto
```diff
@@ -334,6 +334,9 @@ message SchedulingMetadata {
 message ScheduleTaskRequest {
   string task_id = 1;
   SchedulingMetadata metadata = 2;
+
+  // Serialized build.bazel.remote.execution.v2.ExecutionTask. That proto
+  // depends on this one, so using the full type would create a build cycle.
```
bduffany, line 339: nit: we could move `ExecutionTask` to `scheduler.proto` if we wanted to break this cycle, so the note about the dependency cycle feels a bit incidental. (We created this problem ourselves by tacking on `ExecutionTask` to the upstream API protos rather than adding our own package.) It might make sense to omit that part to avoid confusion, or maybe add some additional clarification.

bduffany (approved):

proto/scheduler.proto
vanja-p: Clarified a bit more.

PR #7978, opened 2024-12-02
redis_client: provide flags to enable TLS

sluongng:
Provide a set of flags that would enable TLS when connecting to Redis
using Go Redis client (false by default).

note: In the future, we might want to add support for custom certificate
as well with flags pointing to custom cert+key files.

### DIFF
--- a/enterprise/server/util/redisutil/redisutil.go
+++ b/enterprise/server/util/redisutil/redisutil.go
@@ -68,28 +68,42 @@ func TargetToOptions(redisTarget string) *redis.Options {
 	}
 }
 
-func TargetToOpts(redisTarget string) *Opts {
+func TargetToOpts(redisTarget string, useTLS bool) *Opts {
 	if redisTarget == "" {
 		return nil
 	}
 	libOpts := TargetToOptions(redisTarget)
+	var tlsConfig *tls.Config
+	if useTLS {
+		tlsConfig = &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+	}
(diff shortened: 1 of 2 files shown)


vadimberezniker (approved):

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -68,28 +68,42 @@ func TargetToOptions(redisTarget string) *redis.Options {
 	}
 }
 
-func TargetToOpts(redisTarget string) *Opts {
+func TargetToOpts(redisTarget string, useTLS bool) *Opts {
 	if redisTarget == "" {
 		return nil
 	}
 	libOpts := TargetToOptions(redisTarget)
+	var tlsConfig *tls.Config
```
zolution, line 76: In redis.ParseURL(), as long as the connection string is with scheme "rediss://", the TLS config will be returned in libOpts [1]
In this pull request, it will ignore the redis TLS scheme specified in the connection string, and only refer to the Buildbuddy configuration flags.
Would it be better if we directly use the scheme provided in redis connection string instead of create a seperate flag for all?

[1] 
https://github.com/redis/go-redis/blob/v8.11.5/options.go#L222
https://github.com/redis/go-redis/blob/v8.11.5/options.go#L269

sluongng: @zolution that's a really good call!

Could you please give https://github.com/buildbuddy-io/buildbuddy/pull/8014 a read and see if that would work better for you?

PR #7982, opened 2024-12-02
Add GetPublicKeys RPC to AuthService for retrieving RSA public key(s)

iain-macdonald:


proto/auth.proto
```diff
@@ -11,6 +11,17 @@ message AuthenticateResponse {
...
 }
 
+message GetPublicKeysRequest {}
+
+message GetPublicKeysResponse {
+  repeated PublicKey public_keys = 1;
+}
+
+message PublicKey {
+  optional string key = 1;
```
tylerwilliams, line 21: could you add a comment with an example of a public key?

tylerwilliams (approved):
LGTM, Approval

proto/auth.proto
iain-macdonald: done!

PR #7987, opened 2024-12-02
Simplify stats polling logic

bduffany:
This PR cleans up some stats logic in preparation for tracking firecracker stats using cgroups, instead of tracking stats using the guest vmexec server. The current logic is somewhat tangled up which would make the integration with firecracker a bit more messy than it needs to be.

* Remove the logic to record prometheus metrics for running containers, which manually totals the results of `c.Stats()`. This logic is complex and its value is pretty questionable.
  * On Linux, a better way to do this would be to poll the stats from the action parent cgroup once `executor.child_cgroups_enabled` is rolled out.
  * On macOS we could probably just live with the node metrics instead of trying to separate the executor metrics and task metrics. The accuracy of the task metrics on macOS is especially questionable, since the poll rate is lower, and it requires walking the process tree, which can miss short-lived processes spawned during an execution.
* Simplify the stats polling logic, which was tangled up and somewhat circular.
  * The previous logic would have each container call the `TrackStats` utility function, which would start a loop that calls back to the container's `Stats()` function, which was expected to call `usageStats.Update()` then return `usageStats.TaskStats()`.
  * Now, `TrackStats` is a method of the `UsageStats` struct, and manages these careful state updates internally, without a circular call back to the `container.Stats()` function, and without expecting the caller to call `usageStats.Reset()` and `usageStats.TaskStats()`

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -122,9 +122,6 @@ type Init struct {
 // UsageStats holds usage stats for a container.
 // It is useful for keeping track of usage relative to when the container
 // last executed a task.
-//
-// TODO: see whether its feasible to execute each task in its own cgroup
-// so that we can avoid this bookkeeping and get stats without polling.
 type UsageStats struct {
 	Clock clockwork.Clock
 
@@ -294,9 +291,7 @@ func (s *UsageStats) TrackExecution(ctx context.Context, lifetimeStatsFn func(ct
 		defer func() {
 			// Only log an error if the task ran long enough that we could
 			// reasonably expect to sample stats at least once while it was
-			// executing. Note that we can't sample stats until podman creates
-			// the container, which can take a few hundred ms or possibly longer
-			// if the executor is heavily loaded.
+			// executing.
 			dur := time.Since(start)
 			if dur > 1*time.Second && lastErr != nil && s.TaskStats() == nil {
 				log.CtxWarningf(ctx, "Failed to read container stats: %s", lastErr)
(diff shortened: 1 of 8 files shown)


enterprise/server/remote_execution/container/container.go
vanja-p, line 1: is the TODO on line 126 completed now?

enterprise/server/remote_execution/container/container.go
```diff
@@ -392,7 +298,7 @@ func TrackStats(ctx context.Context, c CommandContainer) (stop func(), res <-cha
 			// the container, which can take a few hundred ms or possibly longer
```
vanja-p, line 298: the mention of podman on the line above might be out of date.

vanja-p (approved):

enterprise/server/remote_execution/container/container.go
bduffany: That part is still accurate, since podman doesn't (yet) have the logic where we create the cgroup then run the container in that cgroup that we created. Instead, podman manages the cgroup lifecycle itself.

I removed this note for now though, since it's not super important. The bigger TODO is to make sure that the container implementations don't delete the cgroup from under us (this is a problem with both ociruntime and podman). That way, we could remove this polling, and reliably read the cgroup stats once at the end of execution. I think we could do this by creating a child cgroup under the container cgroup and run oci/podman containers in the child, so that even after crun/podman delete the child cgroup, we can still read the stats from the parent cgroup.

iain-macdonald (approved):

PR #7989, opened 2024-12-02
[RB] Finish refactoring out go-git from CLI

maggie-lou:
Our usage of go-git in the remote bazel CLI doesn't handle a series of edge cases and has resulted in some performance problems. This PR is a continuation of https://github.com/buildbuddy-io/buildbuddy/pull/7933

Also:
* Fixes an assumption that the default branch had to be named `main` or `master`
* Tries to minimize slow remote git calls by making one `git ls-remote --symref origin` call at the beginning, and passing the output to helper function
* Fixes a bug where the flags weren't parsed when we were trying to use them

Fixes https://github.com/buildbuddy-io/buildbuddy/issues/7896
Fixes https://github.com/buildbuddy-io/buildbuddy-internal/issues/3717

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -399,21 +399,20 @@ func getBaseBranchAndCommit(remoteData string) (branch string, commit string, er
 // HEAD state
 func getCurrentRef() (string, error) {
 	currentBranch, err := runGit("symbolic-ref", "--short", "HEAD")
-	if err != nil {
-		if strings.Contains(err.Error(), "ref HEAD is not a symbolic ref") {
-			// Handle detached head state
-			detachedHeadOutput, _ := runGit("branch")
-			regex := regexp.MustCompile(".*detached at ([^)]+).*")
-			matches := regex.FindStringSubmatch(detachedHeadOutput)
-			if len(matches) != 2 {
-				return "", status.UnknownErrorf("unexpected branch state %s", detachedHeadOutput)
-			}
-			currentBranch = matches[1]
-		} else {
-			return "", status.WrapError(err, "get current branch")
-		}
+	if err == nil {
+		return strings.TrimSpace(currentBranch), nil
+	} else if !strings.Contains(err.Error(), "ref HEAD is not a symbolic ref") {
+		return "", status.WrapError(err, "get current branch")
 	}
-	return strings.TrimSpace(currentBranch), nil
+
+	// Handle detached head state
+	detachedHeadOutput, _ := runGit("branch")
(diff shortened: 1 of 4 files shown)


cli/remotebazel/remotebazel.go
```diff
@@ -303,45 +282,40 @@ func diffUntrackedFile(path string) (string, error) {
...
+func Config() (*RepoConfig, error) {
 	remote, err := determineRemote()
 	if err != nil {
 		return nil, status.WrapError(err, "determine remote")
 	}
 	fetchURL := remote.url
 	log.Debugf("Using fetch URL: %s", fetchURL)
 
-	branch, commit, err := getBaseBranchAndCommit(repo)
+	remoteData, err := runGit("remote", "show", remote.name)
```
sluongng, line 303: Nit: This is not obvious, but it is making a network call to the remote server with `git remote show origin`. We need to add `-n` to avoid network calls. I suspect if we need the remote information, `git ls-remote origin` might give us richer information.

sluongng (approved):

sluongng:
Can we remove go-git from our repo entirely after this? 🤔

cli/remotebazel/remotebazel.go
sluongng: https://github.com/git/git/blob/cc01bad4a9f566cf4453c7edd6b433851b0835e2/builtin/remote.c#L1322-L1329

Looking at this, it seems like both `git remote show` and `git ls-remote` would call `transport_get_remote_refs` to list all remote refs. So I think we might be able to refactor the code to only call `git ls-remote` once, then re-use the result.

cli/remotebazel/remotebazel.go
```diff
@@ -441,6 +395,52 @@ func getBaseBranchAndCommit(repo *git.Repository) (branch string, commit string,
...
+			if len(matches) != 2 {
+				return "", status.UnknownErrorf("unexpected branch state %s", detachedHeadOutput)
+			}
+			currentBranch = matches[1]
+		} else {
+			return "", status.WrapError(err, "get current branch")
+		}
+	}
+	return strings.TrimSpace(currentBranch), nil
+}
```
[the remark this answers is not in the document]
maggie-lou: Done

maggie-lou:
> Can we remove go-git from our repo entirely after this? 🤔

Unfortunately not :/ Looks like we have a couple other usages

cli/remotebazel/remotebazel.go
maggie-lou: Refactored so the CLI makes a single `git ls-remote --symref origin` call

PR #7990, opened 2024-12-02
[raft] refactor driver start and stop code

luluz66:
To prepare the driver for rebalancing lease counts, the driver needs to be
stopped at the start of the store.Stop() method before we try to drop all the
leases held.

### DIFF
--- a/enterprise/server/raft/store/store.go
+++ b/enterprise/server/raft/store/store.go
@@ -517,6 +517,9 @@ func (s *Store) AddEventListener() <-chan events.Event {
 // ranges on this node.
 func (s *Store) Start() error {
 	s.usages.Start()
+	if s.driverQueue != nil {
+		s.driverQueue.Start()
+	}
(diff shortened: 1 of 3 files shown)


enterprise/server/raft/store/store.go
```diff
@@ -567,6 +564,9 @@ func (s *Store) Start() error {
 
 func (s *Store) Stop(ctx context.Context) error {
 	s.log.Info("Store: started to shut down")
+	if s.driverQueue != nil {
```
tylerwilliams, line 567: Why do we need these if not nil checks in start and stop?

tylerwilliams (approved):

enterprise/server/raft/store/store.go
luluz66: because we have the flag to enable driver. if the flag is off, the driverQueue is nil and can cause nil pointer excepter.

PR #7991, opened 2024-12-02
Add memory usage timelines

bduffany:
Example:

image

This screenshot is from running this python script as a test action. It repeatedly doubles the length of a string, starting from a string of length 1, sleeping for a bit before each doubling operation. It runs this whole procedure for 3 total iterations, sleeping a bit between each run.

```python
import gc
import time

for _ in range(3):
  s = "A"
  while len(s) < 100e6:
    s += s
    gc.collect()
    time.sleep(0.25)
  time.sleep(1)
  gc.collect()
```

### DIFF
--- a/enterprise/server/remote_execution/container/container.go
+++ b/enterprise/server/remote_execution/container/container.go
@@ -250,9 +250,9 @@ type UsageStats struct {
 // that TaskStats() can just return a view of the data rather than requiring a
 // full re-encoding.
 type timelineState struct {
-	lastTimestamp    int64
-	lastCPUSample    int64
-	lastMemorySample int64
+	lastTimestampUnixMillis int64
+	lastCPUMillis           int64
+	lastMemoryKB            int64
 	// When adding new fields here, also update:
 	// - The size calculation in updateTimeline()
 	// - The test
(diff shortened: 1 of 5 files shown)


enterprise/server/remote_execution/container/container.go
```diff
@@ -335,9 +338,18 @@ func (s *UsageStats) TaskStats() *repb.UsageStats {
 }
 
 func (s *UsageStats) updateTimeline(now time.Time) {
-	if 8*(len(s.timeline.GetCpuSamples())+len(s.timeline.GetTimestamps())) > timeseriesSizeLimitBytes {
+	totalLength := 0
```
vanja-p, line 341: nit: this loop is neat, but this is shorter and probably easier to read:
```
st := s.timeline
totalLength := st.GetTimestamps() + st.GetCpuSample() + st.GetMemoryKbSamples()
```

enterprise/server/remote_execution/container/container.go
```diff
@@ -250,11 +250,14 @@ type UsageStats struct {
 // that TaskStats() can just return a view of the data rather than requiring a
 // full re-encoding.
 type timelineState struct {
```
vanja-p, line 252: maybe document the units for the fields in this struct

vanja-p (approved):

enterprise/server/remote_execution/container/container.go
bduffany: Renamed the fields to include units

enterprise/server/remote_execution/container/container.go
bduffany: Done

PR #7996, opened 2024-12-03
Add disk IO usage timelines

bduffany:
Screenshot from an action that generates a 5GB file using `head -c 5000000000 /dev/zero > /tmp/zeros.bin`, then creates a copy of it with `cp /tmp/zeros.bin /tmp/zeros_copy.bin`

image


siggisim (approved):
So cool!!

PR #7997, opened 2024-12-03
Upgrade `rules_nodejs` to pick up newer node versions

tempoz:
The current version of `rules_nodejs` does not list in its `node_versions.bzl` any versions newer than `18.13.0`, but we need `18.17.0` or newer for a newer version of `cheerio`.


sluongng (approved):
LGTM

sluongng:
Reference for my future self: https://github.com/bazel-contrib/rules_nodejs/compare/5.8.2...5.8.5

PR #7999, opened 2024-12-03
Upgrade `node` to `18.20.3` to support newer npm deps

tempoz:
`cheerio` version `1.0.0` requires `node` version `18.7.0` or newer.


bduffany:
is this node version also compatible with the //website build + internal website build?

tempoz:
It definitely works for the website build; I forgot to check on the internal website. I'll check it and confirm.

tempoz:
Looks like internal uses node `16.19.0`; seems to me like upgrading that would be out of scope. wdyt?

bduffany (approved):

PR #8006, opened 2024-12-04
test github workflows

sluongng:
https://github.com/actions/runner-images/issues/10636


sluongng:
seems like we should be ok with 24.04 🤔

PR #8007, opened 2024-12-04
Remove unused import in `remote_runner.proto`

tempoz:
`google/rpc/status.proto` is unused in `proto/api/v1/remote_runner.proto`, which causes a logged warning during build.


sluongng:
LGTM but I think @maggie-lou would be a better reviewer for this code path

maggie-lou (approved):

PR #8009, opened 2024-12-04
Write to the action cache in the execution server

vanja-p:
This is done when the executor sends a completed action to PublishOperation. The write from the app is added at the same time that the executor write is removed, so we could have a period of time where we double write some entries, or don't write some entries, depending on whether the app or executor rolls out first.


enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -1007,31 +1007,44 @@ func (s *ExecutionServer) PublishOperation(stream repb.Execution_PublishOperatio
 			return err
 		}
 
+		if op.GetName() == "" {
+			return status.InvalidArgumentError("Operation.name must be non-empty")
+		}
```
tylerwilliams, line 1012: I trust you've checked and this is safe?

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -1007,31 +1007,44 @@ func (s *ExecutionServer) PublishOperation(stream repb.Execution_PublishOperatio
...
 		mu.Unlock()
 
 		log.CtxDebugf(ctx, "PublishOperation: stage: %s", stage)
 
-		var response *repb.ExecuteResponse
+		actionResourceName, err := digest.ParseUploadResourceName(taskID)
+		if err != nil {
+			return status.WrapErrorf(err, "Failed to parse taskID")
+		}
+		actionResourceName.ToProto().CacheType = rspb.CacheType_AC
```
tylerwilliams, line 1026: why do we want to mutate a field on a proto returned by ToProto()?

actionResourceNameProto := actionResourceName.ToProto()
actionResourceNameProto.CacheType = rspb.CacheType_AC

or if you're trying to mutation actionResourceName (the struct) directly, is there a public field or setter we could use?

tylerwilliams (approved):
LGTM but let's test the hell out of this

vanja-p:
> LGTM but let's test the hell out of this

What did you have in mind? Running it locally seems to work. I'll run it in dev for a bit as well.

enterprise/server/remote_execution/execution_server/execution_server.go
vanja-p: by the way, if taskID is empty, I would expect https://github.com/buildbuddy-io/buildbuddy/blob/d06497cd4e1aa94233a327df1d70c844192ae0ab/enterprise/server/remote_execution/execution_server/execution_server.go#L1096 to return an error, and it never does: https://console.cloud.google.com/logs/query;query=SEARCH%2528%22Unparsable%20resource%20name%22%2529;cursorTimestamp=2024-12-03T15:26:10.559419765Z;duration=P30D?project=flame-build&inv=1&invt=AbjUVw

bduffany (approved):

vanja-p:
I ran this in dev overnight and everything seems to be working. The executor stopped sending UpdateActionResult RPCs, and the app started sending more. The cache hit ratio is relatively stable. Running `bazel test enterprise/server/... --config=remote-dev` twice, with a bazel clean in between, resulted in the second build being fully remotely cached.

PR #8011, opened 2024-12-04
[RB] Fix an incorrect reference to remote branch vs local branch

maggie-lou:
I introduced a bug in this commit: https://github.com/buildbuddy-io/buildbuddy/pull/7989

If your local branch does not exist remotely, we want to fallback to using the master branch. In the previous incorrect version, we'd use the remote master head commit. If this commit didn't exist locally, there'd be errors when we tried to diff against it to generate the diff patchsets

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -356,7 +356,7 @@ func getBaseBranchAndCommit(remoteData string) (branch string, commit string, er
 	if currentBranchExistsRemotely {
 		branch = currentBranch
 
-		currentCommitHash, err := runGit("rev-parse", "HEAD")
+		currentCommitHash, err := getHeadCommitForLocalBranch("HEAD")
 		if err != nil {
 			return "", "", status.WrapError(err, "get current commit hash")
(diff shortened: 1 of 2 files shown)


cli/remotebazel/remotebazel.go
```diff
@@ -385,9 +385,9 @@ func getBaseBranchAndCommit(remoteData string) (branch string, commit string, er
 		}
 		branch = defaultBranch
 
-		defaultBranchCommitHash, err := getHeadCommitForRemoteBranch(remoteData, defaultBranch)
+		defaultBranchCommitHash, err := getHeadCommitForLocalBranch(defaultBranch)
```
sluongng, line 388: nit: above this a few lines we are running "git rev-parse HEAD" which could be changed to `getHeadCommitForLocalBranch("HEAD")`

sluongng (approved):
LGTM

cli/remotebazel/remotebazel.go
maggie-lou: Done

PR #8014, opened 2024-12-05
redis: respect rediss://

sluongng:
Effectively revert https://github.com/buildbuddy-io/buildbuddy/pull/7978
as user could already enable TLS by using "rediss://" url scheme.

This would not work in case of a Shared Redis setup using Ring Client,
so ensure that we use the first shard address to setup the Ring Client
options. If the first shard address uses "rediss://", TLS will be
enabled and if it uses "redis://", TLS will be disabled.

### DIFF
--- a/enterprise/server/util/redisutil/redisutil.go
+++ b/enterprise/server/util/redisutil/redisutil.go
@@ -90,19 +90,17 @@ func ShardsToOpts(shards []string, username, password string) *Opts {
 	}
 	var prevShardScheme string
 	for i, shard := range shards {
-		url, err := url.Parse(shard)
+		u, err := url.Parse(shard)
 		if err != nil {
-			log.Errorf("Failed to parse redis shard %q: %s", shard, err)
-			return nil
+			log.Errorf("Failed to parse redis shard %q: %s, ignoring", shard, err)
+			continue
 		}
-
 		if i > 0 {
-			if url.Scheme != prevShardScheme {
-				log.Errorf("All redis shards must use the same url scheme, but found %q and %q", prevShardScheme, url.Scheme)
-				return nil
+			if u.Scheme != prevShardScheme {
+				log.Fatalf("All redis shards must use the same url scheme, but found %q and %q", prevShardScheme, u.Scheme)
 			}
 		}
-		prevShardScheme = url.Scheme
+		prevShardScheme = u.Scheme
 	}
 	opt := TargetToOpts(shards[0])
 	opt.Addrs = shards
@@ -187,7 +185,11 @@ func (o *Opts) toRingOpts() (*redis.RingOptions, error) {
 		TLSConfig:          o.TLSConfig,
 	}
 	for i, addr := range o.Addrs {
-		opts.Addrs[fmt.Sprintf("shard%d", i)] = addr
+		opt, err := redis.ParseURL(addr)
+		if err != nil {
+			return nil, status.FailedPreconditionErrorf("invalid redis shard address %q: %s", addr, err)
+		}
+		opts.Addrs[fmt.Sprintf("shard%d", i)] = opt.Addr
 	}
 	return opts, nil
 }
--- a/enterprise/server/util/redisutil/redisutil.go
+++ b/enterprise/server/util/redisutil/redisutil.go
@@ -185,6 +185,10 @@ func (o *Opts) toRingOpts() (*redis.RingOptions, error) {
 		TLSConfig:          o.TLSConfig,
 	}
 	for i, addr := range o.Addrs {
+		if !strings.HasPrefix(addr, "redis://") || !strings.HasPrefix(addr, "rediss://") || !strings.HasPrefix(addr, "unix://") {
+			// Assume tcp if no scheme is provided.
+			addr = "redis://" + addr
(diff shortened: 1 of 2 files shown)


enterprise/server/util/redisutil/redisutil.go
```diff
@@ -68,43 +68,30 @@ func TargetToOptions(redisTarget string) *redis.Options {
...
-			MinVersion: tls.VersionTLS12,
-		}
-	}
-	return &Opts{
-		Addrs:     shards,
-		Username:  username,
-		Password:  password,
-		TLSConfig: tlsConfig,
-	}
+	opt := TargetToOpts(shards[0])
```
bduffany, line 90: nit: could we return an error here if the shards are using a mix of `rediss://` and `redis://` ?

sluongng: our current code to setup redis does not support returning an error 🤔 

We can log a warning return nil, in which case the non-sharded config will get picked up. If there is no default then there will be no redis client registered and we fail to startup in case of enterprise setup.

We can also do log.fatal directly here, which is what the error return gona do regardless as this failed during registration.

I am leaning toward just log a warning and returning nil, wdyt?

bduffany: If it's too big of a refactor to return an error then logging a warning seems fine as well

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -68,43 +69,46 @@ func TargetToOptions(redisTarget string) *redis.Options {
...
+		url, err := url.Parse(shard)
+		if err != nil {
+			log.Errorf("Failed to parse redis shard %q: %s", shard, err)
+			return nil
 		}
+
+		if i > 0 {
+			if url.Scheme != prevShardScheme {
+				log.Errorf("All redis shards must use the same url scheme, but found %q and %q", prevShardScheme, url.Scheme)
+				return nil
```
bduffany, line 102: returning `nil` seems like it will result in an NPE later on? should we just log and continue?

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -68,43 +69,46 @@ func TargetToOptions(redisTarget string) *redis.Options {
...
+		prevShardScheme = url.Scheme
 	}
-	return &Opts{
-		Addrs:     shards,
-		Username:  username,
-		Password:  password,
-		TLSConfig: tlsConfig,
-	}
+	opt := TargetToOpts(shards[0])
+	opt.Addrs = shards
```
bduffany, line 108: do we need to strip the scheme part of the URL for this `Addrs` list, or does the consuming function support schemes being present? asking because we aren't currently specifying schemes in our dev or prod configs.

sluongng: Good call.

I think it's best to add the stripping logic into `toRingOpts()` and use the official `redis.ParseURL` to better handle unix socket setups. Let me know if you have objection

enterprise/server/util/redisutil/redisutil.go
sluongng: Pushed a change to log + skip invalid url (failed `url.Parse`). But log fatal when the scheme is inconsistent between shards. This should be a sane "fail fast" default because the TLS connection gonna mess up regardless.

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -183,7 +185,15 @@ func (o *Opts) toRingOpts() (*redis.RingOptions, error) {
 		TLSConfig:          o.TLSConfig,
 	}
 	for i, addr := range o.Addrs {
-		opts.Addrs[fmt.Sprintf("shard%d", i)] = addr
+		if !strings.HasPrefix(addr, "redis://") || !strings.HasPrefix(addr, "rediss://") || !strings.HasPrefix(addr, "unix://") {
```
bduffany, line 188:
```suggestion
		if !strings.HasPrefix(addr, "redis://") && !strings.HasPrefix(addr, "rediss://") && !strings.HasPrefix(addr, "unix://") {
```

bduffany: or better yet, `!isRedisURI(addr)`

enterprise/server/util/redisutil/redisutil.go
```diff
@@ -68,43 +69,44 @@ func TargetToOptions(redisTarget string) *redis.Options {
...
+	var prevShardScheme string
+	for i, shard := range shards {
+		u, err := url.Parse(shard)
+		if err != nil {
+			log.Errorf("Failed to parse redis shard %q: %s, ignoring", shard, err)
+			continue
 		}
+		if i > 0 {
+			if u.Scheme != prevShardScheme {
+				log.Fatalf("All redis shards must use the same url scheme, but found %q and %q", prevShardScheme, u.Scheme)
```
bduffany, line 100: I'm not a fan of using `log.Fatal` in places other than `main()` - it's like `panic()` except worse, because `defer` statements won't run. This is only called in 3 places, could we have it return an error?

bduffany (approved):

PR #8017, opened 2024-12-05
ociruntime: fix blocking on unknown tar header

sluongng:
It seems like our "executor-docker-default" image contains quite a few
`/dev/*` entries in a layer on the image.

These are either block or character devices that don't do much on the
container itself and should be safe to ignore.

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -1,6 +1,7 @@
 package ociruntime
 
 import (
+	"archive/tar"
 	"bytes"
 	"context"
 	"crypto/rand"
@@ -8,7 +9,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"io/fs"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1369,7 +1369,7 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
 				return status.UnavailableErrorf("chown file: %s", err)
 			}
 		default:
-			return status.UnavailableErrorf("unsupported tar entry type %q", header.Typeflag)
+			log.CtxWarningf(ctx, "Ignoring unsupported tar header %q type %q in oci layer", header.Name, header.Typeflag)
```
bduffany, line 1372: nit: I think these warnings will be spammy. Would it work to explicitly skip character and block devices?

bduffany: or alternatively just log at info level for now.

bduffany, line 1372:
```suggestion
			log.CtxInfof(ctx, "Ignoring unsupported tar header %q type %q in oci layer", header.Name, header.Typeflag)
```

sluongng:
turning this to draft to do a bit more testing.

We reverted the original PR instead.

sluongng:
Replacing this PR with https://github.com/buildbuddy-io/buildbuddy/pull/8023

PR #8018, opened 2024-12-05
[RB] Fix race condition with fetching completed execution data

maggie-lou:
Fixes https://buildbuddy-corp.slack.com/archives/C07GMM2VBLY/p1733333761499139?thread_ts=1731949897.755029&cid=C07GMM2VBLY

The CLI sometimes reports the wrong exit code (0, even when the remote run failed). This is because there can be a slight delay between when the remote runner execution finishes, and when the execution is successfully updated in the DB with the final exit code. Make sure to check that the execution is complete before using the exit code.

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -777,6 +777,7 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 		return 1, status.UnavailableErrorf("could not connect to BuildBuddy remote bazel service %q: %s", opts.Server, err)
 	}
 	bbClient := bbspb.NewBuildBuddyServiceClient(conn)
+	execClient := repb.NewExecutionClient(conn)
 
 	reqOS := runtime.GOOS
 	if *execOs != "" {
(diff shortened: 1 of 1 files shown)


cli/remotebazel/remotebazel.go
```diff
@@ -912,14 +912,23 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 	})
 	eg.Go(func() error {
 		var err error
-		exRsp, err = bbClient.GetExecution(ctx, &espb.GetExecutionRequest{ExecutionLookup: &espb.ExecutionLookup{
-			InvocationId: iid,
-		}})
-		if err != nil {
-			return fmt.Errorf("could not retrieve ci_runner execution: %s", err)
+		for i := 0; i < 5; i++ {
```
bduffany, line 915: can we use `WaitExecution` instead of polling?

bduffany: (`rexec.NewRetryingStream()` can help here)

maggie-lou: I started out on that approach, but the code feels a lot more complicated to introduce a stream. At this point, we know the execution is complete (it has stopped publishing event logs), so we won't have to poll for hours for example. What do you think of just going with the simpler polling approach?

maggie-lou: Ack - updated

bduffany (approved):

PR #8023, opened 2024-12-06
ociruntime: reapply tar unpack

sluongng:
This change include a few small changes inside:

- Reapply "ociruntime: replace tar extraction subprocess with in-process (#7589)" (#8019)

- Add a basic image pull test and guard it with a separate, manual Bazel test target to help verifying common images.

- Fix issues with the previous implementation.
  Ensure that parent directory is created for file tar entries.
  Ignore unknown tar entry types (i.e. devices).

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/BUILD
+++ b/enterprise/server/remote_execution/containers/ociruntime/BUILD
@@ -52,6 +52,7 @@ go_test(
         "test.workload-isolation-type": "firecracker",
         "test.container-image": "docker://gcr.io/flame-public/net-tools@sha256:ac701954d2c522d0d2b5296323127cacaaf77627e69db848a8d6ecb53149d344",
         "test.EstimatedComputeUnits": "2",
+        "test.EstimatedFreeDiskBytes": "10GB",
     },
     # NOTE: if testing locally, use --test_sharding_strategy=disabled to work
     # around the networking package not supporting cross-process locks.
@@ -92,72 +93,6 @@ go_test(
         "//server/testutil/testtar",
         "//server/util/disk",
         "//server/util/networking",
-        "//server/util/proto",
-        "//server/util/status",
-        "//server/util/testing/flags",
-        "//server/util/uuid",
-        "@com_github_google_go_containerregistry//pkg/crane",
-        "@com_github_google_go_containerregistry//pkg/v1:pkg",
-        "@com_github_google_go_containerregistry//pkg/v1/mutate",
-        "@com_github_stretchr_testify//assert",
-        "@com_github_stretchr_testify//require",
-        "@io_bazel_rules_go//go/runfiles:go_default_library",
-    ],
-)
-
-# keep
-go_test(
-    name = "ociruntime_pull_test",
-    srcs = ["ociruntime_test.go"],
-    args = ["-test.run=TestPullImage"],
-    data = [
-        ":busybox",
-        ":crun",
-        "//enterprise/server/remote_execution/runner/testworker",
-        "@busybox",
-    ],
-    env = {
-        "BB_INTEGRATION": "true",
-    },
-    exec_properties = {
-        "test.workload-isolation-type": "firecracker",
-        "test.container-image": "docker://gcr.io/flame-public/net-tools@sha256:ac701954d2c522d0d2b5296323127cacaaf77627e69db848a8d6ecb53149d344",
-        "test.EstimatedComputeUnits": "2",
-        "test.EstimatedFreeDiskBytes": "10GB",
-    },
-    tags = [
-        "manual",
-    ],
-    target_compatible_with = [
-        "@platforms//os:linux",
-        # TODO: GitHub Actions arm runners are running into cgroup-related issues;
-        # fix and re-enable on arm64.
-        "@platforms//cpu:x86_64",
-    ],
-    x_defs = {
-        "crunRlocationpath": "$(rlocationpath :crun)",
-        "busyboxRlocationpath": "$(rlocationpath :busybox)",
-        "ociBusyboxRlocationpath": "$(rlocationpath @busybox)",
-        "testworkerRlocationpath": "$(rlocationpath //enterprise/server/remote_execution/runner/testworker)",
-    },
-    deps = [
-        ":ociruntime",
-        "//enterprise/server/remote_execution/container",
-        "//enterprise/server/remote_execution/persistentworker",
-        "//enterprise/server/remote_execution/platform",
-        "//enterprise/server/remote_execution/workspace",
-        "//enterprise/server/util/oci",
-        "//proto:remote_execution_go_proto",
-        "//proto:scheduler_go_proto",
-        "//proto:worker_go_proto",
-        "//server/interfaces",
-        "//server/testutil/testenv",
-        "//server/testutil/testfs",
-        "//server/testutil/testnetworking",
-        "//server/testutil/testregistry",
-        "//server/testutil/testshell",
-        "//server/testutil/testtar",
-        "//server/util/disk",
         "//server/util/proto",
         "//server/util/status",
         "//server/util/testing/flags",
--- a/enterprise/server/remote_execution/containers/ociruntime/BUILD
+++ b/enterprise/server/remote_execution/containers/ociruntime/BUILD
@@ -105,6 +105,72 @@ go_test(
     ],
 )
 
+# keep
+go_test(
+    name = "ociruntime_pull_test",
+    srcs = ["ociruntime_test.go"],
+    args = ["-test.run=TestPullImage"],
+    data = [
+        ":busybox",
+        ":crun",
+        "//enterprise/server/remote_execution/runner/testworker",
+        "@busybox",
+    ],
+    env = {
+        "BB_INTEGRATION": "true",
+    },
+    exec_properties = {
+        "test.workload-isolation-type": "firecracker",
+        "test.container-image": "docker://gcr.io/flame-public/net-tools@sha256:ac701954d2c522d0d2b5296323127cacaaf77627e69db848a8d6ecb53149d344",
+        "test.EstimatedComputeUnits": "2",
+        "test.EstimatedFreeDiskBytes": "10GB",
+    },
+    tags = [
+        "manual",
+    ],
+    target_compatible_with = [
+        "@platforms//os:linux",
+        # TODO: GitHub Actions arm runners are running into cgroup-related issues;
+        # fix and re-enable on arm64.
+        "@platforms//cpu:x86_64",
+    ],
+    x_defs = {
+        "crunRlocationpath": "$(rlocationpath :crun)",
+        "busyboxRlocationpath": "$(rlocationpath :busybox)",
+        "ociBusyboxRlocationpath": "$(rlocationpath @busybox)",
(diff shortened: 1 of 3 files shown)


sluongng:
```
--- PASS: TestPullImage (74.35s)
    --- PASS: TestPullImage/dockerhub_busybox (1.09s)
    --- PASS: TestPullImage/ghcr_nix (7.38s)
    --- PASS: TestPullImage/executor_image (17.39s)
    --- PASS: TestPullImage/executor_docker (24.34s)
    --- PASS: TestPullImage/workflow_2004 (14.00s)
    --- PASS: TestPullImage/workflow_2204 (10.16s)
PASS
```

Also tested some public images that our users used and did not find any issues.

enterprise/server/remote_execution/containers/ociruntime/BUILD
```diff
@@ -102,6 +102,72 @@ go_test(
     ],
 )
 
+# keep
```
bduffany, line 105: nit: we have a few of these types of targets elsewhere but they are a bit of a hassle to maintain because gazelle doesn't handle them properly. I think it would be fine to use a flag (instead of env var) to guard the test, and then people could run it by passing the flag to the "normal" test target, like `bazel test enterprise/server/remote_execution/containers/ociruntime:ociruntime_test --config=remote --test_arg=-test_pull=true` (optionally adding `--test_filter=TestPullImage` to run just that test case)

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1293,58 +1294,86 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
...
+			break
+		}
 		if err != nil {
-			return err
+			return status.UnavailableErrorf("download and extract layer tarball: %s", err)
 		}
 
-		if !entry.Type().IsRegular() {
-			return nil
+		if slices.Contains(strings.Split(header.Name, string(os.PathSeparator)), "..") {
```
bduffany, line 1307: nit: I think `filepath.SplitList()` would work here (it's different from `filepath.Split()` that I incorrectly suggested in the last PR)

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1293,58 +1294,86 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
...
-	})
-	if err != nil {
-		return status.UnavailableErrorf("walk layer dir: %s", err)
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err := os.MkdirAll(target, os.FileMode(header.Mode)); err != nil {
+				return status.UnavailableErrorf("create directory: %s", err)
+			}
+		case tar.TypeReg:
+			if err := os.MkdirAll(dir, os.ModePerm); err != nil {
```
bduffany, line 1340: do we need to make the parent dir for TypeSymlink and TypeLink as well?

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1293,58 +1294,86 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
...
+			f.Close()
+		case tar.TypeSymlink:
+			if err := os.Symlink(header.Linkname, target); err != nil {
+				return status.UnavailableErrorf("create symlink: %s", err)
+			}
+			if err := os.Lchown(target, header.Uid, header.Gid); err != nil {
+				return status.UnavailableErrorf("chown link: %s", err)
+			}
+		case tar.TypeLink:
+			if slices.Contains(strings.Split(header.Linkname, string(os.PathSeparator)), "..") {
```
fmeum, line 1364: Could you add a comment to explain why we do this for hardlinks but not symlinks?

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1293,58 +1294,86 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
...
+				return status.UnavailableErrorf("tar entry is not clean: %q", header.Name)
+			}
+			source := filepath.Join(tempUnpackDir, filepath.Clean(header.Linkname))
+			if err := os.Link(source, target); err != nil {
+				return status.UnavailableErrorf("create hard link: %s", err)
+			}
+			if err := os.Chown(target, header.Uid, header.Gid); err != nil {
```
fmeum, line 1371: As far as I understand hardlinks, this would change the metadata of the target, which is shared between all hardlinks. Since we ensure that the target is created by this tar and thus could have had a mode set, should we skip it here to prevent flip-flopping? I don't know how tar handles this.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1293,58 +1294,86 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
...
+			}
+			if err := os.Lchown(target, header.Uid, header.Gid); err != nil {
+				return status.UnavailableErrorf("chown link: %s", err)
+			}
+		case tar.TypeLink:
+			if slices.Contains(strings.Split(header.Linkname, string(os.PathSeparator)), "..") {
+				return status.UnavailableErrorf("tar entry is not clean: %q", header.Name)
+			}
+			source := filepath.Join(tempUnpackDir, filepath.Clean(header.Linkname))
+			if err := os.Link(source, target); err != nil {
```
fmeum, line 1368: Hardlinks can only be created if the target file exists. Depending on the tar entry order, a hardlink may come before its target. We may want to collect these into a list and create them at the end.

enterprise/server/remote_execution/containers/ociruntime/BUILD
sluongng: will fix

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: oh good call on SpliList(). It does exactly what we are doing but cleaner

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: yeah, I can add the create parent dir to all known cases

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: Will do.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: This is a deep rabbit hole to go down to.
I realize that you can create a hard link from an existing hard link
```
touch a
link a b
link b c
```

that means we probably need a recursive implementation similar to https://github.com/google/go-containerregistry/blob/6bce25ecf0297c1aa9072bc665b5cf58d53e1c54/pkg/v1/tarball/image.go#L221 to be safe.


However, I tried to find reference implementations in podman/docker code base and I don't see any special queuing of hard link https://github.com/containers/storage/blob/664fe5d9b95004e1be3eee004d56a1715c8ca790/pkg/archive/archive.go#L707-L729
https://github.com/moby/moby/blob/9633556bef3eb20dfe888903660c3df89a73605b/pkg/archive/archive.go#L726-L735
So I suspect it's relatively rare and probably safe to ignore for now? 🤔

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: As I work on the hardlink change, I also realize that symlinks are not only evaled inside the container, if a hardlink is targeting the symlink, which points to a host-level file, it will result in a breakout 🤔 

Let me study how podman are docker are doing this and mirror their approach.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: will remove this

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: oh shoot, `SplitList()` actually splits on `":"` so I was wrong again :skull:

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: When creating a hardlink to a symlink, the symlink shouldn't be followed, so I don't think this is an issue? Creating a hardlink to a symlink should just create that same symlink at a new path (the hardlink destination), not the file it points to.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
bduffany: I agree that it's safe to ignore

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1294,58 +1299,101 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
...
+			// Ensure that parent dir exists
+			if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+				return status.UnavailableErrorf("create directory: %s", err)
 			}
-			if err := os.Remove(path); err != nil {
-				return fmt.Errorf("remove directory whiteout marker: %w", err)
-			}
-			return nil
+		} else {
+			// TODO(sluongng): switch this to debug level?
```
bduffany, line 1333: nit: this TODO doesn't seem actionable and could be removed

```suggestion
```

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -1294,58 +1299,101 @@ func downloadLayer(ctx context.Context, layer ctr.Layer, destDir string) error {
...
-		base := filepath.Base(path)
-		dir := filepath.Dir(path)
-		const whiteoutPrefix = ".wh."
+		// filepath.Join applies filepath.Clean to all arguments
+		file := filepath.Join(tempUnpackDir, header.Name)
+		base := filepath.Base(file)
+		dir := filepath.Dir(file)
+		if !strings.HasPrefix(file, tempUnpackDir) {
+			return status.UnavailableErrorf("breakout attempt detected with tar entry: %q", header.Name)
+		}
```
bduffany, line 1322: do we need this check? it seems redundant with the check on line 1312.

bduffany (approved):

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: Removed.

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
sluongng: Removed.

bduffany (approved):

PR #8027, opened 2024-12-09
Don't report symlinks in output directory separately with `output_paths`

fmeum:
A symlink in an output directory should not be reported as an `output_{,directory,file}_symlink` unless it has been explicitly requested via `output_paths`. This matches the logic for REAPI < v2.1 using `output_directories` and `output_files`.

This fixes a failure in Bazel with `(Exit 34): /build/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/foo/... (File exists)`.

### DIFF
--- a/server/cache/dirtools/dirtools.go
+++ b/server/cache/dirtools/dirtools.go
@@ -399,34 +399,35 @@ func handleSymlink(dirHelper *DirHelper, rootDir string, cmd *repb.Command, acti
 	if len(cmd.OutputPaths) > 0 && len(cmd.OutputFiles) == 0 && len(cmd.OutputDirectories) == 0 {
 		// REAPI >= v2.1
 		for _, expectedPath := range cmd.OutputPaths {
-			if symlink.Path == expectedPath {
-				actionResult.OutputSymlinks = append(actionResult.OutputSymlinks, symlink)
-				// REAPI specification:
-				//   Servers that wish to be compatible with v2.0 API should still
-				//   populate `output_file_symlinks` and `output_directory_symlinks`
-				//   in addition to `output_symlinks`.
-				//
-				// TODO(sluongng): Since v6.0.0, all the output directories are included in
-				// Action.input_root_digest. So we should be able to save a `stat()` call by
-				// checking wherether the symlink target was included as a directory inside
-				// input root or not.
-				//
-				// Reference:
-				//   https://github.com/bazelbuild/bazel/commit/4310aeb36c134e5fc61ed5cdfdf683f3e95f19b7
-				symlinkInfo, err := os.Stat(fqfn)
-				if err != nil {
-					// When encounter a dangling symlink, skip setting it to legacy fields.
-					// TODO(sluongng): do we care to log this?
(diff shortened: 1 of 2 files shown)


server/cache/dirtools/dirtools.go
```diff
@@ -398,32 +398,36 @@ func handleSymlink(dirHelper *DirHelper, rootDir string, cmd *repb.Command, acti
...
-			// When encounter a dangling symlink, skip setting it to legacy fields.
-			// TODO(sluongng): do we care to log this?
-			log.Warningf("Could not find symlink %q's target: %s, skip legacy fields", fqfn, err)
-			return nil
-		}
-		if symlinkInfo.IsDir() {
-			actionResult.OutputDirectorySymlinks = append(actionResult.OutputDirectorySymlinks, symlink)
-			return nil
+		for _, expectedPath := range cmd.OutputPaths {
+			if symlink.Path == expectedPath {
```
sluongng, line 402: I thought this was already covered by the `dirhelper.ShouldUploadFile(fqfn)` call above. 🤔 

Nit: this if clause could be reversed with `continue` to reduce nested code

server/cache/dirtools/dirtools.go
```diff
@@ -398,32 +398,37 @@ func handleSymlink(dirHelper *DirHelper, rootDir string, cmd *repb.Command, acti
...
-			log.Warningf("Could not find symlink %q's target: %s, skip legacy fields", fqfn, err)
-			return nil
-		}
-		if symlinkInfo.IsDir() {
-			actionResult.OutputDirectorySymlinks = append(actionResult.OutputDirectorySymlinks, symlink)
-			return nil
+		for _, expectedPath := range cmd.OutputPaths {
+			if symlink.Path != expectedPath {
+				continue
+			}
```
sluongng, line 404: discussed on Slack about this: we can replace this with

```go
	if !dirHelper.IsOutputPath(fqfn) {
		return nil
	}
```

server/cache/dirtools/dirtools.go
```diff
@@ -420,10 +423,9 @@ func handleSymlink(dirHelper *DirHelper, rootDir string, cmd *repb.Command, acti
 		}
 		if symlinkInfo.IsDir() {
 			actionResult.OutputDirectorySymlinks = append(actionResult.OutputDirectorySymlinks, symlink)
-			return nil
+		} else {
+			actionResult.OutputFileSymlinks = append(actionResult.OutputFileSymlinks, symlink)
 		}
-
-		actionResult.OutputFileSymlinks = append(actionResult.OutputFileSymlinks, symlink)
```
sluongng, line 426: nit: we probably don't need this style change but I don't mind leaving it in

sluongng (approved):
LGTM

server/cache/dirtools/dirtools.go
fmeum: It's not a functional change, but I think the symmetric form makes it more obvious that exactly one of these is set.

PR #8031, opened 2024-12-09
Increase gRPC connection pool size by 50%

tylerwilliams:


tylerwilliams (commented):
not for this CL, but it would be nice if these were added dynamically so we didn't pay the cost of the connections until we needed them.

vadimberezniker (approved):

vadimberezniker:
> not for this CL, but it would be nice if these were added dynamically so we didn't pay the cost of the connections until we needed them.

I don't think they're expensive. It might be simpler to scale the pool size proportionally to GOMAXPROCS.

PR #8035, opened 2024-12-10
Update ci_runner bundled bazelisk version

bduffany:
Confirmed that `v1.19.0` respects .bazelversion when MODULE.bazel is present, and that `v1.17.0` does not.

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -6518,28 +6518,28 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
         urls = ["https://github.com/bazelbuild/bazel/releases/download/7.1.0/bazel-7.1.0-linux-x86_64"],
     )
     http_file(
-        name = "io_bazel_bazelisk-1.17.0-darwin-amd64",
+        name = "com_github_bazelbuild_bazelisk-bazelisk-darwin-amd64",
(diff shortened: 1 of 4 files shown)


deps.bzl
```diff
@@ -6518,28 +6518,28 @@ def install_static_dependencies(workspace_name = "buildbuddy"):
         urls = ["https://github.com/bazelbuild/bazel/releases/download/7.1.0/bazel-7.1.0-linux-x86_64"],
     )
     http_file(
-        name = "io_bazel_bazelisk-1.17.0-darwin-amd64",
+        name = "com_github_bazelbuild_bazelisk-bazelisk-darwin-amd64",
+        urls = ["https://github.com/bazelbuild/bazelisk/releases/download/v1.19.0/bazelisk-darwin-amd64"],
```
siggisim, line 6522: Any particular reason not to go with the latest release?

siggisim (approved):

deps.bzl
bduffany: This brings us in line with the version that we already have in go.mod (which is used for the CLI) - I could update both versions to the latest release but it felt a little less risky to do a small version bump now just to get the fix commit, then maybe later we can do a bigger version jump

PR #8036, opened 2024-12-10
Log a bunch more fields to clickhouse Executions

vanja-p:
The code is a bit clunky because I have to put a bunch of data in `ExecutionAuxiliaryMetadata`, then remove it for most of the code in `execution_server.PublishOperation`, but keep it around so I can pass it to `execution_server.recordExecution`. Some alternatives I considered:

1. Read this data from Redis. This still can't get everything, like `isolation_type`.
2. Pass the data in the gRPC context. Not a fan of this as an API, since it's very hidden.
3. Use the `Operation.metadata.partial_execution_metadata` field instead of `Operation.response.result.execution_metadata`. Then the executor would be passing metadata in 2 different ways, depending on what it thought the app was doing to do with the data. Doesn't seem great.

I wrote some [notes for myself](https://docs.google.com/document/d/14pCbYj9ERv4G_8lIngUQcN-9XqDyRRgnRH3sUhOAQEg/edit?tab=t.0) about these tradeoffs.

## Estimated disk write increase

I queried clickhouse system tables 2 times, separated by about 24 hours, to find the [daily write throughput for each column in the `Executions` table](https://docs.google.com/spreadsheets/d/13UGAQ79BYCfnytzuWUyvDa51OZra9T98b-I69jl-ZVw/edit?gid=0#gid=0). Related findings:

- We write about 8GB per day, which is about 3TB per year.
- Estimated task size fields write about 1.5% of the total. I'm adding 10 of these fields.
- Boolean fields like `success` write about 0.1% of the total. I'm adding one of these (skip_cache_check).
- Small int32 fields like `stage` write about 0.01% of the total. I'm adding one of these (execution_priority).
- Low cardinality, small string fields like `branch_name` write about 0.9% of the total. I'm adding one of these (isolation_type).

So in total, I think this change would increase write throughput by 16%, with almost all of that coming from the task size fields.

## Other fields I would maybe like to add:
- requested timeout and effective timeout
- more info about the platform, like OS and architecture
- info about the executor pool


server/util/clickhouse/schema/schema.go
```diff
@@ -196,8 +206,12 @@ type Execution struct {
...
 	ExitCode   int32
 
-	CachedResult bool
-	DoNotCache   bool
+	CachedResult    bool
+	DoNotCache      bool
+	SkipCacheLookup bool
+
+	ExecutionPriority int32
+	IsolationType     string
```
bduffany, line 214: I think this would be a great use case for ClickHouse's `LowCardinality(String)` column type - it builds a dictionary to represent the column values as an int, while still allowing the column to be queried as a human readable string. Though idk how to annotate this field so that gorm applies the `LowCardinality` option. (Also, we'd want to check for a known set of values when storing the field, to enforce the low cardinality.)

server/util/clickhouse/schema/schema.go
```diff
@@ -179,8 +179,18 @@ type Execution struct {
...
 
 	// Task sizing
-	EstimatedMemoryBytes int64
-	EstimatedMilliCPU    int64
+	EstimatedMemoryBytes   int64
+	EstimatedMilliCPU      int64
+	EstimatedFreeDiskBytes int64
+	RequestedMemoryBytes   int64
+	RequestedMilliCPU      int64
+	RequestedFreeDiskBytes int64
```
bduffany, line 187: do we store enough information to reconstruct whether these came from `EstimatedComputeUnits` or from `EstimatedCPU` / `EstimatedMemory`? At some point we probably want to simplify EstimatedCPU/EstimatedMemory so that they are actually respected (instead of only respected when EstimatedComputeUnits is set or if the workload-isolation-type is firecracker), and knowing what those values are set to would help us understand the impact of that change

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -199,15 +199,28 @@ func (s *Executor) ExecuteTaskAndStreamResults(ctx context.Context, st *repb.Sch
...
 	taskID := task.GetExecutionId()
 	adInstanceDigest := digest.NewResourceName(req.GetActionDigest(), req.GetInstanceName(), rspb.CacheType_AC, req.GetDigestFunction())
 	digestFunction := adInstanceDigest.GetDigestFunction()
 	task.ExecuteRequest.DigestFunction = digestFunction
 	acClient := s.env.GetActionCacheClient()
 
-	stateChangeFn := operation.GetStateChangeFunc(stream, taskID, adInstanceDigest)
+	auxMetadata := &espb.ExecutionAuxiliaryMetadata{
+		PlatformOverrides:  task.GetPlatformOverrides(),
+		ExecutionTask:      task,
```
bduffany, line 212: do we use the ExecutionTask for anything? seems like SchedulingMetadata has everything we need?

bduffany (commented):
The added fields LGTM. idk if we need to store Action digest / Command digest because those can be reconstructed from the execution_id. Same with digest function. I think it wouldn't hurt to add timeout especially since it will probably be highly compressible. The effective timeout is computed on the executor (we apply a default if the Action.timeout is unset, and also read the `default-timeout` platform property if Action.timeout is unset - we added this as a convenience because bazel doesn't have a way to set Action.timeout for non-test actions).

server/util/clickhouse/schema/schema.go
vanja-p: I was considering that, but I also couldn't figure out how to use it with gorm. I'll take another look.

enterprise/server/remote_execution/executor/executor.go
vanja-p: I'm using it to get at `execute_request.skip_cache_lookup` and `execute_request.execution_policy.priority`. Maybe I should send the ExecuteRequest instead of the ExecutionTask, since we already have access to most of its other fields:
- action and command are fetched from CAS
- execution_id is the operation name
- invocation_id and request_metadata come in the RPC context.
- platform_overrides and queued_timestamp are in the result metadata.
- jwt is the only field I'm not sure where to get.

What do you think?

vanja-p:
> idk if we need to store Action digest / Command digest because those can be reconstructed from the execution_id. Same with digest function.

It does make queries much easier if you don't have to parse string fields, but I'll hold off on logging these fields for now.

> I think it wouldn't hurt to add timeout especially since it will probably be highly compressible. The effective timeout is computed on the executor (we apply a default if the Action.timeout is unset, and also read the `default-timeout` platform property if Action.timeout is unset - we added this as a convenience because bazel doesn't have a way to set Action.timeout for non-test actions).

Then I guess we should probably store both the requested and effective timeout.

enterprise/server/remote_execution/executor/executor.go
bduffany: setting the ExecuteRequest sounds good to me

server/util/clickhouse/schema/schema.go
vanja-p: This isn't part of task size, so I got it by calling platform.ParseProperties.

bduffany: What I'm getting at is that I'm not sure if we can use the current DB columns to differentiate between (A) the user set EstimatedComputeUnits, or (B) the user set only EstimatedCPU and EstimatedMemoryBytes? (I think the answer is no but maybe I'm missing something.)

Here is the context for why I think this is interesting data: we have this logic to skip using the measured task size if either (A) `EstimatedComputeUnits` is set: https://github.com/buildbuddy-io/buildbuddy/blob/2348120722abe758bcdaecf0a2016790a149b739/enterprise/server/tasksize/tasksize.go#L154-L157 or if (B) the isolation type is firecracker: https://github.com/buildbuddy-io/buildbuddy/blob/2348120722abe758bcdaecf0a2016790a149b739/enterprise/server/tasksize/tasksize.go#L165-L167

Case (A) is really unfortunate because it's confusing - I think the user expectation would be that if they set EstimatedComputeUnits _or_ EstimatedCPU _or_ EstimatedMemoryBytes then those should take precedence over the measured task size. We even document these as "overrides" in our public docs (https://www.buildbuddy.io/docs/rbe-platforms#runner-resource-allocation), but that documentation actually isn't correct. `EstimatedComputeUnits` are overrides, but `EstimatedCPU` and `EstimatedMemoryBytes` are effectively just "hints" (except in the firecracker case).

So, to see if we can untangle ourselves from this confusing situation, I wonder if we'll have some way to query whether there are any users that are _relying_ on the current behavior. If no users are relying on it, then we can update the actual behavior to match the docs, without breaking anyone.

vanja-p: Sorry, I made a change but forgot to push it. Take another look. What I'm doing now is getting the requested sized directly from the action/command/platform, and including a separate field for compute units.

server/util/clickhouse/schema/schema.go
vanja-p: Turns out that it's pretty simple. I couldn't find any clear documentation, but gorm supports specifying the type as a [field tag](https://gorm.io/docs/models.html#Fields-Tags), and as long as you specify a clickhouse compatible type, the clickhouse gorm driver respects this. I verified it by running locally and querying the system schema for the field type.

vanja-p: As far as enforcing low cardinality, I don't think we should fail execution when we see a new isolation type, so I think we're left with a few options:

1. Keep track of the observed isolation types in the app, and log or report a metric of the count.
2. Monitor an existing metric, like [buildbuddy_remote_execution_count](https://github.com/buildbuddy-io/buildbuddy/blob/1ff042e870fdfae3b32eefc148d2b8bb98f278b0/server/metrics/metrics.go#L823-L831), which has an isolation type label, and alert on too many types.
3. Query the clickhouse database periodically and somehow report if the cardinality is too high. I'm not sure of the best way to do this, but I think clickhouse might have a built in way to turn queries into metrics.

Regardless, I don't think we have to have this enforcement right away.

vanja-p:
I added tests so this is ready for review.

enterprise/server/remote_execution/execution_server/execution_server.go
```diff
@@ -1020,24 +1039,68 @@ func (s *ExecutionServer) PublishOperation(stream repb.Execution_PublishOperatio
...
+		var actionRN *digest.ResourceName
+
 		if stage == repb.ExecutionStage_COMPLETED {
 			response = operation.ExtractExecuteResponse(op)
-		}
-		if response != nil { // The execution completed
-			arn, err := digest.ParseUploadResourceName(taskID)
+
+			// Remove fields that we don't want to write anywhere other than
+			// logs, and save them for later.
```
bduffany, line 1054: It feels a little cumbersome to have to clear out each of these fields - would it make sense to just clear out all of the auxiliary metadata before storing in the action cache? (I think it would be helpful to keep the isolation_type and scheduling_metadata etc. in the cached execute response, so that we can show these in the UI. e.g. it'd be nice to be able to show why we estimated the task size a certain way)

server/util/clickhouse/schema/schema.go
```diff
@@ -179,8 +179,19 @@ type Execution struct {
...
+	EstimatedMilliCPU      int64
+	EstimatedFreeDiskBytes int64
+	RequestedComputeUnits  float64
+	RequestedMemoryBytes   int64
+	RequestedMilliCPU      int64
+	RequestedFreeDiskBytes int64
+	MeasuredMemoryBytes    int64
+	MeasuredMilliCPU       int64
+	MeasuredFreeDiskBytes  int64
+	PredictedMemoryBytes   int64
```
bduffany, line 192: I feel like these names might be somewhat confusing (was the measurement from the execution of this task, or is this the measurement from a previous execution). wdyt about naming these with `LastMeasured` or `PreviousMeasured` as the prefix?

enterprise/server/remote_execution/execution_server/execution_server.go
vanja-p: I think that makes sense. Another thing this code does is handle cases where there is more than one ExecutionAuxiliaryMetadata, but maybe I shouldn't worry about that.

Are we concerned about sending any of existing (or future) fields in SchedulingMetadata to bazel via the pubsub send?

server/util/clickhouse/schema/schema.go
vanja-p: yep, I used PreviousMeasured

enterprise/server/remote_execution/execution_server/execution_server.go
bduffany: > Are we concerned about sending any of existing (or future) fields in SchedulingMetadata to bazel via the pubsub send?

I think it's fine to clear out the metadata both before sending to bazel and before storing in the action cache if that's what you mean? (but not before cacheExecuteResponse)

vanja-p: Cool, that's what I did.

server/util/clickhouse/schema/schema.go
bduffany: another option is to have a hard-coded list of isolation types (`platform.go` is probably the most convenient place for it) and then coerce any values not in that list to an empty string or the string "unknown"

vanja-p: We chatted and agreed to not use LowCardinality for the user requested isolation type. It should usually be empty, and it might be useful to analyze values which don't belong to a known set.

For the effective isolation type, we can mostly trust the executor, but we decided its better to force the value to belong to a known set, so clickhouse query performance doesn't become worse if we make a mistake.

PR #8041, opened 2024-12-10
[FC] Add new version ID to logs after invalidating a snapshot

maggie-lou:


siggisim (approved):
Should we put group id in here too?

bduffany:
> Should we put group id in here too?

Tyler recently added group_id to all logs when `log.Ctx*` is used https://github.com/buildbuddy-io/buildbuddy/pull/8032

PR #8042, opened 2024-12-10
Remove linux go build constraint for uffd

maggie-lou:
We don't need the go build constraint because the BUILD file contains a `target_compatible_with` platform. This makes it easier to edit the file with an IDE, even on non-linux platforms

This was originally added here https://github.com/buildbuddy-io/buildbuddy/pull/7965, then reverted due to the separate changes in that PR

### DIFF
--- a/enterprise/server/remote_execution/uffd/BUILD
+++ b/enterprise/server/remote_execution/uffd/BUILD
@@ -9,17 +9,14 @@ go_library(
         "@platforms//os:linux",
     ],
     visibility = ["//visibility:public"],
-    deps = select({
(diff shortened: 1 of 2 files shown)


bduffany:
out of curiosity what problems are you running into on the IDE? at least with gopls / vscode, these files are handled properly

bduffany (approved):

sluongng (approved):

maggie-lou:
> out of curiosity what problems are you running into on the IDE? at least with gopls / vscode, these files are handled properly

Just nice-to-have features like code completion and following function names to their implementation etc (I use intellij)

PR #8048, opened 2024-12-11
Remove need to call umoci, use crane tools

tylerwilliams:
OLD     --- PASS: TestOciconv/image=mirror.gcr.io/ubuntu:22.04 (4.10s)
NEW    --- PASS: TestOciconv/image=mirror.gcr.io/ubuntu:22.04 (2.57s)

Images are the same size (125MB) and contain the same number of files and dirs.


bduffany (approved):
:fire:

PR #8051, opened 2024-12-12
Fetch ExecuteResult when WaitExecution is done

vanja-p:
This lets us get all the metadata that is stored in the ExecutedResult, but isn't sent on wait operation.

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -166,7 +166,7 @@ export default class InvocationActionCardComponent extends React.Component<Props
           this.setState({ actionResult: operation.response.result });
         }
         if (operation.done) {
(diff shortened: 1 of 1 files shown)


app/invocation/invocation_action_card.tsx
```diff
@@ -165,6 +165,9 @@ export default class InvocationActionCardComponent extends React.Component<Props
         if (operation.response?.result) {
           this.setState({ actionResult: operation.response.result });
         }
+        if (operation.done) {
```
bduffany, line 168:
```suggestion
        // Fetch the full response from cache, since it contains
        // some additional metadata not sent on the stream.
        if (operation.done) {
```

bduffany (approved):

PR #8057, opened 2024-12-13
Use `flags-as-proto` to determine command line schema

tempoz:
Need to fix how we import the proto in bazelmod


WORKSPACE.bzlmod
```diff
@@ -196,6 +196,23 @@ load("@io_bazel_rules_webtesting//web:go_repositories.bzl", web_test_go_reposito
...
 
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
+
+go_repository(
+    name = "com_github_bazelbuild_bazel",
+    build_file_generation = "clean",
+    build_file_proto_mode = "package",
+    importpath = "github.com/bazelbuild/bazel",
+    tag = "7.4.1",
+)
```
bduffany, line 207: we could probably just copy the proto file into `proto/` like we do for the other protos that we import from bazel

tempoz: Do we want to not depend on bazel? It's only a 24 MB download, and I'm not in love with copying deps into our repo; it makes version management a huge mess and makes upgrading cumbersome.

bduffany: It's more about consistency, since we're already copying the other protos from Bazel. These protos don't change that frequently, and protos are backwards-compatible, so I don't think version management is that messy.

tempoz: kk, then that's what I'll do for now; I'll file an issue for discussion of maybe using (all of) the bazel protos via dependency instead of copying them.

bduffany (commented):
nit: could you update parser_test.go? it currently hard-codes some static `bazel help` file contents (to avoid having to invoke bazel during the test) - maybe we could run `bazel flags-as-proto` and save those contents to the repo, and write a new test using those contents instead of the help contents

cli/parser/parser.go
```diff
@@ -246,6 +249,18 @@ type Option struct {
...
+	if err != nil {
+		return nil
+	}
+	m := &bfpb.FlagCollection{}
+	if err := proto.Unmarshal(b, m); err != nil {
+		return nil
+	}
+	return nil
+}
+
```
bduffany, line 263: I think this is unused?

```suggestion
```

tempoz: Oh, yeah, that was from when I was just trying to get it to import correctly; good catch!

.gitattributes
```diff
@@ -0,0 +1,3 @@
+* text=auto
+
+*.b64 binary
```
bduffany, line 3: The b64 file is a text-format (ASCII) representation of binary content, so I think this change should be reverted

bduffany (approved):
LGTM modulo reverting `.gitattributes` change

.gitattributes
tempoz: This setting tells git not to add new lines to the file or try to display diffs of it as if it is a normal file (which it can't do effectively, as it's all on one line anyway).

See the **Identifying Binary Files** section of https://git-scm.com/book/ms/v2/Customizing-Git-Git-Attributes , which states:

> Some files look like text files but for all intents and purposes are to be treated as binary data.

and then goes on to describe an example of a text file that is not mergeable and where diffs aren't "generally helpful, advising that in these cases the files in question should be labeled `binary` in `.gitattributes`

It later describes how to get `git` to display diffs of "binary" files that have been converted to text, which we could do, but that requires a `git config` setting, which would have to be set for every user who clones the repo, so it's not really practical.

bduffany: OK, seems reasonable - thanks for explaining! maybe worth adding a comment? otherwise LGTM

tempoz: comment is a good idea ^-^

vanja-p:
I'm pretty sure this broke flags like `remote_download_minimal`. Builds with `--config=remote` fail with `failed to evaluate bazelrc configuration: expected value after --remote_download_minimal`.

PR #8067, opened 2024-12-16
Various VFS client side changes

vadimberezniker:
- Use log.Ctx for verbose logging instead of adding a custom prefix which makes the logs easier to read.
 - Add optional per-file FUSE operation stats.
 - Add support for Passthrough.
 - Add flags for controlling various debugging options.

### DIFF
--- a/enterprise/server/remote_execution/vfs/vfs_unix.go
+++ b/enterprise/server/remote_execution/vfs/vfs_unix.go
@@ -131,7 +131,6 @@ func (vfs *VFS) Mount() error {
 			// Disable fallback to fusermount as well, since it can cause
 			// deadlocks. See https://github.com/hanwen/go-fuse/issues/506
 			DirectMountStrict: true,
-			DirectMountFlags:  syscall.MS_NOSUID | syscall.MS_NODEV | syscall.MS_NOATIME,
 			FsName:            "bbvfs",
 			MaxWrite:          fuse.MAX_KERNEL_WRITE,
 			EnableLocks:       true,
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/vfs/vfs_unix.go
```diff
@@ -244,9 +299,26 @@ func (vfs *VFS) Unmount() error {
...
+	}
+
+	vfs.mu.Lock()
+	defer vfs.mu.Unlock()
+	stats, ok := vfs.perFileStats[path]
+	if !ok {
+		stats = make(perOpStats)
+		vfs.perFileStats[path] = stats
+	}
+	stats[op]++
```
tylerwilliams, line 319: should we do this accounting only if it's enabled so we avoid the allocations otherwise?

tylerwilliams: same in startOp

tylerwilliams: i don't mean this to feel nitpicky -- i think we suspect (based on profiles jim sent) that we're doing a lot of allocations during setup of each input root and then the GC is struggling a bit to delete everything after we're done with it.

tylerwilliams (approved):
hot

enterprise/server/remote_execution/vfs/vfs_unix.go
vadimberezniker: `logFUSEPerFileStats` is not enabled by default and we skip this code when it's off...

tylerwilliams: 🤦🏻‍♂️

startOp still applies but maybe the cost there is nbd?

vadimberezniker: I have run profiles during testing and have not seen this come up at all.

PR #8068, opened 2024-12-16
Don't use COW for workspaces or store them in snapshots

bduffany:
It's not necessary to use COW+VBD for the workspace drive or store it as part of the VM snapshot, because we never wind up reading from the snapshotted workspace - instead, we wind up just throwing it away after unpausing, since we create a new workspace ext4 image from the action workspace on the host for every action.

So, instead of snapshotting the workspace, we can stub out the workspace drive with an empty file (`empty.bin` relative to the chroot) just before saving the snapshot, so that the snapshotted state does not contain any references to the workspace. Firecracker will expect this file to exist when resuming the snapshot, so we just need to manually create this file before resuming.

In addition, there is some extra related cleanup in this PR: when initially creating the VM, we currently start out the workspace drive as an empty ext4 image. The filesystem is empty but the block device contents contain some bytes for the basic ext4 metadata. The goinit binary then mounts this filesystem blindly when the VM initially boots. This was done to ensure that the "setup workspace" routine could always unmount, swap, and re-mount. However, this is no longer needed. Instead, we now start out the workspace image as an empty (zero-length) _file_ (not an empty ext4 FS), and don't mount it. This lets us use the same logic every time for attaching a workspace, regardless of whether we're creating a new VM, or resuming one from snapshot. In both cases, the VM starts out with `empty.bin` backing the workspace drive, which is not mounted. Then, calling `createAndAttachWorkspace` creates an ext4 image from the host workspace, patches the VM drive contents to reference this new ext4 file, then mounts the filesystem.

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -111,7 +111,7 @@ const (
 	//
 	// NOTE: this is part of the snapshot cache key, so bumping this version
 	// will make existing cached snapshots unusable.
-	GuestAPIVersion = "13" // TODO: next time we bump this, fully clean up CgroupV2Only
+	GuestAPIVersion = "14"
 
 	// How long to wait when dialing the vmexec server inside the VM.
 	vSocketDialTimeout = 60 * time.Second
@@ -154,6 +154,9 @@ const (
 	// Example:
 	//	"{chrootPath}/workspacefs.vbd/file"
 	vbdMountDirSuffix = ".vbd"
+
+	// Name of the empty file used as a placeholder for the workspace drive.
+	emptyFileName = "empty.bin"
 
 	// The workspacefs image name and drive ID.
 	workspaceFSName  = "workspacefs.ext4"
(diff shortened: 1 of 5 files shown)


tylerwilliams (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -1244,72 +1223,66 @@ func cowChunkSizeBytes() int64 {
...
-// hotSwapWorkspace unmounts the workspace drive from a running firecracker
-// container, updates the workspace block device to an ext4 image pointed to
-// by chrootRelativeImagePath, and re-mounts the drive.
-func (c *FirecrackerContainer) hotSwapWorkspace(ctx context.Context, execClient vmxpb.ExecClient) error {
+// updateWorkspaceDriveToEmptyFile updates the workspace device to point to an
+// empty file, which is located at /empty.bin relative to the chroot dir.
+//
+// This lets us avoid saving the workspace contents as part of the VM snapshot,
+// which would be wasteful because the workspace contents aren't needed for the
+// next action to be executed in the VM.
```
maggie-lou, line 1231: nit add: "because we create a clean workspace for each action"

maggie-lou (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go

PR #8075, opened 2024-12-18
In config=remote, use --remote_download_toplevel 

vanja-p:
I had to do a bunch of work to refactor uses of `remote` and `remote-shared` which relied on the fact that remote-shared specified --remote_download_minimal.

Bazel logs a warning if different version of --remote_download_* are specified for the same invocation, so I tried to avoid that.


tempoz (approved):
This looks great; you even cleaned up the repetition in `probers` and `probers-dev` :D

vanja-p:
I think I'm going to wait until after the holidays to merge this. I don't want to break something when people aren't around.

PR #8081, opened 2024-12-18
Save requested and effective timeouts to clickhouse 

vanja-p:
I expect each of these to add about 1% to the write throughput.


server/util/clickhouse/schema/schema.go
```diff
@@ -215,6 +215,9 @@ type Execution struct {
 	RequestedIsolationType string
 	EffectiveIsolationType string `gorm:"type:LowCardinality(String)"` // This values comes from the executor
 
+	RequestedTimeoutMsec int64
+	EffectiveTimeoutMsec int64
```
bduffany, line 219: I think it would be fine to use `usec` for consistency with the other fields - people are mostly using second-level precision for these so I'd expect the trailing digits will mostly be 0 and these should compress well

vanja-p: yeah, I was debating which way we go. I'm not a fan of using microseconds for multiple reasons, but I guess it's better to be consistent.

bduffany (approved):

PR #8094, opened 2024-12-19
Surface slow interrupt warnings in the executor logs

bduffany:
I'd like to understand if these slow interrupt warnings are correlated with higher action failure rates. By logging these we can get a rough frequency chart from GCP then compare that to the action failure metrics in Grafana.

### DIFF
--- a/enterprise/server/remote_execution/containers/firecracker/firecracker.go
+++ b/enterprise/server/remote_execution/containers/firecracker/firecracker.go
@@ -229,7 +229,8 @@ var (
 	vmIdx   int
 	vmIdxMu sync.Mutex
 
-	fatalErrPattern = regexp.MustCompile(`\b` + fatalInitLogPrefix + `(.*)`)
+	fatalErrPattern             = regexp.MustCompile(`\b` + fatalInitLogPrefix + `(.*)`)
+	slowInterruptWarningPattern = regexp.MustCompile(`hrtimer: interrupt took \d+ ns`)
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


tylerwilliams (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
```diff
@@ -2079,6 +2080,12 @@ func (c *FirecrackerContainer) Exec(ctx context.Context, cmd *repb.Command, stdi
 		if err := c.parseSegFault(result); err != nil {
 			log.CtxWarningf(ctx, "Segfault occurred during task execution (recycled=%v) : %s", c.recycled, err)
 		}
+		// Slow hrtimer interrupts can happen during periods of high contention
+		// and may help explain action failures - surface these in the executor
+		// logs.
+		if warning := c.parseSlowInterruptWarning(); warning != "" {
+			log.CtxWarningf(ctx, "Slow interrupt warning reported in kernel logs: %q", warning)
```
maggie-lou, line 2087: nit: We could consolidate some of the logic in the 3 `parseX` functions. We translate the vmlog bytes buffer to a string 3 times in a row

maggie-lou (approved):

enterprise/server/remote_execution/containers/firecracker/firecracker.go
bduffany: Factored out the logic for converting to string and replacing `\r\n` with `\n`

PR #8112, opened 2025-01-04
Redact password portion of `scheme://username:password@domain` (and nothing else)

dan-stowell:
We're redacting harmless strings in workflow logs:
`ERROR: Error computing the main repository mapping: <REDACTED>@3.16.1 depends on <REDACTED>@2.1.1 with compatibility level 2, but <root> depends on <REDACTED>@1.18.0 with compatibility level 1 which is different`

I'd like to only redact high-entropy strings from URLs, but that breaks tests (and likely breaks expectations). @maggie-lou, how paranoid would you suggest being about secrets in URLs?


maggie-lou:
@dan-stowell Loosening the redaction logic should be okay. I think the stricter original logic was intended to prevent us from printing the git remote URL in the logs, which contains users' github tokens and your PR should be sufficient

dan-stowell:
Thanks, @maggie-lou! I've edited the tests to add high entropy strings wherever we expect redaction. I've tagged all the high-entropy strings with `high_entropy_` or `HIGH_ENTROPY_` to give some indication of what's going on.

sluongng (commented):
Im not too sure how cautious we need to be when rolling this out.
A relatively naive approach would be to only log out the matched regex with low entropy, then inspect the log after 1 week to see if we miss something.

Another way to solve this issue, I think, would be to not applying  the redaction to all of workflow logs.
https://github.com/buildbuddy-io/buildbuddy/blob/f458fe6416c613866ada256fd3364e09cb9f6e6c/enterprise/server/cmd/ci_runner/main.go#L938
But only to the `git()` activities in the workflow.

But Im down for having both.

server/util/redact/redact.go
```diff
@@ -100,8 +101,34 @@ var (
...
+	}
+
+	entropy := 0.0
+	length := float64(len(s))
+	for _, count := range freqs {
+		prob := float64(count) / length
+		entropy -= prob * math.Log2(prob)
+	}
+
+	return entropy >= 3.5
```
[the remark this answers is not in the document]
sluongng: The string "rules_swift" is 3.27 though

server/util/redact/redact.go
```diff
@@ -100,8 +101,34 @@ var (
...
+}
+
 func stripURLSecrets(input string) string {
-	return urlSecretRegex.ReplaceAllString(input, "<REDACTED>@")
+	return urlSecretRegex.ReplaceAllStringFunc(input, func(match string) string {
+		if isHighEntropy(match) {
+			return "<REDACTED>@"
+		} else {
+			return "<REDACTED_LOW_ENTROPY_STRING>@"
```
maggie-lou, line 129: I may be misunderstanding @sluongng 's suggestion, but I think he was suggesting that we don't redact low entropy strings but log them to GCP (`log.Warningf`). Then in a week you can manually review the logs to see if there is anything "low entropy" that we actually should've redacted, and adjust the logic accordingly. That way we fix our over-redaction, but ensure we're not loosening our logic too much

sluongng: Yeah, I think that would be a safer approach than rolling out the redaction directly.
Though I want to get Siggi/Brandon thoughts on this because cleaning up secret/credential leaks in GCP log is not the easiest thing either.

---

My second suggestion was to only apply redaction to the "git" logs.
Right now only our Workflow log is redacted with this. The workflow log is composed of several chunks:

```
<git log>
<invocation/step 1>
<invocation/step 2>
...
<invocation/step n>
<git maintenance>
```

Currently, the invocation 1-n logs are usually captured in separate child invocation BES streams and are _not_ being redacted. So a "safe" approach here would be to just redact the log from the git chunks of the workflow. That would prevent folks from using some GIT VERBOSE env var to leak git secrets to the log.

bduffany: In `invocationLog`, where we call `RedactText`, I think we could just instead just manually redact any known secrets (`BUILDBUDDY_API_KEY`, `REPO_TOKEN`), unless @maggie-lou remembers any reasons why this wouldn't work. I could be wrong but get the sense that calculating entropy will lead to false-positives and false-negatives, whereas manual redaction seems a bit more deterministic and we could explain more easily why something was or wasn't redacted.

We also support custom user secrets, which are specified as environment variables. I think we need some way to differentiate these from "safe" environment variables. That seems like it could be done separately though.

dan-stowell: Keeping redaction deterministic seems like the right move. Thanks for the nudge!

@bduffany, sounds like you're proposing gathering a set of secret strings from the environment, then making sure to redact those strings? Seems straightforward. @maggie-lou, any concerns with that approach?

maggie-lou: From [this older PR](https://github.com/buildbuddy-io/buildbuddy/commit/af0c00744e73dce4b1a5bfb43863345b768dc5a8#diff-92078e1505d58633aa1ea334c16f6f63b1649c5f144d753f3356c0ffc3e35391L75), it seems like we originally added the URL redaction code as a broad attempt to redact passwords and API keys. But I don't know if it was in response to a specific incident.

We used to [explicitly expand env vars and print them in workflow logs](https://github.com/buildbuddy-io/buildbuddy/commit/2f80f1e1db1dec17ead3d7b0cbee30568da32b0e#diff-a1778f788b2996ecb914bb66cf733fe40e53a23e1d348e392d9e19db1f1b8d77L1310) (the `expandEnv` part) which made it more likely that we would expose secrets. Now the env vars will only be printed if the user script explicitly prints them. (For example, if $PASSWORD=hi and a user script was `echo $PASSWORD` we used to manually expand the env var and print `echo hi`, whereas now we will just directly copy over `echo $PASSWORD`)

So TLDR I think it should be okay to only redact a set of expected secrets. If a user exposes another password, it will no longer be redacted, but that feels more like user error.

dan-stowell:
Thank you for the archaeology, @maggie-lou.

Pulling back up, `stripURLSecrets` redacts _anything_ before an ampersand: `anystring_before@`.
We could instead look for `scheme://username:password@host:port/path`, or `scheme://anything_before@anything_after`.

I'm inclined to preserve the redaction logic, and log anytime a string would NOT be redacted if we were looking for `scheme://username:password@host:port/path` (but not log the sensitive string itself).

maggie-lou:
> I'm inclined to preserve the redaction logic, and log anytime a string would NOT be redacted if we were looking for `scheme://username:password@host:port/path` (but not log the sensitive string itself).

I like the idea of matching on scheme. Don't feel strongly, but I'd be a proponent of adding ':' to the regex as well, so we don't match URLs like http://maggie@gmail.com, and we only redact the password and not the username

I'm not sure if a log will be actionable though, given that we're incorrectly redacting a lot of ruleset names, so the log will probably fire a lot. Personally I think it is safe to merge your scheme regex change. I don't think a lot of URLs with secrets pop up in build logs, and your regex should cover the case this was originally intended to solve

dan-stowell:
@maggie-lou @sluongng, it's been a journey (have I said that already?), thank you for your patience. This change _should_ only redact the password portion of URLs that contain a username and password.

maggie-lou (approved):
LGTM. Maybe get @sluongng or @siggisim 's approval as well as a final gut check, as they see more build logs than me

siggisim (approved):

PR #8114, opened 2025-01-06
Fix error banner on action page

bduffany:
In #8051 we added an extra execution request which races with another request, which gets canceled. This PR reorders some logic to avoid concurrent requests. We'll now call `WaitExecution` to live-stream the execution only if we first try and fail to fetch the cached `ExecuteResponse`.

### DIFF
--- a/app/invocation/invocation_action_card.tsx
+++ b/app/invocation/invocation_action_card.tsx
@@ -90,9 +90,6 @@ export default class InvocationActionCardComponent extends React.Component<Props
   componentDidMount() {
     this.fetchAction();
     this.fetchExecuteResponseOrActionResult();
-    if (this.props.search.has("executionId")) {
-      this.streamExecution();
-    }
   }
 
   componentDidUpdate(prevProps: Readonly<Props>): void {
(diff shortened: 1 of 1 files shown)


server/buildbuddy_server/buildbuddy_server.go
```diff
@@ -1960,9 +1960,6 @@ func (s *BuildBuddyServer) serveBytestream(ctx context.Context, w http.ResponseW
 		return http.StatusBadRequest, err
 	}
 
-	// TODO(siggisim): Figure out why this JWT is overriding authority auth and remove.
-	ctx = context.WithValue(ctx, "x-buildbuddy-jwt", nil)
```
jdhollen, line 1964: :o

jdhollen (approved):

server/buildbuddy_server/buildbuddy_server.go
bduffany: I split this out into https://github.com/buildbuddy-io/buildbuddy/pull/8115

dan-stowell (approved):

PR #8117, opened 2025-01-06
Add webdriver test for action page

bduffany:
The action page breaks often because the `/file/download` logic is pretty complex as well as the logic in `invocation_action_card.tsx`. So, it seems worth expanding the existing webdriver test to not just check the executions listing for the test build, but also click through to the action page.

To make this change, we need to update the current test to actually execute the build on an executor. Currently when the test build runs, it fails because there are no registered executors. The current test just asserts that the failed execution is displayed in the UI. By running an executor in the test, we can successfully run the execution and click through to the action page to see the action result.

This PR only tests a single auth scenario: the invocation is authenticated, and the user is logged in while trying to view the action page. I would like to expand this test in future PRs with more auth scenarios:
- Authenticated invocation, view while impersonating
- Anonymous invocation, view while logged in + not logged in
- Authenticated invocation made public, view while logged in + not logged in + impersonating

### DIFF
--- a/enterprise/server/test/integration/remote_execution/remote_execution_test.go
+++ b/enterprise/server/test/integration/remote_execution/remote_execution_test.go
@@ -1637,11 +1637,7 @@ func TestRedisRestart(t *testing.T) {
 	}
 	app := buildbuddy_enterprise.RunWithConfig(t, buildbuddy_enterprise.DefaultAppConfig(t), buildbuddy_enterprise.NoAuthConfig, args...)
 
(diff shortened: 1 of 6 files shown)


jdhollen:
looks like the invocation webdriver test is failing

bduffany:
@jdhollen fixed - the ExecuteResponse is cached asynchronously after we return it to bazel, so the action page would occasionally show "not found" since we're navigating there right after the build completed

maggie-lou (approved):

PR #8120, opened 2025-01-06
Make PQ return a bool indicating presence

tylerwilliams:


server/util/priority_queue/priority_queue.go
```diff
@@ -98,23 +98,23 @@ func (pq *PriorityQueue[V]) Push(v V, priority int) {
 	})
```
bduffany, line 98: I think the `noValueGetter` field could be removed now? otherwise LGTM

bduffany (approved):

server/util/priority_queue/priority_queue.go
tylerwilliams: Done.

PR #8125, opened 2025-01-07
Use `os.Getuid()` instead of `user.Current()` if we only want the uid

tempoz:
`os.Getuid` has no failure mode, whereas `user.Current` can fail even though the uid was successfully retrieved. Since we don't care about any part of the user except for the uid, retrieving the other user info is extraneous work and complicates error handling unnecessarily, reporting a warning and short-cicuiting the `isRoot` logic even though we have the uid.

### DIFF
--- a/enterprise/server/util/ociconv/ociconv.go
+++ b/enterprise/server/util/ociconv/ociconv.go
@@ -31,15 +31,11 @@ const (
 )
 
 var (
-	isRoot bool
+	isRoot = os.Getuid() == 0
 
 	// Single-flight group used to dedupe firecracker image conversions.
 	conversionGroup singleflight.Group[string, string]
 )
-
-func init() {
-	isRoot = os.Getuid() == 0
-}
 
 func hashFile(filename string) (string, error) {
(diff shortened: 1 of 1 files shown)


enterprise/server/util/ociconv/ociconv.go
```diff
@@ -39,12 +38,7 @@ var (
 )
 
 func init() {
```
bduffany, line 40: nice! I think this could be moved out of an init function now:

```go
var isRoot = os.Getuid() == 0
```

(GH wouldn't let me make a clickable suggestion here for some reason :thinking: )

bduffany (approved):

enterprise/server/util/ociconv/ociconv.go
tempoz: > (GH wouldn't let me make a clickable suggestion here for some reason 🤔 )

Yeah, github doesn't let you comment (and thus suggest) on lines outside the diff context.

> I think this could be moved out of an init function now:

Good point! fixed.

PR #8126, opened 2025-01-07
Update `go-git` to `v5.13.1` for dependabot

tempoz:
See https://github.com/buildbuddy-io/buildbuddy/security/dependabot/186 and https://github.com/buildbuddy-io/buildbuddy/security/dependabot/187 for details.


sluongng (approved):
LGTM.

AFAIK, this should only affect our Github App (clone template repos, dev only) and our Workflows (run ls-remote from app). So it should be relatively safe and well-tested.

PR #8128, opened 2025-01-07
Fail webdriver tests if error banner is displayed

bduffany:
We generally want to avoid showing error banners to the user. So, we should fail webdriver tests if the error banner is displayed at any point during the test.

This is a follow-up from https://github.com/buildbuddy-io/buildbuddy/pull/8117 - I verified that the new test logic fails with that commit reverted.

We show a "Logged out!" banner when logging out, so a few tests are opted out of this assertion currently. Removing that banner requires some minor auth changes so I split it into a separate PR.

### DIFF
--- a/app/errors/error_service.ts
+++ b/app/errors/error_service.ts
@@ -3,6 +3,7 @@ import { BuildBuddyError, ErrorCode } from "../util/errors";
 import alertService from "../alert/alert_service";
 
 const ERROR_SEARCH_PARAM = "error";
+
 export class ErrorService {
   register() {
     let searchParams = new URLSearchParams(window.location.search);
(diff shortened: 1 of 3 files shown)


app/errors/error_service.ts
```diff
@@ -14,6 +15,10 @@ export class ErrorService {
   }
 
   handleError(e: any, options?: { ignoreErrorCodes: ErrorCode[] }) {
+    // NOTE: keep this string in sync with webtester.go -
+    // web tests check for this message and fail the test if it's logged.
+    console.error(new Error("Displaying error banner"));
+
     console.error(e);
```
jdhollen, line 22: maybe i'm being unrealistic, but do we actually have any errors popping through at all? if not, could we actually just check the error level instead?

bduffany: We do have a lot of errors being logged unfortunately. I think that's a good idea though - it'd be nice to be able to do that instead, eventually.

jdhollen (approved):

PR #8131, opened 2025-01-07
display successful test cases even if they have text content

dan-stowell:
Addressing a customer issue: https://github.com/buildbuddy-io/buildbuddy-internal/issues/4272

Previously, if a successful test case had any child content (including text content), we would not render an individual card for the test case. Successful test cases with Playwright attachments, for example, would not show up.

This change renders cards for all successful test cases. It does not currently render Playwright attachments. I'm not sure if this will turn many test result pages into an ugly sea of logs with dark backgrounds. @jdhollen, do you know?


app/target/target_test_suite.tsx
```diff
@@ -29,7 +29,8 @@ export default class TargetTestSuiteComponent extends React.Component<Props> {
                     <div className="test-case-message">
                       {child.getAttribute("message")} {child.getAttribute("type")}
                     </div>
-                    {!!child.textContent?.trim() && (
+                    // only render text content if it exists and it's not a Playwright attachment
```
jdhollen, line 32: this is a little quirky--the actual node here is a `system-out` node, right?  i think the assumption we were making is that nobody actually uses `system-out` at the `testcase` level, which is funny.  it might be better to actually detect that and treat it in a special way, since i don't think it will have `message` or `type` defined.

jdhollen (commented):
i don't think this will be spammy, but i have a separate comment

app/target/target_test_suite.tsx
jdhollen: i guess really what you would want to be trimming is CDATA? though i'm not even sure the junit schema requires these weird variable definitions like ATTACHMENT to even be in CDATA

PR #8132, opened 2025-01-07
Move JWT creation into a method of Claims

iain-macdonald:
In a later step I'll add support for creating JWTs using different signing methods so that the auth service can use the UserInfo to create JWTs signed using RSA.

### DIFF
--- a/server/util/claims/claims.go
+++ b/server/util/claims/claims.go
@@ -109,6 +109,22 @@ func (c *Claims) IsSAML() bool {
 	return c.SAML
 }
 
+func (c *Claims) AssembleJWT() (string, error) {
+	expirationTime := time.Now().Add(*jwtDuration)
+	expiresAt := expirationTime.Unix()
+	// Round expiration times down to the nearest minute to improve stability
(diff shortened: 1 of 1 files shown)


bduffany (commented):
I'm having trouble understanding the motivation for this PR on its own - I think I could grok this more easily if you bundled this change with the next PR that adds support for more signing methods?

server/util/claims/claims.go
```diff
@@ -109,6 +109,22 @@ func (c *Claims) IsSAML() bool {
 	return c.SAML
 }
 
+func (c *Claims) AssembleJWT() (string, error) {
```
bduffany, line 112: nit: I think this could be a top-level function in the `claims` package - adding it to the struct's public interface limits our ability to restrict the visibility of this function later on

bduffany (approved):

server/util/claims/claims.go
iain-macdonald: That makes this PR just a couple of lines, so I'll just fold it into the next one

PR #8135, opened 2025-01-07
display individual test cases even if they have `<system-out>` or `<system-err>` children

dan-stowell:
(I mucked up the history for #8131, so am starting over).

Addressing a customer issue: https://github.com/buildbuddy-io/buildbuddy-internal/issues/4272

Previously, if a successful test case had any child content (including text content), we would not render an individual card for the test case. Successful test cases with Playwright attachments, for example, would not show up.

@jdhollen suggested handling `<system-out>` and `<system-err>` specially instead of trying to peek into the `<![CDATA[`.


jdhollen (approved):
fwiw, i wouldn't have minded if you just did a force push over the original commit set in your pr, but i also understand if you split it so that you wouldn't accidentally blow away your previous changes :P

PR #8141, opened 2025-01-08
Improve action ID rendering in Cache tab

fmeum:
Bazel attaches a number of special action IDs to cache requests, such as `prefetcher` when it downloads the inputs for an action that is executed locally. Such actions were frequently rendered as a broken link that treated the special action IDs as digests.

This PR ensures that only actual digests are stylized as links and adds a ` (for local execution)` suffix to prefetcher downloads.

### DIFF
--- a/app/invocation/cache_requests_card.tsx
+++ b/app/invocation/cache_requests_card.tsx
@@ -449,16 +449,40 @@ export default class CacheRequestsCardComponent extends React.Component<CacheReq
         renderContent={() => this.renderResultHovercard(result, startTimeMillis)}>
         {(groupTarget === null || groupActionId === null) && (
           <div className="name-column" title={result.targetId ? `${result.targetId} › ${result.actionMnemonic}` : ""}>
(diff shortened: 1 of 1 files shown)


fmeum:
@bduffany I tried to test this locally, but when I run the same build as against the cloud, my cache tab only has the `Cache stats` and `Cache Misses` cards, but not the `Cache requests` card. Is there anything I need to configure to see it?

fmeum:
Thanks, that worked. It looks like this, which isn't great:
image

Should I replace ` (for local execution)` with ` (local)` or instead add an icon with a tooltip?

bduffany:
I like the `(local)` option, and then in the existing hovercard where we render the "prefetcher" action ID, we could maybe render the longer clarification? e.g. `Action ID: prefetcher (for local execution)`

fmeum:
Done. I also switched to `===` throughout.

bduffany (approved):

PR #8148, opened 2025-01-08
when displaying build errors, remove ANSI escape code that clears to …

dan-stowell:
…end of line


app/invocation/invocation_error_card.tsx
```diff
@@ -224,6 +225,10 @@ function deemphasizeSandboxDebug(text: string): string {
   return text.replaceAll(debugMessage, `\x1b[90m${debugMessage}\x1b[0m\n \n`);
 }
 
+function removeAnsiClearToEndOfLine(text: string): string {
+  return text.replaceAll(/\u001B\[K/g, "");
```
bduffany, line 229: I wonder if we should try to interpret these escape codes the same way that the terminal would, rather than deleting them? Could maybe add a TODO here to look into this, since it's probably more trouble than it's worth for now, but might lead to a better UX that more closely matches what the user would see in their terminal.

bduffany (approved):

app/invocation/invocation_error_card.tsx
dan-stowell: Good call! I'll see if there's a straightforward way to do that.

dan-stowell: Thanks for the nudge, @bduffany! We already have a `parseAnsi()` function that was dropping the rest of the line after encountering an unexpected escape code. Fixed that, left a TODO to handle cursor escape sequences (sometime, not sure when it would make sense).

app/terminal/ansi.tsx
```diff
@@ -119,6 +131,13 @@ export default function parseAnsi(text: string): AnsiTextSpan[] {
         code = "";
         continue;
       }
+      if (cursorEscapeCharacters.includes(char)) {
+        //TODO: handle ANSI cursor escape sequences
+        //for now, treat as a no-op
```
bduffany, line 136: nit: add space after `//`

```suggestion
        // TODO: handle ANSI cursor escape sequences
        // For now, treat as a no-op
```

bduffany (approved):

app/invocation/invocation_error_card.tsx
bduffany: nice! this seems better

PR #8152, opened 2025-01-09
Add region and self-hosted columns to Executions table

bduffany:
Not populating them yet, just adding to the schema.

### DIFF
--- a/server/util/clickhouse/schema/schema.go
+++ b/server/util/clickhouse/schema/schema.go
@@ -157,6 +157,10 @@ type Execution struct {
 	UserID             string
 	Worker             string
 
+	// Executor metadata
+	SelfHosted bool
+	Region     string `gorm:"type:LowCardinality(String)"`
+
 	Stage int64
 
 	// RequestMetadata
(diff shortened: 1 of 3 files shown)


tylerwilliams (approved):
LGTM but added @luluz66 resident clickhouse expert

server/util/clickhouse/schema/schema.go
```diff
@@ -157,6 +157,10 @@ type Execution struct {
 	UserID             string
 	Worker             string
 
+	// Executor metadata
+	SelfHosted bool
+	Region     string `gorm:"type:LowCardinality(String)"`
```
luluz66, line 162: just want to confirm this gorm attributes actually works. I recalled certain gorm attributes with clickhouse don't work in the past.

luluz66 (commented):
lgtm; just one question.

server/util/clickhouse/schema/schema.go
bduffany: yep, double-checked in dev - the `LowCardinality` option is applied to the `effective_isolation_type` column.

luluz66 (approved):

PR #8153, opened 2025-01-10
Revamp of the fuse workspace implementation:

vadimberezniker:
- We no longer rely on the go-fuse library to maintain the directory structure which allows us to store a single reprsentation of the structure in memory.

 - Lookups and operations are now handled via inode numbers, eliminating a lot of unnecessary string allocations.

 - We open files directly from the file cache when they are present.

 - We bypass the gRPC library when both the client and server are runnign in the same process.


enterprise/server/remote_execution/filecache/filecache.go
```diff
@@ -288,6 +288,29 @@ func (c *fileCache) FastLinkFile(ctx context.Context, node *repb.FileNode, outpu
 	return true
 }
 
+func (c *fileCache) Open(ctx context.Context, node *repb.FileNode) (f *os.File, err error) {
```
tylerwilliams, line 291: We've found the named return params to be a little error-prone in the past.

enterprise/server/remote_execution/vfs/vfs_unix.go
```diff
@@ -130,10 +130,11 @@ func (vfs *VFS) Mount() error {
 			// Don't depend on `fusermount`.
 			// Disable fallback to fusermount as well, since it can cause
 			// deadlocks. See https://github.com/hanwen/go-fuse/issues/506
-			DirectMountStrict: true,
-			FsName:            "bbvfs",
-			MaxWrite:          fuse.MAX_KERNEL_WRITE,
-			EnableLocks:       true,
+			DirectMountStrict:  true,
+			DisableReadDirPlus: true,
```
tylerwilliams, line 134: comment would be useful for this one

enterprise/server/remote_execution/vfs/vfs_unix.go
```diff
@@ -551,14 +544,55 @@ func (n *Node) Open(ctx context.Context, flags uint32) (fh fs.FileHandle, fuseFl
...
+		}
+		if int(flags)&os.O_TRUNC != 0 {
+			textFlags = append(textFlags, "O_TRUNC")
+		}
+		if int(flags)&os.O_EXCL != 0 {
+			textFlags = append(textFlags, "O_EXCL")
+		}
+		if int(flags)&os.O_SYNC != 0 {
+			textFlags = append(textFlags, "O_SYNC")
+		}
```
tylerwilliams, line 571: could move this to a func?

enterprise/server/util/vfs_server/vfs_server.go
```diff
@@ -151,189 +271,148 @@ type Server struct {
...
-				Name: name,
-				Attrs: &vfspb.Attrs{
-					Size:      lazyFile.Size,
-					Perm:      uint32(lazyFile.Perms),
-					Immutable: true,
-				},
-			}
-			parent.Files = append(parent.Files, fe)
+		for _, childFileNode := range dir.GetFiles() {
+			log.Infof("%q %t", childFileNode.GetName(), childFileNode.IsExecutable)
```
tylerwilliams, line 344: you want to keep this?

tylerwilliams (approved):

enterprise/server/remote_execution/filecache/filecache.go
vadimberezniker: It's consistent with the other filecache functions so I'm going to keep this one as well. If we change this we should do this for everything in this file.

enterprise/server/remote_execution/vfs/vfs_unix.go
vadimberezniker: I removed this. There's no perf difference now that we implement `Lookup` ourselves.

enterprise/server/remote_execution/vfs/vfs_unix.go
vadimberezniker: Done

PR #8155, opened 2025-01-10
Use `flags-as-proto` to determine command line schema (take two)

tempoz:
We now use `bazel help flags-as-proto` as the primary method of determining the command line schema, only falling back to the usage-based parsing if the bazel version is too old to support `flags-as-proto`.

Switched from using `BoolLike` to the more-specific `RequiresValue` and `HasNegative`, as those are what is present in the `FlagCollection` proto. They also allow us to tell the difference between boolean options (or boolean/enum options) and expansion options, which allows us more precision and accuracy in parsing: we can now remove values specified via `=` to expansion options (which improves our canonicalization).

Added a test to confirm that `flags-as-proto` allows us to correctly handle undocumented options that are specified via `common` in a `bazelrc` file (which was the primary motivation for this change).

Added a mock help function in the tests that uses `flags-as-proto` help and tested it everywhere the usage-style help was tested to ensure there were no regressions.

Added some options to the canonicalization tests to ensure we are handling expansion options and options that use `BoolOrEnumConverter` in bazel (`--subcommands` and `--experimental_convenience_symlinks`) correctly.

Added a test to ensure the options produced through `flags-as-proto` match up with those produced by parsing usage.


tempoz:
Looks like I need to fix the CLI integration tests

bduffany (approved):

PR #8156, opened 2025-01-10
Make cpuleaser numa-aware

tylerwilliams:
Make the CPU Leaser numa aware (on linux).

The leaser will prefer to assign a set of processors from the least loaded numa node. It will never return processors from multiple numa nodes in the same lease -- if more processors are requested than are available on a single node, it will return all the processors from the least loaded node.

Obligatory https://www.youtube.com/watch?v=Cqd1Gvq-RBY

### DIFF
--- a/enterprise/server/util/cpuset/cpuset_test.go
+++ b/enterprise/server/util/cpuset/cpuset_test.go
@@ -43,7 +43,7 @@ func TestEvenDistributionUnderLoad(t *testing.T) {
 	counts := make(map[int]int, 4)
 	for i := 0; i < 100; i++ {
 		task := uuid.New()
-		cpus, cancel := cs.Acquire(3000, task)
+		cpus, cancel := cs.Acquire(1000, task)
 		for _, cpu := range cpus {
 			counts[cpu] += 1
 		}
@@ -51,7 +51,7 @@ func TestEvenDistributionUnderLoad(t *testing.T) {
 		defer cancel()
 	}
 	for _, count := range counts {
-		assert.Equal(t, 75, count)
+		assert.Equal(t, 25, count)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


enterprise/server/util/cpuset/cpuset_test.go
```diff
@@ -104,3 +104,60 @@ func TestCPUSetDisabledManualCPUSet(t *testing.T) {
 	defer cancel1()
 	assert.Equal(t, []int{0, 1, 3}, cpus1)
 }
+
+func TestNumaNodeFairness(t *testing.T) {
+	flags.Set(t, "executor.cpu_leaser.enable", true)
+	flags.Set(t, "executor.cpu_leaser.overhead", 0)
+	flags.Set(t, "executor.cpu_leaser.min_overhead", 0)
+	flags.Set(t, "executor.cpu_leaser.cpuset", "0:0-3,1:4-7")
```
bduffany, line 112: If the parser can handle it, shuffling up the cpu indexes a bit more would give us higher confidence that the NUMA node is being taken into account, instead of e.g. just returning the first N cpus. (Same for the tests below)

```suggestion
	flags.Set(t, "executor.cpu_leaser.cpuset", "0:0-1,1:4-7,0:2-3")
```

enterprise/server/util/cpuset/cpuset_test.go
```diff
@@ -104,3 +104,60 @@ func TestCPUSetDisabledManualCPUSet(t *testing.T) {
...
+	cpus1, cancel1 := cs.Acquire(4000, task1)
+	defer cancel1()
+	allCPUs = append(allCPUs, cpus1...)
+
+	task2 := uuid.New()
+	cpus2, cancel2 := cs.Acquire(4000, task2)
+	defer cancel2()
+	allCPUs = append(allCPUs, cpus2...)
+
+	assert.ElementsMatch(t, []int{0, 1, 2, 3, 4, 5, 6, 7}, allCPUs)
```
bduffany, line 128: probably worth adding a third task and checking that the CPUs are on the same node

bduffany (approved):

enterprise/server/util/cpuset/cpuset_test.go
tylerwilliams: Done. (yuck!)

enterprise/server/util/cpuset/cpuset_test.go
tylerwilliams: Done.

bduffany (approved):

PR #8157, opened 2025-01-10
Set action mnemonic for workflows and remote bazel

bduffany:
This will allow us to more easily distinguish between workflows, remote bazel, and "normal" bazel actions when looking at executions in clickhouse.

### DIFF
--- a/enterprise/server/hostedrunner/hostedrunner.go
+++ b/enterprise/server/hostedrunner/hostedrunner.go
@@ -377,7 +377,10 @@ func (r *runnerService) Run(ctx context.Context, req *rnpb.RunRequest) (*rnpb.Ru
 	}
 	log.Debugf("Uploaded runner action to cache. Digest: %s/%d", actionDigest.GetHash(), actionDigest.GetSizeBytes())
 
-	execCtx, err := bazel_request.WithRequestMetadata(ctx, &repb.RequestMetadata{ToolInvocationId: invocationID})
+	execCtx, err := bazel_request.WithRequestMetadata(ctx, &repb.RequestMetadata{
+		ToolInvocationId: invocationID,
+		ActionMnemonic:   "RemoteBazelRun",
+	})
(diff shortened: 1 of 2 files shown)


vanja-p (approved):
This seems ok to me, but note that the `command` field can already be used for the same purpose. That's usually the bazel command. For remote bazel, it's the bazel command, prefixed by `remote `. For workflows, it's `workflow run`.

bduffany:
@vanja-p ah nice, I didn't think of that. I would like to run some queries that segment by action mnemonic and think it would be convenient to identify these more easily in that case. These actions are relatively rare compared to other actions, so I think it doesn't hurt to set the mnemonic.

enterprise/server/hostedrunner/hostedrunner.go
```diff
@@ -377,7 +377,10 @@ func (r *runnerService) Run(ctx context.Context, req *rnpb.RunRequest) (*rnpb.Ru
 	}
 	log.Debugf("Uploaded runner action to cache. Digest: %s/%d", actionDigest.GetHash(), actionDigest.GetSizeBytes())
 
-	execCtx, err := bazel_request.WithRequestMetadata(ctx, &repb.RequestMetadata{ToolInvocationId: invocationID})
+	execCtx, err := bazel_request.WithRequestMetadata(ctx, &repb.RequestMetadata{
+		ToolInvocationId: invocationID,
+		ActionMnemonic:   "RemoteBazelRun",
```
vanja-p, line 382: could this specify the bazel command? (build, run, test, etc.)

bduffany: I think we'd have to parse the action arguments to figure that out, which seems a bit hairy. I think it's fine for this to be coarse grained and then we can use the `Command` field to differentiate build/run etc.

maggie-lou (approved):

PR #8159, opened 2025-01-11
Return numa node in cpuset

tylerwilliams:


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -656,6 +642,13 @@ func (c *ociContainer) doWithStatsTracking(ctx context.Context, invokeRuntimeFn
 }
 
 func (c *ociContainer) setupCgroup(ctx context.Context) error {
+	// Lease CPUs for task execution, and set cleanup function.
+	leaseID := uuid.New()
+	_, leasedCPUs, cleanupFunc := c.env.GetCPULeaser().Acquire(c.milliCPU, leaseID)
```
bduffany, line 647: I think we could use the returned node here to set `cpuset.mems`? (maybe in a later PR)

bduffany (approved):

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
tylerwilliams: yeah, agreed.

PR #8169, opened 2025-01-14
[raft] refactor zombie clean up code

luluz66:
https://docs.google.com/document/d/1Vk5TqjFYykkmgtXTOmygtN9HI0fP0zYd6WHd8wOa6bo/edit?tab=t.0

### DIFF
--- a/enterprise/server/raft/store/store.go
+++ b/enterprise/server/raft/store/store.go
@@ -582,9 +582,7 @@ func (s *Store) Start() error {
 		return nil
 	})
 	s.eg.Go(func() error {
-		if s.replicaJanitor != nil {
-			s.replicaJanitor.Start(s.egCtx)
-		}
+		s.replicaJanitor.Start(s.egCtx)
 		return nil
 	})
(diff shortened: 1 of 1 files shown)


enterprise/server/raft/store/store.go
```diff
@@ -271,6 +274,9 @@ func NewWithArgs(env environment.Env, rootDir string, nodeHost *dragonboat.NodeH
 		s.driverQueue = driver.NewQueue(s, gossipManager, nhLog, apiClient, clock)
 	}
 	s.deleteSessionWorker = newDeleteSessionsWorker(clock, s)
+	if *zombieNodeScanInterval != 0 {
+		s.replicaJanitor = newReplicaJanitor(clock, s)
+	}
```
tylerwilliams, line 279: It can be a little error prone to have stuff only set sometimes -- easy to hit NPEs. Would it be clearer to always set the janitor but just have it not do anything if *zombieNodeScanInterval == 0?

tylerwilliams (approved):

PR #8172, opened 2025-01-14
Add a mechanism to disable automatic retries for workflows and remote bazel

maggie-lou:
This PR adds a platform property to disable automatic retries by the scheduler. This can be used to ensure non-idempotent tasks are not retried.

For example, someone using remote bazel for a deploy may not want automatic retries

Pre-req to https://github.com/buildbuddy-io/buildbuddy/pull/8195

### DIFF
--- a/cli/remotebazel/remotebazel.go
+++ b/cli/remotebazel/remotebazel.go
@@ -68,6 +68,8 @@ const (
 	// `git remote` output is expected to look like:
 	// `origin	git@github.com:buildbuddy-io/buildbuddy.git (fetch)`
 	gitRemoteRegex = `(.+)\s+(.+)\s+\((push|fetch)\)`
+
+	maxRetries = 5
 )
 
 var (
@@ -86,6 +88,7 @@ var (
 	runFromBranch           = RemoteFlagset.String("run_from_branch", "", "A GitHub branch to base the remote run off. If unset, the remote workspace will mirror your local workspace.")
 	runFromCommit           = RemoteFlagset.String("run_from_commit", "", "A GitHub commit SHA to base the remote run off. If unset, the remote workspace will mirror your local workspace.")
 	script                  = RemoteFlagset.String("script", "", "Shell code to run remotely instead of a Bazel command.")
+	disableRetry            = RemoteFlagset.Bool("disable_retry", false, "By default, transient errors are automatically retried. This behavior can be disabled, if a command is non-idempotent for example.")
 )
 
 func consoleCursorMoveUp(y int) {
@@ -847,6 +850,9 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
 		ExecProperties: platform.Properties,
 		RemoteHeaders:  *remoteHeaders,
 		RunRemotely:    *runRemotely,
+		// In order to detect and notify on retry, this client will implement
+		// retry behavior itself. Direct the server to not retry.
+		DisableRetry: false,
 		Steps: []*rnpb.Step{
 			{
 				Run: opts.Command,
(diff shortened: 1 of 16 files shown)


cli/remotebazel/remotebazel.go
```diff
@@ -1273,3 +1298,12 @@ func getAPIKeyFromConfig() (string, error) {
...
 }
+
+// isTaskMisconfigured returns whether a task failed to execute because of a
+// configuration error that will prevent the action from executing properly,
+// even if retried.
+func isTaskMisconfigured(err error) bool {
+	return status.IsInvalidArgumentError(err) ||
+		status.IsFailedPreconditionError(err) ||
+		status.IsUnauthenticatedError(err)
+}
```
bduffany, line 1309: nit: leave a note both here and in `executor.go` to keep the 2 implementations in sync?

enterprise/server/remote_execution/executor/executor.go
```diff
@@ -158,7 +158,16 @@ func isClientBazel(task *repb.ExecutionTask) bool {
...
 
+func retriesEnabled(task *repb.ExecutionTask) bool {
+	v := platform.FindValue(task.GetPlatformOverrides(), platform.ShouldRetryPropertyName)
+	return v == "true" || v == ""
+}
+
 func shouldRetry(task *repb.ExecutionTask, taskError error) bool {
+	if !retriesEnabled(task) {
+		return false
+	}
```
bduffany, line 169: I think the scheduler server also has some retry logic, e.g. if the executor crashes before we hit this codepath, the scheduler still probably needs some way for the task to be attempted again. Have you looked at that / is there anything we need to change there?

cli/remotebazel/remotebazel.go
```diff
@@ -866,93 +872,112 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
...
+			return 1, status.UnknownErrorf("error running bazel: %s", err)
+		}
+
+		iid := rsp.GetInvocationId()
+		log.Debugf("Invocation ID: %s", iid)
+
+		// If the remote bazel process is canceled or killed, cancel the remote run
+		isInvocationRunning := true
+		go func() {
+			<-ctx.Done()
```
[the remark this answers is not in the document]
maggie-lou: This logic is in a goroutine because it's listening in the background if the parent context is canceled (like if you CTRL-C the remote bazel command)

I added a child context that is canceled after the current retry attempt, so the goroutine is cleaned up after each attempt

cli/remotebazel/remotebazel.go
```diff
@@ -866,93 +872,112 @@ func Run(ctx context.Context, opts RunOpts, repoConfig *RepoConfig) (int, error)
...
 	log.Printf("\nWaiting for available remote runner...\n")
-	rsp, err := bbClient.Run(ctx, req)
-	if err != nil {
-		return 1, status.UnknownErrorf("error running bazel: %s", err)
-	}
 
-	iid := rsp.GetInvocationId()
-	log.Debugf("Invocation ID: %s", iid)
+	retry := !*disableRetry
+	retryCount := 1
```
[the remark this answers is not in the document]
maggie-lou: Fixed

cli/remotebazel/remotebazel.go
maggie-lou: Moved the logic to a shared until func

maggie-lou:
I also moved the CLI code into a separate PR, because it has to go out after the server and executors have rolled out: https://github.com/buildbuddy-io/buildbuddy/pull/8195

enterprise/server/remote_execution/executor/executor.go
maggie-lou: Good call - added some logic to address that

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -96,6 +96,7 @@ const (
 	IncludeSecretsPropertyName           = "include-secrets"
 	DefaultTimeoutPropertyName           = "default-timeout"
 	TerminationGracePeriodPropertyName   = "termination-grace-period"
+	ShouldRetryPropertyName              = "should-retry"
```
iain-macdonald, line 99: WDYT of renaming to just "retry"?

enterprise/server/remote_execution/platform/platform.go
```diff
@@ -768,3 +778,8 @@ func IsCICommand(cmd *repb.Command, platform *repb.Platform) bool {
 	}
 	return false
 }
+
+func RetriesEnabled(task *repb.ExecutionTask) bool {
```
iain-macdonald, line 782: nit: rename to Retryable?

server/util/status/status.go
```diff
@@ -263,3 +263,11 @@ func MetricsLabel(err error) string {
 	// errors).
 	return status.Code(err).String()
 }
+
+// IsTaskMisconfigured returns whether the error is a configuration error
+// that will persist, even despite retries.
+func IsTaskMisconfigured(err error) bool {
```
iain-macdonald, line 269: nit: rename to Retryable?

iain-macdonald (approved):

enterprise/server/remote_execution/platform/platform.go
maggie-lou: Done

enterprise/server/remote_execution/platform/platform.go
maggie-lou: Done

server/util/status/status.go
```diff
@@ -263,3 +263,12 @@ func MetricsLabel(err error) string {
 	// errors).
 	return status.Code(err).String()
 }
+
+// Retryable returns false if the error is a configuration error
+// that will persist, even despite retries.
+func Retryable(err error) bool {
```
bduffany, line 269: this logic is a bit specific to RBE, e.g. we probably wouldn't want to use this to decide whether a cache RPC should get retried - might make sense to put it in the `rexec` package

bduffany (approved):
lgtm, just 1 nit. Would be good to get another review for the scheduler changes, since that code is a bit tricky

maggie-lou:
> lgtm, just 1 nit. Would be good to get another review for the scheduler changes, since that code is a bit tricky

👍 Added Iain because he's worked on the scheduler before. I'll also wait to merge until next week in case anything goes wrong during the release

PR #8174, opened 2025-01-14
ociruntime: handle container sigkill

sluongng:
wip

### DIFF
--- a/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
+++ b/enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
@@ -72,6 +72,9 @@ const (
 
 	// Maximum length of overlayfs mount options string.
 	maxMntOptsLength = 4095
+
+	ociSIGKILLExitCode     = int(128 + unix.SIGKILL)
+	ociExecSIGTERMExitCode = int(128 + unix.SIGTERM)
 )
 
 //go:embed seccomp.json
@@ -297,6 +300,9 @@ type ociContainer struct {
 	forceRoot      bool
 
 	milliCPU int64 // milliCPU allocation from task size
+
+	mu      sync.Mutex
+	removed bool
 }
 
 // Returns the OCI bundle directory for the container.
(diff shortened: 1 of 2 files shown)


enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -504,9 +516,19 @@ func (c *ociContainer) Exec(ctx context.Context, cmd *repb.Command, stdio *inter
...
 
-	return c.doWithStatsTracking(ctx, func(ctx context.Context) *interfaces.CommandResult {
+	res := c.doWithStatsTracking(ctx, func(ctx context.Context) *interfaces.CommandResult {
 		return c.invokeRuntime(ctx, cmd, stdio, 1*time.Microsecond, args...)
 	})
+
+	c.mu.Lock()
+	removed := c.removed
+	c.mu.Unlock()
+	if removed && (res.ExitCode == ociSIGKILLExitCode || res.ExitCode == ociExecSIGTERMExitCode) {
```
bduffany, line 526: in practice, I think `Exec()` and `Remove()` can't be called concurrently - `TracedCommandContainer` (which wraps all container instances) takes an exclusive lock before it calls `Remove()`

enterprise/server/remote_execution/containers/ociruntime/ociruntime.go
```diff
@@ -440,9 +446,15 @@ func (c *ociContainer) Run(ctx context.Context, cmd *repb.Command, workDir strin
...
 	}
 
-	return c.doWithStatsTracking(ctx, func(ctx context.Context) *interfaces.CommandResult {
+	res := c.doWithStatsTracking(ctx, func(ctx context.Context) *interfaces.CommandResult {
 		return c.invokeRuntime(ctx, nil /*=cmd*/, &interfaces.Stdio{}, 0 /*=waitDelay*/, "run", "--bundle="+c.bundlePath(), c.cid)
 	})
+	if res.ExitCode == ociSIGKILLExitCode || res.ExitCode == ociExecSIGTERMExitCode {
+		res.ExitCode = commandutil.KilledExitCode
+		res.Error = commandutil.ErrSIGKILL
+	}
```
bduffany, line 455: I think there is a chance this will break people if they are already returning these exit codes for another reason.

Is there anything logged by `crun` that we could maybe use to distinguish these scenarios?

sluongng: no, it's pretty barebone :(

https://github.com/containers/crun/blob/283ffdf7dc0e92f61d78df362f08e42ba11efa7a/src/libcrun/utils.h#L415-L424

sluongng:
Don't have a good solution to this just yet. Created an internal tracking issue for this. Let's discuss the potential solutions there.

PR #8178, opened 2025-01-14
Populate self-hosted and region fields in ClickHouse

bduffany:
Tested locally with self-hosted executors and shared executors, and also checked the region column.


enterprise/server/remote_execution/operation/operation.go
```diff
@@ -15,13 +16,18 @@ import (
...
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/types/known/anypb"
 
 	repb "github.com/buildbuddy-io/buildbuddy/proto/remote_execution"
 	gstatus "google.golang.org/grpc/status"
 	tspb "google.golang.org/protobuf/types/known/timestamppb"
 )
 
+var (
+	region = flag.String("executor.region", "", "Region metadata associated with executions.")
```
vanja-p, line 28: operation.go is used by both the app and the executor, so it's a bit weird for this flag to live here, with the name prefixed by `executor. Publish() is only used by priority_task_scheduler.go, so maybe that that's the right place for the flag, and for adding the region to the context.

bduffany: good point - done

vanja-p (approved):

vadimberezniker (approved):

